PR #746, opened 2021-10-06
[fix] Update set-value CDK dependency

rcoh:
- Run `npm update`
- verified that this pulls in a non-broken version of set-value


jdisanti (approved):
Pulled down changes locally and reviewed.

PR #759, opened 2021-10-09
Add RFC for publishing to crates.io

jdisanti:
This is an RFC that proposes how publishing to crates.io should work in both the short-term and the long-term.

[Rendered.](https://github.com/jdisanti/smithy-rs/blob/rfc-crates-io/design/src/rfcs/rfc0006_crates_io_publishing.md)


design/src/rfcs/rfc0006_crates_io_publishing.md
```diff
@@ -0,0 +1,269 @@
...
+Concrete Scenarios
+------------------
+
+The following are example scenarios that may occur that the publishing implementation must handle.
+
+1. **AWS models were updated**: This results in `patch` version bumps for all AWS SDK crates _that had changes_.
+2. **Bug fixed in runtime crate**: Runtime crate is `minor` version bumped. All runtime crates that depend on it are
+   also `minor` version bumped, and all of the AWS SDK crates are `minor` version bumped.
+3. **Bug fixed in codegen**: All AWS SDK crates are `minor` version bumped.
+4. **CVE discovered in external dependency**: If the external dependency is fixed in a `patch` revision where it will
```
rcoh, line 84: depending on severity it seems like this may necessitate yanking as well?

design/src/rfcs/rfc0006_crates_io_publishing.md
```diff
@@ -0,0 +1,269 @@
...
+   automatically upgrade via semver, then no action is required. Otherwise, update the dependency where it is
+   included, and then follow the normal version bump process from there.
+5. **CVE discovered in runtime crate**: If it can be fixed in a backwards compatible way, then `patch` revision the
+   runtime crate with the fix. Otherwise, `minor` version bump the runtime crate and follow the normal process from there.
+6. **CVE discovered in generated code**: Same process as a bug fix in codegen.
+7. **Accidental release of breaking change**: Run publish process in reverse, yanking the published packages.
+8. **Buggy update to runtime crate**: If possible, roll forward with a `patch` revision. Otherwise, run publish process
+   in reverse yanking the published packages.
+
+Short-term Proposal
```
rcoh, line 94: a section with the estimated effort in dev weeks for the short term and long term proposal would probably be helpful

design/src/rfcs/rfc0006_crates_io_publishing.md
```diff
@@ -0,0 +1,269 @@
...
+4. Wait for GitHub actions to generate AWS SDK using newly released smithy-rs
+5. Check out aws-sdk-rust, delete existing SDK code, unzip generated SDK in place, and update readme
+6. Tag aws-sdk-rust
+
+To keep things simple:
+- The Smithy runtime crates will have the same smithy-rs version
+- All AWS crates will have the same AWS SDK version
+- There will _NOT_ be `patch` revisions
+
+All runtime crate version numbers in smithy-rs will be locked at `0.1.0-smithy-rs-head`. This is a fake
```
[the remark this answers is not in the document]
jdisanti: It has to have the number or Cargo can't parse it. I'll set it to `0.0.0-smithy-rs-head`.

design/src/rfcs/rfc0006_crates_io_publishing.md
jdisanti: I think this design allows for yanking old vulnerable versions if we decide to do it since we could retrieve old version manifest files out of git history and generate yank commands from them.

design/src/rfcs/rfc0006_crates_io_publishing.md
```diff
@@ -0,0 +1,279 @@
...
+- All AWS crates will have the same AWS SDK version
+- There will _NOT_ be `patch` revisions
+
+All runtime crate version numbers in smithy-rs will be locked at `0.0.0-smithy-rs-head`. This is a fake
+version number that gets replaced when generating the SDK.
+
+The SDK generator script in smithy-rs will be updated to:
+- Replace Smithy runtime crate versions with the smithy-rs version from `aws/sdk/gradle.properties`
+- Replace AWS runtime crate versions with AWS SDK version from `aws/sdk/gradle.properties`
+- Add correct version numbers to all path dependencies in all the final crates that end up in the build artifacts
```
jdisanti, line 119: Now that I'm implementing this, I think I'm going to move this last step into the publish script since that will be easier. We need to do these kinds of updates in the long-term solution anyway, so that should be future proof.

crisidev (approved):
Nice RFC! LGTM!

design/src/rfcs/rfc0006_crates_io_publishing.md
```diff
@@ -0,0 +1,279 @@
...
+
+Cargo uses [semver](https://github.com/dtolnay/semver#requirements) for versioning,
+with a `major.minor.patch-pre` format:
+- `major`: Incompatible API changes
+- `minor`: Added functionality in backwards compatible manner
+- `patch`: Backwards compatible bug fixes
+- `pre`: Pre-release version tag (omitted for normal releases)
+
+The Rust SDK will make one small adjustment to the above requirements: the `patch` version will be incremented
+for backwards compatible AWS model changes, and the `minor` version will be incremented for changes in Smithy runtime
```
carllerche, line 34: Define "change" Is the change backwards compatible? Minor versions cannot be backwards incompatible.

design/src/rfcs/rfc0006_crates_io_publishing.md
```diff
@@ -0,0 +1,269 @@
...
+The latest version of `aws-sdk-s3` will always have the same `major.minor` version as the latest `aws-sdk-dynamodb`,
+for example. The `patch` version is allowed to be different between service crates.
+
+Smithy runtime crates will maintain their own versioning, and their version numbers are independent of each other.
+For example, the latest `smithy-types` could be at `0.3` while `smithy-json` is at `0.2`. The `aws-config` crate
+is an exception to this rule since customers depend on it directly. It should maintain the same `major.minor` as
+the AWS SDK crates.
+
+The `pre` version tag will be `alpha` during the Rust SDK alpha, `preview` during dev preview, and removed once the
+SDK is generally available.
```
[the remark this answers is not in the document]
carllerche: This can work. Another option would be to go 1.0 when GA and 0.* before that. How do you see the pros / cons of this approach?

design/src/rfcs/rfc0006_crates_io_publishing.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Version Bumping
+
+In general, model changes will be frequent (daily) while changes to runtime crates and code generation
+will be infrequent (weekly to monthly). Model changes are supposed to _always_ be backwards compatible, so it
+makes sense to apply these to `patch` revisions as much as possible. That way, if a model change was made only
```
carllerche, line 54: I don't see why this makes sense. The argument is that it is to avoid updating all 260 crates, but then maybe service crates shouldn't have their minor versions synced?

design/src/rfcs/rfc0006_crates_io_publishing.md
```diff
@@ -0,0 +1,279 @@
...
+
+Concrete Scenarios
+------------------
+
+The following are example scenarios that may occur that the publishing implementation must handle.
+
+1. **AWS models were updated**: This results in `patch` version bumps for all AWS SDK crates _that had changes_.
+2. **Bug fixed in runtime crate**: Runtime crate is `minor` version bumped. All runtime crates that depend on it are
+   also `minor` version bumped, and all of the AWS SDK crates are `minor` version bumped.
+3. **Bug fixed in codegen**: All AWS SDK crates are `minor` version bumped.
```
carllerche, line 83: Same, why here?

design/src/rfcs/rfc0006_crates_io_publishing.md
carllerche: Tokio takes the approach that only significant bugs are fixed in patch releases, otherwise they are included in the next minor release. We release minor versions regularly, so it tends to be OK. We take the point of view that patch releases should be as stable as possible, so only introduce a patch release to fix important things.

jdisanti: The only changes we would allow into a patch release are generated code from an updated service model. For example, if an AWS service adds a new API call, we would have some automation that pulls in the model update for that new API call and then generates a new SDK with it. This should only introduce new bugs if there is a bug in the code generator that gets exercised by the updated model (which is definitely possible).

The idea was that this kind of change could be a patch release since the model update is required to be backwards compatible, and thus, the code generated from it will also be backwards compatible. In reality, model updates aren't backwards compatible 100% of the time though, so an argument for putting them in minor releases rather than patches is valid.

Keeping the major/minor consistent across the SDK crates has the nice benefit that it's trivial to know that the correct versions are being used in a project. If we move model updates into minor, then I think we lose that unless we publish minor versions with no changes, or maintain some kind of version reference table for each SDK release.

jdisanti:
Carl raises strong arguments for changing how the minor/patch versions operate, and I'm leaning more towards including model updates in minor versions rather than patch versions now. I think we might be able to automate something in CI to verify a patch bump in a runtime crate works with the previously released SDK versions too, which would enable us to do bug fixes in the runtime crates without a minor version bump in anything.

Before updating the doc, I think we need to decide if we want to try and keep SDK crate major/minor version numbers consistent or not.

jdisanti:
I'm going to split up this PR since the short-term solution is already implemented and not controversial. I will start a new RFC for the long-term versioning and publishing approach that incorporates the feedback from this PR.

PR #766, opened 2021-10-14
Generate service READMEs and fix module descriptions

### LINKED ISSUE #754: Generated module description contains HTML
The code generated `Cargo.toml` description field is currently defaulted to the `@documentation` trait on the ServiceShape being generated: https://github.com/awslabs/smithy-rs/blob/7e477cd95f5079b0b4c6e0bda9358cfba0cc3267/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RustSettings.kt#L95-L96

Used here: https://github.com/awslabs/smithy-rs/blob/7e477cd95f5079b0b4c6e0bda9358cfba0cc3267/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/CargoTomlGenerator.kt#L30

For the AWS models, this documentation trait contains HTML. For example, DynamoDB has:
https://github.com/awslabs/smithy-rs/blob/14f435e864e20dcfe401d39e456e363e12030b94/aws/sdk/aws-models/dynamodb.json#L3196

For the AWS codegen, we should strip HTML and probably only take the first paragraph of text for the description that will end up on crates.io.

S3's model only has `<p></p>`, so S3 should probably have a customization to provide a better description.

jdisanti:
## Description
- Generates README.md files for individual service client crates, and references them from the Cargo.toml files so that they display on crates.io as the overview for each crate.
- Fixes #754.

## Testing
- Unit tested where it made sense
- Manually inspected generated README.md files and Cargo.toml files

## Checklist


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsReadmeDecorator.kt
```diff
@@ -0,0 +1,80 @@
...
+            val description = Jsoup.parse(
+                codegenContext.settings.getService(codegenContext.model).getTrait<DocumentationTrait>()?.value ?: ""
+            ).text()
+            val moduleName = codegenContext.settings.moduleName
+
+            writer.raw(
+                """
+                # $moduleName
+
+                **Please Note: The SDK is currently released as an alpha and is intended strictly for
```
jdisanti, line 39: Long term, I think we should probably generate the entire top-level README.md file in aws-sdk-rust so that it's kept in sync with this one. I omitted a few sections in this README since I didn't think all of them made sense in this context.

For now, we will need to remember to update both READMEs, unless you think the investment is worth it now?

aws/sdk/build.gradle.kts
```diff
@@ -166,7 +169,13 @@ fun discoverServices(): List<AwsService> {
                 } else {
                     listOf()
                 }
-                AwsService(service = service.id.toString(), module = sdkId, modelFile = file, extraFiles = extras)
+                AwsService(
+                    service = service.id.toString(),
+                    module = sdkId,
+                    moduleDescription = "AWS SDK for $title",
```
jdisanti, line 175: Examples for this value:
- AWS SDK for Amazon Simple Storage Service
- AWS SDK for AWS Identity and Access Management
- AWS SDK for Amazon DynamoDB

I think I would prefer to it to be "AWS SDK for S3", but I'm not sure if there's a reliable model attribute for that. One would think the Service `sdkId` would work, but that value doesn't look great for a number of services.

jdisanti:
I'm not sure why these changes would cause the errors CI is seeing for the generated transcribestreaming crate.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsReadmeDecorator.kt
rcoh: maybe we can put an HTML comment in both noting that they both need to be updated

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/CodegenDelegator.kt
```diff
@@ -49,11 +50,16 @@ open class RustCrate(
         }
     }
 
-    fun finalize(settings: RustSettings, libRsCustomizations: List<LibRsCustomization>) {
+    fun finalize(
+        settings: RustSettings,
+        model: Model,
+        manifestCustomizations: Map<String, Any?>,
```
rcoh, line 56: add a comment about what this map is since it's the fairly opaque str->any?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/CargoTomlGenerator.kt
```diff
@@ -38,7 +41,8 @@ class CargoTomlGenerator(
             "dev-dependencies" to dependencies.filter { it.scope == DependencyScope.Dev }
                 .associate { it.name to it.toMap() },
             "features" to cargoFeatures.toMap()
-        )
+        ).deepMergeWith(manifestCustomizations)
```
rcoh, line 44: do we need to do anything to ensure deterministic ordering?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/LibRsGenerator.kt
```diff
@@ -20,13 +24,16 @@ sealed class LibRsSection(name: String) : Section(name) {
...
+    private val settings: RustSettings,
+    private val model: Model,
     private val modules: List<RustModule>,
     private val customizations: List<LibRsCustomization>
 ) {
     fun render(writer: RustWriter) {
         writer.first {
             customizations.forEach { it.section(LibRsSection.Attributes)(this) }
+
+            val libraryDocs = settings.getService(model).getTrait<DocumentationTrait>()?.value ?: settings.moduleName
```
rcoh, line 36: I'd probably lean towards being more explicit here and having LibRsGenerator accept a service, but fine either way. Using the doc trait directly here makes it a little trickier to customize (although I guess we could always do it as a model customization?

I'm just imagining how this will work when we want to start rendering examples in here

jdisanti: We can easily add a section (similar to the attributes section above) for rendering examples.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsReadmeDecorator.kt
jdisanti: I'm skeptical anyone would actually see that when updating one of the READMEs unless they were updating the entire README rather than just a part of it.

PR #775, opened 2021-10-15
2 DVR tests for ECS

rcoh:
## Motivation and Context
aws-sdk-rust#123

## Description
- Updates JSON response parsing to be case-insensitive.

- Adds two new dvr tests:
  - ECS used directly
  - ECS used via assume role
  The ECS responses were taken & sanitized from a real ECS response.

## Testing
- [x] these are tests.
- [x] ECS works IRL

## Checklist
- [x] no changelog

### DIFF
--- a/aws/rust-runtime/aws-config/src/json_credentials.rs
+++ b/aws/rust-runtime/aws-config/src/json_credentials.rs
@@ -89,7 +89,7 @@ pub(crate) enum JsonCredentials<'a> {
 /// distinguishes between a successful response that contains credentials vs. an error with a code and
 /// error message.
 ///
-/// Keys are matched case ignoring case.
+/// Keys are case insensitive.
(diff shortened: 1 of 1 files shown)


aws/rust-runtime/aws-config/src/json_credentials.rs
```diff
@@ -88,6 +88,8 @@ pub(crate) enum JsonCredentials<'a> {
 /// response from the credential provider vs. something invalid / unexpected. The inner error
 /// distinguishes between a successful response that contains credentials vs. an error with a code and
 /// error message.
+///
+/// Keys are matched case ignoring case.
```
Velfi, line 92:
```suggestion
/// Key casing is ignored when matching keys.
```

Velfi (approved):

PR #786, opened 2021-10-20
Refactor server codegen for restJson1

crisidev:
## Description
Refactor server restJson1 and allow generation of HTTP request / response tests.

## Testing
Using the `simple.smithy` model now two unit tests are autogenerated and run, testing serialization and deserialization of HTTP request/response.

## Checklist
I don't think we are already at a level of maturity where it is worth updating the changelog.

### DIFF
--- a/aws/SDK_CHANGELOG.md
+++ b/aws/SDK_CHANGELOG.md
@@ -31,6 +31,8 @@ vNext (Month Day, Year)
 
 - Moved the contents of `aws-auth` into the `aws-http` runtime crate (smithy-rs#783)
 - Add tracing output for resolved AWS endpoint (smithy-rs#784)
+- Add support for the following Glacier customizations:
+  - Set the ApiVersion header (smithy-rs#138, #787)
 
 v0.0.21-alpha (October 15th, 2021)
 ==================================
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt
@@ -10,6 +10,7 @@ import software.amazon.smithy.rust.codegen.smithy.customize.CombinedCodegenDecor
 import software.amazon.smithy.rustsdk.customize.apigateway.ApiGatewayDecorator
 import software.amazon.smithy.rustsdk.customize.auth.DisabledAuthDecorator
 import software.amazon.smithy.rustsdk.customize.ec2.Ec2Decorator
+import software.amazon.smithy.rustsdk.customize.glacier.GlacierDecorator
 import software.amazon.smithy.rustsdk.customize.s3.S3Decorator
 
 val DECORATORS = listOf(
@@ -34,7 +35,8 @@ val DECORATORS = listOf(
     DisabledAuthDecorator(),
     ApiGatewayDecorator(),
     S3Decorator(),
-    Ec2Decorator()
+    Ec2Decorator(),
+    GlacierDecorator()
 )
 
 class AwsCodegenDecorator : CombinedCodegenDecorator(DECORATORS) {
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/glacier/ApiVersionHeader.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/glacier/ApiVersionHeader.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package software.amazon.smithy.rustsdk.customize.glacier
+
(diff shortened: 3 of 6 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/RestJson1.kt
```diff
@@ -256,7 +239,7 @@ class RestJson1HttpSerializerGenerator(
         val responseTests =
             operationShape
                 .getTrait<HttpResponseTestsTrait>()
-                ?.getTestCasesFor(AppliesTo.SERVER)
+                ?.getTestCasesFor(AppliesTo.CLIENT)
```
rcoh, line 242: I think you want server? the client ones are client-only weird behavior

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/RestJson1.kt
```diff
@@ -99,11 +82,11 @@ class RestJson1HttpSerializerGenerator(
             "build_error" to runtimeConfig.operationBuildError(),
             "JsonSerdeError" to error.member("Error"),
             "JsonObjectWriter" to smithyJson.member("serialize::JsonObjectWriter"),
-            "parse_http_response" to smithyHttp.member("response::ParseHttpResponse"),
-            "sdk_body" to RuntimeType.sdkBody(runtimeConfig = runtimeConfig),
+            "ParseHttpResponse" to smithyHttp.member("response::ParseHttpResponse"),
```
rcoh, line 85: 👍🏻

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/RestJson1.kt
```diff
@@ -315,20 +298,14 @@ class RestJson1HttpSerializerGenerator(
             }
             rust(
                 """let response = #T(&expected).expect("failed to serialize response");""",
-                serde.member(fnName)
+                operation.member(fnName)
```
rcoh, line 301: generally, you'll probably want to test the external API instead of testing the serializers directly (but maybe the external API doesn't exist yet?)

That also means you don't need to guess at the `fnName`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/RestJson1.kt
crisidev: I want server yes, and I forget about changing this back during my tests. Thanks for the catch. I agree, sooner or later we will want a separate class. I just made this work, so hopefully I will be able to start working on the real protocol tests already available :)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/RestJson1.kt
crisidev: The external API is not there yet as you rightly guessed. David is working on extractors for Json and once that is done we will start autogenerating the public and more stable API. Little steps 😄

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/RestJson1.kt
```diff
@@ -410,7 +374,7 @@ class RestJson1HttpRequestDeserializerGenerator(
         val deserializerSymbol = jsonParserGenerator.serverInputParser(operationShape)
         if (deserializerSymbol == null) {
             logger.warning(
-                "[rust-server-codegen] $inputShape: response output serialization does not contain any member"
+                "[rust-server-codegen] $inputShape: request input deserialization does not contain any member"
```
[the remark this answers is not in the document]
crisidev: You know what, these warnings are kinda useless, you are right. I am removing them.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/RestJson1.kt
```diff
@@ -618,6 +618,8 @@ class RestJson1HttpDeserializerGenerator(
                 "let request = request.map(|body| #{Bytes}::from(body.bytes().unwrap().to_vec()));",
                 *codegenScope
             )
+            rust("assert_eq!(request.method().as_str(), \"${testCase.method}\");")
+            rust("assert_eq!(request.uri().to_owned(), \"${testCase.uri}\");")
```
rcoh, line 622: I'm not sure how this works...fwiw, when comparing Uris I usually generated `Uri::from_static` for the comparison.

rcoh, line 622: (you'll need to import it)
```suggestion
            rust("assert_eq!(request.method().as_str(), ${testCase.method.dq()});")
            rust("assert_eq!(request.uri().to_owned(), ${testCase.uri.dq()});")
```

PR #804, opened 2021-10-25
Generate simpler operation error types for the server

### LINKED ISSUE #780: Error types for servers
As of writing, `rust-server-codegen` generates (e.g. for the [`simple.smithy` model](https://github.com/awslabs/smithy-rs/blob/main/codegen-server-test/model/simple.smithy)):

```rust
#[non_exhaustive]
#[derive(std::fmt::Debug)]
pub struct RegisterServiceError {
    pub kind: RegisterServiceErrorKind,
    pub(crate) meta: smithy_types::Error,
}
#[non_exhaustive]
#[derive(std::fmt::Debug)]
pub enum RegisterServiceErrorKind {
    ResourceAlreadyExists(crate::error::ResourceAlreadyExists),
    /// An unexpected error, eg. invalid JSON returned by the service or an unknown error code
    Unhandled(Box<dyn std::error::Error + Send + Sync + 'static>),
}

/// Returned when a new resource cannot be created because one already exists.
#[non_exhaustive]
pub struct ResourceAlreadyExists {
    pub message: std::option::Option<std::string::String>,
}
```

Where the  `Error` type for `meta` is:

```rust
/// Generic Error type
///
/// For many services, Errors are modeled. However, many services only partially model errors or don't
/// model errors at all. In these cases, the SDK will return this generic error type to expose the
/// `code`, `message` and `request_id`.
#[derive(Debug, Eq, PartialEq, Default, Clone)]
pub struct Error {
    code: Option<String>,
    message: Option<String>,
    request_id: Option<String>,
    extras: HashMap<&'static str, String>,
}
```

This is too coupled to SDK client errors. In Rust servers there shouldn't really be an `Unhandled` variant. We should also have serde-specific errors that are independent from errors returned by the operation's logic. Additionally, we also don't need the `meta` field.

All in all, I think that error types for servers are far simpler; a simple `RegisterServiceError` `enum` like this would suffice:

```rust
#[non_exhaustive]
#[derive(std::fmt::Debug)]
pub enum RegisterServiceError {
    ResourceAlreadyExists(crate::error::ResourceAlreadyExists),
}
```

I'm opening this issue for discussion and to see if we can reconcile these differences into error types that can be used both by the client and the server, or if having completely different code generation for errors for the server is more appropiate here.

david-perez:
Currently, operation error types are designed to be used by clients.
Clients need to cope with a large scale of possible errors, including
unmodeled errors and errors coming from different layers of the stack.
On the other hand, all operation server errors should be related to
business logic; we want all of them to be correctly modeled and have
compilation fail if any of them is not properly handled.

This commit revamps error generation by the `rust-server-codegen`
plugin, generating a simple Rust enum type for each fallible operation,
holding one variant per modeled error.

Closes #780.

## Checklist

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
@@ -5,11 +5,9 @@
 
 package software.amazon.smithy.rust.codegen.server.smithy.generators
 
-import software.amazon.smithy.codegen.core.Symbol
 import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.knowledge.OperationIndex
 import software.amazon.smithy.model.shapes.OperationShape
-import software.amazon.smithy.model.shapes.Shape
 import software.amazon.smithy.rust.codegen.rustlang.Attribute
 import software.amazon.smithy.rust.codegen.rustlang.RustMetadata
 import software.amazon.smithy.rust.codegen.rustlang.RustWriter
@@ -17,10 +15,8 @@ import software.amazon.smithy.rust.codegen.rustlang.Writable
 import software.amazon.smithy.rust.codegen.rustlang.documentShape
 import software.amazon.smithy.rust.codegen.rustlang.rust
 import software.amazon.smithy.rust.codegen.rustlang.rustBlock
-import software.amazon.smithy.rust.codegen.rustlang.writable
 import software.amazon.smithy.rust.codegen.smithy.RuntimeType
 import software.amazon.smithy.rust.codegen.smithy.RustSymbolProvider
-import software.amazon.smithy.rust.codegen.smithy.customize.Section
 import software.amazon.smithy.rust.codegen.smithy.generators.error.errorSymbol
 import software.amazon.smithy.rust.codegen.util.toSnakeCase
 
@@ -58,7 +54,7 @@ class ServerCombinedErrorGenerator(
         writer.rustBlock("impl #T for ${symbol.name}", RuntimeType.stdfmt.member("Display")) {
             rustBlock("fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result") {
                 delegateToVariants {
-                    writable { rust("_inner.fmt(f)") }
+                    rust("_inner.fmt(f)")
                 }
             }
         }
@@ -77,18 +73,10 @@ class ServerCombinedErrorGenerator(
         writer.rustBlock("impl #T for ${symbol.name}", RuntimeType.StdError) {
             rustBlock("fn source(&self) -> Option<&(dyn #T + 'static)>", RuntimeType.StdError) {
                 delegateToVariants {
-                    writable {
-                        when (it) {
-                            is VariantMatch.Modeled -> rust("Some(_inner)")
-                        }
-                    }
+                    rust("Some(_inner)");
                 }
             }
         }
-    }
-
-    sealed class VariantMatch(name: String) : Section(name) {
-        data class Modeled(val symbol: Symbol, val shape: Shape) : VariantMatch("Modeled")
     }
 
     /**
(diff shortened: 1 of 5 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -155,26 +153,6 @@ private class ServerHttpProtocolImplGenerator(
         operationShape: OperationShape
     ) {
         val errorSymbol = operationShape.errorSymbol(symbolProvider)
-        val successCode = httpBindingResolver.httpTrait(operationShape).code
```
david-perez, line 158: The server doesn't need to parse responses.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -93,7 +93,7 @@ class ServerProtocolTestGenerator(
             val testCases = error.getTrait<HttpResponseTestsTrait>()?.testCases.orEmpty()
             testCases.map { TestCase.ResponseTest(it, error) }
         }
-        val allTests: List<TestCase> = (requestTests + responseTests + errorTests).filterMatching()
```
david-perez, line 96: Note that I have removed generation of [HTTP protocol compliance response tests](https://awslabs.github.io/smithy/1.0/spec/http-protocol-compliance-tests.html?highlight=httpresponsetest#httpresponsetests). The generated tests were of the form:

```rust
/// Test ID: RegisterServiceResponseTest
#[tokio::test]
async fn register_service_response_test_response() {
    let expected_output = crate::output::RegisterServiceOutput::builder()
        .set_id(Some("1".to_string()))
        .set_name(Some("TestService".to_string()))
        .build();
    use aws_smithy_http::response::ParseStrictResponse;
    use aws_smithy_http_server::response::SerializeHttpResponse;
    let op = crate::operation::RegisterService::new();
    let http_response = op
        .serialize(&expected_output)
        .expect("unable to serialize response body");
    let parsed = op.parse(&http_response);
    let parsed = parsed.expect("unable to parse response");
    assert_eq!(parsed.id, expected_output.id, "Unexpected value for `id`");
    assert_eq!(
        parsed.name, expected_output.name,
        "Unexpected value for `name`"
    );
}
```

Of course these tests now fail, because I've removed the `parse()` HTTP responses method on operations, since the server doesn't need to parse responses, only requests.

This led me to read the Smithy spec and the purpose of these tests, and I think we are doing them wrong. I think what we should be doing is instantiating an operation output using the `params` in the model, serializing that into an HTTP response, and directly inspecting the status code, headers, body etc. of that response and asserting that they match whatever was written in the model.

Is my understanding correct? If so, I can readd the tests in another PR but implementing them like I described in the previous paragraph.

crisidev: Your understanding is correct. The protocol tests have to be changed to 1) adhere to the new public API 2) do the right thing, which currently they are not doing.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: Yep, this can go. It was here only to be able to build the protocol tests, which we are changing. Good to see this going :)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
```diff
@@ -0,0 +1,125 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    sealed class VariantMatch(name: String) : Section(name) {
+        data class Modeled(val symbol: Symbol, val shape: Shape) : VariantMatch("Modeled")
+    }
```
crisidev, line 92: I'd rather make this simpler since all errors are modeled.

crisidev (approved):
LGTM

crisidev (approved):
LGTM

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
david-perez: https://github.com/awslabs/smithy-rs/pull/840

PR #814, opened 2021-10-27
Use the same protocol dispatcher used in the client codegen

crisidev:
## Motivation and Context
Right now we copy paste the HttpBoundProtocolGenerator in its own server
specific class ServerHttpProtocolGenerator and expand the ProtocolSupport
to be able to handle server specific configurations. Protocols can
inherit from ServerHttpProtocolGenerator and tune their implementations

## Description
Other class that have been copied and adapted:
* The ProtocolLoader is implemented inside ServerProtocolLoader
  (ServerProtocol.kt)
  It matches the right protocol and casts it to
  a ServerProtocolGenerator that can be used inside the
  ProtocolGeneratorFactory.
* The ProtocolGenerator is implemented inside ServerProtocolGenerator (
  ServerProtocolGenerator.kt)
  It implements serverProtocolOperation() that can build the
  operation definition as we like. This method is than called by
  ServiceGenerator per every operation in the model.
* RestJson and RestJsonFactory are implemented in RestJson.kt and
  inherit from ServerHttpProtocolGenerator
* Implement the HTTP protocol definition and adapt tests

## Checklist


crisidev:
Closed in favour of new set of PRs

PR #822, opened 2021-11-01
Runtime and client codegen support for improved server protocol dispatcher

crisidev:
## Motivation and Context
We are changing how the HTTP protocol operation generation is done in server-codegen. It will mimic how the client-codegen works, but delegating the protocol specifics to their implementations and dispatching them through a protocol loader.

This is the first PR, adding a Rust runtime defining the server operation traits and adapts the client codegen allowing server configurations to the ProtocolSupport and changing the visibility of HttpBoundProtocolGenerator methods to allow server-codegen to use them (will be needed mostly for testing).

This can be reviewed as I'll be waiting to merge this before submitting the dispatcher implementation. 

## Description
Implement Rust runtime defining the traits used by the server to build operations
parsers/serializers:
  * ParseHttpRequest
  * SerializeHttpResponse
  * SerializeHttpError

Refactor ProtocolSupport to allow server configurations:
  * requestDeserialization (bool)
  * requestBodyDeserialization (bool)
  * responseSerialization(bool)
  * errorSerialization(bool)

Change visibility of HttpBoundProtocolGenerator methods to allow to use
them in the server generator

## Testing
Tested on CI

## Checklist

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
@@ -272,23 +272,6 @@ data class RuntimeType(val name: String?, val dependency: RustDependency?, val n
             namespace = "aws_smithy_http::response"
         )
 
-        fun parseHttpRequest(runtimeConfig: RuntimeConfig) = RuntimeType(
-            "ParseHttpRequest",
-            dependency = CargoDependency.SmithyHttpServer(runtimeConfig),
-            namespace = "aws_smithy_http_server::request"
-        )
-
-        fun serializeHttpResponse(runtimeConfig: RuntimeConfig) = RuntimeType(
-            "SerializeHttpResponse",
-            dependency = CargoDependency.SmithyHttpServer(runtimeConfig),
-            namespace = "aws_smithy_http_server::response"
-        )
-
-        fun serializeHttpError(runtimeConfig: RuntimeConfig) = RuntimeType(
-            "SerializeHttpError",
-            dependency = CargoDependency.SmithyHttpServer(runtimeConfig),
-            namespace = "aws_smithy_http_server::response"
-        )
         val Bytes = RuntimeType("Bytes", dependency = CargoDependency.Bytes, namespace = "bytes")
 
         fun forInlineDependency(inlineDependency: InlineDependency) =
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
@@ -266,12 +266,29 @@ data class RuntimeType(val name: String?, val dependency: RustDependency?, val n
         fun sdkBody(runtimeConfig: RuntimeConfig): RuntimeType =
             RuntimeType("SdkBody", dependency = CargoDependency.SmithyHttp(runtimeConfig), "aws_smithy_http::body")
 
-        fun parseStrict(runtimeConfig: RuntimeConfig) = RuntimeType(
+        fun parseStrictResponse(runtimeConfig: RuntimeConfig) = RuntimeType(
             "ParseStrictResponse",
             dependency = CargoDependency.SmithyHttp(runtimeConfig),
             namespace = "aws_smithy_http::response"
         )
 
+        fun parseHttpRequest(runtimeConfig: RuntimeConfig) = RuntimeType(
+            "ParseHttpRequest",
+            dependency = CargoDependency.SmithyHttpServer(runtimeConfig),
+            namespace = "aws_smithy_http_server::request"
+        )
+
+        fun serializeHttpResponse(runtimeConfig: RuntimeConfig) = RuntimeType(
+            "SerializeHttpResponse",
+            dependency = CargoDependency.SmithyHttpServer(runtimeConfig),
+            namespace = "aws_smithy_http_server::response"
+        )
+
+        fun serializeHttpError(runtimeConfig: RuntimeConfig) = RuntimeType(
+            "SerializeHttpError",
+            dependency = CargoDependency.SmithyHttpServer(runtimeConfig),
+            namespace = "aws_smithy_http_server::response"
+        )
         val Bytes = RuntimeType("Bytes", dependency = CargoDependency.Bytes, namespace = "bytes")
 
         fun forInlineDependency(inlineDependency: InlineDependency) =
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolGenerator.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolGenerator.kt
@@ -67,7 +67,7 @@ class HttpBoundProtocolGenerator(
     HttpBoundProtocolTraitImplGenerator(codegenContext, protocol),
 )
 
-private class HttpBoundProtocolTraitImplGenerator(
+class HttpBoundProtocolTraitImplGenerator(
     private val codegenContext: CodegenContext,
     private val protocol: Protocol,
 ) : ProtocolTraitImplGenerator {
@@ -78,7 +78,7 @@ private class HttpBoundProtocolTraitImplGenerator(
     private val operationDeserModule = RustModule.private("operation_deser")
 
     private val codegenScope = arrayOf(
-        "ParseStrict" to RuntimeType.parseStrict(runtimeConfig),
+        "ParseStrict" to RuntimeType.parseStrictResponse(runtimeConfig),
         "ParseResponse" to RuntimeType.parseResponse(runtimeConfig),
         "http" to RuntimeType.http,
         "operation" to RuntimeType.operationModule(runtimeConfig),
@@ -160,7 +160,7 @@ private class HttpBoundProtocolTraitImplGenerator(
         )
     }
 
-    private fun parseError(operationShape: OperationShape): RuntimeType {
+    fun parseError(operationShape: OperationShape): RuntimeType {
         val fnName = "parse_${operationShape.id.name.toSnakeCase()}_error"
         val outputShape = operationShape.outputShape(model)
         val outputSymbol = symbolProvider.toSymbol(outputShape)
@@ -258,7 +258,7 @@ private class HttpBoundProtocolTraitImplGenerator(
         }
     }
 
-    private fun parseResponse(operationShape: OperationShape): RuntimeType {
+    fun parseResponse(operationShape: OperationShape): RuntimeType {
         val fnName = "parse_${operationShape.id.name.toSnakeCase()}_response"
         val outputShape = operationShape.outputShape(model)
         val outputSymbol = symbolProvider.toSymbol(outputShape)
@@ -416,7 +416,7 @@ class HttpBoundProtocolBodyGenerator(
     private val operationSerModule = RustModule.private("operation_ser")
 
     private val codegenScope = arrayOf(
-        "ParseStrict" to RuntimeType.parseStrict(runtimeConfig),
+        "ParseStrict" to RuntimeType.parseStrictResponse(runtimeConfig),
         "ParseResponse" to RuntimeType.parseResponse(runtimeConfig),
         "http" to RuntimeType.http,
         "hyper" to CargoDependency.HyperWithStream.asType(),
--- a/rust-runtime/aws-smithy-http-server/Cargo.toml
+++ b/rust-runtime/aws-smithy-http-server/Cargo.toml
@@ -1,11 +1,15 @@
 [package]
 name = "aws-smithy-http-server"
 version = "0.1.0"
-authors = ["Matteo Bigoi <matbigoi@amazon.com>", "David Pérez <davidpz@amazon.com>"]
+authors = ["Smithy Rust Server <smithy-rs-server@amazon.com>"]
 edition = "2018"
 description = """
 Server traits and utilities used by the code generator.
+
+NOTE: THIS IS HIGHLY EXPERIMENTAL AND SHOULD NOT BE USED YET.
 """
+# until this is not stable, it is not publishable.
+publish = false
 
 [dependencies]
 aws-smithy-http = { path = "../aws-smithy-http" }
--- a/rust-runtime/aws-smithy-http-server/Cargo.toml
+++ b/rust-runtime/aws-smithy-http-server/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "aws-smithy-http-server"
 version = "0.1.0"
-authors = ["Matteo Bigoi <matbigoi@amazon.com>", "David Perez <davidpz@amazon.com>"]
+authors = ["Matteo Bigoi <matbigoi@amazon.com>", "David Pérez <davidpz@amazon.com>"]
 edition = "2018"
 description = """
 Server traits and utilities used by the code generator.
--- a/rust-runtime/aws-smithy-http-server/Cargo.toml
+++ b/rust-runtime/aws-smithy-http-server/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "aws-smithy-http-server"
+version = "0.1.0"
+authors = ["Matteo Bigoi <matbigoi@amazon.com>", "David Perez <davidpz@amazon.com>"]
+edition = "2018"
+description = """
+Server traits and utilities used by the code generator.
+"""
+
+[dependencies]
+aws-smithy-http = { path = "../aws-smithy-http" }
+bytes = "1.1"
+http = "0.2"
+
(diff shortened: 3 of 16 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolGenerator.kt
```diff
@@ -67,7 +67,7 @@ class HttpBoundProtocolGenerator(
     HttpBoundProtocolTraitImplGenerator(codegenContext, protocol),
```
crisidev, line 67: I will be using the code generated by this file in the server codegen to support some of the ser/de tests I am generating. For this reason I need to make some visibility changes to this class. It is not a strict contract since the code generated is used only in tests.

rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "aws-smithy-http-server"
+version = "0.1.0"
+authors = ["Matteo Bigoi <matbigoi@amazon.com>", "David Perez <davidpz@amazon.com>"]
```
david-perez, line 4:
```suggestion
authors = ["Matteo Bigoi <matbigoi@amazon.com>", "David Pérez <davidpz@amazon.com>"]
```

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// response and consumed by the client. However, even in the case of S3::GetObject, errors
+    /// require reading the entire body.
+    ///
+    /// This also facilitates `EventStream` and other streaming HTTP protocols by enabling the
+    /// handler to take ownership of the HTTP request directly.
+    ///
+    /// Currently `parse_unloaded` operates on a borrowed HTTP request to enable
+    /// the caller to provide a raw HTTP response to the caller for inspection after the response is
+    /// returned. For EventStream-like use cases, the caller can use `mem::swap` to replace
+    /// the streaming body with an empty body as long as the body implements default.
```
david-perez, line 42:
```suggestion
    /// the streaming body with an empty body as long as the body implements `std::default::Default`.
```

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,111 @@
...
+    ///
+    /// We should consider if this is too limiting & if this should take an owned response instead.
```
david-perez, line 44:
```suggestion
    /// We should consider if this is too limiting and if this should take an owned request instead.
```

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,111 @@
...
+/// event streams, regular request-response style operations, as well as any other HTTP-based
+/// protocol that we manage to come up with.
+///
+/// It also enables this critical and core trait to avoid being async, and it makes code that uses
+/// the trait easier to test.
+pub trait ParseHttpRequest {
+    /// Input type of the HttpRequest
+    ///
+    /// For request/response style operations, this is typically something like:
+    /// `Result<OperationInput, OperationInputError>`
```
david-perez, line 23: If this already returns an operation input, what do we need `axum`'s `FromRequest` for? Isn't this already the extractor?

Also, won't the error type be fixed? I'd assume the error type to be an enum with generic errors that can occur during parsing that are **not** specific to the particular operation input being extracted, e.g. things like `axum`'s rejection types: https://docs.rs/axum/0.2.8/axum/extract/rejection/index.html

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+///
+/// It is designed to be flexible, because `Output` and `Struct` are unconstrained, it can be used to support
+/// event streams, regular request-response style operations, as well as any other HTTP-based
+/// protocol that we manage to come up with.
+///
+/// It also enables this critical and core trait to avoid being async, and it makes code that uses
+/// the trait easier to test.
+///
+/// TODO: streaming and not-fully loaded body serialization is no developed yet
+pub trait SerializeHttpResponse {
```
david-perez, line 17: So this is analogous to `axum`'s `IntoResponse`?

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Struct instance to be serialized into the HTTP body
+    ///
+    /// For request/response style operations, this is typically something like:
+    /// `OperationOutput`
+    type Struct;
```
david-perez, line 22: Why isn't it this struct the one that directly `impl`s `SerializeHttpResponse`?

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// HTTP response output
+    ///
+    /// For request/response style operations, this is typically something like:
+    /// `Result<ResponseBytes>, Error>`
```
david-perez, line 26:
```suggestion
    /// `Result<ResponseBytes, Error>`
```

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+    type Output;
+
+    /// Serialize an HTTP response from a fully loaded body. This is for standard request/response style
+    /// APIs like RestJson1.
```
david-perez, line 30:
```suggestion
    /// protocols like RestJson1.
```

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+/// `SerializeHttpError` is a generic trait for serializing structured errors into HTTP responses.
+///
+/// It is designed to be flexible, because `Output` and `Struct` are unconstrained, it can be used to support
+/// event streams, regular request-response style operations, as well as any other HTTP-based
+/// protocol that we manage to come up with.
+///
+/// It also enables this critical and core trait to avoid being async, and it makes code that uses
+/// the trait easier to test.
+///
+/// TODO: streaming and not-fully loaded body serialization is no developed yet
```
david-perez, line 43:
```suggestion
/// TODO: streaming and not-fully loaded body serialization is not developed yet
```

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,111 @@
...
+use aws_smithy_http::body::SdkBody;
+use bytes::Bytes;
+
+/// `ParseHttpRequest` is a generic trait for parsing structured data from HTTP requests.
+///
+/// It is designed to be flexible, because `Input` is unconstrained, it can be used to support
+/// event streams, regular request-response style operations, as well as any other HTTP-based
+/// protocol that we manage to come up with.
+///
+/// It also enables this critical and core trait to avoid being async, and it makes code that uses
```
david-perez, line 17: I don't understand these comments in the traits saying that it enables the trait to avoid being `async`.

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+pub trait SerializeHttpError {
+    /// Struct instance to be serialized into the HTTP body
+    ///
+    /// For request/response style operations, this is typically something like:
+    /// `OperationError`
+    type Struct;
+    /// HTTP response output
+    ///
+    /// For request/response style operations, this is typically something like:
+    /// `Result<ResponseBytes>, Error>`
```
david-perez, line 53:
```suggestion
    /// `Result<ResponseBytes, Error>`
```

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+    type Output;
+
+    /// Serialize an HTTP response from a fully loaded body. This is for standard request/response style
+    /// APIs like RestJson1.
```
david-perez, line 57:
```suggestion
    /// protocols like RestJson1.
```

rust-runtime/aws-smithy-http-server/src/request.rs
crisidev: 👍🏼

rust-runtime/aws-smithy-http-server/src/request.rs
crisidev: 👍🏼

rust-runtime/aws-smithy-http-server/src/request.rs
crisidev: 1) I thought about using directly axum traits, but it was easier for me to approach this change by first creating my own simple traits. This allowed me to understand how the code generation works for operation I/O. Once this PR and the one implementing the codegen are landed, I think we should probably switch to the axum traits.

2) The error here should be just the generic server Error type, I'll fix the documentation.

rust-runtime/aws-smithy-http-server/src/response.rs
crisidev: See above about the reasoning behind these traits.

rust-runtime/aws-smithy-http-server/src/response.rs
crisidev: Oh right, this is an excellent idea, I don't know why I did not think about it. I'll fix this.

rust-runtime/aws-smithy-http-server/src/response.rs
crisidev: 👍🏼

rust-runtime/aws-smithy-http-server/src/response.rs
crisidev: 👍🏼

rust-runtime/aws-smithy-http-server/src/response.rs
crisidev: 👍🏼

rust-runtime/aws-smithy-http-server/src/response.rs
crisidev: 👍🏼

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,111 @@
...
+/// the trait easier to test.
+pub trait ParseHttpRequest {
+    /// Input type of the HttpRequest
+    ///
+    /// For request/response style operations, this is typically something like:
+    /// `Result<OperationInput, OperationInputError>`
+    ///
+    /// For streaming operations, this is something like:
+    /// `Result<EventStream<OperationEvent>>, OperationStreamingError>`
+    type Input;
```
david-perez, line 27: Similar idea here: why doesn't the `Input` type directly `impl` `ParseHttpRequest`?

rust-runtime/aws-smithy-http-server/src/request.rs
crisidev: This is a little forward thinking in terms of supporting streaming types. `If` we use the SdkBody type for streaming types, we can avoid async traits as SdkBody type implements directly future polling: https://github.com/awslabs/smithy-rs/blob/main/rust-runtime/aws-smithy-http/src/body.rs#L204

rust-runtime/aws-smithy-http-server/src/request.rs
crisidev: Yep, you are right, will fix it.

david-perez (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -266,12 +266,29 @@ data class RuntimeType(val name: String?, val dependency: RustDependency?, val n
...
+            "SerializeHttpResponse",
+            dependency = CargoDependency.SmithyHttpServer(runtimeConfig),
+            namespace = "aws_smithy_http_server::response"
+        )
+
+        fun serializeHttpError(runtimeConfig: RuntimeConfig) = RuntimeType(
+            "SerializeHttpError",
+            dependency = CargoDependency.SmithyHttpServer(runtimeConfig),
+            namespace = "aws_smithy_http_server::response"
+        )
```
rcoh, line 291: if you want, you may want to make a companion object in `ServerCodegen` and define those there

rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "aws-smithy-http-server"
+version = "0.1.0"
+authors = ["Matteo Bigoi <matbigoi@amazon.com>", "David Pérez <davidpz@amazon.com>"]
```
rcoh, line 4: you should probably make `smithy-rs-server` or similar email alias and include it in the author list of the crates (we do this with `aws-sdk-rust@amazon.com`

rust-runtime/aws-smithy-http-server/src/request.rs
rcoh: that's not totally accurate ^

This trait can avoid being async because _it_ never needs to asynchronously read the body. Instead, it can signal to a driver when the body needs to be written.

However, I'm not sure this is the right design for the server—it doesn't enable any sort of behavior where you'd have a max content length or have parsing interact with whether or not you want to read more of a response body.

rust-runtime/aws-smithy-http-server/src/request.rs
rcoh: doing it this way has some advantages, namely, that you can give the parser state that the input doesn't have. Eg:

```rust
struct MyInputParser {
  a: String
  bunch: bool,
  of: u64
  parsing_settings: Bytes
}

impl ParseHttpRequest for MyInputParser {
  ...
}
```
It also avoids the orphan trait rule if you ever want to define parsers for something that's actually in an external crate.

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+/// It also enables this critical and core trait to avoid being async, and it makes code that uses
+/// the trait easier to test.
+///
+/// TODO: streaming and not-fully loaded body serialization is not developed yet
+pub trait SerializeHttpError {
+    /// Struct instance to be serialized into the HTTP body
+    ///
+    /// For request/response style operations, this is typically something like:
+    /// `OperationError`
+    type Struct;
```
rcoh, line 49: I probably wouldn't name this `Struct` personally. Maybe `Input`?

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// HTTP response output
+    ///
+    /// For request/response style operations, this is typically something like:
+    /// `Result<ResponseBytes, Error>`
+    type Output;
```
rcoh, line 54: I suspect. you'll actually want to not have this be free. By having it like this, you'll need to constrain it everywhere to be something useful (or you won't end up actually using this trait at all)

rcoh (approved):
This looks fine to me. However, I would _strongly_ encourage you to whip up some handwritten versions of the code you eventually plan to generate for the server (including usages / implementations of these traits).

I think that will really help clarify the API you want to have and how it will bridge with Axum.

rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition = "2018"
+description = """
+Server traits and utilities used by the code generator.
```
rcoh, line 7: can you also add some description about how this is extra unstable? And also maybe `publish = false` in the metadata?

rust-runtime/aws-smithy-http-server/Cargo.toml
crisidev: Will do, thanks.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
crisidev: Ah, makes sense.. I haven't used a companion object yet, but I have seen them used around the code. I'll follow your advice!

rust-runtime/aws-smithy-http-server/src/response.rs
crisidev: These traits are temporary and are just here to allow to generate working code for the protocol dispatcher. They will be updated to the real FromRequest / ToReponse traits once they will be formalized. Hopefully soon 😃

rust-runtime/aws-smithy-http-server/src/response.rs
crisidev: These traits are temporary and are just here to allow to generate working code for the protocol dispatcher. They will be updated to the real FromRequest / ToReponse traits once they will be formalized

PR #823, opened 2021-11-01
Smithy 1.13

rcoh:
## Motivation and Context
Upgrade to Smithy 1.13. The new release is breaking CI because the CLI version is being autodiscovered.

## Description
The Smith 1.13 upgrade adds new restJson protocol tests. These demonstrated two issues:

1. We were not properly using the http binding index to compute content types. We should not have fallen back to a default content type—instead we need to use exactly what is given. We were also using the index incorrectly: `document` in the context of the index refers to anything in the HTTP body, _not_ the Smithy document type.
2. We were sending `Content-Length: 0` even when no HTTP body was modeled.
3. We were sending an empty body for structures targeted `@httpPayload` trait in restJson. We should instead be sending on an empty JSON body: `{}`.

## Testing
- [x] Protocol tests
- [ ] Manually review codegen diff

## Checklist

### DIFF
--- a/aws/sdk/integration-tests/s3/tests/signing-it.rs
+++ b/aws/sdk/integration-tests/s3/tests/signing-it.rs
@@ -23,7 +23,7 @@ async fn test_signer() -> Result<(), aws_sdk_s3::Error> {
         .build();
     let conn = TestConnection::new(vec![(
         http::Request::builder()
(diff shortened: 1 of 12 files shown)


Velfi (approved):

Velfi:
That diff generator is so helpful for reviewing these, thanks again for setting that up

crisidev (approved):
LGTM!

PR #825, opened 2021-11-02
Add pre-commit hook to fix code gen block quote indentation

jdisanti:
## Motivation and Context
The block quote indentation in the Kotlin codegen is all over the place, and many PRs make changes to it. This script will keep it normalized and looking neat.

_Note: because I took a line-based approach, there are some cases where block quotes and block comments are combined on the same line that can confuse the script. I added a safeguard to bail out in this case, but I don't anticipate us doing weird stuff like this that would trigger it:_

```kotlin
foo = /* something */ """
   asdf
"""
```

If that does happen, the script won't modify the file, but will tell you to go fix it.

## Testing
- Unit tested the script
- Ran against all of smithy-rs and manually inspected the results
- Verified Ktlint is happy with the output

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,13 @@ repos:
   - id: check-yaml
   - id: end-of-file-fixer
   - id: trailing-whitespace
+- repo: local
+  hooks:
+  - id: kotlin-block-quotes
+    name: Kotlin Block Quotes
+    entry: ./.pre-commit-hooks/kotlin-block-quotes.py
+    language: python
+    files: ^.*\.kt$
 - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
   rev: v1.6.1
   hooks:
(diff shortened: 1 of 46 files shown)


Velfi (approved):

.pre-commit-hooks/kotlin-block-quotes.py
```diff
@@ -0,0 +1,271 @@
...
+    line = line[old_indent.start(0):]
+    return (" " * indent) + line
+
+# Parser state.
+class State(Enum):
+    Default = 0 # Just started, or not inside a block comment or block quote
+    InsideBlockComment = 1
+    InsideBlockQuote = 2
+
+# Fixes block quote indentation and returns True if changes were made
```
jdisanti, line 83:
```suggestion
# Fixes block quote indentation and returns a list of lines changed
```

.pre-commit-config.yaml
```diff
@@ -5,6 +5,13 @@ repos:
   - id: check-yaml
   - id: end-of-file-fixer
   - id: trailing-whitespace
+- repo: local
```
rcoh, line 8: TIL this was a thing!

rcoh (approved):

PR #826, opened 2021-11-02
Fix docs on union converters

rcoh:
## Motivation and Context
Documentation on unions is currently buggy and linking to wrong thing:

Screen Shot 2021-11-02 at 2 40 16 PM

## Description
Fix the generated docs for unions

## Testing

Screen Shot 2021-11-02 at 2 39 53 PM

## Checklist
- [x] I have reviewed the generated code diff

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/UnionGenerator.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/UnionGenerator.kt
@@ -53,7 +53,7 @@ class UnionGenerator(
                 if (sortedMembers.size == 1) {
                     Attribute.Custom("allow(irrefutable_let_patterns)").render(this)
                 }
-                rust("/// Tries to convert the enum instance into [`$variantName`](#T::$variantName).", unionSymbol)
+                rust("/// Tries to convert the enum instance into [`$variantName`](#T::$variantName), extracting the inner #D.", unionSymbol, memberSymbol)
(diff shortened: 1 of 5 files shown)


Velfi (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/UnionGenerator.kt
```diff
@@ -47,18 +47,18 @@ class UnionGenerator(
...
-                val variantName = member.memberName.toPascalCase()
+                val variantName = symbolProvider.toMemberName(member).toPascalCase()
+                val funcNamePart = variantName.toSnakeCase()
 
                 if (sortedMembers.size == 1) {
                     Attribute.Custom("allow(irrefutable_let_patterns)").render(this)
                 }
-                rust("/// Tries to convert the enum instance into its #D variant.", unionSymbol)
-                rust("/// Returns `Err(&Self) if it can't be converted.` ")
+                rust("/// Tries to convert the enum instance into [`$variantName`](#T::$variantName).", unionSymbol)
```
jdisanti, line 56: This looks incorrect to me? In the screenshot, it should have been `bool` instead of `Bool`.

rcoh: hmm that's a good point since the API actually returns the inner type. I linked to the outer type because that's where the docs are

PR #818, opened 2021-10-29
Unknown variants

rcoh:
## Motivation and Context
- Fixes #386 
- Fixes #185 
- Generate better docs when we rename structure fields

## Description
A few changes in sequence here, the commit-by-commit review may be helpful:
1. I refactored the way that we produce member names to make it possible to track the original name for structure members and union variants.
2. Generate errors when serializing unknown union variants in JSON/XML. This required some minor refactorings of the errors we generate during serialization.
3. Add support to JSON and XML for parsing unknown union variants. When code generating in Server mode, an unknown union variant produces an error.

## Testing
- [x] UT/IT

## Checklist
- [x] I have updated `CHANGELOG.md`
- [x] I have updated `aws/SDK_CHANGELOG.md` if applicable

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #827, opened 2021-11-02
Use the same protocol dispatcher logic used in the client codegen

crisidev:
## Motivation and Context
Migrate the server-codegen to the same protocol dispatcher mechanism used in the client. The operation traits are generate by `ServerHttpProtocolGenerator` which uses the more specific `ServerRestJson` implementation for things such has content-type, serializers, parser, etc..

This will open the possibility to implement more protocols.

**NOTE: `The RestJson1.kt` file is completely integrated inside `ServerHttpProtocolGenerator.kt`**

## Description
The server protocol dispatch and code generation is implemented by
having the HTTP logic inside ServerHttpProtocolGenerator and using
specification we can apply specific ser/de and code.
Protocols (like ServerRestJson) can inherit from ServerHttpProtocolGenerator
and tune their implementations.

Other class that have been added:
* The ProtocolLoader is implemented inside ServerProtocolLoader
  (ServerProtocol.kt)
  It matches the right protocol and casts it to
  a ServerProtocolGenerator that can be used inside the
  ProtocolGeneratorFactory.
* The ProtocolGenerator is implemented inside ServerProtocolGenerator (
  ServerProtocolGenerator.kt)
  It implements serverProtocolOperation() that can build the
  operation definition as we like. This method is than called by
  ServiceGenerator per every operation in the model.
* RestJson and RestJsonFactory are implemented in RestJson.kt and
  inherit from ServerHttpProtocolGenerator
* Implement the HTTP protocol definition and adapt tests


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+                                if let Some(message) = $data.message() {
+                                    object.key(${"message".dq()}).string(message);
+                                }
+                            """.trimIndent()
+                        )
+                    }
+                    val bindings = httpBindingResolver.errorResponseBindings(it)
+                    bindings.forEach { binding ->
+                        when (val location = binding.location) {
+                            HttpLocation.RESPONSE_CODE, HttpLocation.DOCUMENT -> {}
```
david-perez, line 333: I don't understand this arm. Why are we not emitting a warning if we're not doing anything with these bindings like we do for the rest?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+            else -> {
+                logger.warning(
+                    "[rust-server-codegen] $operationShape: request deserialization does not currently support $location bindings"
+                )
+                null
+            }
+        }
+    }
+
+    private fun serverRenderPathParser(writer: RustWriter, operationShape: OperationShape) {
```
david-perez, line 482:
```suggestion
    private fun serverRenderUriPathParser(writer: RustWriter, operationShape: OperationShape) {
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+            else -> generateParseLabelPrimitiveFn(binding)
+        }
+    }
+
+    private fun generateParseLabelStringFn(binding: HttpBindingDescriptor): RuntimeType {
+        val output = symbolProvider.toSymbol(binding.member)
+        val fnName = generateParseLabelFnName(binding)
+        return RuntimeType.forInlineFun(fnName, operationDeserModule) { writer ->
+            writer.rustBlockTemplate(
+                "##[allow(dead_code)] pub fn $fnName(value: &str) -> std::result::Result<#{O}, #{Error}>",
```
david-perez, line 566: Why not insert above:

```
Attribute.AllowUnused.render(it)
```

where `Attribute` is defined in `RustTypes.kt`.

Same comment goes for all other uses of `#[allow(dead_code)]` in this file.

codegen-server-test/build.gradle.kts
```diff
@@ -28,9 +28,10 @@ val CodegenTests = listOf(
 )
 
 /**
- * The fluent client is generated to prevent warnings in RustDoc since the client is
- * referenced by multiple documentations.
- * TODO: review client generation in the future.
+ * The fluent HAVE to be set to false as we are not generating all the supporting
+ * code for it.
+ * TODO: Review how can we make this a default in the server so that customers don't
+ *       have to specify it.
```
rcoh, line 34: you'll be able to use the `mode` field in codegen context (it lands in #818 )

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerProtocolLoader.kt
```diff
@@ -0,0 +1,45 @@
...
+        serviceShape: ServiceShape
+    ): Pair<ShapeId, ProtocolGeneratorFactory<ServerProtocolGenerator>> {
+        val protocols: MutableMap<ShapeId, Trait> = ServiceIndex.of(model).getProtocols(serviceShape)
+        val matchingProtocols =
+            protocols.keys.mapNotNull { protocolId -> supportedProtocols[protocolId]?.let { protocolId to it } }
+        if (matchingProtocols.isEmpty()) {
+            throw CodegenException("No matching protocol — service offers: ${protocols.keys}. We offer: ${supportedProtocols.keys}")
+        }
+        val pair = matchingProtocols.first()
+        // TODO: is there a better way than an unsafe cast here?
```
david-perez, line 34: This is happening because `ProtocolMap` is defined as

```kotlin
typealias ProtocolMap = Map<ShapeId, ProtocolGeneratorFactory<ProtocolGenerator>>
```

and this function returns a pair whose second component is of type `ProtocolGeneratorFactory<ServerProtocolGenerator>`. So the cast is an unsafe cast casting `ProtocolGeneratorFactory<ProtocolGenerator>` to `ProtocolGeneratorFactory<ServerProtocolGenerator>`.

---

We can define our own

```
typealias ServerProtocolMap = Map<ShapeId, ProtocolGeneratorFactory<ServerProtocolGenerator>>
```

in this file, or write the type in full in the signature of this function. The cast will then become unnecessary.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerProtocolLoader.kt
```diff
@@ -0,0 +1,45 @@
...
+        return Pair(pair.first, pair.second as ProtocolGeneratorFactory<ServerProtocolGenerator>)
+    }
+
+    companion object {
+        val DefaultProtocols = mapOf(
+            // TODO: support other protocols.
+            RestJson1Trait.ID to ServerRestJsonFactory(),
+        )
+        val Default = ServerProtocolLoader(DefaultProtocols)
```
david-perez, line 43: `Default` is never used in the project, only `DefaultProtocols`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolGenerator.kt
```diff
@@ -0,0 +1,81 @@
...
+     */
+    fun serverRenderOperation(
+        operationWriter: RustWriter,
+        operationShape: OperationShape,
+        customizations: List<OperationCustomization>
+    ) {
+        // impl OperationInputShape { ... }
+        val operationName = symbolProvider.toSymbol(operationShape).name
+        // pub struct Operation { ... }
+        operationWriter.rust(
```
rcoh, line 61: you may find the `docs` method helpful

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerRestJson.kt
```diff
@@ -0,0 +1,113 @@
...
+            requestBodyDeserialization = true,
+            responseSerialization = true,
+            errorSerialization = true
+        )
+    }
+}
+
+/*
+ * RestJson1 implementation. This class is fed to the `ServerHttpProtocolGenerator` and configure things like
+ * wire-protocol serializers, content-type, etc..
```
david-perez, line 58:
```suggestion
 * RestJson1 implementation. This class is fed to the `ServerHttpProtocolGenerator` to configure things like
 * wire-protocol serializers, content-type, etc.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -178,36 +179,22 @@ class ProtocolTestGenerator(
             rust("/* test case disabled for this protocol (not yet supported) */")
             return
         }
-        val customToken = if (inputShape.findMemberWithTrait<IdempotencyTokenTrait>(codegenContext.model) != null) {
-            """.make_token("00000000-0000-4000-8000-000000000000")"""
-        } else ""
-        rust(
-            """let config = #T::Config::builder()$customToken.build();""",
-            RuntimeType.Config
-        )
```
rcoh, line 187: huh, I guess you never need to _generate_ an idempotency token on the server side. Makes sense!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -260,26 +246,19 @@ class ProtocolTestGenerator(
...
         rustTemplate(
             """
-            use #{parse_http_response};
-            let parser = #{op}::new();
-            let parsed = parser.parse_loaded(&http_response);
+            use #{ParseStrictResponse};
+            use #{SerializeHttpResponse};
+            let op = #{op}::new();
+            let http_response = op.serialize(&expected_output).expect("unable to serialize response body");
+            let parsed = op.parse(&http_response);
```
rcoh, line 258: ah! I get it now. We'll need to add response-specific code to `aws-smithy-protocol-test` 👍🏻

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+    protocol: Protocol,
+) : ServerProtocolGenerator(
+    codegenContext,
+    protocol,
+    MakeOperationGenerator(codegenContext, protocol, HttpBoundProtocolBodyGenerator(codegenContext, protocol)),
+    ServerHttpProtocolImplGenerator(codegenContext, protocol),
+)
+
+/*
+ * Class used to expose Rust server traits types. Is is used in ServerHttpProtocolGenerator and ServerProtocolTestGenerator.
```
rcoh, line 69: a very minor note: if you put classes in `[]`, eg. `[ServerHttpProtocolGenerator]`, IDEs will make it a link (at least intellij does)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+ */
+class HttpServerTraits {
```
rcoh, line 71: I would either:
- have `runtimeConfig` be an argument to this class
- put the methods on the companion object

david-perez:
Looks good, minor comments. Thanks for sprinkling comments over the codebase.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
rcoh: the first one would probably make calling this the cleanest

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+        "ParseStrictResponse" to RuntimeType.parseStrictResponse(runtimeConfig),
+        "SerializeHttpResponse" to httpServerTraits.serializeHttpResponse(runtimeConfig),
+        "SerializeHttpError" to httpServerTraits.serializeHttpError(runtimeConfig),
+        "JsonObjectWriter" to smithyJson.member("serialize::JsonObjectWriter"),
+        "Http" to RuntimeType.http,
+        "Bytes" to RuntimeType.Bytes,
+        "Error" to errorType.member("Error"),
+        "LazyStatic" to CargoDependency("lazy_static", CratesIo("1.4")).asType(),
+        "Regex" to CargoDependency("regex", CratesIo("1.0")).asType(),
+        "PercentEncoding" to CargoDependency("percent-encoding", CratesIo("2.1.0")).asType(),
```
rcoh, line 121: I would probably move these to `CargoDependency`. That way, it's obvious what version of each crate we're using and we don't end up with version mismatches

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+    private val smithyJson = CargoDependency.smithyJson(runtimeConfig).asType()
+    private val httpServerTraits = HttpServerTraits()
+
+    private val codegenScope = arrayOf(
+        "ParseHttpRequest" to httpServerTraits.parseHttpRequest(runtimeConfig),
+        "ParseStrictResponse" to RuntimeType.parseStrictResponse(runtimeConfig),
+        "SerializeHttpResponse" to httpServerTraits.serializeHttpResponse(runtimeConfig),
+        "SerializeHttpError" to httpServerTraits.serializeHttpError(runtimeConfig),
+        "JsonObjectWriter" to smithyJson.member("serialize::JsonObjectWriter"),
+        "Http" to RuntimeType.http,
```
rcoh, line 116: nit: I would name this `http` — it's the http module, as opposed to a struct
```suggestion
        "http" to RuntimeType.http,
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+
+    /*
+     * TODO: implement straming traits
+     */
+    private fun RustWriter.renderStreamingTraits(
+        operationName: String,
+        outputSymbol: Symbol,
+        operationShape: OperationShape
+    ) {
+        logger.warning(
```
rcoh, line 232: _generally_ I find that it's useful to generate code with `todo!("some message") /* generated comment indicating what the problem was */` in it, although I guess in this case, that doesn't quite work.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+        }
+    }
+
+    private fun serverSerializeError(operationShape: OperationShape): RuntimeType {
+        val fnName = "serialize_${operationShape.id.name.toSnakeCase()}_error"
+        val errorSymbol = operationShape.errorSymbol(symbolProvider)
+        return RuntimeType.forInlineFun(fnName, operationSerModule) {
+            Attribute.Custom("allow(clippy::unnecessary_wraps)").render(it)
+            it.rustBlockTemplate(
+                "pub fn $fnName(error: &#{E}Kind) -> std::result::Result<#{Http}::Response<#{Bytes}>, #{Error}>",
```
rcoh, line 286: we should probably add a helper to the error generator that can make the error kind symbol...

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+                        )
+                    }
+                    val bindings = httpBindingResolver.errorResponseBindings(it)
+                    bindings.forEach { binding ->
+                        when (val location = binding.location) {
+                            HttpLocation.RESPONSE_CODE, HttpLocation.DOCUMENT -> {}
+                            else -> {
+                                logger.warning(
+                                    "$operationShape: response serialization does not currently support $location bindings"
+                                )
```
rcoh, line 337: rather than generating nothing here, I'd generate a comment or a `todo!()`. Makes it easier to debug the generated code

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -0,0 +1,642 @@
...
+                null
+            }
+            HttpLocation.DOCUMENT -> {
+                // document is handled separately
+                null
+            }
+            HttpLocation.RESPONSE_CODE -> writable {
+                rustTemplate(
+                    """
+                        let status = output.${member.memberName.toLowerCase()}
```
rcoh, line 415: you need to use `symbolProvider.toMemberName(member)`—using shape fields directly won't work for fields that get escaped

rcoh: (It will also handle the lower casing for you), lowercase is also not right, it's snake case, btw.

rcoh (commented):
overall looks fine, a few issues that will cause codegen failures on more complex models

codegen-server-test/build.gradle.kts
```diff
@@ -28,9 +28,10 @@ val CodegenTests = listOf(
 )
 
 /**
- * The fluent client is generated to prevent warnings in RustDoc since the client is
- * referenced by multiple documentations.
- * TODO: review client generation in the future.
+ * The fluent HAVE to be set to false as we are not generating all the supporting
+ * code for it.
+ * TODO: Review how can we make this a default in the server so that customers don't
```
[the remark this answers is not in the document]
crisidev: Agree. Since it is not in the scope of this PR, I have created an issue: https://github.com/awslabs/smithy-rs/issues/828

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
```diff
@@ -224,13 +175,11 @@ class CodegenVisitor(context: PluginContext, private val codegenDecorator: RustC
...
         rustCrate.useShapeWriter(shape) { writer ->
             StructureGenerator(model, symbolProvider, writer, shape).render()
-            if (!shape.hasTrait<SyntheticInputTrait>()) {
-                val builderGenerator =
-                    BuilderGenerator(codegenContext.model, codegenContext.symbolProvider, shape)
-                builderGenerator.render(writer)
-                writer.implBlock(shape, symbolProvider) {
-                    builderGenerator.renderConvenienceMethod(this)
-                }
+            val builderGenerator =
```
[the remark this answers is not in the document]
crisidev: In the client codegen, input builders are generated during operation generation and we do not really need that, so I have just moved the input/output generations here and removed the conditional.

Client codegen ref: https://github.com/awslabs/smithy-rs/blob/main/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/ProtocolGenerator.kt#L125-L164

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolGenerator.kt
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * This primarily relies on the [traitGenerator] to generate implementations of the `ParseHttpRequest`,
+     * `SerializeHttpResponse` and `SerializeHttpError` traits for the operations
+     */
+    fun serverRenderOperation(
+        operationWriter: RustWriter,
+        operationShape: OperationShape,
+        customizations: List<OperationCustomization>
+    ) {
+        // impl OperationInputShape { ... }
```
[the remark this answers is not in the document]
crisidev: Nope, it should go away, will fix!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: Yeah, I see your point, I'll fix it following 1)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: Ah, I did not know. I'll apply suggestions from David and then fix what we have to follow this pattern.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: Makes sense, I'll do it.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: Oh, thanks. I spent a lot on this piece and it never felt right and now I see why :)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: I am not sure why we are just emitting a warning here. I'll follow what Russell wrote below and generate `todo!` in the Rust code.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: Yeah, it's definitely better.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerProtocolLoader.kt
crisidev: Thanks for teaching me some Kotlin 😄 !! I'll do this.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerProtocolLoader.kt
crisidev: Will remove.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
rcoh: (but also, I'm a little suspicious why this code is dead)

crisidev: I think you are right @rcoh.. This code _is_not_ dead. I am removing the markers. This was a leftover from the old RestJson1 implementation

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: In this specific case I don't think I can put a todo! inside rust as it will be not inside a function. This is the top level definition of the streaming type trait.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerProtocolLoader.kt
```diff
@@ -12,18 +12,20 @@ import software.amazon.smithy.model.knowledge.ServiceIndex
 import software.amazon.smithy.model.shapes.ServiceShape
 import software.amazon.smithy.model.shapes.ShapeId
 import software.amazon.smithy.model.traits.Trait
-import software.amazon.smithy.rust.codegen.server.smithy.generators.protocol.ServerProtocolGenerator
+import software.amazon.smithy.rust.codegen.smithy.generators.protocol.ProtocolGenerator
 import software.amazon.smithy.rust.codegen.smithy.protocols.ProtocolGeneratorFactory
 import software.amazon.smithy.rust.codegen.smithy.protocols.ProtocolMap
 
+typealias ServerProtocolMap = Map<ShapeId, ProtocolGeneratorFactory<ServerProtocolGenerator>>
```
david-perez, line 19: Not used.

crisidev: 👍🏼

david-perez (approved):

PR #834, opened 2021-11-04
Fix bug in epoch second parsing and improve tests

jdisanti:
## Motivation and Context
As preparation for removing our dependency on `chrono`, I created a more thorough test suite for the date-time formatting code. Through this, I discovered the epoch seconds parsing has an accuracy bug due to using `f64` for its parsing.

This PR adds the improved test suite, fixes the epoch second parsing issue, and removes trailing zeros from the fraction in HTTP date formatting.

## Testing
- Formatting/parsing test suite generator that compares against a stable implementation (the JDK)

## Checklist


rust-runtime/aws-smithy-types/test_data/date_time_format_test_suite.json
```diff
@@ -0,0 +1,1334 @@
+{
+  "format_date_time": [
+    { "epoch_seconds": -377703389549, "epoch_subsecond_nanos": 1000000, "iso": "-9999-01-20T23:47:31.001Z",
```
rcoh, line 3: I'm a little confused about `formatted`: is that HTTP date? sometimes it's a negative number? Sometimes it's an http date? Can this be named something more descriptive?

rust-runtime/aws-smithy-types/test_data/date_time_format_test_suite.json
```diff
@@ -0,0 +1,1334 @@
+{
+  "format_date_time": [
```
rcoh, line 2: maybe "tests" would be a clearer name for this key?

rust-runtime/aws-smithy-types/test_data/date_time_format_test_suite.json
```diff
@@ -0,0 +1,1334 @@
...
+    { "epoch_seconds": -377703389549, "epoch_subsecond_nanos": 1000000, "iso": "-9999-01-20T23:47:31.001Z",
+      "formatted": null },
+    { "epoch_seconds": -377700707496, "epoch_subsecond_nanos": 10000000, "iso": "-9999-02-21T00:48:24.01Z",
```
rcoh, line 5: seems like fractional epoch seconds should also be a key to parse against?

rust-runtime/aws-smithy-types/test_data/date_time_format_test_suite.json
rcoh: a `description` top level field explaining the fields and the expected test strategy would probably be good

rust-runtime/aws-smithy-types/src/instant/format.rs
```diff
@@ -490,82 +393,193 @@ pub(crate) mod rfc3339 {
...
+        parse_date_time: Vec<TestCase>,
+        parse_http_date: Vec<TestCase>,
+        parse_epoch_seconds: Vec<TestCase>,
+    }
+
+    lazy_static! {
+        static ref TEST_CASES: TestCases = {
+            // This test suite can be regenerated by the following Kotlin class:
+            // `codegen/src/test/kotlin/software/amazon/smithy/rust/tool/TimeTestSuiteGenerator.kt`
+            let json = include_str!("../../test_data/date_time_format_test_suite.json");
```
rcoh, line 429: _slight_ preference for reading it from the filesystem because it can be changed without recompiling

rust-runtime/aws-smithy-types/src/instant/format.rs
```diff
@@ -27,18 +27,69 @@ impl fmt::Display for DateParseError {
...
+                if *b < b'0' || *b > b'9' {
+                    return Err(DateParseError::Invalid("decimal must be alphanumeric"));
+                }
+                decimal_buf[i] = *b;
+            }
+            // Unsafe: the decimal_buf is only populated with ASCII numeric characters above
+            debug_assert!(std::str::from_utf8(&decimal_buf).is_ok());
+            let decimal_str = unsafe { std::str::from_utf8_unchecked(&decimal_buf) };
+            decimal = <u32>::from_str(decimal_str).map_err(|_| DateParseError::IntParseError)?;
+        }
```
rcoh, line 76: should we assert that `parts` is now empty? I think this would parse `123.456.789` and just ignore the end? Could use https://doc.rust-lang.org/nightly/std/primitive.str.html#method.splitn, or:
```rust
let (whole_str, decimal_str) match value.split_once('.') {
  Some((whole, part)) => (whole, Some(part)),
  None => (value, None)
}
```

rust-runtime/aws-smithy-types/src/instant/format.rs
```diff
@@ -27,18 +27,69 @@ impl fmt::Display for DateParseError {
...
+                if i >= 9 {
+                    return Err(DateParseError::Invalid("decimal is too long"));
+                }
+                if *b < b'0' || *b > b'9' {
+                    return Err(DateParseError::Invalid("decimal must be alphanumeric"));
+                }
+                decimal_buf[i] = *b;
+            }
+            // Unsafe: the decimal_buf is only populated with ASCII numeric characters above
+            debug_assert!(std::str::from_utf8(&decimal_buf).is_ok());
```
rcoh, line 73: we can avoid unsafe here but just counting how many zeros need to be added and performing a multiplication on the final result, right?

rcoh (changes requested):

rust-runtime/aws-smithy-types/src/instant/format.rs
jdisanti: Yep. My debate was between this or a for loop that multiplies by 10. Which is uglier?

codegen/src/test/kotlin/software/amazon/smithy/rust/tool/TimeTestSuiteGenerator.kt
```diff
@@ -0,0 +1,176 @@
...
+            val formatted = testCase.formatted?.let { "\"" + it.replace("\"", "\\\"") + "\"" } ?: "null"
+            val secs = testCase.epochSeconds
+            val subsecs = testCase.epochSubsecondNanos
+            """    { "epoch_seconds": $secs, "epoch_subsecond_nanos": $subsecs, "iso": "$iso",
+              |      "formatted": $formatted }""".trimMargin()
+        }
+        result += testCases.joinToString(",\n")
+        result += "\n  ]"
+        result
+    }
```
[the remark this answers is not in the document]
jdisanti: I reached for an `ObjectMapper` and it wasn't there. Didn't think to check Smithy for a transient json library.

rust-runtime/aws-smithy-types/test_data/date_time_format_test_suite.json
```diff
@@ -0,0 +1,1334 @@
...
+      "formatted": null },
+    { "epoch_seconds": -377698198243, "epoch_subsecond_nanos": 100000000, "iso": "-9999-03-22T01:49:17.1Z",
```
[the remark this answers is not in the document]
jdisanti: It already does all of the bullet points listed, so I'm not really sure where the confusion is coming from. Is it just the JSON format that you don't like?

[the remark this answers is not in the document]
jdisanti: Oh, I think I see the confusion now. The `formatted` key is what actually gets parsed or compared against. The `epoch_seconds` key is what is used to construct an `Instant` (so it is canonical). So just confusing naming.

rust-runtime/aws-smithy-types/src/instant/format.rs
rcoh: I'd do the for loop that multiplies by 10, personally. No need for any unsafe or any custom input validation, just multiplying by 10 according to the length of the string.

the wrinkle in this case is trimming spaces from the string

jdisanti:
I added a fuzz test suite for the date time parsing functions, and this revealed that the `http-date` and `date-time` parsers currently panic on bad input. Since some of these panics are happening inside of `chrono`, I'm going to fix them in a follow up PR that removes our dependency on it.

codegen/src/test/kotlin/software/amazon/smithy/rust/tool/TimeTestSuiteGenerator.kt
```diff
@@ -142,35 +145,79 @@ private fun generateDateTimeTests(parsing: Boolean): List<TestCase> {
...
-        }
-        result += testCases.joinToString(",\n")
-        result += "\n  ]"
-        result
-    }
-    println(testSuites.joinToString(",\n"))
-    println("}")
+    val none = SourceLocation.NONE
+    val topLevels = mapOf<String, Node>(
+        "description" to ArrayNode(
```
rcoh, line 150: ❤️ this

codegen/src/test/kotlin/software/amazon/smithy/rust/tool/TimeTestSuiteGenerator.kt
```diff
@@ -142,35 +145,79 @@ private fun generateDateTimeTests(parsing: Boolean): List<TestCase> {
...
+                canonical_seconds: string,
+
+                // The canonical nanosecond adjustment to the canonical number of seconds.
+                // If conversion from (canonical_seconds, canonical_nanos) into a 128-bit integer is required,
+                // DO NOT just add the two together as this will yield an incorrect value when
+                // canonical_seconds is negative.
+                canonical_nanos: number,
+
+                // Will be true if this test case is expected to result in an error or exception
+                error: boolean,
```
rcoh, line 180: don't know if it's worth it, but could make this a string and have a message about why it's an error if it's ever unclear

rcoh (approved):
LGTM!

PR #840, opened 2021-11-08
Improve server HTTP protocol compliance response tests

david-perez:
The response tests were serializing and deserializing back an output
structure. Instead, what they should do is serialize the output
structure into an HTTP response and assert that it has the properties
that were specified in the Smithy model's `@httpResponseTests` trait.

https://awslabs.github.io/smithy/1.0/spec/http-protocol-compliance-tests.html?highlight=httpresponsetest#httpresponsetests

## Checklist


crisidev (approved):
Good!

PR #842, opened 2021-11-08
Add accessor methods to generated structs

jdisanti:
## Motivation and Context
By having accessors rather than direct access to struct members, we will have a lot more flexibility with how data is stored in the structs.

## Checklist

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
@@ -172,6 +172,45 @@ fun RustType.asOptional(): RustType = when (this) {
     else -> RustType.Option(this)
 }
 
+/** Converts type to a reference */
+fun RustType.asRef(): RustType = when (this) {
+    is RustType.Reference -> this
+    is RustType.Option -> RustType.Option(member.asRef())
+    else -> RustType.Reference(null, this)
+}
+
+/** Converts type to its Deref target */
+fun RustType.asDeref(): RustType = when (this) {
+    is RustType.Option -> if (member.isDeref()) {
+        RustType.Option(member.asDeref().asRef())
+    } else {
+        this
+    }
+    is RustType.Box -> RustType.Reference(null, member)
+    is RustType.String -> RustType.Opaque("str")
+    is RustType.Vec -> RustType.Slice(member)
+    else -> this
+}
+
+/** Returns true if the type implements Deref */
+fun RustType.isDeref(): Boolean = when (this) {
+    is RustType.Box -> true
+    is RustType.String -> true
+    is RustType.Vec -> true
+    else -> false
+}
+
+/** Returns true if the type implements Copy */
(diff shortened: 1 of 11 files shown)


jdisanti:
Notes from the generated diff:
- Remove empty impl blocks
- This is wrong:
```rust
pub fn resource_types(&self) -> std::option::Option<&[std::string::String]> {
    self.resource_types.as_deref()
}
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
```diff
@@ -172,6 +172,45 @@ fun RustType.asOptional(): RustType = when (this) {
...
 
+/** Converts type to a reference */
+fun RustType.asRef(): RustType = when (this) {
+    is RustType.Reference -> this
+    is RustType.Option -> RustType.Option(member.asRef())
+    else -> RustType.Reference(null, this)
+}
+
+/** Converts type to its Deref target */
+fun RustType.asDeref(): RustType = when (this) {
```
rcoh, line 183: given ref and deref which are so similar, can you add a couple examples to the docs? of how it's different?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
```diff
@@ -113,6 +120,42 @@ class StructureGenerator(
...
+                            rust("self.$memberName")
+                        } else if (memberType is RustType.Option && memberType.member.isDeref()) {
+                            rust("self.$memberName.as_deref()")
+                        } else if (memberType is RustType.Option) {
+                            rust("self.$memberName.as_ref()")
+                        } else if (memberType.isDeref()) {
+                            rust("self.$memberName.deref()")
+                        } else {
+                            rust("&self.$memberName")
+                        }
```
rcoh, line 152: totally up to you, but for these long chains, sometimes a when block where the LHS and RHS can be on the same line are easier to read:
```kotlin
                        when {
                            memberType.isCopy() -> rust("self.$memberName")
                            // ...
                         }
```

rcoh (approved):
LGTM!

rcoh:
CI failure is because we need to add `builder` to the list of words

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
```diff
@@ -113,6 +120,36 @@ class StructureGenerator(
         }
     }
 
+    private fun renderStructureImpl() {
+        if (members.isEmpty()) {
+            return
+        }
+        writer.rustBlock("impl $name") {
+            // Render field accessor methods
+            forEachMember { member, memberName, memberSymbol ->
```
rcoh, line 129: this isn't quite right. It will make an empty impl block for errors. Could add an extension function to produce a list of members we're actually going to render?

rcoh:
LGTM!

PR #844, opened 2021-11-09
Update MSRV to 1.54.0

crisidev:
## Motivation and Context
The codegen-server project will soon depend on `axum 0.3`, which requires MSRV 1.54.0.
This will allow us to implement `FromRequest` / `IntoResponse` axum traits.

Moving to 1.54.0 requires some changes to make clippy happy, especially around references that get immediately derefenced by the compiler. This required updates on both `rust-runtime` and `codegen`.

## Testing
Tested locally and waiting for CI to run.

## Checklist

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/MakeOperationGenerator.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/MakeOperationGenerator.kt
@@ -69,7 +69,7 @@ open class MakeOperationGenerator(
         val takesOwnership = bodyGenerator.bodyMetadata(shape).takesOwnership
         val mut = customizations.any { it.mutSelf() }
         val consumes = customizations.any { it.consumesSelf() } || takesOwnership
-        val self = "self".letIf(mut) { "mut $it" }.letIf(!consumes) { "&$it" }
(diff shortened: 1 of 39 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/MakeOperationGenerator.kt
```diff
@@ -69,7 +69,7 @@ open class MakeOperationGenerator(
         val takesOwnership = bodyGenerator.bodyMetadata(shape).takesOwnership
         val mut = customizations.any { it.mutSelf() }
         val consumes = customizations.any { it.consumesSelf() } || takesOwnership
-        val self = "self".letIf(mut) { "mut $it" }.letIf(!consumes) { "&$it" }
+        val self = "self".letIf(mut) { "mut $it" }.letIf(!consumes) { "$it" }
```
rcoh, line 72: I think the last branch can be removed since it doesn't actually change the input

rcoh (approved):
LGTM! There is one branch that I think can be removed now.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/MakeOperationGenerator.kt
crisidev: You are right, fixed!

PR #839, opened 2021-11-05
Implement FromRequest / ToResponse traits for inputs/outputs/errors.

crisidev:
Add server rejection and error management.

Signed-off-by: Bigo <1781140+crisidev@users.noreply.github.com>

## Checklist
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated `aws/SDK_CHANGELOG.md` if applicable

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #850, opened 2021-11-10
Add server operation registry and router

david-perez:
This commit adds two things:
1. A runtime router implementing `tower`'s [`Service`](https://docs.rs/tower-service/0.3.1/tower_service/trait.Service.html) that adheres to [Smithy's `http` trait specification](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#http-trait), that is linear in the number of registered routes.
2. A code-generated "operation registry" that allows service implementers to provide Rust functions and declare them as the handlers for their service's operations.

The framework will receive HTTP requests from the server and route them to the corresponding operation handler.

---

~~The PR is far from being mergeable, but it builds. I also want to gather feedback, especially for `OperationRegistry.kt`.~~ PR is now mergeable.

Another PR implementing `axum`'s `FromRequest` and `IntoResponse` for operations needs to be merged first, ~~on which @crisidev is currently working.~~ which was done on #839.

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
@@ -1,0 +1,129 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package software.amazon.smithy.rust.codegen.server.smithy.generators
+
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.rust.codegen.rustlang.*
+import software.amazon.smithy.rust.codegen.server.smithy.protocols.ServerHttpProtocolGenerator
+import software.amazon.smithy.rust.codegen.smithy.CodegenContext
+import software.amazon.smithy.rust.codegen.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.smithy.RuntimeType.Companion.RequestSpecModule
+import software.amazon.smithy.rust.codegen.smithy.generators.error.errorSymbol
+import software.amazon.smithy.rust.codegen.smithy.protocols.HttpBindingResolver
+import software.amazon.smithy.rust.codegen.smithy.protocols.HttpTraitHttpBindingResolver
+import software.amazon.smithy.rust.codegen.smithy.protocols.ProtocolContentTypes
+import software.amazon.smithy.rust.codegen.util.inputShape
+import software.amazon.smithy.rust.codegen.util.outputShape
+import software.amazon.smithy.rust.codegen.util.toSnakeCase
+
+/**
+ * OperationRegistryGenerator
+ */
+class OperationRegistryGenerator(
+    codegenContext: CodegenContext,
+    private val operations: List<OperationShape>,
+) {
+    private val serverCrate = "aws_smithy_http_server"
+    private val service = codegenContext.serviceShape
+    private val model = codegenContext.model
+    private val symbolProvider = codegenContext.symbolProvider
+    private val operationNames = operations.map { symbolProvider.toSymbol(it).name.toSnakeCase() }
+    private val runtimeConfig = codegenContext.runtimeConfig
+    private val codegenScope = arrayOf(
+        "Router" to RuntimeType.Router(runtimeConfig),
+    )
+    private val httpBindingResolver: HttpBindingResolver =
+        HttpTraitHttpBindingResolver(codegenContext.model, ProtocolContentTypes.consistent("application/json"))
+
+    fun render(writer: RustWriter) {
+        Attribute.Derives(setOf(RuntimeType.Debug, RuntimeType.DeriveBuilder)).render(writer)
+        Attribute.Custom("builder(pattern = \"owned\")").render(writer)
+        // Generic arguments of the `OperationRegistryBuilder<Fun0, Fut0, ..., FunN, FutN>`.
+        val operationsGenericArguments = operations.mapIndexed { i, _ -> "Fun$i, Fut$i"}.joinToString()
+        val operationRegistryName = "${service.getContextualName(service)}OperationRegistry<${operationsGenericArguments}>"
+        writer.rustBlock("""
+            pub struct $operationRegistryName
+            where
+                ${operationsTraitBounds()}
+            """.trimIndent()) {
+            val members = operationNames
+                .mapIndexed { i, operationName -> "$operationName: Fun$i" }
+                .joinToString(separator = ",\n")
+            rust(members)
+        }
+
+        writer.rustBlockTemplate("""
+            impl<${operationsGenericArguments}> From<$operationRegistryName> for #{Router}
+            where
+                ${operationsTraitBounds()}
+            """.trimIndent(), *codegenScope) {
+            rustBlock("fn from(registry: ${operationRegistryName}) -> Self") {
+                val operationInOutWrappers = operations.map {
+                    val operationName = symbolProvider.toSymbol(it).name
+                    Pair("crate::operation::$operationName${ServerHttpProtocolGenerator.OPERATION_INPUT_WRAPPER_SUFFIX}",
+                    "crate::operation::$operationName${ServerHttpProtocolGenerator.OPERATION_OUTPUT_WRAPPER_SUFFIX}")
+                }
+                val requestSpecsVarNames = operationNames.map { "${it}_request_spec" }
+                val routes = requestSpecsVarNames.zip(operationNames).zip(operationInOutWrappers) { (requestSpecVarName, operationName), (inputWrapper, outputWrapper) ->
+                    ".route($requestSpecVarName, $serverCrate::routing::operation_handler::operation::<_, _, $inputWrapper, _, $outputWrapper>(registry.$operationName))"
+                }.joinToString(separator = "\n")
+
+                val requestSpecs = requestSpecsVarNames.zip(operations) { requestSpecVarName, operation ->
+                    "let $requestSpecVarName = ${operation.requestSpec()};"
+                }.joinToString(separator = "\n")
+                rustTemplate("""
+                    $requestSpecs
+                    #{Router}::new()
+                        $routes
+                    """.trimIndent(), *codegenScope)
+            }
+        }
+    }
+
+    private fun operationsTraitBounds(): String = operations
+        .mapIndexed { i, operation ->
+            val outputType = if (operation.errors.isNotEmpty()) {
+                "Result<${symbolProvider.toSymbol(operation.outputShape(model)).fullName}, ${operation.errorSymbol(symbolProvider).fullyQualifiedName()}>"
+            } else {
+                symbolProvider.toSymbol(operation.outputShape(model)).fullName
+            }
+            """
+            Fun$i: FnOnce(${symbolProvider.toSymbol(operation.inputShape(model))}) -> Fut$i + Clone + Send + Sync + 'static,
+            Fut$i: std::future::Future<Output = $outputType> + Send
+            """.trimIndent()
+        }.joinToString(separator = ",\n")
+
+    private fun OperationShape.requestSpec(): String {
+        val httpTrait = httpBindingResolver.httpTrait(this)
+        val namespace = RequestSpecModule(runtimeConfig).fullyQualifiedName()
+
+        // TODO Support the `endpoint` trait: https://awslabs.github.io/smithy/1.0/spec/core/endpoint-traits.html#endpoint-trait
+
+        val pathSegments = httpTrait.uri.segments.map {
+            "$namespace::PathSegment::" +
+                if (it.isGreedyLabel) "Greedy"
+                else if (it.isLabel) "Label"
+                else "Literal(String::from(\"${it.content}\"))"
+        }
+        val querySegments = httpTrait.uri.queryLiterals.map {
+            "$namespace::QuerySegment::" +
+                if (it.value == "") "Key(String::from(\"${it.key}\"))"
+                else "KeyValue(String::from(\"${it.key}\"), String::from(\"${it.value}\"))"
+        }
+
+        return """
+            $namespace::RequestSpec::new(
+                http::Method::${httpTrait.method},
+                $namespace::UriSpec {
+                    host_prefix: None,
+                    path_and_query: $namespace::PathAndQuerySpec {
+                        path_segments: $namespace::PathSpec::from_vector_unchecked(vec![${pathSegments.joinToString()}]),
+                        query_segments: $namespace::QuerySpec::from_vector_unchecked(vec![${querySegments.joinToString()}])
+                    }
+                }
+            )""".trimIndent()
+    }
+}
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
@@ -36,7 +36,7 @@ class ServerServiceGenerator(
      */
     fun render() {
         val operations = index.getContainedOperations(context.serviceShape).sortedBy { it.id }
-        operations.map { operation ->
+        for (operation in operations) {
             rustCrate.useShapeWriter(operation) { operationWriter ->
                 protocolGenerator.serverRenderOperation(
                     operationWriter,
@@ -53,6 +53,10 @@ class ServerServiceGenerator(
                         .render(writer)
                 }
             }
+        }
+        rustCrate.withModule(RustModule.public("operation_registry", "A registry of your service's operations.")) { writer ->
+            OperationRegistryGenerator(context, operations)
+                .render(writer)
         }
     }
 }
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
@@ -192,6 +192,7 @@ data class CargoDependency(
         val Axum: CargoDependency = CargoDependency("axum", CratesIo("0.3"))
         val Bytes: CargoDependency = CargoDependency("bytes", CratesIo("1"))
         val BytesUtils: CargoDependency = CargoDependency("bytes-utils", CratesIo("0.1.1"))
+        val DeriveBuilder = CargoDependency("derive_builder", CratesIo("0.10"))
         val FastRand: CargoDependency = CargoDependency("fastrand", CratesIo("1"))
         val Hex: CargoDependency = CargoDependency("hex", CratesIo("0.4.3"))
         val HttpBody: CargoDependency = CargoDependency("http-body", CratesIo("0.4"))
@@ -213,7 +214,6 @@ data class CargoDependency(
         fun SmithyEventStream(runtimeConfig: RuntimeConfig) = runtimeConfig.runtimeCrate("eventstream")
         fun SmithyHttp(runtimeConfig: RuntimeConfig) = runtimeConfig.runtimeCrate("http")
         fun SmithyHttpServer(runtimeConfig: RuntimeConfig) = runtimeConfig.runtimeCrate("http-server")
-        fun SmithyHttpTower(runtimeConfig: RuntimeConfig) = runtimeConfig.runtimeCrate("http-tower")
         fun SmithyProtocolTestHelpers(runtimeConfig: RuntimeConfig) =
             runtimeConfig.runtimeCrate("protocol-test").copy(scope = DependencyScope.Dev)
         fun smithyJson(runtimeConfig: RuntimeConfig): CargoDependency = runtimeConfig.runtimeCrate("json")
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
@@ -176,6 +176,13 @@ data class RuntimeType(val name: String?, val dependency: RustDependency?, val n
         val PartialEq = std.member("cmp::PartialEq")
         val StdError = RuntimeType("Error", dependency = null, namespace = "std::error")
         val String = RuntimeType("String", dependency = null, namespace = "std::string")
+        val DeriveBuilder = RuntimeType("Builder", dependency = CargoDependency.DeriveBuilder, namespace = "derive_builder")
+
+        fun RequestSpecModule(runtimeConfig: RuntimeConfig) =
+            RuntimeType("request_spec", CargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::routing")
+
+        fun Router(runtimeConfig: RuntimeConfig) =
+            RuntimeType("Router", CargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::routing")
 
         fun DateTime(runtimeConfig: RuntimeConfig) =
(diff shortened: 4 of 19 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+    // TODO Is using this like this fine?
+    private val httpBindingResolver: HttpBindingResolver =
+        HttpTraitHttpBindingResolver(codegenContext.model, ProtocolContentTypes.consistent("application/json"))
+
+    fun render(writer: RustWriter) {
+        fromRequestIntoResponseWorkaround(writer)
+
+        Attribute.Derives(setOf(RuntimeType.Debug, RuntimeType.DeriveBuilder)).render(writer)
+        // TODO: Should I use `raw` here? I would not have to escape the `#`.
+        writer.rust("##[builder(pattern = \"owned\")]")
```
rcoh, line 43: you can actually just `Attribute.Custom(...).render()`. Raw is also fine, totally up to you

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+        // TODO: is this the correct way of getting the service name?
```
rcoh, line 44: as in a human readable name of the service? _probably_ not, you probably want something out of `RustSettings`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        // TODO Note I'm writing out the fully namespaced names of the types, which one would do reaching for `RuntimeType`.
+        //  I don't foresee that we will use these types anywhere else in the codebase, so is it really worth adding them
+        //  to `RuntimeType.kt`, or creating them here and converting them to strings using `fullyQualifiedName()`,
+        //  as opposed to just interpolating strings like I'm doing here? I get the value of using `RuntimeType` when it's
+        //  an ubiquitous type used throughout the codebase, since it means it's definition is centralized, making it easy
+        //  to modify, but if the use of a type is localized like here, is there any benefit? Interpolating strings reads
+        //  clearer in my opinion.
+        //  I guess there has to be at least one usage of `RuntimeType`; otherwise the dependency won't appear in `Cargo.toml`.
```
rcoh, line 158: 1. yes, the main reason to use `RuntimeType` is so that you ensure you add the dependency, but only when it's actually used. As the project gets more complex, you may end up with certain services using differing subsets of dependencies
2. yes, you can call `addDependency` directly on the writer

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
```diff
@@ -211,11 +211,14 @@ data class CargoDependency(
...
 
-        val Bytes: RustDependency = CargoDependency("bytes", CratesIo("1"))
+        val Bytes = CargoDependency("bytes", CratesIo("1"))
         val TokioStream = CargoDependency("tokio-stream", CratesIo("0.1.7"))
         val BytesUtils = CargoDependency("bytes-utils", CratesIo("0.1.1"))
         val Hex = CargoDependency("hex", CratesIo("0.4.3"))
         val TempFile = CargoDependency("temp-file", CratesIo("0.1.6"), scope = DependencyScope.Dev)
         val Ring = CargoDependency("ring", CratesIo("0.16"))
+        val DeriveBuilder = CargoDependency("derive_builder", CratesIo("0.10"))
+        // TODO Update to 0.3 once we move to 1.54
```
rcoh, line 221: this should be done on main

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
david-perez: This gives me the service name defined in the Smithy model file, which should be appropiate here

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
```diff
@@ -0,0 +1,179 @@
...
+                }.joinToString(separator = "\n")
+
+                val requestSpecs = requestSpecsVarNames.zip(operations) { requestSpecVarName, operation ->
+                    "let $requestSpecVarName = ${operation.requestSpec()};"
+                }.joinToString(separator = "\n")
+                rustTemplate("""
+                    $requestSpecs
+                    #{router}::new()
+                        $routes
+                    """.trimIndent(), "router" to RuntimeType.Router(runtimeConfig))
```
crisidev, line 84: Same as above

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
```diff
@@ -0,0 +1,179 @@
...
+                type Rejection = http::StatusCode;
+
+                async fn from_request(
+                    _req: &mut #{RequestParts}<B>,
+                ) -> Result<Self, Self::Rejection> {
+                    todo!()
+                }
+            }
+        """.trimIndent()
+    }
```
crisidev, line 138: This can go away now!

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
```diff
@@ -213,7 +214,6 @@ data class CargoDependency(
         fun SmithyEventStream(runtimeConfig: RuntimeConfig) = runtimeConfig.runtimeCrate("eventstream")
         fun SmithyHttp(runtimeConfig: RuntimeConfig) = runtimeConfig.runtimeCrate("http")
         fun SmithyHttpServer(runtimeConfig: RuntimeConfig) = runtimeConfig.runtimeCrate("http-server")
-        fun SmithyHttpTower(runtimeConfig: RuntimeConfig) = runtimeConfig.runtimeCrate("http-tower")
```
crisidev, line 216: Why is this gone?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -176,6 +176,18 @@ data class RuntimeType(val name: String?, val dependency: RustDependency?, val n
...
         val String = RuntimeType("String", dependency = null, namespace = "std::string")
+        val DeriveBuilder = RuntimeType("Builder", dependency = CargoDependency.DeriveBuilder, namespace = "derive_builder")
+
+        // Axum types.
+        val FromRequest = RuntimeType("FromRequest", dependency = CargoDependency.Axum, namespace = "axum::extract")
+        val RequestParts = RuntimeType("RequestParts", dependency = CargoDependency.Axum, namespace = "axum::extract")
+        val IntoResponse = RuntimeType("IntoResponse", dependency = CargoDependency.Axum, namespace = "axum::response")
+        val AxumBody = RuntimeType("Body", dependency = CargoDependency.Axum, namespace = "axum::body")
+        val AxumHttpBody = RuntimeType("HttpBody", dependency = CargoDependency.Axum, namespace = "axum::body")
+        val AsyncTrait = RuntimeType("async_trait", dependency = CargoDependency.Axum, namespace = "axum")
```
crisidev, line 187: Oh, this are much nicer that what I have used in the IntoResponse/FromRequest implementation. We can change them there!

rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -1,14 +1,14 @@
 [package]
 name = "aws-smithy-http-server"
 version = "0.1.0"
-authors = ["Smithy Rust Server <smithy-rs-server@amazon.com>"]
+authors = ["Matteo Bigoi <matbigoi@amazon.com>", "David Pérez <davidpz@amazon.com>"]
```
crisidev, line 4: Please leave the alias instead of names. It will make things easier in the future as the email list is already active and contains me and you right now.

rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -20,11 +20,22 @@ aws-smithy-http = { path = "../aws-smithy-http" }
...
 mime = "0.3"
+pin-project-lite = "0.2.7"
+regex = "1.0"
+serde_urlencoded = "0.7"
+thiserror = "1"
+tokio = { version = "1.0", features = ["full"] }
+tower = { version = "0.4" }
+
+# TODO This can be removed, it's in the service crate.
+derive_builder = "0.10"
```
crisidev, line 38: Is this TODO already applicable?

rust-runtime/aws-smithy-http-server/src/clone_box_service.rs
```diff
@@ -0,0 +1,58 @@
+use futures_util::future::BoxFuture;
```
crisidev, line 1: Missing SPDX licence header and module description for Rust doc. If this comes from axum with modification, you will need to follow what I did in `error.rs` and add the axum license inline.

crisidev: I would also add a description of why this module is here and how it is used in the codebase as it is not the most immediate code we have.

rust-runtime/aws-smithy-http-server/src/extractors.rs
```diff
@@ -0,0 +1,35 @@
+// =============================
```
crisidev, line 1: This can go away since my PR has been merged.

rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -20,11 +20,22 @@ aws-smithy-http = { path = "../aws-smithy-http" }
...
 aws-smithy-json = { path = "../aws-smithy-json" }
 axum = { version = "0.3", features = [ "http1", "http2", "headers", "mime", "tower-log" ], default-features = false }
+async-trait = "0.1"
 bytes = "1.1"
+futures-util = { version = "0.3", default-features = false }
 http = "0.2"
 http-body = "0.4"
 hyper = { version = "0.14", features = ["server", "http1", "http2", "tcp"] }
 mime = "0.3"
+pin-project-lite = "0.2.7"
```
crisidev, line 30: From pin-project-lite documentation:
```This is the only reason to use this crate. However, if you already have proc-macro related dependencies in your crate's dependency graph, there is no benefit from using this crate. (Note: There is almost no difference in the amount of code generated between pin-project and pin-project-lite.)```

We already have proc-macro deps coming in with `async-trait` crate. Is there any other reason to use this instead of standard `pin-project` ?

File: rust-runtime/aws-smithy-http-server/Cargo.toml [hunk 14]
```diff
@@ -20,11 +20,22 @@ aws-smithy-http = { path = "../aws-smithy-http" }
...
+regex = "1.0"
+serde_urlencoded = "0.7"
+thiserror = "1"
+tokio = { version = "1.0", features = ["full"] }
```
crisidev, line 34: Tokio is now at 1.13.0. Any reason to not use "1" so we get the latest stable version?

rust-runtime/aws-smithy-http-server/src/handler/mod.rs
```diff
@@ -0,0 +1,55 @@
+use async_trait::async_trait;
```
crisidev, line 1: Missing SPDX licence header and module description for Rust doc.

rust-runtime/aws-smithy-http-server/src/lib.rs
```diff
@@ -1,25 +1,36 @@
-/*
- * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
- * SPDX-License-Identifier: Apache-2.0.
- */
-
```
crisidev, line 5: Please keep the SPDX license identifier header.

rust-runtime/aws-smithy-http-server/src/lib.rs
```diff
@@ -1,25 +1,36 @@
...
 //! HTTP server runtime and utilities, loosely based on Axum.
 
-#![cfg_attr(docsrs, feature(doc_cfg))]
+// These modules are code-generated by `smithy-rs`, one in each file. They are model-specific.
+pub mod extractors;
+pub mod model;
+pub mod operation_registry;
+pub mod responders;
```
crisidev, line 7: This can go away right?

rust-runtime/aws-smithy-http-server/src/macros.rs
```diff
@@ -171,5 +171,50 @@ macro_rules! composite_rejection {
                 }
             }
         }
+    }
+}
+
+macro_rules! opaque_future {
```
crisidev, line 177: If you add a small doc about what is an opaque future, it will reflect into cargo docs, which is nice.

rust-runtime/aws-smithy-http-server/src/main.rs
```diff
@@ -0,0 +1,44 @@
+// ====================
```
crisidev, line 1: This can go away or should be moved into an example if we want to keep it, right? 

Missing SPDX licence header.

rust-runtime/aws-smithy-http-server/src/responders.rs
```diff
@@ -0,0 +1,40 @@
+// =============================
```
crisidev, line 1: This can go away before merging the PR.

rust-runtime/aws-smithy-http-server/src/routing/future.rs
```diff
@@ -0,0 +1,28 @@
+//! Future types.
```
crisidev, line 1: Missing SPDX licence header.

rust-runtime/aws-smithy-http-server/src/runtime.rs
```diff
@@ -0,0 +1,32 @@
+use axum::extract::RequestParts;
```
crisidev, line 1: This can go away before merging the PR. It is already implemented.

rust-runtime/aws-smithy-http-server/src/routing/route.rs
```diff
@@ -0,0 +1,103 @@
+use crate::{
```
crisidev, line 1: Missing SPDX licence header and module description for Rust doc. If this comes from axum with modification, you will need to follow what I did in error.rs and add the axum license inline.

crisidev (changes requested):
I still need to go deeper into the rust implementation of your PR, but I have published some comments you can look at and work on already!

Nice work so far.

rust-runtime/aws-smithy-http-server/src/model.rs
```diff
@@ -0,0 +1,41 @@
+pub struct HealthcheckInput;
```
crisidev, line 1: This can go away as it is autogenerated now from your previous PR.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
```diff
@@ -0,0 +1,179 @@
...
+ */
+class OperationRegistryGenerator(
+    codegenContext: CodegenContext,
+    private val operations: List<OperationShape>,
+) {
+    private val serverCrate = "aws_smithy_http_server"
+    private val service = codegenContext.serviceShape
+    // TODO Should I use symbol provider?
+    private val operationNames = operations
+        .map { operation -> service.getContextualName(operation).toSnakeCase() }
```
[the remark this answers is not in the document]
david-perez: In the rest of the codebase we're using `symbolProvider.toSymbol(operationShape).name`. I'm changing it to that to be consistent. I wonder  what the preferred way of doing this is.

rust-runtime/aws-smithy-http-server/Cargo.toml
david-perez: Yes, it was here to have the example application in `main.rs` of the `aws-smithy-http-server` build.

rust-runtime/aws-smithy-http-server/Cargo.toml
david-perez: Isn't the behavior the same? From https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

> An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping.

rust-runtime/aws-smithy-http-server/src/main.rs
david-perez: I've removed it. We can move it to an example, but it's going to require us manually writing extractors, responders, model, operation registry (essentially everything that we code-generate). So perhaps it's best we do examples as unit tests in `codegen-server`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/OperationRegistryGenerator.kt
```diff
@@ -0,0 +1,179 @@
...
+        writer.rustBlock("""
+            pub struct $operationRegistryName
+            where
+                ${operationsTraitBounds()}
+            """.trimIndent()) {
+            val members = operationNames
+                .mapIndexed { i, operationName -> "$operationName: Fun$i" }
+                .joinToString(separator = ",\n")
+            rust(members)
+        }
```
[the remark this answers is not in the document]
david-perez: Yeah I don't really know if I like it. I saw it used in the codebase and copied it. But, IMO

```kotlin
rustTemplate("""
    pub struct #{Foo} {
        bar: #{Bar},
    }""", *codegenScope)
```

reads better than:

```kotlin
rustBlockTemplate("pub struct #{Foo}", *codegenScope) {
   rustTemplate("bar: #{Bar},", *codegenScope)
}
```

This is another instance where I think there are too many ways in the codebase to generate the same thing and we need simple rules of thumb to homogenize. Perhaps we can discuss with @rcoh. Ideally we'd have all the methods in `RustWriter.kt` documenting when we should favor one over the other.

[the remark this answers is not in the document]
rcoh: I generally agree that not using a nested block is more readable—it's only in the places where you need to loop over something that the second pattern is clearer.

david-perez:
I have resolved all the comments that I have addressed; ready for a second review round.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
```diff
@@ -36,6 +36,7 @@ class ServerServiceGenerator(
      */
     fun render() {
         val operations = index.getContainedOperations(context.serviceShape).sortedBy { it.id }
+        // TODO for loop
```
crisidev, line 39: Out of curiosity what is wrong with this map?

rust-runtime/aws-smithy-http-server/Cargo.toml
crisidev: There is no caret here, it's just "1.0" so for what I understand no, the behaviour is not the same.

david-perez: From https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

> The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

rust-runtime/aws-smithy-http-server/Cargo.toml
david-perez: Nope. Switched to `pin-project`.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
david-perez: We weren't using it.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
david-perez: A pet peeve of mine: `.map` should be used to map a value to another value _without side-effects_. As written this returns `Unit` and has lots of side effects. A for loop communicates intent better.

crisidev: Got it.

rust-runtime/aws-smithy-http-server/Cargo.toml
crisidev: Ah! Ok, forget what I said here.

crisidev (approved):
LGTM

david-perez:
Resolved some clippy lints; hopefully CI passes now.

crisidev (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
```diff
@@ -192,6 +192,7 @@ data class CargoDependency(
         val Axum: CargoDependency = CargoDependency("axum", CratesIo("0.3"))
         val Bytes: CargoDependency = CargoDependency("bytes", CratesIo("1"))
         val BytesUtils: CargoDependency = CargoDependency("bytes-utils", CratesIo("0.1.1"))
+        val DeriveBuilder = CargoDependency("derive_builder", CratesIo("0.10"))
```
rcoh, line 195: in the long term, you _probably_ just want to code generate builders directly, I suspect. `syn` is a very large dependency that ideally, won't be a dependency of the generated code

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -176,6 +176,13 @@ data class RuntimeType(val name: String?, val dependency: RustDependency?, val n
         val PartialEq = std.member("cmp::PartialEq")
         val StdError = RuntimeType("Error", dependency = null, namespace = "std::error")
         val String = RuntimeType("String", dependency = null, namespace = "std::string")
+        val DeriveBuilder = RuntimeType("Builder", dependency = CargoDependency.DeriveBuilder, namespace = "derive_builder")
+
+        fun RequestSpecModule(runtimeConfig: RuntimeConfig) =
+            RuntimeType("request_spec", CargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::routing")
+
+        fun Router(runtimeConfig: RuntimeConfig) =
+            RuntimeType("Router", CargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::routing")
```
rcoh, line 185: may have mentioned this a while ago, but you may at some point want your own companion object in `codegen-server` that contains server-specific runtime types

rcoh (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
david-perez: Good point. https://github.com/awslabs/smithy-rs/issues/864

PR #852, opened 2021-11-10
Prepare `v0.28.0-alpha` release

jdisanti:
This is in response to the compile failures we've seen with the older versions of the AWS SDK pulling in the later versions of the Smithy runtime crates.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,10 +5,10 @@
 
 # Version number to use for the generated SDK
 # Note: these must always be full 3-segment semver versions
-aws.sdk.version=0.0.24-alpha
+aws.sdk.version=0.0.25-alpha
 
 # Version number to use for the generated runtime crates
-smithy.rs.runtime.crate.version=0.27.0-alpha.2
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -5,10 +5,10 @@
 
 # Version number to use for the generated SDK
 # Note: these must always be full 3-segment semver versions
-aws.sdk.version=0.0.24-alpha
+aws.sdk.version=0.0.25-alpha
 
 # Version number to use for the generated runtime crates
-smithy.rs.runtime.crate.version=0.27.0-alpha.2
+smithy.rs.runtime.crate.version=0.28.0-alpha
```
rcoh, line 11: should we start at `-alpha.0` instead of going from `-alpha` to `-alpha.0`?

rcoh (approved):
LGTM, but see note about version

gradle.properties
jdisanti: We decided to go with `-alpha` in an offline conversation.

PR #853, opened 2021-11-11
Generate a Docs Landing Page

rcoh:
## Motivation and Context
We need a top-level landing page for service documentation. 

## Description
This commit refactors a portion of the build into `buildSrc` and adds a job to generate a docs landing page. The generated SDK now contains a `docs.md` that contains top level links to all services.

Totally unrelated but I realized the aws-config fuzzer didn't compile anymore, so I fixed that

## Testing
This is also visible in the generated code diff.
Screen Shot 2021-11-11 at 12 50 01 PM

## Checklist


Velfi (approved):

buildSrc/src/main/kotlin/DocsLandingPage.kt
```diff
@@ -0,0 +1,53 @@
...
+import software.amazon.smithy.utils.CodeWriter
+import java.io.File
+
+/**
+ * Generate a basic docs landing page into [outputDir]
+ *
+ * The generated docs will include links to crates.io, docs.rs and GitHub examples for all generated services. The generated docs will
+ * be written to `docs.md` in the provided [outputDir].
+ */
+fun Project.docsLandingPage(awsServices: List<AwsService>, outputDir: File) {
```
jdisanti, line 11: Are the services always ordered deterministically when passed into this function?

buildSrc/src/main/kotlin/DocsLandingPage.kt
```diff
@@ -0,0 +1,53 @@
...
+            |a crate implementing configuration loading such as credential providers.""".trimMargin()
+        )
+
+        writer.write("## AWS Services")
+        /* generate a basic markdown table */
+        writer.write("| Service | [docs.rs](https://docs.rs) | [crates.io](https://crates.io) | [Usage Examples](https://github.com/awslabs/aws-sdk-rust/tree/main/examples/) |")
+        writer.write("| ------- | ------- | --------- | ------ |")
+        awsServices.forEach {
+            writer.write(
+                "| ${it.humanName} | ${docsRs(it)} | ${cratesIo(it)} | ${
```
jdisanti, line 27: Do we want to order this by human name or by crate name? If the latter, then I think maybe we should put the crate name on the left rather than the human name.

buildSrc/src/main/kotlin/DocsLandingPage.kt
```diff
@@ -0,0 +1,53 @@
+import org.gradle.api.Project
```
jdisanti, line 1: Take a pass on copyright headers. We may just want to grab the Ktlint rule from the Kotlin SDK.

buildSrc/src/main/kotlin/ServiceLoader.kt
```diff
@@ -0,0 +1,115 @@
...
+            if (services.isEmpty()) {
+                logger.info("${file.name} has no services")
+                null
+            } else {
+                val service = services[0]
+                val title = service.expectTrait(TitleTrait::class.java).value
+                val sdkId = service.expectTrait(ServiceTrait::class.java).sdkId
+                    .toLowerCase()
+                    .replace(" ", "")
+                    // TODO: the smithy models should not include the suffix "service"
```
jdisanti, line 34: Is this TODO comment still relevant?

jdisanti (commented):
I love how you split up the build! This is awesome 😄

buildSrc/src/main/kotlin/DocsLandingPage.kt
rcoh: they are, but I just reorded them by human name to be double sure

buildSrc/src/main/kotlin/ServiceLoader.kt
rcoh: yeah as of the current aws-models the service suffix is still included

PR #861, opened 2021-11-12
Relax profile name validation

jdisanti:
## Motivation and Context
This addresses https://github.com/awslabs/aws-sdk-rust/issues/270.

## Testing
- Added additional test cases

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/test-data/default-provider-chain/profile_name/fs/home/.aws/config
+++ b/aws/rust-runtime/aws-config/test-data/default-provider-chain/profile_name/fs/home/.aws/config
@@ -1,0 +1,9 @@
+[default]
+region = us-east-1
+aws_access_key_id = incorrect_key
+aws_secret_access_key = incorrect_secret
+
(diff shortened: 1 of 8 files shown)


rcoh (approved):
LGTM! Do you mind adding a test to the `default_provider` suite (https://github.com/awslabs/smithy-rs/tree/main/aws/rust-runtime/aws-config/test-data/default-provider-chain) that validates this behavior? I just want to be sure that we support this character-set end-to-end (and it's validated). Should be quick since it doesn't need to hit the network

aws/rust-runtime/aws-config/test-data/default-provider-chain/profile_name/fs/home/.aws/config
```diff
@@ -0,0 +1,9 @@
+[default]
+region = us-east-1
+aws_access_key_id = incorrect_key
```
rcoh, line 3: 👍🏻 good idea to have a default profile as part of the test

PR #865, opened 2021-11-16
Add missing features to user agent formatting

jdisanti:
## Description
This PR adds feature, config, framework, and app name formatting to user agents, and also adds the ability to set the app name from service config, shared config, profile config, and environment variables.

## Testing
- Added integration test
- Ran an example to manually verify it works end to end

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/default_provider.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider.rs
@@ -194,7 +194,7 @@ pub mod app_name {
         Builder::default()
     }
 
-    /// Builder for [DefaultRegionChain]
+    /// Default provider builder for [`AppName`]
     #[derive(Default)]
     pub struct Builder {
         env_provider: EnvironmentVariableAppNameProvider,
@@ -219,7 +219,7 @@ pub mod app_name {
             self
         }
 
-        /// Build a [DefaultRegionChain]
+        /// Build an [`AppName`] from the default chain
         pub async fn app_name(self) -> Option<AppName> {
             self.env_provider
                 .app_name()
@@ -243,7 +243,7 @@ pub mod app_name {
             ]);
             let app_name = Builder::default()
                 .configure(
-                    &ProviderConfig::empty()
+                    &ProviderConfig::no_configuration()
                         .with_fs(fs)
                         .with_env(env)
                         .with_http_connector(no_traffic_connector()),
--- a/aws/rust-runtime/aws-config/src/default_provider.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider.rs
@@ -178,6 +178,101 @@ pub mod retry_config {
     }
 }
 
+/// Default app name provider chain
+pub mod app_name {
+    use crate::environment::app_name::EnvironmentVariableAppNameProvider;
+    use crate::profile::app_name;
+    use crate::provider_config::ProviderConfig;
+    use aws_types::app_name::AppName;
+
+    /// Default App Name Provider chain
+    ///
+    /// This provider will check the following sources in order:
+    /// 1. [Environment variables](EnvironmentVariableAppNameProvider)
+    /// 2. [Profile file](crate::profile::app_name::ProfileFileAppNameProvider)
+    pub fn default_provider() -> Builder {
+        Builder::default()
+    }
+
+    /// Builder for [DefaultRegionChain]
+    #[derive(Default)]
+    pub struct Builder {
+        env_provider: EnvironmentVariableAppNameProvider,
+        profile_file: app_name::Builder,
+    }
+
+    impl Builder {
+        #[doc(hidden)]
+        /// Configure the default chain
+        ///
+        /// Exposed for overriding the environment when unit-testing providers
+        pub fn configure(mut self, configuration: &ProviderConfig) -> Self {
+            self.env_provider =
+                EnvironmentVariableAppNameProvider::new_with_env(configuration.env());
+            self.profile_file = self.profile_file.configure(configuration);
+            self
+        }
+
+        /// Override the profile name used by this provider
+        pub fn profile_name(mut self, name: &str) -> Self {
+            self.profile_file = self.profile_file.profile_name(name);
+            self
+        }
+
+        /// Build a [DefaultRegionChain]
+        pub async fn app_name(self) -> Option<AppName> {
+            self.env_provider
+                .app_name()
+                .or(self.profile_file.build().app_name().await)
+        }
+    }
+
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+        use crate::provider_config::ProviderConfig;
+        use crate::test_case::no_traffic_connector;
+        use aws_types::os_shim_internal::{Env, Fs};
+
+        #[tokio::test]
+        async fn prefer_env_to_profile() {
+            let fs = Fs::from_slice(&[("test_config", "[default]\nsdk-ua-app-id = wrong")]);
+            let env = Env::from_slice(&[
+                ("AWS_CONFIG_FILE", "test_config"),
+                ("AWS_SDK_UA_APP_ID", "correct"),
+            ]);
+            let app_name = Builder::default()
+                .configure(
+                    &ProviderConfig::empty()
+                        .with_fs(fs)
+                        .with_env(env)
+                        .with_http_connector(no_traffic_connector()),
+                )
+                .app_name()
+                .await;
+
+            assert_eq!(Some(AppName::new("correct").unwrap()), app_name);
+        }
+
+        #[tokio::test]
+        async fn load_from_profile() {
+            let fs = Fs::from_slice(&[("test_config", "[default]\nsdk-ua-app-id = correct")]);
+            let env = Env::from_slice(&[("AWS_CONFIG_FILE", "test_config")]);
+            let app_name = Builder::default()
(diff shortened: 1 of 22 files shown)


rcoh (commented):
Formatting and app name stuff looks good. We should probably switch to a builder for actually constructing the user agent, I'm sure it will have more fields in the future.

We'll also need _some_ way of enabling the current use case for QLDB of setting fields directly, but we _could_ use app name if we formatted it specially in `User-Agent` (without the `lib/` for now.

aws/rust-runtime/aws-config/src/default_provider.rs
```diff
@@ -178,6 +178,101 @@ pub mod retry_config {
...
+        #[tokio::test]
+        async fn prefer_env_to_profile() {
+            let fs = Fs::from_slice(&[("test_config", "[default]\nsdk-ua-app-id = wrong")]);
+            let env = Env::from_slice(&[
+                ("AWS_CONFIG_FILE", "test_config"),
+                ("AWS_SDK_UA_APP_ID", "correct"),
+            ]);
+            let app_name = Builder::default()
+                .configure(
+                    &ProviderConfig::empty()
```
rcoh, line 246: this should probably be better named but you want `ProviderConfig::no_configuration()` in general in tests

rcoh: (there is no impact here because you override fs and env anyway), but `no_configuration()` ensures that if you _don't_ override those that you won't end up potentially reading the real FS/env during tests causing flakiness and other weird behavior

aws/rust-runtime/aws-config/src/profile/app_name.rs
```diff
@@ -0,0 +1,184 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+//! Load an app name from an AWS profile
+
+use crate::provider_config::ProviderConfig;
+use aws_sdk_sts::AppName;
```
rcoh, line 9: just to not confuse things, we should probably use the one from `aws-types`. We should probably `pub use` AppName from aws-config if we haven't already.

aws/rust-runtime/aws-config/src/profile/app_name.rs
```diff
@@ -0,0 +1,184 @@
...
+///
+/// This provider will attempt to shared AWS shared configuration and then read the
+/// `sdk-ua-app-id` property from the active profile.
+///
+/// # Examples
+///
+/// **Loads "my-app" as the app name**
+/// ```ini
+/// [default]
+/// sdk-ua-app-id = my-app
```
rcoh, line 22: AFAICT, this isn't implemented in other SDKs yet. I'm not aware of any other kebab case properties, so we may want to try to get this changed.

aws/rust-runtime/aws-config/src/profile/app_name.rs
```diff
@@ -0,0 +1,184 @@
...
+            .profile_override
+            .as_deref()
+            .unwrap_or_else(|| profile.selected_profile());
+        let selected_profile = profile.get_profile(selected_profile_name)?;
+        selected_profile
+            .get("sdk-ua-app-id")
+            .map(|name| match AppName::new(name.to_owned()) {
+                Ok(app_name) => Some(app_name),
+                Err(err) => {
+                    tracing::warn!(err = %err, "`sdk-ua-app-id` property in profile {} was invalid", selected_profile_name);
```
rcoh, line 73:
```suggestion
                    tracing::warn!(err = %err, "`sdk-ua-app-id` property in profile `{}` was invalid", selected_profile_name);
```

just to make the error message a little clearer

aws/rust-runtime/aws-http/Cargo.toml
```diff
@@ -13,7 +13,6 @@ aws-smithy-types = { path = "../../../rust-runtime/aws-smithy-types" }
 aws-types = { path = "../aws-types" }
 http = "0.2.3"
 lazy_static = "1"
-thiserror = "1"
```
rcoh, line 16: 🎉

aws/rust-runtime/aws-http/src/user_agent.rs
```diff
@@ -79,29 +90,58 @@ impl AwsUserAgent {
...
+        self.framework_metadata.push(metadata);
+        self
+    }
+
+    /// Sets the app name for the user agent.
+    pub fn with_app_name(mut self, app_name: AppName) -> Self {
+        self.app_name = Some(app_name);
+        self
+    }
+
```
rcoh, line 129: fine for now, but if we ever actually want to use these, I don't think you can easily get an owned user agent from the property bag so we probably need versions of these that accept `&mut self`

aws/rust-runtime/aws-http/src/user_agent.rs
```diff
@@ -170,38 +213,240 @@ impl ApiMetadata {
...
+        Ok(Self {
+            config: validate_metadata(config.into())?,
+            value,
+        })
+    }
+}
+
+impl fmt::Display for ConfigMetadata {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // config-metadata = "cfg/" config ["/" name]
```
rcoh, line 386:
```suggestion
        // config-metadata = "cfg/" config ["/" value]
```

aws/rust-runtime/aws-types/src/app_name.rs
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+        if !app_name.chars().all(valid_character) {
+            return Err(InvalidAppName);
+        }
+        if app_name.len() > 50 {
+            tracing::warn!(
+                "The `app_name` set when configuring the SDK client is recommended \
+                 to have no more than 50 characters."
+            )
```
rcoh, line 61: I _think_ this is OK, but we should probably ensure this isn't logspam and only gets logged once in practice (not, eg. on every request)

aws/rust-runtime/aws-types/src/app_name.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::AppName;
+
+    #[test]
+    fn validation() {
+        assert!(AppName::new("asdf1234ASDF!#$%&'*+-.^_`|~").is_ok());
+        assert!(AppName::new("foo bar").is_err());
+        assert!(AppName::new("🚀").is_err());
```
rcoh, line 95: sad

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/UserAgentDecorator.kt
```diff
@@ -50,28 +60,102 @@ class UserAgentDecorator : RustCodegenDecorator {
...
+            // Re-export the app name so that it can be specified in config programmatically without an explicit dependency
+            rustTemplate("pub use #{AppName};", "AppName" to runtimeConfig.appName())
+        }
         else -> emptySection
     }
 }
 
-fun RuntimeConfig.awsHttp(): CargoDependency = awsRuntimeDependency("aws-http")
-fun RuntimeConfig.userAgentModule() = awsHttp().asType().copy(name = "user_agent")
+private fun RuntimeConfig.userAgentModule() = awsHttp().asType().copy(name = "user_agent")
```
rcoh, line 81: the `.member("user_agent")` helper maybe clearer here

aws/sdk/integration-tests/s3/tests/user-agent-app-name.rs
```diff
@@ -0,0 +1,42 @@
...
+    let properties = op.properties();
+    let user_agent = properties.get::<AwsUserAgent>().unwrap();
+    let formatted = user_agent.aws_ua_header();
+
+    // verify app name made it to the user agent
+    assert!(
+        formatted.ends_with(" app/test-app-name"),
+        "'{}' didn't end with the app name",
+        formatted
+    );
```
rcoh, line 39: if we're writing an integration test, we might as well use `capture_request` and trace it all the way to the UA header

rcoh (approved):
LGTM! thanks for cleaning up the docs

aws/rust-runtime/aws-config/src/profile/app_name.rs
jdisanti: Definitely. Auto-import kept choosing the re-export from STS, which was pretty frustrating.

aws/rust-runtime/aws-http/src/user_agent.rs
jdisanti: Oh, I was under the impression that anyone using these would be using the Smithy client and insert the property upfront. I suppose they could use a custom middleware to modify it though, so this makes sense.

aws/rust-runtime/aws-config/src/profile/app_name.rs
jdisanti: It looks like this has been implemented for the Javascript V3 SDK: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_util_user_agent_node.html

rcoh (approved):
LGTM!

PR #877, opened 2021-11-19
fix: normalize generated cargo dependency paths

Velfi:
## Motivation and Context

This fixes an issue where we were generating file paths but not normalizing them. This caused issues with IDEs.

## Testing

I visually inspected the new paths generated after the fix

Before the fix:
```
[dependencies.aws-endpoint]
path = "..//aws-endpoint"
version = "0.0.26-alpha"
```
After the fix
```
[dependencies.aws-endpoint]
path = "../aws-endpoint"
version = "0.0.26-alpha"
```

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
@@ -10,7 +10,6 @@ import software.amazon.smithy.codegen.core.SymbolDependencyContainer
 import software.amazon.smithy.rust.codegen.smithy.RuntimeConfig
 import software.amazon.smithy.rust.codegen.smithy.RuntimeType
 import software.amazon.smithy.rust.codegen.util.dq
-import kotlin.io.path.Path
 
 sealed class DependencyScope {
     object Dev : DependencyScope()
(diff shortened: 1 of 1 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
```diff
@@ -10,6 +10,7 @@ import software.amazon.smithy.codegen.core.SymbolDependencyContainer
 import software.amazon.smithy.rust.codegen.smithy.RuntimeConfig
 import software.amazon.smithy.rust.codegen.smithy.RuntimeType
 import software.amazon.smithy.rust.codegen.util.dq
+import kotlin.io.path.Path
```
jdisanti, line 13: Not used? I'm surprised Ktlint didn't catch this...
```suggestion
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
```diff
@@ -150,8 +151,8 @@ data class CargoDependency(
             when (this) {
                 is CratesIo -> attribs["version"] = version
                 is Local -> {
-                    val fullPath = "$basePath/$name"
-                    attribs["path"] = fullPath
+                    val fullPath = java.nio.file.Path.of("$basePath/$name")
```
jdisanti, line 154: Would be nice to import this path rather than fully qualify it.

jdisanti (approved):

PR #881, opened 2021-11-20
Make it possible to turn off the Tokio runtime feature

jdisanti:
## Motivation and Context
This is an incremental step towards supporting multiple async runtimes.

## Description
These changes make it possible to disable the Tokio `runtime` feature by doing the following, for example:
```
[dependencies]
aws-sdk-s3 = { version = "VERSION", default-features = false, features = ["client"] }
```

The approach taken is to change all the runtime crates to no longer default to using Tokio `runtime`, and then have the generated service crates turn on the `rt-tokio` features by default. That way, the features only need to be tweaked on the crates customers will directly depend on.

These changes have not:
- Vetted whether this makes it possible to use async-std
- Done anything with the `aws-config` crate

More work is needed after this PR.

## Testing
- [x] Created a cargo project out of tree that references the generated `aws-sdk-s3`, and used `cargo tree -e features` to determine that nothing was enabling the Tokio `runtime` feature
- [x] Compiled and ran the S3 and DynamoDB examples
- [x] Compiled and ran the S3 integration tests

## Checklist

- [ ] I have updated `CHANGELOG.md` if I made changes to the smithy-rs codegen or runtime crates
- [ ] I have updated `aws/SDK_CHANGELOG.md` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

### DIFF
--- a/aws/rust-runtime/aws-hyper/Cargo.toml
+++ b/aws/rust-runtime/aws-hyper/Cargo.toml
@@ -9,6 +9,7 @@ repository = "https://github.com/awslabs/smithy-rs"
 
 [features]
 default = []
+rt-tokio = ["hyper/runtime", "hyper/tcp"]
 native-tls = ["hyper-tls", "aws-smithy-client/native-tls"]
 rustls = ["hyper-rustls", "aws-smithy-client/rustls"]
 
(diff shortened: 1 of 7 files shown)


jdisanti:
CI has revealed this opens a can of worms with our existing feature matrix. We're going to need to a decent amount of work to get all the runtime crates building with their various combinations of features. To give one example, DVR requires both `rustls` and `test-util`, but currently only states `test-util`. I'm also observing other weird issues such as the timeout tests never timing out when `aws-smithy-client` only has the `test-util` feature enabled in addition to the defaults.

Converting this to a draft for now.

Velfi:
> I'm also observing other weird issues such as the timeout tests never timing out when `aws-smithy-client` only has the `test-util` feature enabled in addition to the defaults.

I saw that pop up intermittently throughout development of the Timeouts feature and it made me feel like I was going nuts. It didn't come up in the final checks before merge so I thought it was fixed, sorry about that.

PR #883, opened 2021-11-22
set appropriate minimum httpbody version

rcoh:
## Motivation and Context
https://github.com/awslabs/aws-sdk-rust/issues/305

## Description
http_body::combinators was added in 0.4.1 but we specify 0.4.0 in the dependencies.

## Testing
- CI

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-hyper/Cargo.toml
+++ b/aws/rust-runtime/aws-hyper/Cargo.toml
@@ -23,7 +23,7 @@ aws-smithy-types = { path = "../../../rust-runtime/aws-smithy-types" }
 bytes = "1"
 fastrand = "1.4.0"
 http = "0.2.3"
-http-body = "0.4.0"
+http-body = "0.4.4"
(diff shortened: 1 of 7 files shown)


aws/rust-runtime/aws-hyper/Cargo.toml
```diff
@@ -23,7 +23,7 @@ aws-smithy-types = { path = "../../../rust-runtime/aws-smithy-types" }
 bytes = "1"
 fastrand = "1.4.0"
 http = "0.2.3"
-http-body = "0.4.0"
+http-body = "0.4.4"
```
Velfi, line 26: do we have to be specific here or could we do `"0.4"`? Did we ever write down guidelines for how we specify dependency versions?

Velfi (approved):

aws/rust-runtime/aws-hyper/Cargo.toml
rcoh: we can't do 0.4—we are not compatible with `0.4.0` (that's the entire genesis of this PR)

PR #886, opened 2021-11-22
Feature: add TimeoutError variant to SdkError

Velfi:
## Motivation and Context

Before this change, service timeouts returned an `SdkError::ConstructionFailure` which is not accurate. This adds a new variant `SdkError::TimeoutError` for them to return instead.

## Description

Add a new `TimeoutError` variant to `SdkError`

## Testing

I updated and ran the extant timeout tests

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/imds/client.rs
+++ b/aws/rust-runtime/aws-config/src/imds/client.rs
@@ -187,6 +187,7 @@ impl Client {
                 Ok(token_failure) => *token_failure,
                 Err(other) => ImdsError::Unexpected(other),
             },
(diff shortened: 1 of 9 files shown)


aws/rust-runtime/aws-config/src/imds/client.rs
```diff
@@ -251,6 +252,9 @@ pub enum ImdsError {
     /// An error occurred communication with IMDS
     IoError(Box<dyn Error + Send + Sync + 'static>),
```
rcoh, line 253: I think in the case of IMDS, I'd probably just put this into IOError instead of adding another variant

rcoh (approved):
LGTM! let me know what you think about IMDS

PR #887, opened 2021-11-23
Update examples folder link

allan2:
The previous link works because GitHub is automatically redirecting it.

This change is to update it to the new link.


Velfi (approved):
This one looks good! Thanks for fixing this

PR #893, opened 2021-11-23
fix: silence home directory warning when running on lambda

Velfi:
update: SDK_CHANGELOG.md

## Motivation and Context

[aws](https://github.com/awslabs/aws-sdk-rust/issues/307)

## Description

This adds a new function that determines whether or not an app is running in an AWS Lambda. If it is, it will silence a warning that gets emitted when no home directory can be found

## Testing

I wrote a test ensuring logs won't be emitted when a lambda-specific env var is set

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/profile/parser/source.rs
+++ b/aws/rust-runtime/aws-config/src/profile/parser/source.rs
@@ -145,9 +145,15 @@ fn expand_home(path: impl AsRef<Path>, home_dir: &Option<String>) -> PathBuf {
                     dir.clone()
                 }
                 None => {
-                    tracing::warn!(
-                        "could not determine home directory but home expansion was requested"
-                    );
+                    // Lambdas don't have home directories and emitting this warning is not helpful
+                    // to users running the SDK from within a Lambda. This warning will be silenced
+                    // if we determine that that is the case.
+                    let is_likely_running_on_a_lambda = check_is_likely_running_on_a_lambda();
+                    if !is_likely_running_on_a_lambda {
+                        tracing::warn!(
+                            "could not determine home directory but home expansion was requested"
+                        );
+                    }
                     // if we can't determine the home directory, just leave it as `~`
                     "~".into()
                 }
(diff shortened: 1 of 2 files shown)


Velfi:
@rcoh It occurred to me that it might be better to accomplish this by just skipping resolution of profile files on Lambda. Is that going too far?

aws/rust-runtime/aws-config/src/profile/parser/source.rs
```diff
@@ -167,6 +178,18 @@ fn expand_home(path: impl AsRef<Path>, home_dir: &Option<String>) -> PathBuf {
     }
 }
 
+// Returns true or false based on whether or not this code is likely running inside an AWS Lambda.
+// [Lambdas set many environment variables](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime)
+// that we can check.
+fn check_is_likely_running_on_a_lambda(environment: &os_shim_internal::Env) -> bool {
+    // AWS_EXECUTION_ENV – The runtime identifier, prefixed by AWS_Lambda_—for example, AWS_Lambda_java8.
+    environment
+        .get("AWS_EXECUTION_ENV")
```
nmoutschen, line 187: This is a different variable than what other libraries use to identify if we are running in a Lambda function.

the AWS X-Ray SDK uses `LAMBDA_TASK_ROOT` for this purpose:

* [Python SDK for X-Ray](https://github.com/aws/aws-xray-sdk-python/blob/d025f5641496efdc9dfad2b270139c224053885c/aws_xray_sdk/core/lambda_launcher.py#L14)
* [Go SDK for X-Ray](https://github.com/aws/aws-xray-sdk-go/blob/c4ccab1d3ca59f5ce66f3dab6522dc1fc4a490c7/xray/lambda.go#L25)
* [Node SDK for X-Ray](https://github.com/aws/aws-xray-sdk-node/blob/a9d0cf9cbd0328e40f30554f61b4bd5fac08bafc/packages/core/lib/aws-xray.js#L380)

The benefit of using `LAMBDA_TASK_ROOT` over `AWS_EXECUTION_ENV` is that the latter is also used for other compute environments like Fargate, so we need to check that the value starts with `AWS_Lambda_`.

Furthermore, after doing a test with `provided.al2`, I don't have an `AWS_EXECUTION_ENV` environment variable set. So it's possible that that variable only exists for managed runtimes, and thus wouldn't work on `provided` and `provided.al2`.

Velfi: Thanks for checking that @nmoutschen, I'll update it to check that that var is set

rcoh (approved):
LGTM!

rcoh:
I think we should _probably_ continue trying to resolve profile files on lambda—I assume that in some cases users will provide them

nmoutschen:
Thanks a lot for fixing this quickly!

> I think we should _probably_ continue trying to resolve profile files on lambda—I assume that in some cases users will provide them

On that note, I don't think I've ever seen anyone use profiles on Lambda. Since you have an IAM execution role attached to the function, you usually rely on that. I've seen cases of using that role to assume another role, but that's it.

PR #894, opened 2021-11-24
Support for share state in Server codegen

crisidev:
## Motivation and Context
Customer of the SSDK may want to use some shared state in their handlers implementation. This PR implements the necessary codegen and runtime and removes the dependency on `derive_builder` by fully codegenerating the OperationRegistry.

### Example usage
```rust
use std::sync::{Arc, Mutex};

use aws_smithy_http_server::routing::Router;
use aws_smithy_http_server::AddExtensionLayer;
use aws_smithy_http_server::Extension;
use simple::error::*;
use simple::input::*;
use simple::operation_registry::SimpleServiceOperationRegistryBuilder;
use simple::output::*;

struct State {
    counter: Mutex<i32>,
}

async fn _healthcheck_operation(_input: HealthcheckInput) -> HealthcheckOutput {
    HealthcheckOutput::builder().build()
}

async fn healthcheck_operation_with_state(
    _input: HealthcheckInput,
    state: Extension<Arc<State>>,
) -> HealthcheckOutput {
    let mut counter = state.0.counter.lock().unwrap();
    *counter += 1;
    dbg!(&counter);

    HealthcheckOutput::builder().build()
}

async fn register_service_operation(
    _input: RegisterServiceInput,
) -> Result<RegisterServiceOutput, RegisterServiceError> {
    Ok(RegisterServiceOutput::builder().build())
}

async fn _register_service_operation_with_state(
    _input: RegisterServiceInput,
    _state: Extension<State>,
) -> Result<RegisterServiceOutput, RegisterServiceError> {
    Ok(RegisterServiceOutput::builder().build())
}

#[tokio::main]
async fn main() {
    let app: Router = SimpleServiceOperationRegistryBuilder::default()
        .healthcheck(healthcheck_operation_with_state)
        .register_service(register_service_operation)
        // Both implementation can be used, depending if the state is required or not.
        // .register_service(_register_service_operation_with_state)
        .build()
        .unwrap()
        .into();

    let shared_state = Arc::new(State {
        counter: Mutex::new(69),
    });

    let app = app.layer(AddExtensionLayer::new(shared_state));

    let server =
        axum::Server::bind(&"0.0.0.0:8080".parse().unwrap()).serve(app.into_make_service());

    if let Err(err) = server.await {
        eprintln!("server error: {}", err);
    }
}

```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #895, opened 2021-11-24
Example application showcasing how stateful handlers could work

david-perez:
This commit adds a minimal example application that showcases how we
could add support for stateful handlers to the framework, if we move the
`Handler` trait from the runtime to code generation.

Service implementers can provide, for each operation, a handler function
that requires shared state _or not_. This is indicated in the signature
of the function:

* a function requiring state takes in two arguments: the first one is
  the modeled input and the second one is an `Extension<S>`, where `S`
  can be any type that is `Clone + Send + Sync`. The order of the
  arguments is fixed.
* a function that does not require state takes in one argument, the
  modeled input.

Service implementers must register the shared state of their application
by calling `.layer(AddExtensionLayer::new(shared_state))` on their
router. The application will panic at runtime if a handler that expects
state is called and the implementer has not registered the state at
router initialization.

---

I have also `cargo expand`ed and tweaked the operation registry's
`derive_builder`, since `derive_builder` does not support
`std::marker::PhantomData`, so that the application can be run:

```sh
➜ cargo run
[...]
```

On another pane:

```
➜ http localhost:8080/path/to/label/healthcheck
HTTP/1.1 200 OK
content-length: 0
date: Wed, 24 Nov 2021 13:03:24 GMT
➜ http localhost:8080/path/to/label/healthcheck
HTTP/1.1 200 OK
content-length: 0
date: Wed, 24 Nov 2021 13:03:24 GMT
➜ http POST "localhost:8080/register-service/gre/ee/dy/suffix?key&foo=bar"
HTTP/1.1 200 OK
content-length: 0
date: Wed, 24 Nov 2021 13:04:43 GMT
```

Observed output on first pane:

```sh
[aws-smithy-http-server/src/main.rs:20] &counter = 70
[aws-smithy-http-server/src/main.rs:20] &counter = 71
```


rust-runtime/aws-smithy-http-server/src/add_extension.rs
```diff
@@ -0,0 +1,67 @@
+// this is vendored from tower-http to reduce public dependencies
```
david-perez, line 1: Copied from `axum`, but let's revisit this choice: I think it's fair to re-export `tower-http`.

rust-runtime/aws-smithy-http-server/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+    _input: RegisterServiceInput,
+    _state: Extension<State>,
+) -> Result<RegisterServiceOutput, RegisterServiceError> {
+    Ok(RegisterServiceOutput)
+}
+
+#[tokio::main]
+async fn main() {
+    let app: Router = SimpleServiceOperationRegistryBuilder::default()
+        // .health_check(_healthcheck_operation)
```
david-perez, line 41: You can swap these lines out and the application runs fine, independently of whether you provide a handler requiring state or not.

crisidev: Nice work!

rust-runtime/aws-smithy-http-server/src/add_extension.rs
crisidev: Since we are not planning modifications I think its good to just re-export `tower-http`.

crisidev: I am going to change https://github.com/awslabs/smithy-rs/pull/894 and re-export.

crisidev:
Closing in favour of #894

PR #906, opened 2021-11-30
Don't capture empty session tokens from env

jdisanti:
## Motivation and Context
This is the fix for https://github.com/awslabs/aws-sdk-rust/issues/316

The AWS CLI won't send session tokens if `AWS_SESSION_TOKEN` is empty or has nothing but whitespace.

## Testing
- Added a unit test
- Manually tested with S3 example

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/environment/credentials.rs
+++ b/aws/rust-runtime/aws-config/src/environment/credentials.rs
@@ -29,7 +29,12 @@ impl EnvironmentVariableCredentialsProvider {
             .get("AWS_SECRET_ACCESS_KEY")
             .or_else(|_| self.env.get("SECRET_ACCESS_KEY"))
             .map_err(to_cred_error)?;
-        let session_token = self.env.get("AWS_SESSION_TOKEN").ok();
+        let session_token = self
+            .env
+            .get("AWS_SESSION_TOKEN")
+            .ok()
+            .map(|token| token.trim().to_string())
+            .filter(|token| !token.is_empty());
         Ok(Credentials::new(
             access_key,
             secret_key,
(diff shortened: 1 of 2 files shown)


aws/rust-runtime/aws-config/src/environment/credentials.rs
```diff
@@ -29,7 +29,12 @@ impl EnvironmentVariableCredentialsProvider {
             .get("AWS_SECRET_ACCESS_KEY")
             .or_else(|_| self.env.get("SECRET_ACCESS_KEY"))
             .map_err(to_cred_error)?;
-        let session_token = self.env.get("AWS_SESSION_TOKEN").ok();
+        let session_token = self
+            .env
+            .get("AWS_SESSION_TOKEN")
+            .ok()
+            .map(|token| token.trim().to_string())
+            .filter(|token| !token.is_empty());
```
Velfi, line 37: Is it worth it to refactor this so we allocate after the filter? We could even use `filter_map`
```
.filter_map(|token| {
  let t = token.trim();
  
  if token.is_empty() {
    None
  } else {
    Some(token.to_string())
  }
})
```

Velfi (approved):

rcoh (approved):
LGTM

aws/rust-runtime/aws-config/src/environment/credentials.rs
jdisanti: It doesn't look like [Option](https://doc.rust-lang.org/stable/std/option/enum.Option.html) has `filter_map`. Could potentially map it to another option and then `flatten` though.

Velfi: I always forget I can't user iterator methods on options and results

PR #907, opened 2021-11-30
Improve log messaging around missing sleep implementation

jdisanti:
## Motivation and Context
This is the fix for https://github.com/awslabs/aws-sdk-rust/issues/317

When constructing a Smithy client manually, or when constructing a fluent client with a service config, the default sleep implementation wasn't being used, which was unintentionally disabling timeouts.

## Testing
- Added unit tests to ensure the builders fall back to the default sleep implementation
- Manually tested it in an local project where I had reproduced the issue

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -13,7 +13,7 @@
 
 [[aws-sdk-rust]]
 message = """
-Removed spamming log message when a client was used without a sleep implementation, and
+Removed inaccurate log message when a client was used without a sleep implementation, and
 improved context and call to action in logged messages around missing sleep implementations.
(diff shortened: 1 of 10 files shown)


jdisanti:
From offline discussion of our approach on this, will do the following:
- Revert back to not defaulting the sleep impl for construction of service clients
- Don't log the warning about missing async sleep if no timeout config is provided
- Update warning to also say retries won't work if there's a missing async sleep
- Add a warning when a client is constructed with _no_ sleep impl that suggests using `aws-config` and shared config

rcoh (approved):
LGTM!

PR #918, opened 2021-12-02
Improve the codegen diff bot

jdisanti:
## Motivation and Context
This updates the codegen diff bot GitHub Action to generate an HTML diff and upload it to a CloudFront fronted S3 bucket so that the diffs can live longer and so that we don't need to push branches to create the diffs anymore.


jdisanti:
The last diff posted above represents what the diffing looks like with this PR. I've pushed a commit that reverts the demo changes, so further diffs won't show anything.

rcoh (approved):
LGTM, just a couple of questions about bucket naming

jdisanti:
I'm going to tear down the old stack in order to rename things. The old diff comments will become broken links. Will also play around with splitting out the SDK vs. the server codegen diffs into separate files, adding whitespace/no-whitespace diffing, and modifying the bot comment to say if a diff has any changes in it.

tools/ci-cdk/README.md
```diff
@@ -0,0 +1,17 @@
+# CI CDK
+
+This is the CDK infrastructure as code for awslabs/smithy-rs and awslabs/aws-sdk-rust
+continuous integration.
+
+The `cdk.json` file tells the CDK Toolkit how to synthesize the infrastructure.
```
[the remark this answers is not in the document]
jdisanti: I think this will get out of date over time. Inclined to leave it as is.

tools/codegen-diff-revisions.py
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env python3
```
rcoh, line 1: rip bash 🥳

tools/codegen-diff-revisions.py
```diff
@@ -0,0 +1,232 @@
...
+# Returns the output from a shell command. Bails if the command failed
+def get_cmd_output(command):
+    result = subprocess.run(command, capture_output=True, shell=True)
+    expect_cmd_success(command, result)
+    return result.stdout.decode("utf-8").strip()
+
+
+# Runs a shell command and returns its exit status
+def get_cmd_status(command):
+    return subprocess.run(command, capture_output=True, shell=True).returncode
```
rcoh, line 228: `shell=True` has some security implications: https://docs.python.org/3/library/subprocess.html#security-considerations

Although the usages currently in this script are fine since it's always a static string, I might use `shlex.split` for all of the examples that _don't_ need pipes and only set `shell=True` for the one command that needs pipes.

tools/codegen-diff-revisions.py
```diff
@@ -0,0 +1,232 @@
...
+    if result.returncode != 0:
+        eprint(f"Command `{command}` failed with status code {result.returncode}")
+        eprint("stderr:\n" + result.stderr.decode("utf-8"))
+        eprint("stdout:\n" + result.stdout.decode("utf-8"))
+        sys.exit(1)
+
+
+# Runs a shell command
+def run(command):
+    result = subprocess.run(command, stdout=sys.stderr, stderr=sys.stderr, shell=True)
```
rcoh, line 215: you can use `subprocees.run(check=True)`

rcoh (approved):

tools/codegen-diff-revisions.py
jdisanti: Bash was doomed the moment I needed a function with arguments...

tools/codegen-diff-revisions.py
```diff
@@ -202,25 +203,16 @@ def eprint(*args, **kwargs):
...
+    if not shell:
+        command = shlex.split(command)
+    subprocess.run(command, stdout=sys.stderr, stderr=sys.stderr, shell=shell, check=True)
 
 
 # Returns the output from a shell command. Bails if the command failed
 def get_cmd_output(command):
-    result = subprocess.run(command, capture_output=True, shell=True)
-    expect_cmd_success(command, result)
+    result = subprocess.run(command, capture_output=True, shell=True, check=True)
```
rcoh, line 215: this is only used once AFAICT and doesn't need `shell=True`

jdisanti: Ah, good catch. Will address as part of the doc preview PR.

PR #924, opened 2021-12-03
Implement a doc preview bot

jdisanti:
## Motivation and Context
This adds a GitHub Action that generates a documentation preview for the runtime crates, STS, and Transcribe Streaming for a pull request.

### DIFF
--- a/.github/workflows/codegen-preview.yml
+++ b/.github/workflows/codegen-preview.yml
@@ -123,14 +123,14 @@ jobs:
 
         cargo doc --no-deps --all-features
         popd
-        ./tools/generate-doc-preview-index.sh
+        ./tools/generate-doc-preview-index.sh ${{ github.event.pull_request.base.sha }}
     - uses: aws-actions/configure-aws-credentials@v1
       name: Acquire credentials for uploading to S3
       with:
         role-to-assume: ${{ secrets.SMITHY_RS_PULL_REQUEST_CDN_ROLE_ARN }}
         role-session-name: GitHubActions
         aws-region: us-west-2
-    - name: Upload diff to S3
+    - name: Upload doc preview to S3
       run: |
         aws s3 cp target/doc "s3://${S3_BUCKET_NAME}/docs/${{ github.event.pull_request.head.sha }}" --recursive
     - uses: actions/github-script@v5
--- a/.github/workflows/codegen-preview.yml
+++ b/.github/workflows/codegen-preview.yml
@@ -8,6 +8,8 @@ on:
     - synchronize
 env:
   java_version: 11
+  rust_version: 1.54.0
+
 jobs:
   generate-diff:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


tools/generate-doc-preview-index.sh
```diff
@@ -0,0 +1,43 @@
...
+
+HEAD_COMMIT_SHA=$(git rev-parse HEAD)
+cd $(git rev-parse --show-toplevel)/target/doc
+
+{
+    echo '<!doctype html>'
+    echo '<html>'
+    echo '<head>'
+    echo '  <metadata charset="utf-8">'
+    echo "  <title>Doc preview for ${HEAD_COMMIT_SHA}</title>"
```
david-perez, line 15: A more human-readable and useful title would be something like `git log -1 --oneline`.

tools/generate-doc-preview-index.sh
```diff
@@ -0,0 +1,43 @@
...
+    echo '</head>'
+    echo '<body>'
+    echo "  <h1>Doc preview for ${HEAD_COMMIT_SHA}</h1>"
+} > index.html
+
+doc_link () {
+    local package=$1
```
david-perez, line 22: `crate` instead of `package` throughout this file would be a better name.

tools/generate-doc-preview-index.sh
```diff
@@ -0,0 +1,43 @@
...
+    if [[ -d ${package} ]]; then
+        echo "    <li><a href='${package}/index.html'>${package}</a></li>" >> index.html
```
david-perez, line 24:
```suggestion
        echo "  <li><a href='${package}/index.html'>${package}</a></li>" >> index.html
```

The rest of the indents are two spaces.

tools/generate-doc-preview-index.sh
```diff
@@ -0,0 +1,43 @@
...
+    fi
+}
+
+echo '  <h3>AWS Services</h3><ul>' >> index.html
+for package in $(ls | grep aws_sdk_ | sort); do doc_link ${package}; done
```
david-perez, line 29:
```suggestion
for package in $(ls -d aws_sdk_*); do doc_link ${package}; done
```

tools/generate-doc-preview-index.sh
```diff
@@ -0,0 +1,43 @@
...
+echo '  </ul>' >> index.html
+
+echo '  <h3>AWS Runtime Crates</h3><ul>' >> index.html
+for package in $(ls | grep aws_ | grep -v _smithy_ | sort); do doc_link ${package}; done
```
david-perez, line 33: This also includes the crates that started with `aws_sdk_`, which were already included in the previous group. You can use OR patterns instead of two invocations to `grep`. And the call to `sort` is unnecessary since `ls` output is sorted alphabetically and `grep` does not reorder matches.

So it can be simplified to:

```suggestion
for package in $(ls -d aws_* | grep -v '_smithy_\|_sdk_'); do doc_link ${package}; done
```

tools/generate-doc-preview-index.sh
```diff
@@ -0,0 +1,43 @@
...
+echo '  </ul>' >> index.html
+
+echo '  <h3>Smithy Runtime Crates</h3><ul>' >> index.html
+for package in $(ls | grep aws_smithy_ | sort); do doc_link ${package}; done
```
david-perez, line 37:
```suggestion
for package in $(ls -d aws_smithy_*); do doc_link ${package}; done
```

david-perez (commented):
This is a great idea!

Can we also please include the server runtime crate (`aws_smithy_http_server`)? It's the only Smithy runtime crate missing because it's not used in the AWS SDK. So I suggest that we generate docs for _all_ crates under `./rust_runtime`.

david-perez:
Another possible improvement would be to only generate docs for crates that changed in the PR compared to the base branch. E.g.

```
git diff --name-only origin/main | grep rust-runtime | cut -d'/' -f2 | sort | uniq
```

Would be more involved and require creating an ad-hoc Cargo workspace (or `cd`ing into each crate and running `cargo docs` separately).

So a perhaps better idea is to do as we're currently doing, but **bolding** the changed crate names.

.github/workflows/codegen-preview.yml
```diff
@@ -64,11 +66,72 @@ jobs:
...
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ env.rust_version }}
+        default: true
+    - name: Generate doc preview
+      # Only generate two of the smallest services since these get huge. One of these must be
+      # STS since aws-config depends on it. STS and Transcribe Streaming were chosen below to stay
+      # small while still representing most features. Combined, they are about 11 MB at time of writing.
+      run: |
+        ./gradlew -Paws.services=+sts,+transcribestreaming :aws:sdk:assemble
```
rcoh, line 111: probably doesn't matter that much but we could also use the generated tier 1 artifact and just prune services out

.github/workflows/codegen-preview.yml
```diff
@@ -64,11 +66,72 @@ jobs:
...
+        pushd aws/sdk/build/aws-sdk
+        # Remove example crates from workspace
+        sed -i '/examples/d' Cargo.toml
+        cargo doc --no-deps --all-features
```
rcoh, line 115: is there a reason you did use `RUSTDOCFLAGS='--enable-index-page -Zunstable-options'` instead of manually making an index? In either case, we should use nightly (because that's what rustdoc uses) and include `--cfg docrs` in the `RUSTDOCFLAGS` environment variable to mimic how the docs will look

.github/workflows/codegen-preview.yml
jdisanti: That won't work if we want to include the server runtime crates, but regardless, I had avoided doing it this way since the preview bot is in its own workflow.

.github/workflows/codegen-preview.yml
jdisanti: I wasn't aware of `--enable-index-page`, although I like David's suggestion of differentiating the crate names that have actual changes in them, which is enabled by generating it ourselves.

rcoh (approved):
replies make sense, 🚢

tools/generate-doc-preview-index.sh
jdisanti: I like it! Thanks for the suggestion.

tools/generate-doc-preview-index.sh
jdisanti: Actually, this one was intentionally at four spaces since it's nested inside a `<ul>`.

tools/generate-doc-preview-index.sh
jdisanti: Good catch!

Issue #930, opened 2021-12-06
Fix aws-config features


jdisanti:
The current features in `aws-config` aren't the right set for the customer and they require a lot of manipulation. This is multi part story:
- [ ] RFC for new feature layout in aws-config
- [ ] redo the features in aws-config

----
previously:
- [x] Remove default features from all smithy runtime crates
- [x] Remove default features from all AWS runtime crates (except aws-config)
- [x] Update generated SDK Cargo.tomls to pull in the correct defaults

PR #935, opened 2021-12-06
remove: default features from runtime crates

Velfi:
## Motivation and Context

This PR addresses #930 

## Description

This PR modifies runtime crates to have no default features. The desired features must now be explicitly defined and many tests and examples have been updated to specify the features they require.

## Testing

I ran the CI tests/lints

## Checklist


CHANGELOG.next.toml
```diff
@@ -10,6 +10,18 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false }
 # author = "rcoh"
+[[aws-sdk-rust]]
+message = "Runtime crates (other than aws-config) no longer have default features. You must now specify the features that you want when you add a dependency to your `Cargo.toml`."
```
jdisanti, line 14: I think this is a breaking change from the perspective of smithy-rs, but not necessarily from the perspective of aws-sdk-rust (depending on which crates customers were depending on). For customers using just `aws-config` and some service packages, they shouldn't notice a change at all. I think we should nuance the language a little.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -82,9 +82,10 @@ class AwsFluentClientDecorator : RustCodegenDecorator {
             AwsFluentClientExtensions(types).render(writer)
         }
         val awsHyper = "aws-hyper"
-        rustCrate.mergeFeature(Feature("client", default = true, listOf(awsHyper, "aws-smithy-client")))
-        rustCrate.mergeFeature(Feature("rustls", default = true, listOf("$awsHyper/rustls")))
-        rustCrate.mergeFeature(Feature("native-tls", default = false, listOf("$awsHyper/native-tls")))
+        val awsSmithyClient = "aws-smithy-client"
+        rustCrate.mergeFeature(Feature("client", default = true, listOf(awsSmithyClient, awsHyper)))
+        rustCrate.mergeFeature(Feature("rustls", default = false, listOf("$awsHyper/rustls", "$awsSmithyClient/rustls")))
```
jdisanti, line 87: I think we probably still want to default to rustls on the service crates so that customers don't need to specify a feature list to use the SDK. They will be able to switch to native-tls by specifying `no-default-features`, and then adding `native-tls` and `client` explicitly.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -82,9 +82,10 @@ class AwsFluentClientDecorator : RustCodegenDecorator {
             AwsFluentClientExtensions(types).render(writer)
         }
         val awsHyper = "aws-hyper"
-        rustCrate.mergeFeature(Feature("client", default = true, listOf(awsHyper, "aws-smithy-client")))
-        rustCrate.mergeFeature(Feature("rustls", default = true, listOf("$awsHyper/rustls")))
-        rustCrate.mergeFeature(Feature("native-tls", default = false, listOf("$awsHyper/native-tls")))
+        val awsSmithyClient = "aws-smithy-client"
+        rustCrate.mergeFeature(Feature("client", default = true, listOf(awsSmithyClient, awsHyper)))
```
jdisanti, line 86: Was the use of crate names instead of feature names intentional in the feature list for `client`?

@rcoh - What is the use-case for the `client` feature? Is this something we could potentially remove?

jdisanti (commented):
This is a great start!

CHANGELOG.next.toml
Velfi: If someone was directly using something like `aws-sigv4` then they'd be affected by this change. I don't think that that'll be an issue in practice but it is a breaking change. That being said, I'm for setting whatever expectations you think are the most helpful. I'll bother you about it tomorrow.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
Velfi: I figured we could remove it but I wanted to ask first since I do think it could affect the Server folks

CHANGELOG.next.toml
jdisanti: Yeah, I would probably mark it as a breaking change, but preface the message with something like, "If you directly depend on AWS or Smithy runtime crates (e.g., AWS crates not named `aws-config` or prefixed with `aws-sdk-`), ...".

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
rcoh: the use case is that you could get a SDK that didn't actually depend on anything (just included the operations)

CHANGELOG.next.toml
```diff
@@ -10,6 +10,18 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false }
 # author = "rcoh"
+[[aws-sdk-rust]]
+message = "If you directly depend on AWS or Smithy runtime crates _(e.g., AWS crates not named aws-config or prefixed with aws-sdk-)_, the default features from those crates must now be explicitly set in your `Cargo.toml`."
```
jdisanti, line 14:
```suggestion
message = "If you directly depend on AWS or Smithy runtime crates _(e.g., AWS crates not named `aws-config` or prefixed with `aws-sdk-`)_, the previously default features from those crates must now be explicitly set in your `Cargo.toml`."
```

jdisanti: Thinking we should probably list out the previous defaults for each runtime crate as a quick upgrade reference here.

CHANGELOG.next.toml
```diff
@@ -10,6 +10,18 @@
...
+references = ["smithy-rs#930"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
+author = "zhessler"
+
+[[smithy-rs]]
+message = "Runtime crates no longer have default features. You must now specify the features that you want when you add a dependency to your `Cargo.toml`."
```
jdisanti, line 20: This should get the same upgrade list

aws/rust-runtime/aws-http/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "Apache-2.0"
 repository = "https://github.com/awslabs/smithy-rs"
 
 [dependencies]
-aws-smithy-http = { path = "../../../rust-runtime/aws-smithy-http" }
+aws-smithy-http = { path = "../../../rust-runtime/aws-smithy-http", features = ["bytestream-util"] }
```
jdisanti, line 11: Another question for @rcoh: If most of our AWS runtime crates pull in the `bytstream-util` feature, is there any point to it being a feature? Was the intent that Smithy generated clients/servers can opt out of it?

rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -4,22 +4,51 @@
...
+use crate::timeout::generate_timeout_service_params_from_timeout_config;
+
 pub mod bounds;
 pub mod erase;
 pub mod retry;
 
 // https://github.com/rust-lang/rust/issues/72081
 #[allow(rustdoc::private_doc_tests)]
 mod builder;
-pub use builder::Builder;
```
jdisanti, line 21: Nit: I prefer these where they were before since it's clearer that the module is split out and then re-exported.

rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -4,22 +4,51 @@
  */
 
 //! A Hyper-based Smithy service client.
+//!
+//! | Feature         | Description |
+//! |-----------------|-------------|
+//! | bytestream-util | ByteStream provides misuse-resistant primitives to make it easier to handle common patterns with streaming data. |
```
jdisanti, line 10: From looking at the doc preview, I think these will look better with the feature name wrapped in backticks.

rust-runtime/aws-smithy-client/src/lib.rs
Velfi: oops, that was my bad. I accidentally ran the default formatter instead of rustfmt. I'll revert that

aws/rust-runtime/aws-http/Cargo.toml
Velfi: I assumed that not all SDKs had streaming data and maybe that's why it exists?

crisidev (approved):
LGTM from the server runtime perspective.

aws/rust-runtime/aws-http/Cargo.toml
rcoh: yeah, it's a feature because it uses `tokio/fs` and `tokio/io`

I'm not sure why anything other than end users would need it though

CHANGELOG.next.toml
```diff
@@ -29,6 +29,42 @@ references = ["aws-sdk-rust#317", "smithy-rs#907"]
...
+
+[[smithy-rs]]
+message = '''
+Runtime crates no longer have default features. You must now specify the features that you want when you add a dependency to your `Cargo.toml`.
+
+**Upgrade guide**
+
+| before                          | after |
+|---------------------------------|-------|
+| `aws-sigv4 = "VERSION"`         | `aws-sigv4 = { version = "VERSION", features = ["sign-http"] }` |
```
jdisanti, line 59: `aws-sigv4` is only for the SDK, so it can be removed from the smithy-rs section.
```suggestion
```

aws/rust-runtime/aws-sigv4/Cargo.toml
```diff
@@ -11,7 +11,6 @@ repository = "https://github.com/awslabs/smithy-rs"
 [features]
 sign-http = ["http", "percent-encoding", "form_urlencoded"]
 sign-eventstream = ["aws-smithy-eventstream", "bytes"]
-default = ["sign-http"]
```
jdisanti, line 14: I think `aws-sigv4` is another of the odd exceptions since it already had users before the alpha SDK was launched. I don't think there's any harm in keeping `sign-http` default here. This crate is intended to be used standalone from the SDK unlike the other runtime crates.

aws/rust-runtime/aws-types/Cargo.toml
```diff
@@ -8,11 +8,10 @@ license = "Apache-2.0"
 repository = "https://github.com/awslabs/smithy-rs"
 
 [features]
-default = []
 hardcoded-credentials = []
 
 [dependencies]
-aws-smithy-async = { path = "../../../rust-runtime/aws-smithy-async" }
+aws-smithy-async = { path = "../../../rust-runtime/aws-smithy-async", features = ["rt-tokio"] }
```
jdisanti, line 14: Why is `rt-tokio` required for `aws-types`?

rust-runtime/aws-smithy-http/README.md
```diff
@@ -5,7 +5,7 @@ Core HTTP primitives for service clients generated by [smithy-rs](https://github
 - Endpoint support
 - HTTP header deserialization
 - Event streams
-- `ByteStream` misuse-resistant abstraction for streaming binary data
+- `ByteStream`: _(supported on crate feature `tokio-rt` only)_ a misuse-resistant abstraction for streaming binary data
```
jdisanti, line 8:
```suggestion
- `ByteStream`: _(supported on crate feature `rt-tokio` only)_ a misuse-resistant abstraction for streaming binary data
```

jdisanti (commented):
This looks really close! Just some minor comments/question.

aws/rust-runtime/aws-types/Cargo.toml
Velfi: it probably isn't, I'll remove it

Velfi: ah, it does depend on the async sleep impl in `aws-smithy-async` so there will be some dep eventually but I think it's fine to remove this until we actually add other sleep impls.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/EventStreamSymbolProvider.kt
```diff
@@ -55,7 +55,7 @@ class EventStreamSymbolProvider(
                     .rustType(rustType)
                     .addReference(error)
                     .addReference(initial)
-                    .addDependency(CargoDependency.SmithyHttp(runtimeConfig).withFeature("event-stream"))
```
rcoh, line 58: do we actually need a full tokio runtime for this? There is a difference between tokio with no default features (which is runtime agnostic) vs the full tokio runtime

rust-runtime/aws-smithy-http/README.md
```diff
@@ -5,7 +5,7 @@ Core HTTP primitives for service clients generated by [smithy-rs](https://github
 - Endpoint support
 - HTTP header deserialization
 - Event streams
-- `ByteStream` misuse-resistant abstraction for streaming binary data
```
rcoh, line 8: this isn't true—the only part that isn't supported is the `from_path` features

rust-runtime/aws-smithy-http/src/lib.rs
```diff
@@ -21,5 +32,8 @@ pub mod retry;
 #[cfg(feature = "event-stream")]
 pub mod event_stream;
 
+#[cfg(feature = "rt-tokio")]
```
rcoh, line 35: I don't think this is right? All of ByteStream works fine without rt-tokio (except the from_path) parts

Velfi: You're right, I'll remove this

rcoh (approved):

PR #936, opened 2021-12-07
Add CI job to check for unused dependencies

jdisanti:
## Motivation and Context
This adds a CI step to check for unused dependencies so that we remove them. Currently, it is checking the SDK codegen output rather than the runtime crates, and this is so that the `aws-config` crate gets included. I'm open to suggestions on how to also include the server codegen, but it may need to be its own CI job.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -318,6 +318,44 @@ jobs:
         # so we have to manually restore the target directory override
         CARGO_TARGET_DIR: ../target
 
+  unused-sdk-dependencies:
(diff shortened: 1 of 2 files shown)


jdisanti:
[CI on commit c2bcd3d](https://github.com/awslabs/smithy-rs/pull/936/commits/c2bcd3d59b35ae257d32c2e5088e52d829544b66) demonstrates it detecting unused dependencies.

Velfi (approved):

rcoh (approved):

PR #942, opened 2021-12-08
Reorganize CI and add per-crate CI checks

jdisanti:
## Motivation and Context

**Reorganizing CI:** Found a way to reduce copy/pasted setup config in the CI configs by using matrices. Also took this as an opportunity to rename all the workflows and walk away from the tiered terminology that we no longer use.

**Per-crate checks:** This adds the ability to run one-off checks against any runtime crate during the Rust Runtime CI. This can be used for checking the dependency graph for duplicate or undesired dependencies, or testing certain combinations of crate features.

Merging this will require updating the required CI Actions in the branch protection rules.


.github/workflows/ci-sdk.yaml
```diff
@@ -0,0 +1,210 @@
...
+        test:
+        - name: Unit Tests
+          run: cargo test $(cat service-with-tests)
+          working-directory: aws-sdk
+        - name: Docs
+          run: cargo doc --no-deps --document-private-items
+          working-directory: aws-sdk
+        - name: Clippy
+          run: cargo clippy
+          working-directory: aws-sdk
```
rcoh, line 70: this is super clever

tools/additional-per-crate-checks.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+#
+
+if [[ $# -ne 1 ]]; then
```
rcoh, line 7: I must be missing something, but I don't see this actually being run in CI?

jdisanti: It's at the very end of ci.yaml, as part of runtime tests.

rcoh: Oh! Github didn't expand that file and I thought it was deleted. Didn't realize it was edited 🤕

rcoh (approved):

PR #945, opened 2021-12-08
Migrate from axum to axum-core

### LINKED ISSUE #939: [Server] Upgrade to Axum 0.4
Axum [0.4](https://github.com/tokio-rs/axum/blob/main/axum/CHANGELOG.md#040-02-december-2021) have moved the `FromRequest` and `IntoResponse` traits into [axum-core](https://docs.rs/axum-core/latest/axum_core/) with the promise of a more stable API and less transitive dependencies.

We should upgrade the server codegen to use this crate instead of the Axum runtime one.

crisidev:
## Motivation and Context
Axum has moved its core traits and functionalities inside [axum-core](https://docs.rs/axum-core/latest/axum_core/) crate, promoting a more stable interface and less dependencies.

This PR aims to upgrade to axum-core and also removes part of our custom runtime implementation.

We also switched from [pin-project](https://crates.io/crates/pin-project) to [pin-project-lite](https://crates.io/crates/pin-project-lite) to lower the amount of dependencies we take and removed our custom implementation of `CloneBoxService` since it has been upstreamed inside Tower as `BoxCloneService`.

The hacky OperationHandler inclusion has been changed from just a complex string containing our Rust implementation to an inlineable module.

Closes #939 


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
```diff
@@ -52,7 +52,7 @@ sealed class RustDependency(open val name: String) : SymbolDependencyContainer {
  *
  * CodegenVisitor deduplicates inline dependencies by (module, name) during code generation.
  */
-class InlineDependency(
+open class InlineDependency(
```
crisidev, line 55: These are the main changes in the client codegen. They promote some code re-use while avoiding the client codegen to depend on the server codegen.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
```diff
@@ -133,88 +134,4 @@ class ServerOperationHandlerGenerator(
             $serverCrate::rejection::SmithyRejection: From<<B as $serverCrate::HttpBody>::Error>
         """
     }
-
```
crisidev, line 136: I am so happy this is gone :D

rust-runtime/aws-smithy-http-server/src/error.rs
```diff
@@ -44,7 +44,8 @@ pub struct Error {
 }
 
 impl Error {
-    pub(crate) fn new(error: impl Into<BoxError>) -> Self {
+    /// Create a new `Error` from a boxable error.
+    pub fn new(error: impl Into<BoxError>) -> Self {
```
crisidev, line 48: Probably this should stay `pub(crate)`.. Lemme check.

crisidev: Yep, it should.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
jdisanti: Why do you want to inherit from this as opposed to creating `ServerInlineDependency` as an `object` since it only has the one companion function on it? Is there a future use-case for polymorphism?

crisidev: Probably this is just my inexperience with Kotlin..I guess what you are saying is that i can just make `forRustFile` public and use it in my `ServerInlineDependency` companion object, avoiding to  open the class..

Thanks for pointing  this out, writing good kotlin is still a challenge for me..

crisidev: I'll fix this ASAP!

jdisanti: You can simplify the companion object out entirely:
```kotlin
object ServerInlineDependency {
    fun serverOperationHandler(runtimeConfig: RuntimeConfig): InlineDependency = // ...
}
```

crisidev: Done, looks nice now!

david-perez (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -204,12 +205,9 @@ private class ServerHttpProtocolImplGenerator(
...
                     Error(#{E})
                 }
-                ##[#{Axum}::async_trait]
-                impl #{Axum}::response::IntoResponse for $outputName {
-                    type Body = #{SmithyHttpServer}::Body;
-                    type BodyError = <Self::Body as #{SmithyHttpServer}::HttpBody>::Error;
-
-                    fn into_response(self) -> #{http}::Response<Self::Body> {
+                ##[#{AsyncTrait}::async_trait]
+                impl #{AxumCore}::response::IntoResponse for $outputName {
```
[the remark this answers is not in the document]
crisidev: Hum.. You are right, it makes the API more consistent for customers. I am going to change this.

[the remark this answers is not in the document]
crisidev: Scratch this, IntoResponse does not accept a generic type anymore.

jdisanti (approved):
Looks good from the client side!

PR #946, opened 2021-12-08
add: example showing how to use native-tls

Velfi:
## Motivation and Context

It's not currently possible to avoid using rustls

## Description

This PR makes it possible to use `native-tls` and disable `rustls` and includes an example for users demonstrating how to do it.

## Testing

I included a test that ensures `rustls` is not in use

## Checklist


aws/rust-runtime/aws-config/Cargo.toml
```diff
@@ -22,8 +22,8 @@ tcp-connector = ["tokio/net", "tower"]
 # SSO is not supported
 sso = []
 
-rustls = ["aws-smithy-client/rustls"]
-native-tls = ["aws-smithy-client/native-tls"]
+rustls = ["aws-smithy-client/rustls", "aws-sdk-sts/rustls", "aws-hyper/rustls"]
+native-tls = ["aws-smithy-client/native-tls", "aws-sdk-sts/native-tls", "aws-hyper/native-tls"]
```
rcoh, line 26: after #947 lands you won't need these extra features anymore 🥳

aws/rust-runtime/aws-endpoint/src/partition/mod.rs
```diff
@@ -51,7 +51,7 @@ impl ResolveAwsEndpoint for PartitionResolver {
 
 #[derive(Debug)]
 pub struct Partition {
-    id: &'static str,
+    _id: &'static str,
```
Velfi, line 54: Clippy was complaining about this locally and the field seemed unused so I added the underscore

aws/sdk/examples/using_native_tls_instead_of_rustls/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# aws-config pulls in rustls and several other things by default. We have to disable defaults in order to use native-tls
+# and then manually bring the other defaults back
+aws-config = { path = "../../build/aws-sdk/sdk/aws-config", default-features = false, features = ["default-provider", "native-tls", "rt-tokio", "dns", "tcp-connector"] }
+# aws-sdk-s3 brings in rustls by default so we disable that in order to use native-tls only
+aws-sdk-s3 = { package = "aws-sdk-s3", path = "../../build/aws-sdk/sdk/s3", default-features = false, features = ["native-tls"] }
+# aws-sdk-sts is the same as aws-sdk-s3
+aws-sdk-sts = { package = "aws-sdk-sts", path = "../../build/aws-sdk/sdk/sts", default-features = false, features = ["native-tls"] }
```
Velfi, line 16: Thanks to John's suggestion of removing default features for runtime crates, we don't even need to include them here!

aws/rust-runtime/aws-config/Cargo.toml
```diff
@@ -22,8 +22,8 @@ tcp-connector = ["tokio/net", "tower"]
 # SSO is not supported
 sso = []
 
-rustls = ["aws-smithy-client/rustls"]
-native-tls = ["aws-smithy-client/native-tls"]
+rustls = ["aws-smithy-client/rustls", "aws-sdk-sts/rustls", "aws-hyper/rustls"]
```
rcoh, line 25: do we actually use `aws-sdk-sts::new` or `from_conf`? If not, we don't even need these features

rust-runtime/aws-smithy-client/src/hyper_ext.rs
```diff
@@ -252,7 +252,7 @@ impl Builder {
     }
 }
 
-#[cfg(any(feature = "rustls", feature = "native_tls"))]
+#[cfg(any(feature = "rustls", feature = "native-tls"))]
```
Velfi, line 255: This was a really annoying thing to solve, I wish the compiler had assisted me here

rcoh: I bet you could add a clippy lint for features that don't exist...

Velfi: I spoke to Esteban about it and he advised me to file [this issue](https://github.com/rust-lang/rust/issues/91674)

aws/rust-runtime/aws-config/Cargo.toml
Velfi: You were right, I have removed these

aws/rust-runtime/aws-config/Cargo.toml
```diff
@@ -32,7 +32,7 @@ dns = ["tokio/rt"]
 default = ["default-provider", "rustls", "rt-tokio", "dns", "tcp-connector"]
 
 [dependencies]
-aws-sdk-sts = { path = "../../sdk/build/aws-sdk/sdk/sts", optional = true }
+aws-sdk-sts = { path = "../../sdk/build/aws-sdk/sdk/sts", default-features = false, optional = true }
```
jdisanti, line 35: Should the `rustls` and `native-tls` features in `aws-config` also enable the equivalent features on the STS crate? Or to reframe the question, is it working by coincidence right now without explicitly enabling those?

aws/sdk/examples/using_native_tls_instead_of_rustls/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// and that nothing is pulling in `rustls` as a dependency
+    #[test]
+    #[should_panic = "error: package ID specification `rustls` did not match any packages"]
+    fn test_rustls_is_not_in_dependency_tree() {
+        let cargo_location = std::env::var("CARGO").unwrap();
+        let cargo_command = std::process::Command::new(&cargo_location)
+            .arg("tree")
+            .arg("--invert")
+            .arg("rustls")
+            .output()
```
jdisanti, line 39: Clever!

Does this differentiate between something being wrong with the cargo binary (not able to execute or something) vs. exiting with a non-zero status?

Velfi: This only checks that Cargo writes a specific message to stderr when the command is run. The possible paths I see are:

1. rustls is not in tree, message is logged to stderr, we panic with that message and match on it with `should_panic`, it's a match so that's a test pass
2. some other failure occurs, message is logged to stderr, we panic with that message and match on it with `should_panic`, It's not a match so that's a test fail
3. rustls **is** in the tree, print tree and exit, test fails cause no panic

So, to answer your question: yes, it does differentiate

aws/rust-runtime/aws-config/Cargo.toml
Velfi: I had that previously and removed it because it didn't seem to matter. The test runs and doesn't detect `rustls` even though `aws-config` is a dependency so I believe we're good. Personally, my gut tells me to include the features but I can't articulate why so I'm ignoring the feeling.

rcoh: @jdisanti they should not: this is because the only purpose of those in the STS crate are to support `from_conf` (constructing the client _without_ a connection). But this crate will always provide a connection when constructing a client.

rcoh (approved):

PR #956, opened 2021-12-09
Implement a canary Lambda for the SDK

jdisanti:
## Motivation and Context
This implements a Rust Lambda that can be used as a canary for the SDK.

## Testing
- Created a Lambda and executed it with the code from this PR


tools/ci-cdk/canary-lambda/src/canary.rs
```diff
@@ -0,0 +1,81 @@
...
+        Self {
+            s3_bucket_name,
+            expected_transcribe_result,
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("{0}")]
+pub struct CanaryError(pub String);
```
rcoh, line 79: probably not a load bearing usage of proc macros

tools/ci-cdk/canary-lambda/write-cargo-toml.py
```diff
@@ -0,0 +1,72 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+#
+
+# Generates a Cargo.toml with the given AWS SDK version for this canary
+
+import argparse
+
+
+def main():
```
rcoh, line 12: I wonder if we should just use an actual template for this like https://docs.rs/handlebars/latest/handlebars/

tools/ci-cdk/canary-lambda/README.md
```diff
@@ -0,0 +1,32 @@
+Canary Lambda
+=============
+
+This is a Rust Lambda that exercises the AWS SDK for Rust in a real Lambda
```
rcoh, line 4: is there a one shot command to build/deploy with CDK? can you include that here if there is?

tools/ci-cdk/canary-lambda/build-bundle.sh
```diff
@@ -0,0 +1,30 @@
...
+set -e
+
+if [[ $# -ne 1 ]]; then
+    echo "Usage: $0 <sdk version>"
+    exit 1
+fi
+SDK_VERSION=$1
+
+cd `dirname $0`
+./write-cargo-toml.py --sdk-version ${SDK_VERSION}
```
rcoh, line 16: could even just be `Cargo.toml.templ` and use `sed`, or use a template language

tools/ci-cdk/canary-lambda/build-bundle.sh
```diff
@@ -0,0 +1,30 @@
...
+
+TARGET_PATH="$(git rev-parse --show-toplevel)/target/x86_64-unknown-linux-musl/release"
+pushd "${TARGET_PATH}" &>/dev/null
+
+BIN_NAME="bootstrap"
+BIN_SHA1_HASH="$(sha1sum ${BIN_NAME} | awk '{ print $1; }')"
+BUNDLE_NAME="canary-lambda-${BIN_SHA1_HASH}.zip"
+
+zip --quiet "${BUNDLE_NAME}" "${BIN_NAME}"
+popd &>/dev/null
```
rcoh, line 27: TIL popd

tools/ci-cdk/canary-lambda/write-cargo-toml.py
```diff
@@ -0,0 +1,72 @@
...
+    args = Args()
+
+    with open("Cargo.toml", "w") as file:
+        lines = [
+            '# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.',
```
rcoh, line 17: reason to not have this just be a multiline string? the template seems static?

tools/ci-cdk/canary-lambda/write-cargo-toml.py
```diff
@@ -0,0 +1,72 @@
...
+
+# Generates a Cargo.toml with the given AWS SDK version for this canary
+
+import argparse
+
+
+def main():
+    args = Args()
+
+    with open("Cargo.toml", "w") as file:
```
rcoh, line 15: seems like this might be a little clearer as a multiline string with string interpolation, (or are you trying to have this work on Python 2?)

rcoh (approved):
LGTM, really the only thing I might change is how the Cargo.toml is generated but this definitely isn't a blocking comment.

tools/ci-cdk/canary-lambda/write-cargo-toml.py
jdisanti: I'm inclined to keep it in this Python script just to avoid having to pull dependencies or compile something in order to generate it.

tools/ci-cdk/canary-lambda/README.md
jdisanti: IIRC, CDK doesn't provide this directly, but you can add a target to the `package.json` that does the build for you, and have the CDK synth or deploy depend on that target.

tools/ci-cdk/canary-lambda/write-cargo-toml.py
jdisanti: Reason I opted for this was because Python doesn't appear to have a built-in `trimIndent()` like Kotlin does, so the indentation would be ugly for the multi-line string.

rcoh: ah. I would probably do:
```python
base_template = """
# copyright ...
[dependencies] 
... 
...
""".strip()

def make_cargo_toml():
   ...
```

PR #957, opened 2021-12-10
Include non-service-specific examples in the generated workspace

jdisanti:
## Motivation and Context
We started adding examples that are not service-specific, but didn't update CI to correctly generate the Cargo workspace with them. This leads to compile errors in aws-sdk-rust when you try to compile these examples, since Cargo thinks they should be in a workspace but they're not.

This PR makes sure these examples are listed in the generated root Cargo.toml.

## Testing
- Tested locally with `./gradlew aws:sdk:assemble` and then `cargo build` with the generated `Cargo.toml` (just to make sure the manifests were valid; didn't wait for it to finish building).
 
## Checklist


aws/sdk/build.gradle.kts
```diff
@@ -284,7 +284,19 @@ fun generateCargoWorkspace(services: List<AwsService>): String {
     val generatedModules = services.map { it.module }.toSet()
     val examples = projectDir.resolve("examples")
         .listFiles { file -> !file.name.startsWith(".") }.orEmpty().toList()
-        .filter { generatedModules.contains(it.name) }
+        .filter { file ->
```
rcoh, line 287: Might be worth refactoring this into buildSrc, I think we even have a toml parser in tree already if you wanted to do it more directly

rcoh (approved):
LGTM!

Velfi (approved):

aws/sdk/build.gradle.kts
jdisanti: Yeah, I agree. I'm going to hold off on it for this PR though, since I'm refactoring this area pretty heavily for splitting the release process.

PR #962, opened 2021-12-13
Fix generation of hash sets to actually use `std::collections::HashSet`

david-perez:

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
@@ -79,8 +79,8 @@ sealed class RustType {
         override val namespace = Namespace
 
         companion object {
-            const val Type = "Vec"
-            const val Namespace = "std::vec"
+            const val Type = "HashSet"
+            const val Namespace = "std::collections"
             val RuntimeType = RuntimeType(name = Type, namespace = Namespace, dependency = null)
         }
     }


rcoh:
You need to be sure this is what you actually want–we didn't use hashsets
for a few reasons, mainly they produce nondeterministic serialization.

rcoh:
ah yeah you won't be able to do this unless you do it only for the server. We made it a `Vec` for a few reasons:
- backwards compatibility. Because floats aren't hashable if a float key was added in the future we would lose derive Hash. There are ways around this (eg. manual hash derivation)
- deterministic behavior: if you store values in a hashset there isn't a practical way to serialize them in a deterministic way. This makes dealing with protocol tests much tricker

Velfi:
> ah yeah you won't be able to do this unless you do it only for the server. We made it a `Vec` for a few reasons:
> 
>     * backwards compatibility. Because floats aren't hashable if a float key was added in the future we would lose derive Hash. There are ways around this (eg. manual hash derivation)
> 
>     * deterministic behavior: if you store values in a hashset there isn't a practical way to serialize them in a deterministic way. This makes dealing with protocol tests much trickier

Maybe the outcome of this PR should be to add comments to that effect in `RustTypes.kt`?

david-perez:
Interesting. I think there's an even stronger reason to use `Vec`. I've just learnt that [the Smithy spec](https://awslabs.github.io/smithy/1.0/spec/core/model.html#set) says "Sets MUST be insertion ordered." The only programming languages whose stdlibs provide insertion-ordered set types that I know of are [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) and Java's `LinkedHashSet`.

Velfi:
Alternatively, we could use something like [IndexSet](https://docs.rs/indexmap/1.7.0/indexmap/set/struct.IndexSet.html)

rcoh:
yeah we considered `IndexSet` but we wouldn't want to expose that publicly (we would've needed to create a wrapper in `aws-smithy-types` (which is still not out of the question).

I don't remember the insertion ordered requirement @david-perez but TIL!

rcoh:
closing, a comment was added clarifying the behavior

PR #963, opened 2021-12-13
[Server] Set http extensions for Request / Response

crisidev:
## Motivation and Context
We are starting to work on metrics reporting and there are certain things we need to pass around Tower layers to be able to record and produce metrics.

## Description
This PR allows to store important information in `http::Request`/`http::Response` `Extensions` that can be later used in a metrics reporting `Tower::Layer`.

In the `Request` we are storing:
* Operation name
* Model namespace

In the `Response` we are storing:
* Framework errors (IE deserialization errors, unmodeled errors)
* User errors (modeled errors)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
```diff
@@ -68,12 +68,18 @@ class ServerCombinedErrorGenerator(
                     rust("matches!(&self, ${symbol.name}::${errorSymbol.name}(_))")
                 }
             }
+            writer.rust("/// Returns the error name string by matching the correct variant.")
+            writer.rustBlock("pub fn name(&self) -> String") {
+                delegateToVariants {
+                    rust("_inner.name()")
+                }
+            }
```
crisidev, line 76: I have added this `name()` method which returns the error name as a String. This is the only change to the codegen and from what I can see in the errorTrait spec, it shouldn't impact anything. Let's see what @awslabs/rust-sdk-owners think about this and what the CI will say 😄

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
```diff
@@ -68,12 +68,18 @@ class ServerCombinedErrorGenerator(
...
         }
 
         writer.rustBlock("impl #T for ${symbol.name}", RuntimeType.StdError) {
             rustBlock("fn source(&self) -> Option<&(dyn #T + 'static)>", RuntimeType.StdError) {
                 delegateToVariants {
-                    rust("Some(_inner)");
+                    rust("Some(_inner)")
```
crisidev, line 82: Hum, I have to double check why I have removed this `;`. It is probably just an error from my side.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/ErrorGenerator.kt
```diff
@@ -85,6 +85,12 @@ class ErrorGenerator(
                 """
                 /// Returns the error message.
                 pub fn message(&self) -> Option<&str> { $message }
+
+                ##[doc(hidden)]
+                /// Returns the error name.
+                pub fn name(&self) -> String {
+                    String::from("${shape.id.name}")
+                }
```
crisidev, line 93: I have added this `name()` method which returns the error name as a String. The server is using this to store any modeled error name inside `http::Extensions` to help with metrics reporting.

This is the only change to the codegen and from what I can see in the errorTrait spec, it shouldn't impact anything. Let's see what @awslabs/rust-sdk-owners think about this and what the CI will say 😄

crisidev: Also I was thinking that this can probably become

```rust
    ##[doc(hidden)]
    /// Returns the error name.
    pub fn name(&self) -> &'static str {
        ${shape.id.name.dq()}
    }
```

Thoughts?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
crisidev, line 76: Also this can return a `&'static str` TBH..

rcoh (approved):
LGTM, we _may_ want to only render that method on the server. I'm not sure what benefit it has for clients and it will add a decent amount of generated code.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/ErrorGenerator.kt
crisidev: 👍🏼

crisidev:
> LGTM, we _may_ want to only render that method on the server. I'm not sure what benefit it has for clients and it will add a decent amount of generated code.

I'll figure out how to do it :) The CI is barfing a little on the SDK gen since `name()` is already defined for some types, so we need to ensure this happens only on the server. https://github.com/awslabs/smithy-rs/runs/4506867653?check_suite_focus=true

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
crisidev: Yep, it was the autoformat I think.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -289,6 +302,21 @@ private class ServerHttpProtocolImplGenerator(
...
     }
 
+    /*
+     * Set `http::Extensions` for the current request. They can be used later for things like metrics, logging, etc..
+     */
+    private fun setHttpExtensions(operationShape: OperationShape): String {
+        val namespace = operationShape.id.getNamespace()
+        val operationName = symbolProvider.toSymbol(operationShape).name
+        // TODO: remove when we support streaming types
+        val requestPrefix = if (operationShape.inputShape(model).hasStreamingMember(model)) { "_" } else { "" }
```
david-perez, line 312: What is this prefix for?

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -39,6 +39,24 @@ use async_trait::async_trait;
...
 
+/// Extension type used to store the Smithy model namespace.
+#[derive(Debug, Clone)]
+pub struct ExtensionNamespace(pub &'static str);
+
+/// Extension type used to store the Smithy operation name.
+#[derive(Debug, Clone)]
+pub struct ExtensionOperationName(pub &'static str);
+
+/// Extension type used to store the type of framework error caught during exection.
```
david-perez, line 50:
```suggestion
/// Extension type used to store the type of framework error caught during execution.
```

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -39,6 +39,24 @@ use async_trait::async_trait;
...
+/// These are unmodeled error, defined in the runtime crates.
+#[derive(Debug, Clone)]
+pub struct ExtensionUnmodeledError(pub String);
```
david-perez, line 53: Maybe `ExtensionRejection` is a better name? We've been calling rejections the "errors" returned by the framework. I don't like calling them errors because dealing with malformed requests and rejecting them is frequent enough to be considered part of normal execution flow.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: Forget about this, I just noticed I made a mistake and this is not needed at all.

rust-runtime/aws-smithy-http-server/src/extension.rs
crisidev: Yeah, good suggestion, I'll fix this. Naming things is hard 😛

crisidev: Refactored, thanks for the suggestion!

david-perez (approved):

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -34,10 +34,51 @@
...
+#[derive(Debug, Clone)]
+pub struct ExtensionRejection(pub Cow<'static, str>);
+
+/// Extension type used to store the type of user defined error returned by an operation.
+/// These are modeled errors, defined in the Smithy model.
+#[derive(Debug, Clone)]
+pub struct ExtensionModeledError(pub Cow<'static, str>);
+
+/// Implement new for all Extension holding a Cow<'static, str>.
+macro_rules! impl_extension_new_static {
```
david-perez, line 61: Move to `macros.rs`

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -34,10 +34,51 @@
...
+/// These are unmodeled error, or rejection, defined in the runtime crates.
+#[derive(Debug, Clone)]
+pub struct ExtensionRejection(pub Cow<'static, str>);
+
+/// Extension type used to store the type of user defined error returned by an operation.
+/// These are modeled errors, defined in the Smithy model.
+#[derive(Debug, Clone)]
+pub struct ExtensionModeledError(pub Cow<'static, str>);
+
+/// Implement new for all Extension holding a Cow<'static, str>.
```
david-perez, line 60:
```suggestion
/// Implement `new` for all `Extension`s holding a `Cow<'static, str>`.
```

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -34,10 +34,51 @@
...
+            #[doc = "Returns a new `"]
+            #[doc = $sname]
+            #[doc = "`."]
+            pub fn new<S: Into<String>>(value: S) -> $name {
+                $name(std::borrow::Cow::from(value.into()))
+            }
+        }
+    };
+
+    ($name:tt) => {
```
david-perez, line 73: Better to use `concat!` for this: https://stackoverflow.com/a/68783193

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -34,10 +34,51 @@
...
-use super::rejection::{ExtensionRejection, ExtensionsAlreadyExtracted, MissingExtension};
+use super::rejection::{ExtensionHandlingRejection, ExtensionsAlreadyExtracted, MissingExtension};
 use async_trait::async_trait;
 use axum_core::extract::{FromRequest, RequestParts};
-use std::ops::Deref;
+use std::{borrow::Cow, ops::Deref};
+
+/// Extension type used to store the Smithy model namespace.
+#[derive(Debug, Clone)]
+pub struct ExtensionNamespace(pub Cow<'static, str>);
```
david-perez, line 44: If we have `new()` methods, we can make the wrapped type private in all of these.

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -34,10 +34,51 @@
...
+
+/// Implement new for all Extension holding a Cow<'static, str>.
+macro_rules! impl_extension_new_static {
+    ($name:ident, $sname:expr) => {
+        impl $name {
+            #[doc = "Returns a new `"]
+            #[doc = $sname]
+            #[doc = "`."]
+            pub fn new<S: Into<String>>(value: S) -> $name {
+                $name(std::borrow::Cow::from(value.into()))
```
david-perez, line 68: I'm confused. This will now convert the `&str` into `String` performing heap-allocation before wrapping it in the `Cow`.

rust-runtime/aws-smithy-http-server/src/extension.rs
crisidev: Huh, right, let me change.

rust-runtime/aws-smithy-http-server/src/extension.rs
crisidev: Didn't know, will check.

rust-runtime/aws-smithy-http-server/src/extension.rs
crisidev: Ok.

rust-runtime/aws-smithy-http-server/src/extension.rs
crisidev: Yes, you are right, it does. I wanted to have a single implementation, but as you mentioned it will do allocation also when it is not needed. I am going to change the macro and only use it in the types that are not allocated and do a manual implementation for `ExtensionModeledError`.

PR #965, opened 2021-12-13
Fix query string signing bug

Velfi:
## Motivation and Context

fixes [aws-sdk-rust #330](https://github.com/awslabs/aws-sdk-rust/issues/330) 

## Description

We weren't percent encoding all necessary characters so those have been added to the list ([see this wiki article for more context](https://en.wikipedia.org/wiki/Percent-encoding#Character_data)). This PR updates `aws-sigv4` to use the percent encoding defined in the smithy runtime crates and also adds some tests to ensure proper signing.

## Testing

I wrote tests to ensure the issue is fixed

## Checklist


aws/sdk/integration-tests/s3/Cargo.toml
```diff
@@ -8,18 +8,20 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dev-dependencies]
+aws-config = { path = "../../build/aws-sdk/sdk/aws-config" }
 aws-http = { path = "../../build/aws-sdk/sdk/aws-http" }
 aws-hyper = { path = "../../build/aws-sdk/sdk/aws-hyper" }
 aws-sdk-s3 = { path = "../../build/aws-sdk/sdk/s3" }
+aws-sdk-sts = { path = "../../build/aws-sdk/sdk/sts" }
```
jdisanti, line 15: Unused?

aws/rust-runtime/aws-sigv4/src/http_request/url_escape.rs
```diff
@@ -3,41 +3,12 @@
...
-    .add(b'=')
-    .add(b'%');
-
-const QUERY_SET: &AsciiSet = &BASE_SET.add(b'/');
-const PATH_SET: &AsciiSet = BASE_SET;
+use aws_smithy_http::{label, query};
 
 pub(super) fn percent_encode_query(value: &str) -> String {
-    percent_encoding::percent_encode(value.as_bytes(), QUERY_SET).to_string()
+    query::fmt_string(value)
```
jdisanti, line 9: Where were the `query::fmt_string` and `label::fmt_string` functions added?

aws/sdk/integration-tests/s3/tests/query-strings-are-correctly-encoded.rs
```diff
@@ -0,0 +1,165 @@
...
+use aws_http::user_agent::AwsUserAgent;
+use aws_hyper::AwsMiddleware;
+use aws_sdk_s3::operation::ListObjectsV2;
+use aws_sdk_s3::{Credentials, Region};
+use aws_smithy_client::test_connection::capture_request;
+use aws_smithy_client::Client as CoreClient;
+use std::time::{Duration, UNIX_EPOCH};
+
+pub type Client<C> = CoreClient<C, AwsMiddleware>;
+
```
jdisanti, line 15: Seems to me like these kinds of tests should be in `aws-sigv4`

aws/rust-runtime/aws-sigv4/src/http_request/url_escape.rs
Velfi: They were already defined in `aws-smith-http`. I'm using those ones.

aws/sdk/integration-tests/s3/Cargo.toml
Velfi: I don't know exactly why but if I don't include this, it errors out when trying to build and run the test because it can't find aws_smithy_client. Once I added this, the issue went away.

aws/sdk/integration-tests/s3/tests/query-strings-are-correctly-encoded.rs
Velfi: I can move it there, I only put it in S3 because that's where the user originally encountered the issue

rcoh: IMO, both is good—It's nice to test S3 directly so that we have a test that covers end-to-end codegen and wiring but a specific test is also not a bad idea

aws/sdk/integration-tests/s3/tests/query-strings-are-correctly-encoded.rs
```diff
@@ -0,0 +1,165 @@
...
+        .credentials_provider(creds)
+        .region(Region::new("us-east-1"))
+        .build();
+    let (conn, rcvr) = capture_request(None);
+
+    let client = Client::new(conn.clone());
+
+    let mut op = ListObjectsV2::builder()
+        .bucket("test-bucket")
+        .prefix(r#":/?#[]@!$&'"(){}~`*+,;=%<>"#)
```
rcoh, line 35: can we generate this with every printable ascii character for completeness?

rust-runtime/aws-smithy-http/src/urlencode.rs
```diff
@@ -33,7 +36,10 @@ pub const BASE_SET: &AsciiSet = &CONTROLS
     .add(b'%')
     .add(b'<')
     .add(b'>')
-    .add(b'"');
+    .add(b'"')
+    .add(b'^')
+    .add(b'`')
+    .add(b'\\');
 
 pub const LABEL_SET: &AsciiSet = &BASE_SET.add(b'`').add(b'?').add(b' ');
```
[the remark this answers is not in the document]
Velfi: We actually now include all of these chars in the base set, I'll just remove this

PR #970, opened 2021-12-14
RFC: Split the release process between `smithy-rs` and `aws-sdk-rust`

jdisanti:
[Rendered](https://github.com/awslabs/smithy-rs/blob/jdisanti-split-release-process-rfc/design/src/rfcs/rfc0007_split_release_process.md)


design/src/rfcs/rfc0007_split_release_process.md
```diff
@@ -0,0 +1,143 @@
...
+We could keep a copy of the Smithy runtime crates used by the SDK in `aws-sdk-rust`.
+This would look mostly the same as the proposed solution, except that rather than cloning
+`smithy-rs` in CI, we would use the local snapshot. The CI for the `main` branch CI would
+need to modify the crate manifests so that the Smithy runtime crates published to
+crates.io are used instead of the local copies.
+
+This has the advantage that a customer having an issue with the current release could get
+a fix sooner by patching their own project's crate manifest to use the `aws-sdk-rust/next`
+branch before a release is cut. However, it also has the disadvantage that the changes
+CI has to make while verifying the `main` branch are more intrusive.
```
jdisanti, line 129: We should discuss this alternative vs. the proposed solution in depth. I think the alternative might actually be more compelling, and the intrusiveness of the manifest changes is pretty minor.

design/src/rfcs/rfc0007_split_release_process.md
```diff
@@ -0,0 +1,143 @@
...
+step beforehand to ensure the correct Smithy runtime crates are in place. The process as a whole
+looks as follows:
+
+1. Make a shallow clone of `aws-sdk-rust/next`
+2. Retrieve the `smithy-rs` commit hash that was used to generate the SDK from a file
+   that was generated alongside the rest of the build artifacts from `smithy-rs` and
+   copied into `aws-sdk-rust`.
+3. Make a shallow clone of `smithy-rs` at the correct commit hash
+4. Use a script to add a `[patch]` section to all the AWS SDK crates to point to the
+   Smithy runtime crates from the local clone of `smithy-rs`.
```
jdisanti, line 84: To preempt questions on this: It needs to be a local clone of `smithy-rs` rather than a git dependency since cargo doesn't provide any way to specify the path within a git repository (the package must be locate-able from the repository root).

design/src/rfcs/rfc0007_split_release_process.md
```diff
@@ -0,0 +1,143 @@
...
+   generated against _both before AND after they have been published to crates.io_.
+
+Background: How Publishing Worked Before
+----------------------------------------
+
+The publish process to crates.io relied on copying all the Smithy runtime crates
+into the final `aws-sdk-rust` repository. Overall, the process looked as follows:
+
+1. `smithy-rs` generates a complete `aws-sdk-rust` source bundle at CI time
+2. The releaser copies the generated bundle over to `aws-sdk-rust`
```
rcoh, line 53: Maybe replace releaser with engineer to clarify that it's a human and not an automated job

design/src/rfcs/rfc0007_split_release_process.md
```diff
@@ -0,0 +1,143 @@
...
+-----------------
+
+CI in `smithy-rs` will be revised to generate two separate build artifacts where it generates
+just an SDK artifact previously. Now, it will have two build targets that get executed from CI
+to generate these artifacts:
+
+- `rust-runtime:assemble` - Generates a publish-ready bundle of Smithy runtime crates.
+- `aws:sdk:assemble` - Generates a publish-ready bundle of AWS runtime crates and SDK crates
+  _without_ the Smithy runtime crates. This build artifact will now also include a file that has
+  the commit hash of `smithy-rs` that it was generated from.
```
rcoh, line 69: And maybe if the working directory is dirty the resulting artifact is intentionally unpublishable?

design/src/rfcs/rfc0007_split_release_process.md
rcoh, line 84: This is a clever approach, I like this. I guess the only downside is that the next branch is not something that actually "works" though right? (As in: it won't work as a target for cargo)

design/src/rfcs/rfc0007_split_release_process.md
```diff
@@ -0,0 +1,143 @@
...
+
+The `aws-sdk-rust` repository will have a new `next` branch that has its own set of CI workflows
+and branch protection rules. The releaser will take the `aws:sdk:assemble` artifact and apply it
+directly to this `next` branch as would have previously been done against the `main` branch.
+The `next` branch's CI will behave similar to the previous CI, except that it runs an additional
+step beforehand to ensure the correct Smithy runtime crates are in place. The process as a whole
+looks as follows:
+
+1. Make a shallow clone of `aws-sdk-rust/next`
```
rcoh, line 78: What if a CI job maintains a `next` branch on smithy-rs that can be used as a target for the aws-sdk-rust crates? We could also produce automated tags of the next branch for releases...too complex?

The advantage would be that the next branch would continue to work. The autodeployer would have the runtime dependencies point to a specific commit on the smithy-rs/next branch...

design/src/rfcs/rfc0007_split_release_process.md
crisidev: My main question about the suggested vs alternative approach is what do we gain by not having a local copy of the runtime crates inside `aws-sdk-rust`? Separation of concerns?

I am asking this because the alternative approach seems to be involving less "machinery" in the release process. It is however true that the CI changes will be more intrusive..

jdisanti: After thinking about this more, I'm preferring the alternate approach to the proposed approach. I will swap the two in the next revision.

design/src/rfcs/rfc0007_split_release_process.md
jdisanti: @rcoh proved me wrong here 😅

design/src/rfcs/rfc0007_split_release_process.md
jdisanti: "Releaser" was defined above as either an engineer or an automated process (or a combination of the two), so I think this is fine.

design/src/rfcs/rfc0007_split_release_process.md
jdisanti: This is a good idea! Will incorporate.

design/src/rfcs/rfc0007_split_release_process.md
jdisanti: This seems like too many moving parts to me.

design/src/rfcs/rfc0007_split_release_process.md
jdisanti: I think we should keep this in our toolbox for the future, but the alternate approach is simpler over all and accomplishes our requirements while also making the `next` branch directly consumable from cargo.

rcoh (approved):
LGTM!

PR #971, opened 2021-12-14
`rust-server-codegen`: add support for `httpQuery` and `httpQueryParams` traits

david-perez:

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
@@ -688,17 +688,14 @@ private class ServerHttpProtocolImplGenerator(
                 rust("let mut seen_${it.memberName.toSnakeCase()} = false;")
             }
             queryBindingsTargettingCollection.forEach {
-                val foo = model.expectShape(it.member.target)
-                rust("let mut ${it.memberName.toSnakeCase()}: Vec<String> = Vec::new();")
+                rust("let mut ${it.memberName.toSnakeCase()} = Vec::new();")
             }
 
             rustBlock("for (k, v) in pairs") {
-                val (queryBindingsTargettingCollection, queryBindingsTargettingSimple) =
-                    queryBindings.partition { model.expectShape(it.member.target) is CollectionShape }
                 queryBindingsTargettingSimple.forEach {
                     val deserializer = generateParsePercentEncodedStrFn(it)
                     rustTemplate("""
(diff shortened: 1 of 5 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -605,13 +611,164 @@ private class ServerHttpProtocolImplGenerator(
...
+            QueryParamsTargetMapValueType.LIST
+        } else if (targetMapValue.isSetShape) {
+            QueryParamsTargetMapValueType.SET
+        } else {
+            throw ExpectationNotMetException("""
+                @httpQueryParams trait applied to non-supported target
+                $targetMapValue of type ${targetMapValue.type}
+                """.trimIndent(),
+                targetMapValue.sourceLocation)
+        }
```
crisidev, line 648: Probably just style, but I like more `when` than chained if / else if / else. Not a blocker.

crisidev (approved):
LGTM. I'd like to merge this, so I can do some cleanup this week as I believe we are still logging warnings about not supported protocols that can go away now.

crisidev:
Close #971

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -605,13 +611,164 @@ private class ServerHttpProtocolImplGenerator(
...
+                    "HashMap" to RustType.HashMap.RuntimeType,
+                )
+            }
+            val (queryBindingsTargettingCollection, queryBindingsTargettingSimple) =
+                queryBindings.partition { model.expectShape(it.member.target) is CollectionShape }
+            queryBindingsTargettingSimple.forEach {
+                rust("let mut seen_${it.memberName.toSnakeCase()} = false;")
+            }
+            queryBindingsTargettingCollection.forEach {
+                rust("let mut ${it.memberName.toSnakeCase()} = Vec::new();")
```
david-perez, line 691: Note here that we always use a `Vec`. If the member shape targets a Smithy set, we're handing `Vec`s to the user anyway. See the discussion in #962.

This means however that we're not deduplicating in case the query string contains repeated values with the same key, and the user chose a Smithy set.

david-perez: I filed https://github.com/awslabs/smithy-rs/issues/972

PR #979, opened 2021-12-15
Smithy Client cannot be an optional dependency

rcoh:
## Motivation and Context
When the client feature was removed, the smithy-client dependency was not made required. This causes compilation failures in workspaces where Smithy-client is not already included (eg. if `aws-config` is not in tree)

## Description
Update smithy-client dep to never be optional

## Testing

- [x] manually build client with --no-default-features. We should add this to the additional-checks CI

## Checklist


Velfi (approved):

Velfi:
Sorry about that

PR #984, opened 2021-12-15
Improve configurability of connectors

rcoh:
## Motivation and Context
- https://github.com/awslabs/aws-sdk-rust/issues/334
- https://github.com/awslabs/aws-sdk-rust/issues/338

## Description
Currently it's possible to use a fully custom connector with generated clients, but it's more cumbersome when using `aws-config`. Furthermore, the type of `TcpConnector` wasn't ideal since it didn't actually work when HTTPS was needed.

This change:
- Allows `tcp_connector` to accept anything that Hyper can use
- Adds a `provider_config` method (perhaps this should be named `configure` for symmetry?) to `ConfigLoader`
- Adds an example of how one can set a custom CA store with Rustls

## Testing
- [x] cargo hack
- [x] it/ut 

## Checklist


aws/rust-runtime/aws-config/src/provider_config.rs
```diff
@@ -266,7 +266,7 @@ impl ProviderConfig {
     /// Override the HTTPS connector for this configuration
     ///
     /// **Warning**: Use of this method will prevent you from taking advantage of the timeout machinery.
-    /// Consider `with_tcp_connector`.
+    /// Consider [`ProviderConfig::with_tcp_connector`].
```
Velfi, line 269: Will this prevent users from taking advantage of all timeouts or just the HTTP-specific ones?

Also, perhaps we could note this as a possible way to implement non-standard timeout mechanisms?

aws/sdk/examples/ca-certs/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+authors = ["rcoh@amazon.com>"]
+edition = "2018"
+
+description = "An example demonstrating setting a custom root certificate with rustls"
+
+[dependencies]
+aws-config = { path = "../../build/aws-sdk/sdk/aws-config" }
+aws-smithy-client = { path = "../../build/aws-sdk/sdk/aws-smithy-client" }
+# bringing out own HTTPs so no need for the default features
```
Velfi, line 12:
```suggestion
# bringing our own HTTPs so no need for the default features
```

aws/rust-runtime/aws-config/src/provider_config.rs
rcoh: just the HTTP ones

aws/sdk/examples/ca-certs/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+use aws_config::provider_config::ProviderConfig;
+use aws_smithy_client::erase::boxclone::BoxCloneService;
+use aws_smithy_client::hyper_ext;
+use rustls::RootCertStore;
+
+#[tokio::main]
+async fn main() {
+    let root_store: RootCertStore = todo!();
```
Velfi, line 13: Is this meant to be a `todo`? Shouldn't there at least be a comment telling people what to do?

rust-runtime/aws-smithy-client/src/never.rs
```diff
@@ -54,10 +54,8 @@ impl<Req, Resp, Err> NeverService<Req, Resp, Err> {
 pub type NeverConnector =
     NeverService<http::Request<SdkBody>, http::Response<SdkBody>, ConnectorError>;
 
-#[cfg(feature = "rt-tokio")]
-#[allow(dead_code)]
-/// A service where the underlying TCP connection never connects
-pub type NeverConnected = NeverService<Uri, tokio::net::TcpStream, BoxError>;
+/// A service where the underlying TCP connection never connects.
+pub type NeverConnected = NeverService<Uri, stream::EmptyStream, BoxError>;
```
Velfi, line 58: nice

Velfi (approved):

PR #986, opened 2021-12-16
Produce two publishable bundles in CI

jdisanti:
## Motivation and Context
This implements the `smithy-rs` part of https://github.com/awslabs/smithy-rs/pull/970.

## Testing
- Tested publish tool changes by hand
- Manually inspected and verified build artifacts


tools/publisher/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
```
rcoh, line 1: assuming this commit is straight copy/paste I did not review it

aws/sdk/build.gradle.kts
```diff
@@ -260,10 +200,9 @@ tasks.register("relocateAwsRuntime") {
     dependsOn("copyAllRuntimes")
     doLast {
         // Patch the Cargo.toml files
-        awsModules.forEach { moduleName ->
+        Crates.AWS_RUNTIME.forEach { moduleName ->
```
rcoh: might be worth thinking of a different naming scheme here, my brain doesn't expect this to be a list.

buildSrc/src/main/kotlin/Crates.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+object Crates {
+    val AWS_RUNTIME = listOf(
```
rcoh: I'm not sure about pushing this information into buildSrc, this feels like maybe it belongs in the individual builds? Could be convinced otherwise though

buildSrc/src/main/kotlin/ManifestPatcher.kt
```diff
@@ -0,0 +1,31 @@
+import java.io.File
+
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+private fun rewriteCrateVersion(line: String, version: String): String = line.replace(
```
rcoh, line 8: since this stuff is in buildSrc now I wonder if we should add some unit tests...

aws/sdk/build.gradle.kts
```diff
@@ -239,6 +241,17 @@ task("generateCargoWorkspace") {
     outputs.upToDateWhen { false }
 }
 
+tasks.register<Exec>("fixManifests") {
+    description = "Run the publisher tool's `fix-manifests` sub-command on the generated services"
+    workingDir(rootProject.projectDir.resolve("tools/publisher"))
+    commandLine("cargo", "run", "--", "fix-manifests", "--location", outputDir.absolutePath)
```
rcoh, line 247: I think this needs `outputDir` registered as an output and `tools/publisher` registered as an input

buildSrc/src/main/kotlin/Crates.kt
```diff
@@ -4,7 +4,7 @@
  */
 
 object Crates {
-    val AWS_RUNTIME = listOf(
+    val AWS_SDK_RUNTIME = listOf(
```
rcoh, line 7: just curious way all caps and underscores for these instead of `AwsSdkRuntime`?

buildSrc/src/main/kotlin/Crates.kt
```diff
@@ -4,7 +4,7 @@
  */
 
 object Crates {
```
rcoh, line 6: I think my brain would expect these to be lists if the object was named `CrateSet` or something like that

tools/publisher/src/package.rs
```diff
@@ -194,15 +194,18 @@ pub enum Error {
 }
 
 /// Discovers all Cargo.toml files under the given path with a depth limit of 1.
```
rcoh, line 196: is `depth limit of 1` still correct here?

.github/workflows/ci.yaml
```diff
@@ -115,3 +115,32 @@ jobs:
         RUSTDOCFLAGS: -D warnings
     - name: Additional per-crate checks
       run: ./tools/additional-per-crate-checks.sh ${{ matrix.runtime }}/rust-runtime/
+
+  runtime-bundle:
+    name: Produce smithy-rs runtime bundle
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
rcoh, line 124: seems like potentially a good use of matrix?

tools/publisher/src/subcommand/fix_manifests.rs
```diff
@@ -125,15 +126,36 @@ fn fix_dep_sets(versions: &BTreeMap<String, Version>, metadata: &mut toml::Value
     Ok(changed)
 }
 
+fn fix_publish_flag(manifest_path: &Path, metadata: &mut toml::Value) -> Result<bool> {
```
rcoh, line 129: nit: naming: I'd name this something a bit more specific `fix` implies it's broken, maybe `block_local_publish` or something?

.github/workflows/ci-sdk.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
       matrix:
         test:
         - name: Unit Tests
-          run: cargo test $(cat service-with-tests)
+          run: cargo test --all-features
```
rcoh, line 65: is this irrelevant now? (`services-with-tests`)?

tools/publisher/Cargo.toml
jdisanti: That's correct.

buildSrc/src/main/kotlin/Crates.kt
jdisanti: Just what I'm used to for constants from previous projects. No strong preference either way.

.github/workflows/ci.yaml
jdisanti: In this case, I don't think so since it needs additional tar/upload steps that can't be a part of the matrix.

.github/workflows/ci-sdk.yaml
jdisanti: As far as I can tell, nothing is generating this file, and that's been the case for a while.

buildSrc/src/main/kotlin/Crates.kt
jdisanti: I'm fine with `CrateSet` 👍

buildSrc/src/main/kotlin/Crates.kt
jdisanti: I moved it into buildSrc since I needed it from both `aws/sdk` and `rust-runtime`. I prefer to have only one place to update the list.

buildSrc/src/main/kotlin/ManifestPatcher.kt
jdisanti: Good call!

rcoh (approved):

PR #993, opened 2021-12-17
Improve the canary Lambda implementation

jdisanti:
## Description
This fixes a few problems that were uncovered when trying to set up the canary in GitHub Actions:
- The Lambda execution role didn't have permission to log to CloudWatch, so there was no log output on failure
- The Lambda was logging the S3 bucket name
- The build for the Lambda would fail because it didn't have an empty `[workspace]` in the Cargo.toml
- The S3 metadata comparison failure didn't give enough information (still not sure why it's failing now since it used to pass)

## Testing
- Ran the Lambda and verified it logs correctly now


tools/ci-cdk/canary-lambda/src/s3_canary.rs
```diff
@@ -63,8 +63,14 @@ pub async fn s3_canary(client: s3::Client, s3_bucket_name: String) -> anyhow::Re
     let mut result = Ok(());
     match output.metadata() {
         Some(map) => {
-            if map.get("something") != Some(&"テスト テスト!".to_string()) {
-                result = Err(CanaryError("S3 metadata was incorrect".into()).into());
+            // Option::as_deref doesn't work here since the deref of &String is String
+            let value = map.get("something").map(|s| s.as_str()).unwrap_or("");
```
rcoh, line 67: I don't totally understand why the old code didn't work, can you explain?

tools/ci-cdk/canary-lambda/src/s3_canary.rs
```diff
@@ -63,8 +63,14 @@ pub async fn s3_canary(client: s3::Client, s3_bucket_name: String) -> anyhow::Re
...
+            if value != "テスト テスト!" {
+                result = Err(CanaryError(format!(
+                    "S3 metadata was incorrect. Expected `テスト テスト!` but got `{}`.",
+                    value
+                ))
+                .into());
```
rcoh, line 73: might be nice to add a `#[test]` that can run this locally for a little easier testing

rcoh (approved):
LGTM!

tools/ci-cdk/canary-lambda/src/s3_canary.rs
jdisanti: There's no change in the logic. I just rearranged things so that I could output the actual value received as part of the error.

tools/ci-cdk/canary-lambda/src/s3_canary.rs
jdisanti: This is a good idea. I'll keep it in mind for the next time I need to run this.

PR #995, opened 2021-12-19
Setup server default RustSettings

crisidev:
## Motivation and Context
This PR involves 2 changes:
1) We depend on client codegen `RustSettings` to configure things like `includeFluentClient`. This PR gives the server its own configurable `ServerRustSettings`, with good defaults. For example `includeFluentClient = true` will generate a server SDK that doesn't build properly.
2) Move store of operation and namespace inside the response extensions or it will be unusable inside other layers. 

Signed-off-by: Bigo <1781140+crisidev@users.noreply.github.com>

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
@@ -25,9 +25,9 @@ import java.util.Optional
  * [formatTimeoutSeconds]: Timeout for running cargo fmt at the end of code generation
  */
 data class ServerCodegenConfig(
-    val renameExceptions: Boolean = true,
+    val renameExceptions: Boolean = false,
     val includeFluentClient: Boolean = false,
-    val addMessageToErrors: Boolean = true,
+    val addMessageToErrors: Boolean = false,
     val formatTimeoutSeconds: Int = 20,
     // TODO(EventStream): [CLEANUP] Remove this property when turning on Event Stream for all services
     val eventStreamAllowList: Set<String> = emptySet(),
@@ -38,9 +38,9 @@ data class ServerCodegenConfig(
                 CodegenConfig.fromNode(node)
             } else {
                 CodegenConfig(
-                    true,
                     false,
-                    true,
+                    false,
+                    false,
                     20,
                     emptySet()
                 )
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
@@ -5,17 +5,14 @@
 
 package software.amazon.smithy.rust.codegen.server.smithy
 
-import software.amazon.smithy.codegen.core.CodegenException
 import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.node.ObjectNode
-import software.amazon.smithy.model.shapes.ServiceShape
 import software.amazon.smithy.model.shapes.ShapeId
 import software.amazon.smithy.rust.codegen.smithy.CODEGEN_SETTINGS
 import software.amazon.smithy.rust.codegen.smithy.CodegenConfig
 import software.amazon.smithy.rust.codegen.smithy.RuntimeConfig
 import software.amazon.smithy.rust.codegen.smithy.RustSettings
 import java.util.Optional
-import java.util.logging.Logger
 
 /**
  * Configuration of codegen settings
@@ -68,23 +65,7 @@ class ServerRustSettings(
     val examplesUri: String? = null,
     private val model: Model
 ) {
-
-    /**
-     * Get the corresponding [ServiceShape] from a model.
-     * @return Returns the found `Service`
-     * @throws CodegenException if the service is invalid or not found
-     */
-    fun getService(model: Model): ServiceShape {
-        return model
-            .getShape(service)
-            .orElseThrow { CodegenException("Service shape not found: $service") }
-            .asServiceShape()
-            .orElseThrow { CodegenException("Shape is not a service: $service") }
-    }
-
     companion object {
-        private val LOGGER: Logger = Logger.getLogger(ServerRustSettings::class.java.name)
-
         /**
          * Create settings from a configuration object node.
          *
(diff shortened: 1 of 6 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
```diff
@@ -0,0 +1,83 @@
...
+ * [includeFluentClient]: Generate a `client` module in the generated SDK (currently the AWS SDK sets this to false
+ *   and generates its own client)
+ *
+ * [addMessageToErrors]: Adds a `message` field automatically to all error shapes
+ * [formatTimeoutSeconds]: Timeout for running cargo fmt at the end of code generation
+ */
+data class ServerCodegenConfig(
+    val renameExceptions: Boolean = true,
+    val includeFluentClient: Boolean = false,
+    val addMessageToErrors: Boolean = true,
```
david-perez, line 30: I think this should be removed or set to `false` in the server. They added it in the client because some AWS services were including error messages despite their models not modeling them. New Rust servers should model their error messages.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * [renameExceptions]: Rename `Exception` to `Error` in the generated SDK
+ * [includeFluentClient]: Generate a `client` module in the generated SDK (currently the AWS SDK sets this to false
+ *   and generates its own client)
+ *
+ * [addMessageToErrors]: Adds a `message` field automatically to all error shapes
+ * [formatTimeoutSeconds]: Timeout for running cargo fmt at the end of code generation
+ */
+data class ServerCodegenConfig(
+    val renameExceptions: Boolean = true,
```
david-perez, line 28: I think this should be removed from the server. If a user used "exception" in their Smithy model and they then built a Rust server with that model, they definitely want to use exception.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
```diff
@@ -0,0 +1,83 @@
...
+    val includeFluentClient: Boolean = false,
+    val addMessageToErrors: Boolean = true,
+    val formatTimeoutSeconds: Int = 20,
+    // TODO(EventStream): [CLEANUP] Remove this property when turning on Event Stream for all services
+    val eventStreamAllowList: Set<String> = emptySet(),
```
david-perez, line 33: What is this for?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
```diff
@@ -0,0 +1,83 @@
...
+) {
+    companion object {
+        fun fromNode(node: Optional<ObjectNode>): CodegenConfig {
```
david-perez, line 36: I find it very weird that a method called `fromNode()` on an object does not return itself. Same goes for `from()` on `ServerRustSettings`.

Can we make our settings "inherit" from the client ones and use our settings in the server codegen everywhere? I suspect that in the future we will want to add settings for us that won't be in the client.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -328,8 +328,7 @@ class ServerProtocolTestGenerator(
     private fun checkHttpExtensions(rustWriter: RustWriter) {
         rustWriter.rust(
             """
-            let extensions = http_request.extensions().expect("unable to extract http request extensions");
-            let request_extensions = extensions.get::<aws_smithy_http_server::RequestExtensions>().expect("extension RequestExtensions not found");
+            let request_extensions = http_response.extensions().get::<aws_smithy_http_server::RequestExtensions>().expect("extension RequestExtensions not found");
```
david-perez, line 331:
```suggestion
            let request_extensions = http_response.extensions().get::<aws_smithy_http_server::RequestExtensions>().expect("extension `RequestExtensions` not found");
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
david-perez: If not, I think a better design is to have server settings be projected into the client ones where needed in the codebase.

crisidev: Initially I wanted to go for the inheritance as you said. Since we are not changing anything in the client I went for this middle ground approach. Let's hear the @awslabs/rust-sdk-owners thoughts on this and I can refactor.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
crisidev: Let's see what the rest of the folks say about inheritance and I will refactor.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
crisidev: Let's see what the rest of the folks say about inheritance and I will refactor.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
crisidev: No matter what we still need to fill all the values inside CodegenConfig since we use them in the generation as we share part of the code with the client codegen.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
rcoh: agreed, this should also probably be removed from client

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt
rcoh: there are only a subset of event stream operations currently supported in client codegen

rcoh (approved):

Issue #938, opened 2021-12-07
[Server] Reimplement httpLabel trait deserializer without regexes


crisidev:
Currently we use regexes to deserialize [httpLabel](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httplabel-trait) traits: https://github.com/awslabs/smithy-rs/blob/6e423810ab8ed6db32a306df5f2189a7d72f668c/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt#L572

We should use the same parsing mechanism we will use for #937.

PR #996, opened 2021-12-20
Implement httpLabel with nom

82marbag:
Move to `nom` to implement httpLabel instead of using regexes.

Issue: #938

Signed-off-by: Daniele Ahmed <ahmeddan@amazon.com>

## Description

See issue: https://github.com/awslabs/smithy-rs/issues/938

## Testing
I run: `gradlew :codegen-server-test:assemble`  and checked the generated output.
Example output:
```
pub async fn parse_constant_query_string_request<B>(
    #[allow(unused_variables)] request: &mut axum_core::extract::RequestParts<B>,
) -> std::result::Result<
    crate::input::ConstantQueryStringInput,
>
where
    B: aws_smithy_http_server::HttpBody + Send,
    B::Data: Send,
    B::Error: Into<aws_smithy_http_server::BoxError>,
    aws_smithy_http_server::rejection::SmithyRejection:
        From<<B as aws_smithy_http_server::HttpBody>::Error>,
{
    Ok({
        #[allow(unused_mut)]
        let mut input = crate::input::constant_query_string_input::Builder::default();
        let input_string = request.uri().path();
        let (input_string, (_tag, _m)) = nom::sequence::tuple::<&str, _, (_, _), _>((
            nom::bytes::complete::tag("/"),
            nom::bytes::complete::tag("ConstantQueryString"),
        ))(input_string)
        .unwrap();
        let (_input_string, (_tag, m)) = nom::sequence::tuple::<&str, _, (_, _), _>((
            nom::bytes::complete::tag("/"),
            nom::branch::alt((
                nom::bytes::complete::take_until1("/"),
                nom::combinator::rest,
            )),
        ))(input_string)
        .unwrap();
        input = input
            .set_hello(crate::operation_deser::parse_str_constant_query_string_input_hello(m)?);
        input.build()?
    })
}
```

## Checklist


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt
```diff
@@ -200,6 +200,7 @@ data class CargoDependency(
         val HyperWithStream: CargoDependency = Hyper.withFeature("stream")
         val LazyStatic: CargoDependency = CargoDependency("lazy_static", CratesIo("1.4"))
         val Md5: CargoDependency = CargoDependency("md5", CratesIo("0.7"))
+        val Nom: CargoDependency = CargoDependency("nom", CratesIo("7"))
```
crisidev, line 203: This should go inside ServerCargoDependency.kt in the server-codegen folder as it is a dependency used only by the server right now.

crisidev:
Wow, nice, thanks a lot for the contribution!

crisidev:
If you haven't, you probably want to install `pre-commit.ci` as written in the readme and add a new commit. it will run a bunch of validations for you.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -577,42 +578,47 @@ private class ServerHttpProtocolImplGenerator(
...
+                val tag = "#{Nom}::bytes::complete::tag(\"/\")"
+                val nomParser =
+                    if (segment.isLabel && segment.isGreedyLabel) {
+                        "#{Nom}::combinator::rest"
+                    } else if (segment.isLabel) {
+                        "#{Nom}::branch::alt((" +
+                                "#{Nom}::bytes::complete::take_until1(\"/\")," +
+                                "#{Nom}::combinator::rest))"
+                    } else {
+                        "#{Nom}::bytes::complete::tag(\"${segment.content}\")"
```
rcoh, line 601:
```suggestion
                        "#{Nom}::bytes::complete::tag(${segment.content.dq()})"
```

rcoh: (a few other spots can be cleaned up with this)

rcoh: you can also use a raw string literal `"""` to avoid needing to escape quotes

crisidev:
The fact 2 of the CI steps are failing because the PR comes from a fork, so don't worry about it.

david-perez (commented):
Thanks for opening the PR, it looks promising.

I've never used `nom`, but my idea from reading the docs was that we would create a single parser that, given the URI path, will extract all the bound `@httpLabel` fields in one go with a single call. This PR is creating one parser per binding though, and feeding it and consuming the input little by little. We should be able to combine all these small parsers into one that represents the expected URI path for the Smithy operation.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -577,42 +578,47 @@ private class ServerHttpProtocolImplGenerator(
...
-        val pattern = StringBuilder()
         val httpTrait = httpBindingResolver.httpTrait(operationShape)
-        httpTrait.uri.segments.forEach {
-            pattern.append("/")
-            if (it.isLabel) {
-                pattern.append("(?P<${it.content}>")
-                if (it.isGreedyLabel) {
-                    pattern.append(".+")
-                } else {
-                    pattern.append("[^/]+")
```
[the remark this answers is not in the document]
david-perez: Writing my thoughts down here. I guess done this way because anecdotally servers accept empty path segments e.g. https://github.com//awslabs///////smithy-typescript///commits works fine.

But when extracting labels, I'm not sure about whether this behavior should be the correct one, and I don't see it specified in the Smithy spec. I'm thinking of possible ambiguous scenarios e.g. say the URI pattern is `/foo/{label}/bar` and the server receives a request with URI path `/foo//bar`. Should the `label` field be bound to an empty string? Or should we strip away the empty segment and reject the request?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -577,42 +578,47 @@ private class ServerHttpProtocolImplGenerator(
...
-            rustBlock("if let Some(captures) = RE.captures(request.uri().path())") {
-                pathBindings.forEach {
-                    val deserializer = generateParsePercentEncodedStrFn(it)
+            httpTrait.uri.segments.forEachIndexed { index, segment ->
+                val name = segment.content
+                val bindings = pathBindings.filter { it.memberName == name }
+                val tag = """#{Nom}::bytes::complete::tag("/")"""
+                val nomParser =
+                    if (segment.isLabel && segment.isGreedyLabel) {
+                        "#{Nom}::combinator::rest"
```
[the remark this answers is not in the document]
david-perez: The fact that greedy labels match greedily should also be taken into account. Consider the URI spec `/{greedy+}/foo/bar` and the request `/quux/foo/baz/foo/bar/foo/bar`. Here we need to capture `"quux/foo/baz/foo/bar"` in the label.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -577,42 +578,47 @@ private class ServerHttpProtocolImplGenerator(
...
-                        }
+                        let (${inputStringPrefix}input_string, (_tag, _m)) = #{Nom}::sequence::tuple::<&str, _, (_, _), _>(($tag, ${nomParser}))(input_string).unwrap();
+                        """.trimIndent(),
+                        *codegenScope,
+                    )
+                } else {
+                    val binding = bindings[0]
+                    val deserializer = generateParsePercentEncodedStrFn(binding)
+                    rustTemplate(
+                        """
```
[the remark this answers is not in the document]
82marbag: memberName is in camelCase, but Rust prefers (warns if not in) snake_case. If it's ok, I'd leave it like this here for simplicity. If you think it's best, I can add a function to convert to snake case

82marbag:
> Thanks for opening the PR, it looks promising.
> 
> I've never used `nom`, but my idea from reading the docs was that we would create a single parser that, given the URI path, will extract all the bound `@httpLabel` fields in one go with a single call. This PR is creating one parser per binding though, and feeding it and consuming the input little by little. We should be able to combine all these small parsers into one that represents the expected URI path for the Smithy operation.

If I understand correctly, I could have any number of labels to parse. With nom I can parse at most 21 as it is limited by the number of elements in a tuple. A workaround would be to parse with tuples in tuples, but I think it's easier to parse consume the input one at a time.

> Thanks for opening the PR, it looks promising.
> 
> I've never used `nom`, but my idea from reading the docs was that we would create a single parser that, given the URI path, will extract all the bound `@httpLabel` fields in one go with a single call. This PR is creating one parser per binding though, and feeding it and consuming the input little by little. We should be able to combine all these small parsers into one that represents the expected URI path for the Smithy operation.

I see. For what I know, nom isn't a good choice here; tuples for parsing in one go are restricted to 21 items. The alternative could be to chain `map`s and setting them, but it feels like multiple calls anyway. I can't find a better way with nom for this.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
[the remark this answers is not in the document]
rcoh: You may want symbolProvder.toMemberName

Note that you can't use memberName directly in code because it may need to be escaped

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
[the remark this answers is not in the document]
david-perez: My opinion on this and interpretation of the Smithy spec is that we were indeed doing it wrong and empty path segments **should not** be stripped away, but I've opened an issue with the Smithy team to clarify the spec: https://github.com/awslabs/smithy/issues/1024

82marbag:
I performed some benchmarks manually and with criterion:

* nom parsing one piece of input at a time: 2.1us
* nom parsing in a tuple until a greedy label is found: 1.8us
* regex (current approach): 410us

These are over between 20k and 2M iterations.

When a greedy label is found, both tests need to manually parse the string because nom does not support finding the last occurrence of a string (if the greedy label appears before another string and it's not at the end).

I believe these tests prove we should move to nom. I'm pushing a new revision; if `unwrap()` is not the right approach, tell me where I can leave the implementation for the error

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -572,42 +573,70 @@ private class ServerHttpProtocolImplGenerator(
...
+                    *codegenScope
+                )
+                if (segment.isGreedyLabel) {
+                    val end = httpTrait.uri.segments.slice((index + 1) until httpTrait.uri.segments.size).joinToString(prefix = "/", separator = "/")
+                    if (end.length > 1)
+                        rustTemplate(
+                            """
+                            let (${bindingPrefix}m, ${inputStringPrefix}input_string) = match input_string.rfind(${end.dq()}) {
+                                Some(n) => (&input_string[..n], &input_string[n..]),
+                                None => (input_string, "")
```
david-perez, line 603: * Here we're extracting to the first pair component, which is the reverse of nom's API that we use in the other places, so it's confusing.
* If the suffix is not found, we're assigning the field bound to the greedy label to the entire `input_string` when we should error out.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -572,42 +573,70 @@ private class ServerHttpProtocolImplGenerator(
...
+                        let (${inputStringPrefix}input_string, ${bindingPrefix}m) = #{Nom}::branch::alt::<_, &str, #{Nom}::error::Error<&str>, _>((
+                            #{Nom}::bytes::complete::take_until1("/"),
+                            #{Nom}::combinator::rest))(input_string).unwrap();
+                        """.trimIndent(),
+                        *codegenScope
+                    )
+                } else {
+                    rustTemplate(
+                        """
+                        let (input_string, ${bindingPrefix}m) = #{Nom}::bytes::complete::tag::<_, &str, #{Nom}::error::Error<&str>>(${segment.content.dq()})(input_string).unwrap();
```
david-perez, line 626:
```suggestion
                        let (${inputStringPrefix}input_string, ${bindingPrefix}m) = #{Nom}::bytes::complete::tag::<_, &str, #{Nom}::error::Error<&str>>(${segment.content.dq()})(input_string).unwrap();
```

I think clippy is failing in CI because of this.

david-perez (commented):
Are you going to try out what I suggested about building a single big parser for the entire URI path and calling it once instead of having one parser per segment?

As of now because of how you implemented the greedy label case with `rfind` I don't think it'd be easy/possible to build one big parser, but a simpler implementation is perhaps as follows:

If the URI pattern has a greedy label, the first thing we do at the very beginning is check for the (possibly empty) suffix using [`str::ends_with`](https://doc.rust-lang.org/std/primitive.str.html#method.ends_with), if the suffix is not there we can error out early without doing any parsing. If the suffix is there, we can strip it away and work with the string slice `input_string[..input_string.len() - s]` thereafter, where `s` is the length of the suffix, which we know at codegen time. It then becomes easy to build a big parser, since the greedy label can be extracted with `::combinator::rest`.

david-perez:
Thanks for the benchmarking! Would you mind linking a pastebin to this PR for historical purposes? Thanks!

82marbag:
> Thanks for the benchmarking! Would you mind linking a pastebin to this PR for historical purposes? Thanks!

This is the benchmark test: https://pastebin.com/raw/v9r3qwqS

I've changed the code, the only two things remaining to see are:
* a max of 21 labels / strings can be parsed now (agreed it is fine here)
* with `ends_with`, some `/` at the end won't be fine

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -572,46 +573,77 @@ private class ServerHttpProtocolImplGenerator(
...
-                    val deserializer = generateParsePercentEncodedStrFn(it)
-                    rustTemplate(
-                        """
-                        if let Some(m) = captures.name("${it.locationName}") {
-                            input = input.${it.member.setterName()}(
-                                #{deserializer}(m.as_str())?
+            segments
+                .forEachIndexed { index, segment ->
+                    val binding = pathBindings.find { it.memberName == segment.content }
+                    if (binding != null && segment.isLabel) {
```
david-perez, line 634: If `segment.isLabel` is `true`, then `binding` must be non-null, right? Because `pathBindings` contains all the `@httpLabel` bindings.

rust-runtime/aws-smithy-http-server/src/rejection.rs
```diff
@@ -227,3 +227,9 @@ impl From<serde_urlencoded::de::Error> for SmithyRejection {
         SmithyRejection::Deserialize(Deserialize::from_err(err))
     }
 }
+
+impl From<nom::Err<nom::error::Error<&str>>> for SmithyRejection {
+    fn from(_err: nom::Err<nom::error::Error<&str>>) -> Self {
+        SmithyRejection::MissingQueryString(MissingQueryString)
```
david-perez, line 233: `MissingQueryString` is the wrong rejection type, since we're not looking in the query string, but the URI path. `SmithyRejection::Deserialize` should be fine for now.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -572,46 +573,77 @@ private class ServerHttpProtocolImplGenerator(
...
+                    if (segment.isGreedyLabel) {
+                        "#{Nom}::combinator::rest"
+                    } else if (segment.isLabel) {
+                        """#{Nom}::branch::alt((#{Nom}::bytes::complete::take_until1("/"), #{Nom}::combinator::rest))"""
+                    } else {
+                        """#{Nom}::bytes::complete::tag("${segment.content}")"""
+                    }
+                }
+                .joinToString(
+                    prefix = "#{Nom}::sequence::tuple::<_, _, #{Nom}::error::Error<&str>, _>((",
```
david-perez, line 605: CI is failing here https://github.com/awslabs/smithy-rs/runs/4692069662?check_suite_focus=true#step:7:1064

In the case where there is only one label in the URI pattern and nothing else, using `sequence::tuple` doesn't seem to compile.

david-perez (changes requested):
This looks much better; only minor things left now.

> with ends_with, some / at the end won't be fine

This should be correct, trailing slashes at the end do have meaning. So if the URI pattern ends with `/` requests must end with `/`; if it doesn't, requests must not end with `/`. This might seem very strict but if the URI pattern is `/foo/{label}` then the request `/foo/` is binding `""` to `label`. Same goes for empty path segments in the middle of the URI, which we were previously accepting in the regex approach, but with this PR we will be strict and interpret them. See this issue https://github.com/awslabs/smithy/issues/1024.

> a max of 21 labels / strings can be parsed now (agreed it is fine here)

This is fine IMO, I don't foresee users will need so many labels. Leave a comment somewhere in the source code documenting this limitation, lest we forget or decide to tackle it in the future.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
82marbag: In some cases, such as `/foo/{foo}`, `foo` is found as binding too because I look for the name

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -572,46 +573,77 @@ private class ServerHttpProtocolImplGenerator(
...
+                else
+                    httpTrait.uri.segments
+            val restAfterGreedyLabel =
+                if (greedyLabelIndex >= 0)
+                    httpTrait.uri.segments.slice((greedyLabelIndex + 1) until httpTrait.uri.segments.size).joinToString(prefix = "/", separator = "/")
+                else
+                    ""
+            val labeledNames = segments
+                .mapIndexed { index, segment ->
+                    if (segment.isLabel) { "m$index" } else { "_" }
```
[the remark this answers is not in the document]
82marbag: `safeName(prefix="var")` basically returns `var_$i` with `i` always increasing. It'd still have to leave `_` at the beginning for those labels/constants we don't use and keep the two problems on the naming (like camel case) and on the possible characters in the variable names if we decide to use any other prefix. It seems to me like it won't make it necessarily simpler, but I can add that too.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -572,46 +573,77 @@ private class ServerHttpProtocolImplGenerator(
...
+                    }
+                )
+            rustTemplate("let input_string = request.uri().path();")
+            if (restAfterGreedyLabel.isNotEmpty()) {
+                rustTemplate(
+                    """
+                    if ! input_string.ends_with(${restAfterGreedyLabel.dq()}) {
+                        return std::result::Result::Err(#{SmithyRejection}::MissingQueryString(aws_smithy_http_server::rejection::MissingQueryString));
+                    }
+                    let input_string = input_string[..(input_string.len() - ${restAfterGreedyLabel.dq()}.len())].into();
```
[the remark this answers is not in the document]
82marbag: It's a str without a size known at compile time https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dfafab556132a719eb28b50ab42cf70e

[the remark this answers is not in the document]
david-perez: `input_string` is a `&str`, `input_string[..7]` is a `str`, and the question was, why the `input_string[..7].into()`? What is it converting into? By not annotating the type of the new binding for `input_string`, it's relying on the compiler's type inference to pick an `Into` impl that makes the program compile. And it's hard to know what it's inferring from reading the code, since the next time `input_string` is used is as the input to the result of `nom::sequence::tuple`, which is very generic. Surely it must be a `&str` though, but then the call to `into()` is a bit misleading since the conversion from `str` to `&str` is very straightforward and best conveyed by simply creating a reference with `&`:

```
let input_string = &input_string[..(input_string.len() - ${restAfterGreedyLabel.dq()}.len())];
```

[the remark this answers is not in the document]
82marbag: I see. Yes, `&input_string[..]` also works. Updated, thanks!

82marbag:
I saw the linked PR https://github.com/awslabs/smithy-rs/pull/1029 and I have changed the code here accordingly. Empty labels are fine now, such as `/{foo}/str` for `//str`

david-perez (approved):
Thank you!

PR #998, opened 2021-12-21
[fix] Sleep not passed to ECS provider

rcoh:
## Motivation and Context
When constructing an ECS credential provider, sleep_impl was not properly passed along from the provider config.

- https://github.com/awslabs/aws-sdk-rust/issues/359

## Description
This adds a test that asserts that retries will be made, and fixes the bug.

## Testing
- [x] New UT
- [x] audited other client construction to ensure a sleep implementation was passed
 
## Checklist


aws/rust-runtime/aws-config/src/ecs.rs
```diff
@@ -451,6 +451,7 @@ mod test {
     use aws_types::os_shim_internal::Env;
     use aws_types::Credentials;
 
+    use aws_smithy_async::rt::sleep::TokioSleep;
```
Velfi, line 454: this sleep impl is feature gated right? Does this test work if it's not being compiled with the `rt-tokio` feature?

Velfi (approved):

aws/rust-runtime/aws-config/src/ecs.rs
rcoh: dev dependencies pull in rt-tokio

PR #1000, opened 2021-12-21
`rust-server-codegen`: don't return 200 status code with rejections

david-perez:
Rejections themselves implement `IntoResponse` and set their status
code on it.


rust-runtime/aws-smithy-http-server/src/macros.rs
```diff
@@ -46,12 +46,13 @@ macro_rules! define_rejection {
         #[derive(Debug)]
         pub struct $name;
 
-        #[allow(deprecated)]
```
david-perez, line 49: Don't know why this is in axum's codebase.

crisidev (approved):

PR #1004, opened 2021-12-22
add: tool to sync aws-sdk-rust/next with smithy-rs

Velfi:
## Motivation and Context

smithy-rs#928

## Description

Add a tool that will grab commits from smithy-rs/main, and replay those commits to the aws-sdk-rust/next branch. This will give us a better commit history when reviewing the SDK and will make releases quicker too. The GitHub workflow that uses this tool will be defined in `aws-sdk-rust` (it's easier to pull than to push).

## Testing

I ran this locally and verified the output


tools/smithy-rs-sync/Cargo.lock
```diff
@@ -0,0 +1,366 @@
+# This file is automatically @generated by Cargo.
```
crisidev, line 1: Is it expected to commit Cargo.lock in vcs?

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+use std::fs::OpenOptions;
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::time::Instant;
+use structopt::StructOpt;
+
+#[derive(StructOpt, Debug)]
+#[structopt(name = "smithy-rs-sync")]
+/// A CLI tool to replay commits from smithy-rs, generate code, and commit that code to aws-rust-sdk
```
crisidev, line 17: Nit: most comments / doc are missing the carriage return.

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+    #[structopt(long, default_value = "next")]
+    branch: String,
+}
+
+const BOT_NAME: &str = "AWS SDK Rust Bot";
+const BOT_EMAIL: &str = "aws-sdk-rust-primary@amazon.com";
+const COMMIT_HASH_FILENAME: &str = ".smithyrs-githash";
+
+/// A macro for attaching info to error messages pointing to the line of code responsible for the error
+/// [Thanks to dtolnay for this macro](https://github.com/dtolnay/anyhow/issues/22#issuecomment-542309452)
```
crisidev, line 35: Uh, this is neat.

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+
+/// Read the file from aws-sdk-rust that tracks the last smithy-rs commit it was synced with.
+/// Returns the hash of that commit.
+fn get_last_synced_commit(repo_path: &Path) -> Result<Oid> {
+    let path = repo_path.join(COMMIT_HASH_FILENAME);
+
+    let mut file = OpenOptions::new()
+        .read(true)
+        .open(&path)
+        .with_context(|| format!("couldn't open file at '{}'", path.to_string_lossy()))?;
```
crisidev, line 177: I am not sure I understand why are you using `with_context` and a closure vs just `context` and the format string..

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+    let remove_sdk_command_output = Command::new("rm")
+        .arg("-rf")
+        .arg(&sdk_path)
+        .current_dir(aws_sdk_path)
+        .output()?;
+    if !remove_sdk_command_output.status.success() {
+        bail!("failed to clean out the SDK folder at {}", sdk_path);
+    }
+
+    let examples_path = format!("{}/example/*", aws_sdk_path.to_string_lossy());
```
crisidev, line 273: I always advocate about using `Path` instead of raw strings form paths.

crisidev (approved):

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+/// Run this app in order to keep aws-sdk-rust in sync with smithy-rs.
+///
+/// pre-requisites:
+/// - an up-to-date local copy of smithy-rs repo
+/// - an up-to-date local copy of aws-sdk-rs repo
+/// - a Unix-ey system (for the `cp` and `rf` commands to work)
+/// - Java Runtime Environment v11 (in order to run gradle commands)
+///
+/// ```sh
+/// cargo run -- \
```
david-perez, line 51:
```suggestion
/// cargo run \
```

tools/smithy-rs-sync/src/main.rs
david-perez: It's a neat pattern; if you don't use a closure, the formatted string will always be allocated independently of whether the error path is taken or not.

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+    println!("\tsuccessfully built the SDK in {:?}", start.elapsed());
+    Ok(build_artifact_path)
+}
+
+/// Delete any current SDK files in aws-sdk-rust. Run this before copying over new files.
+fn clean_out_existing_sdk(aws_sdk_path: &Path) -> Result<()> {
+    println!("\tcleaning out previously built SDK...");
+
+    let sdk_path = format!("{}/sdk/*", aws_sdk_path.to_string_lossy());
+    let remove_sdk_command_output = Command::new("rm")
```
david-perez, line 264: It's a bit verbose to do it with stdlib, but instead of shelling out to copy and delete recursively, this crate is nice https://github.com/webdesus/fs_extra

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+    let mut index = aws_sdk_repo.index().context(here!())?;
+    // The equivalent of `git add .`
+    index
+        .add_all(["."].iter(), IndexAddOption::DEFAULT, None)
+        .context(here!())?;
+    let oid = index.write_tree().context(here!())?;
+    let parent_commit = find_last_commit(&aws_sdk_repo).context(here!())?;
+    let tree = aws_sdk_repo.find_tree(oid).context(here!())?;
+
+    let _ = aws_sdk_repo
```
david-perez, line 339:
```suggestion
    aws_sdk_repo
```

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+    println!("Updating 'commit at last sync' to {}", last_synced_commit);
+
+    // Update the file containing the commit hash
+    set_last_synced_commit(&aws_sdk_repo, last_synced_commit).with_context(|| {
+        format!(
+            "couldn't write last synced commit hash ({}) to aws-sdk-rust/{}",
+            last_synced_commit, COMMIT_HASH_FILENAME,
+        )
+    })?;
+    // Commit the file containg the commit hash
```
david-perez, line 134:
```suggestion
    // Commit the file containing the commit hash
```

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+/// this tool was run.
+fn commits_to_be_applied(smithy_rs_repo: &Repository, since_commit: &Oid) -> Result<Vec<Oid>> {
+    let rev_range = format!("{}..HEAD", since_commit);
+    println!("Checking for smithy-rs commits in range {}", rev_range);
+
+    let mut rev_walk = smithy_rs_repo.revwalk().context(here!())?;
+    rev_walk.push_range(&rev_range).context(here!())?;
+
+    let mut commit_revs = Vec::new();
+    for rev in rev_walk {
```
david-perez, line 158: `rev_walk.map(...).collect()` maybe?

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+        .context(here!())?;
+    let oid = index.write_tree().context(here!())?;
+    let parent_commit = find_last_commit(&aws_sdk_repo).context(here!())?;
+    let tree = aws_sdk_repo.find_tree(oid).context(here!())?;
+
+    let _ = aws_sdk_repo
+        .commit(
+            Some("HEAD"),
+            &based_on_commit.author(),
+            &based_on_commit.committer(),
```
david-perez, line 343: nit: in Git the committer is the one who applies the commit, so this really should be the bot's name and email. Author is used to retain original attribution.

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+    let mut index = aws_sdk_repo.index().context(here!())?;
+    index
+        .add_path(Path::new(COMMIT_HASH_FILENAME))
+        .context(here!())?;
+    let signature = Signature::now(BOT_NAME, BOT_EMAIL)?;
+    let oid = index.write_tree().context(here!())?;
+    let parent_commit = find_last_commit(&aws_sdk_repo).context(here!())?;
+    let tree = aws_sdk_repo.find_tree(oid).context(here!())?;
+
+    let _ = aws_sdk_repo
```
david-perez, line 366:
```suggestion
    aws_sdk_repo
```

david-perez (commented):
I love that this repo is slowly gathering a suite of well-rounded nifty little Rust programs instead of the usual mishmash of shell script to do this kind of tasks. I'm for sure going to pilfer this one for my projects!

tools/smithy-rs-sync/Cargo.lock
david-perez: Yup https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

crisidev: Ah, TIL.

tools/smithy-rs-sync/src/main.rs
rcoh: context only takes a static string I think

rcoh: context only takes a static string I think

tools/smithy-rs-sync/src/main.rs
rcoh: so this is something we should discuss: I thought the clearest thing would be to have the effective commits attributed to the person who made the codegen change.

But I could definitely be convinced we should be doing something different.

tools/smithy-rs-sync/src/main.rs
rcoh: you need the `--` so that the arguments go to the binary and not to cargo

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+        "Found {} unsynced commit(s), syncing now...",
+        number_of_commits
+    );
+    // Run through all the new commits, syncing them one by one
+    for (i, rev) in commit_revs.iter().enumerate() {
+        let commit = smithy_rs_repo
+            .find_commit(*rev)
+            .with_context(|| format!("couldn't find commit {} in smithy-rs", rev))?;
+
+        println!("[{}/{}]\tsyncing {}...", i + 1, number_of_commits, rev);
```
rcoh, line 104: I'd probably use `eprintln` or tracing everywhere in case we decide to use stdout for structured output

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+            "couldn't write last synced commit hash ({}) to aws-sdk-rust/{}",
+            last_synced_commit, COMMIT_HASH_FILENAME,
+        )
+    })?;
+    // Commit the file containg the commit hash
+    commit_last_synced_commit_file(&aws_sdk_repo)
+        .context("couldn't commit the last synced commit hash file to aws-sdk-rust")?;
+
+    println!(
+        "Successfully synced all mirror commits to aws-sdk-rust/{}. Don't forget to push them",
```
rcoh, line 139: might be nice to print the number of commits here.

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+    let path = repo_path.join(COMMIT_HASH_FILENAME);
+
+    let mut file = OpenOptions::new()
+        .read(true)
+        .open(&path)
+        .with_context(|| format!("couldn't open file at '{}'", path.to_string_lossy()))?;
+    // Commit hashes are 40 chars long
+    let mut commit_hash = String::with_capacity(40);
+    file.read_to_string(&mut commit_hash)
+        .with_context(|| format!("couldn't read file at '{}'", path.to_string_lossy()))?;
```
rcoh, line 181: is there a reason you didn't use `std::file::read_to_string(path)`?

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,409 @@
...
+        .arg("-Paws.fullsdk=true")
+        .arg(":aws:sdk:clean")
+        .current_dir(smithy_rs_path)
+        .output()
+        .with_context(|| {
+            format!(
+                "failed to execute '{} -Paws.fullsdk=true :aws:sdk:clean'",
+                gradlew.to_string_lossy()
+            )
+        })?;
```
rcoh, line 219: probably worth adding something like:

`fn run(args: &[str]) -> ...` that runs the command and handles the error wrapping

tools/smithy-rs-sync/src/main.rs
rcoh, line 273: I wonder if we should have an allow-list of protected non-synced files. Seems like this could be the home of a future bug if we added another generated directory

rcoh (approved):
This is awesome!!!

tools/smithy-rs-sync/Cargo.lock
Velfi: In my opinion, yes because it's a binary crate. If it was a library I wouldn't have committed it. I'm cool removing it if y'all want.

Velfi: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

tools/smithy-rs-sync/src/main.rs
Velfi: because I don't want to format and allocate unless this fails

tools/smithy-rs-sync/src/main.rs
Velfi: The reason I didn't do that here is because of the wildcard. I wasn't sure how this would interact with Rust's `Path` because afaik it's not actually a path but a 'glob'. I'm open to suggestions tho

tools/smithy-rs-sync/src/main.rs
Velfi: I need that there so that cargo doesn't eat up the args. The `--` makes them get interpreted by the program being run instead.

tools/smithy-rs-sync/src/main.rs
Velfi: The `commit` method being called on the next line returns an `Oid`. I'm assigning to `_` to be explicit about throwing away that value.

tools/smithy-rs-sync/src/main.rs
Velfi: I'm using a for loop so I can early return if there's an error when reading a rev

tools/smithy-rs-sync/src/main.rs
Velfi: The commit method being called on the next line returns an Oid. I'm assigning to _ to be explicit about throwing away that value.

tools/smithy-rs-sync/src/main.rs
Velfi: Not really, just a habit. Pedantically, I think this format is faster because we know the length of the string and can allocate it all at once.

tools/smithy-rs-sync/src/main.rs
Velfi: This is a good callout, let me think about how to implement that.

tools/smithy-rs-sync/src/main.rs
Velfi: This crate is basically a Bash script, I'm going to leave it as is for simplicity's sake

tools/smithy-rs-sync/src/main.rs
Velfi: Do you mean that I should have split this comment into 2 lines?

tools/smithy-rs-sync/src/main.rs
Velfi: My intuition was incorrect. Under the hood, `std::fs::read_to_string` does smartly allocate

tools/smithy-rs-sync/src/main.rs
david-perez: The author field is precisely to retain authorship attribution, while the committer is the one who applies the commit. At least those are the semantics that Git ascribes: https://stackoverflow.com/questions/6755824/what-is-the-difference-between-author-and-committer-in-git Having the committer be the bot is ideal for this use case, since it lets you know that the commit was done by this program.

tools/smithy-rs-sync/src/main.rs
david-perez: An iterator over `Result<T, E>` can be `collect()`-ed directly into a `Result<Vec<T>, E>`, on which you can then use `?`, and it will return early upon encountering the first error.

tools/smithy-rs-sync/src/main.rs
crisidev: Ah, fair enough, you want globbing here. Disregard my comment.

tools/smithy-rs-sync/src/main.rs
crisidev: Nope, they miss the "." at the end of the line. Again, nitpicking here, not a blocker.

tools/smithy-rs-sync/src/main.rs
Velfi: @rcoh I implemented it, lmk what you think

tools/smithy-rs-sync/src/fs.rs
```diff
@@ -0,0 +1,308 @@
...
+use super::here;
+use anyhow::Context;
+use std::error::Error;
+use std::fmt::Display;
+use std::path::{Path, PathBuf};
+
+pub static HANDWRITTEN_DOTFILE: &str = ".handwritten";
+
+pub fn delete_all_generated_files_and_folders(directory: &Path) -> anyhow::Result<()> {
+    fn callback(path: &Path) -> Result<(), std::io::Error> {
```
rcoh, line 15: could rename these to be more descriptive, eg. `rm_path`

tools/smithy-rs-sync/src/fs.rs
```diff
@@ -0,0 +1,308 @@
...
+    ) -> Result<(), HandwrittenFilesError> {
+        let read_dir = std::fs::read_dir(directory)?;
+
+        for entry in read_dir {
+            let path = &entry.map(|e| e.path())?;
+
+            if self.is_handwritten(path)? {
+                callback(path)?;
+            }
+        }
```
rcoh, line 125: this is fine. One option could be to restructure it a bit. Instead of having this use callbacks, you could use `impl Iterator` instead, something like `read_dir.filter(|...|)`. Then you could have `handwritten_files(&self) -> impl Iterator<Item=PathBuf>` which might be a little cleaner to use elsewhere

tools/smithy-rs-sync/src/main.rs
```diff
@@ -0,0 +1,439 @@
...
+        let build_artifacts = build_sdk(smithy_rs).context("couldn't build SDK")?;
+        clean_out_existing_sdk(aws_sdk)
+            .context("couldn't clean out existing SDK from aws-sdk-rust")?;
+
+        // Check that we aren't generating any files that we've marked as "handwritten"
+        let handwritten_files_in_generated_sdk_folder =
+            find_handwritten_files_and_folders(aws_sdk, &build_artifacts)?;
+        if !handwritten_files_in_generated_sdk_folder.is_empty() {
+            bail!(
+                "found one or more 'handwritten' files/folders in generated code: {:#?}",
```
rcoh, line 128: "hint: if this file is newly generated, remove it from `.handwritten`"

rcoh (approved):
LGTM, could consider a refactoring to produce iterators instead of callbacks but it's not critical

tools/smithy-rs-sync/src/fs.rs
Velfi: The thing that makes this difficult is the `Result`s that need to be propagated. I refactored this to use the scan iterator instead but it's kinda gross and still requires a `collect` because of the lifetime of `gitignore::File` prevents me from easily passing an iterator with an internal ref to `HandwrittenFiles`. I'm sure it's possible, I just didn't want to get too sucked in.

tools/smithy-rs-sync/src/fs.rs
Velfi: I deleted the callback style functions so this is a moot point

tools/smithy-rs-sync/src/main.rs
Velfi: I have since added these

tools/smithy-rs-sync/src/main.rs
Velfi: I have updated this to mark the bot as the committer

rcoh (approved):

PR #1006, opened 2021-12-23
Add support for Paginators

rcoh:
## Motivation and Context
- https://github.com/awslabs/aws-sdk-rust/issues/47

## Description
This commit adds pagination support for clients. It does this via the `FnStream` abstraction which combines a rendezvous-channel with a closure to produce a stream. When the stream is polled, the underlying closure is polled to advance the computation which then pushes data into the channel.

Two paginators are generated:
1. `.paginate()` which produces a stream of `Result<Page, Err>`
2. `.paginate().items()` which produces a stream of `Result<Item, Err>` where items are flattened from the individual pages. This uses `TryFlatMap` to support conveniently generating a flat page iterator.

In addition, docs are generated to point customers towards the paginators.
## Testing
- [x] manual integration testing
- [x] integration tests using mocked traffic
- [x] assorted UTs

## Checklist


rust-runtime/aws-smithy-async/src/future/fn_stream.rs
```diff
@@ -0,0 +1,265 @@
...
+use crate::future::rendezvous;
+use futures_util::StreamExt;
+use pin_project_lite::pin_project;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use tokio_stream::{Iter, Once, Stream};
+
+pin_project! {
+    /// Utility to drive a string with an async function and a channel.
```
Velfi, line 14:
```suggestion
    /// Utility to drive a stream with an async function and a channel.
```

rust-runtime/aws-smithy-async/src/future/fn_stream.rs
```diff
@@ -0,0 +1,265 @@
...
+    ///
+    /// The closure is passed a reference to a `Sender` which acts as a rendezvous channel. Messages
+    /// sent the sender will be emitted to the stream. Because the stream is 1-bounded, the function
```
Velfi, line 17:
```suggestion
    /// sent to the sender will be emitted to the stream. Because the stream is 1-bounded, the function
```

aws/sdk/build.gradle.kts
```diff
@@ -266,8 +268,8 @@ task("finalizeSdk") {
         "relocateRuntime",
         "relocateAwsRuntime",
         "relocateExamples",
-        "generateDocs",
-        "fixManifests"
+        "generateDocs" // ,
+        // "fixManifests"
```
Velfi, line 272: It produces a good amount of noise doesn't it? I noticed that too when working on the `impl Into` thing

aws/sdk/examples/dynamodb/src/bin/list-items.rs
```diff
@@ -58,11 +59,18 @@ async fn main() -> Result<(), Error> {
         println!();
     }
 
-    let resp = client.scan().table_name(table).send().await?;
+    let items: Result<Vec<_>, _> = client
+        .scan()
+        .table_name(table)
+        .paginate()
+        .items()
```
Velfi, line 66: In cases where a user just wants all the data and doesn't care about pages, do they still need to paginate?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+
+    /** Type of the inner item of the paginator */
+    private fun itemType(): String {
+        val members = paginationInfo.itemsMemberPath
+        val type = symbolProvider.toSymbol(model.expectShape(members.last().target)).rustType()
+        check(type is RustType.Vec || type is RustType.HashMap)
+        return when (type) {
+            is RustType.Vec -> type.member.render(true)
+            is RustType.HashMap -> "(${type.key.render(true)}, ${type.member.render(true)})"
+            else -> PANIC("only HashMaps or Vecs may be used item pagination.")
```
Velfi, line 208:
```suggestion
            else -> PANIC("only HashMaps or Vecs may be used for item pagination.")
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    /** Generate a `.items()` function to expose flattened pagination when modeled */
```
Velfi, line 212:
```suggestion
    /** Generate an `.items()` function to expose flattened pagination when modeled */
```

design/src/rfcs/rfc0008_paginators.md
```diff
@@ -0,0 +1,191 @@
...
+. Customers of Smithy generated code & the Rust SDK will have an improved user experience if code is generated to
+support this.
+
+In this RFC, we propose modeling paginated data as
+a  [`Stream`](https://docs.rs/tokio-stream/0.1.5/tokio_stream/#traits) of output shapes.
+
+- When an output is paginated, a `paginate()` method will be added to the high level builder
+- An `<OperationName>Paginator` struct will be generated into the `paginator` module.
+- If `items` is modeled, `paginate().items()` will be added to produce the paginated
+  items. `<OperationName>PaginatorItems` will be generated into the `paginator` module.
```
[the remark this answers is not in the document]
rcoh: some operations only model how to get the next page—other operations both how to get the page but also a nested field that contains the actual items. `items()` isn't generated if it isn't modeled (so it just won't exist)

design/src/rfcs/rfc0008_paginators.md
```diff
@@ -0,0 +1,191 @@
...
+    // holds the low-level client and configuration
+    handle: Arc<Handle<C, M, R>>,
+
+    // input builder to construct the actual input on demand
+    input: ListTablesInputBuilder
+}
+```
+
+In addition to the basic usage example above, when `pageSize` is modeled, customers can specify the page size during
+pagination:
```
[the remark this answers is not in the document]
rcoh: yeah it's all about what's modeled via the pagination trait

design/src/rfcs/rfc0008_paginators.md
```diff
@@ -0,0 +1,191 @@
...
+  ///
+  /// This paginator automatically flattens results using `table_names`. Queries to the underlying service
+  /// are dispatched lazily.
+  pub fn items(self) -> crate::paginator::ListTablesPaginatorItems<C, M, R> {
+    crate::paginator::ListTablesPaginatorItems(self)
+  }
+
+  /// Create the pagination stream
+  ///
+  /// _Note:_ No requests will be dispatched until the stream is used (eg. with [`.next()`](tokio_stream::StreamExt::next)).
```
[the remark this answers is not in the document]
rcoh: anything that actually polls the stream will dispatch a request, eg. `next().await`, `try_next()`, `collect()`, etc.

aws/sdk/examples/dynamodb/src/bin/list-items.rs
rcoh: I don't follow, if you want all the data and don't care about pages, you could use `collect()` (which will drive the paginator to completion but you won't have to interact with the pages)

aws/sdk/build.gradle.kts
rcoh: it also makes stuff really slow. I need to back this change out but we should find a way to avoid running fixManifests locally

design/src/rfcs/rfc0008_paginators.md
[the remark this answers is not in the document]
rcoh: but yeah you're correct that `await` is the thing that actually makes something happen, will update

design/src/rfcs/rfc0008_paginators.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+**On Box::pin**: The stream returned by `AsyncStream` does not implement `Unpin`. Unfortunately, this makes iteration
+require an invocation of `pin_mut!` and generates several hundred lines of compiler errors. Box::pin seems a worthwhile
+trade off to improve the user experience.
+
+**On the `+ Unpin` bound**: Because auto-traits leak across `impl Trait` boundaries, `+ Unpin` prevents accidental
+regressions in the generated code which would break users.
+
+**On the crate::reflens::...**: We use `LensGenerator.kt` to generate potentially complex accessors to deeply nested fields.
```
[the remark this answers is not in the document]
rcoh: yeah I guess we could call them `NestedAccessor` or something. Admittedly, these aren't true lenses since lenses also support `set` usually

aws/sdk/examples/dynamodb/src/bin/list-items.rs
Velfi: What I meant is that I see two use cases:
- I want to look through the data page by page and potentially load less total data
- I just want it all

In the second case, do they need to create the paginated operation at all or can they just send the operation they have?

rcoh: if you send the operation you have, you only get the first page

Velfi: Ah! I understand now

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+        val inputTokenMember = symbolProvider.toMemberName(paginationInfo.inputTokenMember)
+        rustTemplate(
+            """
+            /// Paginator for #{operation:D}
+            pub struct $paginatorName#{generics:W} {
+                handle: std::sync::Arc<crate::client::Handle${generics.inst}>,
+                builder: #{Builder}
+            }
+
+            impl ${generics.inst}${paginatorName}${generics.inst} where #{bounds:W} {
```
jdisanti, line 128:
```suggestion
            impl${generics.inst} ${paginatorName}${generics.inst} where #{bounds:W} {
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+                /// Create a new paginator-wrapper
+                pub(crate) fn new(handle: std::sync::Arc<crate::client::Handle${generics.inst}>, builder: #{Builder}) -> Self {
+                    Self {
+                        handle,
+                        builder,
+                    }
+                }
+
+                #{page_size_setter:W}
```
jdisanti, line 137: This new `W` formatter is nice!

Should there also be a getter for page size?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+                        #{Input}OperationRetryAlias
+                    >, {
+                    // Move individual fields out of self for the borrow checker
+                    let builder = self.builder;
+                    let handle = self.handle;
+                    #{fn_stream}::FnStream::new(move |tx| Box::pin(async move {
+                        // Build the input for the first time. If required fields are missing, this is where we'll produce an early error.
+                        let mut input = match builder.build().map_err(|err|#{SdkError}::ConstructionFailure(err.into())) {
+                            Ok(input) => input,
+                            Err(e) =>  { let _ = tx.send(Err(e)).await; return; }
```
jdisanti, line 160:
```suggestion
                        let mut input = match builder.build().map_err(|err| #{SdkError}::ConstructionFailure(err.into())) {
                            Ok(input) => input,
                            Err(e) => { let _ = tx.send(Err(e)).await; return; }
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+                        };
+                        loop {
+                            let op = match input.make_operation(&handle.conf)
+                                .await
+                                .map_err(|err|#{SdkError}::ConstructionFailure(err.into())) {
```
jdisanti, line 165:
```suggestion
                                .map_err(|err| #{SdkError}::ConstructionFailure(err.into())) {
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+                """
+                /// Create a flattened paginator
+                ///
+                /// This paginator automatically flattens results using `$documentedPath`. Queries to the underlying service
+                /// are dispatched lazily.
+                pub fn items(
+                        self
+                    ) -> #{ItemPaginator}${generics.inst} {
+                        #{ItemPaginator}(self)
+                    }
```
jdisanti, line 227:
```suggestion
                pub fn items(self) -> #{ItemPaginator}${generics.inst} {
                    #{ItemPaginator}(self)
                }
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /** Generate a struct with a `items()` method that flattens the paginator **/
+    private fun itemsPaginator(): RuntimeType? = if (paginationInfo.itemsMemberPath.isEmpty()) {
+        null
+    } else {
+        RuntimeType.forInlineFun("${paginatorName}Items", module) {
+            it.rustTemplate(
+                """
+                /// Flattened paginator for for $paginatorName
```
jdisanti, line 241:
```suggestion
                /// Flattened paginator for $paginatorName
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -0,0 +1,289 @@
...
+                    /// To read the entirety of the paginator, use [`.collect::<Result<Vec<_>, _>()`](tokio_stream::StreamExt::collect).
+                    pub fn send(self) -> impl #{Stream}<Item = std::result::Result<${itemType()}, #{SdkError}<#{Error}>>> + Unpin
+                    where
+                        R::Policy: #{client}::bounds::SmithyRetryPolicy<
+                            #{Input}OperationOutputAlias,
+                            #{Output},
+                            #{Error},
+                            #{Input}OperationRetryAlias
+                        >, {
+                        #{fn_stream}::TryFlatMap::new(self.0.send()).flat_map(|page|#{extract_items}(page).unwrap_or_default().into_iter())
```
jdisanti, line 260:
```suggestion
                        #{fn_stream}::TryFlatMap::new(self.0.send()).flat_map(|page| #{extract_items}(page).unwrap_or_default().into_iter())
```

rust-runtime/aws-smithy-async/src/future/rendezvous.rs
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+//! Rendezvous channel implementation
+//!
+//! Rendezvous channels are equivalent to a channel with a 0-sized buffer: A sender cannot send
+//! until this is an active receiver waiting. This implementation uses a Semaphore to record demand
```
jdisanti, line 9:
```suggestion
//! until there is an active receiver waiting. This implementation uses a Semaphore to record demand
```

rust-runtime/aws-smithy-async/src/future/rendezvous.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Sender-half of a channel
+pub struct Sender<T> {
+    semaphore: Arc<Semaphore>,
+    chan: tokio::sync::mpsc::Sender<T>,
+}
+
+impl<T> Sender<T> {
+    /// Send `item` into the channel.
+    ///
+    /// `send` will block until there is matching demand.
```
jdisanti, line 50: This doc comment is a bit confusing since the function by definition is non-blocking. Maybe say, "`send` will wake when there is matching demand"?

jdisanti (commented):
This is awesome! I have mostly minor comments.

We should probably think some more about the `paginate()` function name. I think the current name implies that the operation isn't paginated by default, and by calling `paginate`, we enable pagination.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/FluentClientDecorator.kt
```diff
@@ -443,8 +453,8 @@ class FluentClientGenerator(
                             #{operation_err},
                             #{input}OperationRetryAlias>,
                         {
-                            let input = self.inner.build().map_err(|err|#{sdk_err}::ConstructionFailure(err.into()))?;
-                            let op = input.make_operation(&self.handle.conf)
+                            let op = self.inner.build().map_err(|err|#{sdk_err}::ConstructionFailure(err.into()))?
+                                .make_operation(&self.handle.conf)
```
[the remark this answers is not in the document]
rcoh: this is actually required so for paginating glacier operations where make_operation takes `&mut self`. By refactoring the code like this, we don't need to conditionally add `mut input`. (Could also allow unused mut)

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/FluentClientDecorator.kt
```diff
@@ -361,12 +367,16 @@ class FluentClientGenerator(
         ) {
             operations.forEach { operation ->
                 val name = symbolProvider.toSymbol(operation).name
+                val fullPath = "crate::client::fluent_builders::$name"
+                val maybePaginated = if (operation.isPaginated(model)) {
+                    "\n/// This operation supports pagination. See [`paginate()`]($fullPath::paginate)."
```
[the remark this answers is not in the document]
rcoh: added dynamodb to the doc preview

.github/workflows/pull-request-bot.yml
```diff
@@ -108,7 +108,7 @@ jobs:
       # STS since aws-config depends on it. STS and Transcribe Streaming were chosen below to stay
       # small while still representing most features. Combined, they are about 11 MB at time of writing.
       run: |
-        ./gradlew -Paws.services=+sts,+transcribestreaming :aws:sdk:assemble
+        ./gradlew -Paws.services=+sts,+transcribestreaming,+dynamodb :aws:sdk:assemble
```
jdisanti, line 111: Comment above this should get updated.

jdisanti (approved):
LGTM!

PR #1014, opened 2021-12-29
fix: exit with error code on failure

Velfi:
add: print directories at start

## Motivation and Context

The tool was exiting with failure but wasn't communicating that failure to the OS

## Description

When app would close due to an error, `call std::process::exit(1)`

## Testing

I ran it locally

### DIFF
--- a/tools/smithy-rs-sync/src/main.rs
+++ b/tools/smithy-rs-sync/src/main.rs
@@ -59,20 +59,15 @@ pub(crate) use here;
 /// --smithy-rs /Users/zhessler/Documents/smithy-rs-test/ \
 /// --aws-sdk /Users/zhessler/Documents/aws-sdk-rust-test/
 /// ```
-fn main() {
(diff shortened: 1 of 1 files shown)


tools/smithy-rs-sync/src/main.rs
```diff
@@ -66,13 +66,26 @@ fn main() {
...
-        eprintln!("Sync failed with error: {:?}", e);
-    };
+    match sync_aws_sdk_with_smithy_rs(&smithy_rs, &aws_sdk, &branch) {
+        Ok(_) => std::process::exit(0),
+        Err(e) => {
+            eprintln!("Sync failed with error: {:?}", e);
+            std::process::exit(1);
+        }
+    }
 }
```
rcoh, line 76: fwiw, I think you can just have `main` return a Result. Probably more future proof

rcoh (approved):

PR #1019, opened 2021-12-30
fix: in generated docs, convert bare anchor tags to pre

Velfi:
## Motivation and Context

https://github.com/awslabs/aws-sdk-rust/issues/357

## Description

In generated docs, convert any `anchor` tags without an `href` attribute to be `pre` tags instead

## Testing

The generated docs can be compared against [these docs](https://docs.rs/aws-sdk-sts/latest/aws_sdk_sts/output/struct.AssumeRoleOutput.html)

## Checklist


Velfi:
An example from the generated diff to prove it works
Screen Shot 2021-12-30 at 1 39 57 PM

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
```diff
@@ -260,6 +262,29 @@ fun <T : CodeWriter> T.docs(text: String, vararg args: Any, newlinePrefix: Strin
...
+    }
+
+    return doc.body().html()
+}
+
+private fun Element.normalizeAnchors() {
+    getElementsByTag("a").forEach {
+        val link = it.attr("href")
+        if (link.isBlank()) {
+            it.changeInto("pre")
```
Velfi, line 279:
```suggestion
            it.changeInto("code")
```

rcoh (approved):

PR #1021, opened 2021-12-30
add: function that exposes profile names

kiiadi:
## Motivation and Context

Tooling will often want the ability to list available profiles for someone to select from. Currently the `ProfileSet` does not expose this.

## Description

Added `fn profiles()` function to `ProfileSet`

## Testing

New unit test added

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/profile/parser.rs
+++ b/aws/rust-runtime/aws-config/src/profile/parser.rs
@@ -134,7 +134,7 @@ impl ProfileSet {
 
     /// Returns the names of the profiles in this profile set
     pub fn profiles(&self) -> impl Iterator<Item=&str> {
-        self.profiles.keys().into_iter().map(|k| k.as_ref())
+        self.profiles.keys().map(String::as_ref)
     }
 
     fn parse(source: Source) -> Result<Self, ProfileParseError> {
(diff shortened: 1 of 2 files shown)


aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -132,6 +132,11 @@ impl ProfileSet {
         self.profiles.is_empty()
     }
 
+    /// Returns the names of the profiles in this profile set
+    pub fn profiles(&self) -> impl Iterator<Item=&str> {
+        self.profiles.keys().into_iter().map(|k| k.as_ref())
+    }
```
Velfi, line 138:
```suggestion
    pub fn profiles(&self) -> impl Iterator<Item=&str> {
        self.profiles.keys().map(String::as_ref)
    }
```

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -235,21 +240,29 @@ mod test {
...
+    #[test]
+    fn profile_names_are_exposed() {
+        let source = make_source(ParserInput {
+            config_file: Some("[profile foo]\n[profile bar]".to_string()),
+            credentials_file: Some("".to_string())
+        });
+
+        let profile_set = ProfileSet::parse(source).expect("profiles loaded");
+
+        let mut profile_names: Vec<&str> = profile_set.profiles().collect();
```
Velfi, line 261:
```suggestion
        let mut profile_names: Vec<_> = profile_set.profiles().collect();
```

You only need to specify the container when collecting (Vec is container in this case)

Velfi:
Make sure you have `precommit` installed per the main README.md
it'll run the formatter for you, plus some other stuff

aws/rust-runtime/aws-config/src/profile/parser.rs
kiiadi: OK cool - will fix.

Velfi (approved):

PR #1023, opened 2022-01-02
Server streaming body

guymguym:

## Motivation and Context
1. Support streaming body in server requests and responses.

## Testing
Generated s3 server codegen and tested get-object/put-object with streams.

## Checklist

CC @crisidev @david-perez

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
@@ -18,6 +18,8 @@ import software.amazon.smithy.rust.codegen.server.smithy.protocols.ServerHttpPro
 import software.amazon.smithy.rust.codegen.smithy.CodegenContext
 import software.amazon.smithy.rust.codegen.smithy.RuntimeType
 import software.amazon.smithy.rust.codegen.smithy.generators.error.errorSymbol
+import software.amazon.smithy.rust.codegen.util.hasStreamingMember
+import software.amazon.smithy.rust.codegen.util.inputShape
 import software.amazon.smithy.rust.codegen.util.outputShape
 
 /**
@@ -132,13 +134,18 @@ class ServerOperationHandlerGenerator(
         } else {
             symbolProvider.toSymbol(operation.outputShape(model)).fullName
         }
+        val streamingBodyTraitBounds = if (operation.inputShape(model).hasStreamingMember(model)) {
+            "\n B: Into<#{SmithyHttpServer}::ByteStream>,"
+        } else {
+            ""
+        }
         return """
             $inputFn
             Fut: std::future::Future<Output = $outputType> + Send,
-            B: $serverCrate::HttpBody + Send + 'static,
+            B: $serverCrate::HttpBody + Send + 'static, $streamingBodyTraitBounds
             B::Data: Send,
             B::Error: Into<$serverCrate::BoxError>,
             $serverCrate::rejection::SmithyRejection: From<<B as $serverCrate::HttpBody>::Error>
-        """
+        """.trimIndent()
     }
 }
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
@@ -41,6 +41,7 @@ class ServerOperationHandlerGenerator(
         "PinProjectLite" to ServerCargoDependency.PinProjectLite.asType(),
         "Tower" to ServerCargoDependency.Tower.asType(),
         "FuturesUtil" to ServerCargoDependency.FuturesUtil.asType(),
+        "SmithyHttp" to CargoDependency.SmithyHttp(runtimeConfig).asType(),
         "SmithyHttpServer" to CargoDependency.SmithyHttpServer(runtimeConfig).asType(),
         "SmithyRejection" to ServerHttpProtocolGenerator.smithyRejection(runtimeConfig),
         "Phantom" to ServerRuntimeType.Phantom,
@@ -135,7 +136,7 @@ class ServerOperationHandlerGenerator(
             symbolProvider.toSymbol(operation.outputShape(model)).fullName
         }
         val streamingBodyTraitBounds = if (operation.inputShape(model).hasStreamingMember(model)) {
-            "\n B: Into<#{SmithyHttpServer}::ByteStream>,"
+            "\n B: Into<#{SmithyHttp}::byte_stream::ByteStream>,"
         } else {
             ""
         }
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
@@ -380,7 +380,7 @@ class ServerProtocolTestGenerator(
             assertOk(rustWriter) {
                 rustWriter.write(
                     "#T(&body, ${
-                        rustWriter.escape(body).dq()
+                    rustWriter.escape(body).dq()
                     }, #T::from(${(mediaType ?: "unknown").dq()}))",
                     RuntimeType.ProtocolTestHelper(codegenContext.runtimeConfig, "validate_body"),
                     RuntimeType.ProtocolTestHelper(codegenContext.runtimeConfig, "MediaType")
@@ -534,10 +534,6 @@ class ServerProtocolTestGenerator(
             FailingTest(RestJson, "RestJsonNoInputAndNoOutput", Action.Response),
             FailingTest(RestJson, "RestJsonNoInputAndOutputWithJson", Action.Response),
             FailingTest(RestJson, "RestJsonSupportsNaNFloatInputs", Action.Request),
-            FailingTest(RestJson, "RestJsonSupportsNaNFloatInputs", Action.Response),
-            FailingTest(RestJson, "RestJsonSimpleScalarProperties", Action.Response),
-            FailingTest(RestJson, "RestJsonSupportsInfinityFloatInputs", Action.Response),
-            FailingTest(RestJson, "RestJsonSupportsNegativeInfinityFloatInputs", Action.Response),
             FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithBlob", Action.Response),
             FailingTest(RestJson, "RestJsonHttpWithEmptyBlobPayload", Action.Request),
             FailingTest(RestJson, "RestJsonHttpWithEmptyStructurePayload", Action.Request),
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
@@ -39,6 +39,7 @@ import software.amazon.smithy.rust.codegen.util.getTrait
 import software.amazon.smithy.rust.codegen.util.hasStreamingMember
 import software.amazon.smithy.rust.codegen.util.hasTrait
 import software.amazon.smithy.rust.codegen.util.inputShape
+import software.amazon.smithy.rust.codegen.util.isStreaming
 import software.amazon.smithy.rust.codegen.util.orNull
 import software.amazon.smithy.rust.codegen.util.outputShape
 import software.amazon.smithy.rust.codegen.util.toSnakeCase
@@ -326,15 +327,37 @@ class ServerProtocolTestGenerator(
             """
             use #{AxumCore}::extract::FromRequest;
             let mut http_request = #{AxumCore}::extract::RequestParts::new(http_request);
-            let input_wrapper = super::$operationName::from_request(&mut http_request).await.expect("failed to parse request");
-            let input = input_wrapper.0;
+            let parsed = super::$operationName::from_request(&mut http_request).await.expect("failed to parse request").0;
             """,
             *codegenScope,
         )
-        if (operationShape.outputShape(model).hasStreamingMember(model)) {
-            rustWriter.rust("""todo!("streaming types aren't supported yet");""")
+
+        if (inputShape.hasStreamingMember(model)) {
+            // A streaming shape does not implement `PartialEq`, so we have to iterate over the input shape's members
+            // and handle the equality assertion separately.
+            for (member in inputShape.members()) {
+                val memberName = codegenContext.symbolProvider.toMemberName(member)
+                if (member.isStreaming(codegenContext.model)) {
+                    rustWriter.rustTemplate(
+                        """
+                        #{AssertEq}(
+                            parsed.$memberName.collect().await.unwrap().into_bytes(),
+                            expected.$memberName.collect().await.unwrap().into_bytes()
+                        );
+                        """,
+                        *codegenScope
+                    )
+                } else {
+                    rustWriter.rustTemplate(
+                        """
+                        #{AssertEq}(parsed.$memberName, expected.$memberName, "Unexpected value for `$memberName`");
+                        """,
+                        *codegenScope
+                    )
+                }
+            }
         } else {
-            rustWriter.rustTemplate("#{AssertEq}(input, expected);", *codegenScope)
+            rustWriter.rustTemplate("#{AssertEq}(parsed, expected);", *codegenScope)
         }
     }
 
@@ -511,12 +534,11 @@ class ServerProtocolTestGenerator(
             FailingTest(RestJson, "RestJsonNoInputAndNoOutput", Action.Response),
             FailingTest(RestJson, "RestJsonNoInputAndOutputWithJson", Action.Response),
             FailingTest(RestJson, "RestJsonSupportsNaNFloatInputs", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithBlob", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithNoBlobBody", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithBlob", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithNoBlobBody", Action.Request),
+            FailingTest(RestJson, "RestJsonSupportsNaNFloatInputs", Action.Response),
+            FailingTest(RestJson, "RestJsonSimpleScalarProperties", Action.Response),
+            FailingTest(RestJson, "RestJsonSupportsInfinityFloatInputs", Action.Response),
+            FailingTest(RestJson, "RestJsonSupportsNegativeInfinityFloatInputs", Action.Response),
             FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithBlob", Action.Response),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithMediaTypeWithBlob", Action.Request),
             FailingTest(RestJson, "RestJsonHttpWithEmptyBlobPayload", Action.Request),
             FailingTest(RestJson, "RestJsonHttpWithEmptyStructurePayload", Action.Request),
 
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
@@ -212,16 +212,16 @@ class ServerProtocolTestGenerator(
 
         rustTemplate(
             """
-                ##[allow(unused_mut)] let mut http_request = http::Request::builder()
-                    .uri("${httpRequestTestCase.uri}")
-                """,
-                *codegenScope
+            ##[allow(unused_mut)] let mut http_request = http::Request::builder()
+                .uri("${httpRequestTestCase.uri}")
+            """,
+            *codegenScope
         )
         for (header in httpRequestTestCase.headers) {
             rust(".header(${header.key.dq()}, ${header.value.dq()})")
         }
         rustTemplate(
-    """
+            """
             .body(#{SmithyHttpServer}::Body::from(#{Bytes}::from_static(b${httpRequestTestCase.body.orNull()?.dq()})))
             .unwrap();
             """,
@@ -386,9 +386,9 @@ class ServerProtocolTestGenerator(
         basicCheck(
             requireHeaders,
             rustWriter,
-           "required_headers",
-           actualExpression,
-           "require_headers"
+            "required_headers",
+            actualExpression,
+            "require_headers"
         )
     }
 
@@ -396,9 +396,9 @@ class ServerProtocolTestGenerator(
         basicCheck(
             forbidHeaders,
             rustWriter,
-           "forbidden_headers",
+            "forbidden_headers",
             actualExpression,
-           "forbid_headers"
+            "forbid_headers"
         )
     }
 
@@ -513,14 +513,10 @@ class ServerProtocolTestGenerator(
             FailingTest(RestJson, "RestJsonSupportsNaNFloatInputs", Action.Request),
             FailingTest(RestJson, "RestJsonStreamingTraitsWithBlob", Action.Request),
             FailingTest(RestJson, "RestJsonStreamingTraitsWithNoBlobBody", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithBlob", Action.Response),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithNoBlobBody", Action.Response),
             FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithBlob", Action.Request),
             FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithNoBlobBody", Action.Request),
             FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithBlob", Action.Response),
-            FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithNoBlobBody", Action.Response),
             FailingTest(RestJson, "RestJsonStreamingTraitsWithMediaTypeWithBlob", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithMediaTypeWithBlob", Action.Response),
             FailingTest(RestJson, "RestJsonHttpWithEmptyBlobPayload", Action.Request),
             FailingTest(RestJson, "RestJsonHttpWithEmptyStructurePayload", Action.Request),
 
@@ -591,56 +587,64 @@ class ServerProtocolTestGenerator(
                 ).asObjectNode().get()
             ).build()
         private fun fixRestJsonAllQueryStringTypes(testCase: HttpRequestTestCase): HttpRequestTestCase =
-             testCase.toBuilder().params(
-                 Node.parse("""{
-                    "queryString": "Hello there",
-                    "queryStringList": ["a", "b", "c"],
-                    "queryStringSet": ["a", "b", "c"],
-                    "queryByte": 1,
-                    "queryShort": 2,
-                    "queryInteger": 3,
-                    "queryIntegerList": [1, 2, 3],
-                    "queryIntegerSet": [1, 2, 3],
-                    "queryLong": 4,
-                    "queryFloat": 1.1,
-                    "queryDouble": 1.1,
-                    "queryDoubleList": [1.1, 2.1, 3.1],
-                    "queryBoolean": true,
-                    "queryBooleanList": [true, false, true],
-                    "queryTimestamp": 1,
-                    "queryTimestampList": [1, 2, 3],
-                    "queryEnum": "Foo",
-                    "queryEnumList": ["Foo", "Baz", "Bar"],
-                    "queryParamsMapOfStringList": {
-                        "String": ["Hello there"],
-                        "StringList": ["a", "b", "c"],
-                        "StringSet": ["a", "b", "c"],
-                        "Byte": ["1"],
-                        "Short": ["2"],
-                        "Integer": ["3"],
-                        "IntegerList": ["1", "2", "3"],
-                        "IntegerSet": ["1", "2", "3"],
-                        "Long": ["4"],
-                        "Float": ["1.1"],
-                        "Double": ["1.1"],
-                        "DoubleList": ["1.1", "2.1", "3.1"],
-                        "Boolean": ["true"],
-                        "BooleanList": ["true", "false", "true"],
-                        "Timestamp": ["1970-01-01T00:00:01Z"],
-                        "TimestampList": ["1970-01-01T00:00:01Z", "1970-01-01T00:00:02Z", "1970-01-01T00:00:03Z"],
-                        "Enum": ["Foo"],
-                        "EnumList": ["Foo", "Baz", "Bar"]
+            testCase.toBuilder().params(
+                Node.parse(
+                    """
+                    {
+                        "queryString": "Hello there",
+                        "queryStringList": ["a", "b", "c"],
+                        "queryStringSet": ["a", "b", "c"],
+                        "queryByte": 1,
+                        "queryShort": 2,
+                        "queryInteger": 3,
+                        "queryIntegerList": [1, 2, 3],
+                        "queryIntegerSet": [1, 2, 3],
+                        "queryLong": 4,
+                        "queryFloat": 1.1,
+                        "queryDouble": 1.1,
+                        "queryDoubleList": [1.1, 2.1, 3.1],
+                        "queryBoolean": true,
+                        "queryBooleanList": [true, false, true],
+                        "queryTimestamp": 1,
+                        "queryTimestampList": [1, 2, 3],
+                        "queryEnum": "Foo",
+                        "queryEnumList": ["Foo", "Baz", "Bar"],
+                        "queryParamsMapOfStringList": {
+                            "String": ["Hello there"],
+                            "StringList": ["a", "b", "c"],
+                            "StringSet": ["a", "b", "c"],
+                            "Byte": ["1"],
+                            "Short": ["2"],
+                            "Integer": ["3"],
+                            "IntegerList": ["1", "2", "3"],
+                            "IntegerSet": ["1", "2", "3"],
+                            "Long": ["4"],
+                            "Float": ["1.1"],
+                            "Double": ["1.1"],
+                            "DoubleList": ["1.1", "2.1", "3.1"],
+                            "Boolean": ["true"],
+                            "BooleanList": ["true", "false", "true"],
+                            "Timestamp": ["1970-01-01T00:00:01Z"],
+                            "TimestampList": ["1970-01-01T00:00:01Z", "1970-01-01T00:00:02Z", "1970-01-01T00:00:03Z"],
+                            "Enum": ["Foo"],
+                            "EnumList": ["Foo", "Baz", "Bar"]
+                        }
                     }
-                }""".trimMargin()).asObjectNode().get()
-             ).build()
+                    """.trimMargin()
+                ).asObjectNode().get()
+            ).build()
         private fun fixRestJsonQueryStringEscaping(testCase: HttpRequestTestCase): HttpRequestTestCase =
             testCase.toBuilder().params(
-                Node.parse("""{
-                   "queryString": "%:/?#[]@!${'$'}&'()*+,;=😹",
-                   "queryParamsMapOfStringList": {
-                       "String": ["%:/?#[]@!${'$'}&'()*+,;=😹"]
-                   }
-               }""".trimMargin()).asObjectNode().get()
+                Node.parse(
+                    """
+                    {
+                        "queryString": "%:/?#[]@!${'$'}&'()*+,;=😹",
+                        "queryParamsMapOfStringList": {
+                            "String": ["%:/?#[]@!${'$'}&'()*+,;=😹"]
+                        }
+                    }
+                    """.trimMargin()
+                ).asObjectNode().get()
             ).build()
         // This test assumes that errors in responses are identified by an `X-Amzn-Errortype` header with the error shape name.
         // However, Smithy specifications for AWS protocols that serialize to JSON recommend that new server implementations
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
@@ -119,6 +119,7 @@ private class ServerHttpProtocolImplGenerator(
         "PercentEncoding" to CargoDependency.PercentEncoding.asType(),
         "Regex" to CargoDependency.Regex.asType(),
         "SerdeUrlEncoded" to ServerCargoDependency.SerdeUrlEncoded.asType(),
+        "SmithyHttp" to CargoDependency.SmithyHttp(runtimeConfig).asType(),
         "SmithyHttpServer" to CargoDependency.SmithyHttpServer(runtimeConfig).asType(),
         "SmithyRejection" to ServerHttpProtocolGenerator.smithyRejection(runtimeConfig),
         "http" to RuntimeType.http
@@ -1056,7 +1057,7 @@ private class ServerHttpProtocolImplGenerator(
 
     private fun getStreamingBodyTraitBounds(operationShape: OperationShape): String {
         if (operationShape.inputShape(model).hasStreamingMember(model)) {
-            return "\n B: Into<#{SmithyHttpServer}::ByteStream>,"
+            return "\n B: Into<#{SmithyHttp}::byte_stream::ByteStream>,"
         } else {
             return ""
         }
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
@@ -8,7 +8,6 @@ package software.amazon.smithy.rust.codegen.server.smithy.protocols
 import software.amazon.smithy.aws.traits.protocols.RestJson1Trait
 import software.amazon.smithy.aws.traits.protocols.RestXmlTrait
 import software.amazon.smithy.codegen.core.Symbol
-import software.amazon.smithy.model.knowledge.HttpBinding
 import software.amazon.smithy.model.knowledge.HttpBindingIndex
 import software.amazon.smithy.model.node.ExpectationNotMetException
 import software.amazon.smithy.model.shapes.CollectionShape
@@ -55,6 +54,7 @@ import software.amazon.smithy.rust.codegen.smithy.protocols.parse.StructuredData
 import software.amazon.smithy.rust.codegen.util.UNREACHABLE
 import software.amazon.smithy.rust.codegen.util.dq
 import software.amazon.smithy.rust.codegen.util.expectTrait
+import software.amazon.smithy.rust.codegen.util.findStreamingMember
 import software.amazon.smithy.rust.codegen.util.getTrait
 import software.amazon.smithy.rust.codegen.util.hasStreamingMember
 import software.amazon.smithy.rust.codegen.util.inputShape
@@ -132,13 +132,12 @@ private class ServerHttpProtocolImplGenerator(
     }
 
     /*
-     * Generation of `FromRequest` and `IntoResponse`. They are currently only implemented for non-streaming request
-     * and response bodies, that is, models without streaming traits
-     * (https://awslabs.github.io/smithy/1.0/spec/core/stream-traits.html).
-     * For non-streaming request bodies, we require the HTTP body to be fully read in memory before parsing or
-     * deserialization. From a server perspective we need a way to parse an HTTP request from `Bytes` and serialize
+     * Generation of `FromRequest` and `IntoResponse`.
+     * For non-streaming request bodies, that is, models without streaming traits
+     * (https://awslabs.github.io/smithy/1.0/spec/core/stream-traits.html)
+     * we require the HTTP body to be fully read in memory before parsing or deserialization.
+     * From a server perspective we need a way to parse an HTTP request from `Bytes` and serialize
      * an HTTP response to `Bytes`.
-     * TODO Add support for streaming.
      * These traits are the public entrypoint of the ser/de logic of the `aws-smithy-http-server` server.
      */
     private fun RustWriter.renderTraits(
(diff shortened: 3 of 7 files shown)


guymguym:
Hmm, I wonder why `hyper::Body::wrap_stream` is not found when building codegen-server-test
```
error[E0599]: no function or associated item named `wrap_stream` found for struct `Body` in the current scope
    --> rest_json/rust-server-codegen/src/operation_ser.rs:1742:59
     |
1742 |         let payload = aws_smithy_http_server::body::Body::wrap_stream(output.blob);
     |                                                           ^^^^^^^^^^^ function or associated item not found in `Body`

error[E0599]: no function or associated item named `wrap_stream` found for struct `Body` in the current scope
    --> rest_json/rust-server-codegen/src/operation_ser.rs:1771:59
     |
1771 |         let payload = aws_smithy_http_server::body::Body::wrap_stream(output.blob);
     |                                                           ^^^^^^^^^^^ function or associated item not found in `Body`

error[E0599]: no function or associated item named `wrap_stream` found for struct `Body` in the current scope
    --> rest_json/rust-server-codegen/src/operation_ser.rs:1800:59
     |
1800 |         let payload = aws_smithy_http_server::body::Body::wrap_stream(output.blob);
     |                                                           ^^^^^^^^^^^ function or associated item not found in `Body`
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
```diff
@@ -132,13 +135,19 @@ class ServerOperationHandlerGenerator(
         } else {
             symbolProvider.toSymbol(operation.outputShape(model)).fullName
         }
+        val streamingBodyTraitBounds = if (operation.inputShape(model).hasStreamingMember(model)) {
+            "B: Into<#{SmithyHttp}::byte_stream::ByteStream>,"
```
crisidev, line 139: The gut feeling is that we should re-export ByteStream from smithy-http into smithy-http-server. This way we can "control" the API a little more and prevent breaking changes since the server codegen does not own smithy-http.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -801,6 +810,109 @@ private class ServerHttpProtocolImplGenerator(
...
+                    writer.rustTemplate(
+                        """
+                        use std::convert::TryFrom;
+                        let header_value = $safeName;
+                        let header_value = http::header::HeaderValue::try_from(&*header_value)
+                            .map_err(|err| {
+                                #{SmithyHttpServer}::rejection::Serialize::from(
+                                   ${(memberName + " cannot be used as a header value").dq()}
+                                )
+                            })?;
```
crisidev, line 837: Nit: `|err|` is unused here.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -801,6 +810,109 @@ private class ServerHttpProtocolImplGenerator(
...
+                """
+                for (k, v) in $field {
+                    use std::str::FromStr;
+                    let header_name = http::header::HeaderName::from_str(
+                            &format!("{}{}", ${binding.locationName.dq()}, &k)
+                        )
+                        .map_err(|err| {
+                            #{SmithyHttpServer}::rejection::Serialize::from(
+                                ${(memberName + " cannot be used as a header name").dq()}
+                            )
```
crisidev, line 868: Nit: `|err|` is unused here.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -801,6 +810,109 @@ private class ServerHttpProtocolImplGenerator(
...
+                        })?;
+                    use std::convert::TryFrom;
+                    let header_value = ${headerFmtFun(writer, target, memberShape, "v")};
+                    let header_value = http::header::HeaderValue::try_from(header_value)
+                        .map_err(|err| {
+                            #{SmithyHttpServer}::rejection::Serialize::from(
+                                ${(memberName + " cannot be used as a header value").dq()}
+                            )
+                        })?;
```
crisidev, line 877: Nit: `|err|` is unused here.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -801,6 +810,109 @@ private class ServerHttpProtocolImplGenerator(
...
+                val timestampFormatType = RuntimeType.TimestampFormat(runtimeConfig, timestampFormat)
+                "$targetName.fmt(${writer.format(timestampFormatType)})?"
+            }
+            target.isListShape || target.isMemberShape -> {
+                throw IllegalArgumentException("lists should be handled at a higher level")
+            }
+            target.isPrimitive() -> {
+                "encoder.encode()"
+            }
+            else -> throw CodegenException("unexpected shape: $target")
```
crisidev, line 912: I am not 100% sure we want the codegen to fail here. I think I agree with bailing out in case of an unexpected shape, but for the list shape we probably want to codegen a `todo!` in rust and let it fail when the crate is built. @david-perez WDYT?

crisidev:
> Hmm, I wonder why `hyper::Body::wrap_stream` is not found when building codegen-server-test
> 
> ```
> error[E0599]: no function or associated item named `wrap_stream` found for struct `Body` in the current scope
>     --> rest_json/rust-server-codegen/src/operation_ser.rs:1742:59
>      |
> 1742 |         let payload = aws_smithy_http_server::body::Body::wrap_stream(output.blob);
>      |                                                           ^^^^^^^^^^^ function or associated item not found in `Body`
> 
> error[E0599]: no function or associated item named `wrap_stream` found for struct `Body` in the current scope
>     --> rest_json/rust-server-codegen/src/operation_ser.rs:1771:59
>      |
> 1771 |         let payload = aws_smithy_http_server::body::Body::wrap_stream(output.blob);
>      |                                                           ^^^^^^^^^^^ function or associated item not found in `Body`
> 
> error[E0599]: no function or associated item named `wrap_stream` found for struct `Body` in the current scope
>     --> rest_json/rust-server-codegen/src/operation_ser.rs:1800:59
>      |
> 1800 |         let payload = aws_smithy_http_server::body::Body::wrap_stream(output.blob);
>      |                                                           ^^^^^^^^^^^ function or associated item not found in `Body`
> ```

[wrap_stream](https://docs.rs/hyper/latest/src/hyper/body/body.rs.html#188) is available only with the `stream` feature. You will need to add the feature to [aws-smithy-http-server Cargo.toml](https://github.com/awslabs/smithy-rs/blob/main/rust-runtime/aws-smithy-http-server/Cargo.toml#L29).

crisidev (changes requested):
This is amazing @guymguym !! I left a couple of comments, nothing major   :)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
crisidev: @david-perez thoughts?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: BTW I copied this behavior from RequestBindingGenerator so it might be considered in the same way -

https://github.com/awslabs/smithy-rs/blob/36faba8ee6860208b821c384f8a611fbd87f8fae/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/RequestBindingGenerator.kt#L263-L269

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
rcoh: we re-export ByteStream from generated services FYI

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
rcoh: on the client side parsing lists is quite challenging so we handle it at the `List<T>` level—I'm not sure how you've handled it here (haven't read the PR yet), but I would say if:

- Lists aren't supported: Log a big warning and generate `todo!()`
- Lists are supported and handled elsewhere: `PANIC` or `UNREACHABLE`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
guymguym: that's right - I see it re-exported in the lib.rs of any generated server crate like so:
```
/// Crate version number.
pub static PKG_VERSION: &str = env!("CARGO_PKG_VERSION");
pub use aws_smithy_http::byte_stream::ByteStream;
pub use aws_smithy_http::result::SdkError;
pub use aws_smithy_types::Blob;
pub use aws_smithy_types::DateTime;
```

guymguym: @david-perez notice that also SdkBody is exported there ^ not sure if there's a reason behind it or not

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -906,4 +1018,12 @@ private class ServerHttpProtocolImplGenerator(
             }
         }
     }
+
+    private fun getStreamingBodyTraitBounds(operationShape: OperationShape): String {
+        if (operationShape.inputShape(model).hasStreamingMember(model)) {
+            return "B: Into<#{SmithyHttp}::byte_stream::ByteStream>,"
```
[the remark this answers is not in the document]
guymguym: Does the conversion I added look OK to you @david-perez ?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -178,21 +176,20 @@ private class ServerHttpProtocolImplGenerator(
         )
 
         // Implement Axum `IntoResponse` for output types.
+        // For streaming response bodies, the service implementer will return a `Stream` from their handler.
+
```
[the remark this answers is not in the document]
guymguym: I tried to fix the comments which I stumbled across, but probably missed a few.
Let me know if you notice any comments that should be amended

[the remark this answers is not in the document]
guymguym: accepted suggestion. thanks

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -361,7 +353,7 @@ private class ServerHttpProtocolImplGenerator(
         return RuntimeType.forInlineFun(fnName, operationSerModule) {
             Attribute.Custom("allow(clippy::unnecessary_wraps)").render(it)
             it.rustBlockTemplate(
```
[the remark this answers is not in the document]
guymguym: accepted suggestion. thanks

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: fixed

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: fixed

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: fixed

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
guymguym: For now I re-exported from aws_smithy_http_server too and used the re-exported type instead.

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -326,6 +326,14 @@ impl From<Vec<u8>> for ByteStream {
     }
 }
 
+impl From<hyper::Body> for ByteStream {
```
rcoh, line 329: SdkBody actually implements `From<hyper::Body>`. So you can do `ByteStream::new(SdkBody::from(input))`

rcoh (changes requested):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -525,6 +525,15 @@ private class ServerHttpProtocolImplGenerator(
                 *codegenScope,
                 "parser" to parser,
             )
+        } else if (streamingMember != null) {
+            // TODO Need contentTypeCheck here?
```
david-perez, line 529: Why wouldn't we?

rust-runtime/aws-smithy-http/src/byte_stream.rs
guymguym: @rcoh that doesn't work because of the boxed error type. I tried that at first.
You can see how I converted from hyper::Body to ByteStream :
```
        ByteStream::new(SdkBody::from_dyn(
            input.map_err(|e| e.into_cause().unwrap()).boxed(),
        ))
```

So I preferred to have that specifically done outside of generated code, but can change to either way.

rcoh: what issues are caused by the box error type?

this seems to compile fine:
```rust
impl From<hyper::Body> for ByteStream {
    fn from(body: hyper::Body) -> Self {
        ByteStream::new(SdkBody::from(body))
    }
}
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: @david-perez The previous case uses getContentTypeCheck() to verify the content-type is json/xml which is parseable by the protocol. But with blob streams the content-type can be anything at all... So I think this TODO can just be removed and we won't need to verify the content-type in the streaming case.

rust-runtime/aws-smithy-http/src/byte_stream.rs
guymguym: Let me capture the error in a few minutes and paste. I think this is related to the generated code trait bounds which are not explicitly using hyper::Body.

guymguym: Here is the generated code for example:
```
pub async fn parse_put_object_request<B>(
    #[allow(unused_variables)] request: &mut axum_core::extract::RequestParts<B>,
) -> std::result::Result<
    crate::input::PutObjectInput,
>
where
    B: aws_smithy_http_server::HttpBody + Send,
    B: Into<aws_smithy_http::byte_stream::ByteStream>,
    B::Data: Send,
    B::Error: Into<aws_smithy_http_server::BoxError>,
    aws_smithy_http_server::rejection::SmithyRejection:
        From<<B as aws_smithy_http_server::HttpBody>::Error>,
{
    Ok({
        #[allow(unused_mut)]
        let mut input = crate::input::put_object_input::Builder::default();
        let body = request
            .take_body()
            .ok_or(aws_smithy_http_server::rejection::BodyAlreadyExtracted)?;
        input = input.set_body(Some(
            aws_smithy_http::byte_stream::ByteStream::new(
                aws_smithy_http::body::SdkBody::from(body)
            )
        ));
        // input = input.set_body(Some(body.into()));
```

And the error is:
```
error[E0277]: the trait bound `SdkBody: From<B>` is not satisfied
    --> codegen/build/smithyprojections/s3d/s3/rust-server-codegen/src/operation_deser.rs:4173:17
     |
4173 |                 aws_smithy_http::body::SdkBody::from(body)
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<B>` is not implemented for `SdkBody`
     |
help: consider extending the `where` bound, but there might be an alternative better way to express this requirement
     |
4163 |         From<<B as aws_smithy_http_server::HttpBody>::Error>, SdkBody: From<B>
     |                                                             ~~~~~~~~~~~~~~~~~~
```

So instead of the trait bound that I added `B: Into<aws_smithy_http::byte_stream::ByteStream>` I could do `B: Into<aws_smithy_http::body::SdkBody>` and then do as you suggested. The only question is why we need SdkBody here and not skip directly to the stream. CC @david-perez

rcoh: to be honest, the client used to have generic bodies but it ended up being a big hassle. It might be worth just using SdkBody directly and removing the generated generic

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: @crisidev Would be great if you can direct me here

rust-runtime/aws-smithy-http/src/byte_stream.rs
rcoh: anyway, maybe I'm missing something but adding the `From<hyper::Body>` doesn't seem related to this function which uses `HttpBody`?

guymguym: I think the confusion is that it doesn't use HttpBody trait, which doesn't have any conversion to a stream, instead it uses the trait bound that I added `B: Into<ByteStream>` which I then had to implement for the hyper::Body struct to make a hyper server compile.

guymguym: In any case I can try any direction you have to improve and simplify, but I think the generic nature of the generated server code that allows any concrete body type that implements the required bounds is the reason for this complexity, so if we want to change the generated server code to fixate on SdkBody struct that will make the code simpler, but less flexible if theres a need to support another server stack.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
david-perez: @guymguym So @crisidev and I took the design decision a while back to **not** use `SdkBody` in the server, since it's a special body type that caters to the [AWS Rust SDK's](https://github.com/awslabs/aws-sdk-rust) needs for _dispatching_ HTTP requests. From a technical perspective there's no impediment to using it in the server at the moment, but in the future this type might evolve further to satisfy the requirements of the AWS Rust SDK.

We also considered having a server body type of our own, but eventually we decided that since we don't have any special requirements, we should write our interfaces against the trait [`http_body::Body`](https://docs.rs/http-body/latest/http_body/trait.Body.html) wherever possible. We also decided to reexport an HTTP body trait from `aws_smithy_http_server` that endorsed [`hyper::body::HttpBody`](https://docs.rs/hyper/latest/hyper/body/trait.HttpBody.html), and unboxed and boxed dyn body types that endorsed [`hyper::body::Body`](https://docs.rs/hyper/latest/hyper/body/struct.Body.html) and [`http_body::comibnators::UnsyncBoxBody`](https://docs.rs/http-body/0.4.4/http_body/combinators/struct.UnsyncBoxBody.html), respectively. The rationale was that these crates have good stable types that are used throughout the Rust ecosystem, and as such users would have a better experience if they wanted to use these bodies in other libraries, and we retain some degree of autonomy by reexporting them in case in the future we'd like to expose a body type of our own. It's also the approach that other frameworks took, [like `axum`](https://docs.rs/axum/latest/axum/body/index.html).

Truth be told, Smithy service implementers shouldn't have to worry too much about body types, since they are hidden well inside the framework and they just code against modeled input and output types, but it's inevitable to expose them in the public API (e.g. in the Router). So in that sense we differ from the typical web framework. Another place where they seep in is here: in streaming requests and responses. I didn't even realize we were exposing `SdkBody` via `ByteStream` in service's `lib.rs`; I guess because we hadn't implemented streaming so we've never used `ByteStream`.

But now that you're adding streaming support, we need to decide and commit more firmly on this design aspect. Personally, I think the server should _still continue_ to use the `http_body` and `hyper` types and not expose `SdkBody`, at least until we have a need for `SdkBody` or our own server body type. This means we can't expose `ByteStream`, since it's tied to `SdkBody`, but the `ByteStream` abstraction is perfect and just what we need for this feature (aside: this is a recurring pattern you'll notice in the codebase, since it got started for the needs of the AWS Rust SDK and we've been slowly refactoring it and stealing useful utilities for the server side where applicable). So to move forward and land this feature, I think I lean towards wrapping `ByteStream` in `aws_smithy_http_server` to hide away the fact that it's using `SdkBody` underneath, and convert between body types in our implementation if we need to until we figure out how best to write it for service implementors. For instance, I haven't looked too closely at `ByteStream` implementation, but one thing I'm missing that I foresee service implementors will need is the ability for them to create and return a `ByteStream` from an arbitrary [`std::stream::Stream`](https://doc.rust-lang.org/std/stream/trait.Stream.html) from their handlers. We can also have the server's `ByteStream` start very simple (just taking in `&'static [u8]`, for instance) and slowly expand it in future PRs. This is where **you can greatly help us**, since we don't have any customers that are using streaming in their Rust services. Whatever your application's needs will be in this regard, we can slowly add them to the streaming support.

But what I'm certain of is that we should not expose / endorse **both** `SdkBody` and the other body types, since it will make things confusing for users.

---

Sorry for the long comment and thanks for the PR! You're helping us tremendously with these and I'm excited to see what you will build with this project.

guymguym: @david-perez Thanks for all the context! That's very helpful for me as an external collaborator. 

I understand you design direction, but this also requires changing the type in the generated structs. Do you want me to take a stab at creating the server byte_stream in this PR? Or do you prefer to commit and that I make another PR for that change?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
david-perez: @guymguym `@streaming` can only be applied to blob or union, and the specs say what `Content-Type` to use in those cases ([example for AWS XML](https://awslabs.github.io/smithy/1.0/spec/aws/aws-restxml-protocol.html#content-type)).

guymguym: That won't work for s3 which relies on the sent content-type header as the object type...

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
david-perez: @guymguym 

> but this also requires changing the type in the generated structs

That's right. In struct generation, instead of generating:

```rust
#[non_exhaustive]
pub struct StreamingTraitsOutput {
    pub foo: std::option::Option<std::string::String>,
    pub blob: aws_smithy_http::byte_stream::ByteStream,
}
```

we would generate `pub blob: aws_smithy_http_server::byte_stream::ByteStream`.

> Do you want me to take a stab at creating the server byte_stream in this PR?

Yeah, let's try that first. Make separate commits so that in case the diff gets too big we can easily split the PR and merge what you've done already.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
guymguym: Alrighty then. Lets go for it! I'll update soon and ping for your review.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
david-perez: @guymguym Ah, crap. I'm guessing you already verified the S3 service is accepting any MIME type for PutObject's `Content-Type`. Do they also return the MIME type you provide in PutObject when calling GetObject, or do they always return `application/octet-stream`? This seems like it's another of these instances where S3 deviates from the protocol they say they adhere to (AWS restXml). S3 is one of the earliest AWS services and as such I suspect you will encounter these idiosyncrasies if you're trying to build a service that mimics its behavior 100%.

My idea to support these deviations would be to add "escape-hatches" in [`ServerCodegenConfig`](https://github.com/awslabs/smithy-rs/blob/main/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustSettings.kt#L27) that would slightly modify codegen behavior (e.g. a `doNotCheckContentType` boolean). In the beginning we were using `CodegenConfg` (i.e. the client's codegen settings) but we realized we were going to need a completely different set of settings, so we refactored in https://github.com/awslabs/smithy-rs/pull/995 to start having our own data class. However, we soon realized that `CodegenConfig` is used everywhere in the codebase, which would thus need a large refactor, so our `ServerCodegenConfig` just converts to a `CodegenConfig` under the hood 😅.

I'm OK with not checking `Content-Type` _for the moment_ in this PR, since that refactor is probably large and outside the scope of this PR, but just bear in mind that we have to adhere to the specs, so it eventually will have to go in; I'm working on adding [a suite of protocol tests](https://github.com/awslabs/smithy/tree/main/smithy-aws-protocol-tests/model/restJson1) for restJson1 and we will then have to whittle down the test failures until we are fully compliant.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
david-perez: @guymguym 

> One question is if we can reuse the code from RequestBindingGenerator.kt where I took the headers generation code to ServerHttpProtocolGenerator.kt.

Yes, we should aim for reuse here since our server implementation just deviates in minor things as far as I can see:

1. The variable name is `output` instead of `input`,
2. The variable name is `response` instead of `builder`,
3. We're mapping directly to `rejection::Serialize` via `http::header::HeaderValue::try_from(&*header_value).map_err` instead of using [`aws_smithy_http::operation::BuildError`](https://docs.rs/aws-smithy-http/latest/aws_smithy_http/operation/enum.BuildError.html); and
4. We're not calling `redactIfNecessary` in case of header serialize error.

We should make `renderHeaders` and `renderPrefixHeaders` from `RequestBindingGenerator.kt` public and reconcile. (4) we need to do as well. (1) and (2) we can parametrize and document the methods so that they take in the variable names (alternatively, for (2), we can rename our `response` to `builder` to simplify a bit). For (3), I suggest we keep using `BuildError` and write a `From` impl [in rejection.rs](https://github.com/awslabs/smithy-rs/blob/main/rust-runtime/aws-smithy-http-server/src/rejection.rs) to put it inside a `Serialize` rejection, so we can keep using `?` to bubble it up.

---

@rcoh lists _are supported_ in `@httpHeader` and `@httpPrefixHeader` like in the client and they are being handled higher via [`listForEach` here](https://github.com/awslabs/smithy-rs/blob/main/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/RequestBindingGenerator.kt#L213); this code block has been copied from there. So throwing `IllegalArgumentException` or `PANIC` or `UNREACHABLE` is correct here; does the codebase have a convention on which to use when?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: @david-perez Yeah I know, S3 is a bit different, and I also understand your approach with "escape-hatches" and it sounds like a perfect fit for my use case because I have no problems including s3 specific logic in my code, while I am just trying to get all the generic smithy codegen automated. I'll check out ServerCodegenConfig.

And btw thanks for opening the issue for smithy about the s3 model - I wonder what is the answer - if the model itself was not fully translated to smithy, or perhaps it's focused on the client perspective of the service and these are the requirements they want from *new* clients.

rust-runtime/aws-smithy-http/src/byte_stream.rs
rcoh: got it! ok I understand now. Here's what I would do:

Being able to construct a ByteStream from HttpBody is generally useful, we should support that. We should add a purpose built convert to create a ByteStream from the server generic—since this will be just be a fat pointer to whatever the server is using, it maintains flexibility with minimal overhead. 

I probably _would not_ change to smithy_http_server::ByteStream but that is up to @david-perez

david-perez: @rcoh 

> I probably would not change to smithy_http_server::ByteStream but that is up to @david-perez

The only reasons why I suggested this is because (1) `ByteStream` is tied to `SdkBody` and (2) we would expose `SdkBody` in the public API and server users will be confused. If we can make `ByteStream` generic over the body type and thus reuse code across client and server that would be perfect.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
rcoh: `UNREACHABLE` is intended for cases where if kotlin had exhaustiveness checking, the branch wouldn't be required

`PANIC` is for a bug in the code

rust-runtime/aws-smithy-http/src/byte_stream.rs
rcoh: I tried this for the client—it just causes loads of complexity for very little gain (IMO), especially since SDK Body is capable of holding a `dyn Body`.

We _could_ add a generic to ByteStream with a default parameter actually...but I'd need to be a little more convinced of it's utility

guymguym: @rcoh @david-perez 
I don't think I have the context to assist in this decision at this time. 

Would it make sense that for this PR I just keep using ByteStream as is (or re-export as is from smithy-http-server) and we can continue adjusting this in subsequent PRs?

guymguym: What do you say @david-perez ?

guymguym: @david-perez ?

david-perez: @guymguym Sure, we can leave that change for a subsequent PR if it makes it easier for you.

guymguym: That will be good. Thanks

david-perez: @guymguym We have decided that we will use [`aws_smithy_http::byte_stream::ByteStream`](https://docs.rs/aws-smithy-http/latest/aws_smithy_http/byte_stream/struct.ByteStream.html), making all methods referencing `SdkBody` `#[doc(hidden)]` (`new`, `into_inner`, `From<SdkBody>` impl), and we will add generic methods that don’t reference `SdkBody`.

cc @crisidev @rcoh

guymguym: @david-perez Sounds good. 
Should I change anything in this `impl From<hyper::Body> for ByteStream` code?

rust-runtime/aws-smithy-http-server/src/lib.rs
```diff
@@ -30,6 +30,9 @@ pub use self::routing::Router;
 #[doc(inline)]
 pub use tower_http::add_extension::{AddExtension, AddExtensionLayer};
 
+#[doc(inline)]
+pub use aws_smithy_http::byte_stream::ByteStream;
```
guymguym, line 34: @david-perez @crisidev Should I keep this re-export of ByteStream in the server?

crisidev: I think this is fine for now. We are discussing about creating our own BytesStream object vs using the SDK provided one. 

Having the export IMO moves us in the right direction even if we will change the implementation in the near future.

Thoughts @david-perez?

I think it should not be the aim of this PR to sort out the ByteStream implementation but rather have an initial implementation of the codegen needed for streaming types.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: This is being refactored in #1086. I'll remove my code once rebased. Resolving.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
guymguym: Conclusion in https://github.com/awslabs/smithy-rs/pull/1023#discussion_r784751006. Resolving.

rust-runtime/aws-smithy-http-server/src/lib.rs
david-perez: @crisidev 

> Thoughts @david-perez?

https://github.com/awslabs/smithy-rs/pull/1023#discussion_r784751006

> I think it should not be the aim of this PR to sort out the ByteStream implementation but rather have an initial implementation of the codegen needed for streaming types.

OK. Let's track that here https://github.com/awslabs/smithy-rs/issues/1125.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
```diff
@@ -132,13 +134,18 @@ class ServerOperationHandlerGenerator(
         } else {
             symbolProvider.toSymbol(operation.outputShape(model)).fullName
         }
+        val streamingBodyTraitBounds = if (operation.inputShape(model).hasStreamingMember(model)) {
+            "\n B: Into<#{SmithyHttpServer}::ByteStream>,"
```
david-perez, line 138: You can remove the `\n` and interpolate the variable in the next line.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -1052,4 +1052,12 @@ private class ServerHttpProtocolImplGenerator(
             }
         }
     }
+
+    private fun getStreamingBodyTraitBounds(operationShape: OperationShape): String {
+        if (operationShape.inputShape(model).hasStreamingMember(model)) {
+            return "\n B: Into<#{SmithyHttpServer}::ByteStream>,"
```
david-perez, line 1058: You can remove the `\n` and interpolate the variable in the next line.

david-perez (approved):
I'd like to merge this PR this week.

I did another pass and only found one more nitpick.

I'm approving. Can you please merge and resolve conflicts?

crisidev:
I'll give this another pass this morning as well.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
guymguym: @david-perez the reason I did this was that the output is usually not including this streaming member and then it generates an empty line, which is not causing any compilation issues, but is weird to watch...

crisidev: We run "cargo fmt" at the end of the codegen, so the weird look should go away.. Is it not?

crisidev (approved):
Looks good!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
guymguym: Nope, the empty line is still there. Perhaps because it's just an empty line then cargo fmt will not remove it.

guymguym: is there a better way to generate an optional line?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -525,14 +525,14 @@ class ServerProtocolTestGenerator(
             FailingTest(RestJson, "RestJsonSupportsNegativeInfinityFloatInputs", Action.Response),
             FailingTest(RestJson, "RestJsonStreamingTraitsWithBlob", Action.Request),
             FailingTest(RestJson, "RestJsonStreamingTraitsWithNoBlobBody", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithBlob", Action.Response),
```
guymguym, line 528: @david-perez @crisidev I was not sure what to do here - should I remove these tests from the expect-fail list as these now seem to pass...?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -579,13 +576,23 @@ private class ServerHttpProtocolImplGenerator(
...
+                *codegenScope
+            )
         }
-        for (binding in bindings) {
-            val member = binding.member
-            val parsedValue = serverRenderBindingParser(binding, operationShape, httpBindingGenerator, structuredDataParser)
-            if (parsedValue != null) {
-                withBlock("input = input.${member.setterName()}(", ");") {
-                    parsedValue(this)
+        if (streamingMember == null) {
```
guymguym, line 588: I wanted to highlight this condition I added - I had to check it so that with streaming there will be no bindings parsers generated. The weird thing is just the flow of this function - the conditions seems a bit detached from one another but I wasn't sure if there's a way to simplify:
```kotlin
if (parser != null) { ... } 
else if (streamingMember != null) { ... }
if (streamingMember == null) { ... }
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: I'll fix the same as we decide for https://github.com/awslabs/smithy-rs/pull/1023/files#r795990687
Resolving this one in the meanwhile.

guymguym:
@david-perez I'm done rebasing and updating. Couple questions for you in the comments above. Thanks

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
crisidev: Yep, if it passes it can be removed from the failing list.

crisidev:
Wonderful work @guymguym!! @rcoh, when you have some time, would you mind reviewing this from a client codegen perspective? We'd like to merge this. There is still some work to do to have a good streaming experience, but this PR is paving the way!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
david-perez: @guymguym Thanks for calling this out. We still need to generate HTTP bindings for blob streaming (e.g. HTTP header bindings). This also led me to discover there's an error in Smithy that is now being fixed over at https://github.com/awslabs/smithy/issues/1075

I have pushed a new commit to this branch, [Move blob streaming handling to httpPayload binding generation](https://github.com/awslabs/smithy-rs/pull/1023/commits/d86443ee023e3685622df9f9c96971b7fed8eef8), that refactors this part and generates the stream handling when we handle `@httpPayload`.

Let me know if you agree with the fix or have any questions.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
david-perez: I manually verified the now-passing tests and they look good.

I also had a look at the remaining failing request streaming protocol tests and pushed a new commit to this branch, [Add blob streaming support for server request protocol tests](https://github.com/awslabs/smithy-rs/pull/1023/commits/0033f7e942b4d37beb6ceb671ce9b1c20749608c), to get rid of them.

The only remaining failing protocol test for streaming is `RestJsonStreamingTraitsRequireLengthWithBlob`, but that is due to us not setting `Content-Length` appropriately, so I've cut https://github.com/awslabs/smithy-rs/issues/1146 to track that.

Let me know if you agree with the fix or have any questions.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
david-perez: Not that I know of. Maybe @rcoh knows.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
guymguym: That's great @david-perez. Thank you for making it so easy!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
guymguym: Ah now I see. I think my problem was probably that the previous headers code that I used was not very accurate, but since your PR to implement server response headers that might have fixed this issue. I will test a bit later and report back if I still see any issues in the S3 server codegen, but it looks good to me. Thanks!

david-perez:
One last thought: given that we decided on using `aws_smithy_http::byte_stream::ByteStream`, I don't see any value in `aws_smithy_http_server` re-exporting `ByteStream`. We should just use `ByteStream` directly and work on https://github.com/awslabs/smithy-rs/issues/1125 to hide `SdkBody` from its public API.

guymguym:
done @david-perez

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -511,16 +534,11 @@ class ServerProtocolTestGenerator(
...
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithBlob", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithNoBlobBody", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithBlob", Action.Response),
-            FailingTest(RestJson, "RestJsonStreamingTraitsWithNoBlobBody", Action.Response),
-            FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithBlob", Action.Request),
-            FailingTest(RestJson, "RestJsonStreamingTraitsRequireLengthWithNoBlobBody", Action.Request),
+            FailingTest(RestJson, "RestJsonSupportsNaNFloatInputs", Action.Response),
+            FailingTest(RestJson, "RestJsonSimpleScalarProperties", Action.Response),
+            FailingTest(RestJson, "RestJsonSupportsInfinityFloatInputs", Action.Response),
+            FailingTest(RestJson, "RestJsonSupportsNegativeInfinityFloatInputs", Action.Response),
```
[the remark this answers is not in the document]
david-perez: Yes, earlier today I had pushed to this branch a commit merging from `main` https://github.com/awslabs/smithy-rs/commit/907c0f3ff9c10c44d946d25a5af5069beab45e43 (which removes these now-passing tests), so that you wouldn't have to deal with conflicts.

But you didn't fetch it (?), and you force-pushed the latest commit.

In general once a PR is open I _think_ it's best to _not_ rebase and force-push, only merge from main. Because AFAIK GitHub's UI won't show the previous history, only the rewrriten one, and that might confuse other people looking at the PR. The downside is that merging makes the history non-linear, but we always squash before merging, so I think it's a small price to pay for having an entire's PR history.

---

Anyway, these tests can be removed from the failing list.

rcoh (approved):
Approved on the client side

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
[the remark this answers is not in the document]
guymguym: got it. thanks. pushed a commit to remove those.

david-perez:
@guymguym Thank you!

PR #1029, opened 2022-01-04
`aws-smithy-http-server`: don't ignore empty path segments when routing

david-perez:
In #996 [0], we realized that we were ignoring empty URI path segments
when doing label extraction. It turns out we are also ignoring them when
doing routing, as the TypeScript sSDK currently does [1], from which the
initial implementation was copied.

However, a discussion with the Smithy team in awslabs/smithy#1024 [2]
revealed that we _must not_ ignore empty URI path segments when routing
or doing label extraction, since empty strings (`""`) should be assigned
to the labels in those segments.

This commit fixes the behavior so that we don't ignore empty path
segments _when doing routing_. #996 will take care of fixing the behavior
when doing label extraction.

[0]: https://github.com/awslabs/smithy-rs/pull/996#discussion_r772624998
[1]: https://github.com/awslabs/smithy-typescript/blob/d263078b81485a6a2013d243639c0c680343ff47/smithy-typescript-ssdk-libs/server-common/src/httpbinding/mux.ts#L78
[2]: https://github.com/awslabs/smithy/issues/1024

## Testing

Test suite has been expanded.

---


crisidev (approved):
LGTM

PR #1031, opened 2022-01-04
Fix sync tool bug

Velfi:
## Motivation and Context

found another issue where I was setting last synced commit twice. The second time it was getting set, it was getting set to the latest smithy-rs commit instead of the actual last-synced commit

## Description

delete the sync step leftover from before we started limiting max commits

## Testing

I ran it locally and looked at the output


Velfi:
let me cherry pick this instead

PR #1036, opened 2022-01-04
`aws-smithy-http-server`: implement basic URI pattern conflict disambiguation

### LINKED ISSUE #1009: [Server] Unreachable routes in operation_registry when only httpLabels are different
When generating codegen-server for S3, operation_registry has list_buckets and list_objects operations and the only difference in their routing is that list_objects specifies an http label to capture the bucket name from the uri path, whereas list_buckets specifies no labels. However the generated order (alphabetical) puts the list_buckets route first, which hides the second route altogether. It seems like any subsequent route is unreachable because the route of list_buckets matches any uri path, while it should match just for root path `/`.

This might require a fix in request_spec.rs where it builds the path regex:
https://github.com/awslabs/smithy-rs/blob/cbd61ab1fff94775e112f78181653fe040d08e24/rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs#L78-L91

I tried to restrict it to match the entire path by changing the last line to the following code, which seems to help in some cases but I'm not sure this is correct - `Regex::new(&format!("^/*{}$", re)).unwrap()` (I had to use `/*` instead of `/+` because it didn't work otherwise but I'm quite puzzled as to why).

This issue might be related to existing PR #996 Implement httpLabel with nom - so worth noting this issue when working on it.

For reference here are the router specs for list_buckets and list_objects which is only different in having a path label -

```json
        "com.amazonaws.s3#ListBuckets": {
            "type": "operation",
            "output": {
                "target": "com.amazonaws.s3#ListBucketsOutput"
            },
            "traits": {
                "smithy.api#documentation": "<p>Returns a list of all buckets owned by the authenticated sender of the request.</p>",
                "smithy.api#http": {
                    "method": "GET",
                    "uri": "/",
                    "code": 200
                }
            }
        },

        "com.amazonaws.s3#ListObjects": {
            "type": "operation",
            "input": {
                "target": "com.amazonaws.s3#ListObjectsRequest"
            },
            "output": {
                "target": "com.amazonaws.s3#ListObjectsOutput"
            },
            "errors": [
                {
                    "target": "com.amazonaws.s3#NoSuchBucket"
                }
            ],
            "traits": {
                "smithy.api#documentation": "...",
                "smithy.api#http": {
                    "method": "GET",
                    "uri": "/{Bucket}",
                    "code": 200
                }
            }
        },

```

CC @crisidev

david-perez:
This commit implements basic URI pattern conflict disambiguation when
routing incoming requests to service operations.

It does this by introducing a measure of how "important" a `RequestSpec`
is. The more specific a `RequestSpec` is, the higher it ranks in
importance. Specificity is measured by the number of segments plus the
number of query string literals in its URI pattern, so
`/{Bucket}/{Key}?query` is more specific than `/{Bucket}/{Key}`, which
is more specific than `/{Bucket}`, which is more specific than `/`.

Why do we need this?

Note that:

1. the Smithy spec does not define how servers should route incoming
   requests in the case of pattern conflicts; and
2. the Smithy spec even outright rejects conflicting patterns that can
   be easily disambiguated e.g. `/{a}` and `/{label}/b` cannot coexist.

We can't to anything about (2) since the Smithy CLI will refuse to build
a model with those kind of conflicts. However, the Smithy CLI does allow
_other_ conflicting patterns to coexist, e.g. `/` and `/{label}`. We
therefore have to take a stance on (1), since if we route arbitrarily we
render basic usage impossible (see issue #1009).

So this ranking of routes implements some basic pattern conflict
disambiguation with some common sense. It's also the same behavior that
the TypeScript sSDK is implementing [0].

This commit also:

* makes the `future` module private,
* hides documentation for the `request_spec` module; and
* makes some fields from some structs in the `request_spec` module
  private and adds constructors.

Closes #1009.

[1]: https://github.com/awslabs/smithy-typescript/blob/d263078b81485a6a2013d243639c0c680343ff47/smithy-typescript-ssdk-libs/server-common/src/httpbinding/mux.ts#L59.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationRegistryGenerator.kt
```diff
@@ -216,18 +215,17 @@ class ServerOperationRegistryGenerator(
...
-                val routes = requestSpecsVarNames.zip(operationNames) { requestSpecVarName, operationName ->
-                    ".route($requestSpecVarName, #{ServerOperationHandler}::operation(registry.$operationName))"
-                }.joinToString(separator = "\n")
-
                 val requestSpecs = requestSpecsVarNames.zip(operations) { requestSpecVarName, operation ->
                     "let $requestSpecVarName = ${operation.requestSpec()};"
                 }.joinToString(separator = "\n")
+                val towerServices = requestSpecsVarNames.zip(operationNames) { requestSpecVarName, operationName ->
+                    "(#{Tower}::util::BoxCloneService::new(#{ServerOperationHandler}::operation(registry.$operationName)), $requestSpecVarName)"
+                }.joinToString(prefix = "vec![", separator = ",\n", postfix = "]")
```
crisidev, line 223: I am not sure I understand the need behind cloning the boxed service here..

crisidev (approved):
LGTM

PR #1037, opened 2022-01-05
Explicitly denylist failing server protocol tests

david-perez:
We are currently ignoring all S3 and REST JSON protocol tests.
Explicitly denylisting each failing test is slightly better since the
build will fail if we accidentally fix a denylisted test, or regress and
make a currently-passing test fail.


rcoh (approved):

crisidev (commented):
LGTM.

PR #1041, opened 2022-01-05
Add RFC for SDK example consolidation

jdisanti:
This is an RFC outlining the changes to make [`awsdocs/aws-doc-sdk-examples`](https://github.com/awsdocs/aws-doc-sdk-examples) the source of truth for the SDK's examples.

[Rendered.](https://github.com/awslabs/smithy-rs/blob/jdisanti-rfc-examples-consolidation/design/src/rfcs/rfc0008_example_consolidation.md)


rcoh:
Can you elaborate on your proposed workflow to fix examples? I guess I can point to `next` for testing?

design/src/rfcs/rfc0008_example_consolidation.md
```diff
@@ -0,0 +1,87 @@
...
+There are a couple of risks with this approach:
+
+1. **Risk:** Examples are broken and an urgent fix needs to be released.
+
+   **Possible mitigations:**
+
+     1. Revert the change that broke the examples and then add the urgent fix
+     2. Create a patch branch in `aws-sdk-rust`, apply the fix to that based off an older
+        version of `smithy-rs` with the fix applied, and merge that into `main`.
+
```
rcoh, line 68: this is a great callout in general—we should make sure we have a way of quickly getting a minimal fix out.

design/src/rfcs/rfc0008_example_consolidation.md
```diff
@@ -0,0 +1,87 @@
...
+2. **Risk:** A larger project requires changes to examples prior to GA, but multiple releases
+   need to occur before the project completion.
+
+   **Possible mitigations:**
+
+     1. If the required changes compile against the older SDK, then just make the changes
+        to the examples.
+     2. Feature gate any incremental new functionality in `smithy-rs`, and work on example
+        changes on a branch in `aws-doc-sdk-examples`. When wrapping up the project,
+        remove the feature gating and merge the examples into the `main` branch.
```
rcoh, line 78: it would be good to have some alternatives discussed as well–one downside with the current approach is the fact that the examples can't be PR'd which isn't ideal. It would be good to at least consider how things would look if it was reversed with examples in aws-sdk-rust as the source of truth, sync'd into the doc examples repo.

rcoh (approved):
Overall looks good—would like to see (at least for posterities sake) the other options that exist and the pros/cons.

design/src/rfcs/rfc0008_example_consolidation.md
jdisanti: I don't think the RFC (as written) impacts our ability to do patch releases. We could run into situations where we need to patch an example rather than the SDK, but that seems like an unlikely event that we should be able to handle when it comes up.

rcoh (approved):
LGTM

PR #1042, opened 2022-01-05
Fix: strange commit behavior in sync tool

Velfi:
## Motivation and Context

Previously, the sync tool relied on git2 to create git commits. For an unknown reason (i.e. my lack of knowledge regarding git porcelain) extra changes would be staged that weren't desired. This change sidesteps that by shelling out to the git CLI.

## Description

update: create mirror commit now works by shelling out
update: rm -rf instead of gradle clean
update: find_last_commit to work by shelling out

## Testing

ran locally and verified output


jdisanti (approved):
LGTM!

rcoh (approved):
LGTM, you may want to add a comment to the function explaining why most of the code uses git2 but this function does not

PR #1049, opened 2022-01-07
Fix bug in publisher topological sort

rcoh:
## Motivation and Context
The publisher script had a bug where it erroneously reported loops when they did not exist (and failed to generate a dag as a result).

## Description
- Create failing test case based on a minimized version of the real life failure
- fix topological sort algorithm (port algorithm from https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search)
- test passes 🎉
 
## Testing
- [x] new UT
- [x] it works on the crate graph


Velfi (approved):

tools/publisher/src/sort.rs
```diff
@@ -55,21 +55,19 @@ fn dependency_order_visit(
...
-        }
-        if package_handle != dependency
-            && packages.contains_key(dependency)
-            && !visited.contains(dependency)
-        {
-            dependency_order_visit(dependency, packages, stack, visited, result)?;
-        }
+        dependency_order_visit(dependency, packages, stack, visited, result)?;
     }
+    stack.remove(&package_handle);
```
Velfi, line 69:
```suggestion
    stack.remove(package_handle);
```
for clippy

PR #1050, opened 2022-01-07
Deserialize empty collections from the query string into `None`

david-perez:
We were instead providing empty collections e.g. `Some(Vec::new())`,
which makes protocol tests fail.

This commit also passes Smithy member names through the symbol
provider to obtain safe variable names.

This commit also adds some plumbing to the server protocol test case
generation to fix broken request test definitions in the
`awslabs/smithy` project.

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
@@ -7,6 +7,7 @@ package software.amazon.smithy.rust.codegen.server.smithy.generators.protocol
 
 import software.amazon.smithy.codegen.core.CodegenException
 import software.amazon.smithy.model.knowledge.OperationIndex
+import software.amazon.smithy.model.node.Node
 import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.model.shapes.StructureShape
 import software.amazon.smithy.model.traits.ErrorTrait
@@ -95,7 +96,7 @@ class ServerProtocolTestGenerator(
             val testCases = error.getTrait<HttpResponseTestsTrait>()?.testCases.orEmpty()
             testCases.map { TestCase.ResponseTest(it, error) }
         }
-        val allTests: List<TestCase> = (requestTests + responseTests + errorTests).filterMatching()
+        val allTests: List<TestCase> = (requestTests + responseTests + errorTests).filterMatching().fixBroken()
 
         if (allTests.isNotEmpty()) {
             val operationName = operationSymbol.name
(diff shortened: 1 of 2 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -138,6 +139,21 @@ class ServerProtocolTestGenerator(
...
+                it
+            } else {
+                val fixed = howToFixIt(it.testCase)
+                TestCase.RequestTest(fixed, it.targetShape)
+            }
+        }
+        is TestCase.ResponseTest -> {
+            it
+        }
+    } }
```
crisidev, line 155: Probably a comment would help here. It is clear in the PR, but it will be come harder when it is just part of a big kotlin source file like this one.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -549,5 +567,51 @@ class ServerProtocolTestGenerator(
...
+            val params = Node.parse("""{
+                "queryFloat": "NaN",
+                "queryDouble": "NaN",
+                "queryParamsMapOfStringList": {
+                    "Float": ["NaN"],
+                    "Double": ["NaN"]
+                }
+            }""".trimIndent()).asObjectNode().get()
+
+            return testCase.toBuilder().params(params).build()
```
crisidev, line 582: I am not sure why are you doing this if it doesn't work. Is the comment outdated?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -810,9 +811,16 @@ private class ServerHttpProtocolImplGenerator(
                 rust("input = input.${queryParamsBinding.member.setterName()}(Some(query_params));")
             }
             queryBindingsTargettingCollection.forEach {
+                val memberName = symbolProvider.toMemberName(it.member)
                 rustTemplate(
                     """
-                    input = input.${it.member.setterName()}(Some(${it.memberName.toSnakeCase()}));
+                    input = input.${it.member.setterName()}(
```
crisidev, line 817: Good catch!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
david-perez: Clarified the comment.

PR #1051, opened 2022-01-07
Add Support for SSO

rcoh:
This commit adds support for the SSO credential provider, which enables the aws-config to support using SSO when specified in `~/.aws/config`.

## Motivation and Context
https://github.com/awslabs/aws-sdk-rust/issues/4

## Description
This adds support for SSO, both as a standalone credentials provider, and with AWS config. 
## Testing
- [x] http traffic test
- [x] assorted UTs
- [x] profile tests  

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/json_credentials.rs
+++ b/aws/rust-runtime/aws-config/src/json_credentials.rs
@@ -20,6 +20,12 @@ pub(crate) enum InvalidJsonCredentials {
     /// The response was missing a required field
     MissingField(&'static str),
 
+    /// A field was invalid
+    InvalidField {
+        field: &'static str,
+        err: Box<dyn Error + Send + Sync>,
+    },
+
     /// Another unhandled error occurred
     Other(Cow<'static, str>),
 }
@@ -48,6 +54,9 @@ impl Display for InvalidJsonCredentials {
                 field
             ),
             InvalidJsonCredentials::Other(msg) => write!(f, "{}", msg),
+            InvalidJsonCredentials::InvalidField { field, err } => {
+                write!(f, "Invalid field in response: `{}`. {}", field, err)
+            }
         }
     }
 }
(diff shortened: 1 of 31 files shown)


aws/rust-runtime/aws-config/src/json_credentials.rs
```diff
@@ -197,6 +175,42 @@ pub(crate) fn parse_json_credentials(
...
+                if let Some(Ok(Token::ValueString { value, .. })) = tokens.peek() {
+                    let key = key.to_unescaped()?;
+                    let value = value.to_unescaped()?;
+                    f(key, value)
+                }
+                skip_value(&mut tokens)?;
+            }
+            other => {
+                return Err(InvalidJsonCredentials::Other(
+                    format!("expected object key, found: {:?}", other,).into(),
```
jdisanti, line 201:
```suggestion
                    format!("expected object key, found: {:?}", other).into(),
```

aws/rust-runtime/aws-config/src/sso.rs
```diff
@@ -0,0 +1,432 @@
...
+        self
+    }
+
+    /// Set the role name used for SSO
+    pub fn role_name(mut self, role_name: impl Into<String>) -> Self {
+        self.role_name = Some(role_name.into());
+        self
+    }
+
+    /// Set the start url used for SSO
```
jdisanti, line 127:
```suggestion
    /// Set the start URL used for SSO
```

aws/rust-runtime/aws-config/src/sso.rs
```diff
@@ -0,0 +1,432 @@
...
+    IoError(io::Error),
+}
+
+impl Display for LoadTokenError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            LoadTokenError::InvalidCredentials(err) => write!(f, "Token was invalid: {}", err),
+            LoadTokenError::NoHomeDirectory => write!(f, "Could not resolve a home directory"),
+            LoadTokenError::IoError(io_failure) => {
+                write!(f, "failed to read the token file: {}", io_failure)
```
jdisanti, line 172: Should this include the file name for context?

aws/rust-runtime/aws-config/src/sso.rs
```diff
@@ -0,0 +1,432 @@
...
+pub(crate) enum LoadTokenError {
+    InvalidCredentials(InvalidJsonCredentials),
+    NoHomeDirectory,
+    IoError(io::Error),
+}
+
+impl Display for LoadTokenError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            LoadTokenError::InvalidCredentials(err) => write!(f, "Token was invalid: {}", err),
```
jdisanti, line 169: Should this say "SSO Token"?

aws/rust-runtime/aws-config/src/sso.rs
```diff
@@ -0,0 +1,432 @@
...
+        .read_to_end(path)
+        .await
+        .map_err(LoadTokenError::IoError)?;
+    let token = parse_token_json(&data).map_err(LoadTokenError::InvalidCredentials)?;
+    Ok(token)
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct SsoToken {
+    access_token: String,
```
jdisanti, line 263: Should this be zeroized and excluded from debug output?

jdisanti (commented):
Overall looks great! I just have some questions and some minor fixes.

jdisanti (approved):
Awesome!

naftulikay:
Excited to see this passed! This will certainly help in building applications which leverage SSO.

I'm trying to find out how the AWS CLI does `aws sso login` to see if I can implement that logic in Rust, as the flow I'd like to emulate is:
  1. If cached credentials that are not expired exist for the given profile, use them, use STS to get role credentials, proceed.
  2. If cached credentials do not exist or are expired, optionally implement the browser launch flow to renew credentials, then proceed.

My goal is naturally that for a given AWS utility, I want it to _work_, whether credentials are coming from environment variables, from the credentials file(s), from an SSO configuration, or from the EC2 metadata server. Does anyone know enough to point me to where `aws sso login` is implemented in the Python AWS CLI? I have dug around and the functionality has _not_ been easy to find.

rcoh:
I ran `aws sso login --debug`, which lead to this callstack:
```
Traceback (most recent call last):
  File "awscli/clidriver.py", line 457, in main
  File "awscli/clidriver.py", line 586, in __call__
  File "awscli/customizations/commands.py", line 191, in __call__
  File "awscli/customizations/sso/login.py", line 40, in _run_main
  File "awscli/customizations/sso/login.py", line 64, in _get_sso_config
```

this code was hard to track down! I eventually found it here: https://fossies.org/linux/www/aws-cli-2.4.10.tar.gz/

naftulikay:
After loads of reverse engineering and dropping debug points, it's all handled in the `aws_sdk_ssooidc` crate APIs. The flow looks like this:

### Register your Client

First, call [`RegisterClient`](https://docs.rs/aws-sdk-ssooidc/0.4.1/aws_sdk_ssooidc/client/fluent_builders/struct.RegisterClient.html) with a client name of your choice and a client type of `public`. This will return a [`RegisterClientOutput`](https://docs.rs/aws-sdk-ssooidc/0.4.1/aws_sdk_ssooidc/output/struct.RegisterClientOutput.html), which contains a couple things you'll need to care about, the `client_id` and the `client_secret`. The `client_secret_expires_at` is useful if you're going to cache these credentials. `botocore` caches these intermediate credentials in JSON at `~/.aws/sso/cache/botocore-client-id-${AWS_REGION}.json` if you'd like to take a look.

### Start Device Authorization

Next, call [`StartDeviceAuthorization`](https://docs.rs/aws-sdk-ssooidc/0.4.1/aws_sdk_ssooidc/client/fluent_builders/struct.StartDeviceAuthorization.html) with the `client_id`, `client_secret`, and the SSO start URL from your SSO configuration (usually something like `https://d-${SOMETHING}.awsapps.com/start`. You'll get a `device_code`, `user_code`, `verification_uri`, and `verification_uri_complete`. An `expires_in` will also be returned demonstrating when _these_ credentials expire. Finally, save the `interval` value, as you'll use this interval to respectfully poll for token creation.

You're now ready to launch the browser.

### Launch the Browser

Launch your browser to `verification_uri_complete`, and it's time to start polling to await completion of the user login flow.

### Create and Obtain the Token

In a loop, and respectfully, attempt to [`CreateToken`](https://docs.rs/aws-sdk-ssooidc/0.4.1/aws_sdk_ssooidc/client/fluent_builders/struct.CreateToken.html) by passing in `urn:ietf:params:oauth:grant-type:device-code` as the `grant_type`, pass in your `client_id` and `client_secret` from before, and finally pass the `device_code`. 

You'll need to match on error responses, because [`CreateTokenErrorKind`](https://docs.rs/aws-sdk-ssooidc/0.4.1/aws_sdk_ssooidc/error/enum.CreateTokenErrorKind.html) has a lot of variants that you care about, namely `AuthorizationPendingException`, which means that you should keep polling, `SlowDownException` which means you should poll less often, and the rest of them largely mean that the request failed.

### Caching the Token

You'll end up with a [`CreateTokenOutput`](https://docs.rs/aws-sdk-ssooidc/0.4.1/aws_sdk_ssooidc/output/struct.CreateTokenOutput.html), which you can then use for all the rest of the magic you'll need to do, see my repository for the additional steps needed: [naftulikay/aws-sso-env](https://github.com/naftulikay/aws-sso-env).

The `CreateTokenOutput` will contain an `access_token` and an `expires_in` field to denote how long the token is valid for. To cache this value and play nicely with other tools, render it to `~/.aws/sso/cache/$(shasum "$SSO_START_URL").json` in the following JSON format:

```json
[
 {
    "startUrl": "$START_URL",
    "region": "$REGION",
    "accessToken": "$ACCESS_TOKEN",
    "expiresAt": "$RFC3339_EXPIRY_TIMESTAMP"
 }
]
```

### Obtaining Actual IAM Credentials

Now that you've gone through all of this, you can use the access token with `aws_sdk_sso` to get role credentials:

```rust
use aws_sdk_sso::{Client as SsoClient};
use aws_sdk_sso::model::RoleCredentials;

let client = SsoClient::from_conf(config);

let role_credentials: RoleCredentials = client
    .get_role_credentials()
    .account_id(sso_account_id)
    .role_name(sso_role_name)
    .access_token(access_token)
    .send()
    .await?
    .role_credentials
    .unwrap();
```

This was a fun little deep-dive for me and hopefully it helps someone else as well. I'm not too familiar with how OIDC works, but I think I'm starting to understand a little bit.

sauhaardac:
@naftulikay You are the hero we need but do not deserve... thank you so much. This is actually saving my company right now!

naftulikay:
o7

If only I had the time to tinker with it for my own registry 😂 This is one of the first times I have written code that didn't function at the time and then just started working at some later date.

PR #1058, opened 2022-01-11
`rust-server-codegen`: fix deserialization of escaped query string data

david-perez:
We are currently deserializing the query string into `Vec<(&str,
&str)>`. `serde_urlencoded` panics if the input string slice contains
escaped data, since in that case it needs to allocate a new `String` to
unescape the input string slice's contents.

Instead of deserializing to `Vec<(String, String)>`, we can instead use
`Cow<'a, str>` so that deserialization only allocates when strictly
required.

Reference: https://github.com/serde-rs/serde/issues/1413#issuecomment-494892266


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -103,6 +104,7 @@ private class ServerHttpProtocolImplGenerator(
     private val codegenScope = arrayOf(
         "AsyncTrait" to ServerCargoDependency.AsyncTrait.asType(),
         "AxumCore" to ServerCargoDependency.AxumCore.asType(),
+        "Cow" to ServerRuntimeType.Cow,
```
crisidev, line 107: I thought `Cow` was also part of `RuntimeType` but it is not..

crisidev (approved):
LGTM

PR #1064, opened 2022-01-12
update: client input links to point to fluent builder

Velfi:
## Motivation and Context

This is more helpful than pointing to an operation's Input struct

## Description

update operation input doc links generated for `client::Client` to point to the fluent builder and related methods

## Testing

I built the docs locally and read them

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
@@ -133,10 +133,12 @@ fun RustType.qualifiedName(): String {
     return "$namespace$name"
 }
 
+/** Format this Rust type as an `impl Into<T>` */
 fun RustType.implInto(fullyQualified: Boolean = true): String {
     return "impl Into<${this.render(fullyQualified)}>"
 }
 
+/** Format this Rust type so that it may be used as an argument type in a function definition */
 fun RustType.asArgumentType(fullyQualified: Boolean = true): String {
     return when (this) {
         is RustType.String,
@@ -145,6 +147,10 @@ fun RustType.asArgumentType(fullyQualified: Boolean = true): String {
     }
 }
 
+/**
+ * For a given name, generate an `Argument` data class containing pre-formatted strings for using this type when
+ * writing a Rust function
+ */
 fun RustType.asArgument(name: String): Argument {
     return Argument(
         "$name: ${this.asArgumentType()}",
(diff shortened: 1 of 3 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/FluentClientDecorator.kt
```diff
@@ -563,6 +562,30 @@ class FluentClientGenerator(
...
+    val fluentBuilderFullyQualifiedName = "crate::client::fluent_builders::$operationName"
+    return input.members().map { member ->
+        val memberName = symbolProvider.toMemberName(member)
+        val memberSymbol = symbolProvider.toSymbol(member)
+        val outerType = memberSymbol.rustType()
+
+        val builderInput = "$memberName(${outerType.stripOuter<RustType.Option>().render(fullyQualified = false)})"
+        val builderInputLink = "$fluentBuilderFullyQualifiedName::$memberName"
+        val builderSetter = "${member.setterName()}(${outerType.render(fullyQualified = false)})"
+        val builderSetterLink = "$fluentBuilderFullyQualifiedName::${member.setterName()}"
```
jdisanti, line 577: I think there's a lot of potential for this to break, for instance, if we modify the codegen for the actual functions but forget that these docs exist also. Consider pulling out some helper functions for this in the function codegen, and then calling those same functions here to get these.

Velfi: @jdisanti I updated things a bit, can you let me know if you like it more?

Velfi:
My refactors broke how bools get passed into setters, I'm looking into it

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
```diff
@@ -364,4 +381,4 @@ sealed class Attribute {
     }
 }
 
-data class Argument(val argument: String, val value: String)
+data class Argument(val argument: String, val value: String, val type: String)
```
jdisanti, line 384: I think an argument could be made to store the `RustType` rather than a `String` for `type`, but it's not needed right now and not blocking.

jdisanti (approved):
LGTM!

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
Velfi: I didn't do that because I wasn't sure how to represent `into Impl` as a rust type

PR #1065, opened 2022-01-12
fix: silence warnings in Lambda environment

nmoutschen:
## Motivation and Context

This silence some unnecessary warnings in a Lambda environment (ref: https://github.com/awslabs/aws-sdk-rust/issues/398).

## Description

* Silence warnings in `aws_config::profile::parser::timeout_config` and `aws_config::profile::parser::retry_config` when running in a Lambda function.
* Move `is_likely_running_on_a_lambda` function from `aws_config::profile::parser::source` to `aws_config::profile::parser`.

## Testing

I have a temporary issue setting the test environment, so I couldn't test this right now. See the error message below when running `./gradlew`:

```
A problem occurred configuring project ':buildSrc'.
> Could not resolve all artifacts for configuration ':buildSrc:classpath'.
   > Could not resolve org.gradle.kotlin:plugins:1.3.6.
     Required by:
         project :buildSrc > org.gradle.kotlin.kotlin-dsl:org.gradle.kotlin.kotlin-dsl.gradle.plugin:1.3.6
      > Could not resolve org.gradle.kotlin:plugins:1.3.6.
         > Could not get resource 'https://plugins.gradle.org/m2/org/gradle/kotlin/plugins/1.3.6/plugins-1.3.6.module'.
            > Could not GET 'https://jcenter.bintray.com/org/gradle/kotlin/plugins/1.3.6/plugins-1.3.6.module'. Received status code 502 from server: Bad Gatewa
```

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/profile/mod.rs
+++ b/aws/rust-runtime/aws-config/src/profile/mod.rs
@@ -22,3 +22,11 @@ pub mod timeout_config;
 pub use credentials::ProfileFileCredentialsProvider;
 #[doc(inline)]
 pub use region::ProfileFileRegionProvider;
+
+/// Returns true or false based on whether or not this code is likely running inside an AWS Lambda.
+/// [Lambdas set many environment variables](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime)
+/// that we can check.
+fn check_is_likely_running_on_a_lambda(environment: &os_shim_internal::Env) -> bool {
+    // LAMBDA_TASK_ROOT – The path to your Lambda function code.
+    environment.get("LAMBDA_TASK_ROOT").is_ok()
+}
--- a/aws/rust-runtime/aws-config/src/profile/mod.rs
+++ b/aws/rust-runtime/aws-config/src/profile/mod.rs
@@ -26,7 +26,7 @@ pub use region::ProfileFileRegionProvider;
 /// Returns true or false based on whether or not this code is likely running inside an AWS Lambda.
 /// [Lambdas set many environment variables](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime)
 /// that we can check.
-fn check_is_likely_running_on_a_lambda(environment: &os_shim_internal::Env) -> bool {
+fn check_is_likely_running_on_a_lambda(environment: &aws_types::os_shim_internal::Env) -> bool {
     // LAMBDA_TASK_ROOT – The path to your Lambda function code.
     environment.get("LAMBDA_TASK_ROOT").is_ok()
 }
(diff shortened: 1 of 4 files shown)


aws/rust-runtime/aws-config/src/profile/retry_config.rs
```diff
@@ -106,7 +107,13 @@ impl ProfileFileRetryConfigProvider {
         let selected_profile = match profile.get_profile(selected_profile) {
             Some(profile) => profile,
             None => {
-                tracing::warn!("failed to get selected '{}' profile", selected_profile);
+                // Lambdas don't have home directories and emitting this warning is not helpful
+                // to users running the SDK from within a Lambda. This warning will be silenced
+                // if we determine that that is the case.
+                let is_likely_running_on_a_lambda = check_is_likely_running_on_a_lambda(self);
```
jdisanti, line 113: I think this is actually a bug for non-Lambda use-cases too. For example, if I'm not using profile credentials on EC2, it's still going to log this warning. To fix this everywhere, I think we should instead check to see if `self.profile_override.is_some()`. That is: if a customer specified a profile name to use, then we should warn if that profile isn't found.

jdisanti (commented):
Thank you for the PR! :heart:

I think we should move the INFO messages to the DEBUG level in addition to fixing the WARN. Let us know if you want to make these fixes, or if you would like us to take over. Thanks!

nmoutschen:
Thanks for the quick review @jdisanti ! I'll implement the proposed changes (both INFO -> DEBUG and profile_override) tomorrow. :)

aws/rust-runtime/aws-config/src/profile/timeout_config.rs
```diff
@@ -117,10 +117,16 @@ impl ProfileFileTimeoutConfigProvider {
...
             None => {
-                tracing::warn!(
-                    "failed to get selected '{}' profile, skipping it",
-                    selected_profile
-                );
+                // Lambdas don't have home directories and emitting this warning is not helpful
+                // to users running the SDK from within a Lambda. This warning will be silenced
+                // if we determine that that is the case.
+                let is_likely_running_on_a_lambda = check_is_likely_running_on_a_lambda(&self.env);
+                if !is_likely_running_on_a_lambda {
```
jdisanti, line 124: This should behave the same as in `retry_config.rs` where it checks if the `profile_override` is set.

aws/rust-runtime/aws-config/src/profile/mod.rs
```diff
@@ -22,3 +22,11 @@ pub mod timeout_config;
 pub use credentials::ProfileFileCredentialsProvider;
 #[doc(inline)]
 pub use region::ProfileFileRegionProvider;
+
+/// Returns true or false based on whether or not this code is likely running inside an AWS Lambda.
+/// [Lambdas set many environment variables](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime)
+/// that we can check.
+fn check_is_likely_running_on_a_lambda(environment: &aws_types::os_shim_internal::Env) -> bool {
```
jdisanti, line 29: This can move back into `profile/parser/source.rs`.

jdisanti (approved):
This looks great! Thank you!

PR #1066, opened 2022-01-12
Add headers to the requests in the server protocol tests

### LINKED ISSUE #1062: Add input headers to requests in server protocol compliance tests
In the requests we're generating, we're not adding the input headers from the test case. Example:

```rust
async fn rest_json_input_and_output_with_quoted_string_headers_request() {
    let expected = crate::input::InputAndOutputWithHeadersInput::builder()  » Builder
        .set_header_string_list(Some(vec![
            "b,c".to_string(),
            "\"def\"".to_string(),
            "a".to_string(),
        ]))  » Builder
        .build()  » Result<InputAndOutputWithHeadersInput, …>
        .unwrap();
    #[allow(unused_mut)]
    let mut http_request = http::Request::builder()  » Builder
        .uri("/InputAndOutputWithHeaders")  » Builder
        .body(aws_smithy_http_server::Body::from(
            bytes::Bytes::from_static(b""),
        ))  » Result<Request<Body>, Error>
        .unwrap();
    let expected_headers = [("X-StringList", "\"b,c\", \"\\\"def\\\"\", a")];
    aws_smithy_protocol_test::assert_ok(aws_smithy_protocol_test::validate_headers(
        &http_request,
        expected_headers,
    ));
    ...
}
```

Note how the `X-StringList` header is never added to `http_request`, so the assertion fails.

david-perez:
Closes #1062.

### DIFF
--- a/codegen-server-test/model/simple.smithy
+++ b/codegen-server-test/model/simple.smithy
@@ -50,7 +50,9 @@ resource Service {
         id: "RegisterServiceRequestTest",
         protocol: "aws.protocols#restJson1",
         uri: "/service/1",
-        bodyMediaType: "application/json",
+        headers: {
+            "Content-Type": "application/json",
+        },
         params: { id: "1", name: "TestService" },
         body: "{\"name\":\"TestService\"}",
         method: "PUT",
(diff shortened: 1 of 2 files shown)


codegen-server-test/model/simple.smithy
```diff
@@ -50,7 +50,9 @@ resource Service {
         id: "RegisterServiceRequestTest",
         protocol: "aws.protocols#restJson1",
         uri: "/service/1",
-        bodyMediaType: "application/json",
+        headers: {
+            "Content-Type": "application/json",
+        },
```
crisidev, line 55: Ohhh this makes a lot of sense..

crisidev (approved):

david-perez:
PR became stale, I've rebased and moved all header tests at the beginning.

PR #1067, opened 2022-01-12
RFC: How Cargo "features" should be used in the SDK and runtime crates

Velfi:
This is an early version of the features RFC in case I get hit by a bus. Feel free to comment.


design/src/rfcs/rfc0009_using_features_responsibly.md
```diff
@@ -0,0 +1,164 @@
...
+  None
+}
+```
+
+While verbose, this code will compile without error no matter what features are active. It also makes the effect of activated features clearer to reason about.
+
+### Don't default to defining "default features"
+
+Because Cargo will use the union of all features enabled on a dependency when building it, we should be wary of marking features as default. Once we do mark features as default, users that want to exclude code and dependencies brought in by those features will have a difficult time doing so. One need look no further than [this issue][remove rustls from crate graph] submitted by a user that wanted to use Native TLS and struggled to make sure that Rustls was actually disabled *(This issue was resolved in [this PR][remove default features from runtime crates] which removed default features from our runtime crates.)* This is not to say that we should never use them, as having defaults for the most common use cases means less work for those users.
+
```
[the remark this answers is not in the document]
Velfi: good callout

design/src/rfcs/rfc00014_using_features_responsibly.md
```diff
@@ -0,0 +1,172 @@
...
+#[cfg(not(feature = "rt-tokio"))]
+pub fn default_async_sleep() -> Option<Arc<dyn AsyncSleep>> {
+    None
+}
+```
+
+In this example, code that calls the `default_async_sleep` function will continue to compile when the `rt-tokio` feature is inactive. This moves the responsibility of handling "missing" functionality out of the places where that functionality is called upon into the place where that functionality is defined. This results in less redundant code. However, once more than one mutually exclusive feature is added, it's better to use the [`cfg!` macro][cfg! macro] with an if-else statement:
+
+```rust
+pub fn default_async_sleep() -> Option<Arc<dyn AsyncSleep>> {
```
jdisanti, line 112: The "When a default feature providing some functionality is disabled, active features must not automatically replace that functionality" section below basically says don't do this, so maybe we just want to remove this example?

design/src/SUMMARY.md
```diff
@@ -27,4 +27,5 @@
   - [RFC-0011: Publishing Alpha to Crates.io](./rfcs/rfc0011_crates_io_alpha_publishing.md)
   - [RFC-0012: Independent Crate Versioning](./rfcs/rfc0012_independent_crate_versioning.md)
   - [RFC-0013: Body Callback APIs](./rfcs/rfc0013_body_callback_apis.md)
-  - [RFC-0014: How Cargo "features" should be used in the SDK and runtime crates](./rfcs/rfc00014_using_features_responsibly.md)
+  - [RFC-0013: Fine-grained timeout configuration](./rfcs/rfc00014_timeout_config.md)
+  - [RFC-0015: How Cargo "features" should be used in the SDK and runtime crates](./rfcs/rfc00014_using_features_responsibly.md)
```
jdisanti, line 31: Version mismatch here

jdisanti (approved):

design/src/SUMMARY.md
Velfi: good catch

PR #1069, opened 2022-01-13
Fix several warnings and reduce Rust target clobbering

jdisanti:
## Motivation and Context
The protocol test output had several warnings in it, one of which was called out in https://github.com/awslabs/smithy-rs/issues/1040. Additionally, local development has involved a lot of recompiling of Rust dependencies lately due to clobbering of the `target/` directory due to Gradle targets invoking Rust with different compiler flags.

This PR addresses both issues and stops short of disallowing warnings from Rustdoc, which isn't quite possible until we implement https://github.com/awslabs/smithy-rs/issues/1068.

These changes should also make the license header lint run faster since it will recompile the tool less frequently.

Note: One protocol test Rustdoc warning is left in place since it didn't seem worth the effort to fix right now. This is a warning about the `crate::model` module not existing. There's a doc link to this module, but the module doesn't get generated if there are no structs or enums that would be placed in it (which is determined by the Smithy model). We could fix this by always producing an empty module in the future.

## Testing
- Ran protocol tests and verified there is only one warning left that wasn't worth the time investment to fix
- Ran `codegen-test:check` followed by `aws:sdk:assemble` followed by `codegen-test:check` and verified that it didn't recompile all dependencies for the last `codegen-test:check` invocation.

## Checklist

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
@@ -305,6 +305,12 @@ fun Writable.isEmpty(): Boolean {
     return writer.toString() == RustWriter.root().toString()
 }
 
+/**
+ * Rustdoc doesn't support `r#` for raw identifiers.
+ * This function adjusts doc links to refer to raw identifiers directly.
+ */
+fun docLink(docLink: String): String = docLink.replace("::r##", "::").replace("::r#", "::")
+
 class RustWriter private constructor(
     private val filename: String,
     val namespace: String,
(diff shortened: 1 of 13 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
```diff
@@ -305,6 +305,12 @@ fun Writable.isEmpty(): Boolean {
     return writer.toString() == RustWriter.root().toString()
 }
 
+/**
+ * Rustdoc doesn't support `r#` for raw identifiers.
+ * This function adjusts doc links to refer to raw identifiers directly.
+ */
+fun docLink(docLink: String): String = docLink.replace("::r##", "::").replace("::r#", "::")
```
rcoh, line 312: ~there _may_ be other places where we make doc links that need to be audited (but maybe those aren't dynamic?)~

Nevermind, looks like you got em

rcoh (approved):

PR #1072, opened 2022-01-13
`rust-server-codegen`: add `ResponseExtensions` to non-fallible operations

### LINKED ISSUE #1063: Add `ResponseExtensions` to _all_ server responses
[We're only adding `RequestExtensions` to responses of fallible operations](https://github.com/awslabs/smithy-rs/blob/8fe1083faac510f385dba7e4aca5f63dc929c9b5/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt#L215).

We should also rename the struct to `ResponseExtensions`, since we only add them to responses, not requests.

david-perez:
This commit also renames `RequestExtensions` to `ResponseExtensions`,
since we add this extension type only to HTTP responses. It also makes
its members private.

Closes #1063.


crisidev (approved):
I am so happy to see all these protocol tests succeeding now!

PR #1078, opened 2022-01-14
Make psuedo jobs fail if dependency jobs failed in CI

jdisanti:
## Motivation and Context
The psuedo jobs are getting skipped when the jobs they depend on fail, and that's allowing merge with failing tests. This makes the psuedo jobs always run and check the status of the jobs they depend on.

## Testing
Testing as part of this PR by intentionally breaking one of the jobs the psuedo jobs depend on. Will revert this test break once it's verified.

### DIFF
--- a/.github/workflows/ci-sdk.yaml
+++ b/.github/workflows/ci-sdk.yaml
@@ -71,6 +71,8 @@ jobs:
           run: cargo +nightly udeps
         - name: Additional per-crate checks
           run: ../tools/additional-per-crate-checks.sh ./sdk/
+        - name: Failure test
+          run: exit 1
     env:
       # Disable incremental compilation to reduce disk space use
       CARGO_INCREMENTAL: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-sdk.yaml
```diff
@@ -71,6 +71,8 @@ jobs:
           run: cargo +nightly udeps
         - name: Additional per-crate checks
           run: ../tools/additional-per-crate-checks.sh ./sdk/
+        - name: Failure test
+          run: exit 1
```
crisidev, line 75: I am not sure I understand why this _have_ to fail..

crisidev (approved):
Nice!

.github/workflows/ci-sdk.yaml
jdisanti: I'm just testing failure with that 😄 Will remove it once I know it works.

rcoh (approved):

.github/workflows/ci-sdk.yaml
crisidev: Ah! Oook 😄

crisidev (approved):

PR #1079, opened 2022-01-14
Fix the canary Lambda build and add it to CI

jdisanti:
The canary is failing at build time in `aws-sdk-rust/next` due to some missing imports: https://github.com/awslabs/aws-sdk-rust/runs/4822595979?check_suite_focus=true

This PR fixes those failures, and also adds the canary to smithy-rs CI so that we detect these problems sooner.

### DIFF
--- a/tools/ci-cdk/canary-lambda/ci-check.sh
+++ b/tools/ci-cdk/canary-lambda/ci-check.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+#
+# Run by CI to check the canary-lambda
+set -e
+cd "$(dirname $0)"
+
+SDK_PATH="$(git rev-parse --show-toplevel)"/aws/sdk/build/aws-sdk/sdk
+if [[ "${GITHUB_ACTIONS}" == "true" ]]; then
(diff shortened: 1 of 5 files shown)


tools/ci-cdk/canary-lambda/ci-check.sh
```diff
@@ -0,0 +1,17 @@
...
+set -e
+cd "$(dirname $0)"
+
+SDK_PATH="$(git rev-parse --show-toplevel)"/aws/sdk/build/aws-sdk/sdk
+if [[ "${GITHUB_ACTIONS}" == "true" ]]; then
+   SDK_PATH="$(git rev-parse --show-toplevel)"/aws-sdk/sdk
+fi
+
+./write-cargo-toml.py --path "${SDK_PATH}"
+cargo check
```
rcoh, line 16: I don't think we need to run both? also, can this use `additional-ci` instead?

rcoh (approved):

tools/ci-cdk/canary-lambda/ci-check.sh
jdisanti: Yeah, I think you're right about not needing to run both. I will address this and attempt to consolidate it into `additional-ci` as a follow-up pull request to unblock the release.

jdisanti: Addressed in https://github.com/awslabs/smithy-rs/pull/1084

PR #1083, opened 2022-01-18
Fix paginator bug where `None` was returned immediately

rcoh:
## Motivation and Context
- https://github.com/awslabs/aws-sdk-rust/issues/405

## Description
The escape hatch added by aws-sdk-rust#391 did not properly handle the case where the first response was `None` and _not_ the empty string. This diff:
- Checks for emptiness for both maps and strings
- Fixes the check so that an initial `None` input does not cause an incorrect paginator error

## Testing
- [x] added to canary
- [x] added mocked-traffic IT
## Checklist

### DIFF
--- a/aws/sdk/integration-tests/ec2/tests/paginators.rs
+++ b/aws/sdk/integration-tests/ec2/tests/paginators.rs
@@ -50,3 +50,39 @@ async fn paginators_handle_empty_tokens() {
     assert_eq!(first_item, None);
     conn.assert_requests_match(&[]);
 }
+
+/// See https://github.com/awslabs/aws-sdk-rust/issues/405
+///
+/// EC2 can also reply with the token truly unset which will be interpreted as `None`
+#[tokio::test]
+async fn paginators_handle_unset_tokens() {
+    let request= "Action=DescribeSpotPriceHistory&Version=2016-11-15&AvailabilityZone=eu-north-1a&InstanceType.1=g5.48xlarge&ProductDescription.1=Linux%2FUNIX";
+    let response = r#"<?xml version="1.0" encoding="UTF-8"?>
+        <DescribeSpotPriceHistoryResponse xmlns="http://ec2.amazonaws.com/doc/2016-11-15/">
+            <requestId>edf3e86c-4baf-47c1-9228-9a5ea09542e8</requestId>
+            <spotPriceHistorySet/>
+        </DescribeSpotPriceHistoryResponse>"#;
+    let conn = TestConnection::<&str>::new(vec![(
+        http::Request::builder()
+            .uri("https://ec2.us-east-1.amazonaws.com/")
+            .body(request.into())
+            .unwrap(),
+        http::Response::builder()
+            .status(200)
+            .body(response)
+            .unwrap(),
+    )]);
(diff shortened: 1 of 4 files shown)


aws/sdk/integration-tests/ec2/tests/paginators.rs
```diff
@@ -50,3 +50,39 @@ async fn paginators_handle_empty_tokens() {
     assert_eq!(first_item, None);
     conn.assert_requests_match(&[]);
 }
+
+/// See https://github.com/awslabs/aws-sdk-rust/issues/405
+///
+/// EC2 can also reply with the token truly unset which will be interpreted as `None`
```
Velfi, line 56: What exactly does this mean? How many "kinds" of unset are there? I can think of
a. field returned but empty or set to a value signifying empty
b. no field returned

Were we only covering case a previously and now we're covering case b too?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -276,12 +275,7 @@ class PaginatorGenerator private constructor(
     }
 
     private fun nextTokenEmpty(token: String): String {
-        val tokenType = model.expectShape(paginationInfo.outputTokenMemberPath.last().target)
-        if (tokenType is StringShape) {
-            return "$token.as_deref().unwrap_or_default().is_empty()"
-        } else {
-            return "$token.is_none()"
-        }
+        return "$token.map(|token|token.is_empty()).unwrap_or(true)"
```
Velfi, line 278: I think we should prefer function pointers in place of closures where possible:
```suggestion
        return "$token.map(String::is_empty).unwrap_or(true)"
```
It might not be possible depending on how type inference interprets this though

tools/ci-cdk/canary-lambda/src/paginator_canary.rs
```diff
@@ -43,6 +43,15 @@ pub async fn paginator_canary(client: ec2::Client, page_size: usize) -> anyhow::
         bail!("should be ~60 of pages of results")
```
Velfi, line 43:
```suggestion
        bail!("expected 3+ pages containing ~60 results but got {} pages", num_pages)
```

Velfi (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
rcoh: this can be either a string or a hashmap

aws/sdk/integration-tests/ec2/tests/paginators.rs
rcoh: yeah it can either be `<nextToken></nextToken>`, or as is the case here, `<nextToken>` is not present in the response at all. In one case, we get `Some("")`, in the other, we get `None`. `None` used to be handled properly, but regressed when I added the failsafe to avoid infinite loops

PR #1085, opened 2022-01-19
Move re-exports into `types` and re-export `AggregatedBytes`

jdisanti:
## Motivation and Context
This implements https://github.com/awslabs/smithy-rs/issues/944.

## Checklist


jdisanti:
Opening this PR as a draft to gather feedback. This change will also require updating all the examples, which I want to hold off on until consolidation is complete.

Open questions: 
1. Should the re-exported `SdkError` type follow other re-exports and be in the `types` module, or should it remain in root?
2. Should we re-export types even if they're not used? For example, should `ByteStream` be re-exported by a service that has no streaming operations?

Velfi:
> - Should the re-exported `SdkError` type follow other re-exports and be in the types module, or should it remain in root?

I like setting and sticking to conventions so putting it in types makes sense to me unless that somehow makes codegen more difficult.

> - Should we re-export types even if they're not used? For example, should `ByteStream` be re-exported by a service that has no streaming operations?

On first thought, I think it shouldn't be re-exported. However, I'm worried about SDKs that add and then remove a bytestream based api. Is that something we need to consider or will that never happen?

rcoh (approved):

rcoh:
generally APIs can't be removed

jdisanti:
Will proceed with:
1. All re-exports go into the `types` namespace
2. Don't re-export unused types

jdisanti:
Examples fixed in https://github.com/awsdocs/aws-doc-sdk-examples/pull/2776

rcoh (approved):

PR #1096, opened 2022-01-20
Fix #1011 - Server allows empty query

### LINKED ISSUE #1011: [Server] MissingQueryString rejection for empty query which is not required
When generating codegen-server for S3, the ListObjects operation fails with MissingQueryString rejection when no search query is specified in the uri, but it does work with empty search term (just by adding an ending `?`). 

This operation input has members with the `smithy.api#httpQuery` trait,  but none of them is required, so I don't think this should fail when the search query is missing because it is a valid request uri AFAIK.

WDYT @crisidev @david-perez 

```
$ curl -X GET http://127.0.0.1:3333/lala
Expected query string in URI but none found

$ curl -X GET http://127.0.0.1:3333/lala?
<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"></ListBucketResult>
```

The rejection itself originates from here:
https://github.com/awslabs/smithy-rs/blob/b828dc80feddfa23bca904ddf1f1de2cd7a6f441/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt#L674

guymguym:
## Motivation and Context
- Fix #1011 

## Description
- Avoid failing server requests on empty query.
- Temporary fix as described in https://github.com/awslabs/smithy-rs/issues/1011#issuecomment-1017394740

## Checklist

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
@@ -33,13 +33,13 @@ import software.amazon.smithy.rust.codegen.rustlang.withBlock
 import software.amazon.smithy.rust.codegen.rustlang.writable
 import software.amazon.smithy.rust.codegen.server.smithy.ServerCargoDependency
 import software.amazon.smithy.rust.codegen.server.smithy.ServerRuntimeType
+import software.amazon.smithy.rust.codegen.server.smithy.generators.http.ServerRequestBindingGenerator
 import software.amazon.smithy.rust.codegen.smithy.CodegenContext
 import software.amazon.smithy.rust.codegen.smithy.RuntimeConfig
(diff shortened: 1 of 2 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -746,7 +747,7 @@ private class ServerHttpProtocolImplGenerator(
         with(writer) {
             rustTemplate(
                 """
-                let query_string = request.uri().query().ok_or(#{SmithyHttpServer}::rejection::MissingQueryString)?;
```
david-perez, line 749: Can you also remove the struct from `rust-runtime/aws-smithy-http-server/src/rejection.rs`? This is the only place where it was used.

guymguym: Sure, done

david-perez (approved):
@guymguym Thank you.

PR #1097, opened 2022-01-20
`aws-smithy-http-server`: fix routing collisions on empty path segments

guymguym:

## Motivation and Context
- Followup on #1009 and #1029

## Description
- Avoid empty string segments as single labels too.
- Match entire path with regexp - `^RE$`
- Make path weight higher than query for rank sorting of routes

## Testing
- Tested routing between S3 ListBuckets vs ListObjects.

## Checklist

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
+++ b/rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
@@ -107,14 +107,13 @@ impl From<&PathSpec> for Regex {
                 .map(|segment_spec| match segment_spec {
                     PathSegment::Literal(literal) => literal,
                     // TODO(https://github.com/awslabs/smithy/issues/975) URL spec says it should be ASCII but this regex accepts UTF-8:
-                    // `*` instead of `+` because the empty string `""` can be bound to a label.
-                    PathSegment::Label => "[^/]*",
+                    PathSegment::Label => "[^/]+",
                     PathSegment::Greedy => ".*",
                 })
                 .fold(String::new(), |a, b| a + sep + b)
         };
 
-        Regex::new(&format!("{}$", re)).unwrap()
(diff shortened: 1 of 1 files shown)


rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
```diff
@@ -107,14 +107,13 @@ impl From<&PathSpec> for Regex {
                 .map(|segment_spec| match segment_spec {
                     PathSegment::Literal(literal) => literal,
                     // TODO(https://github.com/awslabs/smithy/issues/975) URL spec says it should be ASCII but this regex accepts UTF-8:
-                    // `*` instead of `+` because the empty string `""` can be bound to a label.
-                    PathSegment::Label => "[^/]*",
+                    PathSegment::Label => "[^/]+",
```
david-perez, line 110: Unfortunately we don't know whether this is correct. The Smithy spec should detail what to do in the case of empty path segments. We are tracking this in https://github.com/awslabs/smithy/issues/1024. Feel free to share your thoughts there.

For the moment we're allowing empty path segments, since otherwise the empty string `""` can never be bound to a label. Note that this is also the approach we've taken when doing label deserialization (note it's [`take_until`](https://docs.rs/nom/latest/nom/bytes/complete/fn.take_until.html) and not [`take_until1`](https://docs.rs/nom/latest/nom/bytes/complete/fn.take_until1.html)):

https://github.com/awslabs/smithy-rs/blob/main/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt#L639

However, we're aware that this is at odds with the S3 model https://github.com/awslabs/smithy/issues/1024#issuecomment-1012066749.

guymguym: Got it. Let me take another look.

PR #1098, opened 2022-01-20
Server XML response use correct members and return null if nothing to serialize

### LINKED ISSUE #1101: Server RestXML generates all structure serializers empty in xml_ser
I see that in operation_ser.rs the operation code is creating the root xml tag OK:

```rust
pub fn serialize_operation_crate_operation_list_buckets(
    output: &crate::output::ListBucketsOutput,
) -> Result<String, aws_smithy_http::operation::SerializationError> {
    let mut out = String::new();
    {
        let mut writer = aws_smithy_xml::encode::XmlWriter::new(&mut out);
        #[allow(unused_mut)]
        let mut root = writer
            .start_el("ListAllMyBucketsResult")
            .write_ns("http://s3.amazonaws.com/doc/2006-03-01/", None);
        crate::xml_ser::serialize_structure_crate_output_list_buckets_output(output, root)?
    }
    Ok(out)
}
```

But the xml_ser.rs functions are doing nothing - it does have all the functions, but they are all the same and practically empty:

```rust
pub fn serialize_structure_crate_output_list_buckets_output(
    input: &crate::output::ListBucketsOutput,
    writer: aws_smithy_xml::encode::ElWriter,
) -> Result<(), aws_smithy_http::operation::SerializationError> {
    let _ = input;
    #[allow(unused_mut)]
    let mut scope = writer.finish();
    scope.finish();
    Ok(())
}
```

@david-perez Any idea where things might have changed or perhaps this was so before and I missed it...

guymguym:
## Motivation and Context
- Fix #1101

## Description
- fix to use response members for server output
- return null if no members

## Testing
- Tested with S3 API server codegen.

## Checklist

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
@@ -451,12 +451,12 @@ private class ServerHttpProtocolImplGenerator(
         bindings: List<HttpBindingDescriptor>,
     ) {
         val structuredDataSerializer = protocol.structuredDataSerializer(operationShape)
-        structuredDataSerializer.serverOutputSerializer(operationShape).also { serializer ->
+        structuredDataSerializer.serverOutputSerializer(operationShape)?.let { serializer ->
             rust(
                 "let payload = #T(output)?;",
                 serializer
             )
-        }
+        } ?: rust("""let payload = "";""")
         // avoid non-usage warnings for response
         Attribute.AllowUnusedMut.render(this)
         rustTemplate("let mut response = #{http}::Response::builder();", *codegenScope)
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/XmlBindingTraitSerializerGenerator.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/XmlBindingTraitSerializerGenerator.kt
@@ -106,8 +106,8 @@ class XmlBindingTraitSerializerGenerator(
     override fun operationSerializer(operationShape: OperationShape): RuntimeType? {
         val fnName = symbolProvider.serializeFunctionName(operationShape)
         val inputShape = operationShape.inputShape(model)
-        val xmlMembers = operationShape.operationXmlMembers()
-        if (!xmlMembers.isNotEmpty()) {
+        val xmlMembers = operationShape.requestBodyMembers()
+        if (xmlMembers.isEmpty()) {
             return null
         }
         val operationXmlName = xmlIndex.operationInputShapeName(operationShape)
@@ -191,10 +191,13 @@ class XmlBindingTraitSerializerGenerator(
         }
     }
 
-    override fun serverOutputSerializer(operationShape: OperationShape): RuntimeType {
+    override fun serverOutputSerializer(operationShape: OperationShape): RuntimeType? {
         val fnName = symbolProvider.serializeFunctionName(operationShape)
         val outputShape = operationShape.outputShape(model)
-        val xmlMembers = operationShape.operationXmlMembers()
+        val xmlMembers = operationShape.responseBodyMembers()
+        if (xmlMembers.isEmpty()) {
+            return null
+        }
         val operationXmlName = xmlIndex.operationOutputShapeName(operationShape)
             ?: throw CodegenException("operation must have a name if it has members")
         return RuntimeType.forInlineFun(fnName, operationSerModule) {
@@ -466,8 +469,11 @@ class XmlBindingTraitSerializerGenerator(
         }
     }
 
-    private fun OperationShape.operationXmlMembers(): XmlMemberIndex =
+    private fun OperationShape.requestBodyMembers(): XmlMemberIndex =
         XmlMemberIndex.fromMembers(httpBindingResolver.requestMembers(this, HttpLocation.DOCUMENT))
+
+    private fun OperationShape.responseBodyMembers(): XmlMemberIndex =
+        XmlMemberIndex.fromMembers(httpBindingResolver.responseMembers(this, HttpLocation.DOCUMENT))
 
     private fun Shape.xmlNamespace(root: Boolean): XmlNamespaceTrait? {
         return this.getTrait<XmlNamespaceTrait>().letIf(root) { it ?: rootNamespace }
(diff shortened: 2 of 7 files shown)


david-perez:
I think the correct fix is to copy what the client is doing:

https://github.com/awslabs/smithy-rs/blob/f4d3910f227ceb4962eadac994540edb45137709/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/XmlBindingTraitSerializerGenerator.kt#L106-L115

That is, if there are no `xmlMembers`, early return `null`. That way no function to serialize will be generated. Your fix works, but the serializer function will be generated and called when in reality it has no work to do (there are no operation output members to serialize to the body).

guymguym:
@david-perez Since this fix and #1101 are almost related, I updated this PR to fix both.

Thanks for all your explanations - you've made me so much more productive!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -451,12 +451,12 @@ private class ServerHttpProtocolImplGenerator(
...
         val structuredDataSerializer = protocol.structuredDataSerializer(operationShape)
-        structuredDataSerializer.serverOutputSerializer(operationShape).also { serializer ->
+        structuredDataSerializer.serverOutputSerializer(operationShape)?.let { serializer ->
             rust(
-                "let payload = #T(output)?;",
+                "let payload = #T(&output)?;",
                 serializer
             )
-        }
+        } ?: rust("let payload = \"\";")
```
guymguym, line 459: @david-perez I wonder - would you use an empty string here or some other form of "empty" body (I saw somewhere examples with `T=()` as body type...)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -451,12 +451,12 @@ private class ServerHttpProtocolImplGenerator(
...
     ) {
         val structuredDataSerializer = protocol.structuredDataSerializer(operationShape)
-        structuredDataSerializer.serverOutputSerializer(operationShape).also { serializer ->
+        structuredDataSerializer.serverOutputSerializer(operationShape)?.let { serializer ->
             rust(
                 "let payload = #T(output)?;",
                 serializer
             )
-        }
+        } ?: rust("let payload = \"\";")
```
rcoh, line 459:
```suggestion
        } ?: rust("""let payload = "";""")
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/XmlBindingTraitSerializerGenerator.kt
```diff
@@ -191,10 +191,13 @@ class XmlBindingTraitSerializerGenerator(
         }
     }
 
-    override fun serverOutputSerializer(operationShape: OperationShape): RuntimeType {
+    override fun serverOutputSerializer(operationShape: OperationShape): RuntimeType? {
         val fnName = symbolProvider.serializeFunctionName(operationShape)
         val outputShape = operationShape.outputShape(model)
-        val xmlMembers = operationShape.operationXmlMembers()
+        val xmlMembers = operationShape.serverOutputXmlMembers()
+        if (!xmlMembers.isNotEmpty()) {
```
rcoh, line 198:
```suggestion
        if (xmlMembers.isEmpty()) {
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/XmlBindingTraitSerializerGenerator.kt
```diff
@@ -469,6 +472,9 @@ class XmlBindingTraitSerializerGenerator(
     private fun OperationShape.operationXmlMembers(): XmlMemberIndex =
         XmlMemberIndex.fromMembers(httpBindingResolver.requestMembers(this, HttpLocation.DOCUMENT))
 
+    private fun OperationShape.serverOutputXmlMembers(): XmlMemberIndex =
```
rcoh, line 475: what do you think about renaming these to `requestBodyMembers` and `responseBodyMembers`?

rcoh (approved):
lgtm, just some small cleanups suggested. Final approval delegated to server team.

guymguym:
Thanks @rcoh ! updated with your suggestions.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/XmlBindingTraitSerializerGenerator.kt
guymguym: @rcoh maybe you can help me out here - I get these compilation errors for calling isEmpty -

```
e: /home/runner/work/smithy-rs/smithy-rs/codegen/src
  /main/kotlin/software/amazon/smithy/rust/codegen/smithy
  /protocols/serialize/XmlBindingTraitSerializerGenerator.kt: (110, 24):

Unresolved reference. None of the following candidates is applicable
  because of receiver type mismatch:

public inline fun <T> Array<out TypeVariable(T)>.isEmpty(): Boolean defined in kotlin.collections
public inline fun BooleanArray.isEmpty(): Boolean defined in kotlin.collections
public inline fun ByteArray.isEmpty(): Boolean defined in kotlin.collections
public inline fun CharArray.isEmpty(): Boolean defined in kotlin.collections
public inline fun CharSequence.isEmpty(): Boolean defined in kotlin.text
public inline fun DoubleArray.isEmpty(): Boolean defined in kotlin.collections
public inline fun FloatArray.isEmpty(): Boolean defined in kotlin.collections
public inline fun IntArray.isEmpty(): Boolean defined in kotlin.collections
public inline fun LongArray.isEmpty(): Boolean defined in kotlin.collections
public inline fun ShortArray.isEmpty(): Boolean defined in kotlin.collections
```

Am I missing an import or something?

guymguym: @rcoh Thanks! added.

crisidev (approved):
Look good. Thanks a million for the contribution!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: I think you are referring to https://github.com/awslabs/smithy-rs/blob/main/rust-runtime/aws-smithy-http-server/src/body.rs#L16

A string is fine for now IMO, we can change it in the future if needed.

PR #1104, opened 2022-01-21
Improve publisher tool

jdisanti:
## Motivation and Context
These changes are for https://github.com/awslabs/smithy-rs/issues/1102.

- Updated `publish` sub-command to display crate version numbers in the publish plan
- In `fix-manifests`, added validations that:
  - Sanity check version number consistency _before_ fixing manifests
  - Check for dependency cycles _after_ fixing manifests
- Updated `yank` sub-command to take a path to `aws-sdk-rust`, and added a validation that the version in `aws-sdk-rust` matches the version to yank

## Testing
- Ran `publish` locally (without confirming) to verify crate number display
- Ran `./gradlew aws:sdk:assemble` to verify new `fix-manifests` validations. This triggered the version sanity check immediately because of local gradle caching
- Reverted the `0.5.1` fix that removed the cyclical dependency, and verified `GITHUB_ACTIONS=true ./gradlew aws:sdk:assemble` detected the cycle:
  ```
  2022-01-21T02:27:15.425813Z ERROR publisher::sort: dependency cycle! stack={PackageHandle { name: "aws-config", version: Version { major: 0, minor: 5, patch: 2 } }, PackageHandle { name: "aws-sdk-sso", version: Version { major: 0, minor: 5, patch: 2 } }, PackageHandle { name: "config-code-examples", version: Version { major: 0, minor: 1, patch: 0 } }} handle=PackageHandle { name: "aws-config", version: Version { major: 0, minor: 5, patch: 2 } }
  Error: dependency cycle detected

  > Task :aws:sdk:fixManifests FAILED
  ```
  This required the `GITHUB_ACTIONS=true` env var since otherwise the generated packages would have `package.publish` set to false, and would be filtered out from the validation
- Ran `yank` locally (without confirming) and verified its new validation works

## Open questions
- Do we care that the dependency cycle check will only fail in CI and not in local dev without setting `GITHUB_ACTIONS=true`? My original intent was to only validate crates that will actually get published (exclude examples), but then realized later that our examples don't currently have `publish = false` in their manifests.


tools/publisher/src/main.rs
```diff
@@ -108,6 +108,15 @@ fn clap_app() -> clap::App<'static, 'static> {
...
+                )
+                .arg(
+                    clap::Arg::with_name("location")
+                        .required(true)
+                        .takes_value(true)
+                        .long("location")
+                        .help("Path to `aws-sdk-rust` repo. The repo should be checked out at the \
+                               version that is being yanked so that the correct list of crate names \
+                               is used. This will be validated.")
+                )
```
rcoh, line 120: this actually needs to be `aws-sdk-rust/sdk`—should probably either update the help text or the tool

tools/publisher/src/subcommand/fix_manifests/validate.rs
```diff
@@ -0,0 +1,124 @@
...
+    let expected_sdk_version = versions
+        .get("aws-config")
+        .ok_or_else(|| anyhow!("`aws-config` crate missing"))?;
+    let expected_smithy_version = versions
+        .get("aws-smithy-types")
+        .ok_or_else(|| anyhow!("`aws-smithy-types` crate missing"))?;
+
+    for (name, version) in versions {
+        if name.starts_with("aws-smithy-") {
+            confirm_version(name, expected_smithy_version, version)?;
```
rcoh, line 29: I know we had some crate classification logic somewhere already, should we share that here?

tools/publisher/src/subcommand/fix_manifests/validate.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Validations that run before the manifests are fixed.
+///
+/// For now, this validates:
+/// - `aws-config` version number matches all `aws-sdk-` prefixed versions
+/// - `aws-smithy-` prefixed versions match `aws-` (NOT `aws-sdk-`) prefixed versions
+pub(super) fn pre_validate_manifests(versions: &BTreeMap<String, Version>) -> Result<()> {
+    info!("Pre-validation manifests...");
+    let expected_sdk_version = versions
+        .get("aws-config")
+        .ok_or_else(|| anyhow!("`aws-config` crate missing"))?;
```
rcoh, line 22: fyi, this fails on CI when CI tries to publish the smithy-rs runtime bundle

tools/publisher/src/subcommand/fix_manifests.rs
```diff
@@ -29,7 +31,11 @@ pub async fn subcommand_fix_manifests(mode: Mode, location: &str) -> Result<()>
     let manifest_paths = discover_package_manifests(location.into()).await?;
     let mut manifests = read_manifests(Fs::Real, manifest_paths).await?;
     let versions = package_versions(&manifests)?;
+
+    validate::pre_validate_manifests(&versions)?;
```
rcoh, line 35: might name these something to clarify what they are pre and post. "pre_validate_manifests" sounds like you're doing something before validation, not that you're validating the state of manifests, or you could update to be something like:

```
validate_manifests(&versions, |&mut manifests| fix_manifests(...))
```

tools/publisher/src/subcommand/fix_manifests/validate.rs
```diff
@@ -0,0 +1,124 @@
...
+use semver::Version;
+use std::collections::BTreeMap;
+use tracing::info;
+
+/// Validations that run before the manifests are fixed.
+///
+/// For now, this validates:
+/// - `aws-config` version number matches all `aws-sdk-` prefixed versions
+/// - `aws-smithy-` prefixed versions match `aws-` (NOT `aws-sdk-`) prefixed versions
+pub(super) fn pre_validate_manifests(versions: &BTreeMap<String, Version>) -> Result<()> {
```
rcoh, line 18: while we're at it, should we squash the next COE and check that the git repository is at the matching tag?

tools/publisher/src/subcommand/publish.rs
```diff
@@ -6,8 +6,8 @@
 use crate::cargo::{self, CargoOperation};
 use crate::fs::Fs;
 use crate::package::{
-    continue_batches_from, discover_package_batches, Package, PackageBatch, PackageHandle,
-    PackageStats,
+    continue_batches_from, discover_and_validate_package_batches, Package, PackageBatch,
```
rcoh, line 9: imo we should remove continue_batches_from—it doesn't actually work and could lead to use failing to publish packages.

tools/publisher/src/package.rs
```diff
@@ -60,13 +60,16 @@ pub struct Package {
     pub manifest_path: PathBuf,
     /// Dependencies used by this package
     pub local_dependencies: BTreeSet<PackageHandle>,
+    /// Whether or not the package should be published
+    pub publish_enabled: bool,
```
rcoh, line 64: I might consider an enum for this, `publish_enabled` seems _pretty_ clear, but I wouldn't be surprised if someone gets it backwards somewhere

rcoh (approved):

rcoh (approved):

tools/publisher/src/main.rs
jdisanti: Great catch! I brought back the git repository resolution logic and had it detect the case where it was given `aws-sdk-rust` to resolve `sdk/`.

tools/publisher/src/subcommand/fix_manifests/validate.rs
jdisanti: Ah, yeah. Forgot about that. Will fix.

tools/publisher/src/subcommand/fix_manifests/validate.rs
jdisanti: Fixed

tools/publisher/src/subcommand/fix_manifests/validate.rs
jdisanti: Added!
```
Error: Current tag `v0.5.0-3-gf4165a00742` in the local `aws-sdk-rust` repository didn't match expected release tag `v0.5.1`
```

tools/publisher/src/subcommand/publish.rs
jdisanti: I'm onboard with this. Will remove.

tools/publisher/src/subcommand/fix_manifests.rs
jdisanti: Renamed the functions to clarify their intent.

rcoh (approved):

PR #1109, opened 2022-01-22
Generate readme/changelog as build artifact and prevent example publish

jdisanti:
## Motivation and Context
This is more progress towards https://github.com/awslabs/smithy-rs/issues/1102. This PR:
- Inflates a README template with the correct version and MSRV for `aws-sdk-rust` into the SDK build artifacts
- Copies the SDK changelog into the SDK build artifacts
- Prevents accidental publish of examples to crates.io by adding `publish = false` to their manifests


jdisanti:
Looks like inflating the README fails in GitHub Actions due to failing to get the smithy-rs repo's current tag, but it doesn't need that, so I will update it to not load the tag for this use-case.

aws/SDK_README.md.hb
```diff
@@ -0,0 +1,109 @@
...
+    tokio = { version = "1", features = ["full"] }
+    ```
+
+3. Provide your AWS credentials with the default credential provider chain, which currently looks in:
+   - Environment variables: `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_REGION`
+   - Web Identity Token credentials from the environment or container (including EKS)
+   - The default credentials files located in `~/.aws/config` and `~/.aws/credentials` (location can vary per platform)
+   - ECS Container Credentials (IAM roles for tasks)
+   - EC2 Instance Metadata Service (IAM Roles attached to instance)
+**Note:** SSO is not supported yet.
```
Velfi, line 47: We support this now right?

aws/sdk/build.gradle.kts
```diff
@@ -263,6 +273,21 @@ tasks.register<Exec>("fixManifests") {
...
+
+    workingDir(publisherToolPath)
+    commandLine(
+        "cargo", "run", "--","inflate-readme",
+        "--sdk-version", getSdkVersion(),
+        "--msrv", getRustMSRV(),
+        "--output", outputDir.resolve("README.md").absolutePath
+    )
+}
+
```
Velfi, line 290: A possibly better term than inflate is hydrate:
- https://en.wikipedia.org/wiki/Hydration_(web_development)
- https://stackoverflow.com/questions/6991135/what-does-it-mean-to-hydrate-an-object

It's what I used to call this sort of thing in web dev world (that or just "template" although I don't quite when the noun and verb for something are the same word.)

Velfi (approved):

aws/SDK_README.md.hb
jdisanti: Good catch! Looks like this needs to be updated.

aws/sdk/build.gradle.kts
jdisanti: Probably the word I was reaching for but couldn't quite remember 😄

PR #1115, opened 2022-01-25
Rebase `next` off `main` before syncing

jdisanti:
## Motivation and Context
This is intended to address the sync issues between `next`/`main` in https://github.com/awslabs/smithy-rs/issues/1102. I think this approach should have the desired end result without the complexity of detecting/running at the end of a release.

## Description

Basically, every time the sync workflow runs, the sync tool will rebase `next` on top of `main` before doing anything else. Below, I attempt to think of scenarios we could run into and explain why they work with this approach.

### Scenario 1: `next` was recently merged into `main` for a release, and the sync tool runs

In this scenario, the `next` branch was rebase-merged into `main` via GitHub PR. When the sync tool runs, `main` and `next` will have the same effective commits, but potentially with different commit hashes since `main` will have rebased commits.

When the sync tool rebases in this scenario, git is able to determine the commits are the same despite their different hashes and make `next` match `main` (verified locally).

### Scenario 2: `main` has a commit that updates a handwritten file that doesn't conflict with `next`, and the sync tool runs

The `next` branch will acquire the commit from `main` underneath the rest of its synced, unreleased, changes.

### Scenario 3: `main` has a commit that modifies a file, and `next` has another commit that modifies that same file differently since the last release.

The sync tool will run in this case, but immediately fail because of the conflict. A developer will need to manually fix the `next` branch. This should be rare since generated files should not be modified directly in `main`. It could happen if, for example, a GitHub workflow was modified both in `main` and `next`.

### How is this different from the rebase-merge that broke during the last release?

This is keeping the `next` branch up-to-date with `main` such that its changes can always be fast-forward merged into `main`, or if they can't, the sync job will fail until that is rectified. In the last release, the `next` branch was allowed to drift away from `main` since its commits were rebase-merged into `main`, but never synced back into `next` after that rebase.

### DIFF
--- a/tools/smithy-rs-sync/src/main.rs
+++ b/tools/smithy-rs-sync/src/main.rs
@@ -80,19 +80,11 @@ fn sync_aws_sdk_with_smithy_rs(
     branch: &str,
     max_commits_to_sync: usize,
 ) -> Result<()> {
-    // In case these are relative paths, canonicalize them into absolute paths
-    let aws_sdk = aws_sdk.canonicalize().context(here!())?;
-    let smithy_rs = smithy_rs.canonicalize().context(here!())?;
+    let aws_sdk = resolve_git_repo("aws-sdk-rust", aws_sdk)?;
+    let smithy_rs = resolve_git_repo("smithy-rs", smithy_rs)?;
 
-    eprintln!("aws-sdk-rust path:\t{}", aws_sdk.display());
-    if !is_a_git_repository(&aws_sdk) {
-        eprintln!("warning: aws-sdk-rust is not a git repository");
-    }
-
-    eprintln!("smithy-rs path:\t\t{}", smithy_rs.display());
-    if !is_a_git_repository(&aws_sdk) {
-        eprintln!("warning: smithy-rs is not a git repository");
-    }
+    // Rebase aws-sdk-rust's target branch on top of main
+    rebase_on_main(&aws_sdk, branch).context(here!())?;
 
     // Open the repositories we'll be working with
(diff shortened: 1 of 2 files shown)


tools/smithy-rs-sync/src/main.rs
```diff
@@ -170,6 +162,43 @@ fn sync_aws_sdk_with_smithy_rs(
     Ok(())
 }
 
+fn resolve_git_repo(repo: &str, path: &Path) -> Result<PathBuf> {
+    // In case these are relative paths, canonicalize them into absolute paths
```
Velfi, line 166:
```suggestion
    // In case this is a relative path, canonicalize it into an absolute path
```

tools/smithy-rs-sync/src/main.rs
```diff
@@ -170,6 +162,43 @@ fn sync_aws_sdk_with_smithy_rs(
...
+///
+/// Running this every sync should ensure `next` will always rebase-merge cleanly
+/// onto `main` when it's time for a release, and will also ensure history is common
+/// between `main` and `next` after a rebase-merge occurs for release.
+///
+/// The reason this works is because rebasing on main will produce the exact same
+/// commits as the rebase-merge pull-request will into main so long as no conflicts
+/// need to be resolved. Since the sync is run regularly, this will catch conflicts
+/// before syncing a commit into the target branch.
+fn rebase_on_main(aws_sdk_path: &Path, branch: &str) -> Result<()> {
```
Velfi, line 185: This is a good solution, thanks for fixing this.

Velfi (approved):

PR #1116, opened 2022-01-25
Add conversion methods for `PresignedRequest` and update docs

Velfi:
## Motivation and Context

This PR is a result of some papercuts I experienced while investigating aws-sdk-rust#423

## Description

update: s3 presigning example with curl command generator
add: impl into for PresignedRequest -> http::request::Builder
sort: cargo deps

## Testing

I ran it locally

## Checklist


aws/rust-runtime/aws-inlineable/src/presigning.rs
```diff
@@ -186,6 +186,20 @@ pub mod request {
...
+
+    impl Into<http::request::Builder> for PresignedRequest {
+        fn into(self) -> http::request::Builder {
+            let mut builder = http::request::Builder::new()
+                .uri(self.uri())
+                .method(self.method());
+
+            if let Some(headers) = builder.headers_mut() {
+                *headers = self.headers().clone();
+            }
```
rcoh, line 198: this will only be `None` if the builder is invalid, so I don't think this is the right way to go here. We should probably either implement `TryInto` or use `expect` under the pretense that `uri` and method are both valid so the builder _must_ be valid

aws/rust-runtime/aws-inlineable/src/presigning.rs
```diff
@@ -186,6 +186,20 @@ pub mod request {
                 .finish()
         }
     }
+
+    impl Into<http::request::Builder> for PresignedRequest {
```
rcoh, line 190: let's add an example of using this to the docs—finding this `Into` methods is often challenging

aws/rust-runtime/aws-inlineable/src/presigning.rs
```diff
@@ -186,6 +186,20 @@ pub mod request {
                 .finish()
         }
     }
+
```
rcoh, line 189: let's also add something like `.with_body<B>(self, body: B) -> http::Request<B> { self.0.map(|_|B) }` as a direct route without needing to go through the builder

aws/sdk/examples/s3/src/bin/put-object-presigned.rs
```diff
@@ -75,21 +75,50 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
         .build()?
         .presigned(
             &Config::from(&shared_config),
             PresigningConfig::expires_in(expires_in)?,
         )
         .await?;
-    println!("From operation input: {:?}", presigned_request.uri());
+    println!("From operation input: {:?}\n", presigned_request.uri());
+
+    print_as_curl_request(&presigned_request, Some("some test content"));
```
rcoh, line 96: we should probably also demonstrate the conversion to `http::Request` that you just added

Velfi:
@rcoh I removed the examples per @jdisanti's changes. Looking at the example repo, they're behind us a bit and will need to be brought up to speed. I'll make a PR for the example updates to that repo after I talk to our example writer about it.

Velfi:
The related example PR is here, not to be merged until we release the changes in this PR

jdisanti (approved):
LGTM!

PR #1122, opened 2022-01-26
Prepare smithy-rs `0.36.0` release

jdisanti:


gradle.properties
```diff
@@ -8,10 +8,10 @@ rust.msrv=1.54
 
 # Version number to use for the generated SDK
 # Note: these must always be full 3-segment semver versions
-aws.sdk.version=0.5.2
+aws.sdk.version=0.6.0
```
rcoh, line 11: going to 0.6 because of model updates? Could consider going to 0.35.3 but no strong preference

rcoh (approved):

Velfi (approved):

gradle.properties
Velfi: That does seem more in line with the RFC https://github.com/awslabs/smithy-rs/pull/759/files
> Bump the `patch` version if the smithy-rs version remained the same _AND_ if there are code changes to the SDK crate. This allows backwards compatible model changes to apply as patch revisions.

jdisanti: I think its safer to do minor version bumps until we have the long-term versioning scheme established, and reserve patch releases for incredibly tiny changes. Essentially, stick to the original short-term versioning scheme established in the [crates.io publishing RFC](https://github.com/awslabs/smithy-rs/pull/759).

PR #1131, opened 2022-01-28
Always set `Content-Type` header in RestJson1 responses

david-perez:
The protocol specifies it should be set by default to `application/json`
in HTTP responses, even when there are no output shape members, since in
that case an empty JSON payload `{}` is serialized. There are server
protocol tests asserting this behavior.

To implement this behavior, the `RestJsonHttpBindingResolver` class has
been introduced, which mostly delegates to the
`HttpTraitHttpBindingResolver`, but tweaks the response `Content-Type`
header.

Note this behavior is different than the client's, which should not
serialize a JSON payload if there are no input shape members to be
serialized to the body, and as such does not need to set the
`Content-Type` header.

This commit also deletes the `ServerRestJson` and `ServerRestXml`
classes, since they were exact copies of their client counterparts. The
latter was hosted in `ServerRustXml.kt`, which has been renamed to
`ServerRestXml.kt`.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -918,7 +917,7 @@ private class ServerHttpProtocolImplGenerator(
     private fun serverRenderHeaderParser(writer: RustWriter, binding: HttpBindingDescriptor, operationShape: OperationShape) {
         val httpBindingGenerator =
             ServerRequestBindingGenerator(
-                ServerRestJson(codegenContext),
+                protocol,
```
crisidev, line 920: Nice, much better.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerRestXml.kt
```diff
@@ -0,0 +1,38 @@
...
+package software.amazon.smithy.rust.codegen.server.smithy.protocols
+
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.rust.codegen.smithy.CodegenContext
+import software.amazon.smithy.rust.codegen.smithy.generators.protocol.ProtocolSupport
+import software.amazon.smithy.rust.codegen.smithy.protocols.Protocol
+import software.amazon.smithy.rust.codegen.smithy.protocols.ProtocolGeneratorFactory
+import software.amazon.smithy.rust.codegen.smithy.protocols.RestXml
+
+class ServerRestXmlFactory(private val generator: (CodegenContext) -> Protocol = { RestXml(it) }) :
```
crisidev, line 15: I would probably add a documentation to this class as it is done for restJson1.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
```diff
@@ -47,6 +49,44 @@ class RestJsonFactory : ProtocolGeneratorFactory<HttpBoundProtocolGenerator> {
     }
 }
 
+/**
+ * This [HttpBindingResolver] mostly delegates to [HttpTraitHttpBindingResolver].
```
crisidev, line 53:
```suggestion
 * This [HttpBindingResolver] mostly delegates to [HttpBindingResolver].
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
```diff
@@ -47,6 +49,44 @@ class RestJsonFactory : ProtocolGeneratorFactory<HttpBoundProtocolGenerator> {
...
+ */
+class RestJsonHttpBindingResolver(
+    model: Model,
+    contentTypes: ProtocolContentTypes,
+) : HttpBindingResolver {
```
crisidev, line 58: Why do we need this new class delegating to `HttpBindingResolver`? I understand the idea behind setting the content-type in responses.. That said, shouldn't we do this for any protocol in an agnostic way?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
david-perez: `HttpBindingResolver` is the interface; this class delegates to the `HttpTraitHttpBindingResolver` class. I've reworded the docs.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
david-perez: It is a custom detail of the protocol. Until now we have not needed a `RestJson.kt` class because the client behavior of the protocol can be entirely described by Smithy's vanilla HTTP binding traits, which is what `HttpTraitHttpBindingResolver` is in charge of. But for the server, default response `Content-Type` is a protocol-specific behavior that is not configurable with Smithy traits alone. In any case, it's best that we have a separate class for `RestJson1` like we do for the other protocols, even if the implementation is very simple and just delegates.

I've clarified the docs.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
crisidev: Ok thanks.

crisidev (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
```diff
@@ -47,6 +49,48 @@ class RestJsonFactory : ProtocolGeneratorFactory<HttpBoundProtocolGenerator> {
...
+ * The only protocol-specific behavior that is truly custom is the response `Content-Type` header, which defaults to
+ * `application/json` if not overridden.
+ */
+class RestJsonHttpBindingResolver(
+    model: Model,
+    contentTypes: ProtocolContentTypes,
+) : HttpBindingResolver {
+    private val innerResolver = HttpTraitHttpBindingResolver(model, contentTypes)
+
+    override fun httpTrait(operationShape: OperationShape): HttpTrait = innerResolver.httpTrait(operationShape)
```
rcoh, line 65: I wonder if we should add a `default` to `ProtocolContentTypes` and just handle this in `httpBindingResolver` by falling back to the default instead of `.orNull()`

rcoh (approved):
LGTM, see one potential improvement inline

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
david-perez: @rcoh I don't know. I liked the explicitness of having default response `Content-Type` be a property of `RestJson1` configured in its separate HTTP binding resolver class and not something all protocols would need to configure (even if implicitly).

david-perez: @rcoh I've realized I can achieve the same behavior while still using a separate class by using inheritance instead of composition, with less code. I've pushed another commit, let me know if you like this better.

crisidev:
I know you don't like the precommit config, but https://github.com/awslabs/smithy-rs/runs/5065001814?check_suite_focus=true 😛

crisidev (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
rcoh: Is not setting a content type ever the right thing?

rcoh (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
david-perez: @rcoh The RFC says 

> Is not setting a content type ever the right thing?

It shouldn't be set if the HTTP body is empty.

> Any HTTP/1.1 message containing an entity-body SHOULD include a Content-Type header field defining the media type of that body.

It's also technically conceivable for a protocol to specify that no `Content-Type` header should be set in an HTTP response under some circumstances, even if the HTTP body is not empty, since it's a SHOULD.

PR #1132, opened 2022-01-28
Eliminate SDK client generics

jdisanti:
## Motivation and Context
The motivation here is to simplify implementation of waiters by reducing the amount of generic bounds required to implement them. The connector and middleware can still be customized since they are dynamic.

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -89,6 +89,54 @@ author = "jdisanti"
 message = """
 Fluent clients for all services no longer have generics, and now use `DynConnector` and `DynMiddleware` to allow
 for connector/middleware customization. This should only break references to the client that specified generic types for it.
+
+If you customized the AWS client's connector or middleware with something like the following:
+```rust
+let client = aws_sdk_s3::Client::with_config(
+    aws_sdk_s3::client::Builder::new()
+        .connector(my_custom_connector) // Connector customization
+        .middleware(my_custom_middleware) // Middleware customization
+        .default_async_sleep()
+        .build(),
+    config
+);
+```
+Then you will need to wrap the custom connector or middleware in
+[`DynConnector`](https://docs.rs/aws-smithy-client/0.36.0/aws_smithy_client/erase/struct.DynConnector.html)
+and
+[`DynMiddleware`](https://docs.rs/aws-smithy-client/0.36.0/aws_smithy_client/erase/struct.DynMiddleware.html)
+respectively:
+```rust
+let client = aws_sdk_s3::Client::with_config(
+    aws_sdk_s3::client::Builder::new()
+        .connector(DynConnector::new(my_custom_connector)) // Now with `DynConnector`
+        .middleware(DynMiddleware::new(my_custom_middleware)) // Now with `DynMiddleware`
+        .default_async_sleep()
+        .build(),
+    config
+);
+```
+
+If you had functions that took a generic connector, such as the following:
+```rust
+fn some_function<C, E>(conn: C) -> Result<()>
+where
+    C: aws_smithy_client::bounds::SmithyConnector<Error = E> + Send + 'static,
+    E: Into<aws_smithy_http::result::ConnectorError>
+{
+    // ...
+}
+```
+
+Then the generics and trait bounds will no longer be necessary:
+```rust
+fn some_function(conn: DynConnector) -> Result<()> {
+    // ...
+}
+```
+
+Similarly, functions that took a generic middleware can replace the generic with `DynMiddleware` and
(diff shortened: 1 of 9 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -125,7 +129,7 @@ class PaginatorGenerator private constructor(
                 builder: #{Builder}
             }
 
-            impl${generics.inst} ${paginatorName}${generics.inst} where #{bounds:W} {
+            impl${generics.inst} ${paginatorName}${generics.inst} #{bounds:W} {
```
rcoh, line 132: nit: small inconsistency that one includes the `where` and the other does not

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -23,24 +24,52 @@ import software.amazon.smithy.rust.codegen.smithy.RuntimeConfig
...
+    }
+
+    /** Instantiation */
+    override val inst = ""
+
+    /** Trait bounds */
+    override val bounds = writable { }
+
+    /** Bounds for generated `send()` functions */
+    override fun sendBounds(input: Symbol, output: Symbol, error: RuntimeType): Writable = writable { }
```
rcoh, line 72: I'm really surprised this compiles! It generates `fn send(...) where { body }`. Probably worth including the `where` in the bounds just so we generate a slightly more normal looking code

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -88,33 +112,34 @@ class AwsFluentClientDecorator : RustCodegenDecorator {
...
+                    E: Into<#{ConnectorError}>,
+                {
                     let retry_config = conf.retry_config.as_ref().cloned().unwrap_or_default();
                     let timeout_config = conf.timeout_config.as_ref().cloned().unwrap_or_default();
                     let sleep_impl = conf.sleep_impl.clone();
                     let mut builder = #{aws_smithy_client}::Builder::new()
-                        .connector(conn)
-                        .middleware(#{Middleware}::new());
+                        .connector(#{DynConnector}::new(conn))
+                        .middleware(#{DynMiddleware}::new(#{Middleware}::new()));
```
rcoh, line 142: we _could_ consider sticking these types into the client generics for a little `DRY` but not sure if it's worth it

CHANGELOG.next.toml
```diff
@@ -78,3 +78,12 @@ message = "MSRV increased from `1.54` to `1.56.1` per our 2-behind MSRV policy."
 references = ["smithy-rs#1130"]
 meta = { "breaking" = true, "tada" = false, "bug" = false }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = """
+`Client` no longer has generics, and now uses `DynConnector` and `DynMiddleware` to allow for connector/middleware customization.
```
rcoh, line 84: can you update the changelog message to clarify that this is all generated service clients? Might also be worth calling out that this simplifies the code and docs. 

The only other call out might be custom retry policies? Not sure.

rcoh (approved):
LGTM! This is huge IMO.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
jdisanti: Wow, good catch! I'm also surprised 😅

rcoh:
LGTM!

CHANGELOG.next.toml
```diff
@@ -83,4 +83,61 @@ author = "jdisanti"
...
+    // ...
+}
+```
+
+Then the generics and trait bounds will no longer be necessary:
+```rust
+fn some_function(conn: DynConnector) -> Result<()> {
+    // ...
+}
+```
```
Velfi, line 136: > I felt a great disturbance in the Force, as if millions of ~voices~ **`where` clauses** suddenly cried out in terror and were suddenly silenced. I fear something ~terrible~ **awesome** has happened.

Velfi (approved):

PR #1133, opened 2022-01-28
Do not check headers nor query params in server protocol request tests

david-perez:
This is a holdover from when we copied the test generation code from the
client's protocol tests, which assert against the expected serialized
HTTP request from a given operation input shape. For server protocol
tests, it doesn't make sense to check that the headers or the query
params are set in the HTTP request, since the request is the input we
build from the test case configuration, not the output we expect and
assert against.

Despite the checks being unnecessary, one might be tempted to think that
it doesn't harm to check that the HTTP request we construct is the one
that is defined in the test case before attempting to deserialize it.
However, the protocol tests' `require*` and `forbid*` fields might
assert things that are derived from the input configuration: for
example, an HTTP request test case might specify `Content-Length` in
`requireHeaders` but not in `headers`. This makes a bunch of tests fail
that are perfectly valid and would pass otherwise (note that in the
previous example, `Content-Length` is not strictly required according to
the RFC, although recommended [0]). This commit removes those tests from
the list of failing tests.

[0]: https://datatracker.ietf.org/doc/html/rfc2616/#section-14.13

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
@@ -119,7 +119,7 @@ class ServerProtocolTestGenerator(
         allTests.forEach {
             renderTestCaseBlock(it.testCase, this) {
                 when (it) {
-                    is TestCase.RequestTest -> this.renderHttpRequestTestCase(it.testCase, it.targetShape)
+                    is TestCase.RequestTest -> this.renderHttpRequestTestCase(it.testCase)
                     is TestCase.ResponseTest -> this.renderHttpResponseTestCase(it.testCase, it.targetShape)
                 }
             }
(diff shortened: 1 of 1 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -233,17 +234,8 @@ class ServerProtocolTestGenerator(
...
-        checkForbidQueryParams(this, httpRequestTestCase.forbidQueryParams)
-        checkRequiredQueryParams(this, httpRequestTestCase.requireQueryParams)
-        checkHeaders(this, "&http_request.headers()", httpRequestTestCase.headers)
-        checkForbidHeaders(this, "&http_request.headers()", httpRequestTestCase.forbidHeaders)
-        checkRequiredHeaders(this, "&http_request.headers()", httpRequestTestCase.requireHeaders)
         if (protocolSupport.requestBodyDeserialization) {
-            // "If no request body is defined, then no assertions are made about the body of the message."
-            httpRequestTestCase.body.orNull()?.also { body ->
-                checkBody(this, body, httpRequestTestCase)
-            }
```
crisidev, line 246: This is what I was talking about when I wanted to remove the client behaviour!! Thanks!

crisidev (approved):

Issue #1139, opened 2022-01-31
Multiple values for the same header name in for `httpPrefixHeaders`


david-perez:
This modified Smithy protocol test

```diff
diff --git a/smithy-aws-protocol-tests/model/restJson1/http-prefix-headers.smithy b/smithy-aws-protocol-tests/model/restJson1/http-prefix-headers.smithy
index 3f02c6412..cb31f9f7b 100644
--- a/smithy-aws-protocol-tests/model/restJson1/http-prefix-headers.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-prefix-headers.smithy
@@ -105,20 +105,21 @@ operation HttpPrefixHeadersInResponse  {
 }
 
 apply HttpPrefixHeadersInResponse @httpResponseTests([
     {
         id: "HttpPrefixHeadersResponse",
         documentation: "(de)serializes all response headers",
         protocol: restJson1,
         code: 200,
         headers: {
             "X-Foo": "Foo",
+            "X-Foo": "Bar",
             "Hello": "Hello"
         },
         params: {
             prefixHeaders: {
                 "X-Foo": "Foo",
                 "Hello": "Hello"
             }
         }
     },
 ])
```

causes a panic

```
thread 'operation::http_prefix_headers_in_response_request_test::http_prefix_headers_response_response' panicked at 'called `Result::unwrap()` on an `Err` value: HttpPrefixHeadersInResponseError { kind: Unhandled("Failed to parse prefixHeaders from prefix header `"), meta: Error { code: None, message: None, request_id: None, extras: {} } }',
rest_json/rust-codegen/src/operation.rs:2574:29
```

because we're calling [`aws_smithy_http::header::one_or_none`](https://docs.rs/aws-smithy-http/latest/aws_smithy_http/header/fn.one_or_none.html), which returns an error in case there is more than one value for the same header name in the input header iterator.

The client will experience an HTTP parse error and reject the response. Is this desired behavior for a client? I think yes, but we should at least have a test.

Should a server implementation reject such a request? Or take the first / last header value? https://github.com/awslabs/smithy/issues/1069

PR #1140, opened 2022-01-31
Add support for `httpPrefixHeaders` in server requests

david-perez:
The implemented behavior is to fail to deserialize HTTP requests if
there is more than one value for the same header name. See #1139.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -911,7 +912,7 @@ private class ServerHttpProtocolImplGenerator(
     private fun serverRenderHeaderParser(writer: RustWriter, binding: HttpBindingDescriptor, operationShape: OperationShape) {
         val httpBindingGenerator =
             ServerRequestBindingGenerator(
-                ServerRestJson(codegenContext),
+                protocol,
```
crisidev, line 915: Didn't you already fix this in a PR I already reviewed? 😕

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -157,7 +157,7 @@ class HttpBindingGenerator(
                     let headers = #T::headers_for_prefix(header_map, ${binding.locationName.dq()});
                     let out: std::result::Result<_, _> = headers.map(|(key, header_name)| {
                         let values = header_map.get_all(header_name);
-                        #T(values.iter()).map(|v| (key.to_string(), v.unwrap()))
+                        #T(values.iter()).map(|v| (key.to_string(), v.expect("we've checked there's at least one value for this header name")))
```
crisidev, line 160:
```suggestion
                        #T(values.iter()).map(|v| (key.to_string(), v.expect("We have checked there is at least one value for this header name")))
```

crisidev (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
david-perez: I don't know if there are API guidelines for panic error messages, but I start all of these with lowercase as per https://doc.rust-lang.org/nightly/std/error/trait.Error.html:

> Error messages are typically concise lowercase sentences without trailing punctuation:

I'd only include punctuation (full-stops / starting with uppercase) in multi-sentence error messages.

---

I've removed the contractions though.

crisidev (approved):

PR #1141, opened 2022-02-01
Fix server protocol response body assertion

david-perez:
In server protocol tests, we were asserting that the returned server
response body matched *exactly* the body field in the protocol tests.
However, Smithy protocol tests are written with readability in mind, and
as such this field contains formatted bodies with abundant spacing and
newlines. For example:

```
Diff < left / right > :
<"{\n    \"stringValue\": \"string\",\n    \"documentValue\": {\n        \"foo\": \"bar\"\n    }\n}"
>b"{\"documentValue\":{\"foo\":\"bar\"},\"stringValue\":\"string\"}"
```

This commit changes the response body assertion to use
`aws_smithy_protocol_test::validate_body`, a helper function that takes
these formatting differences into account using the configured protocol
test `bodyMediaType` field.

This change makes many response protocol tests pass.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -310,14 +310,9 @@ class ServerProtocolTestGenerator(
...
         checkHttpResponseExtensions(this)
-        if (!testCase.body.isEmpty) {
-            rustTemplate(
-                """
-                let body = #{Hyper}::body::to_bytes(http_response.into_body()).await.expect("unable to extract body to bytes");
-                #{AssertEq}(${escape(testCase.body.get()).dq()}, body);
-                """,
-                *codegenScope
-            )
+        // "If no request body is defined, then no assertions are made about the body of the message."
```
crisidev, line 313:
```suggestion
        // If no request body is defined, then no assertions are made about the body of the message.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -343,6 +338,34 @@ class ServerProtocolTestGenerator(
...
+            """
+            let body = #{Hyper}::body::to_bytes(http_response.into_body()).await.expect("unable to extract body to bytes");
+            """,
+            *codegenScope
+        )
+        if (body == "") {
+            rustWriter.rustTemplate(
+                """
+                // No body
+                #{AssertEq}(std::str::from_utf8(&body).unwrap(), "");
```
crisidev, line 352:
```suggestion
                #{AssertEq}(std::str::from_utf8(&body).expect("Should be an empty body string"), "");
```

crisidev: Not sure if the comment is the right one, but I would use `expect()` instead of `unwrap()`

crisidev (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
david-perez: This will panic if the body is not UTF8, not if it's empty.

`unwrap()` is fine here because it's a test, and `from_utf8()` already returns an `Err` with a reason as to why the body is not UTF8.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
david-perez: This is intentionally in quotes because I'm citing the Smithy spec.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
crisidev: Ok, makes sense.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
crisidev: Don't worry, it just felt weird to me..

PR #1144, opened 2022-02-01
HTTP Connectors Refactor pt.1 - Laying the groundwork

Velfi:
## Motivation and Context
This is the first PR in a series that will change the way that HTTP connector are created, configured, and shared throughout the SDK. The goal is providing a better experience for users that want to easily configure various aspects of how the SDK makes HTTP requests.

Because this is such a large PR, here's a high-level breakdown of the changes:
1. The `default_providers` module has be broken up into multiple files, one per submodule.
2. The `sts` module has be broken up into multiple files, one per submodule.
3. The `aws_config::connector` module now has it's own file.
4. The `http_provider` module has been renamed to `http_credential_provider` to better reflect the fact that it contains the `HttpCredentialProvider` and related code.
5. Imports in several `aws_config` modules have be separated into separate groups (libstd imports, same-crate imports, aws and smithy runtime crates). This is an unspoken pattern that we've been following elsewhere.
6. **BREAKING CHANGE** The `aws_smithy_client::timeout::Settings` struct has been fully replaced by `aws_smithy_types::timeout::TimeoutConfig`. I should have done this when I originally added timeout support but better late than never, right? All structs and functions that referenced `aws_smithy_client::timeout::Settings` have been updated.
7. **BREAKING CHANGE** `MakeConnectorFn`, `HttpConnector`, and `HttpSettings` have been moved from `aws_config::provider_config` to `aws_smithy_client::http_connector`. This is in preparation for a later PR that will change how connectors are created and configured. Also, I had to update the way that `HttpSetting` is created because it's a non-exhaustive struct.
8. **BREAKING CHANGE** `HttpConnector::make_connector` has been renamed to `HttpConnector::connector` in adherence with how we tend to name the methods on providers that return the thing the provider provides. I'm open to workshopping this name or even reverting this change if y'all don't think it's helpful.
9. As a result of the changes in point 7, `aws_types` now depends on `aws_smithy_client`.
10. `aws_types::Config` now has a builder + setter method for setting an `HttpConnector`. Nothing uses this yet, but it will be used when I do the actual "Config owns Connector" refactor in a later PR. I'm OK with removing this change for now if y'all think it would mislead people.

_NOTE: This PR breaks our example for setting timeouts, I'll create a PR to the docs repo addressing that once this one is merged. It'll be a one-line change._

The next PR in this series will focus on an RFC for "Config owns Connector" + an implementation.

## Testing
- I ran the existing tests we had locally and in CI.


jdisanti:
I love the splitting out the modules! I think that's going to make it easier to find things.

> As a result of the changes in point 6, aws_types now depends on aws_smithy_client.

I understand why this is needed, but at the same time, I'm slightly concerned about tying `aws-types` to `aws-smithy-client` since we aim to keep `aws-types` as stable as possible (from [the RFC it was introduced in](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0001_shared_config.md), "`aws-types` should be an interface-only crate that is extremely stable").

Should we have an `aws-types`-like stable interface crate in the Smithy runtime to avoid locking ourselves into certain API decisions? Seems like `aws-smithy-types` could be this crate, and it does have the `TimeoutConfig`, so maybe that location makes sense.

> Imports in several aws_config modules have be separated into separate groups (libstd imports, same-crate imports, aws and smithy runtime crates). This is an unspoken pattern that we've been following elsewhere.

It's unfortunate that the options to automate this in rustfmt still haven't been stabilized--perhaps we should opt into them anyway. I want to avoid developers changing import style back and forth as my tendency is to lump them all together alphabetically. I don't have a strong preference on the style used; just that its consistency should be enforced/corrected by a computer.

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -376,63 +397,10 @@ mod loader {
...
-    #[cfg(not(any(feature = "rustls", feature = "native-tls")))]
-    pub(crate) fn default_connector(
-        _settings: &HttpSettings,
-        _sleep: Option<Arc<dyn AsyncSleep>>,
-    ) -> Option<DynConnector> {
-        None
-    }
+// unused when all crate features are disabled
+/// Unwrap an [`Option<DynConnector>`](aws_smithy_client::erase::DynConnector), and panic with a helpful error message if it's `None`
+pub fn expect_connector(
```
jdisanti, line 402: This was private to the crate before, and I think it can still be that way from what I see in the diff. If it remains private, then its organization doesn't matter as much.
```suggestion
pub(crate) fn expect_connector(
```

jdisanti (approved):
LGTM!

aws/rust-runtime/aws-config/src/lib.rs
Velfi: Since I moved the default connector stuff back into aws-config, I put those and `expect_connector` in a new `aws_config::connector` module. I then added the `pub(crate)` you suggested.

Issue #1137, opened 2022-01-31
Generated Server code shouldn't use non-exhaustive tag on struct


llgerard:
When using the generated code to write the server side implementation, I would prefer to have the structures defined without the non-exhaustive tag.

This tag makes a lot of sense for client code which needs to be able to work with newer versions of the server code (so new fields are not something the client code has to be aware of).

For server code, I would want updates to the model to mean server code fails to compile until it is updated to match the changes, the most obvious case being for a new output field. 

Notes for myself:
In the generator code, the tag is added as a `RustMetadata` with `additionalAttributes = listOf(NonExhaustive)`.
The tag also forces the use of `.builder()`.

llgerard:
@crisidev seems like you are actively looking at that part of the code, what are your thoughts?

david-perez:
I agree. We're working on improving the ergonomics of using generated model `struct`s in the server implementation in #1095. Since the server works with the latest version of the model, service implementers should be able to construct model structs with and without builders.

crisidev:
@llgerard, I almost have a PR ready to make @required trait behave as expected in the server generation. I agree with you and David, we should remove the non-exaustive tag in server codegen. I will work on it as part of #1095

Issue #1095, opened 2022-01-20
Non-`Option`al struct members for `@required` shape members in the server


david-perez:
Since [adding `@required` to a shape member](https://awslabs.github.io/smithy/1.0/guides/evolving-models.html?highlight=backwards#updating-structures) is **not** a backwards-compatible change, the server implementation can generate non-`Option`al struct members for `@required` shape members.

PR #1148, opened 2022-02-02
[Server] Respect @required trait in model generation

crisidev:
## Motivation and Context
Since adding [@required to a shape member](https://awslabs.github.io/smithy/1.0/guides/evolving-models.html?highlight=backwards#updating-structures) is not a backwards-compatible change, the server implementation can generate non-Optional structure members for @required shape members.

On top of this generated server code shouldn't use non-exhaustive tag on structures for the same non-backwards compatibility and because it forces to constructs instances through builders.

## Description
This change introduces the new parameter `handleRequired` when constructing the `SymbolVisitor`. This parameter is set to false for client codegen and to true for server codegen.

With this new parameter we control whether to always treat symbols as `Option` or to follow the model directives more strictly remove optionality for @required members.

## Testing
See CI and codegen diff.

Closes: #1095
Closes: #1137


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
```diff
@@ -281,11 +313,16 @@ class SymbolVisitor(
...
+        if (config.handleOptionality && config.handleRequired) {
+            throw CodegenException("CodegenVisitor 'handleOptionality' and 'handleRequired' configuration options cannot be both true at the same time")
+        }
         // Handle boxing first so we end up with Option<Box<_>>, not Box<Option<_>>
         return targetSymbol.letIf(config.handleRustBoxing) {
             handleRustBoxing(it, shape)
         }.letIf(config.handleOptionality) {
             handleOptionality(it, shape)
+        }.letIf(config.handleRequired) {
+            handleRequiredTrait(it, shape)
```
rcoh, line 325: could merge these and handle optionality in one place based on the flags

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
```diff
@@ -281,11 +313,16 @@ class SymbolVisitor(
     override fun memberShape(shape: MemberShape): Symbol {
         val target = model.expectShape(shape.target)
         val targetSymbol = this.toSymbol(target)
+        if (config.handleOptionality && config.handleRequired) {
+            throw CodegenException("CodegenVisitor 'handleOptionality' and 'handleRequired' configuration options cannot be both true at the same time")
+        }
```
rcoh, line 318: this makes me wonder if we should have a mode flag in codegenConfig instead of booleans

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
```diff
@@ -63,14 +64,16 @@ data class SymbolVisitorConfig(
...
     val handleOptionality: Boolean = true,
-    val handleRustBoxing: Boolean = true
+    val handleRustBoxing: Boolean = true,
+    val handleRequired: Boolean = false
 )
 
 val DefaultConfig =
     SymbolVisitorConfig(
         runtimeConfig = RuntimeConfig(),
         handleOptionality = true,
```
rcoh, line 74: handleOptionalityFalse is kinda useless, I think we should just remove it.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/BuilderGenerator.kt
```diff
@@ -136,12 +136,15 @@ class BuilderGenerator(
...
     ) {
-        // Render a `set_foo` method. This is useful as a target for code generation, because the argument type
-        // is the same as the resulting member type, and is always optional.
-        val inputType = outerType.asOptional()
+        // Render a `set_foo` method.
+        val (inputType, inputVal) = if (symbolProvider.isRequiredTraitHandled(member)) {
+            outerType to "Some(input)"
+        } else {
+            outerType.asOptional() to "input"
+        }
```
rcoh, line 144: same here, can use the output type + `isOptional`

rcoh: If you need to bridge a possibly optional input with a possible optional output I'd do something like:

fn bridgeOptionality(input: Symbol, output: Symbol, field: String) {
  if (input.isOptional() && output.isOptional()) { ... } // the rest of the 4 cases etc.
}

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCodegenServerPlugin.kt
```diff
@@ -66,7 +66,7 @@ class RustCodegenServerPlugin : SmithyBuildPlugin {
                 // Generate [ByteStream] instead of `Blob` for streaming binary shapes (e.g. S3 GetObject)
                 .let { StreamingShapeSymbolProvider(it, model) }
                 // Add Rust attributes (like `#[derive(PartialEq)]`) to generated shapes
-                .let { BaseSymbolMetadataProvider(it) }
+                .let { BaseSymbolMetadataProvider(it, true) }
```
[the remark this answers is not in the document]
crisidev: Yep, makes sense.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -1047,4 +1046,8 @@ private class ServerHttpProtocolImplGenerator(
             }
         }
     }
+
+    private fun getReturnTypeForMember(member: MemberShape): String {
+        return if (symbolProvider.isRequiredTraitHandled(member)) { "value" } else { "Some(value)" }
+    }
```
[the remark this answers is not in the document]
crisidev: All you are saying here makes sense.. I am not super happy about this method and the proposed changes will make this much nicer to read. Thanks, will fix.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolMetadataProvider.kt
```diff
@@ -67,10 +71,12 @@ abstract class SymbolMetadataProvider(private val base: RustSymbolProvider) : Wr
     abstract fun enumMeta(stringShape: StringShape): RustMetadata
 }
 
-class BaseSymbolMetadataProvider(base: RustSymbolProvider) : SymbolMetadataProvider(base) {
+class BaseSymbolMetadataProvider(base: RustSymbolProvider, handleRequired: Boolean = false) : SymbolMetadataProvider(base) {
+    // Server structure should not have #[non_exaustive] tag as model changes are always breaking changes.
+    val additionalAttributes = if (handleRequired) { listOf() } else { listOf(NonExhaustive) }
```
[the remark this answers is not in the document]
crisidev: I think you are right, it is better to be explicit here.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
crisidev: I think I prefer to have the full logical division to be honest, but I don't have a very strong reasoning behind it.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -510,7 +509,7 @@ private class ServerHttpProtocolImplGenerator(
         }
 
         val bindingGenerator = ServerResponseBindingGenerator(protocol, codegenContext, operationShape)
-        val addHeadersFn = bindingGenerator.generateAddHeadersFn(errorShape?: operationShape)
+        val addHeadersFn = bindingGenerator.generateAddHeadersFn(errorShape ?: operationShape)
```
Velfi, line 512: As someone still relatively new to Kotlin, what does this syntax mean? Is this the Kotlin equivalent of `unwrap_or_else`?

crisidev: I present you the "elvis operator": https://kotlinlang.org/docs/null-safety.html#elvis-operator 😃

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/BuilderGenerator.kt
crisidev: Same as above, I believe your suggestion makes sense. Will fix.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
crisidev: I am not sure I understand..

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
```diff
@@ -281,11 +313,16 @@ class SymbolVisitor(
     override fun memberShape(shape: MemberShape): Symbol {
         val target = model.expectShape(shape.target)
         val targetSymbol = this.toSymbol(target)
+        if (config.handleOptionality && config.handleRequired) {
```
david-perez, line 316: If we end up using two flags, I'd check this once upon `SymbolVisitorConfig` creation and never again.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -191,6 +191,12 @@ class JsonParserGenerator(
         return structureParser(fnName, inputShape, includedMembers)
     }
 
+    private fun getSuffixForRequiredTrait(shape: MemberShape): String {
+        return if (symbolProvider.isRequiredTraitHandled(shape, false)) {
+            """.ok_or_else(|| aws_smithy_json::deserialize::Error::custom("Shape ${shape.memberName} is required"))?"""
+        } else { "" }
+    }
```
david-perez, line 198:
```suggestion
    private fun getSuffixForRequiredTrait(shape: MemberShape) = 
        if (symbolProvider.isRequiredTraitHandled(shape, false)) {
            """.ok_or_else(|| aws_smithy_json::deserialize::Error::custom("shape ${shape.memberName} is required"))?"""
        } else { 
            "" 
        }
```

david-perez: Consider also renaming to `suffixForRequiredTrait`.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -372,17 +376,24 @@ class HttpBindingGenerator(
                     })
                     """
                 )
-            else -> rustTemplate(
-                """
-                if $parsedValue.len() > 1 {
-                    Err(#{header_util}::ParseError::new_with_message(format!("expected one item but found {}", $parsedValue.len())))
+            else -> {
+                val returnValue = if (symbolProvider.isRequiredTraitHandled(memberShape)) {
+                    """$parsedValue.pop().ok_or_else(|| #{header_util}::ParseError::new_with_message("missing mandatory header '$locationName'"))"""
```
david-perez, line 381:
```suggestion
                    """$parsedValue.pop().ok_or_else(|| #{header_util}::ParseError::new_with_message("missing required header '$locationName'"))"""
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
```diff
@@ -63,14 +63,16 @@ data class SymbolVisitorConfig(
     val runtimeConfig: RuntimeConfig,
     val codegenConfig: CodegenConfig,
     val handleOptionality: Boolean = true,
```
rcoh, line 65: let's fully delete `handleOptionality` as an option and just have it be what we do when `handleRequired` is false

rcoh: currently there's a bit of a weird case where if you set `handleOptionality` to false right now, we also won't handle required

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/BuilderGenerator.kt
```diff
@@ -136,8 +136,7 @@ class BuilderGenerator(
         member: MemberShape,
         memberName: String
     ) {
-        // Render a `set_foo` method. This is useful as a target for code generation, because the argument type
-        // is the same as the resulting member type, and is always optional.
+        // Render a `set_foo` method.
```
rcoh, line 139: Looks like I removed this comment by accident in my commit

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/ErrorGenerator.kt
```diff
@@ -82,8 +82,14 @@ class ErrorGenerator(
     private fun renderError(isServer: Boolean) {
         val symbol = symbolProvider.toSymbol(shape)
         val messageShape = shape.errorMessageMember()
-        val message = messageShape?.let { "self.${symbolProvider.toMemberName(it)}.as_deref()" } ?: "None"
         val errorKindT = RuntimeType.errorKind(symbolProvider.config().runtimeConfig)
+        val (returnType, message) = messageShape?.let {
+            if (symbolProvider.config().handleRequired && messageShape.isRequired()) {
```
rcoh, line 87: I think this can use the same pattern of `symbolProvider.toSymbol(messageShape).isOptional()`

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -337,10 +337,7 @@ class JsonSerializerGenerator(
         val writer = context.writerExpression
         val value = context.valueExpression
         when (target) {
-            is StringShape -> when (target.hasTrait<EnumTrait>()) {
-                true -> rust("$writer.string(${value.name}.as_str());")
-                false -> rust("$writer.string(${value.name});")
-            }
+            is StringShape -> rust("$writer.string(${value.name}.as_str());")
```
rcoh, line 340: sneaking suspicion clippy doesn't like this (but we can just add a clippy ignore)

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
crisidev: Yep, you are right, it can go away. I kept it for compatibility only.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/BuilderGenerator.kt
crisidev: Will add it back.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
```diff
@@ -144,14 +144,15 @@ class StructureGenerator(
...
+                    memberType.isDeref() -> memberType.asDeref().asRef()
                     else -> memberType.asRef()
                 }
                 rustBlock("pub fn $memberName(&self) -> ${returnType.render()}") {
                     when {
                         memberType.isCopy() -> rust("self.$memberName")
                         memberType is RustType.Option && memberType.member.isDeref() -> rust("self.$memberName.as_deref()")
                         memberType is RustType.Option -> rust("self.$memberName.as_ref()")
-                        memberType.isDeref() -> rust("self.$memberName.deref()")
+                        memberType.isDeref() -> rust("use std::ops::Deref; self.$memberName.deref()")
```
rcoh, line 155: I think this is equivalent to `*self.$memberName`? (deref coercion)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -967,6 +967,7 @@ private class ServerHttpProtocolImplGenerator(
     private fun generateParsePercentEncodedStrAsStringFn(binding: HttpBindingDescriptor): RuntimeType {
         val output = symbolProvider.toSymbol(binding.member)
         val fnName = generateParseStrFnName(binding)
+        val returnVal = if (symbolProvider.toSymbol(binding.member).isOptional()) { "Some(value)" } else { "value" }
```
rcoh, line 970: this can be `wrapOptional` now I think

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/ErrorGenerator.kt
crisidev: It looks to be working, I probably had some leftover that a good clean removed..

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -728,7 +728,7 @@ private class ServerHttpProtocolImplGenerator(
                         rustTemplate(
                             """
                             input = input.${binding.member.setterName()}(
-                                #{deserializer}(m$index)?
+                                ${symbolProvider.toOptional(binding.member, "#{deserializer}(m$index)?")}
```
rcoh, line 731: maybe we should just make #{deserializer} always return `Option<T>`?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolMetadataProvider.kt
```diff
@@ -67,10 +67,18 @@ abstract class SymbolMetadataProvider(private val base: RustSymbolProvider) : Wr
...
 
-class BaseSymbolMetadataProvider(base: RustSymbolProvider) : SymbolMetadataProvider(base) {
+/**
+ * The base metadata supports a list of attributes that are used by generators to decorate code.
+ * By default we apply #[non_exhaustive] only to client structures since model changes should
+ * be considered as breaking only when generating server code.
+ */
+class BaseSymbolMetadataProvider(
+    base: RustSymbolProvider,
+    additionalAttributes: List<Attribute> = listOf(NonExhaustive)
```
rcoh, line 77: rather than having this be the default, lets set it explicitly in `RustCodegenPlugin` in the client and remove the default

rcoh (approved):
Awesome! You may want to consider just making all the server parser functions return option to remove the need to do any conditional option wrapping

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
crisidev: Yep it is the same I think. I preferred the explicit version, but I can change this.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
crisidev: I had the same feeling, but SDK tests (and clippy) are not complaining on my local environment. Let's see what the CI have to say.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: Yes! Old habits, I need to get used to the two new methods :(

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
crisidev: I think I'd prefer to keep the `#{deserializer}` as it is and handle this in the caller here TBH.. It looks logically more correct to not return always an `Option<T>` from it if it is not needed..

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
crisidev: CI is happy 💪🏼

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
crisidev: Turns out deref coercion doesn't work for all cases and now I remember that I already tried in one of my many iterations.

You can end up with stuff like this:
```
   --> ebs/rust-server-codegen/src/input.rs:874:9
    |
873 |     pub fn snapshot_id(&self) -> &str {
    |                                  ---- expected `&str` because of return type
874 |         *self.snapshot_id
    |         ^^^^^^^^^^^^^^^^^
    |         |
    |         expected `&str`, found `str`
    |         help: consider borrowing here: `&*self.snapshot_id`
```

Of course `&*` works for some cases, but not for all of them, while explicit `deref()` always works.

david-perez (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -603,7 +602,7 @@ private class ServerHttpProtocolImplGenerator(
     ): Writable? {
         val errorSymbol = getDeserializeErrorSymbol(binding)
         return when (binding.location) {
-            HttpLocation.HEADER -> writable { serverRenderHeaderParser(this, binding, operationShape) }
+            HttpLocation.HEADER -> writable { serverRenderHeaderParser(this, binding, operationShape) } // always optional
```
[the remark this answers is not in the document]
crisidev: It's a leftover, I'll fix this.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
crisidev: `deref()` is not implemented for all types without importing the `Deref` trait. I have tried, it doesn't work and it is not worth it to make the logic to differentiate whether we should import the trait or not.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
crisidev: Hum you are probably right, it should be called only if the target is an Enum.. I'll give it a try to narrow this down.

PR #1149, opened 2022-02-03
RFC: Waiters

jdisanti:
This is an RFC that outlines how we can add waiters to the SDK without making any breaking changes to the current API.

[Rendered](https://github.com/awslabs/smithy-rs/blob/jdisanti-rfc-waiters/design/src/rfcs/rfc0010_waiters.md)

The overall approach has been proved out by forking the generated S3 client locally and (mostly) hand implementing a "bucket exists" waiter. The RFC doesn't dive into the JMESPath response classification code generation since those details aren't needed to assess whether or not breaking changes are required to implement waiters.


Velfi (approved):
This looks simple and understandable; I like it.

PR #1155, opened 2022-02-04
Fix the canary runner

jdisanti:
## Motivation and Context
Moving the canary runner into smithy-rs introduced some bugs. This PR fixes them so that the canary can pass again. [Passing run here using this branch](https://github.com/awslabs/aws-sdk-rust/runs/5060244375).


Velfi (approved):
Thanks for improving the `here` macro

PR #1158, opened 2022-02-04
Add support for `httpPayload` in server responses

david-perez:
Note that with this change, payload serializers now return `Vec<u8>` or
`ByteStream` instead of always `SdkBody`, since the server does not use
`SdkBody`. The caller then needs to convert the byte slab into the
appropiate HTTP body type. To this end, `HttpBoundProtocolBodyGenerator`
has been extracted into its own file and renamed to
`HttpBoundProtocolPayloadGenerator`.

This commit also makes the body types exports from
`aws-smithy-http-server` `#[doc(hidden)]`, since they should only be
used by generated code, and exports all of them from `body.rs` instead
of `lib.rs`.

Comments have been added where appropiate to document which serializers
are called by the client / server and how their use differs.


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
```diff
@@ -0,0 +1,262 @@
...
+        serializer: StructuredDataSerializerGenerator
+    ) {
+        when (val targetShape = model.expectShape(member.target)) {
+            is StringShape -> {
+                // Write the raw string to the payload.
+                // TODO Can we get rid of this allocation and make the function return a slice of bytes?
+                if (targetShape.hasTrait<EnumTrait>()) {
+                    rust("$payloadName.as_str().as_bytes().to_vec()")
+                } else {
+                    // TODO Cannot use `into_bytes()` because payload is behind shared reference.
```
crisidev, line 228: Should we have make these todos issues and reference them?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolGenerator.kt
```diff
@@ -399,199 +378,3 @@ class HttpBoundProtocolTraitImplGenerator(
...
-        serializer: StructuredDataSerializerGenerator
-    ) {
-        when (val targetShape = model.expectShape(member.target)) {
-            // Write the raw string to the payload
-            is StringShape -> {
-                if (targetShape.hasTrait<EnumTrait>()) {
-                    rust("$payloadName.as_str()")
-                } else {
-                    rust("""$payloadName.to_string()""")
-                }
```
crisidev, line 568: I am not sure I understand the implication of changing the return type of the payload from a str / String to a vec.. (see same function in HttpBoundProtocolPayloadGenerator)

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
```diff
@@ -36,7 +36,7 @@ import software.amazon.smithy.rust.codegen.smithy.generators.client.FluentClient
 import software.amazon.smithy.rust.codegen.smithy.generators.client.FluentClientSection
```
crisidev, line 36: Just a general question: I took a look the SDK diff and it seems there are no breaking changes there with this code change. Did you verify this?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -1075,11 +1059,11 @@ private class ServerHttpProtocolImplGenerator(
             }
         }
     }
-    private fun getStreamingBodyTraitBounds(operationShape: OperationShape): String {
+
+    private fun streamingBodyTraitBounds(operationShape: OperationShape) =
```
crisidev, line 1063: TIL: if you use "=" instead of a function with brackets, you can avoid to fill the return type.. Interesting.

crisidev (approved):
LGTM, just some minor questions and you have to run the linting :)

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
david-perez: I've fixed these. The answer to whether we can is yes for these particular serializer functions, but it's suboptimal because we then need to take ownership to put the data in the body, where we'd have to clone. Don't know what I was thinking when I wrote these.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolGenerator.kt
david-perez: I added the explanation to the PR description and the commit message of https://github.com/awslabs/smithy-rs/pull/1158/commits/0861b19eaf2394026dcff60a05ff3aeb9d1f2631, which is the one I'll use when I squash the whole patchset.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
david-perez: It's only considered idiomatic for one-liner functions. You avoid the return type and the `return` keyword, you can just use an expression.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
david-perez: There shouldn't be, it's just a class rename. I've now called out the rename in the commit message and PR description.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
```diff
@@ -0,0 +1,269 @@
...
+                is DocumentShape, is StructureShape, is UnionShape -> ProtocolPayloadGenerator.PayloadMetadata(
+                    takesOwnership = false
+                )
+                is StringShape, is BlobShape -> ProtocolPayloadGenerator.PayloadMetadata(takesOwnership = true)
+                else -> UNREACHABLE("Unexpected payload target type: $type")
+            }
+        }
+    }
+
+    override fun generatePayload(writer: RustWriter, self: String, operationShape: OperationShape) {
```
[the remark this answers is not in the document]
david-perez: I agree in retrospect, I think it's best to split this method. I've done so in https://github.com/awslabs/smithy-rs/pull/1158/commits/48085e11a3a946262d7c4902afd1f531264a865f. I don't think the other methods need to be split.

PR #1163, opened 2022-02-07
Fix duplicate content-type in server response

guymguym:

## Motivation and Context
- Content-Type header was set twice in the server response to S3 GetObject/HeadObject causing the client to fail.
- This happens for operations that bind an output member to the content-type header. 
- Even though the call to `set_response_header_if_absent` checks for dup header, it was called before the output members, and the output members code did not check for dups, and added the header twice.

## Description
- By replacing the order so that the default content-type is added after setting the output members and only if absent, the duplicate is detected on runtime and avoided.

## Testing
- Tested with S3 GetObject

## Checklist


david-perez (approved):
@guymguym Thank you.

I've made a note in https://github.com/awslabs/smithy-rs/issues/1164#issuecomment-1031385029 to ensure we don't forget to add a protocol test to not regress on this, once we have the "extras" suite hooked up.

Issue #1159, opened 2022-02-05
Server deserialize operation with required member expects enum Option


guymguym:
With #1148 codegen server of S3 model from `aws/sdk/aws-models/s3.json` gives a bunch of errors like the below :

```
error[E0308]: `?` operator has incompatible types
    --> src/operation_deser.rs:1238:13
     |
1238 |             crate::http_serde::deser_payload_delete_objects_delete_objects_input_delete(&bytes)?
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `Option`, found struct `Delete`
     |
     = note: `?` operator cannot convert from `Delete` to `Option<Delete>`
     = note: expected enum `Option<Delete>`
              found struct `Delete`
```

This is the operation and member -

```
        "com.amazonaws.s3#DeleteObjectsRequest": {
            "type": "structure",
            "members": {
                ...
                "Delete": {
                    "target": "com.amazonaws.s3#Delete",
                    "traits": {
                        "smithy.api#documentation": "<p>Container for the request.</p>",
                        "smithy.api#httpPayload": {},
                        "smithy.api#required": {},
                        "smithy.api#xmlName": "Delete"
                    }
                },
```

crisidev:
I guess we need to do another pass at optionality while using the full S3 model as you mentioned in the related issue. 

Sorry about the issue and thanks for reporting this!

crisidev:
All right, I was able to reproduce this issue by using the full S3 model. I'll work on the codegen to understand what did we miss on #1148

crisidev:
@guymguym when you have some time do you mind to validate this is actually fixed by my change?

guymguym:
@crisidev I am not sure, maybe I am doing it wrong? Here is the error:
```
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> src/s3_server.rs:85:34
    |
85  |                         unsafe { std::mem::transmute::<ServerInput, ClientInput>(i) }
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
149 |     register_s3_gateway_op2!(PutObjectTagging);
    |     ------------------------------------------ in this macro invocation
    |
[...]

PR #1166, opened 2022-02-07
Fix payload deserialization where we need to return an Optional type when the member is non-streaming

crisidev:
Fixes: #1159

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
@@ -181,10 +181,10 @@ class HttpBindingGenerator(
         httpMessageType: HttpMessageType = HttpMessageType.RESPONSE
     ): RuntimeType {
         check(binding.location == HttpBinding.Location.PAYLOAD)
-        val outputT = symbolProvider.toSymbol(binding.member)
         val fnName = "deser_payload_${fnName(operationShape, binding)}"
         return RuntimeType.forInlineFun(fnName, httpSerdeModule) { rustWriter ->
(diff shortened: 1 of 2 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -181,10 +181,10 @@ class HttpBindingGenerator(
         httpMessageType: HttpMessageType = HttpMessageType.RESPONSE
     ): RuntimeType {
```
david-perez, line 182: Unrelated to this PR, but we could fix it here:

The docblock for this function reads (GitHub won't let me add the comment since it's above the modified hunk):

```
/**
* Generate a function to deserialize `[binding]` from the response payload.
*/
```

We should update it to say "request/response payload".

david-perez (approved):
The commit + PR message seems inaccurate. It reads "Fix payload serialization where we need to return an Optional type when the member is streaming", but:
* the change affects payload **de**serialization.
* the return type is optional when the output member is **non**-streaming.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
crisidev: I can fix this. Regarding the commit message, it is kinda backward, sorry about it. I'll fix.

crisidev: I will use the right commit message when I squash all of this.

jdisanti (approved):
LGTM!

PR #1169, opened 2022-02-08
Fix git repo root detection in tools

jdisanti:
## Motivation and Context
This fixes https://github.com/awslabs/smithy-rs/issues/1162.

## Testing
- Ran hydrate-readme locally to verify repo detection still works


crisidev (approved):

tools/smithy-rs-tool-common/src/git.rs
```diff
@@ -16,26 +17,15 @@ pub use get_current_tag::GetCurrentTag;
 mod get_last_commit;
 pub use get_last_commit::GetLastCommit;
 
+mod get_repo_root;
+pub use get_repo_root::GetRepoRoot;
+
 /// Attempts to find git repository root from the given location.
 pub fn find_git_repository_root(repo_name: &str, location: impl AsRef<Path>) -> Result<PathBuf> {
```
david-perez, line 24: Is `repo_name` useful? Could we remove it? I see it's only called with either `aws-sdk-rust` or `smithy-rs`, and it is checked that the name of the directory coincides with what `git rev-parse --show-top-level` returns. I don't see the check of the `repo_name` coinciding with the directory name adding value.

I ask because I sometimes have `smithy-rs` checked out with in a dir not named `smithy-rs`.

jdisanti: It's a fair point. The reason it's in here is we often checkout smithy-rs into aws-sdk-rust in CI so that we can use parts of both, and this is a sanity check that our CI script is set up correctly. It's also helpful in code to communicate which repository we're expecting to be working with. None of this is essential though.

tools/smithy-rs-tool-common/src/git/get_repo_root.rs
```diff
@@ -0,0 +1,73 @@
...
+            program: "./git_revparse_show_toplevel",
+            start_path: "./fake_git".into(),
+        }
+        .run()
+        .unwrap();
+        assert_eq!(PathBuf::from("/git/repo/root/path"), last_commit);
+    }
+
+    #[test]
+    fn get_last_commit_faijlure() {
```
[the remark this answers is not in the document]
jdisanti: Copy paste error 😅 

It should have a different name entirely

tools/smithy-rs-tool-common/src/git.rs
david-perez: This is the patch I apply to my tree whenever I want to build the SDK https://github.com/awslabs/smithy-rs/pull/1158/commits/f76ad73a46ae1bf5b68c77ac4652838ae979785f It'd be nice if the script worked with any directory name.

jdisanti: Made it log a warning instead of failing.

PR #1174, opened 2022-02-10
Add `ServerCombinedErrorGeneratorTest.kt`

david-perez:

### DIFF
--- a/codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/ServerCombinedErrorGeneratorTest.kt
+++ b/codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/ServerCombinedErrorGeneratorTest.kt
@@ -15,7 +15,6 @@ import software.amazon.smithy.rust.codegen.testutil.TestWorkspace
 import software.amazon.smithy.rust.codegen.testutil.asSmithyModel
 import software.amazon.smithy.rust.codegen.testutil.compileAndTest
 import software.amazon.smithy.rust.codegen.testutil.renderWithModelBuilder
-import software.amazon.smithy.rust.codegen.testutil.testSymbolProvider
 import software.amazon.smithy.rust.codegen.testutil.unitTest
 import software.amazon.smithy.rust.codegen.util.lookup
 
@@ -51,7 +50,7 @@ class ServerCombinedErrorGeneratorTest {
         val project = TestWorkspace.testProject(symbolProvider)
         project.withModule(RustModule.public("error")) { writer ->
             listOf("FooException", "ComplexError", "InvalidGreeting").forEach {
-                model.lookup<StructureShape>("error#$it").renderWithModelBuilder(model, symbolProvider, writer, TYPE.SERVER)
+                model.lookup<StructureShape>("error#$it").renderWithModelBuilder(model, symbolProvider, writer, CodegenTarget.SERVER)
             }
             val generator = ServerCombinedErrorGenerator(model, symbolProvider, model.lookup("error#Greeting"))
             generator.render(writer)
--- a/codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/ServerCombinedErrorGeneratorTest.kt
+++ b/codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/ServerCombinedErrorGeneratorTest.kt
@@ -1,0 +1,87 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package software.amazon.smithy.rust.codegen.smithy.generators
+
+import org.junit.jupiter.api.Test
+import software.amazon.smithy.model.shapes.StructureShape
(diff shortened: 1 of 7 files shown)


codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/ServerCombinedErrorGeneratorTest.kt
```diff
@@ -0,0 +1,87 @@
...
+                    assert_eq!(error.is_complex_error(), false);
+
+                    // Indicate the original name in the display output.
+                    let error = FooException::builder().build();
+                    assert_eq!(format!("{}", error), "FooException")
+                """
+            )
+
+            println("file:///${project.baseDir}/src/lib.rs")
+            println("file:///${project.baseDir}/src/error.rs")
```
crisidev, line 83: Why this print? Is it a leftover or is it there for debugging purposes?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
```diff
@@ -65,17 +72,10 @@ class StructureGenerator(
...
 
-    fun render() {
-        renderStructure()
-        errorTrait?.also { errorTrait ->
-            ErrorGenerator(symbolProvider, writer, shape, errorTrait).render()
-        }
-    }
-
-    fun renderServer() {
+    fun render(forWhom: TYPE = TYPE.CLIENT) {
```
crisidev, line 75: Do you really need to specify that `fromWom` is a `TYPE` ? Doesn't the compiler guess it by itself?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/ErrorGenerator.kt
```diff
@@ -65,22 +66,7 @@ class ErrorGenerator(
...
-    /*
-     * Renders an error specific for the server implementation, where the [name] method is added to allow
-     * to record encoutered error types inside `http::Extensions`.
-     */
-    fun renderServer() {
-        renderError(true)
-    }
-
-    private fun renderError(isServer: Boolean) {
+    fun render(forWhom: TYPE = TYPE.CLIENT) {
```
crisidev, line 69: To be honest this is much nicer than using a boolean! 👍🏼

crisidev (approved):

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/ServerCombinedErrorGeneratorTest.kt
david-perez: @rcoh Is there a way of having the test infrastructure be smarter about these and always print all the generated sources automatically?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
crisidev: The fact that `fromWhom` is of type `TYPE`. I would expect this to also work: `fun render(fromWhom = TYPE.CLIENT)` but I am probably wrong.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/StructureGenerator.kt
```diff
@@ -50,6 +50,13 @@ fun redactIfNecessary(member: MemberShape, model: Model, safeToPrint: String): S
     }
 }
 
+/**
+ * The artifact type for whom we are generating the structure.
+ */
+public enum class TYPE {
```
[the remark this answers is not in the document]
david-perez: > In general, enum classes shouldn't be named with all caps. Thinking about this name though, I'm concerned Type will be too easy to confuse with the other various representations of type in the codegen. Perhaps we should call this Target or CodegenTarget?

Agree, I was hoping someone could come up with a better name. I've renamed to `CodegenTarget`. https://github.com/awslabs/smithy-rs/pull/1174/commits/c63b52efaaa450c931a51d818311d6a6b449110c

> It's probably useful to have outside of StructureGenerator so it can be used in other places, but we can always move it when using it again.

The classpath looks meaningful. Other places are importing it already via:

```
import software.amazon.smithy.rust.codegen.smithy.generators.CodegenTarget
```

You're right though in that the _file_ where the `enum` is placed might not be intuitive. I didn't want to create a `Utils.kt` file given that we've managed without one so far, but I couldn't think of any other more appropriate file.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/ServerCombinedErrorGeneratorTest.kt
david-perez: I've cut https://github.com/awslabs/smithy-rs/issues/1178 to track this improvement.

jdisanti (approved):

PR #1175, opened 2022-02-10
Add converters into server error enums

### LINKED ISSUE #1160: Server codegen could provide From Trait for operations' error types
Each operation gets a dedicated error enum generated making error handling quite verbose with the current lack of `From` trait implementations.

Example when using the following smithy model:
```
@httpError(500) @error("server")
structure InternalServerException { message: String }

operation Op1 {
    input: Input,
    output: Output,
    errors: [
        InternalServerException,
    ],
}
```
The generated code will have in `errors.rs`:
```
#[non_exhaustive]
pub struct InternalServerError {
    pub message: std::option::Option<std::string::String>,
}

#[derive(std::fmt::Debug)]
pub enum Op1Error {
    InternalServerError(crate::error::InternalServerError),
}
```

Having the `From<InternalServerError>` trait there would be nice (and should always be possible since the Enum is created from a list of error types in smithy):
```
impl From<InternalServerError> for Op1Error {
  fn from(e : InternalServerError) -> Self { Self::InternalServerError(e) }
}
```

david-perez:
Closes #1160.

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
@@ -83,6 +83,15 @@ class ServerCombinedErrorGenerator(
                 }
             }
         }
+
+        for (error in errors) {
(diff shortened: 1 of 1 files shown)


crisidev (approved):

llgerard (commented):
Nice, thank you.

Issue #1147, opened 2022-02-02
Protocol tests and floating point comparisons


david-perez:
Server protocol tests are using the `PartialEq` implementation on floating point types because they `assert_eq` on the containing operation input shape.

Client protocol tests, on the other hand, _always_ iterate over the shape's members and use [`aws_smithy_protocol_test::FloatEquals`](https://docs.rs/aws-smithy-protocol-test/latest/aws_smithy_protocol_test/trait.FloatEquals.html).

> This trait differs in float’s default PartialEq implementation by considering all NaN values to be equal.

I stumbled upon this earlier in https://github.com/awslabs/smithy/pull/1040#discussion_r780418707. I think this fix is fair.

However, I don't see this trait being used _recursively_ in the client tests i.e. the shape containing floating point types could be nested.

https://github.com/awslabs/smithy-rs/blob/4b8381b22c8d4b8eebea60bdc6a666bb75dece57/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/ProtocolTestGenerator.kt#L326-L335

PR #1177, opened 2022-02-10
Consider `NaN` to be equal to itself in server protocol tests

david-perez:
The client protocol tests use the
`aws_smithy_protocol_test::FloatEquals` for this [0].

Note we're only applying this to direct floating point shape members,
i.e. this commit _does not_ address #1147.

[0]: https://docs.rs/aws-smithy-protocol-test/latest/aws_smithy_protocol_test/trait.FloatEquals.html


crisidev (commented):
Isn't this semantically wrong? https://doc.rust-lang.org/std/primitive.f32.html

david-perez:
@crisidev In protocol tests we don't want to do any math on `NaN` values, nor do we care about the specific bits of a _particular_ `NaN`, or about what calculation poisoned the result. We do need however a mechanism to verify that we deserialize `"NaN"` on the wire to a `NaN` floating point value, i.e. to check for the _presence_ of _a_ (any) `NaN` value. Using `aws_smithy_protocol_test::FloatEquals` for this purpose is fair use.

I found this post interesting https://stackoverflow.com/questions/10034149/why-is-nan-not-equal-to-nan

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -357,7 +359,40 @@ class ServerProtocolTestGenerator(
...
+                                    #{AssertEq}(parsed.$memberName, expected.$memberName, "Unexpected value for `$memberName`");
+                                    """,
+                                *codegenScope
+                            )
+                        }
+                    }
+                }
+            } else {
+                rustWriter.rustTemplate("#{AssertEq}(parsed, expected);", *codegenScope)
+            }
```
rcoh, line 395: should we being extracting some sort of `ProtocolTestCore`? The client has this behavior too right?

rcoh (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
david-perez: 100%. It's slightly different in that the client _always_ iterates over the members to do the equality comparison.

But yes, once I get all the protocol tests for restJson1 passing, I plan on extracting a lot of shared functionality between client and server protocol tests.

crisidev (approved):

PR #1179, opened 2022-02-10
Autodump generated files

### LINKED ISSUE #1178: Print generated sources automatically in unit tests
Currently, we have to manually print generated sources at the end of unit tests. Example below:

https://github.com/awslabs/smithy-rs/blob/c17d58fd0862194f625bcc0e8f4697d231a3ca73/codegen/src/test/kotlin/software/amazon/smithy/rust/codegen/generators/http/RequestBindingGeneratorTest.kt#L324-L328

It would be nice if the unit tests knew what files they generated and automatically printed these paths. https://github.com/awslabs/smithy-rs/pull/1174#discussion_r803944823

> The TestWriterDelegate has the baseDir, and from that, we should be able to walk through all of src/*.rs in compileAndTest() and print out links automatically. I don't think it would have a way to know all the file names upfront since the generator classes can create their own files without it being aware.

rcoh:
## Motivation and Context
- Fixes #1178 

## Description
Print `file:///` automatically from TestWriterDelegator

## Testing
- [x] ran test locally, verified that the list of files was printed and the links worked

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/testutil/Rust.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/testutil/Rust.kt
@@ -158,7 +158,7 @@ class TestWriterDelegator(private val fileManifest: FileManifest, symbolProvider
     RustCrate(fileManifest, symbolProvider, DefaultPublicModules) {
     val baseDir: Path = fileManifest.baseDir
 
-    fun generatedFiles(): List<Path> = fileManifest.files.toList()
+    fun generatedFiles(): List<Path> = fileManifest.files.toList().sorted()
 }
 
 /**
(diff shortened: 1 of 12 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/testutil/Rust.kt
```diff
@@ -177,6 +179,10 @@ fun TestWriterDelegator.compileAndTest(runClippy: Boolean = false) {
         manifestCustomizations = emptyMap(),
         libRsCustomizations = listOf(),
     )
+    println("Generated files:")
+    generatedFiles().forEach { path ->
```
jdisanti, line 183: Is this already sorted by name?

jdisanti (approved):
Love it!

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/testutil/Rust.kt
rcoh: no it's actually nondeterministic 😅 probably nice for usability to sort it

PR #1186, opened 2022-02-13
Server allow `unused_variables` in `serverSerializeResponse` and `serverParseRequest`

guymguym:
## Motivation and Context
- Get rid of warnings in server-codegen operation_ser.rs.

## Description
- Allow unused_variables for output arg in `serverSerializeResponse()`.
- Used the same approach as found in `serverParseRequest()` -
https://github.com/awslabs/smithy-rs/blob/548f6ed4307faae5bb98e431db96ae90a59e489d/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt#L318

## Testing
- Compiled and observed the warnings go away.

## Checklist

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
@@ -316,14 +316,13 @@ private class ServerHttpProtocolImplGenerator(
         val inputShape = operationShape.inputShape(model)
         val inputSymbol = symbolProvider.toSymbol(inputShape)
         val includedMembers = httpBindingResolver.requestMembers(operationShape, HttpLocation.DOCUMENT)
-        val unusedVars = if (includedMembers.isEmpty()) "##[allow(unused_variables)] " else ""
 
         return RuntimeType.forInlineFun(fnName, operationDeserModule) {
             Attribute.Custom("allow(clippy::unnecessary_wraps)").render(it)
             it.rustBlockTemplate(
                 """
                 pub async fn $fnName<B>(
-                    ${unusedVars}request: &mut #{AxumCore}::extract::RequestParts<B>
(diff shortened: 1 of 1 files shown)


david-perez (changes requested):
Not having members bound to the document is _not a sufficient condition_ for the `output` argument to end up unused.

An example illustrating this is given by:

```
$version: "1.0"

use aws.protocols#restJson1

@restJson1
@title("SimpleService")
service SimpleService {
    operations: [
        AnOperation,
    ],
}

@http(uri: "/operation", method: "GET")
operation AnOperation {
    input: AnOperationInput,
    output: AnOperationOutput,
}

structure AnOperationInput { }

structure AnOperationOutput {
    @httpHeader("My-Header")
    headerBoundData: String
}
```

Here, there aren't any output members bound to the document, so with this patch, `#[allow(unused_variables)]` will get added; however, `output` _is_ used because it has a header-bound member.

```rust
pub fn serialize_an_operation_response(
    #[allow(unused_variables)] output: crate::output::AnOperationOutput,
) -> std::result::Result<
    axum_core::response::Response,
> {
    Ok({
        #[allow(unused_mut)]
        let mut builder = http::Response::builder();
        builder = crate::http_serde::add_headers_an_operation(&output, builder)?;
        builder = aws_smithy_http::header::set_response_header_if_absent(
            builder,
            http::header::CONTENT_TYPE,
            "application/json",
        );
        let payload =
            crate::operation_ser::serialize_structure_crate_output_an_operation_output(&output)?;
        builder.body(aws_smithy_http_server::body::to_boxed(payload))?
    })
}
```


I think it's best we not be smart and simplify the codegen, and _unconditionally_ add `#[allow(unused_variables)]` both here and in `serverParseRequest`. There are multiple instances of the client adding attributes like these unconditionally just to simplify the codegen.

https://github.com/awslabs/smithy-rs/blob/548f6ed4307faae5bb98e431db96ae90a59e489d/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt#L318

david-perez (approved):
Thank you.

Issue #1150, opened 2022-02-03
Server operation handlers need to be able to result with internal error


guymguym:
Servers might have internal errors which are not defined in the API as functional errors.
However, currently if the operation has no errors defined in the API, the handler type will not allow to return any error, and the return type of the handler will be expected to be the output type directly, which makes it impossible to signal an internal error has occurred. Only panic is possible which is not desired for most servers.
Is there's any way to map such errors in the server which does not panic?
Maybe adding an `unhandled` error type like the client has?

That said - when I look at the S3 smithy model, I find that it lacks a lot of error definitions for operations.
Here is a list of all the operations that have errors defined:
```
$ node -e 'm=require("../smithy-rs/aws/sdk/aws-models/s3.json"); Object.entries(m.shapes).forEach(([n,s])=>{ if (s.type==="operation" && s.errors) console.log(n, "---", s.errors.map(x=>x.target).join(",")); })'
com.amazonaws.s3#AbortMultipartUpload --- com.amazonaws.s3#NoSuchUpload
com.amazonaws.s3#CopyObject --- com.amazonaws.s3#ObjectNotInActiveTierError
com.amazonaws.s3#CreateBucket --- com.amazonaws.s3#BucketAlreadyExists,com.amazonaws.s3#BucketAlreadyOwnedByYou
com.amazonaws.s3#GetObject --- com.amazonaws.s3#InvalidObjectState,com.amazonaws.s3#NoSuchKey
com.amazonaws.s3#GetObjectAcl --- com.amazonaws.s3#NoSuchKey
com.amazonaws.s3#HeadBucket --- com.amazonaws.s3#NotFound
com.amazonaws.s3#HeadObject --- com.amazonaws.s3#NotFound
com.amazonaws.s3#ListObjects --- com.amazonaws.s3#NoSuchBucket
com.amazonaws.s3#ListObjectsV2 --- com.amazonaws.s3#NoSuchBucket
com.amazonaws.s3#PutObjectAcl --- com.amazonaws.s3#NoSuchKey
com.amazonaws.s3#RestoreObject --- com.amazonaws.s3#ObjectAlreadyInActiveTierError
```
This is clearly a tiny fraction of the real errors that this API can return.
See https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

I think that I could bypass this limitation if the server handlers could return a generic error in addition to the predefined ones...
@david-perez @crisidev WDYT?

crisidev:
[...]

PR #1190, opened 2022-02-15
Add decorator to allow to make all operations fallible

crisidev:
## Motivation and Context
Force all operations in the model to be fallible.

When this decorator is applied, operations that do not have a Smithy error attatched, will return `Result<OperationOutput, InternalServerError>`.

To enable this decorator, create a file called `META-INF/service/software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator` containing `codegen-server/src/main/resources/META-INF/services/software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator`.

Closes: #1150

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/FallibleOperationsDecorator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/FallibleOperationsDecorator.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package software.amazon.smithy.rust.codegen.server.smithy.customizations
+
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.MemberShape
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.model.traits.ErrorTrait
+import software.amazon.smithy.model.traits.RequiredTrait
+import software.amazon.smithy.model.transform.ModelTransformer
+import software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator
+
+/**
+ * Force all operations in the model to be fallible.
+ *
+ * When this decorator is applied, operations that do not have a Smithy error attatched,
+ * will return `Result<OperationOutput, InternalServerError>`.
+ *
+ * To enable this decorator, create a file called `META-INF/services/software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator`
+ * containing `codegen-server/src/main/resources/META-INF/services/software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator`.
+ */
+class FallibleOperationsDecorator : RustCodegenDecorator {
+    override val name: String = "FallibleOperations"
+    override val order: Byte = 0
+
+    override fun transformModel(service: ServiceShape, model: Model): Model {
+        val namespace = service.id.getNamespace()
+        val message = MemberShape.builder().id("$namespace#InternalServerError\$message")
+            .target("smithy.api#String").addTrait(RequiredTrait()).build()
+        val errorShape = StructureShape.builder().id("$namespace#InternalServerError")
+            .addTrait(ErrorTrait("server")).addMember(message).build()
+        val modelShapes = model.toBuilder().addShapes(listOf(errorShape)).build()
+        return ModelTransformer.create().mapShapes(modelShapes) { shape ->
+            if (shape is OperationShape && shape.errors.isEmpty()) {
+                shape.toBuilder().addError(errorShape).build()
+            } else {
+                shape
+            }
+        }
+    }
+}


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/FallibleOperationsDecorator.kt
```diff
@@ -0,0 +1,46 @@
...
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.model.traits.ErrorTrait
+import software.amazon.smithy.model.traits.RequiredTrait
+import software.amazon.smithy.model.transform.ModelTransformer
+import software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator
+
+/**
+ * Force all operations in the model to be fallible.
```
rcoh, line 19: `Add at least one error to all operations in the model`? A bit more explicit about what this does

rcoh (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/FallibleOperationsDecorator.kt
```diff
@@ -0,0 +1,46 @@
...
+        val errorShape = StructureShape.builder().id("$namespace#InternalServerError")
+            .addTrait(ErrorTrait("server")).addMember(message).build()
+        val modelShapes = model.toBuilder().addShapes(listOf(errorShape)).build()
+        return ModelTransformer.create().mapShapes(modelShapes) { shape ->
+            if (shape is OperationShape && shape.errors.isEmpty()) {
+                shape.toBuilder().addError(errorShape).build()
+            } else {
+                shape
+            }
+        }
```
rcoh, line 44: you might want to refactor this out into at least a function so it's testable (It should be pretty quick to make a model and run it through this function to verify that all operations have errors)

crisidev: I will add a test, you are right.

david-perez (approved):

guymguym:
@crisidev is this path `META-INF/service/software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator` meant to be from the smithy-rs project root? or from the root of the build task, and then it will read from under codegen-server-test/...?

rcoh:
assuming you're using the plugin from a separate gradle project, this `META-INF` would be within resources—you can see the `aws/sdk` directory as an example: https://github.com/awslabs/smithy-rs/blob/main/aws/sdk-codegen/src/main/resources/META-INF/services/software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator

rcoh (approved):
LGTM!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/FallibleOperationsDecorator.kt
```diff
@@ -0,0 +1,54 @@
...
+import software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator
+
+/**
+ * Add at least one error to all operations in the model.
+ *
+ * When this decorator is applied, operations that do not have a Smithy error attatched,
+ * will return `Result<OperationOutput, InternalServerError>`.
+ *
+ * To enable this decorator, create a file called `META-INF/services/software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator`
+ * containing `codegen-server/src/main/resources/META-INF/services/software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator`.
```
guymguym, line 25: @crisidev shouldn't this change to match the updated decorator class name - FallibleOperationsDecorator?

I am trying to get this to work from my separate gradle build root, and I see the jar containing the meta-inf file, but the decorator is not running (I added a print to make sure). I am probably missing something regarding the gradle build structure but hard to figure out what. Any chance you could show me how this can be applied to the codegen-server-test project perhaps? Sorry for the hassle, but I'm a bit lost after trying a bunch of things. Thanks

crisidev: Damn, this is completely wrong.. it should contain the java path of the decorator you want to use: `software.amazon.smithy.rust.codegen.server.smithy.customizations.FaillibleOperationsDecorator`.

guymguym:
Hey @crisidev - I still can't make this work for me.

These are the steps I use to apply the decorator to codegen-server-test:

```bash
C1="software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator"
C2="software.amazon.smithy.rust.codegen.server.smithy.customizations.FallibleOperationsDecorator"
D="codegen-server-test/src/main/resources/META-INF/services"
mkdir -p "$D"
echo $C2 > "$D/$C1"
./gradlew :codegen-server-test:test
grep -r InternalServerError codegen-server-test/build/
```

Does this work for you? 
I expected at least `operation Healthcheck` from simple.smithy to have the internal error because it has no errors of its own.

I already tried a bunch of my monkey-in-cockpit tricks on the build.gradle.kts but no luck.
Maybe you see what I'm missing?
Thanks

guymguym:
@crisidev OK I found that when I put the META-INF service under `codegen-server/` then it works and the decorator is loaded and the internal errors are generated !!! 🥳

(I just change in the script above to `D="codegen-server/src/main/resources/META-INF/services"`)

But I can't make it work when I add it to the upper project codegen-server-test....
Must be a gradle hurdle.

crisidev:
To make it work everywhere, you can probably add it under the META-INF of `codegen-server`.. Not sure if that will make it work..

guymguym:
It seems to work, but is that how you mean that a server project will enable optional features - by changing the server-codegen project itself? Sounds hard to maintain. Isn't there a way to enable decorators from the caller project?

PR #1194, opened 2022-02-16
RFC: Independent Crate Versioning

jdisanti:
## Motivation and Context
This adds an RFC describing how we should approach versioning and publishing to crates.io long-term. It incorporates feedback [from a previous RFC](https://github.com/awslabs/smithy-rs/pull/759) by breaking the requirement that `major.minor` versions should be the same across all crates in the SDK.

[Rendered](https://github.com/awslabs/smithy-rs/blob/jdisanti-rfc-crates-io/design/src/rfcs/rfc0012_independent_crate_versioning.md)


design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+- `minor`:
+    - Changes that aren't breaking
+    - Dependency updates for dependencies that are part of the public API
+    - Model updates with API changes
+    - For code-generated crates: when a newer version of [smithy-rs] is used to generate the crate
+- `patch`:
+    - Bug fixes that do not break backwards compatibility
+    - Model updates that _only_ have documentation changes
+
+During phase 3, bumps to the `major` version must be coordinated across all SDK and runtime crates.
```
Velfi, line 65: Another way of saying this is that "if we break one SDK/runtime crate, we break them all" right?

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Release Identification
+
+Since there will no longer be one SDK "version", release tags will be dates in `YYYY-MM-DD` format
+rather than version numbers. Additionally, the SDK's user agent string will need to include a separate
+service version number (this requirement has already been implemented).
+
+### Yanking
+
+It must be possible to yank an entire release with a single command-line. The publisher tool must
```
Velfi, line 75: suggestion:
"with a single command-line" -> "with a single action"

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+be updated to understand which crate versions were released with a given release tag, and be able to
+yank that entire tag worth of crates.
```
Velfi, line 77: suggestion:
"tag worth" -> "tag's worth"

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+[aws-sdk-s3]
+version = "0.89.0"
+model_hash = "<hash>"
+
+# ...
+```
+
+The auto-sync tool is responsible for maintaining this file. When it generates a new SDK, it will take
+the version numbers from runtime crates directly, and it will use the rules from the next section to determine
+the version numbers for the generated crates.
```
Velfi, line 114: This is gonna be so cool!

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Versioning
+
+Cargo uses [semver] for versioning, with a `major.minor.patch-pre` format:
+- `major`: Incompatible API changes
+- `minor`: Added functionality in backwards compatible manner
+- `patch`: Backwards compatible bug fixes
+- `pre`: Pre-release version tag (omitted for normal releases)
+
+In the new versioning strategy, the `minor` version number will no longer be coordinated across
```
rcoh, line 40: is this the case for all three phases?

design/src/rfcs/rfc0012_crates_io_publishing.md
rcoh, line 65: is this a requirement of this proposal? I could definitely imagine wanting to major version bump an individual service that made a breaking change, eg.

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+    - For code-generated crates: when a newer version of [smithy-rs] is used to generate the crate
+- `patch`:
+    - Bug fixes that do not break backwards compatibility
+    - Model updates that _only_ have documentation changes
+
+During phase 3, bumps to the `major` version must be coordinated across all SDK and runtime crates.
+
+### Release Identification
+
+Since there will no longer be one SDK "version", release tags will be dates in `YYYY-MM-DD` format
```
rcoh, line 69: this refers to tags on GitHub, not cargo versions, right?

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+- Creation of version bump process for code generated crates
+- Enforcement of version bump process in runtime crates
+- Yanking of versions associated with a release tag
+
+### Version Tracking
+
+A new manifest file will be introduced in the root of [aws-sdk-rust] named `versions.toml` that describes
+all versioning information for any given commit in the repository. In the main branch, the `versions.toml`
+in tagged commits will become the source of truth for which crate versions belong to that release, as well
+as additional metadata that's required for maintaining version process in the future.
```
rcoh, line 93: recommend adding a callout here about removing the smithy-rs-head version

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+version = "0.40.0"
+
+[aws-sdk-s3]
+version = "0.89.0"
+model_hash = "<hash>"
+
+# ...
+```
+
+The auto-sync tool is responsible for maintaining this file. When it generates a new SDK, it will take
```
rcoh, line 112: will there be some way to override / hint?

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+the version numbers from runtime crates directly, and it will use the rules from the next section to determine
+the version numbers for the generated crates.
+
+### Versioning for Code Generated Crates
```
rcoh, line 116:
```suggestion
### Versioning for Code Generated (SDK Service) Crates
```

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+
+Code generated crates will have their `minor` version bumped when the version of [smithy-rs] used to generate
+them changes, or when model updates with API changes are made. Three pieces of information are required to
+handle this process: the previously released version number, the [smithy-rs] version used to generate the code,
+and the level of model updates being applied. For this last one, if there are multiple model updates that
+affect only documentation, but then one model update that affects an API, then as a whole they will be
+considered as affecting an API and require a `minor` version bump.
+
+The previously released version number will be retrieved from crates.io using its API. The [smithy-rs] version
```
rcoh, line 125: conceivably we could track the previously deployed version in versions.toml right?

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+all versioning information for any given commit in the repository. In the main branch, the `versions.toml`
+in tagged commits will become the source of truth for which crate versions belong to that release, as well
+as additional metadata that's required for maintaining version process in the future.
+
+This format will look as follows:
+```toml
+smithy_rs_version = "<release-tag|commit-hash>"
+
+[aws-smithy-types]
+version = "0.50.1"
```
rcoh, line 100: the runtime crate versions here are just for consistency right? they aren't actually required for the system to function?

We _could_ consider tracking `published = ...` and `next = ...` in this file which removes the crates.io API dependency (or would just provide a spot to verify)

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+
+This introduces a new problem where a developer can forget to bump a runtime crate version, so a method of
+process enforcement needs to be introduced. This will be done through CI when merging into `smithy-rs/main`
+and repeated when merging into `aws-sdk-rust/main`.
+
+The following checks need to be run for runtime crates:
+
+![Phase 1: How to validate a runtime version bump](rfc0012_crates_io_publishing/phase1_runtime_crate_version_checks.svg)
+
+- __A: Crate has changed?__ The crate's source files and manifest will be hashed for the previous version
```
rcoh, line 152: whats manifest here? `Cargo.toml`?

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+- [ ] Establish initial `versions.toml` in `aws-sdk-rust/main`
+- [ ] Set version numbers in runtime crates in [smithy-rs]
+- [ ] Update the auto-sync tool to generate `versions.toml`
+- [ ] Create CI tool to check runtime crate version
+  - [ ] Integrate with `smithy-rs/main` CI
+  - [ ] Integrate with `aws-sdk-rust/main` CI
+- [ ] Update CI to verify no older runtime crates are used. For example, if `aws-smithy-client` is bumped to
+      `0.50.0`, then verify no crates (generated or runtime) depend on `0.49.0` or lower.
+
+**Estimate:** 2-4 dev weeks
```
rcoh, line 189: is this a 50th-95th bound?

design/src/rfcs/rfc0012_crates_io_publishing.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Yanking
+
+The publisher tool will be updated to read the `versions.toml` to yank all versions published in a release.
+This process will look as follows:
+1. Take a path to a local clone of the [aws-sdk-rust] repository
+2. Confirm the branch is `main` and that HEAD is tagged as a release
+3. Read `versions.toml` and print out summary of crates to yank
+4. Confirm with user before proceeding
+5. Yank crates
```
rcoh, line 176: I'm a little confused about this—I think I'm missing how versions.toml tells you which crates were released and which were not given the current schema—do you have to diff the files?

design/src/rfcs/rfc0012_crates_io_publishing.md
jdisanti: Yeah, essentially.

design/src/rfcs/rfc0012_crates_io_publishing.md
jdisanti: Yes

design/src/rfcs/rfc0012_crates_io_publishing.md
jdisanti: I suppose this isn't a hard requirement, but we would be walking away from the ability to make a large refactor and coordinate it under a major version release like the other SDKs do with their V2/V3 releases.

design/src/rfcs/rfc0012_crates_io_publishing.md
jdisanti: That's right

design/src/rfcs/rfc0012_crates_io_publishing.md
jdisanti: The idea was to have all the version numbers for a given release tag centralized in one place, so that's why the runtime crate versions are in here. Nothing else is requiring them to be in this file.

design/src/rfcs/rfc0012_crates_io_publishing.md
jdisanti: Yeah. _Manifest_ is the name for a `Cargo.toml` file according to https://doc.rust-lang.org/cargo/reference/manifest.html:

> The Cargo.toml file for each package is called its _manifest_.

design/src/rfcs/rfc0012_crates_io_publishing.md
jdisanti: Yes

design/src/rfcs/rfc0012_crates_io_publishing.md
jdisanti: Let's say we have a bad release at tag `bad` that we want to yank. The `versions.toml` file at that tag will have all the crate version numbers associated with that release, but if the publish failed half way through, then some of those might be published to crates.io while others might not be. But that's OK since the yank tool can just check that the version is actually published using the crates.io API before yanking it.

That said, I'm now confused by (2) above. I think I'll revise this to make sure you're on a release tag rather than on `main`.

design/src/rfcs/rfc0012_unsynchronized_versioning.md
```diff
@@ -0,0 +1,207 @@
...
+the version numbers for the generated crates.
+
+### Versioning for Code Generated (SDK Service) Crates
+
+Code generated crates will have their `minor` version bumped when the version of [smithy-rs] used to generate
+them changes, or when model updates with API changes are made. Three pieces of information are required to
+handle this process: the previously released version number, the [smithy-rs] version used to generate the code,
+and the level of model updates being applied. For this last one, if there are multiple model updates that
+affect only documentation, but then one model update that affects an API, then as a whole they will be
+considered as affecting an API and require a `minor` version bump.
```
Velfi, line 122: _paraphrasing this to make sure I understand_
>"if we update 2+ models in a release, and at least 1 of those models has API changes, all updated models will receive a `minor` version bump"

Is that right?

design/src/rfcs/rfc0012_unsynchronized_versioning.md
```diff
@@ -0,0 +1,207 @@
...
+
+The previously released version number will be retrieved from crates.io using its API. The [smithy-rs] version
+used during code generation will become a build artifact that is saved to `versions.toml` in [aws-sdk-rust].
+During phase 1, the tooling required to know if a model is a documentation-only change will not be available,
+so all model changes will result in a `minor` version bump during this phase.
```
Velfi, line 127: Are we ever going to run into a situation where we start releasing a version and then it chokes on the crates.io upload because of some problem with our code? If yes, would the versioning get messed up?

design/src/rfcs/rfc0012_unsynchronized_versioning.md
```diff
@@ -0,0 +1,207 @@
...
+
+- __A: smithy-rs changed?__: Compare the `smithy_rs_version` in the previous `versions.toml` with the
+  next `versions.toml` file, and if the values are different, consider [smithy-rs] to have changed.
+- __B: model changed?__: Similarly, compare the `model_hash` for the crate in `versions.toml`.
+
+### Versioning for Runtime Crates
+
+The old scheme of all runtime crates in [smithy-rs] having a fake `0.0.0-smithy-rs-head` version number with
+a build step to replace those with a consistent `major.minor` will be removed. These runtime crates will begin
+having their actual next version number in the Cargo.toml file in smithy-rs.
```
Velfi, line 141: I like this change

design/src/rfcs/rfc0012_unsynchronized_versioning.md
```diff
@@ -0,0 +1,207 @@
...
+This introduces a new problem where a developer can forget to bump a runtime crate version, so a method of
+process enforcement needs to be introduced. This will be done through CI when merging into `smithy-rs/main`
+and repeated when merging into `aws-sdk-rust/main`.
+
+The following checks need to be run for runtime crates:
+
+![Phase 1: How to validate a runtime version bump](rfc0012_unsynchronized_versioning/phase1_runtime_crate_version_checks.svg)
+
+- __A: Crate has changed?__ The crate's source files and manifest will be hashed for the previous version
+  and the next version. If these hashes match, then the crate is considered unchanged.
```
Velfi, line 152: Cargo has something like this for incremental compilation right? It'd be kinda neat if we could use that functionality to do this.

design/src/rfcs/rfc0012_unsynchronized_versioning.md
```diff
@@ -0,0 +1,207 @@
...
+- __B: Minor bumped?__ The previous version is compared against the next version to see if the minor version
+  number was bumped.
+- __C: Version changed?__ The previous version is compared against the next version to see if it changed.
+- __D: Patch bumped?__ The previous version is compared against the next version to see if the patch version
+  number was bumped.
+- __E: Semverver passes?__ Runs [rust-semverver] against the old and new versions of the crate.
+  - If semverver fails to run (for example, if it needs to be updated to the latest nightly to succeed),
+    then fail CI saying that either semverver needs to be updated, or that a minor version bump is required.
```
Velfi, line 160:
```suggestion
    then fail CI saying that semverver needs maintenance.
```

design/src/rfcs/rfc0012_unsynchronized_versioning.md
```diff
@@ -0,0 +1,207 @@
...
+- [ ] Update CI to verify no older runtime crates are used. For example, if `aws-smithy-client` is bumped to
+      `0.50.0`, then verify no crates (generated or runtime) depend on `0.49.0` or lower.
+
+**Estimate:** 2-4 dev weeks
+
+Phase 2: Stability and 1.x
+--------------------------
+
+When stabilizing to 1.x, the version process will stay the same, but with the introduction of the greater than zero
+major version number, the minor version bumps caused by version bumping runtime crates, updating models, or changing
```
Velfi, line 200:
```suggestion
When stabilizing to 1.x, the version process will stay the same, but with the minor version bumps caused by version bumping runtime crates, updating models, or changing
```

The current wording feels a bit redundant but I don't feel super strongly about this

design/src/rfcs/rfc0012_unsynchronized_versioning.md
```diff
@@ -0,0 +1,207 @@
+RFC: Unsynchronized Crate Versioning
```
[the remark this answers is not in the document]
jdisanti: I like it.

design/src/rfcs/rfc0012_unsynchronized_versioning.md
jdisanti: So imagine we have the following updates to the S3 model, and the S3 version was previously `1.2.3`:

1. Documentation
2. Documentation
3. API change
4. API change

The new version in this case, will be exactly one minor version bump: `1.3.0`

design/src/rfcs/rfc0012_unsynchronized_versioning.md
jdisanti: It would be neat, but I think it would require building a tool that utilizes the compiler internals, and introduce the maintenance burden that comes with that kind of tooling.

design/src/rfcs/rfc0012_unsynchronized_versioning.md
```diff
@@ -0,0 +1,207 @@
...
+    - For code-generated crates: when a newer version of [smithy-rs] is used to generate the crate
+- `patch`:
+    - Bug fixes that do not break backwards compatibility
+    - Model updates that _only_ have documentation changes
+
+During phase 3, bumps to the `major` version must be coordinated across all SDK and runtime crates.
+
+### Release Identification
+
+Since there will no longer be one SDK "version", release tags will be dates in `YYYY-MM-DD` format
```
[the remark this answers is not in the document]
jdisanti: SDK release automation is limited to one release per day, but there is the question of what this would look like in the event of a manual release taking place in addition to an automatic release. I don't really have a good answer for this currently.

design/src/rfcs/rfc0012_unsynchronized_versioning.md
jdisanti: The point about a minor version bump being an option should stay, but I like the suggestion of "needs maintenance" over "to be updated".

rcoh (approved):

PR #1197, opened 2022-02-17
Replenish cross-request retry allowance on successful response

jdisanti:
## Motivation and Context
This PR fixes an issue discovered in https://github.com/awslabs/aws-sdk-rust/issues/440 where the cross-request retry allowance was never reimbursed after a successful response, so eventually it would zero out and cause retry to become permanently disabled for a client.

## Testing
- Added a unit test to verify reimbursement

## Checklist


aws/sdk/integration-tests/dynamodb/tests/movies.rs
```diff
@@ -125,8 +125,9 @@ where
         &self,
         response: Result<&SdkSuccess<DescribeTableOutput>, &SdkError<DescribeTableError>>,
     ) -> RetryKind {
-        match self.inner.classify(response.clone()) {
+        match self.inner.classify(response) {
```
rcoh, line 128: we have a few other classify implementations (imds at least) that need the same treatment

rcoh (commented):
LGTM—can you audit the other classifiers? I like the idea of adding another variant—we could use `Ok` to do that, but this is more flexible.

aws/sdk/integration-tests/dynamodb/tests/movies.rs
jdisanti: I'm worried about this breaking change silently breaking customers' customer retry behavior, so I'm going to rename `NotRetryable` to something else to force a compilation error so that they revisit their retry classification implementations. This should also have the side effect of revealing all of the broken classify implementations in our own code. It looks like there are a few of them in our credential providers chain.

aws/rust-runtime/aws-config/src/imds/client.rs
```diff
@@ -977,6 +982,36 @@ pub(crate) mod test {
         connection.assert_requests_match(&[]);
     }
 
+    /// Successful responses should classify as `RetryKind::Unnecessary`
+    #[test]
+    fn successful_response_properly_classified() {
+        use aws_smithy_http::retry::ClassifyResponse;
+
+        struct GenericAppeaser; // Need a type to appease the generics; don't care what it is
```
rcoh, line 990: can probably just use `()`?

rcoh (approved):

PR #1199, opened 2022-02-17
Add server SDK constraint traits RFC

david-perez:
[Rendered](https://github.com/awslabs/smithy-rs/blob/118ba860a3f989e65362179bb815deb641b1a299/design/src/rfcs/rfc0025_constraint_traits.md).


design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. Should we code-generate unsigned integer types (`u16`, `u32`, `u64`) when
+   the `range` trait is applied with `min` set to a value greater than or equal
+   to 0?
+    - A user has even suggested to use the `std::num::NonZeroUX` types (e.g.
+      [`NonZeroU64`]) when `range` is applied with `min` set to a value greater
+      than 0.
+2. In request deserialization, should we fail with the first violation and
+   immediately render a response, or attempt to parse the entire request and
+   provide a complete and structured report?
```
crisidev, line 380: For the sake of security I think we should bail out as soon as we find the first validation error.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+* Validation remains an internal detail of the framework. If the semantics of a
+  constraint trait change, the behavior of the service is still
+  backwards-incompatibly affected, but user code is not.
+* Less "invasive". Baking validation in the generated type might be deemed as
+  the service framework overreaching responsibilities.
+
+Comparative disadvantages:
+
+* It becomes _possible_ to send responses with invalid operation outputs. All
+  the service framework could do is log the validation errors.
```
crisidev, line 416: Should we really care about this? At the end the server is the authority and we do not validate outputs..

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+------------------
+
+An alternative design with less public API surface would be to perform
+constraint validation at request deserialization, but hand over a regular
+"loose" type (e.g. `String` instead of `NiceString`) that allows for values
+violating the constraints. If we were to implement this approach, we can
+implement it by wrapping the incoming value in the aforementioned tuple struct
+to perform the validation, and immediately unwrap it.
+
+Comparative advantages:
```
crisidev, line 405: I think the first option follows a better approach in terms of design, but I would be fine also doing 2. if we find problems with 1. Both options solve the problem in a good way :)

design/src/rfcs/rfc0012_constraint_traits.md
david-perez: At the very least I'd consider it to likely be an implementation error from the user with regards to what they modeled, and as such performing validation and logging warnings would help them identify it. So we _should_ care, but very little.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+        Self::try_from(inner)
+    }
+}
+
+impl TryFrom<String> for NiceString {
+    type Error = NiceStringValidationError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let num_code_points = value.chars().count();
+        if 1 <= num_code_points && num_code_points <= 69 {
```
Velfi, line 121: Consider using [`RangeInclusive::contains`](https://doc.rust-lang.org/nightly/std/ops/struct.RangeInclusive.html) here

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+            Ok(Self(value))
+        } else {
+            Err(NiceStringValidationError::LengthViolation(num_code_points))
+        }
+    }
+}
+```
+
+(Note that we're using the _linear_ time check `chars().count()` instead of
```
Velfi, line 130: Good catch! Unicode is hard.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+[`TryFrom`]`<InnerType>` for the generated struct, which makes it convenient to use
+the [`?` operator for error propagation]. Each constrained struct will have a
+related `std::error::Error` enum type to signal the _first_ parsing failure,
+with one enum variant per applied constraint trait:
+
+```rust
+pub enum NiceStringValidationError {
+    /// Validation error holding the number of Unicode code points found, when a value between `1` and
+    /// `69` (inclusive) was expected.
+    LengthViolation(usize),
```
Velfi, line 148: It might be too much information, but I always prefer errors that explain not only what the invalid value was, but what counts as a valid value.

Velfi: Nevermind, I see in the next paragraph that you say you'll do this.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+
+`pattern` trait
+---------------
+
+The [`pattern` trait]
+
+> restricts string shape values to a specified regular expression.
+
+We will implement this by using the `regex`'s crate [`is_match`]. We will use
+[`lazy_static!`] to compile the regex only the first time it is required.
```
Velfi, line 309: [`once_cell`](https://lib.rs/crates/once_cell) is now preferred to `lazy_static`

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+
+[the spec]: https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html?highlight=enum#trait-precedence
+
+Unresolved questions
+--------------------
+
+1. Should we code-generate unsigned integer types (`u16`, `u32`, `u64`) when
+   the `range` trait is applied with `min` set to a value greater than or equal
+   to 0?
+    - A user has even suggested to use the `std::num::NonZeroUX` types (e.g.
```
Velfi, line 375: Personally, I'm all for using the `num` crate in situations like this.

david-perez: What specifically would you use from the `num` crate?

design/src/rfcs/rfc0012_constraint_traits.md
adamthom-amzn: I have the opposite opinion. Validation doesn't protect the parser (you've already parsed), and it's extremely useful to applications that are gathering user input, such as web form, to indicate all invalid inputs so that they can be fixed by the user at once. This is why the [ValidationException built into Smithy](https://github.com/awslabs/smithy/blob/main/smithy-validation-model/model/smithy.framework.validation.smithy) has the structure that it does.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+   struct is in a transient invalid state. However:
+    - (2) is arguably a modelling mistake and a separate struct to represent
+      the transient state would be a better approach,
+    - the user can always use `unsafe` Rust to bypass the validation; and
+    - adding this constructor is a backwards-compatible change, so it can
+      always be added later if this feature is requested.
+
+[`NonZeroU64`]: https://doc.rust-lang.org/std/num/struct.NonZeroU64.html
+
+Alternative design
```
jonhoo, line 395: I prefer the primary design, though with the addendum that I think we should provide conversions back to the primitive type. Something like `fn into_unconstrained`.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+We will enforce the length constraint by calling `len()` on Rust's `Vec`
+(`list` and `set` shapes), `HashMap` (`map` shapes), `String` (`string`
+shapes), and our [`aws_smithy_types::Blob`] (`bytes` shapes).
+
+`pattern` trait
+---------------
+
+The [`pattern` trait]
+
+> restricts string shape values to a specified regular expression.
```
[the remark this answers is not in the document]
adamthom-amzn: https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html#pattern-trait - yes, ECMA 262 is the format of choice. You can, and should, prefer a non-backtracking implementation, like `re2` (I'm not sure what the `regex` crate uses).

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+        let mut formatter = f.debug_struct("NiceStringValidationError");
+        formatter.finish()
+    }
+}
+```
+
+`Display` is used to produce human-friendlier representations. Its
+implementation will be called when formatting a [400 HTTP response message] in
+certain protocols, for example. The format of the error messages will _not_ be
+configurable by service implementers and will be
```
adamthom-amzn, line 170: Preceding frameworks have allowed validation to be disabled or customized - from message all the way up to a completely different error. If a service implementer wanted to take control of the display of validation failures without opting out of the Rust SSDK validation framework entirely, how would they do so?

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+[`TryFrom`]: https://doc.rust-lang.org/std/convert/trait.TryFrom.html
+[`Display`]: https://doc.rust-lang.org/std/fmt/trait.Display.html
+[`Debug`]: https://doc.rust-lang.org/std/fmt/trait.Debug.html
+
+### Request deserialization
+
+We will continue to deserialize the different parts of the HTTP message into
+the regular Rust standard library types. However, just before the
+deserialization function returns, we will convert the type into the wrapper
+tuple struct that we will eventually be handed over to the operation handler.
```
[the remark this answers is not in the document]
david-perez: How can we do it in one step? Ultimately the wrapper tuple struct holds e.g. a `String`, so don't we first have to deserialize the bytes slice into a `String`?

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+   When serializing this error into an HTTP response, the error message that
+   naturally arises by calling `std::fmt::Display::fmt()` will trace what
+   exactly went wrong for the caller client.
+
+`length` trait
+--------------
+
+We will enforce the length constraint by calling `len()` on Rust's `Vec`
+(`list` and `set` shapes), `HashMap` (`map` shapes), `String` (`string`
+shapes), and our [`aws_smithy_types::Blob`] (`bytes` shapes).
```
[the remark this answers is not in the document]
david-perez: 😅

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+
+> indicates that the items in a [`List`] MUST be unique.
+
+If the list shape is `sparse`, more than one `null` value violates this
+constraint.
+
+We will enforce this by copying _references_ to the `Vec`'s elements into a
+`HashSet` and checking that the sizes of both containers coincide. Smithy
+allows you to apply `uniqueItems` to _any_ list, [even if its members are not
+equipped with an equivalence relation], so we would have to manually check that
```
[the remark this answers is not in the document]
david-perez: Basically to check at code-generation time the members are not floating point types. The rest of the [simple shapes](https://awslabs.github.io/smithy/1.0/spec/core/model.html#simple-shapes) cause no issues.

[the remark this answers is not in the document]
david-perez: It's unspecified. I'd like the spec to be explicit and provide notions of equivalence: https://github.com/awslabs/smithy/issues/1072

For now, I'd just fail the codegen if the container is of floating point numbers.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+
+When the constraint trait is applied to a member shape, the tuple struct's name
+will be the PascalCased name of the member shape, `NumberOfItems`.
+
+[the spec]: https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html?highlight=enum#trait-precedence
+
+Unresolved questions
+--------------------
+
+1. Should we code-generate unsigned integer types (`u16`, `u32`, `u64`) when
```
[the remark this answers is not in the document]
david-perez: Very good point. I'm of the opinion that if you add / remove a constraint, you're backwards-incompatibly affecting your service's behavior, so a backwards-incompatible change in server user code seems appropriate.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+   the `range` trait is applied with `min` set to a value greater than or equal
+   to 0?
+    - A user has even suggested to use the `std::num::NonZeroUX` types (e.g.
+      [`NonZeroU64`]) when `range` is applied with `min` set to a value greater
+      than 0.
+2. In request deserialization, should we fail with the first violation and
+   immediately render a response, or attempt to parse the entire request and
+   provide a complete and structured report?
+3. Should we provide a mechanism for the service implementer to construct a
```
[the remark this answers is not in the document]
david-perez: What (testing) scenarios do you think would make this feature valuable?

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+   Rust type violating the modeled constraints in their business logic e.g. a
+   `T::new_unchecked()` constructor? This could be useful (1) when the user
+   _knows_ the provided inner value does not violate the constraints and
+   doesn't want to incur the performance penalty of the check; (2) when the
+   struct is in a transient invalid state. However:
+    - (2) is arguably a modelling mistake and a separate struct to represent
+      the transient state would be a better approach,
+    - the user can always use `unsafe` Rust to bypass the validation; and
```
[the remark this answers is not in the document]
david-perez: The wrapper types that we're generating are not particularly complex so the memory layout could be predictable, so I was thinking that that a sufficiently motivated user that wanted to bypass validation could use `std::mem::transmute`.

design/src/rfcs/rfc0012_constraint_traits.md
david-perez: Good point, need to add this to the doc. I was thinking of `into_inner` but `into_unconstrained` is more explicit.

design/src/rfcs/rfc0012_constraint_traits.md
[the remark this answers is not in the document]
david-perez: I didn't see it documented in the spec so I tested it; Smithy does not allow you to specify the same constraint more than once on the same shape.

```
ERROR: com.amazonaws.simple#HeaderBoundData (Model)
     @ /local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/codegen-server-test/model/simple.smithy
     |
  29 | @length(min: 1)
     |         ^
     = Conflicting `smithy.api#length` trait found on shape `com.amazonaws.simple#HeaderBoundData`. The previous trait was defined at `/local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/codegen-server-test/model/simple.smithy [28, 9]`, and a conflicting trait was defined at `/local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/codegen-server-test/model/simple.smithy [29, 9]`.
```

design/src/rfcs/rfc0012_constraint_traits.md
adamthom-amzn: This is the sort of thing that is much easier to do if you don't bake validation into the type system - in TS, handlers can be configured to accept a callback that is passed an array of `ValidationFailure` objects, and the user can turn those into an error of their choice that is then rendered by the framework. They still get the benefit of validation without giving up the presentation of modeled exceptions coming from their service.

Which reminds me of another point - we have largely done away with synthetic exceptions in Smithy. Validation is no difference - are you requiring that a service's operations be associated with a particular error structure?

design/src/rfcs/rfc0012_constraint_traits.md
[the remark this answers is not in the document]
adamthom-amzn: Yeah, I think breaking the build is fine if a constraint has been tightened, but if it's been loosened, that's really unfair. In my opinion, the types in the target language should be derived solely from the smithy types, and they shouldn't be impacted by the constraint traits outside of `@required`/`@default`.

design/src/rfcs/rfc0012_constraint_traits.md
david-perez: > in TS, handlers can be configured to accept a callback that is passed an array of ValidationFailure objects, and the user can turn those into an error of their choice that is then rendered by the framework.

Ah, I hadn't thought about that idea; I had imagined they'd want the errors to be configured ahead-of-time, not at runtime. That's neat. I don't see why we would not be able to build it into the Rust framework. We could have users return `struct`s (or functions that return `struct`s) that implement `std::error::Error` (or some trait of our own) for each of the `ValidationFailure`s in that callback, and we would use those representations when rendering the response.

> Which reminds me of another point - we have largely done away with synthetic exceptions in Smithy.

What are synthetic exceptions?

> are you requiring that a service's operations be associated with a particular error structure?

If the operation in the Smithy model is fallible (has an `errors` field), we require that the operation handlers return a [`Result<T, E>`](https://doc.rust-lang.org/std/result/enum.Result.html), where the error has to be the enum that we generate holding all the variants of their Smithy `errors` field.

adamthom-amzn: Synthetic exceptions are an idea from a legacy framework - exceptions that every service throws that don't have to be declared in the model. In that framework, `ValidationException` is synthetic, and can always be thrown. We'd like to restrict that idea to only things that are _unavoidably_ thrown, either by virtue of "this is how programming works" (`SerializationException`, `InternalFailure`) or the underlying protocol (incorrect values for `Accept`, etc).

In TS, you have two options:
- If every operation in the service has `smithy.framework#ValidationException` attached, and no customization is done in smithy-build.json, then the default validation behavior is engaged and the handlers cannot customize error rendering.
- If `smithy-build.json` has `disableDefaultValidation` set to `true`, code generation adds a **required** parameter to handlers for a callback that can turn `ValidationFailure[]` into an error (or undefined, if it wants to suppress the error)

If you don't attach `smithy.framework#ValidationException` to every operation in the service and you don't disable default validation in `smithy-build`, then code generation fails.

All this to say that you have to guarantee your validation failures result in modeled exceptions being rendered, and not unmodeled 400s.

design/src/rfcs/rfc0012_constraint_traits.md
llgerard: As a service owner you probably do care that your outputs do not match your model: your server is breaking the contract you want to set with your clients.
So being able to have clear metrics about it seems important. Consistent logging would allow that.

design/src/rfcs/rfc0012_constraint_traits.md
llgerard: +1 for having the option to provide the function doing `ValidationFailure[] -> CustomErrorAlwaysMappedTo400`

oddg: > you have to guarantee your validation failures result in modeled exceptions being rendered, and not unmodeled 400s.

I agree. Otherwise the server implementation breaks the Smithy contract. (I am actually curious about the current behavior if the input does not parse at all.)

As you mentioned, one approach to let the user choose which [error the operation will return](https://awslabs.github.io/smithy/1.0/spec/core/model.html#operation) is to generate for each operation input a type for validation error and ask the user to provide a function mapping this type to the operation error type.

Taking the [weather forecast example](https://awslabs.github.io/smithy/quickstart.html#complete-example) it would look like.

```rust
    enum GetCityInputError {
        CityId(CityIdValidationError)
    }

    enum CityIdValidationError {
        Pattern(String)
    }
```

A user would have to implement a mapping `fn(GetCityInputError) -> GetCityError`.

design/src/rfcs/rfc0012_constraint_traits.md
oddg: I also prefer the primary design as it fosters correctness.

design/src/rfcs/rfc0012_constraint_traits.md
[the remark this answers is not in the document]
oddg: Would it be possible to regard those "custom Rust features" as Smithy traits that live outside of the official Smithy implementation?

For example one could decide to create the following trait
```smithy
@trait(selector: "integer")
structure rustInteger {
      @required
      int: String
}

@rustInteger("u8")
Integer MyInteger
```
and them modify smithy-rs to support that `rustInteger` trait?

That would allow a team using Rust to apply custom traits to a model shared with other team via [the external trait application](https://awslabs.github.io/smithy/1.0/spec/core/model.html#applying-traits-externally).

design/src/rfcs/rfc0012_constraint_traits.md
crisidev: Thanks for the doc Adam, i understand the reasoning and I agree with you and Jon..

design/src/rfcs/rfc0012_constraint_traits.md
david-perez: @adamthom-amzn I didn't know about `ValidationException`, I don't see it documented in the docsite. Is the intention for this shape to be special, to be the default way for service frameworks to report validation errors?

> adds a required parameter to handlers for a callback that can turn ValidationFailure[] into an error

I'm assuming this error is modeled and associated to the operation. If the operation in the model does not define at least one error, code generation fails, right?

---

@oddg Would that mapping be static per operation (i.e. two identical invalid requests resulting in the same `GetCityError`)? Would it be a simple mapping (i.e. no access to state)?

If I were to implement this, I'm thinking that if default validation is disabled (e.g. in `smithy-build.json`) we could generate the operation registry to take in both the operation handler and the error mapping function per operation registration. Any other ideas as to where the user could provide the error mapping function?

design/src/rfcs/rfc0012_constraint_traits.md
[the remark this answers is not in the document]
david-perez: @oddg Certainly, you can do all sorts of custom traits in Smithy. I think I like more the idea of having an `@unsigned` trait for this. 

---

One thing to note is that if we were to generate unsigned types we would be expanding the range of accepted values by Smithy. For example, an [`integer` shape in Smithy](https://awslabs.github.io/smithy/1.0/spec/core/model.html#simple-shapes) is a 32-bit integer with range from -2^31 to (2^31)-1. That's _exactly_ the range of [`i32`](https://doc.rust-lang.org/std/primitive.i32.html) in Rust. If we tack on `@unsigned` or `@range(min: 0)` to the shape, it's unclear that it's too helpful to generate an [`u32`](https://doc.rust-lang.org/std/primitive.u32.html), because the upper bound of that type is (2^32)-1. The wrapper tuple `MyInteger(u32)` struct would only have to check the upper bound.

I'm now inclined to drop this feature and have `range` on integers _not_ affect whether we generate signed or unsigned integers.

design/src/rfcs/rfc0012_constraint_traits.md
adamthom-amzn: > Is the intention for this shape to be special, to be the default way for service frameworks to report validation errors?

Special? No, we're trying not to have anything be special. Think of it as a built-in convenience that service frameworks are free to use as a default :)

> I'm assuming this error is modeled and associated to the operation. If the operation in the model does not define at least one error, code generation fails, right?

This would be a fine behavior to have. Another is to just use your normal error serialization logic, which should turn any exception that is not associated with the operation into an `InternalFailure`.

> I am actually curious about the current behavior if the input does not parse at all.

@oddg by spec, here meaning protocol tests, unparseable input should result in a 400 with the error code of `SerializationException`. It's a synthetic, which as I outlined above, means non-modeled (and does not need to have a corresponding specialized type generated client-side).

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+For constrained shapes we would do the same to parse the inner deserialized
+value into the wrapper tuple struct, except for these differences:
+
+1. For enums, the client generates an `Unknown` variant that "contains new
+   variants that have been added since this code was generated". The server
+   does not need such a variant ([#1187]).
+2. Conversions into the tuple struct are fallible (`try_from()` instead of
+   `from()`). These errors will result in a `MyStructValidationError`, wrapped
+   in the runtime type error
+   `aws_smithy_http_server::rejection::ConstraintViolation` rejection. The
```
jdisanti, line 283: Will there be a way to hook these different failures for metrics so that service owners can discern between failure to parse and failure to validate in their metrics?

david-perez: Yes; the latter will be `Deserialize` rejections containing `ConstraintViolation` rejections, the former will contain other inner rejections, so it will be possible to distinguish.

design/src/rfcs/rfc0012_constraint_traits.md
oddg: > Would that mapping be static per operation (i.e. two identical invalid requests resulting in the same GetCityError)? Would it be a simple mapping (i.e. no access to state)?

Yes. In my opinion, it makes sense that two identical invalid requests result in the same errors. I agree that it should not depend on any state. (Any error that is is state dependent should be generated in the core implementation of the handler for that operation.)

> We could generate the operation registry to take in both the operation handler and the error mapping function per operation registration.

That makes a lot of sense to me.

> by spec, here meaning protocol tests, unparseable input should result in a 400 with the error code of SerializationException. It's a synthetic, which as I outlined above, means non-modeled (and does not need to have a corresponding specialized type generated client-side).

Thanks.

design/src/rfcs/rfc0012_constraint_traits.md
```diff
@@ -0,0 +1,424 @@
...
+
+The central idea to the implementation of constraint traits is: [parse, don't
+validate]. Instead of code-generating a Rust `String` to represent `NiceString`
+values and perform the _validation_ at request deserialization, we can
+[leverage Rust's type system to guarantee domain invariants]. We can generate a
+wrapper [tuple struct] that _parses_ the string's value and is "tight" in the
+set of values it can accept:
+
+```rust
+pub struct NiceString(String);
```
rcoh, line 108: it's worth calling out that adding a constraint trait is now a breaking change for servers since it adds this type wrapper. I haven't thought about the trade off deeply, but it seems worth mentioning / considering since it will make it a bit more of a hassle to add constraints retroactively

design/src/rfcs/rfc0012_constraint_traits.md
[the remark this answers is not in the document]
rcoh: if you wanted to, you could impl an `.unsigned()` method when the model allowed to enable converting from i32 to u32 fearlessly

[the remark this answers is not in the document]
jdisanti: Worth mentioning that the clients should also support this if the server does, or there will be issues when a server responds with something outside of the signed int range.

david-perez:
I've updated the RFC to reflect what we ended up implementing.

LukeMathWalker (approved):

PR #1209, opened 2022-02-19
Hide external buffer types used by `primitive::Encoder` in aws-smithy-types

jdisanti:
## Motivation and Context
The buffer types are an implementation detail for the primitive encoding, and should not be tied to the public API. This PR wraps them in some lightweight structs to hide them.

## Testing
- Added `api-linter` to CI for the `aws-smithy-types` crate now that it reveals no external types in its API

## Checklist


rust-runtime/aws-smithy-types/src/primitive.rs
```diff
@@ -107,25 +119,25 @@ pub enum Encoder {
     Bool(bool),
     /// 8-bit signed integer
     #[non_exhaustive]
-    I8(i8, itoa::Buffer),
```
rcoh, line 110: do we actually even need these public variants on encoder? Can we do `struct Encoder(Inner); `enum Inner { ... }`?

rcoh: since we have the from impls already

rcoh (approved):
lgtm, lmk what you think about just making it a private tuple struct though

rust-runtime/aws-smithy-types/src/primitive.rs
jdisanti: I like this idea! Changed over.

rcoh (approved):

PR #1213, opened 2022-02-22
Generate `httpMalformedRequestTests`

david-perez:
Smithy defines one trait specifically for the behavior of server
protocol implementations.

From https://awslabs.github.io/smithy/1.0/spec/http-protocol-compliance-tests.html#httpmalformedrequesttests:

> The `httpMalformedRequestTests` trait is used to define how a malformed
> HTTP request is rejected given a specific protocol and HTTP message.
> Protocol implementations MUST assert that requests are rejected during
> request processing.

This commit implements code generation for these tests. The ones for the
restJson1 protocol have been added to the list of failing tests. These
should begin passing once we start implementing constraint traits
https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -298,22 +296,77 @@ class ServerProtocolTestGenerator(
...
+            .body(${
+                if (body != null) {
+                    // The `replace` is necessary to fix the malformed request test `RestJsonInvalidJsonBody`.
+                    // https://github.com/awslabs/smithy/blob/887ae4f6d118e55937105583a07deb90d8fabe1c/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-request-body.smithy#L47
+                    //
+                    // Smithy is written in Java, which parses `\u000c` within a `String` as a single char given by the
+                    // corresponding Unicode code point. That is the "form feed" 0x0c character. When printing it,
+                    // it gets written as "\f", which is an invalid Rust escape sequence: https://static.rust-lang.org/doc/master/reference.html#literals
+                    // So we need to write the corresponding Rust Unicode escape sequence to make the program compile.
+                    "#{SmithyHttpServer}::body::Body::from(#{Bytes}::from_static(b${body.replace("\u000c", "\\u{000c}").dq()}))"
```
david-perez, line 356: Is there a better way to do this?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -154,6 +154,7 @@ private class ServerHttpProtocolImplGenerator(
         // Implement Axum `FromRequest` trait for input types.
         rustTemplate(
             """
+            ##[derive(Debug)]
```
david-perez, line 157: This is now needed because of the call to `expect_err()` in the protocol tests.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
crisidev: The detailed explanation helps a lot. If it makes you feel better, I would have implemented it the same way :) Probably there is some crate out in the wild that can help with this, but since it is a single occurrence in our code (so far), I believe we are good for now.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -624,7 +625,538 @@ class ServerProtocolTestGenerator(
             FailingTest(RestJson, "RestJsonHttpWithEmptyBlobPayload", TestType.Request),
             FailingTest(RestJson, "RestJsonHttpWithEmptyStructurePayload", TestType.Request),
 
-            // Malformed requests.
+            FailingTest(RestJson, "RestJsonComplexErrorWithNoMessageResponse", TestType.MalformedRequest),
```
crisidev:
```
20 little bugs in the code
20 little bugs in the code
Take one down, patch it around
613 little bugs in the code
```

crisidev (approved):
LGTM

PR #1216, opened 2022-02-23
Fix presigning bug with `content-length` and `content-type` in S3

jdisanti:
## Motivation and Context
This PR fixes https://github.com/awslabs/aws-sdk-rust/issues/466.

## Description
Rather than remove `content-length` and `content-type` when signing and creating a presigned request, this PR refactors `MakeOperationGenerator` to not produce default values for those headers when making an operation for presigning.

## Testing
- Updated S3 presign tests
- Ran Polly presign tests
- Manually tested S3 presigning:
   - Set `content-length` to a value in the input, and then verified there is a signing error when uploading a file that is not that length, and that it succeeds with the matching file size.
   - Set `content-type` to a value in the input, and then verified there is a signing error with a different `content-type`, and that it succeeds when it matches.

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-inlineable/src/presigning.rs
+++ b/aws/rust-runtime/aws-inlineable/src/presigning.rs
@@ -219,7 +219,7 @@ pub mod request {
 pub(crate) mod service {
     use crate::presigning::request::PresignedRequest;
     use aws_smithy_http::operation;
-    use http::header::{CONTENT_LENGTH, CONTENT_TYPE, USER_AGENT};
+    use http::header::USER_AGENT;
     use std::future::{ready, Ready};
     use std::marker::PhantomData;
     use std::task::{Context, Poll};
@@ -261,12 +261,6 @@ pub(crate) mod service {
 
         fn call(&mut self, req: operation::Request) -> Self::Future {
             let (mut req, _) = req.into_parts();
-
-            // Remove headers from input serialization that shouldn't be part of the presigned
-            // request since the request body is unsigned and left up to the person making the final
-            // HTTP request.
-            req.headers_mut().remove(CONTENT_LENGTH);
-            req.headers_mut().remove(CONTENT_TYPE);
 
             // Remove user agent headers since the request will not be executed by the AWS Rust SDK.
             req.headers_mut().remove(USER_AGENT);
--- a/aws/sdk/integration-tests/s3/tests/presigning.rs
+++ b/aws/sdk/integration-tests/s3/tests/presigning.rs
@@ -4,39 +4,47 @@
  */
 
 use aws_sdk_s3 as s3;
+use aws_sdk_s3::presigning::request::PresignedRequest;
+use http::header::{CONTENT_LENGTH, CONTENT_TYPE};
+use http::{HeaderMap, HeaderValue};
 use s3::presigning::config::PresigningConfig;
 use std::error::Error;
 use std::time::{Duration, SystemTime};
 
+macro_rules! presign_input {
+    ($input:expr) => {{
+        let creds = s3::Credentials::new(
+            "ANOTREAL",
+            "notrealrnrELgWzOk3IfjzDKtFBhDby",
+            Some("notarealsessiontoken".to_string()),
+            None,
+            "test",
+        );
+        let config = s3::Config::builder()
+            .credentials_provider(creds)
+            .region(s3::Region::new("us-east-1"))
+            .build();
+
+        let req: PresignedRequest = $input
+            .presigned(
+                &config,
+                PresigningConfig::builder()
+                    .start_time(SystemTime::UNIX_EPOCH + Duration::from_secs(1234567891))
+                    .expires_in(Duration::from_secs(30))
+                    .build()
+                    .unwrap(),
+            )
+            .await?;
+        req
+    }};
+}
+
 #[tokio::test]
 async fn test_presigning() -> Result<(), Box<dyn Error>> {
-    let creds = s3::Credentials::new(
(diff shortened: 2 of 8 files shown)


CHANGELOG.next.toml
```diff
@@ -34,3 +34,15 @@ message = "`aws_smithy_types::primitive::Encoder` is now a struct rather than an
...
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "`aws-sigv4` no longer skips the `content-length` and `content-type` headers when signing with `SignatureLocation::QueryParams`"
+references = ["smithy-rs#1216"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
+author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Fixed a bug in S3 that prevented the `content-length` and `content-type` inputs from being included in a presigned request signature"
```
rcoh, line 45: a sentence about the impact may be helpful, "With this fix, customers can generate presigned URLs that enforce content-length and content type" or something

aws/rust-runtime/aws-inlineable/src/presigning.rs
```diff
@@ -262,12 +262,6 @@ pub(crate) mod service {
...
 
-            // Remove headers from input serialization that shouldn't be part of the presigned
-            // request since the request body is unsigned and left up to the person making the final
-            // HTTP request.
-            req.headers_mut().remove(CONTENT_LENGTH);
-            req.headers_mut().remove(CONTENT_TYPE);
-
             // Remove user agent headers since the request will not be executed by the AWS Rust SDK.
             req.headers_mut().remove(USER_AGENT);
             req.headers_mut().remove("X-Amz-User-Agent");
```
rcoh, line 267: maybe this is hard...this is probably hard...ideally we just wouldn't hit the UA middleware

aws/sdk/integration-tests/s3/tests/presigning.rs
```diff
@@ -4,39 +4,47 @@
...
 
 use aws_sdk_s3 as s3;
+use aws_sdk_s3::presigning::request::PresignedRequest;
+use http::header::{CONTENT_LENGTH, CONTENT_TYPE};
+use http::{HeaderMap, HeaderValue};
 use s3::presigning::config::PresigningConfig;
 use std::error::Error;
 use std::time::{Duration, SystemTime};
 
+macro_rules! presign_input {
```
rcoh, line 14: comment on test macros is often helpful

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/MakeOperationGenerator.kt
```diff
@@ -38,7 +39,9 @@ open class MakeOperationGenerator(
     private val protocol: Protocol,
     private val bodyGenerator: ProtocolPayloadGenerator,
     private val functionName: String = "make_operation",
-    private val public: Boolean = true
+    private val public: Boolean = true,
+    /** Whether or not to include default values for content-length and content-type */
+    private val includeDefaultPayloadHeaders: Boolean = true,
```
rcoh, line 44: consider removing the default values here—since this function is only invoked in 2 (?) places, explicit is probably clearer on both sides

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/MakeOperationGenerator.kt
```diff
@@ -138,43 +153,39 @@ open class MakeOperationGenerator(
...
-            rust("Ok(builder)")
+
+        writer.rust("let mut builder = update_http_builder(&self, #T::new())?;", RuntimeType.HttpRequestBuilder)
+        if (includeDefaultPayloadHeaders && contentType != null) {
+            writer.rustTemplate(
+                "builder = #{header_util}::set_request_header_if_absent(builder, #{http}::header::CONTENT_TYPE, ${contentType.dq()});",
+                *codegenScope
+            )
+        }
+        for (header in protocol.additionalRequestHeaders(operationShape)) {
```
rcoh, line 177: are _these_ ones we want to always set? I guess so right? this is stuff like `operation` for JSON RPC?

rcoh (approved):
LGTM! I like the design we landed on for this, thanks for the iteration!

jdisanti:
Going to add S3 to the codegen diff so that we can see these changes in the PR.

jdisanti:
It looks like it's not so simple to just include S3. It should already be in the diff output, but it seems like diff2html is dropping files when the diff is exceedingly large. Going to punt on this for now.

aws/rust-runtime/aws-inlineable/src/presigning.rs
jdisanti: We could potentially add something to the property bag that tells the UA middleware to no-op, but I don't think it would be easy to customize the middleware in the presigning function as it currently is.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/MakeOperationGenerator.kt
jdisanti: Yeah, it looks like it's only used by AWS JSON to add the `x-amz-target` header, which is needed to identify the operation to the service. We wouldn't want to remove these for this current use-case.

crisidev (approved):
LGTM from a server perspective.

PR #1217, opened 2022-02-23
Make `rustls` and `native_tls` client builder helpers dyn

jdisanti:
## Motivation and Context
This PR fixes https://github.com/awslabs/aws-sdk-rust/issues/467 which was caused by making the SDK clients use `DynConnector` in https://github.com/awslabs/smithy-rs/pull/1132.

There's a bit of conflict here between smithy-rs and SDK use-cases. Developers using smithy-rs want the flexibility to use static dispatch on the client's connector, but the SDK does not need this. I'm opting to make the SDK use-case easy and convenient, and it should even work for most smithy-rs use-cases. Those who need to use static dispatch will need to manually construct an adapter (for example, `builder.connector(Adapter::builder().build(aws_smithy_client::conns::https()))`) if the performance difference is a concern.

I considered adding `rustls_static` and `native_tls_static` equivalents for static dispatch, but I thought those might cause confusion.

## Testing
- Added a test to check the return types

## Checklist

### DIFF
--- a/rust-runtime/aws-smithy-client/src/hyper_ext.rs
+++ b/rust-runtime/aws-smithy-client/src/hyper_ext.rs
@@ -304,19 +304,22 @@ where
 
 #[cfg(feature = "rustls")]
 impl<M, R> ClientBuilder<(), M, R> {
-    /// Connect to the service over HTTPS using Rustls.
-    pub fn rustls(self) -> ClientBuilder<Adapter<crate::conns::Https>, M, R> {
-        self.connector(Adapter::builder().build(crate::conns::https()))
+    /// Connect to the service over HTTPS using Rustls using dynamic dispatch.
+    pub fn rustls(self) -> ClientBuilder<DynConnector, M, R> {
+        self.connector(DynConnector::new(
+            Adapter::builder().build(crate::conns::https()),
+        ))
     }
 }
 
 #[cfg(feature = "native-tls")]
 impl<M, R> ClientBuilder<(), M, R> {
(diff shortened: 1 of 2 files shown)


rust-runtime/aws-smithy-client/src/hyper_ext.rs
```diff
@@ -304,19 +304,22 @@ where
...
-    pub fn native_tls(
-        self,
-    ) -> ClientBuilder<Adapter<hyper_tls::HttpsConnector<hyper::client::HttpConnector>>, M, R> {
-        self.connector(Adapter::builder().build(crate::conns::native_tls()))
+    /// Connect to the service over HTTPS using the native TLS library on your
+    /// platform using dynamic dispatch.
+    pub fn native_tls(self) -> ClientBuilder<DynConnector, M, R> {
+        self.connector(DynConnector::new(
+            Adapter::builder().build(crate::conns::native_tls()),
+        ))
```
rcoh, line 322: I wonder if we should examples in these methods about how _not_ to use them when required (basically including examples of custom HTTPS, or at least linking to them)

rcoh (approved):
LGTM!

rust-runtime/aws-smithy-client/src/hyper_ext.rs
jdisanti: Filed an issue for improving these docs overall: https://github.com/awslabs/aws-sdk-rust/issues/469

PR #1220, opened 2022-02-24
Make it possible to configure the default credentials cache

jdisanti:
## Motivation and Context
This is for https://github.com/awslabs/aws-sdk-rust/issues/462. It is just making it possible to change settings on the lazy caching builder used by the default credentials provider builder.

This should enable customizing the credential load timeout with:
```rust
let default_provider = DefaultCredentialsChain::builder()
    .load_timeout(Duration::from_secs(20))
    .build();
```

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/default_provider/credentials.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider/credentials.rs
@@ -114,7 +114,7 @@ impl Builder {
         self
     }
 
-    /// Timeout for the given [`ProvideCredentials`] implementation.
+    /// Timeout for the entire credential loading chain.
     ///
     /// Defaults to 5 seconds.
     pub fn load_timeout(mut self, timeout: Duration) -> Self {
@@ -122,7 +122,7 @@ impl Builder {
         self
     }
 
-    /// Timeout for the given [`ProvideCredentials`] implementation.
+    /// Timeout for the entire credential loading chain.
     ///
     /// Defaults to 5 seconds.
     pub fn set_load_timeout(&mut self, timeout: Option<Duration>) -> &mut Self {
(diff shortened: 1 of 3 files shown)


aws/rust-runtime/aws-config/src/default_provider/credentials.rs
```diff
@@ -114,6 +114,71 @@ impl Builder {
         self
     }
 
+    /// Timeout for the given [`ProvideCredentials`] implementation.
```
rcoh, line 117: should this say "default credentials chain"? We should clarify that this is the entire chain and not an individual provider (but maybe we should actually add that too...)

aws/rust-runtime/aws-config/src/default_provider/credentials.rs
```diff
@@ -114,6 +114,71 @@ impl Builder {
...
+    pub fn set_load_timeout(&mut self, timeout: Option<Duration>) -> &mut Self {
+        self.credential_cache.set_load_timeout(timeout);
+        self
+    }
+
+    /// Amount of time before the actual credential expiration time
+    /// where credentials are considered expired.
+    ///
+    /// For example, if credentials are expiring in 15 minutes, and the buffer time is 10 seconds,
+    /// then any requests made after 14 minutes and 50 seconds will load new credentials.
```
rcoh, line 137: should update to have a first line that can stand alone

rcoh (approved):
LGTM, just some proposed docs improvements.

aws/rust-runtime/aws-config/src/default_provider/credentials.rs
jdisanti: Good call. Fixed.

PR #1221, opened 2022-02-24
Pin nightly to `2022-02-22` to fix CI

jdisanti:
## Motivation and Context
The latest nightly introduced a new rustdoc format version that breaks the api-linter tool. This PR pins nightly to give us more time to update to the latest format version.

### DIFF
--- a/.github/workflows/ci-sdk.yaml
+++ b/.github/workflows/ci-sdk.yaml
@@ -68,7 +68,7 @@ jobs:
         - name: Clippy
           run: cargo clippy --all-features
         - name: Unused Dependencies
-          run: cargo +nightly udeps
+          run: cargo +nightly-2022-02-22 udeps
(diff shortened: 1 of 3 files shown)


crisidev (approved):

Velfi (approved):

Velfi:
Consider adding a code comment explaining why the pin is necessary / how to tell if the pin can be removed

aDotInTheVoid:
caused by https://github.com/rust-lang/rust/pull/94137. You can upgrade to `rustdoc-types` 0.7 to move to the lastest nightly, as you don't look at headers/abi. But it's still a good idea to pin nightly.

jdisanti:
Hey @aDotInTheVoid, definitely planning to upgrade to the latest as soon as I have a chance! 😄 Thanks

Issue #1200, opened 2022-02-17
Document usage of `FallibleOperationsDecorator`


crisidev:
Document the usage of the decorator introducted by #1190.

guymguym:
Hey @crisidev - Why did you choose to make this decorator only add the `InternalServerError` to operations that have no other errors defined? When an API server is subject to internal errors (for example network issues connecting to underlying services) I would expect it to add it unconditionally to all operations regardless of this check for `shape.errors.isEmpty()` simply because there is no native mapping of internal errors to the functional errors of the API. WDYT?
 
https://github.com/awslabs/smithy-rs/blob/0a13f76568ed1869b5e6f280a62a7354fe7cd421/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/FallibleOperationsDecorator.kt#L47

crisidev:
I probably did not understand your problem completely. I thought you needed to make the operation fallible even if it was not, not also change operations that can already return errors.

That said, we created this decorator specifically for your use case. If it doesn't fit we can just change it.

guymguym:
Hey, got it, I probably did not explain it clearly.
I am trying to convince that this is needed for other servers too.
I guess that if you write your own API you could add errors as desired.
However in case the API already exists and you would like to create a server for it, it can be quite difficult to map internal errors to one of the API errors if it was not designed to represent these errors, and in this cases an InternalServerError is desired to be returned, and clients will anyhow have to be able to map unknown errors (which can also just be a new API error that was added to the server after the client was generated).

If you agree with that I can update this decorator to `InternalServerErrorDecorator`.
If not, I can either add a new decorator, or it can live outside of this project if you feel it doesn't belong, or if you believe this is just the wrong approach I would be happy to learn how it is recommended to handle server errors with smithy.

crisidev:
I am happy to change the decorator (or get a pull request) and fulfill the behaviour you are searching for. 

On a more general view, I still think that new API should not make use of this as it is an anti-pattern, but for old services I can definitely see the use for it.

I'd rather keep the decorators in smithy-rs repo, so they can be shared with other projects, but I am still confident they should be opt-in only and never a default.

PR #1224, opened 2022-02-25
[Server] AdditionalErrorsDecorator - refactored from FallibleOperationsDecorator

guymguym:

## Motivation and Context
- Followup of the discussion in #1200

## Description
- Rename the decorator and update to always add the internal server error to all operations.

## Testing
- Updated also the test.

## Checklist


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/InternalServerErrorDecorator.kt
```diff
@@ -14,20 +14,37 @@ import software.amazon.smithy.model.traits.ErrorTrait
 import software.amazon.smithy.model.traits.RequiredTrait
```
crisidev, line 14: Why not keeping both the decorators in the same file? They are similar for sure, but they can both be useful. WDYT?

crisidev (commented):
Code looks good.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/InternalServerErrorDecorator.kt
guymguym: Sure. I'm just not sure how to name the file... 😵‍💫

crisidev: Ahahaha ok.. I would call the file AdditionalErrorsDecorator.kt..

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/AdditionalErrorsDecorator.kt
```diff
@@ -0,0 +1,87 @@
...
+import software.amazon.smithy.model.transform.ModelTransformer
+import software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator
+
+/**
+ * Add at least one error to all operations in the model.
+ *
+ * When this decorator is applied, even operations that do not have a Smithy error attatched,
+ * will return `Result<OperationOutput, OperationError>`.
+ *
+ * To enable this decorator write its class name to a resource file like this:
```
crisidev, line 24:
```suggestion
/**
 * Add at least one error to all operations in the model.
 *
 * When this decorator is applied, operations that do not have a Smithy error attatched,
 * will return `Result<OperationOutput, InternalServerError>`.
 *
 * To enable this decorator write its class name to a resource file like this:
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/AdditionalErrorsDecorator.kt
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Add an internal server error to all operations in the model.
+ *
+ * When a server for an API is subject to internal errors, for example underlying network problems,
+ * and there is no native mapping of these actual errors to the API errors, servers can generate
+ * the code with this decorator to add an internal error shape on-the-fly to all the operations.
+ *
+ * When this decorator is applied, even operations that do not have a Smithy error attatched,
+ * will return `Result<OperationOutput, OperationError>`.
```
crisidev, line 48:
```suggestion
 * When this decorator is applied, even operations that do not have a Smithy error attatched,
 * will return `Result<OperationOutput, InternalServerError>`.
```

crisidev (approved):
LGTM, just 2 small changes you can apply if you want. I am not super fond of the naming of the decorators, but I don't have anything better in mind :)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/AdditionalErrorsDecorator.kt
guymguym: I changed it to OperationError because InternalServerError is not really the error type, every operation gets its own enum and then the result is actually returning `Result<OperationOutput, OperationError>`

For example in my error.rs:

```
/// Error type for the `PutObject` operation.
/// Each variant represents an error that can occur for the `PutObject` operation.
#[derive(std::fmt::Debug)]
pub enum PutObjectError {
    InternalServerError(crate::error::InternalServerError),
}
```

and in operation_handler.rs:
```
#[async_trait::async_trait]
impl<B, Fun, Fut>
    crate::server_operation_handler_trait::Handler<B, (), crate::input::PutObjectInput> for Fun
where
    Fun: FnOnce(crate::input::PutObjectInput) -> Fut + Clone + Send + 'static,
    Fut: std::future::Future<
            Output = Result<crate::output::PutObjectOutput, crate::error::PutObjectError>,
        > + Send,
```

guymguym:
@crisidev Yeah I don't mind changing the names to anything at all.
Maybe long imperative names like this? 
```
AddInternalServerErrorToInfallibleOpsDecorator
AddInternalServerErrorToAllOpsDecorator
```

crisidev:
> @crisidev Yeah I don't mind changing the names to anything at all. Maybe long imperative names like this?
> 
> ```
> AddInternalServerErrorToInfallibleOpsDecorator
> AddInternalServerErrorToAllOpsDecorator
> ```

These look great IMHO!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/AdditionalErrorsDecorator.kt
crisidev: Ok, works for me :)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/AdditionalErrorsDecorator.kt
```diff
@@ -0,0 +1,87 @@
...
+ * To enable this decorator write its class name to a resource file like this:
+ * ```
+ * C="software.amazon.smithy.rust.codegen.server.smithy.customizations.AddInternalServerErrorToInfallibleOpsDecorator"
+ * F="software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator"
+ * D="codegen-server/src/main/resources/META-INF/services"
+ * mkdir -p "$D" && echo "$C" > "$D/$F"
+ * ```
+ */
+class AddInternalServerErrorToInfallibleOpsDecorator : RustCodegenDecorator {
+    override val name: String = "FallibleOperations"
```
crisidev, line 33: Sorry for being a nitpick, but the name here should be the same as the class IMHO. Same for the other decorator..

guymguym: oh my bad, didn't notice, fixing

crisidev (approved):

PR #1227, opened 2022-02-25
[Draft] Github action for server integration tests

crisidev:
## Motivation and Context
This new action uses the Pokemon service model, generates both server and client SDKs and runs an end-to-end integration test on localhost.

### DIFF
--- a/.github/workflows/server-benchmark.yml
+++ b/.github/workflows/server-benchmark.yml
@@ -29,6 +29,11 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: ${{ env.java_version }}
+    # Install Rust
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ env.rust_version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/server-benchmark.yml
```diff
@@ -0,0 +1,35 @@
...
+# This job will run the server SDK integration tests amd benchmarks using the Pokémon service model.
+on:
+  pull_request:
+    types:
+    - opened
+    - reopened
+    - synchronize
+env:
+  java_version: 11
+  rust_version: 1.56.1
```
jdisanti, line 11: It looks like you're not installing the Rust toolchain in the job, so it's going to use whatever is installed by default on the image rather than this version.

crisidev: Ah, thanks for pointing this out. I'll fix.

jdisanti (approved):

PR #1233, opened 2022-03-03
Revamp error system

david-perez:
This commit rewrites the server's error system so that it can render
accurate protocol-specific HTTP responses adhering to the malformed
requests protocol tests' expectations. It also makes 400+ of the tests
relating to request deserialization pass.

Previously, we were rendering the `SmithyRejection` type in our HTTP responses
directly as we desired. However, AWS protocols' errors are not as granular:
many different request deserialization failure causes are conlated and
signalled to clients under a single `SerializationException` in the response.
This commit introduces a new concept, the `RuntimeEror` type, which groups
rejections and renders the responses that these protocols expect. Rejection
types now become an internal detail of the framework that only serve for us
maintainers to keep track and inventory all possible error causes in great
detail.

Extension types have also been renamed to better reflect their purpose.

For more information about this patch, read the added documentation to the
`rejection`, `extension`, and `runtime_error` modules.


david-perez:
There's a few places in the diff that have `TODO`s without an associated GitHub issue; these are the places I need feedback on.

david-perez:
This branch currently has the diff from https://github.com/awslabs/smithy-rs/pull/1218 as well. Once that gets merged, I'll rebase and this PR's diff will become a bit lighter.

rust-runtime/aws-smithy-http-server/src/exception.rs
```diff
@@ -0,0 +1,104 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+//! Exception types.
+//!
+//! This module contains "exception types". There is only one such type, [`SmithyFrameworkException`].
+//!
+//! As opposed to rejection types (see [`crate::rejection`]), which are an internal detail about
+//! the framework, exceptions are surfaced to clients in HTTP responses: indeed, they implement
+//! [`axum_core::response::IntoResponse`]. Rejections can be "grouped" and converted into an
```
crisidev, line 12:
```suggestion
//! As opposed to rejection types (see [`crate::rejection`]), which are an internal detail about
//! the framework, exceptions are surfaced to clients in HTTP responses as they implement
//! [`axum_core::response::IntoResponse`]. Rejections can be "grouped" and converted into an
```

rust-runtime/aws-smithy-http-server/src/exception.rs
```diff
@@ -0,0 +1,104 @@
...
+//! exception type: for example, all request rejections due to serialization issues can be
+//! conflated under the [`SmithyFrameworkExceptionType::Serialization`] exception type enum
+//! variant.
```
crisidev, line 15:
```suggestion
//! conflated under the [`SmithyFrameworkExceptionType::Serialization`] enum
//! variant.
```

rust-runtime/aws-smithy-http-server/src/exception.rs
```diff
@@ -0,0 +1,104 @@
...
+//! exceptions in the RestJson1 protocol set the `X-Amzn-Errortype` header.
+//!
+//! Exception types' behavior doesn't have anything to do with how _programming language exceptions_
+//! behave: there are no exceptions in Rust, `[SmithyFrameworkException`] is just a regular Rust
+//! enum. Their name comes from their representation in all AWS protocols: for example,
+//! [`SmithyFrameworkExceptionType::Serialization`] renders an `X-Amzn-Errortype` header with the
+//! value `SerializationException` when using the RestJson1 protocol.
+//!
+//! The way generated code works is it always works with [`crate::rejection`] types when
+//! deserializing requests and serializing response. Just before a response needs to be sent, the
```
crisidev, line 27:
```suggestion
//! Generated code always works with [`crate::rejection`] types when
//! deserializing requests and serializing response. Just before a response needs to be sent, the
```

rust-runtime/aws-smithy-http-server/src/exception.rs
```diff
@@ -0,0 +1,104 @@
...
+//! response.
+
+use crate::protocols::Protocol;
+
+#[derive(Debug)]
+pub enum SmithyFrameworkExceptionType {
+    // UnknownOperation,
+    Serialization(crate::Error),
+    // UnsupportedMediaType,
+    // NotAcceptable,
```
crisidev, line 39: What about these commented values?

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -4,47 +4,53 @@
...
 //! Protocol helpers.
-use crate::rejection::{ContentTypeRejection, MimeParsingFailed, MissingJsonContentType, MissingXmlContentType};
+use crate::rejection::RequestRejection;
 use axum_core::extract::RequestParts;
 
+#[derive(Debug)]
+pub enum Protocol {
+    RestJson1,
+    RestXml,
+}
```
crisidev, line 14: This should probably be non_exhaustive... I am also wondering if we need an "Unknown" variant that will signal us we got a protocol we do not support yet.. Thoughts?

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -69,8 +69,11 @@ impl ResponseExtensions {
 pub struct ExtensionModeledError(&'static str);
 impl_extension_new_and_deref!(ExtensionModeledError);
 
+// TODO This should probably be renamed to `ExtensionException`, since we now surface exceptions.
+// Although I prefer to put `Extension` at the end for these, like we do for the types in
+// `rejection.rs`, so, `ExceptionExtension`.
```
crisidev, line 74: I agree about the naming, `ExceptionExtension` works for me here.

rust-runtime/aws-smithy-http-server/src/rejection.rs
```diff
@@ -3,80 +3,45 @@
...
-    #[status = BAD_REQUEST]
-    #[body = "Failed to parse request MIME type"]
-    /// Rejection type used if the MIME type parsing failed.
-    pub struct MimeParsingFailed;
-}
+//! Rejection types.
+//!
+//! This module contains types that are commonly used as the `E` error type in functions that
+//! handle requests and responses that return `Result<T, E>` throughout the framework. These
+//! include functions to deserialize incoming requests and serialize outgoing responses.
```
crisidev, line 10:
```suggestion
//! include functions used to deserialize incoming requests and serialize outgoing responses.
```

crisidev (commented):
Nice work David, some comments and questions, nothing major.

crisidev:
Clippy is complaining: https://github.com/awslabs/smithy-rs/runs/5407662775?check_suite_focus=true#step:7:1403

rust-runtime/aws-smithy-http-server/src/exception.rs
```diff
@@ -0,0 +1,104 @@
...
+            Protocol::RestXml => {
+                builder = builder.header("Content-Type", "application/xml");
+            }
+        }
+
+        // TODO What extension type should we use here?
+        // TODO `ResponseExtensions` should probably be renamed, as it's something
+        // operation-specific. `SmithyFrameworkException` might not have reached an operation
+        // (think `UnknownOperationException`).
+        builder = builder.extension(crate::ResponseExtensions::new("TODO", "TODO"));
```
david-perez, line 88: What do we do here? Do we make it so that exceptions take in an _optional_ (namespace, operation name) pair so that we can fill this in? It must be optional because some exceptions could be raised before code execution gets to a particular operation deserialization routine (e.g. `UnknownOperationException`).

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -464,7 +465,8 @@ class ServerProtocolTestGenerator(
     private fun checkResponse(rustWriter: RustWriter, testCase: HttpMalformedResponseDefinition) {
         checkStatusCode(rustWriter, testCase.code)
         checkHeaders(rustWriter, "&http_response.headers()", testCase.headers)
-        checkHttpResponseExtensions(rustWriter)
+        // TODO Enable.
+        // checkHttpResponseExtensions(rustWriter)
```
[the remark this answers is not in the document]
david-perez: It should be enabled in this PR, but it currently fails for malformed request tests because I don't know what to do here https://github.com/awslabs/smithy-rs/pull/1233/files#r818735202

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt
```diff
@@ -208,7 +214,10 @@ private class ServerHttpProtocolImplGenerator(
                                 response
                             },
                             Err(e) => {
-                                e.into_response()
+                                #{SmithyFrameworkException} {
+                                    protocol: #{SmithyHttpServer}::protocols::Protocol::RestJson1,
+                                    exception_type: e.into()
+                                }.into_response()
```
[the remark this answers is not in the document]
david-perez: Yes, good catch.

rust-runtime/aws-smithy-http-server/src/exception.rs
david-perez: They will get uncommented when I get to work on the protocol tests that exercise them; didn't want to make this PR bigger. I've copied them from the TypeScript sSDK https://github.com/awslabs/smithy-typescript/blob/2f013f6aad91851e0d4309b3ffbf90f60d3f0dd7/smithy-typescript-ssdk-libs/server-common/src/errors.ts#L18-L18 (except for the internal one, which I don't think we will need in Rust).

rust-runtime/aws-smithy-http-server/src/exception.rs
```diff
@@ -0,0 +1,102 @@
...
+impl SmithyFrameworkExceptionType {
+    pub fn name(&self) -> &'static str {
+        match self {
+            SmithyFrameworkExceptionType::Serialization(_) => "SerializationException",
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct SmithyFrameworkException {
```
david-perez, line 52: I'm tempted to rename this, it might not be a good name. I don't like putting in the word "framework", and the concept of exceptions is not inherent to Smithy, it's just a side-product of how AWS protocols serialize errors in responses.

What if we just name it `Exception`?

rust-runtime/aws-smithy-http-server/src/protocols.rs
crisidev: Ok, makes sense.

rust-runtime/aws-smithy-http-server/src/exception.rs
crisidev: IMO since they are not exercised and commented, they should just be removed until we need them..

rust-runtime/aws-smithy-http-server/src/exception.rs
crisidev: What if we make it very simple and use "unknown", "unknown" in case we did not reach the point where we know operation and namespace? I am saying this because from a metrics perspective, you want to emit that an error happened, but you don't really know for which operation, hence unknown.

rust-runtime/aws-smithy-http-server/src/exception.rs
crisidev: I am ok naming this `Exception`.. it can still be very confusing for many people as it will be associated with the concept of exception in java / python / etc.. I believe we just need to document this very clearly and we can rename it to just Exception.

rust-runtime/aws-smithy-http-server/src/exception.rs
david-perez: Let's use this summary thread to talk about the extensions mess https://github.com/awslabs/smithy-rs/pull/1233/files#r818792743

rust-runtime/aws-smithy-http-server/src/extension.rs
crisidev: I like 1) and 2) from your proposal.

Regarding 3), shouldn't we keep a way to signal the user than something bad happened at the framework level? In my mind a `ModeledError` is something defined in the user Smithy model, while the other can be stuff like deserialization errors and so on..

rust-runtime/aws-smithy-http-server/src/exception.rs
david-perez: An alternative is to come up with a name of our own, different than exception, rejection, or error. So, name this type `X` and make the fact that `X::Z` in responses gets written as `ZException` just a detail of the protocol. After all, the fact that the word "exception" has been used in all AWS protocol errors might does not imply that it will continue to do so. Perhaps in the future a protocol we want to support (AWS or not) serializes an error as "Mistake", so maybe it's better to come up with our own name.

crisidev: The more I think about it, the more I want to get far away from `Exception` for the reason you provided above.. 

I think it is better if we come up with our name. In my mind this is an error, hence it should be suffixed with `Error`. What about `SmithyRuntimeError` or just `RuntimeError` ?

rust-runtime/aws-smithy-http-server/src/extension.rs
david-perez: We will rename `ResponseExtension` to `OperationExtension` and `ExtensionModeledError` to `ModeledErrorExtension`. We will add `RuntimeErrorExtension`. We're unsure whether we should panic or not when a request extension is not found. For now, let's keep returning a 500 response and include a `RequestExtensionNotFoundExtension` in the response.

rust-runtime/aws-smithy-http-server/src/exception.rs
david-perez: We think coming up with a new name is better. For now we'll go with `RuntimeError`.

rust-runtime/aws-smithy-http-server/src/exception.rs
```diff
@@ -0,0 +1,102 @@
...
+//!
+//! Generated code works always works with [`crate::rejection`] types when deserializing requests
+//! and serializing response. Just before a response needs to be sent, the generated code looks up
+//! and converts into the corresponding exception type, and then it uses the exception's
+//! [`axum_core::response::IntoResponse`] implementation to render and send a response.
+
+use crate::protocols::Protocol;
+
+#[derive(Debug)]
+pub enum SmithyFrameworkExceptionType {
```
david-perez, line 33: A better name for this would be to use `Kind` instead of `Type`.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -410,14 +410,14 @@ class HttpBindingGenerator(
...
+                index.getRequestBindings(shape, HttpLocation.PREFIX_HEADERS).getOrNull(0)
+            HttpMessageType.RESPONSE -> index.getResponseBindings(shape, HttpLocation.HEADER) to
+                index.getResponseBindings(shape, HttpLocation.PREFIX_HEADERS).getOrNull(0)
         }
-        val prefixHeaderBinding = when (httpMessageType) {
-            HttpMessageType.REQUEST -> index.getRequestBindings(shape, HttpLocation.PREFIX_HEADERS)
-            HttpMessageType.RESPONSE -> index.getResponseBindings(shape, HttpLocation.PREFIX_HEADERS)
-        }.getOrNull(0) // Only a single structure member can be bound to `httpPrefixHeaders`.
+
         if (headerBindings.isEmpty() && prefixHeaderBinding == null) {
```
rcoh, line 421: I think this is also in another PR?

rcoh (approved):
client changes are identical to #1218

crisidev (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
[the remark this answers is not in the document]
david-perez: In the end I had to remove this check from _all_ test types, because we now set extensions in the operation handler trait implementation, a code path that no tests exercise. See the added comments here https://github.com/awslabs/smithy-rs/pull/1233/files#diff-82b743775b074473ce5ed5ad9493413249796ca370ace0b1fa5921cb40211e28R473-R480.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
david-perez: Yes, now that I've merged `main`, the diff got slightly lighter.

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -32,24 +32,39 @@
...
 
-//! Extension extraction to share state across handlers.
-
-use super::rejection::{ExtensionHandlingRejection, ExtensionsAlreadyExtracted, MissingExtension};
-use async_trait::async_trait;
-use axum_core::extract::{FromRequest, RequestParts};
+//! Extension types.
+//!
+//! Extension types are types that are stored in and extracted from _both_ requests and
+//! responses.
```
crisidev, line 38:
```suggestion
//! Extension types are stored in and extracted from _both_ requests and
//! responses.
```

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -63,72 +78,70 @@ impl ResponseExtensions {
...
-pub struct ExtensionModeledError(&'static str);
-impl_extension_new_and_deref!(ExtensionModeledError);
+pub struct ModeledErrorExtension(&'static str);
+impl_extension_new_and_deref!(ModeledErrorExtension);
 
-/// Extension type used to store the type of framework error caught during execution.
-/// These are unmodeled error, or rejection, defined in the runtime crates.
+/// Extension type used to store the _name_ of the [`crate::runtime_error::RuntimeError`] that
+/// occurred during request handling (see [`crate::runtime_error::RuntimeErrorKind::name`]).
+/// These are _unmodeled_ errors; the operation handler was not invoked.
```
crisidev, line 89:
```suggestion
/// These are _unmodeled_ errors where the operation handler was not invoked.
```

rust-runtime/aws-smithy-http-server/src/runtime_error.rs
```diff
@@ -0,0 +1,106 @@
...
+//! and converts into the corresponding `RuntimeError`, and then it uses the its
+//! [`axum_core::response::IntoResponse`] implementation to render and send a response.
+
+use crate::protocols::Protocol;
+
+#[derive(Debug)]
+pub enum RuntimeErrorKind {
+    // UnknownOperation,
+    /// Request failed to deserialize or response failed to serialize.
+    Serialization(crate::Error),
```
crisidev, line 30: What about calling this Deserialization? It can be confusing otherwise.

rust-runtime/aws-smithy-http-server/src/rejection.rs
```diff
@@ -3,231 +3,222 @@
...
+
+    /// Used when the service implementer provides an integer outside the 100-999 range for a
+    /// member targeted by `httpResponseCode`.
+    InvalidHttpStatusCode,
+
+    /// Used when an invalid HTTP header value (a value that cannot be parsed as an
+    /// `[http::header::HeaderValue]`) is provided for a shape member bound to an HTTP header with
+    /// `httpHeader` or `httpPrefixHeaders`.
+    /// Used when failing to serialize an `httpPayload`-bound struct into an HTTP response body.
+    Build(crate::Error),
```
crisidev, line 79: Are you sure about this documentation? The first 3 lines look to be for something else.. The name `Build` is also not very significative to me.

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -63,72 +78,70 @@ impl ResponseExtensions {
...
-
 impl<T> Deref for Extension<T> {
     type Target = T;
 
     fn deref(&self) -> &Self::Target {
         &self.0
     }
 }
+
+pub async fn extract_extension<T, B>(
```
crisidev, line 127: Since this is public and used by the codegen, I'd add documentation.

rust-runtime/aws-smithy-http-server/src/lib.rs
```diff
@@ -11,25 +11,28 @@
 pub(crate) mod macros;
 
 pub mod body;
-pub mod error;
-mod extension;
+pub(crate) mod error;
 pub mod routing;
 
+#[doc(hidden)]
+pub mod extension;
```
crisidev, line 18: The extension module can be used by external entities such as a Tower metrics layer, so I believe it should be public and with docs as it will be used outside of the codegen.

rust-runtime/aws-smithy-http-server/src/runtime_error.rs
david-perez: This error is used both when a request fails to deserialize and when a response fails to serialize; AWS protocols use `SerializationException` for both.

rust-runtime/aws-smithy-http-server/src/rejection.rs
crisidev: Ah ok, got it. Thanks for the explanation.

PR #1235, opened 2022-03-03
Panic when converting float to int

david-perez:
## Checklist

### DIFF
--- a/rust-runtime/aws-smithy-types/src/lib.rs
+++ b/rust-runtime/aws-smithy-types/src/lib.rs
@@ -100,19 +100,33 @@ macro_rules! to_num_fn {
     };
 }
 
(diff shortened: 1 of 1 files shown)


david-perez:
I noticed that we were lossily converting from floating to integer types in both the client and the server, and raised this PR to see if CI exercised the conversion. It does not (the only test that fails is the one I added), which is not good.

Is this intentional? There are HTTP malformed request tests that test that floating point values are _rejected_ as integer values.

https://github.com/awslabs/smithy/blob/887ae4f6d118e55937105583a07deb90d8fabe1c/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-integer.smithy#L133-L106

PR #1238, opened 2022-03-04
Allow to specify a read buffer initial capacity when creating ByteStream from a file

tanguylebarzic:
## Motivation and Context

Tokio's ReaderStream::new called under the hood by ByteStream::from_file / ByteStream::from_path uses [a default buffer of 4k](https://github.com/tokio-rs/tokio/blob/d0dd74a0583b3ed7c8c7f68a4279fe9709bbf158/tokio-util/src/io/reader_stream.rs#L8). Users may desire to provide a different value, to trade off CPU usage against memory.

## Description

This PR adds 2 variations to ByteStream, `ByteStream::from_file_with_buffer_size` and `ByteStream::from_path_with_buffer_size`, that allow to specify a different buffer size. For a simple app that uploads many large-ish files (65Mb) to S3, using a higher value for the buffer size (8k, 16k, 32k...), I observed a strong decrease in CPU spent (both system and user), at the cost of an increase in memory usage (some figures below).

The behaviour of the existing ByteStream::from_file / ByteStream::from_path is unchanged (using a default buffer capacity of 4k, which corresponds to Tokio's ReaderStream default buffer capacity).

## Testing

Uploading files to S3 from disk. Some figures from my testing: 

| Buffer size (bytes) | Wall time (seconds) | User CPU time (seconds)  | System CPU time (seconds)  | Total CPU (seconds)  | Max memory (Mb) |
|---|---|---|---|---|---|
| 4096 (default) | 32.7 | 28.14 | 22.3  | 50.44 | 47  |
| 8196  |  33.06 | 14.44  | 11.69  |  26.13 |  97 |
| 16392  |  32.57 | 14.51  |  12.36 | 26.87  |  89 |
| 32784  |  32.78 | 11.39  |  8.93 | 20.32  |  103 |

As you can see, while there's no reduction in the time to upload these files, using higher buffer sizes reduces the CPU usage considerably (divided by 2.5x), at a cost of more memory usage (2x). In my case, I mostly care about limiting the CPU, so making this trade-off makes sense.

## Checklist


Velfi:
This is a great idea, thanks for submitting! We'll get to reviewing this as soon as we're able.

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -24,19 +24,22 @@ use tokio_util::io::ReaderStream;
 pub struct PathBody {
     state: State,
     len: u64,
```
rcoh, line 26: can we rename this to `file_size` (now that `buffer_size` is also there it's slightly confusing)

rcoh (commented):
Overall, I really like the new functionality here. What I'm less sure about is the potential for a bit of API explosion. (eg. if we ever end up with another parameter on loading bodies from a path) Thinking about some other options, revolving around making a builder for `PathBody`. 

We would make `PathBody` public, and create a builder for it to allow code like:

`PathBody::with_capacity(1024).from_file(...).byte_stream()`. For "vanilla" use cases, `from_path` and `from_file` would still provide a direct interface.

still considering all the possible options though, let me know if you have any other thoughts.

tanguylebarzic:
> Overall, I really like the new functionality here. What I'm less sure about is the potential for a bit of API explosion. (eg. if we ever end up with another parameter on loading bodies from a path) Thinking about some other options, revolving around making a builder for `PathBody`.
> 
> We would make `PathBody` public, and create a builder for it to allow code like:
> 
> `PathBody::with_capacity(1024).from_file(...).byte_stream()`. For "vanilla" use cases, `from_path` and `from_file` would still provide a direct interface.
> 
> still considering all the possible options though, let me know if you have any other thoughts.

Agreed that just adding a function each time there's a new parameter isn't great, and makes for poor discoverability - I like the builder idea, will give it a shot!

rcoh:
we also recently discussed the need to specify an offset/length into the path. Let's do a builder!

tanguylebarzic:
@rcoh Sorry for the delay, updated the PR with a builder. I used the opportunity to also accept the file size as an optional argument, for cases where it's known by the caller (no need to call the metadata in this case).
Not a fan of the naming (`PathBodyBuilder`). I also hesitated to add a function `from_builder(builder: PathBodyBuilder)` in ByteStream, to ease discovery (and have it appear along with `from_path` and `from_file` in the doc), but ended up just adding a comment to from_path.

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -95,6 +95,30 @@
 //! }
 //! }
 //! ```
+//!
+//! If you want more control over how the file is read, you can use PathBodyBuilder. This allows to specify
+//! options such as the size of the buffer used to read the file, the size of the file to read if known...
+// !
```
Velfi, line 101:
```suggestion
//!
```

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -535,4 +547,46 @@ mod tests {
...
+        use std::io::Write;
+        use tempfile::NamedTempFile;
+        let mut file = NamedTempFile::new()?;
+
+        for i in 0..10000 {
+            writeln!(file, "Brian was here. Briefly. {}", i)?;
+        }
+        let body = PathBodyBuilder::from_path(&file)
+            .with_buffer_size(16384)
+            // This isn't the right file length - one shouln't done that in real life
```
Velfi, line 566:
```suggestion
            // This isn't the right file length - one shouldn't do this in real code
```

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -8,13 +8,19 @@ use futures_core::{ready, Stream};
...
 use std::task::{Context, Poll};
 use tokio::fs::File;
 use tokio::io;
 use tokio_util::io::ReaderStream;
 
+use crate::body::SdkBody;
+
+use super::{ByteStream, Error};
+
+const DEFAULT_BUFFER_SIZE: usize = 4096;
```
Velfi, line 22: Could you write a comment about how this buffer size was chosen?

tanguylebarzic: Done

rust-runtime/aws-smithy-http/src/byte_stream.rs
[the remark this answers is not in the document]
tanguylebarzic: Fixed

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -23,20 +30,117 @@ use tokio_util::io::ReaderStream;
...
-            state: State::Unloaded(path.to_path_buf()),
-            len,
+            state: State::Unloaded(path_buf),
+            file_size,
+            buffer_size,
         }
     }
-    pub fn from_file(file: File, len: u64) -> Self {
+    pub fn from_file(file: File, file_size: u64, buffer_size: usize) -> Self {
         PathBody {
```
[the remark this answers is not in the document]
tanguylebarzic: I believe they actually need to be public to be able to call PathBodyBuilder directly (as in the example usage).

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -23,20 +30,117 @@ use tokio_util::io::ReaderStream;
...
-            len,
+            state: State::Loaded(ReaderStream::with_capacity(file, buffer_size)),
+            file_size,
+            buffer_size,
+        }
+    }
+}
+
+/// Builder for creating [ByteStreams](crate::byte_stream::ByteStream) from a file/path, with full control over advanced options.
+///
```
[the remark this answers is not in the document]
tanguylebarzic: Added a usage example (which mentions the default buffer size).

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -23,20 +30,117 @@ use tokio_util::io::ReaderStream;
...
+pub struct PathBodyBuilder {
+    file: Option<tokio::fs::File>,
+    path: Option<PathBuf>,
+    file_size: Option<u64>,
+    buffer_size: usize,
+}
+
+impl PathBodyBuilder {
+    /// Create a PathBodyBuilder from a path.
+    ///
```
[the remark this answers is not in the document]
tanguylebarzic: Done

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -23,20 +30,117 @@ use tokio_util::io::ReaderStream;
...
+            path: None,
+            file_size: None,
+            buffer_size: DEFAULT_BUFFER_SIZE,
+        }
+    }
+
+    /// Specify the length of the file to read (in bytes).
+    ///
+    /// If not used, [byte_stream](PathBodyBuilder::byte_stream) will require an extra call to query the file's metadata.
+    ///
```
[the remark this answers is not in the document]
tanguylebarzic: Agreed, updated.

rcoh (commented):
Thanks for your continued iteration on this! The team got a chance to bike shed the API and this is what we came up with in terms of the UX:

```rust
// keep the current API
ByteStream::from_path("myfile.txt").await

// deprecate `ByteStream::from_file` and refer people to `read_from`

// New API `ByteStream::read_from` which returns an empty builder:
// - adds `.path` and `.file` methods to the builder
// - renames `with_buffer_size` to `buffer_size`
// - renames `with_file_size` to `file_size`
// - renames `byte_stream` to `build` (since most people will access this _from_ `ByteStream`
ByteStream::read_from().path(path).build().await?
ByteStream::read_from().file(file).build().await
ByteStream::read_from().path(path).buffer_size(16092).build().await

// rename to `PathBodyBuilder` to `FsBuilder` (which uses can get with `byte_stream::FsBuilder`

Thanks for your patience while we tweak this!

File: rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs (hunk 5 above)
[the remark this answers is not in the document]
Comment (tanguylebarzic) [reply rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs:46]: You are right, I mistakenly thought you were referring to the PathBodyBuilder functions.

**Comment by tanguylebarzic** [comment]
> Thanks for your continued iteration on this! The team got a chance to bike shed the API and this is what we came up with in terms of the UX:
> 
> ```rust
> // keep the current API
> ByteStream::from_path("myfile.txt").await
> 
> // deprecate `ByteStream::from_file` and refer people to `read_from`
> 
> // New API `ByteStream::read_from` which returns an empty builder:
> // - adds `.path` and `.file` methods to the builder
> // - renames `with_buffer_size` to `buffer_size`
> // - renames `with_file_size` to `file_size`
> // - renames `byte_stream` to `build` (since most people will access this _from_ `ByteStream`
> ByteStream::read_from().path(path).build().await?
> ByteStream::read_from().file(file).build().await?
> ByteStream::read_from().path(path).buffer_size(16092).build().await?
> 
> // rename to `PathBodyBuilder` to `FsBuilder` (which uses can get with `byte_stream::FsBuilder`
> 
> Thanks for your patience while we tweak this!
> ```

Updated the PR to implement this API. I like it, especially as it it's more discoverable directly from ByteStream. It does expose to 2 potential misuses though - calling neither path nor file on the FsBuilder, or calling both of them. Happy to change the implemented behaviour - I decided to panic when neither has been called (as it's probably a programmer error, it seemed better than returning an error) and favor `path` over `file` when both have been called.

File: rust-runtime/aws-smithy-http/src/byte_stream.rs [hunk 9]
```diff
@@ -262,36 +293,27 @@ impl ByteStream {
...
-        let body = SdkBody::from_dyn(http_body::combinators::BoxBody::new(
-            bytestream_util::PathBody::from_file(file, sz),
-        ));
-        Ok(ByteStream::new(body))
+        FsBuilder::new().file(file).build().await
+    }
+
+    #[cfg(feature = "rt-tokio")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "rt-tokio")))]
+    pub fn read_from() -> FsBuilder {
```
Velfi, line 315: Would you mind adding a short doc here sending people to the `FsBuilder` docs?

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -8,35 +8,169 @@ use futures_core::{ready, Stream};
...
+    /// Increasing the read buffer capacity to higher values than the default (4096 bytes) can result in a large reduction
+    /// in CPU usage, at the cost of memory increase.
+    pub fn buffer_size(mut self, buffer_size: usize) -> Self {
+        self.buffer_size = buffer_size;
+        self
+    }
+
+    /// Returns a [`ByteStream`](crate::byte_stream::ByteStream) from this builder.
+    /// NOTE: If both [`file`](FsBuilder::file) and [`path`](FsBuilder::path) have been called for this FsBuilder, `build` will
+    /// behave read from the path specified by [`path`](FsBuilder::path).
```
Velfi, line 138: Would you mind leaving a short note here about how this can panic if no setters were called before building? Something like
```suggestion
    /// Returns a [`ByteStream`](crate::byte_stream::ByteStream) from this builder.
    /// NOTE: If both [`file`](FsBuilder::file) and [`path`](FsBuilder::path) have been called for this FsBuilder, `build` will
    /// behave read from the path specified by [`path`](FsBuilder::path).
    ///
    /// # Panics
    ///
    /// Panics if neither of the `file` or`path` setters were called.
```

Velfi (changes requested):
Looks good, I just have two suggestion for docs and then I'm ready to approve and merge.

rust-runtime/aws-smithy-http/src/byte_stream.rs
rcoh: an example (can be the same one from above) would also be great.

And a small side nit: can you put this method above the two deprecated ones? it makes the docs a little nicer

rcoh (commented):
this is looking great! Thanks again!

rust-runtime/aws-smithy-http/src/byte_stream.rs
tanguylebarzic: Added comment + example. 

> can you put this method above the two deprecated ones?
Moved it higher. To be clear, there's only one deprecated method at the method (from_file). Do you want me to deprecate from_path as well (and encourage the use of `ByteStream::read_from().path(x)` instead?

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -95,6 +95,30 @@
 //! }
 //! }
 //! ```
+//!
+//! If you want more control over how the file is read, such as specifying the size of the buffer used to read the file
+//! or the length of the file, use a [`FsBuilder`](crate::byte_stream::FsBuilder).
```
rcoh, line 100:
```suggestion
//! or the length of the file, use an [`FsBuilder`](crate::byte_stream::FsBuilder).
```

rcoh (approved):
LGTM!

Velfi (approved):

rcoh:
CI failure is Clippy being upset

rcoh:
I fixed the clippy errors (hope that's alright). Will try to get this merged today. Thanks again @tanguylebarzic !

tanguylebarzic:
> I fixed the clippy errors (hope that's alright). Will try to get this merged today. Thanks again @tanguylebarzic !

Missed the clippy errors, thank you for taking care of it and merging!

PR #1255, opened 2022-03-11
Use Docker build image in CI

jdisanti:
## Motivation and Context
This creates a Docker image stored in Public ECR for use with smithy-rs CI so that tools don't need to be recompiled for every CI run.


.github/workflows/ci.yaml
```diff
@@ -7,100 +7,192 @@ on:
 
 name: CI
 
+concurrency:
+  group: ci-yaml-${{ github.ref }}
+  cancel-in-progress: true
```
Velfi, line 12: ❤️

buildSrc/src/main/kotlin/ExecRustBuildTool.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    private fun runCli(commandLine: List<String>, workingDirectory: File? = null) {
+        getExecActionFactory().newExecAction().let { action ->
+            getObjectFactory()
+                .newInstance(DefaultExecSpec::class.java)
+                .apply {
+                    commandLine(commandLine)
+                    if (workingDirectory != null) {
+                        workingDir(workingDirectory)
+                    }
```
Velfi, line 53: Since we disallow leaving the working dir as `null` is disallowed by `run`, why not just make the `workingDirectory` non-optional for this function?

tools/ci-build/build-image.dockerfile
```diff
@@ -0,0 +1,124 @@
...
+# SPDX-License-Identifier: Apache-2.0.
+#
+ARG base_image=public.ecr.aws/amazonlinux/amazonlinux:2
+FROM ${base_image} AS bare_base_image
+
+#
+# Node Installation Stage
+#
+FROM bare_base_image AS install_node
+ARG node_version=v16.14.0
```
Velfi, line 12: For stuff like this, should we be using our ticket-creating bot to encouraged ourselves to bump this version with some regularity?

tools/ci-build/build-image.dockerfile
```diff
@@ -0,0 +1,124 @@
...
+ARG node_bundle_sha256=0570b9354959f651b814e56a4ce98d4a067bf2385b9a0e6be075739bc65b0fae
```
Velfi, line 13: This is the SHA of the `node.tar.xz` for the version on the line above right?

Velfi (approved):
In awe of the size of this lad

buildSrc/src/main/kotlin/ExecRustBuildTool.kt
jdisanti: I think you're confusing the tool path and the working directory for this function. It's set to null explicitly on line 36 for the case where we don't need to invoke cargo.

tools/ci-build/build-image.dockerfile
jdisanti: Yeah

tools/ci-build/README.md
```diff
@@ -0,0 +1,39 @@
+ci-build
+========
+
+This directory includes the build Docker image and scripts to use it in CI.
+- `acquire-build-image`: Script that retrieves the build image from public ECR or builds one locally
+  depending on the state of the tools directory. If the tools have changed (in git history), then it opts
+  to build a new image rather than reuse an existing one.
+- `add-local-user.dockerfile`: Creates a user in the build image with the host's user ID
+- `build-image.dockerfile`: Dockerfile used to create the build image
```
rcoh, line 9: not required, but a more typical design would probably be a directory with a dockerfile in it instead of named dockerfiles

```
base-image/Dockerfile
add-local-user/Dockerfile
...
```
this is because there is tooling around a dockerfile being named `Dockerfile`

tools/ci-build/README.md
```diff
@@ -0,0 +1,39 @@
+ci-build
```
rcoh, line 1: not sure how this ends up working, but I usually end up also needing a `.dockerignore` file to prevent unnecessary files being sent to the docker daemon when building. If docker builds stall for a few seconds when you start them locally, that may be why

tools/ci-build/README.md
```diff
@@ -0,0 +1,39 @@
...
+  this is equivalent to the GitHub Actions working directory.
+- **Action space:** Temporary directory maintained by `ci-action` where the CI checks actually run.
+
+To create the starting space, do the following:
+
+```bash
+cd /path/to/my/starting-space
+git clone https://github.com/awslabs/smithy-rs.git
+# Optionally check out the revision you want to work with in the checked out smithy-rs.
+# Just make sure you are in /path/to/my/starting-space (or whatever you called it) after.
```
rcoh, line 27: seems like we might eventually want to automate this to provide testing directly against a specified githash or tag

tools/ci-build/add-local-user.dockerfile
```diff
@@ -0,0 +1,11 @@
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+#
+ARG base_image=smithy-rs-build-image-original
+FROM ${base_image} AS bare_base_image
+
```
rcoh, line 7: definitely worth a comment about why this is needed

tools/ci-build/acquire-build-image
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+#
+
+set -ue
+
+REMOTE_IMAGE_NAME="public.ecr.aws/w0m4q9l7/github-awslabs-smithy-rs-ci:latest"
```
rcoh, line 9: I'm a _little_ hesitant about always using latest as the tag of the image. If we fail to publish the docker image on main, we would never find out.

If we have main always tag with the commit hash, it seems like we could use the commit has of main here instead?

tools/ci-build/build-image.dockerfile
```diff
@@ -0,0 +1,124 @@
...
+    cargo install --path tools/sdk-lints; \
+    cargo install --path tools/sdk-versioner; \
+    chmod g+rw -R /opt/cargo/registry
+
+#
+# Final image
+#
+FROM bare_base_image AS final_image
+ARG rust_stable_version=1.56.1
+ARG rust_nightly_version=nightly-2022-03-03
```
rcoh, line 91: does using `ARG` for all these things make them easily configurable without modifying the image?

tools/ci-build/build-image.dockerfile
```diff
@@ -0,0 +1,124 @@
...
+    rm rustup-init; \
+    rustup --version; \
+    rustup component add rustfmt; \
+    rustup component add clippy; \
+    rustup install ${rust_nightly_version}; \
+    cargo --version; \
+    cargo +${rust_nightly_version} --version;
+RUN set -eux; \
+    cargo +${rust_nightly_version} install cargo-udeps --version ${cargo_udeps_version}; \
+    cargo install cargo-hack --version ${cargo_hack_version}; \
```
rcoh, line 76: hack doesn't require nightly?

tools/ci-build/build.docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+  smithy-rs-build:
+    image: smithy-rs-build-image:latest
+    user: ${USER_ID}:build
+    volumes:
+    - type: bind
+      source: ./workspace
+      target: /home/build/workspace
+    - type: bind
+      source: ${GRADLE_CACHE_PATH}
+      target: /home/build/.gradle
```
rcoh, line 16: this basically just exists for local dev, right? probably worth a comment

tools/ci-build/ci-action
```diff
@@ -0,0 +1,58 @@
...
+
+cd "${ACTION_PATH}"
+export USER_ID="$(id -u)"
+export GROUP_ID="$(id -g)"
+export GRADLE_CACHE_PATH="${START_PATH}/gradle"
+docker-compose -f build.docker-compose.yml up -d
+if [[ "${ACTION_NAME}" == "--shell" ]]; then
+    docker-compose -f build.docker-compose.yml exec smithy-rs-build /bin/bash
+else
+    docker-compose -f build.docker-compose.yml exec -T smithy-rs-build "./scripts/${ACTION_NAME}" "${ACTION_ARGS}"
```
rcoh, line 56: does this work? I assume we would _want_ arg splitting here?

tools/ci-build/create-local-build-image
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
rcoh, line 1: not opposed to all these bash scripts necessarily but we should definitely start using shellcheck

rcoh (commented):
Overall looks great!
Some general thoughts:
- this adds a lot of bash scripts. I guess there isn't really a better option? Using Python or Rust probably isn't much better, but I'm curious of you have thoughts
- In general, a sentence or two at the top of each file explaining the reason they exist would probably be helpful

Do we also need to update the required jobs after this?

.github/workflows/ci.yaml
```diff
@@ -110,61 +204,76 @@ jobs:
...
+        for runtime_path in "rust-runtime" "aws/rust-runtime"; do
+          pushd "${runtime_path}" &>/dev/null
+          cargo test --all-features
+          cargo doc --no-deps --document-private-items --all-features
+          popd &>/dev/null
+        done
 
-  runtime-bundle:
-    name: Produce smithy-rs runtime bundle
+  check-sdk-examples:
```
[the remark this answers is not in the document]
jdisanti: If the examples fail, we don't want to prevent a PR from being merged, so they have to run as their own (optional) job.

.github/workflows/github-pages.yaml
```diff
@@ -5,6 +5,10 @@ on:
     paths:
     - design/**
 
+concurrency:
+  group: github-pages-yml
```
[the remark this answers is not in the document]
jdisanti: This workflow only runs on push to main, so this limits it to one build per main push and there won't be a ref available.

tools/ci-build/build-image.dockerfile
jdisanti: Mostly it keeps these values in one place so that it's obvious where to change their values. I don't have anything instrumented to override these.

tools/ci-build/build-image.dockerfile
jdisanti: I don't think so.

tools/ci-build/build.docker-compose.yml
jdisanti: It's actually being used by the `ci-action` script in CI.

tools/ci-build/scripts/check-style-and-lints
```diff
@@ -8,6 +8,6 @@ set -eux
 cd workspace/smithy-rs
 sdk-lints check --all
 sdk-lints fix --all
-rustfmt --check --edition 2018 $(find -name '*.rs' -print | grep -v /target/)
+rustfmt --check --edition 2018 "$(find . -name '*.rs' -print | grep -v /target/)"
```
rcoh, line 11: hmm...I think one of these works on gnu find and one doesn't IIRC?

rcoh (approved):
LGTM!

.github/workflows/ci.yaml
```diff
@@ -110,61 +204,76 @@ jobs:
...
-        jobs: ${{ toJSON(needs) }}
+    - name: Run tests
+      shell: bash
+      run: |
+        for runtime_path in "rust-runtime" "aws/rust-runtime"; do
+          pushd "${runtime_path}" &>/dev/null
+          cargo test --all-features
+          cargo doc --no-deps --document-private-items --all-features
+          popd &>/dev/null
+        done
```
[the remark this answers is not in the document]
jdisanti: It's similar to, but not the same as, the `check-rust-runtimes-and-tools` script. It doesn't run clippy on the Rust runtimes though, and it also doesn't check the tools since we don't use the tools on Windows.

tools/ci-build/README.md
jdisanti: I haven't noticed any issues here. If it sends the directory the Dockerfile is from, then those files are trivially small in this case.

tools/ci-build/scripts/check-style-and-lints
jdisanti: I tried the command out on a Mac, and it seems to work fine there. Regardless, these scripts get run in the Docker build image, which has the GNU coreutils.

Velfi (approved):

Velfi (approved):

PR #1264, opened 2022-03-17
Improve Pokémon Service `Makefile`

david-perez:
* Code-generate only the Pokémon Service SDK (as opposed to all targets
  in the `codegen-server-test` subproject like we do currently).
* Add `run` target to run the service.
* Add `doc-open` target to generate and open Rustdoc documentation.


rust-runtime/aws-smithy-http-server/examples/Makefile
```diff
@@ -9,15 +9,20 @@ CLIENT_SDK_SRC := $(SRC_DIR)/codegen-test/build/smithyprojections/codegen-test/p
 all: codegen
 
 codegen:
-	$(GRADLE) --project-dir $(SRC_DIR) :codegen-test:assemble
```
crisidev, line 12: This line have to remain as we need to generate the client bindings as well.

crisidev (approved):

PR #1267, opened 2022-03-22
Add miscellaneous protocol test with nested required shape

david-perez:
This commit adds another suite of protocol tests to test miscellaneous
aspects of code generation where protocol selection should not be
relevant.

The first test is code-generating an operation with a nested required
shape, minimally reproducing the bug that was fixed in 75056c, so that
we don't regress on it. I've added it to both the `codegen-test` and
`codegen-server-test` subprojects, even though the bug only affected the
server, in part to also incentivize us into adding more tests to this
Smithy model for both clients and servers.

## Testing

Cherry-picked this commit on top of v0.38, and verified that it fails the server integration tests.

---


rcoh (approved):

crisidev:
Nice! Once this is merged, I will add another test here to cover what has been found in #1269 and #1273

crisidev (approved):

PR #1269, opened 2022-03-22
Bugfix: prevent errors like "move occurs because `input.element` has type `Blob`, which does not implement the `Copy` trait"

crisidev:
I believe we do not have a protocol test checking this code path..

Let's see what the CI think about this.. my local tests succeeded..


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -347,7 +347,7 @@ class JsonSerializerGenerator(
                 )
             }
             is BlobShape -> rust(
-                "$writer.string_unchecked(&#T(${value.name}));",
+                "$writer.string_unchecked(&#T(&${value.name}));",
```
rcoh, line 350: I think this should be `context.valueExpression.asRef`

rcoh (approved):
LGTM to get a fix out, but this changes the generated client SDK so we should triple check it won't cause any lint errors. See inline comment.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
crisidev: Makes sense. I will fix this before merging tomorrow morning. Thanks for the approval.

PR #1273, opened 2022-03-23
Use `value.asRef()` to let Kotlin do the right thing in case of values or references

crisidev:
## Motivation and Context
This change properly fixes the bug found in #1269 by using `asRef()` from `ValueExpression.kt`.


david-perez (approved):

crisidev:
Superseded by #1275

PR #1274, opened 2022-03-23
Refactor converters to numeric types for `aws_smithy_types::Number`

david-perez:
Currently, conversions from `aws_smithy_types::Number` into numeric Rust
types (`{i,u}{8, 16, 32, 64}` and `f{32, 64}`) are always lossy, because
they use the `as` Rust keyword to cast into the target type. This means
that clients and servers are accepting lossy data: for example, if an
operation is modeled to take in a 32-bit integer as input, and a client
incorrectly sends an integer number that does not fit in 32 bits, the
server will silently accept the truncated input. There are malformed
request protocol tests that verify that servers must reject these
requests.

This commit removes the lossy `to_*` methods on `Number` and instead
implements `TryFrom<Number> for $typ` for the target numeric type
`$typ`. These converters will attempt their best to perform the
conversion safely, and fail if it is lossy.

The code-generated JSON parsers will now fail with
`aws_smithy_json::deserialize::ErrorReason::InvalidNumber` if the number
in the JSON document cannot be converted into the modeled integer type
without losing precision. For floating point target types, lossy
conversions are still performed, via `Number::to_f32_lossy` and
`Number::to_f64_lossy`.

## Checklist


rust-runtime/aws-smithy-types/src/lib.rs
```diff
@@ -80,46 +80,452 @@ pub enum Document {
...
+    UnexpectedNegativeInteger(i64),
+}
+
+impl std::fmt::Display for TryFromNumberError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            TryFromNumberError::OutsideIntegerRange(err) => write!(f, "integer too large: {}", err),
+            TryFromNumberError::UnexpectedFloat(v) => write!(
+                f,
+                "cannot convert {} floating point number into an integer",
```
crisidev, line 137:
```suggestion
                "cannot convert floating point number {} into an integer",
```

rust-runtime/aws-smithy-types/src/lib.rs
```diff
@@ -80,46 +80,452 @@ pub enum Document {
...
+                v
+            ),
+            TryFromNumberError::UnexpectedNegativeInteger(v) => write!(
+                f,
+                "cannot convert {} negative integer into an unsigned integer type",
```
crisidev, line 142:
```suggestion
                "cannot convert negative integer {} into an unsigned integer type",
```

rust-runtime/aws-smithy-types/src/lib.rs
```diff
@@ -80,46 +80,452 @@ pub enum Document {
...
+to_signed_integer_converter!(i8);
+
+/// Converts to an `f64`. The conversion can only fail if it is lossy.
+impl TryFrom<Number> for f64 {
+    type Error = TryFromNumberError;
+
+    fn try_from(value: Number) -> Result<Self, Self::Error> {
+        match value {
+            // Integers can only be represented with full precision in a float if they fit in the
+            // significand, which is 24 bits in `f32` and 53 bits in `f64`.
```
crisidev, line 258: TIL: significand == mantissa

crisidev (approved):
I left a couple of simple comments and suggestions. LGTM overall. Nice test suite!

rcoh (approved):
Overall LGTM. It seems like it may be possible to refactor the converters to share macros but probably not worth it.

I may consider adding a proptest of some invariants as well.

rust-runtime/aws-smithy-json/src/deserialize/error.rs
```diff
@@ -82,3 +82,12 @@ impl From<EscapeError> for Error {
         }
     }
 }
+
+impl From<aws_smithy_types::TryFromNumberError> for Error {
+    fn from(_: aws_smithy_types::TryFromNumberError) -> Self {
+        Error {
+            reason: ErrorReason::InvalidNumber,
+            offset: None,
```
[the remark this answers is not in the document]
david-perez: We don't have the offset here. We only have the offset when working with the token directly in the code-generated deserializer, or when calling the `expect_` functions which auto-fill it from the token in case of errors. When unescaping strings, we're also bubbling up errors without offsets. Making both cases bubble up offsets would require some refactoring.

rust-runtime/aws-smithy-types/src/lib.rs
```diff
@@ -80,46 +80,452 @@ pub enum Document {
...
 pub enum Number {
-    /// Unsigned 64-bit integer value
+    /// Unsigned 64-bit integer value.
     PosInt(u64),
-    /// Signed 64-bit integer value
+    /// Signed 64-bit integer value. The wrapped value is _always_ negative.
     NegInt(i64),
-    /// 64-bit floating-point value
+    /// 64-bit floating-point value.
     Float(f64),
```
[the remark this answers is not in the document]
david-perez: What invariants? I don't understand the suggestion. These variants are being directly instantiated here:

https://github.com/awslabs/smithy-rs/blob/6130fcb9bd14e385dcd9be7cc6832771804c487a/rust-runtime/aws-smithy-json/src/deserialize.rs#L313-L349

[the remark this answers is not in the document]
rcoh: by invariants I mean: `    /// Signed 64-bit integer value. The wrapped value is _always_ negative.` 
We would add constructor functions, `::pos_int`, `::neg_int`, `::float`.

We would need to refactor code to use those. `neg_int` could panic if you passed in a positive int (or return a result, eg.)

not a blocker, just a possibility

CHANGELOG.next.toml
```diff
@@ -34,3 +34,9 @@ message = "Update all SDK and runtime crates to [edition 2021](https://blog.rust
 references = ["aws-sdk-rust#490"]
 meta = { "breaking" = true, "tada" = false, "bug" = false }
 author = "Velfi"
+
+[[smithy-rs]]
+message = "Refactor converters to numeric types for `aws_smithy_types::Number`"
```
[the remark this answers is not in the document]
david-perez: I doubt any are performing these conversions themselves, but here's the updated changelog: https://github.com/awslabs/smithy-rs/pull/1274/commits/c21572dab734d43892e35cb5a17d46cab7fd9ca4

david-perez:
Should I add proptesting? In my tests I tested one or two values within the target type's range, as well as the values in the edges of the range (plus `NaN` and `+-Infinity`). If I add proptesting it would look like:

```rust
assert_eq!($typ::try_from(Number::PosInt(v)).unwrap(), v);
```

with `v` a random value within the range of `$typ`. Same with random values outside the range.

CHANGELOG.next.toml
```diff
@@ -58,3 +58,41 @@ message = "Update urlencoding crate to v2.1.0"
 references = ["smithy-rs#1301"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "benesch"
+
+[[smithy-rs]]
```
jdisanti, line 62: This needs to be duplicated in the SDK changelog with `[[aws-sdk-rust]]` since it will impact SDK customers as well.

I want to see the codegen diff as part of reviewing this, so I'll merge the latest from main and just make this change as part of the merge conflict resolution.

rust-runtime/aws-smithy-types/src/lib.rs
```diff
@@ -80,49 +80,455 @@ pub enum Document {
...
+    U64ToFloatLossyConversion(u64),
+    /// Used when the conversion from an `i64` into a floating point type would be lossy.
+    I64ToFloatLossyConversion(i64),
+    /// Used when attempting to convert an `f64` into an `f32`.
+    F64ToF32LossyConversion(f64),
+    /// Used when attempting to convert a decimal, infinite, or `NaN` floating point type into an
+    /// integer type.
+    UnexpectedFloat(f64),
+    /// Used when attempting to convert a negative [`Number`] into an unsigned integer type.
+    UnexpectedNegativeInteger(i64),
```
jdisanti, line 130: I think these should get renamed to something like `FloatToIntegerLossyConversion` and `NegativeToUnsignedLossyConversion`, since "unexpected" really only applies to deserialization.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -262,8 +262,23 @@ class JsonParserGenerator(
...
+        if (target.isFloatShape) {
+            rustTemplate("#{expect_number_or_null}(tokens.next())?.map(|v| v.to_f32_lossy())", *codegenScope)
+        } else if (target.isDoubleShape) {
+            rustTemplate("#{expect_number_or_null}(tokens.next())?.map(|v| v.to_f64_lossy())", *codegenScope)
+        } else {
+            rustTemplate(
+                """
+                #{expect_number_or_null}(tokens.next())?
+                    .map(|v| {
+                        use std::convert::TryInto;
```
jdisanti, line 274: I think this can be removed now since we upgraded to edition 2021, and removing it should fix CI.
```suggestion
```

jdisanti (approved):
Overall, this looks good to me. Just a couple minor things.

rust-runtime/aws-smithy-types/src/lib.rs
david-perez: https://github.com/awslabs/smithy-rs/pull/1274/commits/e17da696d2b01e44dd7031df00150bad6e87335b

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/JsonParserGenerator.kt
david-perez: https://github.com/awslabs/smithy-rs/pull/1274/commits/afffb9c275bd8a56792c3d7f3e681cbf156e9016

david-perez:
@jdisanti @rcoh I'll merge this tomorrow if there are no further comments.

PR #1275, opened 2022-03-23
Fix serialization of required shapes (timestamps and nested shapes)

david-perez:
In #1148, `@required` started being strictly interpreted by server SDKs.

That meant that when serializing values from structures, we no longer
borrow for every shape when looking into `Option`s as in:

```
if let Some(var_37) = &input.value
```

(See for example `RustWriter.serializeStructure` from
`JsonSerializerGenerator.kt` for the relevant serialization
code-generation routine)

Instead, we attempt to serialize the unborrowed required shape value.
This works well for booleans, numeric types (since they are `Copy`),
collections (since we borrow their items while iterating), and strings
(since we call `as_str()`). It also worked for unions and documents,
since we already borrowed (note that `ValueExpression.kt` makes sure to
not borrow twice if the value is already a reference, like when clients
look into `Option`s).

We fixed borrowing of blobs in #1269.

However, we currently don't borrow for timestamps and nested shapes.
This commit fixes that, and adds a comprehensive protocol test to ensure
we exercise all the lines relevant to serialization of required shapes.

---

### DIFF
--- a/codegen-test/model/misc.smithy
+++ b/codegen-test/model/misc.smithy
@@ -15,7 +15,8 @@ service MiscService {
     ],
 }
 
-/// To not regress on https://github.com/awslabs/smithy-rs/pull/1266
+/// This operation tests that (de)serializing required values from a nested
+/// shape works correctly.
 @http(uri: "/operation", method: "GET")
 operation OperationWithInnerRequiredShape {
     input: OperationWithInnerRequiredShapeInput,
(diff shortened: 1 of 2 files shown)


rcoh (approved):

david-perez:
I amended commit message and PR title; this only affects **se**rialization.

crisidev (approved):
Nice! The added protocol tests will allow us to prevent regressions on this!

PR #1278, opened 2022-03-25
Smithy async docs

rcoh:
## Motivation and Context
Adds more docs to the runtime crates as well as the codegen

## Testing
- ci


rust-runtime/aws-smithy-async/src/lib.rs
```diff
@@ -3,6 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
+#![warn(
```
jdisanti, line 6: Nice!

rust-runtime/aws-smithy-eventstream/src/smithy.rs
```diff
@@ -32,14 +35,28 @@ expect_shape_fn!(fn expect_byte_array[ByteArray] -> Blob { bytes -> Blob::new(by
 expect_shape_fn!(fn expect_string[String] -> String { value -> value.as_str().into() });
 expect_shape_fn!(fn expect_timestamp[Timestamp] -> DateTime { value -> *value });
 
+/// Structured header data from a [`Message`]
 #[derive(Debug)]
 pub struct ResponseHeaders<'a> {
+    /// Content Type of the message
+    ///
+    /// For binary data, this will be `application/octet-stream`. For strings, this will be `text/plain`.
```
jdisanti, line 43: This can be a number of things depending on the protocol. For example, if the protocol is AwsJson1, then this could be `application/json`, or `application/xml` for RestXml. It'll be `application/octet-stream` if there is a `Blob` payload shape, and `text/plain` if there is a `String` payload shape.

jdisanti (approved):
Thanks for adding these!

PR #1279, opened 2022-03-28
Initial awsJson 1.0 and 1.1 server implementation

crisidev:
## Motivation and Context
This PR add the initial implementation of awsJson [1.0](https://awslabs.github.io/smithy/1.0/spec/aws/aws-json-1_0-protocol.html) and [1.1](https://awslabs.github.io/smithy/1.0/spec/aws/aws-json-1_1-protocol.html) protocols.

## Description
To support these new protocols, we also changed the runtime `Router` definition as for both awsJson 1.0 and 1.1, every request MUST be sent to the root URL (/) using the HTTP "POST" method. 

The runtime `Router` now supports explicitly all the available protocols.

There are still some protocol tests failing for awsJson 1.0 and 1.1. The failure are caused by the missing implementation of `@endpoint` trait and date parsing. Protocol tests for these 2 protocols are heavily biased towards `Responses`. Before announcing support for awsJson 1.0 and 1.1, we should increase the protocol tests coverage.

## Testing
Testing done generating protocol tests.


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -151,11 +152,19 @@ class JsonSerializerGenerator(
...
      * This function is only used by the server, the client uses directly [serializeStructure].
+     *
+     * To obey to protocol implementations such as AwsJson 1.0 and 1.1, this methods accepts two boolean
+     * arguments, `includeErrorType` and `useErrorNamespace`, controlling the generation of the `__type`
+     * field in the response.
+     * See: https://awslabs.github.io/smithy/1.0/spec/aws/aws-json-1_0-protocol.html#operation-error-serialization
+     * See: https://awslabs.github.io/smithy/1.0/spec/aws/aws-json-1_0-protocol.html#differences-between-awsjson1-0-and-awsjson1-1
      */
-    private fun serverStructureSerializer(
+    fun serverStructureSerializer(
```
rcoh, line 162: rather than hard coding this, I'd recommend accepting a customization in the serializer generator which we set in the AwsJson serializer. This will enable other customers to do similar sorts of custom behavior for other JSON protocols without hard coding this `__type` pattern. 

I would probably accept the customization as a list argument to JsonSerializer similar to how it works in `serviceGenerator`

rcoh: I'm happy to pair on this if you need more guidance

rcoh (commented):
I can let this go for now as something the server team can own, but for general future flexibility, I'd strongly recommend going with the customization approach instead of hard coding `__type` into the JSON generator

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
rcoh: another option could be a refactor like I started in here: https://github.com/awslabs/smithy-rs/compare/neutrino?expand=1 which would enable writing an `AwsJsonSerializer` and a `RestJsonSerializer` as separate classes without much duplication

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -166,6 +175,10 @@ class JsonSerializerGenerator(
                 rust("let mut out = String::new();")
                 rustTemplate("let mut object = #{JsonObjectWriter}::new(&mut out);", *codegenScope)
                 serializeStructure(StructContext("object", "value", structureShape), includedMembers)
+                if (includeErrorType && structureShape.hasTrait<ErrorTrait>()) {
+                    val typeId = if (useErrorNamespace) { structureShape.getId() } else { structureShape.id.name }
+                    writeWithNoFormatting("""object.key("__type").string("$typeId");""")
```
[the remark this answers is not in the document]
crisidev: Sure, will do.

rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
```diff
@@ -118,50 +111,102 @@ impl From<&PathSpec> for Regex {
...
+        Self { x_amzn_target }
+    }
+}
+
+impl RequestSpec for AwsJsonRequestSpec {
+    fn matches<B>(&self, req: &Request<B>) -> Match {
+        // We first check for the method and URI of the current HTTP request.
+        // Every request for the AwsJson 1.0 and 1.1 protocol MUST be sent to the root URL (/)
+        // using the HTTP "POST" method.
+        if req.method() != http::Method::POST {
```
david-perez, line 148: `MethodNotAllowed` should be sent when the resource (i.e. URI) is correct, but the method is not, so this check should go below checking for `/`.

Add a unit test to not regress on this.

rust-runtime/aws-smithy-http-server/src/routing/mod.rs
```diff
@@ -161,38 +175,46 @@ where
...
     use crate::{body::boxed, routing::request_spec::*};
     use futures_util::Future;
-    use http::Method;
+    use http::{HeaderMap, Method};
     use std::pin::Pin;
 
     /// Helper function to build a `Request`. Used in other test modules.
-    pub fn req(method: &Method, uri: &str) -> Request<()> {
-        Request::builder().method(method).uri(uri).body(()).unwrap()
+    pub fn req(method: &Method, uri: &str, headers: Option<HeaderMap>) -> Request<()> {
```
david-perez, line 205: Use the request builder to add the headers instead of building the request and then modifying its headers.

rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
```diff
@@ -404,42 +464,85 @@ mod tests {
...
+
+    fn aws_json_spec() -> AwsJsonRequestSpec {
+        AwsJsonRequestSpec::new(String::from("Service.operation"))
+    }
+
+    #[test]
+    fn aws_json_matches() {
+        let reqs = vec![
+            (
+                &Method::POST,
```
david-perez, line 514: No need for taking references here.

rust-runtime/aws-smithy-http-server/src/runtime_error.rs
```diff
@@ -25,7 +25,8 @@ use crate::protocols::Protocol;
 
 #[derive(Debug)]
 pub enum RuntimeErrorKind {
-    // UnknownOperation,
+    /// The requested operation does not exist.
+    UnknownOperation,
```
david-perez, line 29: Is there a unit test in the router checking that this gets hit when the operation does not exist?

david-perez (commented):
> Protocol tests for these 2 protocols are heavily biased towards Responses. Before announcing support for awsJson 1.0 and 1.1, we should increase the protocol tests coverage.

I agree with the sentiment but with restXml, because it uses a different serialization format.

I expect though that the differences in coverage between restJson1 and awsJson10/11 would be minimal. The bulk of the request tests test deserialization, and JSON deser is identical (?)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -90,21 +90,21 @@ class ServerProtocolTestGenerator(
         abstract val protocol: ShapeId
         abstract val testType: TestType
 
-        data class RequestTest(val testCase: HttpRequestTestCase): TestCase() {
+        data class RequestTest(val testCase: HttpRequestTestCase) : TestCase() {
```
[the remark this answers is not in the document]
crisidev: You really cannot do that :) I am running pre-commit and the code formatter runs before the commit and if you don't run pre-commit, the CI will fail. This is a little chicken-egg problem right now.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerAwsJson.kt
```diff
@@ -0,0 +1,142 @@
...
+            fnName,
+            errorShape,
+            includedMembers,
+            true,
+            awsJsonVersion == AwsJsonVersion.Json10
+        )
+    }
+}
+
+class ServerAwsJson(
```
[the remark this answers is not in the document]
crisidev: We can indeed to it. Before doing it I'd like to ask @rcoh if they are ok with us "opening" `AwsJson` class.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -266,7 +279,7 @@ class JsonSerializerGenerator(
 
         val outputShape = operationShape.outputShape(model)
         val fnName = symbolProvider.serializeFunctionName(outputShape)
-        return serverStructureSerializer(fnName, outputShape, httpDocumentMembers)
+        return serverStructureSerializer(fnName, outputShape, httpDocumentMembers, false, false)
     }
 
     override fun serverErrorSerializer(shape: ShapeId): RuntimeType {
```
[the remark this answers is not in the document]
crisidev: I don't have issues with the proposal, but first I'll try to work on the suggestion from @rcoh about accepting a customization in the serializer generator. I'll see what that refactoring will require first.

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -7,21 +7,25 @@
...
 pub enum Protocol {
     RestJson1,
     RestXml,
+    AwsJson10,
+    AwsJson11,
 }
 
-/// Validate that the request had the standard JSON content-type header.
-pub fn check_json_content_type<B>(req: &RequestParts<B>) -> Result<(), RequestRejection> {
+/// TODO: This check should be more strict and codegenerated.
```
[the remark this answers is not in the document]
crisidev: Yes, I agree we can write a macro. Regarding being more strict, the comment is more general as restJson1 can have different content-types depending on the shapes: https://awslabs.github.io/smithy/1.0/spec/aws/aws-restjson1-protocol.html#content-type

I will update the comment.

rust-runtime/aws-smithy-http-server/src/routing/mod.rs
```diff
@@ -31,56 +35,62 @@ mod route;
...
-
-impl<B> Clone for Router<B> {
-    fn clone(&self) -> Self {
-        Self {
-            routes: self.routes.clone(),
-        }
-    }
+pub struct Router<S, B = Body> {
+    routes: Vec<(Route<B>, S)>,
+    protocol: Protocol,
```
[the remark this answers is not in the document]
crisidev: I understand and agree about the "tightness". 

What I am not sure is the second part. Having a `RequestSpec<Protocol>` will require the `RequestSpec` structure to have all the info for AwsJson (x-amazon-target) and from RestJson / RestXml (method, uri_spec, and uri_path_spec). I think I prefer the complete decoupling I did here.. Thoughts?

rust-runtime/aws-smithy-http-server/src/routing/mod.rs
crisidev: It is functionally identical and inside a test. I prefer it this way to be honest.

rust-runtime/aws-smithy-http-server/src/routing/mod.rs
```diff
@@ -161,38 +175,46 @@ where
...
+    pub fn req(method: &Method, uri: &str, headers: Option<HeaderMap>) -> Request<()> {
+        let mut r = Request::builder().method(method).uri(uri).body(()).unwrap();
+        if let Some(headers) = headers {
+            *r.headers_mut() = headers
+        }
+        r
     }
 
-    /// A service that returns its name and the request's URI in the response body.
+    /// A REST service that returns its name and the request's URI in the response body.
```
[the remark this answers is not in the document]
crisidev: Sure, I'll remove.

rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
```diff
@@ -76,15 +76,8 @@ impl UriSpec {
...
-#[derive(Debug, Clone)]
-pub struct RequestSpec {
-    method: http::Method,
-    uri_spec: UriSpec,
-    uri_path_regex: Regex,
-}
-
 #[derive(Debug, PartialEq)]
-pub(super) enum Match {
+pub enum Match {
```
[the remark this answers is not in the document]
crisidev: Since the `Router` is `pub`, both the trait `RequestSpec` and `Match` have to be public as you can't leak crate-private trait.

rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
crisidev: Good catch!

rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
crisidev: Will fix.

rust-runtime/aws-smithy-http-server/src/runtime_error.rs
crisidev: Nope, there isn't. I'll add one.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
crisidev: I'll probably have some questions once I try out the first solution. Thanks for the suggestion.

rust-runtime/aws-smithy-http-server/src/routing/mod.rs
crisidev: Also I don't understand how to do what you are suggesting. [Builder](https://docs.rs/http/0.2.6/http/request/struct.Builder.html) doesn't have a method to set a `HeaderMap`, so I would need to do something like:

```rust
       let r = Request::builder().method(method).uri(uri);
       if let Some(headers) = headers.as_ref() {
           for (k, v) in headers.iter() {
               r.header(k, v);
           }
       }
       r.body(()).unwrap()
```

This doesn't compile for different reasons and looks quite ugly.

Am I missing something?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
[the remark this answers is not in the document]
crisidev: Nice! Using escape I was able to get rid of `writeWithNoFormatting` and just use `rust()`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerAwsJson.kt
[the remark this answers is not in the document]
crisidev: To be honest this saves a lot of code, so I am just going to do it.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -49,11 +51,27 @@ import software.amazon.smithy.rust.codegen.util.hasTrait
...
 import software.amazon.smithy.rust.codegen.util.outputShape
 
+/**
+ * Class describing a JSON section that can be used in a customization.
+ */
+sealed class JsonSection(name: String) : Section(name) {
+    /**
+     * Write the customization just before `object.finish()`.
+     */
+    data class FinalizeObject(val structureShape: StructureShape) : JsonSection("FinalizeObject")
```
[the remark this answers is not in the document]
rcoh: pass the name of the `object` as a string so that implementors don't need to guess / hardcode it (they can inspect the section then use that during codegen)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerAwsJson.kt
```diff
@@ -0,0 +1,96 @@
...
+ */
+class ServerAwsJsonError(private val awsJsonVersion: AwsJsonVersion) : JsonCustomization() {
+    override fun section(section: JsonSection): Writable = when (section) {
+        is JsonSection.FinalizeObject -> writable {
+            if (section.structureShape.hasTrait<ErrorTrait>()) {
+                val typeId = when (awsJsonVersion) {
+                    AwsJsonVersion.Json10 -> section.structureShape.id.toString()
+                    AwsJsonVersion.Json11 -> section.structureShape.id.name.toString()
+                }
+                rust("""object.key("__type").string("${escape(typeId)}");""")
```
rcoh, line 70: re: a comment above this would be `"""${section.object}.key("__type")......`

rcoh (commented):
looks great! just one small suggestion

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -55,10 +55,8 @@ import software.amazon.smithy.rust.codegen.util.outputShape
  * Class describing a JSON section that can be used in a customization.
  */
 sealed class JsonSection(name: String) : Section(name) {
-    /**
-     * Write the customization just before `object.finish()`.
-     */
-    data class FinalizeObject(val structureShape: StructureShape) : JsonSection("FinalizeObject")
+    /** Write the server error __type customization just before `object.finish()` */
```
rcoh, line 58: this customization is `__type` specific right, that's just an example? Maybe
```
Mutate the server error object prior to finalization. Eg: this can be used to inject `__type` to record the error type.

Use `jsonObject` to refer to the JSON builder.
```

rcoh (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt
crisidev: Yeah, you are right..

rust-runtime/aws-smithy-http-server/src/routing/mod.rs
```diff
@@ -150,42 +313,84 @@ where
...
+                                // Search inside the HashMap for a route for the target.
+                                let route = routes.get(target);
+                                if let Some(route) = route {
+                                    // We have a match, return the Router future.
+                                    return RouterFuture::from_oneshot(route.clone().oneshot(req));
+                                }
+                            }
+                        }
+                    } else {
+                        // The HTTP method is not POST.
```
david-perez, line 357: This will return `405 Method Not Allowed` when we don't know if the target resource (indicated by the `x-amz-target` header) exists.

I also identified another problem according to the HTTP spec: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405

> The server must generate an Allow header field in a 405 status code response. The field must contain a list of methods that the target resource currently supports.

---

Maybe the AWS protocols don't use `Method Not Allowed` at all, it's not in their spec. What does `smithy-typescript` do?

crisidev: AFAIK this is the first language implementing awsJson spec, I was not able to find any reference of awsJson inside the smithy-typescript repo.

rust-runtime/aws-smithy-http-server/src/routing/mod.rs
```diff
@@ -31,34 +36,60 @@ mod route;
...
+/// RestJson1 and RestXml routes are implemented using a `Vec` of handlers because there can be multiple
+/// matches on the URI and thus we need to iterate the whole list and use a ranking mechanism to choose.
+///
+/// AwsJson 1.0 and 1.1 can be implemented using a `HashMap` since operations cannot be ambiguous as they
+/// are all registered on POST /.
+#[derive(Debug)]
+pub(crate) enum Routes<B = Body> {
+    RestXml(Vec<(Route<B>, RequestSpec)>),
+    RestJson1(Vec<(Route<B>, RequestSpec)>),
+    AwsJson10(HashMap<String, Route<B>>),
```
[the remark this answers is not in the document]
crisidev: Done: https://github.com/awslabs/smithy-rs/issues/1326

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationRegistryGenerator.kt
```diff
@@ -241,12 +247,42 @@ class ServerOperationRegistryGenerator(
...
+        when (protocol) {
+            RestJson1Trait.ID, RestXmlTrait.ID -> restRequestSpec()
+            AwsJson1_0Trait.ID, AwsJson1_1Trait.ID -> awsJsonRequestSpec()
+            else -> TODO("Protocol $protocol not supported yet")
+        }
+
+    /*
+     * Returns an AwsJson specific runtime `RequestSpec`.
+     */
+    private fun OperationShape.awsJsonRequestSpec(): String {
```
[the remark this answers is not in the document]
crisidev: Done.

david-perez (approved):

PR #1283, opened 2022-03-28
Simplify CI with composite actions and new image tagging approach

jdisanti:
## Motivation and Context
This PR does three things:
1. Refactors CI to use composite actions so that there's less duplication in the workflow yaml files.
2. Switches to a different image tagging approach for the Docker build image: now the git hash of the `tools/` directory is the image tag rather than the commit hash of `HEAD`. This will result in more image reuse if the tools haven't changed, and makes the logic for deciding whether a new image must be built much simpler.
3. When running CI on main, ensures the Docker build image is built and uploaded to ECR before running the CI tests.


.github/actions/docker-build/action.yml
```diff
@@ -0,0 +1,51 @@
...
+  - name: Prepare build image
+    shell: bash
+    run: |
+      set -x
+      if [[ -d smithy-rs-base-image ]]; then
+        IMAGE_TAG="$(cd smithy-rs; git ls-files -s --full-name "tools" | git hash-object --stdin)"
+        docker load -i smithy-rs-base-image/smithy-rs-base-image
+        docker tag "smithy-rs-base-image:${IMAGE_TAG}" "smithy-rs-base-image:local"
+      fi
+      ALLOW_LOCAL_BUILD=false ./smithy-rs/tools/ci-build/acquire-build-image
```
rcoh, line 38: do you mind adding some comments on this? I'm staring at it, but not really groking the logic. I think it's:

"if `smithy-rs-base-image` exists" (although I don't understand when it would or would not exist": load and tag `:IMAGE_TAG` with `:local`

run "acquire-build-image" but don't allow it to actually run docker build?

.github/actions/docker-build/action.yml
```diff
@@ -0,0 +1,51 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+
+name: smithy-rs Docker Build
```
rcoh, line 4: might be useful to emit the artifact name as an output

.github/workflows/ci-main.yml
```diff
@@ -26,9 +28,12 @@ jobs:
       uses: actions/checkout@v2
     - name: Build image
       run: |
-        IMAGE_TAG="$(git rev-parse HEAD)"
+        IMAGE_TAG="$(git ls-files -s --full-name "tools" | git hash-object --stdin)"
```
rcoh, line 31: should this be a script for unification & to enable adding docs explaining why it exists?

.github/actions/docker-build/action.yml
```diff
@@ -0,0 +1,51 @@
...
+        gradle/wrapper
+      key: ${{ runner.os }}-gradle-${{ hashFiles('gradle/caches/**/*', 'gradle/wrapper/**/*') }}
+      restore-keys: |
+        ${{ runner.os }}-gradle-
+      # Pinned to the commit hash of v1.3.0
+  - uses: Swatinem/rust-cache@842ef286fff290e445b90b4002cc9807c3669641
+    with:
+      sharedKey: ${{ runner.os }}-${{ github.job }}
+      target-dir: ./smithy-rs-target
+  - name: Get artifacts
```
rcoh, line 27: what artifacts will these be?

rcoh (approved):
overall, LGTM but a diagram / README explaining the CI structure would be helpful

.github/actions/docker-build/action.yml
jdisanti: Renamed this to make it clearer.

.github/workflows/ci-main.yml
jdisanti: Moved into a script.

PR #1285, opened 2022-03-30
Update pinned nightly to fix `cargo udeps` checks

jdisanti:
## Motivation and Context
CI started failing due to the release of `lock_api-0.4.7` (see [failed build here](https://github.com/awslabs/smithy-rs/runs/5759821814)), which does not appear to compile against `nightly-2022-03-03`. This PR upgrades the nightly to a version that it successfully compiles against.


rust-runtime/aws-smithy-types/additional-ci
```diff
@@ -12,4 +12,4 @@ echo "### Checking for duplicate dependency versions in the normal dependency gr
 cargo tree -d --edges normal --all-features
 
 echo "### Running api-linter"
-cargo +nightly-2022-03-03 api-linter --all-features
+cargo "+${RUST_NIGHTLY_VERSION}" api-linter --all-features
```
rcoh, line 15: Where does this var come from?

rcoh (approved):

rust-runtime/aws-smithy-types/additional-ci
jdisanti: It's an environment variable set in the build Docker image here: https://github.com/awslabs/smithy-rs/blob/main/tools/ci-build/base-image.dockerfile#L117

It just occurred to me that using it here will mean we can't directly invoke these additional-ci scripts anymore... What's your preference on having more locations to update vs. having to jump through hoops to run one of these test scripts? I'm leaning towards having more places to update until it's easier to run the Docker CI actions.

rcoh: I assumed we'd be using docker to run the actions locally. You can also define a default in this file with something like:
`RUST_NIGHTLY_VERSION=${blah:-$RUST_NIGHTLY_VERSION}`

rcoh (approved):

tools/Dockerfile
```diff
@@ -36,10 +39,12 @@ RUN set -eux; \
...
 FROM bare_base_image AS install_rust
 ARG rust_stable_version=1.56.1
-ARG rust_nightly_version=nightly-2022-03-03
+ARG rust_nightly_version=nightly-2022-03-29
 ARG cargo_udeps_version=0.1.27
 ARG cargo_hack_version=0.5.12
-ARG smithy_rs_revision=main
+# If the `checkout_smithy_rs_tools` arg is set to true, then the Dockerfile will acquire the tools
+# source code by checking out awslabs/smithy-rs/main rather than copying them from the local directory.
+ARG checkout_smithy_rs_tools=false
```
rcoh, line 47: do we even need this arg 🤔 

We could have the other docker build handle the repo clone separately from building...

rcoh, line 47: do we even need this arg 🤔 

We could have the other docker build handle the repo clone separately from building...

rcoh (approved):

PR #1294, opened 2022-04-01
Switch to `base64` crate

### LINKED ISSUE #1277: Incorrect base64 implementation?
Consider this patch:

```
diff --git a/rust-runtime/aws-smithy-types/src/base64.rs b/rust-runtime/aws-smithy-types/src/base64.rs
index 846c7eb1..5c49b708 100644
--- a/rust-runtime/aws-smithy-types/src/base64.rs
+++ b/rust-runtime/aws-smithy-types/src/base64.rs
@@ -196,6 +196,19 @@ mod test {
         assert_eq!(encode("anything you wan"), "YW55dGhpbmcgeW91IHdhbg==");
     }

+    #[test]
+    fn test_base64_misc() {
+        let us_1 = decode("xyz");
+        let them_1 = base64::decode("xyz");
+        dbg!(&us_1);
+        dbg!(&them_1);
+
+        let us_2 = decode("YmxvYg=");
+        let them_2 = base64::decode("YmxvYg=");
+        dbg!(&us_2);
+        dbg!(&them_2);
+    }
+
     #[test]
     fn test_invalid_padding() {
         // no internal padding
```

`cargo test test_base64_misc -- --nocapture` yields:

```
running 1 test
[aws-smithy-types/src/base64.rs:203] &us_1 = Ok(
    [
        199,
        44,
    ],
)
[aws-smithy-types/src/base64.rs:204] &them_1 = Err(
    InvalidLastSymbol(
        2,
        122,
    ),
)
[aws-smithy-types/src/base64.rs:208] &us_2 = Ok(
    [
        98,
        108,
        111,
        98,
    ],
)
[aws-smithy-types/src/base64.rs:209] &them_2 = Ok(
    [
        98,
        108,
        111,
        98,
    ],
)
test base64::test::test_base64_misc ... ok
```

So the first case seems to be invalid but our implementation is accepting it. The second case is _technically_ not valid base64 (note code point length is not divisible by 4), since it's missing an extra padding `=` character, but I'm guessing both implementations can unambiguously decode it. However, other implementations are stricter:

```
>> atob("YmxvYg=");
Uncaught DOMException: String contains an invalid character
```

These two cases are deemed as invalid base64 [according to Smithy protocol tests](https://github.com/awslabs/smithy/blob/887ae4f6d118e55937105583a07deb90d8fabe1c/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-blob.smithy#L37-L37).

david-perez:
This removes our implementation of base64 encoding/decoding from
`aws-smithy-types`, switching to the `base64` Rust crate.

Our implementation was written at a time when we wanted to keep
dependencies at an absolute minimum. However, a performant base64
implementation is now needed for the server.

Our implementation was also accepting invalid base64 (see #1277).

Fixes #1277.

## Checklist


CHANGELOG.next.toml
```diff
@@ -11,6 +11,12 @@
 # meta = { "breaking" = false, "tada" = false, "bug" = false }
 # author = "rcoh"
 
+[[smithy-rs]]
+message = "Switch to `base64` crate"
+references = ["smithy-rs#1294"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
```
david-perez, line 17: Where do I add this message?

> The base64 implementation from `aws-smithy-types` has been removed. Please switch to another implementation (e.g. the [`base64`](https://crates.io/crates/base64) crate).

I wanted to add it to `aws/SDK_CHANGELOG.md`, but that file is autogenerated from `update-changelog` from `tools/sdk-lints`. How are the migration guides from that file added if the file is autogenerated? Do they get added manually just before preparing a release?

rcoh: if you see the comments above, you use `[[aws-sdk-rust]]` instead of `smithy-rs`.

Anything you want in the message goes in the message field—TOML allows multiline strings. **However**: I think we should keep the same function signatures and just delegate to the base64 crate behind the scenes. Having a "blessed" base64 implementation inside of `aws-smithy-types` is probably generally a good thing to have.

rcoh (approved):

CHANGELOG.next.toml
david-perez: > There's an example above—[[aws-sdk-rust]] instead of smithy-rs.

But how do I add a longer migration guide e.g. https://github.com/awslabs/smithy-rs/blob/775d227c5fbc46c2e3c2878b270b8e8226ab7ef1/aws/SDK_CHANGELOG.md?plain=1#L5-L33 ? Do I plop all of that Markdown into `message`?

---

>  However: I think we should keep the same function signatures and just delegate to the base64 crate behind the scenes. Having a "blessed" base64 implementation inside of aws-smithy-types is probably generally a good thing to have.

I thought a bit about this before making the change and concluded that there's no benefit. base64 is standard and I can't foresee us or Smithy wanting to customize anything that is not already customizable via the `base64` crate. I also don't think that SDK users should have public access to a base64-encoding routine from a runtime crate; they should have no need for it after all, since they should just be using `aws_smithy_types::Blob`. I think having a "blessed" base64 implementation is unnecessarily increasing public API surface.

rcoh: This makes sense!

rcoh: Maybe blob should add a from_base64 method...

Velfi: yeah, just use triple quote strings for long messages

david-perez: Ok, I've added it here https://github.com/awslabs/smithy-rs/pull/1294/commits/f1c00cbd41c05c5be8dab3bc5392c4583a6ae35d

david-perez:
**Let's hold off merging this until we know [Smithy's firm stance on extant base64-encoded data](https://github.com/awslabs/smithy/pull/1168)**.

rcoh:
@david-perez https://github.com/awslabs/smithy/pull/1168 was merged—what's the plan here?

rcoh:
We also just fix our impl for now

Nugine:
Hi. The [base64-simd](https://docs.rs/base64-simd) crate may be what you want.

david-perez:
Superseded by https://github.com/awslabs/smithy-rs/pull/1938.

PR #1296, opened 2022-04-01
Improve AssumeRoleProvider

rcoh:
## Motivation and Context
AssumeRoleProvider is used directly by a number of customers, but it isn't cached and is missing some key features.
- aws-sdk-rust#479

## Description

This adds to improvements to `AssumeRoleProvider`:
1. This adds a cache aroudn this provider. Since this provider is intended to be used directly by customers,
it should be cached to improve performance and predictability.

2. Add support for configuring the role session length.

## Testing
- [x] Adds UT

## Checklist

### DIFF
--- a/aws/rust-runtime/aws-config/src/sts/assume_role.rs
+++ b/aws/rust-runtime/aws-config/src/sts/assume_role.rs
@@ -8,7 +8,6 @@
 use aws_sdk_sts::error::AssumeRoleErrorKind;
 use aws_sdk_sts::middleware::DefaultMiddleware;
 use aws_sdk_sts::operation::AssumeRole;
-use aws_smithy_async::rt::sleep::default_async_sleep;
 use aws_smithy_client::erase::DynConnector;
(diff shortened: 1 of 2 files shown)


aws/rust-runtime/aws-config/src/sts/assume_role.rs
```diff
@@ -154,24 +188,24 @@ impl AssumeRoleProviderBuilder {
             .set_role_arn(Some(self.role_arn))
             .set_external_id(self.external_id)
             .set_role_session_name(Some(session_name))
+            .set_duration_seconds(self.session_length.map(|dur| dur.as_secs() as i32))
```
Velfi, line 191: it'd be kinda neat if we had customizations that provided `into` impls for these "duration-like" fields. I think we'd have to do it manually though

Velfi (approved):

jdisanti (approved):
LGTM!

PR #1298, opened 2022-04-01
Split codegen tests out into their own job

jdisanti:
## Motivation and Context
This should give faster feedback on codegen tests since they will no longer need to wait for the SDK to be generated in order to run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,6 @@ jobs:
       matrix:
         # These correspond to scripts in tools/ci-build/scripts that will be run in the Docker build image
         actions:
-        - action: check-style-and-lints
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -66,8 +65,35 @@ jobs:
       with:
         action: ${{ matrix.actions.action }}
 
-  test:
-    name: Test
+  test-codegen:
```
rcoh, line 68: Can we add a longer form description here to explain the grouping? (Test of white label smithy codegen behavior, no aws sdk dependency, etc.)

We may be able to eventually go one step further and not run these at all for changes in the aws directory

rcoh (approved):
LGTM!

PR #1300, opened 2022-04-02
Add endpoint resolver to SdkConfig

benesch:
## Motivation and Context

This change fixes issue aws-labs/aws-sdk-rust#396.

## Description

This commit adds an `endpoint_resolver` field to SdkConfig, which flows
through to the `endpoint_resolver` field on each individual service
configuration. To avoid a circular dependency between the aws-types
crate and the aws-endpoint crate, the latter is merged into the former.

**Note to reviewer:** I tried very hard to avoid merging `aws-endpoint` into `aws-types`, but I just didn't see another way. Not sure if this is acceptable.

## Testing

Letting CI handle it.

## Checklist


rcoh:
Woo! Overall looks great. Instead of deleting aws-endpoint, we should add a constructor to AwsEndoint so it can be built outside of the crate, then move AwsEndpoint and ResolveAwsEndpoint into aws-config while keeping the partition logic & middleware in aws-endpoint. (But there could definitely be a reason that won't work I'm not seeing)

benesch:
> Woo! Overall looks great. Instead of deleting aws-endpoint, we should add a constructor to AwsEndoint so it can be built outside of the crate, then move AwsEndpoint and ResolveAwsEndpoint into aws-types while keeping the partition logic & middleware in aws-endpoint. (But there could definitely be a reason that won't work I'm not seeing)

Ah yeah, I went down that road originally but the `partition` module wants to be able to access the private `endpoint` and `credential_scope` fields of `AwsEndpoint`, which it loses access to if it moves to a separate crate. Ok to make those fields public?

benesch:
I added public accessor methods for the necessary fields. This is a much more targeted change now, and is mergeable from my perspective as soon as tests pass!

rcoh (commented):
This looks great! I'm going to give it a close review this afternoon but very excited about this!

rcoh:
to save some spelunking, this is where we need to add `Region::from_static`:
https://github.com/awslabs/smithy-rs/blob/9f1ef3d4087db5bdd0f84fb283ca8a1e5845265c/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt#L343-L345

benesch:
> This looks great! I'm going to give it a close review this afternoon but very excited about this!

🎉 thanks very much, @rcoh! Please feel free to just push at my branch, or open a new PR with changes, if easier.

benesch:
I pushed up a fix for the thing you noticed and added a changelog entry.

benesch:
Looks like pretty much everything is passing. There are a few authentication failures that I assume I can ignore, and then one failure about this example

https://github.com/awsdocs/aws-doc-sdk-examples/blob/94d63a982bdf197b17eda60e7c85b45b27f69df7/rust_dev_preview/s3/src/bin/s3-object-lambda.rs#L101

which now fails to compile due to the breaking change in the `CredentialScope` builder. How do you normally handle that?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
```diff
@@ -90,7 +90,15 @@ class EndpointConfigCustomization(private val codegenContext: CodegenContext, pr
                         self.endpoint_resolver = Some(::std::sync::Arc::new(endpoint_resolver));
                         self
                     }
+
+                    // TODO(docs): include an example of using a static endpoint
```
Velfi, line 94: @rcoh Can you add these docs please?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
```diff
@@ -90,7 +90,15 @@ class EndpointConfigCustomization(private val codegenContext: CodegenContext, pr
...
+                    /// Sets the endpoint resolver to use when making requests.
+                    pub fn set_endpoint_resolver(&mut self, endpoint_resolver: Option<::std::sync::Arc<dyn #T>>) -> &mut Self {
```
Velfi, line 96:
```suggestion
                    pub fn set_endpoint_resolver(&mut self, endpoint_resolver: Option<std::sync::Arc<dyn #T>>) -> &mut Self {
```

rcoh:
added the missing docs, cleaned up some stale docs, ended up refactoring all the tests (😅 ), and used `impl Trait` to keep the old API backwards compatible

rcoh (approved):

PR #1303, opened 2022-04-05
Enable running the canary with a path to a generated SDK

jdisanti:
## Motivation and Context
This makes it possible to run the canary against an unreleased SDK by kicking off the `canary-runner` with a `--sdk-path` argument instead of `--sdk-version`.


rcoh (approved):
LGTM! I think this also enables canaries on PRs if we want to add that

Velfi (approved):

PR #1307, opened 2022-04-06
RFC: Callbacks for `ByteStream` and `SdkBody`

Velfi:
## Motivation and Context

This RFC demonstrates how we could add callback that get triggered when reading chunks from a `ByteStream` or `SdkBody`. The RFC also contains one example of how we could use these callbacks to implement a checksum validator for requests.


design/src/rfcs/rfc0012_bytestream_callback_api.md
```diff
@@ -0,0 +1,112 @@
...
+```rust
+// Each trait method defaults to doing nothing. It's up to implementors to
+// implement one or both of the trait methods
+/// Structs and enums implementing this trait can be inserted into a `ByteStream`,
+/// and will then be called in reaction to various events during a `ByteStream`'s
+/// lifecycle.
+pub trait ByteStreamReadCallback
+// I don't really like this bound but we need it because this will be inserted into an `Inner`
+where
+    Self: Debug + Clone + PartialEq + Eq,
```
jdisanti, line 21: Would we want to include callbacks in the body equality checks? Seems like we should implement a custom equals comparison for ByteStream that excludes the callbacks.

If we `Arc` them in inner instead of `Box`, we can avoid the `Clone` bound.

design/src/rfcs/rfc0012_bytestream_callback_api.md
```diff
@@ -0,0 +1,112 @@
...
+
+    // Read callbacks can only be inserted, not removed or reordered. If users
+    // desire extra management functions, we can add them in a later update.
+    // Callbacks are actually stored and called from the `Inner` object
+    pub fn insert_read_callback(&mut self, callback: Box<dyn ByteStreamReadCallback>) {
+        self.inner.insert_read_callback(callback);
+    }
+
+    // Alternatively, we could add a "builder-style" method for setting callbacks
+    pub fn with_callback(&mut self) -> &mut Self {
```
jdisanti, line 50: Missing the callback arg

design/src/rfcs/rfc0012_bytestream_callback_api.md
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// We add a new method to `ByteStream` for inserting callbacks
+impl ByteStream {
+    // ...other impls omitted
+
+    // Read callbacks can only be inserted, not removed or reordered. If users
+    // desire extra management functions, we can add them in a later update.
+    // Callbacks are actually stored and called from the `Inner` object
+    pub fn insert_read_callback(&mut self, callback: Box<dyn ByteStreamReadCallback>) {
```
jdisanti, line 45: If we don't have remove, then I think the verb shouldn't be insert. The `with_callback` function is probably sufficient.

design/src/rfcs/rfc0012_bytestream_callback_api.md
```diff
@@ -0,0 +1,112 @@
...
+    /// an implementing struct/enum's internal state.
+    // In order to stop the compiler complaining about these empty default impls,
+    // we allow unused variables.
+    fn on_read_chunk(&mut self, #[allow(unused_variables)] chunk: &Bytes) {}
+
+    /// This callback is called once all chunks have been successfully read.
+    /// It's passed a reference to the chunks in the form of an `AggregatedBytes`.
+    /// This function takes `&mut self` so that implementors may modify an
+    /// implementing struct/enum's internal state.
+    fn finally(&mut self, #[allow(unused_variables)] aggregated_bytes: &AggregatedBytes) {}
```
jdisanti, line 35: Should this return something so that checksums can be applied to trailors, or is there another mechanism to make that work?

I don't think this can take a reference to all the data, especially in cases where the data downloaded is too large to store in memory.

design/src/rfcs/rfc0012_bytestream_callback_api.md
```diff
@@ -0,0 +1,112 @@
+RFC: A callback API for `ByteStream`
+====================================
+
+> Status: RFC
+
+Adding a callback API to `ByteStream` will enable developers using the SDK to implement things like checksum validations and 'read progress' callbacks.
```
jdisanti, line 6: The "read progress" use-case is well covered by the doc, but there's little detail about enabling the checksum use-case.

jdisanti (commented):
Great start!

design/src/rfcs/rfc0012_bytestream_callback_api.md
Velfi: What is useful to pass here then? Just the total amount of bytes read? The number of chunks?

design/src/rfcs/rfc0012_bytestream_callback_api.md
Velfi: derp

design/src/rfcs/rfc0012_bytestream_callback_api.md
Velfi: Yeah, I was being lazy mostly. I actually like `Arc` more because I think it's one way we could solve the issue of retrieving state from the callback struct.

design/src/rfcs/rfc0012_bytestream_callback_api.md
Velfi: In the newest design, `finally` can be optionally used to return errors. This would be used in order to return an error if a response body's checksum was requested but it didn't match what was expected. I'm open to a name change, though, if you have suggestions.

design/src/rfcs/rfc0012_bytestream_callback_api.md
Velfi: I added a possible request checksum implementation example in the latest update.

design/src/rfcs/rfc0012_body_callback_apis.md
```diff
@@ -0,0 +1,259 @@
...
+                    // https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.into_iter-2
+                    match (last_header_key_seen, key) {
+                        (_, Some(key)) => {
+                            left_header_map.append(key, value);
+                            last_header_key_seen = Some(key);
+                        }
+                        (Some(key), None) => {
+                            left_header_map.append(key, value);
+                        }
+                        (None, None) => unreachable!(),
```
jdisanti, line 192: I think this is one of those cases where a basic `for` loop is probably easier to follow.

design/src/rfcs/rfc0012_body_callback_apis.md
```diff
@@ -0,0 +1,259 @@
+RFC: Callback APIs for `ByteStream` and `SdkBody`
+=================================================
+
+> Status: RFC
+
+Adding a callback APIs to `ByteStream` and `SdkBody` will enable developers using the SDK to implement things like checksum validations and 'read progress' callbacks.
```
jdisanti, line 6: Are the validations on requests or responses? Is the `ReadCallback` intended for both, or just for requests?

Velfi: It can work for both, it just gets used slightly differently each time. I'll clarify that more.

design/src/rfcs/rfc0012_body_callback_apis.md
```diff
@@ -0,0 +1,353 @@
...
+      match self {
+         Callback::Send(send_callback) => send_callback.make_new(),
+         Callback::Receive(receive_callback) => receive_callback.make_new(),
+      }
+   }
+}
+
+/// A callback that, when inserted into a request body, will be called for corresponding lifecycle events.
+// Docs for these methods will mostly be the same as the docs on `Callback` so I've omitted them.
+trait SendCallback: Send + Sync {
```
rcoh, line 60: does this need to be `Sync`? I guess Sha256 is Send + Sync, so seems like it's fine

rcoh (approved):
LGTM!

design/src/rfcs/rfc0012_body_callback_apis.md
Velfi: I'll remove the bound

design/src/rfcs/rfc0012_body_callback_apis.md
```diff
@@ -0,0 +1,353 @@
...
+   }
+}
+
+/// A callback that, when inserted into a request body, will be called for corresponding lifecycle events.
+// Docs for these methods will mostly be the same as the docs on `Callback` so I've omitted them.
+trait SendCallback: Send {
+   fn update(&mut self, #[allow(unused_variables)] bytes: &[u8]) -> Result<(), BoxError> { Ok(()) }
+   fn headers(
+      &self,
+   ) -> Result<Option<HeaderMap<HeaderValue>>, BoxError> { Ok(None) }
```
jdisanti, line 64: Hmmm... I'm getting confused by the headers/trailers methods. In the case where we're sending a streaming request to S3, we want to send the checksum as a trailer, right? So is this backwards?

What is the use-case for headers?

design/src/rfcs/rfc0012_body_callback_apis.md
```diff
@@ -0,0 +1,353 @@
...
+    fn update(&mut self, bytes: &[u8]) -> Result<(), BoxError> {
+        self.state = match self.state {
+            Some(crc) => { self.state = Some(crc32c_append(crc, bytes)) }
+            None => { Some(crc32c(&bytes)) }
+        };
+
+       Ok(())
+    }
+
+    fn finally(&self) ->
```
jdisanti, line 319: Method name doesn't match the trait

design/src/rfcs/rfc0012_body_callback_apis.md
Velfi: I'll remove headers

design/src/rfcs/rfc0012_body_callback_apis.md
```diff
@@ -0,0 +1,297 @@
...
+trait BodyCallback: Send {
+   /// This lifecycle function is called for each chunk **successfully** read. If an error occurs while reading a chunk,
+   /// this will not be called. This function takes `&mut self` so that implementors may modify an implementing
+   /// struct/enum's internal state. Implementors may return an error.
+   fn update(&mut self, #[allow(unused_variables)] bytes: &[u8]) -> Result<(), BoxError> { Ok(()) }
+
+   /// This callback is called once all chunks have been read. If the callback encountered 1 or more errors
+   /// while running `update`s, this is how those errors are raised. Implementors may return a [`HeaderMap`][HeaderMap]
+   /// that will be appended to the HTTP body as a trailer. This is only useful to do for streaming requests.
+   fn trailers(&self) -> Result<Option<HeaderMap<HeaderValue>>, BoxError> { Ok(None) }
```
Velfi, line 26: I think that for `SdkBody` I can have it inspect itself and only append trailers if the inner variant is `Inner::Streaming` but I'm not sure that covers every streaming request case.

design/src/rfcs/rfc0012_body_callback_apis.md
```diff
@@ -0,0 +1,297 @@
...
+
+## The Implementation
+
+*Note that comments starting with '//' are not necessarily going to be included in the actual implementation and are intended as clarifying comments for the purposes of this RFC.*
+
+```rust
+// in aws_smithy_http::callbacks...
+
+/// A callback that, when inserted into a request body, will be called for corresponding lifecycle events.
+// Docs for these methods will mostly be the same as the docs on `Callback` so I've omitted them.
```
jdisanti, line 16:
```suggestion
```

design/src/rfcs/rfc0012_body_callback_apis.md
```diff
@@ -0,0 +1,297 @@
...
+
+   /// This callback is called once all chunks have been read. If the callback encountered 1 or more errors
+   /// while running `update`s, this is how those errors are raised. Implementors may return a [`HeaderMap`][HeaderMap]
+   /// that will be appended to the HTTP body as a trailer. This is only useful to do for streaming requests.
+   fn trailers(&self) -> Result<Option<HeaderMap<HeaderValue>>, BoxError> { Ok(None) }
+
+   /// Create a new `Callback` from an existing one. This is called when a `Callback` needs to be
+   /// re-initialized with default state. For example: when a request has a body that need to be
+   /// rebuilt, all read callbacks on that body need to be run again but with a fresh internal state.
+   fn make_new(&self) -> Box<dyn SendCallback>;
```
jdisanti, line 31: Recommend taking a pass over the comments and type names. There's some mixing and matching from the previous iterations, such as `Callback` and `SendCallback` still being referenced.

jdisanti (approved):
LGTM!

design/src/rfcs/rfc0012_body_callback_apis.md
Velfi: oops

PR #1311, opened 2022-04-09
Generate a `versions.toml` with the SDK

jdisanti:
## Motivation and Context
This is prerequisite work for tracking SDK crate versions independently, as described in the [RFC](https://github.com/awslabs/smithy-rs/pull/1194). This PR:
- Introduces a `crate-hasher` tool that builds against Cargo and uses the same code paths that find files for the `cargo package` subcommand, and then SHA-256 hashes those files in a deterministic way
- Introduces a `generate-version-manifest` subcommand in the `publisher` tool which examines the generated SDK and runs the `crate-hasher` against it in order to produce a `versions.toml` file
- Incorporates the new `generate-version-manifest` subcommand into the Gradle `aws:sdk:assemble` task so that the generated SDK now has a `versions.toml` manifest


rcoh (commented):
as discussed offline:
- simplify file listing logic

Otherwise, seems good!

.github/actions/docker-build/action.yml
```diff
@@ -9,6 +9,9 @@ inputs:
   action:
     description: What action to run in the Docker build
     required: true
+  action-arguments:
```
rcoh, line 12: maybe just `arguments`? up to you

aws/sdk/build.gradle.kts
```diff
@@ -314,6 +315,32 @@ tasks.register<ExecRustBuildTool>("hydrateReadme") {
...
+    toolPath = publisherToolPath
+    binaryName = "publisher"
+    arguments = listOf(
+        "generate-version-manifest",
+        "--location",
+        outputDir.absolutePath,
+        "--smithy-build",
+        outputDir.resolve("../../smithy-build.json").normalize().absolutePath,
+        "--examples-revision",
+        properties.get("aws.sdk.examples.revision") ?: "missing"
```
rcoh, line 340: will this be confusing when it happens? should we hard fail if this is missing?

tools/Dockerfile
```diff
@@ -73,7 +73,7 @@ RUN set -eux; \
     rustup --version; \
     rustup component add rustfmt; \
     rustup component add clippy; \
-    rustup install ${rust_nightly_version}; \
+    rustup toolchain install ${rust_nightly_version} --component clippy; \
```
rcoh, line 76: side note: we might want to be explicit about the toolchains we install. I've run into some surprises from time to time relying on the defaults (and this might also install components we _don't_ need)

rcoh: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file may also be handy

tools/ci-build/scripts/generate-aws-sdk
```diff
@@ -10,11 +10,12 @@ C_RESET='\033[0m'
 set -eux
 
 echo -e "${C_YELLOW}Taking examples from 'awsdocs/aws-doc-sdk-examples'...${C_RESET}"
+examples_revision=$(cd aws-doc-sdk-examples; git rev-parse HEAD)
```
rcoh, line 13: shellcheck would probably want quotes here?

tools/publisher/src/subcommand/generate_version_manifest.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    let mut crates = BTreeMap::new();
+    for package in packages {
+        // Don't include examples
+        if let PackageCategory::Unknown = package.category {
+            continue;
+        }
+
+        let mut model_hash = None;
+        if package.handle.name.starts_with("aws-sdk-") {
```
rcoh, line 47: how is this different than `PackageCategory::AwsSdk`?

rcoh (approved):

aws/sdk/build.gradle.kts
jdisanti: It's definitely not ideal, but the alternative means not being able to run `./gradlew aws:sdk:assemble` locally without checking out a copy of `aws-doc-sdk-examples`

tools/ci-build/scripts/generate-aws-sdk
jdisanti: Surprisingly not in this case...

tools/publisher/src/subcommand/generate_version_manifest.rs
jdisanti: It isn't. Not sure why I did it that way. Will switch to using the category.

tools/Dockerfile
jdisanti: I think the `--profile minimal` on `rustup-init` is making sure it won't install any undesired components.

PR #1313, opened 2022-04-11
Fix copy-paste errors in generated docs

jonhoo:
I happened across this when trying to implement my own middleware that looked a lot like `AwsMiddleware`. It seems some IDE type hints were copied over along with the code itself, so I removes those and fixed a `s/,/;/`, and the example code works again.

The reason these weren't caused by CI/tests btw is that this code block is marked as ignored, presumably because it'd require a bunch of dependencies to compile.


jdisanti (approved):
Thanks for fixing this! It looks like the CI failure is the intermittent failure I am fixing in https://github.com/awslabs/smithy-rs/pull/1311. I'll retry the failing steps as needed.

jonhoo:
It actually looks like the default stack has changed a bit since this was last written. Should it be updated to something more like

https://github.com/awslabs/smithy-rs/blob/9708aa830730956a5a5a1500c37a6f95fffc9bfb/aws/rust-runtime/aws-inlineable/src/middleware.rs#L21-L33

jdisanti:
Yeah, it should match https://github.com/awslabs/smithy-rs/blob/9708aa830730956a5a5a1500c37a6f95fffc9bfb/aws/rust-runtime/aws-inlineable/src/middleware.rs#L57-L73

jonhoo:
Thinking some more about this, is there a reason the AWS middleware definition is in `aws-inlineable` rather than being a public type in one of the smithy crates? That would certainly make it easier to re-use.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/client/FluentClientDecorator.kt
```diff
@@ -184,37 +184,77 @@ class GenericFluentClient(codegenContext: CodegenContext) : FluentClientCustomiz
...
-                    /// ##[derive(Debug, Default)]
+                    /// /// AWS Middleware Stack
+                    /// ///
+                    /// /// This implements the middleware stack for this service. It will:
+                    /// /// 1. Load credentials asynchronously into the property bag
+                    /// /// 2. Sign the request with SigV4
+                    /// /// 3. Resolve an Endpoint for the request
+                    /// /// 4. Add a user agent to the request
+                    /// #[derive(Debug, Default, Clone)]
+                    /// #[non_exhaustive]
```
jdisanti, line 220: I think CI is failing because these need to be `##` rather than `#`

jonhoo: Ah, good catch!

jonhoo:
> I think there are two reasons the default middleware is currently in `aws-inlineable`:
> 
>     1. Both `aws-http` and `aws-sig-auth` declare middleware layers for the stack, and the generated service crate is the only thing that depends on both of these and can consume all the middleware layers. To have default middleware in the runtime crates would require introducing another crate since `aws-config` is an optional dependency for customers.

Hmm, yeah, I suppose it would need to be its own crate. That said, there are also a fair number of `aws-*` crates, so is having one more actually a problem? It's rare that users will actually take dependencies on these manually, but it _would_ give them the option. There may also be other things from `aws-inlineable` that could be moved to such a crate, I'm not sure?

jdisanti:
I'm not opposed to creating a new crate, but I think it probably deserves a RFC to lay out what it should be named and what should go into it. @rcoh probably has some strong opinions in this area.

rcoh:
it isn't public because it's really a property of an _individual_ service—having a public stack that we claim is "the stack" is a little misleading. There might be a "probably right" stack, but ultimately I expect that as things evolve different services may have different requirements

rcoh:
that said, I'm not opposed to a service builder we expose in `aws-http` or something like that that _does_ wire up the "normal" stuff

PR #1315, opened 2022-04-13
Only retry transient failures for idempotent operations

jdisanti:
## Motivation and Context
Transient failures can only safely be retried for operations modeled as  idempotent or readonly. This PR revises the `make_operation` code generator to conditionally add a marker struct to the operation property bag based on the Smithy `@idempotent` or `@readonly` operation traits, and modifies the Tower retry handler to check for that marker when retrying a transient error.

## Testing
- Unit tested the addition of the marker struct in MakeOperationGeneratorTest
- Unit tested the use of the marker struct in `retry_test`

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -64,3 +64,15 @@ message = "Add endpoint resolver to SdkConfig. This enables overriding the endpo
 references = ["smithy-rs#1300"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "benesch"
+
+[[smithy-rs]]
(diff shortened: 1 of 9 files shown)


jdisanti:
Tests are failing on `aws-config` now because there is a test exercising retry on profile assume role failures, and these are no longer retried since STS's AssumeRole isn't modeled as `@idempotent` or `@readonly`. Need to figure out what the correct behavior for this is.

Velfi (approved):

jdisanti:
From an offline conversation, it may be simpler to do this by replacing the retry policy for the operation with one that is already aware of the model allowing transient retries or not.

PR #1316, opened 2022-04-13
Split `ci.yml` to avoid max workflow depth error

jdisanti:
## Motivation and Context
CI on the `main` branch failed due to the max workflow depth of 2 being exceeded (see https://github.com/awslabs/smithy-rs/actions/runs/2158200204). This PR breaks up the CI workflows to avoid this error.

Successful run of the modified `ci-main.yml` here: https://github.com/awslabs/smithy-rs/actions/runs/2158395230


jdisanti:
Made CI a little bit easier to run locally while waiting to see the result of the build on `main` with these changes. This is incorporated into this PR.

ci.mk
```diff
@@ -0,0 +1,78 @@
+CI_BUILD=./smithy-rs/tools/ci-build
```
rcoh, line 1: add a comment about what this makefile does / what the targets do (if any should be used by a user)

rcoh (approved):

PR #1320, opened 2022-04-13
Release 2022-04-13 (patch)

Velfi:
This updates the MSRV in some CI scripts that I missed in the earlier release. It also adds a release note highlighting the MSRV update.

### DIFF
--- a/aws/SDK_CHANGELOG.md
+++ b/aws/SDK_CHANGELOG.md
@@ -2,6 +2,12 @@
 0.10.0 (April 13th, 2022)
 =========================
 **Breaking Changes:**
+- ⚠ ([smithy-rs#1318](https://github.com/awslabs/smithy-rs/issues/1318)) We updated the MSRV from 1.56.1 to 1.58.1 in the last release but failed to mention it in the changelog.
+
+
+0.10.0 (April 13th, 2022)
+=========================
+**Breaking Changes:**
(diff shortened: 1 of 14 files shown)


aws/SDK_CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
 <!-- Do not manually edit this file, use `update-changelogs` -->
+0.10.0 (April 13th, 2022)
+=========================
+**Breaking Changes:**
+- ⚠ ([smithy-rs#1318](https://github.com/awslabs/smithy-rs/issues/1318)) We updated the MSRV from 1.56.1 to 1.58.1 in the last release but failed to mention it in the changelog.
+
+
```
jdisanti, line 7: We should just merge the two `0.10.0` sections, and also demote the 2021 edition change from breaking changes since it's not a breaking change. We don't need to say we forgot for the SDK since we didn't release the SDK.

jdisanti (approved):

Velfi:
I made a change to our docs codegen that will HTML escape square braces so that rustdoc doesn't interpret them as links to Rust types. I chose to do it with string formatting b/c I couldn't find the right way to do it with jsoup. Here's an example of the generated docs after the change (note that visually it looks right and rustdoc no longer complains in CI)
Screen Shot 2022-04-13 at 3 03 34 PM

jdisanti (approved):

Velfi:
This PR has turned into something of a debacle, I'm going to revert the MSRV changes in another PR and get the release out.

PR #1327, opened 2022-04-19
Run `ktlint` from repository root

david-perez:
This is the result of running `ktlint` from the repository root. We have
`ktlint` running on pre-commit and CI on changed files, but these
violations slipped through the cracks.


david-perez:
> I'll submit a follow-up PR upgrading ktlint to the latest version.

Provided we can upgrade Kotlin version to 1.6.20.

Velfi (approved):
❤️

PR #1329, opened 2022-04-19
add: callback trait proposed in rfc#0012

Velfi:
## Motivation and Context

Implements the [Body Callback RFC](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0012_body_callback_apis.md)

## Description

add: callback setting API to ByteStream
add: callback setting API to SdkBody
add: callback tests for streaming and buffered data
add: private callback fns for calculating checksums
udpate: pub methods on private struct `Inner` to be private
update: changelog
remove: Sync bound from SdkBody test "sdk_body_is_send()"
add: fn to merge `HeaderMap`s by appending
add: tests for new functionality

## Testing

ran existing tests and created new ones

## Checklist


Velfi:
I'm putting this up for initial review. It has two TODOs that I could use some advising on. Currently, the checksum callbacks are private because I don't want people to start messing with them. This PR, once merged, will be followed by another PR that implements the codegen side of things and makes the callbacks public.

rust-runtime/aws-smithy-http/Cargo.toml
```diff
@@ -29,16 +32,30 @@ hyper = "0.14"
...
 futures-core = "0.3.14"
 tokio = { version = "1.6", optional = true }
-tokio-util = { version = "0.6", optional = true}
+tokio-util = { version = "0.6", optional = true }
+
+# Checksum algos
+crc32fast = "1.3"
+crc32c = "0.6"
+sha1 = "0.10"
+sha2 = "0.10"
```
rcoh, line 41: I think ideally we'd keep this separate so we didn't compile all these deps when they weren't needed

rust-runtime/aws-smithy-http/src/callback.rs
```diff
@@ -0,0 +1,172 @@
...
+use http::{HeaderMap, HeaderValue};
+
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+/// A callback that, when inserted into a request body, will be called for corresponding lifecycle events.
+pub trait BodyCallback: Send {
+    /// This lifecycle function is called for each chunk **successfully** read. If an error occurs while reading a chunk,
+    /// this method will not be called. This method takes `&mut self` so that implementors may modify an implementing
+    /// struct/enum's internal state. Implementors may return an error.
+    fn update(&mut self, #[allow(unused_variables)] bytes: &[u8]) -> Result<(), BoxError> {
```
rcoh, line 19: just for visual noise, might be nicer to add `let _ = bytes;` in the body instead of the unused variables allow

rust-runtime/aws-smithy-http/src/callback/checksum.rs
```diff
@@ -0,0 +1,260 @@
+/*
```
rcoh, line 1: we should probably keep these in another crate we can pull in when checksums are actually used

rust-runtime/aws-smithy-http/src/callback.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        let body = ByteStream::from(body).collect().await.unwrap().into_bytes();
+        let body = std::str::from_utf8(&body).unwrap();
+
+        // Assert that the body that went in is the same as the body coming out.
+        assert_eq!(body, "test");
+
+        assert!(logs_contain("callback A was called"));
+        assert!(logs_contain("callback B was called"));
+        assert!(!logs_contain("callback was called"));
```
jdisanti, line 98: What would have logged this?

rust-runtime/aws-smithy-http/src/header.rs
```diff
@@ -295,16 +294,75 @@ pub fn quote_header_value<'a>(value: impl Into<Cow<'a, str>>) -> Cow<'a, str> {
...
+///     append_merge_header_maps(left_hand_side_headers, right_hand_side_headers);
+/// let merged_values: Vec<_> = merged_header_map
+///     .get_all(header_name.clone())
+///     .into_iter()
+///     .collect();
+///
+/// // Will print 'some_key: ["lhs value", "rhs value"]'
+/// println!("{}: {:?}", header_name.as_str(), merged_values);
+/// ```
+pub fn append_merge_header_maps(
```
jdisanti, line 330: This should be `pub(crate)`

rust-runtime/aws-smithy-http/src/header.rs
```diff
@@ -560,4 +618,28 @@ mod test {
         assert_eq!("\"(\"", &quote_header_value("("));
         assert_eq!("\")\"", &quote_header_value(")"));
     }
+
+    #[test]
+    fn test_append_merge_header_maps() {
```
jdisanti, line 623: There should also be a test for multiple header values with the same header name in the same header map. Probably also a good idea to have tests with empty header maps.

rust-runtime/aws-smithy-checksum-callbacks/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "aws-smithy-checksum-callbacks"
```
jdisanti, line 2: Maybe just `aws-smithy-checksums`? The callback part seems like an implementation detail.

rust-runtime/aws-smithy-http/src/callback.rs
```diff
@@ -0,0 +1,172 @@
...
+            .with_callback(Box::new(CallbackB));
+
+        let body = ByteStream::from(body).collect().await.unwrap().into_bytes();
+        let body = std::str::from_utf8(&body).unwrap();
+
+        // Assert that the body that went in is the same as the body coming out.
+        assert_eq!(body, "test");
+
+        assert!(logs_contain("callback A was called"));
+        assert!(logs_contain("callback B was called"));
```
jdisanti, line 97: Do we care what order these were called in? I don't think the log testing will inform on that.

jdisanti (commented):
Looks great! I just have some minor comments and bike shedding.

rust-runtime/aws-smithy-http/src/callback.rs
Velfi: I think no. I'm uncomfortable with users depending on callbacks running in a specific order because it just seems like that would be a source of bugs/confusion. This test exists to show that all callbacks get called when setting more than one.

rust-runtime/aws-smithy-http/src/callback.rs
Velfi: nothing, and I can't recall why I added it. I'll remove it.

rust-runtime/aws-smithy-checksums/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+        header_map.insert(key, value);
+
+        Ok(Some(header_map))
+    }
+}
+
+impl BodyCallback for Crc32callback {
+    fn update(&mut self, bytes: &[u8]) -> Result<(), BoxError> {
+        self.update(bytes)
+    }
```
jdisanti, line 51: On all of these: why implement `update` and `trailers` in a separate impl block?

jdisanti (approved):
LGTM!

rust-runtime/aws-smithy-checksums/src/lib.rs
Velfi: The original idea would be that it allows you to call these functions without needing to use the trait. This would happen within the "mutate request" section for non-streaming requests. For streaming requests they'd all be boxed trait objects.

I'm fine with making them only usable behind the trait if you like. Also, I don't think I've mentioned that these callbacks are going to be marked `pub` when I do the codegen half of things. I made them private to start so people wouldn't touch them.

PR #1331, opened 2022-04-20
Upgrade Kotlin, Kotest, `ktlint` and Jacoco

david-perez:
This commit upgrades:

* Kotlin to 1.6.20,
* Kotest to 5.2.3,
* Ktlint to 0.45.2; and
* Jacoco to 0.8.8.

Perhaps the only noteworthy change is that Ktlint now has a lint to
detect that the name of a class/object defined solely in a file has to
coincide with the filename.


buildSrc/src/main/kotlin/PropertyRetriever.kt
```diff
@@ -5,7 +5,7 @@
 
 import org.gradle.api.Project
 import java.io.File
-import java.util.Properties
+import java.util.*
```
crisidev, line 8: This doesn't look right.. Shouldn't we just import what we need instead of *?

crisidev (approved):
LGTM, just a question about import *.

buildSrc/src/main/kotlin/PropertyRetriever.kt
david-perez: Fixed the regression in https://github.com/awslabs/smithy-rs/pull/1331/commits/73b34d0ec0c0c4c2aaf0e160c5fd03ab9567f404.

jdisanti (approved):
This is awesome! Thank you!

rcoh (approved):

PR #1333, opened 2022-04-20
Make server pass naming obstacle course protocol tests

david-perez:
This commit fixes two bugs that were present in the server
implementation:

* We were incorrectly referring to the operation input shape name
  instead of using the symbol provider in
  `ServerOperationHandlerGenerator.kt` and
  `ServerOperationRegistryGenerator.kt`.
* We were not escaping Rust reserved keywords in operation names in
  `ServerOperationRegistryGenerator.kt`.

The second bug would have been caught had we been generating server
crates for the two naming obstacle course test suites
(`naming-obstacle-course-ops.smithy` and
`naming-obstacle-course-structs.smithy`) that the client is already
passing. This commit adds these suites to the `codegen-server-test`
subproject, and expands `naming-obstacle-course-ops.smithy` to exercise
the logic that was causing the first bug.

## Testing

Verified that the tests fail if the bugfixes are reverted.

---


crisidev (approved):
Very good!

Velfi (approved):

crisidev (approved):

PR #1336, opened 2022-04-22
Implement new `sdk-sync` tool

jdisanti:
## Motivation and Context
This tool will eventually replace the `smithy-rs-sync` tool and be used in release automation that includes automated model updates. These are the notable differences from `smithy-rs-sync`:
- Operates directly on `aws-sdk-rust/main` rather than on `next`.
- Interactions with other processes have been abstracted behind traits so they can be mocked in tests.
- The previous smithy-rs revision is loaded from `versions.toml` rather than from `.smithyrs-githash`.
- Correct attribution is maintained for example syncing.
- Commits in smithy-rs that produce no SDK changes are no longer included in the sync.
- Model changes are now expected to already be commited to smithy-rs before the tool runs, and are properly incorporated.

## Testing
- Unit tested individual git interactions.
- Thoroughly unit tested with mocked process interactions.
- Created an E2E test that creates fake `aws-sdk-rust`, `smithy-rs`, and `aws-doc-sdk-examples` repositories with test data to verify function against actual git repositories.
- Manually tested against the actual  `aws-sdk-rust`, `smithy-rs`, and `aws-doc-sdk-examples` repositories. Examined the synced commits and verified the model commit is retained in `smithy-rs`.


tools/sdk-sync/src/sync.rs
```diff
@@ -0,0 +1,685 @@
...
+
+        self.build_and_copy_sdk(&examples_head).context(here!())?;
+        self.aws_sdk_rust
+            .stage(&PathBuf::from("."))
+            .context(here!())?;
+
+        let example_commits: Vec<Commit> = example_revisions
+            .iter()
+            .map(|commit_hash| self.aws_doc_sdk_examples.show(commit_hash.as_ref()))
+            .collect::<Result<_, _>>()?;
```
Velfi, line 243: I've never seen both the type annotation AND a turbofish but it totally makes sense in this case.

Velfi (approved):

PR #1341, opened 2022-04-25
include link to RFC-12

rcoh:
## Motivation and Context
a summary link is required for files to show up in design docs


jdisanti:
It looks like we have two RFC-12s...

PR #1342, opened 2022-04-25
Builders of builders

### LINKED ISSUE #1714: Server structures do not honor `required` members if their symbol is marked with `Default`
This is the case for all simple shapes.

https://github.com/awslabs/smithy-rs/blob/e3239e1a17ea0165849ce5379ae0fe40bdb16577/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt#L236

For example, consider this model:

```smithy
$version: "1.0"

use aws.protocols#restJson1

@restJson1
service SimpleService {
    operations: [
        AnOperation,
    ],
}

@http(uri: "/operation", method: "GET")
operation AnOperation {
    input: AnOperationInputOutput,
    output: AnOperationInputOutput,
}

structure AnOperationInputOutput {
    @required
    string: String,

    @required
    int: Integer
}
```

This generates the following structure:

```rust
pub struct AnOperationInput {
    #[doc(hidden)]
    pub string: std::string::String,
    #[doc(hidden)]
    pub int: i32,
}
```

And its builder's `build()` implementation is:

```rust
pub fn build(
    self,
) -> Result<crate::input::AnOperationInput, aws_smithy_http::operation::BuildError>
{
    Ok(crate::input::AnOperationInput {
        string: self.string.unwrap_or_default(),
        int: self.int.unwrap_or_default(),
    })
}
```

Notice the `unwrap_or_default()`s! This entails that if a client sends an empty `AnOperationInput` on the wire, the server will gladly accept it and default `string` to `""` and `int` to `0`, which is obviously incorrect behavior.

This is a result of the server reusing the client's structure builder code. In the client, all members are `Option`al in IDL v1, regardless of `required`, and so `unwrap_or_default()` sets them to `None` if the client receives a response without these members set. That is clearly correct.

I'm surprised we never caught this huge violation of the spec. I guess it's because in IDL v1, `required` is a constraint trait, and we've been ignoring the entire validation test suite because constraint traits are not yet implemented. Although now that I take a look, there doesn't seem to be much coverage for `required` member shapes anyway; these are the only three tests with `Required` in their name that we're ignoring:

https://github.com/awslabs/smithy-rs/blob/e3239e1a17ea0165849ce5379ae0fe40bdb16577/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt#L751-L753

david-perez:
This patchset, affectionately called "Builders of builders", lays the
groundwork for fully implementing [Constraint traits] in the server SDK
generator. [The RFC] illustrates what the end goal looks like, and is
recommended prerrequisite reading to understanding this cover letter.

This commit makes the sever deserializers work with _unconstrained_ types
during request parsing, and only after the entire request is parsed are
constraints enforced. Values for a constrained shape are stored in the
correspondingly unconstrained shape, and right before the operation input is
built, the values are constrained via a `TryFrom<UnconstrainedShape> for
ConstrainedShape` implementation that all unconstrained types enjoy. The
service owner only interacts with constrained types, the unconstrained ones are
`pub(crate)` and for use by the framework only.

In the case of structure shapes, the corresponding unconstrained shape is their
builders. This is what gives this commit its title: during request
deserialization, arbitrarily nested structures are parsed into _builders that
hold builders_. Builders keep track of whether their members are constrained or
not by storing its members in a `MaybeConstrained`
[Cow](https://doc.rust-lang.org/std/borrow/enum.Cow.html)-like `enum` type:

```rust
pub(crate) trait Constrained {
    type Unconstrained;
}

#[derive(Debug, Clone)]
pub(crate) enum MaybeConstrained<T: Constrained> {
    Constrained(T),
    Unconstrained(T::Unconstrained),
}
```

Consult the documentation for the generator in `ServerBuilderGenerator.kt` for
more implementation details and for the differences with the builder types the
server has been using, generated by `BuilderGenerator.kt`, which after this
commit are exclusively used by clients.

Other shape types, when they are constrained, get generated with their
correspondingly unconstrained counterparts. Their Rust types are essentially
wrapper newtypes, and similarly enjoy `TryFrom` converters to constrain them.
See the documentation in `UnconstrainedShapeSymbolProvider.kt` for details and
an example.

When constraints are not met, the converters raise _constraint violations_.
These are currently `enum`s holding the _first_ encountered violation.

When a shape is _transitively but not directly_ constrained, newtype wrappers
are also generated to hold the nested constrained values. To illustrate their
need, consider for example a list of `@length` strings. Upon request parsing,
the server deserializers need a way to hold a vector of unconstrained regular
`String`s, and a vector of the constrained newtyped `LengthString`s. The former
requirement is already satisfied by the generated unconstrained types, but for
the latter we need to generate an intermediate constrained
`ListUnconstrained(Vec<LengthString>)` newtype that will eventually be
unwrapped into the `Vec<LengthString>` the user is handed. This is the purpose
of the `PubCrate*` generators: consult the documentation in
`PubCrateConstrainedShapeSymbolProvider.kt`,
`PubCrateConstrainedCollectionGenerator.kt`, and
`PubCrateConstrainedMapGenerator.kt` for more details. As their name implies,
all of these types are `pub(crate)`, and the user never interacts with them.

For users that would not like their application code to make use of constrained
newtypes for their modeled constrained shapes, a `codegenConfig` setting
`publicConstrainedTypes` has been added. They opt out of these by setting it to
`false`, and use the inner types directly: the framework will still enforce
constraints upon request deserialization, but once execution enters an
application handler, the user is on their own to honor (or not) the modeled
constraints. No user interest has been expressed for this feature, but I expect
we will see demand for it. Moreover, it's a good stepping stone for users that
want their services to honor constraints, but are not ready to migrate their
application code to constrained newtypes. As for how it's implemented, several
parts of the codebase inspect the setting and toggle or tweak generators based
on its value. Perhaps the only detail worth mentioning in this commit message
is that the structure shape builder types are generated by a much simpler and
entirely different generator, in
`ServerBuilderGeneratorWithoutPublicConstrainedTypes.kt`. Note that this
builder _does not_ enforce constraints, except for `required` and `enum`, which
are always (and already) baked into the type system. When
`publicConstrainedTypes` is disabled, this is the builder that end users
interact with, while the one that enforces all constraints,
`ServerBuilderGenerator`, is now generated as `pub(crate)` and left for
exclusive use by the deserializers. See the relevant documentation for the
details and differences among the builder types.

As proof that these foundations are sound, this commit also implements the
`length` constraint trait on Smithy map and string shapes. Likewise, the
`required` and `enum` traits, which were already baked in the generated types
as non-`Option`al and `enum` Rust types, respectively, are now also treated
like the rest of constraint traits upon request deserialization. See the
documentation in `ConstrainedMapGenerator.kt` and
`ConstrainedStringGenerator.kt` for details.

The rest of the constraint traits and target shapes are left as an exercise to
the reader, but hopefully the reader has been convinced that all of them can be
enforced within this framework, paving the way for straightforward
implementations. The diff is already large as it is. Any reamining work is
being tracked in #1401; this and other issues are referenced in the code as
TODOs.

So as to not give users the impression that the server SDK plugin _fully_
honors constraints as per the Smithy specification, a validator in
`ValidateUnsupportedConstraintsAreNotUsed.kt` has been added. This traverses
the model and detects yet-unsupported parts of the spec, aborting code
generation and printing informative warnings referencing the relevant tracking
issues. This is a regression in that models that used constraint traits
previously built fine (even though the constraint traits were silently not
being honored), and now they will break. To unblock generation of these models,
this commit adds another `codegenConfig` setting,
`ignoreUnsupportedConstraints`, that users can opt into.

Closes #1714.

Testing
-------

Several Kotlin unit test classes exercising the finer details of the added
generators and symbol providers have been added. However, the best way to test
is to generate server SDKs from models making use of constraint traits. The
biggest assurances come from the newly added `constraints.smithy` model, an
"academic" service that _heavily_ exercises constraint traits. It's a
`restJson1` service that also tests binding of constrained shapes to different
parts of the HTTP message. Deeply nested hierarchies and recursive shapes are
also featured.

```sh
./gradlew -P modules='constraints' codegen-server-test:build
```

This model is _additionally_ generated in CI with the `publicConstrainedTypes`
setting disabled:

```sh
./gradlew -P modules='constraints_without_public_constrained_types' codegen-server-test:build
``````

Similarly, models using currently unsupported constraints are now being
generated with the `ignoreUnsupportedConstraints` setting enabled.

See `codegen-server-test/build.gradle.kts` for more details.

[Constraint traits]: https://awslabs.github.io/smithy/2.0/spec/constraint-traits.html
[The RFC]: https://github.com/awslabs/smithy-rs/pull/1199

Prerequisite PRs
----------------

That make the diff of this one lighter:

* https://github.com/awslabs/smithy-rs/pull/1438
* https://github.com/awslabs/smithy-rs/pull/1439
* https://github.com/awslabs/smithy-rs/pull/1440
* https://github.com/awslabs/smithy-rs/pull/1441
* https://github.com/awslabs/smithy-rs/pull/1444
* https://github.com/awslabs/smithy-rs/pull/1445
* https://github.com/awslabs/smithy-rs/pull/1746
* https://github.com/awslabs/smithy-rs/pull/1805
* https://github.com/awslabs/smithy-rs/pull/1827
* https://github.com/awslabs/smithy-rs/pull/1832
* https://github.com/awslabs/smithy-rs/pull/1841
* https://github.com/awslabs/smithy-rs/pull/1863

### CODE REVIEW

File: codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt [hunk 1]
```diff
@@ -166,12 +184,53 @@ class ServerCodegenVisitor(context: PluginContext, private val codegenDecorator:
...
+            rustCrate.withModule(unconstrainedModule) { writer ->
+                UnconstrainedMapGenerator(
+                    model,
+                    symbolProvider,
+                    unconstrainedShapeSymbolProvider,
+                    constraintViolationSymbolProvider,
+                    writer,
+                    shape
+                ).render()
+            }
```
crisidev, line 233: I am probably missing something here.. Can't there be constrained maps and lists?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -0,0 +1,487 @@
...
+import software.amazon.smithy.rust.codegen.smithy.traits.SyntheticInputTrait
+import software.amazon.smithy.rust.codegen.smithy.wrapMaybeConstrained
+import software.amazon.smithy.rust.codegen.util.hasTrait
+import software.amazon.smithy.rust.codegen.util.toPascalCase
+import software.amazon.smithy.rust.codegen.util.toSnakeCase
+
+// TODO Document differences:
+//     - This one takes in codegenContext.
+//     - Unlike in `BuilderGenerator.kt`, we don't add helper methods to add items to vectors and hash maps.
+//     - This builder is not `PartialEq`.
```
crisidev, line 49: Why is it not `PartialEq`? Below you derive `PartialEq` for the builder structure..

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -0,0 +1,487 @@
...
+        writer.docs("See #D.", structureSymbol)
+        writer.withModule(moduleName) {
+            renderBuilder(this)
+        }
+    }
+
+    private fun renderBuilder(writer: RustWriter) {
+        if (isBuilderFallible) {
+            Attribute.Derives(setOf(RuntimeType.Debug, RuntimeType.PartialEq)).render(writer)
+            // TODO(): `#[non_exhaustive] unless we commit to making builders of builders public.
```
crisidev, line 75: This can go away, you already added `#[non_exhaustive]` two lines below.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -0,0 +1,487 @@
...
+            writer.docs("Holds one variant for each of the ways the builder can fail.")
```
crisidev, line 76:
```suggestion
            writer.docs("Holds one variant for each types of failure")
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -0,0 +1,487 @@
...
+
+                if (takeInUnconstrainedTypes && member.targetCanReachConstrainedShape(model, symbolProvider)) {
+                    renderBuilderMemberSetterFn(this, member)
+                }
+            }
+            renderBuildFn(this)
+        }
+    }
+
+    // TODO This impl does not take into account sensitive trait.
```
crisidev, line 122: Do we have an issue open for this?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -0,0 +1,487 @@
...
+            structureSymbol.wrapMaybeConstrained()
+        )
+    }
+
+    private fun renderBuildFn(implBlockWriter: RustWriter) {
+        val returnType = when (isBuilderFallible) {
+            true -> "Result<${implBlockWriter.format(structureSymbol)}, ConstraintViolation>"
+            false -> implBlockWriter.format(structureSymbol)
+        }
+        implBlockWriter.docs("""Consumes the builder and constructs a #D.""", structureSymbol)
```
crisidev, line 171:
```suggestion
        implBlockWriter.docs("Consumes the builder and constructs a #D.", structureSymbol)
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedListGenerator.kt
```diff
@@ -0,0 +1,90 @@
...
+import software.amazon.smithy.rust.codegen.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.server.smithy.ConstraintViolationSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.smithy.RustSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.UnconstrainedShapeSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.canReachConstrainedShape
+import software.amazon.smithy.rust.codegen.smithy.wrapMaybeConstrained
+
+// TODO Docs
+class UnconstrainedListGenerator(
```
crisidev, line 22: Both the class name and the file should be prefixed as the rest with `Server`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedMapGenerator.kt
```diff
@@ -0,0 +1,131 @@
...
+import software.amazon.smithy.rust.codegen.server.smithy.ConstraintViolationSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.smithy.RustSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.UnconstrainedShapeSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.canReachConstrainedShape
+import software.amazon.smithy.rust.codegen.smithy.hasConstraintTrait
+import software.amazon.smithy.rust.codegen.smithy.wrapMaybeConstrained
+
+// TODO Docs
+class UnconstrainedMapGenerator(
```
crisidev, line 27: Both the class name and the file should be prefixed as the rest with Server.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedListGeneratorTest.kt
```diff
@@ -0,0 +1,141 @@
...
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverRenderWithModelBuilder
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverTestSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.UnconstrainedShapeSymbolProvider
+import software.amazon.smithy.rust.codegen.testutil.TestWorkspace
+import software.amazon.smithy.rust.codegen.testutil.asSmithyModel
+import software.amazon.smithy.rust.codegen.testutil.compileAndTest
+import software.amazon.smithy.rust.codegen.testutil.unitTest
+import software.amazon.smithy.rust.codegen.util.lookup
+
+class UnconstrainedListGeneratorTest {
```
crisidev, line 23: Both the class name and the file should be prefixed as the rest with Server.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/UnconstrainedShapeSymbolProviderTest.kt
```diff
@@ -0,0 +1,105 @@
...
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverTestSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.UnconstrainedShapeSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.rustType
+import software.amazon.smithy.rust.codegen.testutil.asSmithyModel
+import software.amazon.smithy.rust.codegen.util.lookup
+
+// TODO I can't move this file to `codegen` subproject because `serverTestSymbolProvider` is only in the `codegen-server`
+//  subproject, but I need `UnconstrainedSymbolProvider` to be in the `codegen` subproject because it's used in the
+//  `JsonParserGenerator`.
+class UnconstrainedShapeSymbolProviderTest {
```
crisidev, line 24: Both the class name and the file should be prefixed as the rest with Server.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedMapGeneratorTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverRenderWithModelBuilder
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverTestSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.UnconstrainedShapeSymbolProvider
+import software.amazon.smithy.rust.codegen.testutil.TestWorkspace
+import software.amazon.smithy.rust.codegen.testutil.asSmithyModel
+import software.amazon.smithy.rust.codegen.testutil.compileAndTest
+import software.amazon.smithy.rust.codegen.testutil.unitTest
+import software.amazon.smithy.rust.codegen.util.lookup
+
+class UnconstrainedMapGeneratorTest {
```
crisidev, line 23: Both the class name and the file should be prefixed as the rest with Server.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
david-perez: Yes, but constraint traits are not implemented in this PR. This PR only generates the "unconstrained type" corresponding to a constrained shape.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
david-perez: The comment is supposed to convey that we should remove `#[non_exhaustive]` if we make builders of builders public.

crisidev: Oh ok.. It was not clear to me :)

david-perez:
Ok, this finally has all the functionality that I want to add; I don't plan on expanding the scope of this. https://github.com/awslabs/smithy-rs/pull/1342/commits/d34af6482a0b1cc183f2a1935bb3d5f7aff2d77e is the first commit to pass CI with all the functionality. 

I'm punting some things in https://github.com/awslabs/smithy-rs/issues/1401 so as to not make this PR any bigger.

From now on my efforts will be directed towards improving the documentation and the unit tests, and pulling out some small patches into separate piecemeal PRs to make the diff that needs to be reviewed by @awslabs/rust-sdk-owners lighter.

david-perez:
Commit range c2e2b56b..91cce7a8 is an attempt at introducing a `codegenConfig.publicConstrainedTypes` setting in `rust-server-codegen` plugin's `smithy-build.json` to allow for disabling the generated wrapper newtypes for constrained shapes: the constraint traits would still be honored upon parsing incoming requests, but the user would be handed an unconstrained Rust type. So, for example, given the model:

```smithy
@length(min: 2, max: 69)
string LengthString
```

The corresponding constrained wrapper newtype:

```rust
pub(crate) struct LengthString(pub(crate) std::string::String);
#[allow(unused)]
impl LengthString {
    /// Constructs a `LengthString` from an [`std::string::String`], failing when the provided value does not satisfy the modeled constraints.
    pub fn parse(
        value: std::string::String,
    ) -> Result<Self, crate::model::length_string::ConstraintViolation> {
        Self::try_from(value)
    }

    /// Returns an immutable reference to the underlying [`std::string::String`].
    pub fn inner(&self) -> &std::string::String {
        &self.0
    }

    /// Consumes the value, returning the underlying [`std::string::String`].
    pub fn into_inner(self) -> std::string::String {
        self.0
    }
}
```

would only be used by request parsing framework code to enforce the constraint trait, but the user's operation handler would receive a regular unconstrained `String`.

Now, my intention had been to honor constraint traits _in user code too_, when the user constructs an aggregate shape. So if we have:

```smithy
structure ConA {
    lengthString: LengthString,
}
```

then `ConA`'s builder would check the constraint trait upon `.build()`. This would entail that `struct ConA`'s members would have to be declared at least `pub(crate)`, since otherwise user code can violate constraint traits, bypassing the builder:

```rust
pub struct ConA {
    #[doc(hidden)]
    pub length_string: std::option::Option<std::string::String>, // This would need to not be `pub`.
}
```

This would also apply when the member contained a very nested constrained shape. For example, consider that `ConA` holds a member shape `m` that is a list of a list of a list of a map whose keys are strings and whose values are `LengthString`s. Then the builder's member setter method takes in a `Vec<Vec<Vec<HashMap<String, String>>>>`, and validation needs to convert that into an analogous structure holding `LengthString` first before reconverting back after performing the validation. Fine. I was halfway through the pain of implementing that.

But then I tackled union shapes whose members can be (transitively) constrained. Union shapes are generated as `enum`s in Rust:

```smithy
/// A union with constrained members.
union ConstrainedUnion {
    lengthString: LengthString,
}
```

```rust
/// A union with constrained members.
#[derive(std::clone::Clone, std::cmp::PartialEq, std::fmt::Debug)]
pub enum ConstrainedUnion {
    LengthString(std::string::String),
}
```

I don't think there is an ergonomically viable solution that allows for the variants to hold regular types when `publicConstrainedTypes` is disabled _and_ that honors constraint traits at the same time.

So far I had honored constraint traits _upon request parsing_ by generating an analogous `pub(crate) enum ConstrainedUnionUnconstrained` that I first parsed into that then implements `impl TryFrom<ConstrainedUnionUnconstrained> for ConstrainedUnion`. This type plays the analogous role of the structure builder for structure shapes above. However, what to do to prevent constraint violations in user code? We can't limit the visibility of tuple enum variants.

The canonical way of preventing enum variant construction in Rust is:

```rust
pub enum ConstrainedUnion {
    LengthString(LengthStringInner),
}

// Private.
struct LengthStringInner(String);

// LengthStringInner has constructors that honor constraint traits.
```

But then users can't `match` on `ConstrainedUnion`s without using extra syntax to access the inner `String` like they would be used to. Notice `LengthStringInner` is isomorphic to our `LengthString` with `publicConstrainedTypes` enabled too. Constructors for `ConstrainedUnion` would make use of the constructors for the private structs to honor the constraint traits... It all becomes very unergonomic.

Bottom line is, I don't think we can prevent the user from instantiating union shapes (Rust `enum`s) with data violating constraints, _without making very substantial changes that would make `enum`s not feel like `enum`s_. The API delta between `publicConstrainedTypes` enabled and disabled is too big and unintuitive.

For all the above, I'm leaning to not pursuing `publicConstrainedTypes` any more. At least not with the semantics described above. I now think `publicConstrainedTypes` set to `false` should work exactly like in the Smithy TypeScript server SDK: constraint traits should be honored upon request parsing as validation (internally, the framework will go through the constrained wrapper newtypes) and the user receives an _unconstrained_ type. From then onwards, the user is on their own: they'll be able to construct and work with unconstrained types thereafter, _and modeled constraint traits won't ever be checked thereafter_.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerCodegenVisitor.kt
```diff
@@ -61,14 +61,43 @@ class PythonServerCodegenVisitor(
             )
                 .protocolFor(context.model, service)
         protocolGeneratorFactory = generator
+
         model = codegenDecorator.transformModel(service, baseModel)
-        symbolProvider = PythonCodegenServerPlugin.baseSymbolProvider(model, service, symbolVisitorConfig)
 
-        // Override `codegenContext` which carries the symbolProvider.
-        codegenContext = ServerCodegenContext(model, symbolProvider, service, protocol, settings)
+        // `publicConstrainedTypes` must always be `false` for the Python server.
```
crisidev, line 67: I think I understand, but care to explain in the comment why it must be always false?

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerEnumGenerator.kt
```diff
@@ -48,11 +42,6 @@ class PythonServerEnumGenerator(
         Attribute.Custom("pyo3::pyclass", symbols = pyo3Symbols).render(writer)
     }
 
-    override fun renderFromForStr() {
-        renderPyClass()
-        super.renderFromForStr()
-    }
```
crisidev, line 54: Nice, this was a ugly workaround, I am glad is gone.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedShapeGeneratorCommon.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package software.amazon.smithy.rust.codegen.server.smithy.generators
+
+/**
+ * Functions shared amongst the constrained shape generators, to keep them DRY and consistent.
+ */
+
+fun rustDocsNote(typeName: String) =
+    "this is a constrained type because its corresponding modeled Smithy shape has one or more " +
```
crisidev, line 13: Capital T?

crisidev (commented):
I made it through more or less 50% of the review. I'll post the comments I have for now and continue.

hlbarber (commented):
> This commit makes the sever deserializers work with unconstrained types
during request parsing, and only after the entire request is parsed are
constraints enforced.

Why isn't it better to enforce the constraints eagerly? e.g. If I have a constrained list of constrained strings halt parsing as soon as I hit a violating string.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedMapGenerator.kt
```diff
@@ -0,0 +1,166 @@
...
+import software.amazon.smithy.rust.codegen.client.rustlang.RustWriter
+import software.amazon.smithy.rust.codegen.client.rustlang.Visibility
+import software.amazon.smithy.rust.codegen.client.rustlang.documentShape
+import software.amazon.smithy.rust.codegen.client.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.client.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.client.smithy.ServerCodegenContext
+import software.amazon.smithy.rust.codegen.core.util.expectTrait
+import software.amazon.smithy.rust.codegen.server.smithy.PubCrateConstraintViolationSymbolProvider
+
+/**
```
hlbarber, line 22: Do wrappers need to be codegenerated?

```rust
pub struct Constrained<Inner, C> {
    inner: Inner,
    _constraint: PhantomData<C>
}

impl<Inner, C> TryFrom<Inner> for Constrained<Inner, C>
where
    C: Constraint<Inner>
{
    type Error = C::Error;

    fn try_from(value: Inner) -> Result<Self, Self::Error> {
        C::check(&value)?;
        Ok(Self { inner, _constraint: PhantomData })
    }
}

trait Constraint<Source> {
    type Error;

    fn check(source: &Source) -> Result<(), Self::Error>;
}

pub struct Length<const MIN: Option<usize>, const MAX: Option<usize>>;

impl<const MIN: Option<usize>, const MAX: Option<usize>> Constraint<String> for Length<const MIN: Option<usize>, const MAX: Option<usize>> {
    /* impl here */    
}

impl<const MIN: Option<usize>, const MAX: Option<usize>, K, V> Constraint<HashMap<K, V>> for Length<const MIN: Option<usize>, const MAX: Option<usize>> {
    /* impl here */    
}

impl<Source, C1, C2> Constraint<Source> for (C1, C2) {
    type Error = /* either C1::Error or C2::Error */;

    fn check(source: &Source) -> Result<(), Self::Error> {
        C1::check(source)?;
        C2::check(source)?;
    }
}
```

david-perez:
> Why isn't it better to enforce the constraints eagerly? e.g. If I have a constrained list of constrained strings then halt parsing as soon as I hit a violating string.

See https://github.com/awslabs/smithy-rs/pull/1199#discussion_r809300673.

hlbarber:
Do we want to hide the constraint wrappers entirely? Using them just as internal machinery:

- When parsing a request, use them internally but only expose the unconstrained type and validation error to the user.
- When constructing a response use them internal to the setter, take the unconstrained type as an argument and make the setter return the validation error to the user.

codegen-core/common-test-models/constraints.smithy
```diff
@@ -0,0 +1,441 @@
...
+namespace com.amazonaws.constraints
+
+use aws.protocols#restJson1
+
+/// A service to test aspects of code generation where shapes have constraint traits.
+@restJson1
+@title("ConstraintsService")
+service ConstraintsService {
+    operations: [
+      ConstrainedShapesOperation,
```
hlbarber, line 12: Should these operation names adhere to the same standard as the Pokemon service example? 

> {Verb}[{Qualifier}]{Noun}

david-perez:
> What would we lose under this approach?

It's more idiomatic to work with wrapper newtypes that signal that the underlying data is valid. If your business logic handles `NiceString`s, you can be sure the string is nice by virtue of the type system's mechanics. It fosters correctness. [The RFC](https://github.com/awslabs/smithy-rs/blob/1f5853d619ffaa22762f392944e9257a66829c83/design/src/rfcs/rfc0012_constraint_traits.md#alternative-design) draws other comparisons and tradeoffs.

> When constructing a response use them internal to the body of the setter, take the unconstrained type as an argument, and make the setter return the validation error to the user.

This is already how it works in user application code. If this is supposed to happen within the framework after the user application code has returned an output and relinquished execution, I don't think it's feasible to return execution back to the handler. The most the framework can/will do at this point is to log the violations, but at that point it _must_ render a response.

---

> Do we want to hide the constraint wrappers entirely? Using them just as internal machinery:
> This is would be similar to not having wrappers at all and instead having validation checks in key places.

Unless I'm misunderstanding the proposal, this is exactly how this patch works when `publicUnconstrainedTypes` is set to `false`: the wrappers are used internally, so constraints are enforced, but right before the application handler kicks in, the underlying data is unwrapped before being handed to the user.

Velfi (approved):

LukeMathWalker:
> When constructing a response use them internal to the body of the setter, take the unconstrained type as an argument, and make the setter return the validation error to the user.
What would we lose under this approach?

Performance could also be a concern.
If some data is flowing from the request into the response, you might end up doing multiple rounds of (unnecessary) validation by handling over raw types to the business logic code and then re-validating everything as it enters back into the framework for response creation. 
If you have a wrapper type, you retain in the type system the information that the value has already been validated and you don't have to re-validate perfectly valid data again.

jdisanti (commented):
Here are my comments so far. I need to allocate more time to more thoroughly read through this.

hlbarber (commented):
First impressions:

- Can't say I'm a fan of the two phase parse:
   - Possible performance concerns with doing the initial parse and then converting them in the other phase?
   - DoS concerns with not eagerly exiting?
- Well documented locally, but we need something akin to the PR description somewhere to onboard people if they want to contribute.
- Code looks great, tests look great.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ConstraintViolationSymbolProvider.kt
```diff
@@ -0,0 +1,119 @@
...
+                }"
+                val rustType = RustType.Opaque(constraintViolationName, namespace)
+                Symbol.builder()
+                    .rustType(rustType)
+                    .name(rustType.name)
+                    .namespace(rustType.namespace, "::")
+                    .definitionFile(Models.filename)
+                    .build()
+            }
+            else -> TODO("Constraint traits on other shapes not implemented yet: $shape")
```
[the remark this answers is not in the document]
david-perez: It can't be hit because constraints on other shapes have not been implemented yet, so there are no callers providing other shape types.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/Constraints.kt
```diff
@@ -0,0 +1,125 @@
...
+ * Note that an aggregate shape whose member shapes do not have constraint traits but that has a member whose target is
+ * a constrained shape is _not_ directly constrained.
+ *
+ * At the moment only a subset of constraint traits are implemented on a subset of shapes; that's why we match against
+ * a subset of shapes in each arm, and check for a subset of constraint traits attached to the shape in the arm's
+ * (with these subsets being smaller than what [the spec] accounts for).
+ *
+ * Note `uniqueItems` is deprecated, so we won't ever implement it.
+ *
+ * [the spec]: https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html
```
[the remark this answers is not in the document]
david-perez: I'd say we lean towards linking the 2.0 docs, unless we want to call out a specific difference among the two IDL versions, in which case we cite both.

Yes, Kotlin does not support Markdown formatting when creating Javadocs, but the codebase uses it extensively.

https://github.com/awslabs/smithy-rs/pull/1342/commits/925bd700d749f8edbae01ae3fb7102c146498cf9

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/Constraints.kt
```diff
@@ -0,0 +1,125 @@
...
+
+fun Shape.hasPublicConstrainedWrapperTupleType(model: Model, publicConstrainedTypes: Boolean): Boolean = when (this) {
+    is MapShape -> publicConstrainedTypes && this.hasTrait<LengthTrait>()
+    is StringShape -> !this.hasTrait<EnumTrait>() && (publicConstrainedTypes && this.hasTrait<LengthTrait>())
+    is MemberShape -> model.expectShape(this.target).hasPublicConstrainedWrapperTupleType(model, publicConstrainedTypes)
+    else -> false
+}
+
+fun Shape.wouldHaveConstrainedWrapperTupleTypeWerePublicConstrainedTypesEnabled(model: Model) =
+    hasPublicConstrainedWrapperTupleType(model, true)
```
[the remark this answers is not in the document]
david-perez: https://github.com/awslabs/smithy-rs/pull/1342/commits/b421e0509e24f5f56cec373c0500eb1fc1b56faf

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/PubCrateConstrainedShapeSymbolProvider.kt
```diff
@@ -0,0 +1,120 @@
...
+ * tuple newtype holding a container over the inner constrained type. For
+ * member shapes, it's whatever their target shape resolves to.
+ *
+ * The class name is prefixed with `PubCrate` because the symbols it returns
+ * have associated types that are generated as `pub(crate)`. See the
+ * `PubCrate*Generator` classes to see how these types are generated.
+ *
+ * It is important that this symbol provider _not_ wrap
+ * [ConstrainedShapeSymbolProvider], since otherwise it will eventually
+ * delegate to it and generate a symbol with a `pub` type.
```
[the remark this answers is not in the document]
david-perez: https://github.com/awslabs/smithy-rs/pull/1342/commits/5396e734ac4f8ef321aae5cb2246c0d42339deab

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/PubCrateConstrainedShapeSymbolProvider.kt
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Note simple shapes cannot be transitively but not directly constrained, so
+ * this symbol provider is only implemented for aggregate shapes. The symbol
+ * provider will intentionally crash in such a case to avoid the caller
+ * incorrectly using it.
```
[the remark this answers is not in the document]
david-perez: I meant to convey that simple shapes cannot be transitively constrained, but not directly, since they're... simple. I've reworded this in https://github.com/awslabs/smithy-rs/pull/1342/commits/881ecb3f11fcdd4b5f9e09bdd02fff147476ebc5.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/StructureGenerator.kt
```diff
@@ -74,17 +90,54 @@ open class StructureGenerator(
...
+        /**
+         * Returns whether a structure shape, whose builder has been generated with [ServerBuilderGeneratorWithoutPublicConstrainedTypes],
+         * requires a fallible builder to be constructed.
+         *
+         * This builder only enforces the `required` trait.
+         */
+        fun serverHasFallibleBuilderWithoutPublicConstrainedTypes(
+            structureShape: StructureShape,
+            symbolProvider: SymbolProvider,
+        ) =
```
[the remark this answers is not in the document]
david-perez: https://github.com/awslabs/smithy-rs/pull/1342/commits/ef4b7ba0041e4539e3b1e84302cf73123ace33f6

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/ErrorGenerator.kt
```diff
@@ -78,10 +84,26 @@ class ErrorGenerator(
...
+                    .stripOuter<RustType.Option>() is RustType.Opaque
+                ) {
+                    // The string shape has a constraint trait that makes its symbol be a wrapper tuple struct.
+                    if (messageSymbol.isOptional()) {
+                        "Option<&${
+                        messageSymbol.rustType().stripOuter<RustType.Option>().render()
+                        }>" to "self.${symbolProvider.toMemberName(messageShape)}.as_ref()"
+                    } else {
+                        "&${messageSymbol.rustType().render()}" to "&self.${symbolProvider.toMemberName(messageShape)}"
+                    }
```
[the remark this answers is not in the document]
david-perez: Hmm not that I can think of. https://github.com/awslabs/smithy-rs/pull/1342/commits/48920b539f1f0c43f38d0b9b7a159017b9291c00 is as far as I was able to simplify it.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -522,7 +555,7 @@ class HttpBindingGenerator(
             val listHeader = memberType is CollectionShape
             rustTemplate(
                 """
-                for (k, v) in $field {
+                for (k, v) in ${ if (workingWithPublicConstrainedWrapperTupleType(memberShape, coreCodegenContext)) "&$field.0" else field } {
```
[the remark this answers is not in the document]
david-perez: https://github.com/awslabs/smithy-rs/pull/1342/commits/2c0c0d6dac09ad3ae367b89e761e104518a52929

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/PubCrateConstrainedShapeSymbolProvider.kt
```diff
@@ -0,0 +1,120 @@
...
+    override fun toSymbol(shape: Shape): Symbol {
+        require(shape.isTransitivelyButNotDirectlyConstrained(model, base)) { errorMessage(shape) }
+
+        return when (shape) {
+            is CollectionShape, is MapShape -> {
+                constrainedSymbolForCollectionOrMapShape(shape)
+            }
+            is MemberShape -> {
+                require(model.expectShape(shape.container).isStructureShape) {
+                    "This arm is only exercised by `ServerBuilderGenerator`"
```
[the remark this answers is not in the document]
david-perez: Yep, this is saying that we only care about _structure_ member shapes. A `MemberShape`'s container can be e.g. a list shape too.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/InstantiatorTest.kt
```diff
@@ -67,29 +67,6 @@ class InstantiatorTest {
             value: Integer
         }
 
-        structure MyStructRequired {
```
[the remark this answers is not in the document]
david-perez: It has just been moved to `ServerInstantiatorTest.kt`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/traits/SyntheticAggregateShapeReachableFromOperationInputTagTrait.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import software.amazon.smithy.model.node.Node
+import software.amazon.smithy.model.shapes.ShapeId
+import software.amazon.smithy.model.traits.AnnotationTrait
+
+class SyntheticAggregateShapeReachableFromOperationInputTagTrait() : AnnotationTrait(ID, Node.objectNode()) {
+    companion object {
+        val ID = ShapeId.from("smithy.api.internal#syntheticStructureReachableFromOperationInputTag")
+    }
+}
```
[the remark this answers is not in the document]
david-perez: This has been renamed as `codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/traits/ShapeReachableFromOperationInputTagTrait.kt` and I've added documentation.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedShapeGeneratorCommon.kt
david-perez: Nope, this message gets rendered as in `Note: <message>` in Rust docs.

rust-runtime/inlineable/src/constrained.rs
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+pub(crate) trait Constrained {
```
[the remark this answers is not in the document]
david-perez: Good question. I remember I went down this rabbit hole, but I forgot about it, so I went down again. This time I documented my efforts. I tried to do so in https://github.com/awslabs/smithy-rs/pull/1342/commits/1de207c39039ae614cdb5f813ed913c29b824de2.

However, as the commit message notes:

> it's not possible, because if we do, then we need both
`RuntimeType`s to declare a dependency on
`ServerCargoDependency.SmithyHttpServer`, but we can't do that in
`RuntimeType.kt`, which is in `codegen-core`, since we can't make
`codegen-core` take a dependency on `codegen-server`.
>
> But if we put them in `ServerRuntimeType`, we need to move the `data
class MaybeConstrained` (which would move the `MaybeConstrained`
`RuntimeType`) out of `RustTypes`, since `RustTypes` is in in
`codegen-core`. But `RustTypes` is `sealed`, and we can't inherit from a
sealed class outside the module where the class is defined. We'd have to
make `RustTypes` `abstract`. But then `RustType.render` can't be
exhaustive, and I don't know how to continue.

I reverted the commit, but I'd be happy if someone smarter than me with Kotlin could shuffle the right bits around to enable moving this to `aws-smithy-http-server`.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/Constraints.kt
```diff
@@ -0,0 +1,125 @@
...
+ *     - in the case of it being an aggregate shape, one of its member shapes has a constraint trait.
+ *
+ * Note that an aggregate shape whose member shapes do not have constraint traits but that has a member whose target is
+ * a constrained shape is _not_ directly constrained.
+ *
+ * At the moment only a subset of constraint traits are implemented on a subset of shapes; that's why we match against
+ * a subset of shapes in each arm, and check for a subset of constraint traits attached to the shape in the arm's
+ * (with these subsets being smaller than what [the spec] accounts for).
+ *
+ * Note `uniqueItems` is deprecated, so we won't ever implement it.
```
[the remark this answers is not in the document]
david-perez: Yep, I've removed this sentence and I now link against the 2.0 spec.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/Constraints.kt
```diff
@@ -0,0 +1,125 @@
...
+import software.amazon.smithy.model.traits.EnumTrait
+import software.amazon.smithy.model.traits.LengthTrait
+import software.amazon.smithy.rust.codegen.client.smithy.generators.CodegenTarget
+import software.amazon.smithy.rust.codegen.core.util.UNREACHABLE
+import software.amazon.smithy.rust.codegen.core.util.hasTrait
+
+/**
+ * This file contains utilities to work with constrained shapes.
+ *
+ * TODO Move this file to `core` or `server`.
```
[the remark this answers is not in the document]
david-perez: Only `ConstraintsButThisShouldNotStayInCore.kt` is now in `codegen-core` but it should be in `codegen-server`.

I'd also like to move `RequestBindingGenerator`, `ResponseBindingGenerator`, and `BuilderGenerator` from `codegen-core` to `codegen-client`, but those moves are proving to be tricky. I think we need to split `Instantiator` among client and server, and things should untangle more easily. I'll see if I can tackle that in a separate PR before this one lands.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/rustlang/RustTypes.kt
```diff
@@ -237,6 +243,7 @@ fun RustType.render(fullyQualified: Boolean = true): String {
         is RustType.Box -> "${this.name}<${this.member.render(fullyQualified)}>"
         is RustType.Dyn -> "${this.name} ${this.member.render(fullyQualified)}"
         is RustType.Opaque -> this.name
+        is RustType.MaybeConstrained -> "${this.name}<${this.member.render(fullyQualified)}>"
```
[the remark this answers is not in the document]
david-perez: I wouldn't characterize this class as such. This class contains Rust types that are mapped to from Smithy shapes (that is, a symbol provider inserts it into a `Symbol`). `MaybeConstrained` falls into that category. It just so happens to be our first non-std type.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -424,9 +431,12 @@ class JsonSerializerGenerator(
         val valueName = safeName("value")
         rustBlock("for ($keyName, $valueName) in ${context.valueExpression.asRef()}") {
             val keyTarget = model.expectShape(context.shape.key.target)
-            val keyExpression = when (keyTarget.hasTrait<EnumTrait>()) {
-                true -> "$keyName.as_str()"
-                else -> keyName
+            val keyExpression = if (workingWithPublicConstrainedWrapperTupleType(keyTarget, coreCodegenContext)) {
```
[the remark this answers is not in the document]
david-perez: I've tucked away all of these conditionals in the core generators inside customization classes that only the server makes use of. So now all core generators should be constraint trait agnostic.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt
```diff
@@ -259,36 +262,37 @@ class EventStreamUnmarshallerGenerator(
...
 
     private fun RustWriter.renderParseProtocolPayload(member: MemberShape) {
         val parser = protocol.structuredDataParser(operationShape).payloadParser(member)
-        val memberName = symbolProvider.toMemberName(member)
         rustTemplate(
             """
             #{parser}(&message.payload()[..])
                 .map_err(|err| {
-                    #{Error}::Unmarshalling(format!("failed to unmarshall $memberName: {}", err))
+                    #{Error}::Unmarshalling(format!("failed to unmarshall ${member.memberName}: {}", err))
```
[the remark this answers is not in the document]
david-perez: We're using `member.memberName` as the discriminator and in error messages pervasively. Union member shape symbol names are just being pascal cased.

I'll make a note here that this needs further investigation; I suspect we may already have inconsistencies of the sort I uncovered in https://github.com/awslabs/smithy-rs/issues/1826 for operation names.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedStringGenerator.kt
```diff
@@ -0,0 +1,158 @@
...
+        constrainedTypeMetadata.render(writer)
+        writer.rust("struct $name(pub(crate) $inner);")
+        if (constrainedTypeVisibility == Visibility.PUBCRATE) {
+            Attribute.AllowUnused.render(writer)
+        }
+        writer.rustTemplate(
+            """
+            impl $name {
+                /// ${rustDocsParseMethod(name, inner)}
+                pub fn parse(value: $inner) -> Result<Self, #{ConstraintViolation}> {
```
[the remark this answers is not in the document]
david-perez: Good callout. I wanted all newtype wrappers to have a `parse` constructor that took in the wrapped type as is. Shall I make an exception with string shapes with constraint traits? We need to take ownership anyway; if we took in a `&str` we'll clone it if parsing succeeds.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderConstraintViolations.kt
```diff
@@ -0,0 +1,168 @@
...
+                        constraintViolationSymbolProvider.toSymbol(targetShape)
+                            // If the corresponding structure's member is boxed, box this constraint violation symbol too.
+                            .letIf(constraintViolation.forMember.hasTrait<RustBoxTrait>()) {
+                                it.makeRustBoxed()
+                            }
+
+                    // Note we cannot express the inner constraint violation as `<T as TryFrom<T>>::Error`, because `T` might
+                    // be `pub(crate)` and that would leak `T` in a public interface.
+                    writer.docs("${constraintViolation.message(symbolProvider, model)}.")
+                    Attribute.DocHidden.render(writer)
```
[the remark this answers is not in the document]
david-perez: Right; constraint violation structs need work. Some variants can only arise upon parsing a request (think a transitively but not directly constrained shape), while only the constraint violations that arise when you build a directly constrained shape are the ones that service owners will interact with in their application code. This are two distinct use cases, but in an effort to not make this PR any bigger I'm reusing the same enum, marking it `#[non_exhaustive]`, and making the former variants `#[doc(hidden)]`.

There are other considerations that we need to take into account when we split this enum into the final constraint violation types, which I still haven't decided upon.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedMapGenerator.kt
```diff
@@ -0,0 +1,166 @@
...
+            """,
+            *codegenScope,
+        )
+
+        if (!publicConstrainedTypes && isValueConstrained(shape, model, symbolProvider)) {
+            writer.rustTemplate(
+                """
+                impl #{From}<$name> for #{FullyUnconstrainedSymbol} {
+                    fn from(value: $name) -> Self {
+                        value
```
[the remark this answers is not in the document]
david-perez: Yes, these are definitely _not_ free. The compiler is not as smart as I had imagined; even `<T as Into<T>>::into` is _not_ a no-op.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt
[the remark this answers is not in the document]
jdisanti: This may be a case where the SDK and server want different things. Servers will want to use the modeled name in errors since the people working on them will be familiar with the model, whereas, customers of the SDK don't see and aren't familiar with the model at all, so the Rust name is actually better for them.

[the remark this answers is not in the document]
david-perez: I've reverted this.

david-perez:
>  DoS concerns with not eagerly exiting? e.g. Customers might expect the memory usage to bound by the constraints given when in reality these constraints are not imposed on the first phase.

We should probably special-case `length` and short-circuit.

codegen-core/common-test-models/constraints.smithy
david-perez: I'm struggling coming up with reasonable suggestions for `Verb` and `Noun` for such academic examples.

david-perez:
@hlbarber @jdisanti @crisidev This is ready for another review round. I believe all comments have been addressed; I'm tracking everything that needs to be done prior to merging in the form of [TODOs without context](https://github.com/awslabs/smithy-rs/actions/runs/3315179892/jobs/5475447664#step:3:267) so that CI fails. If something you want addressed is not captured there, please leave a comment; GitHub's UI makes it difficult to see all open conversation threads with so much going on.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -76,6 +78,16 @@ open class Instantiator(
...
     )
 
+    /**
+     * Client and server structures have different builder types. `Instantiator` needs to know how the builder
+     * type behaves to generate code for it.
+     */
+    interface BuilderKindBehavior {
+        fun hasFallibleBuilder(shape: StructureShape): Boolean
+        fun setterName(memberShape: MemberShape): String
+        fun doesSetterTakeInOption(memberShape: MemberShape): Boolean
```
jdisanti, line 88: I think these deserve some comments since builders (at least for the client) have two setter methods per field, one of which takes an option if it's an optional field.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -165,7 +177,9 @@ open class Instantiator(
             writer.conditionalBlock(
                 "Some(",
                 ")",
-                conditional = model.expectShape(memberShape.container) is StructureShape || symbol.isOptional(),
+                // The conditions are not commutative: note client builders always take in `Option<T>`.
```
jdisanti, line 180: > note client builders always take in `Option<T>`

This might be changing soon with IDLv2 nullability support. Although, if it's calling the `set_*` builder method, those will likely stay the same since someone should be able to unset a field on the builder.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/error/ErrorGenerator.kt
```diff
@@ -82,10 +88,22 @@ class ErrorGenerator(
...
             if (messageShape != null) {
-                val (returnType, message) = if (symbolProvider.toSymbol(messageShape).isOptional()) {
-                    "Option<&str>" to "self.${symbolProvider.toMemberName(messageShape)}.as_deref()"
+                val messageSymbol = symbolProvider.toSymbol(messageShape).mapRustType { t -> t.asDeref() }
+                val messageType = messageSymbol.rustType()
+                val (returnType, message) = if (messageType.stripOuter<RustType.Option>() is RustType.Opaque) {
+                    // The string shape has a constraint trait that makes its symbol be a wrapper tuple struct.
+                    if (messageSymbol.isOptional()) {
+                        "Option<&${messageType.stripOuter<RustType.Option>().render()}>" to
+                            "self.${symbolProvider.toMemberName(messageShape)}.as_ref()"
```
jdisanti, line 97: Nit: Move `symbolProvider.toMemberName(messageShape)` into a local variable to improve readability since it's used in every branch.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -386,17 +411,37 @@ class HttpBindingGenerator(
                     })
                     """,
                 )
-            else -> rustTemplate(
-                """
-                if $parsedValue.len() > 1 {
-                    Err(#{header_util}::ParseError::new_with_message(format!("expected one item but found {}", $parsedValue.len())))
+            else -> {
+                if (targetShape is ListShape) {
+                    // This is a constrained list shape and we must therefore be generating a server SDK.
```
jdisanti, line 416: Should we match on the `RustType.MaybeContrained` type to be absolutely sure, and have a catch all `else` that panics since the codegen encountered something it doesn't understand?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -386,17 +411,37 @@ class HttpBindingGenerator(
...
+                        if $parsedValue.len() > 1 {
+                            Err(#{header_util}::ParseError::new_with_message(format!("expected one item but found {}", $parsedValue.len())))
+                        } else {
+                            let mut $parsedValue = $parsedValue;
+                            Ok($parsedValue.pop())
+                        }
+                        """,
+                        "header_util" to headerUtil,
+                    )
+//                }
```
jdisanti, line 442:
```suggestion
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -208,8 +236,29 @@ class JsonParserGenerator(
...
+                                if (symbolProvider.toSymbol(member).isOptional()) {
+                                    withBlock("builder = builder.${member.setterName()}(", ");") {
+                                        deserializeMember(member)
+                                    }
+                                } else {
+                                    rust("if let Some(v) = ")
+                                    deserializeMember(member)
+                                    rust(
+                                        """
+                                        {
```
jdisanti, line 255: Nit: I think this could be made more readable if `deserializeMember` is refactored to return a `Writable`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -54,16 +57,41 @@ import software.amazon.smithy.rust.codegen.core.util.inputShape
...
+    /**
+     * Whether we should parse a value for a shape into its associated unconstrained type. For example, when the shape
+     * is a `StructureShape`, we should construct and return a builder instead of building into the final `struct` the
+     * user gets. This is only relevant for the server, that parses the incoming request and only after enforces
+     * constraint traits.
+     *
+     * The function returns a pair where the second component is the return symbol that should be parsed, and the first
+     * component is whether such symbol is unconstrained or not.
+     */
+    private val returnSymbolToParse: (Shape) -> Pair<Boolean, Symbol> = { shape -> false to codegenContext.symbolProvider.toSymbol(shape) },
```
jdisanti, line 88: I think we should create a light `data class` to represent the pair so that the boolean and symbol can both be named.

jdisanti (approved):
I only have minor comments for the client/core side of things. Good work on this!

rust-runtime/inlineable/src/constrained.rs
```diff
@@ -0,0 +1,15 @@
...
+    type Unconstrained;
+}
+
+#[derive(Debug, Clone)]
+#[allow(dead_code)]
+pub(crate) enum MaybeConstrained<T: Constrained> {
```
hlbarber, line 12: > This commit makes the sever deserializers work with unconstrained types
during request parsing, and only after the entire request is parsed are
constraints enforced.

If

```rust
trait Constrained {
    type Unconstrained; 
    type Error;

    fn constrain(unconstrained: Self::Unconstrained) -> Result<Self, (Self::Unconstrained, Self::Error)>;
}
```

or something similar. The constrain logic gets code generated in this impl.

Then we replaced `MaybeConstrained` with

```rust
enum ConstraintAttempt<T: Constrained> {
    Success(T),
    Failure {
        unconstrained: T::Unconstrained,
        error: T::ConstraintError
    }
}
```

which fails to deserialize exactly when `T::Unconstrained` fails to deserialize:

- `T::Unconstrained` fails to deserialize then return an error,
- `T::Unconstrained` fails to constrain return an `Ok(ConstraintAttempt::Failure { ... })`,
- `T::Unconstrained` does contrain, return an `Ok(Success(T))`.


For example something like this could live alongside `ConstraintAttempt`:

```rust
impl<T> FromStr for ConstraintAttempt<T> where T: Constrained, FromStr {
    type Error = <T as FromStr>::Error;

    fn from_str(value: &str) -> Result<Self, Self::Error> {
        let unconstrained = <T as Constrained>::Unconstrained::from_str(value)?;
        let value = match T::constrain(unconstrained) {
            Ok(ok) => Self::Success(ok),
            Err((unconstrained, error)) => Self::Failure { unconstrained, error }
        };
        Ok(value)
    }
}
```

Then we can "check" constraints during request parsing, but be lazy about collecting errors?

> and only after the entire request is parsed are
constraints enforced.

This then becomes just checking whether `ConstraintAttempt::Success` or not.

Does this make things more compact?

rust-runtime/inlineable/src/constrained.rs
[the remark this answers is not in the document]
david-perez: Yep that's correct.

rust-runtime/inlineable/src/constrained.rs
david-perez: > Does this make things more compact?

I don't see how. It's just swapping `MaybeConstrained` for `ConstrainAttempt`.

> `T::Unconstrained` fails to parse then return an error

I don't see how this can happen. At this point the JSON deserializer has already deserialized a valid node into a Rust data type e.g. a `Vec<String>`, now it's just a matter of whether it satisfies the modeled constraints. Or are you suggesting that `FromStr for ConstraintAttempt<T>` do wire format parsing _and_ deserialization at the same time?

> Then we can "check" constraints during request parsing, but be lazy about collecting errors?

* Lazy error collection is currently how it works: error collection is done when the builder is built, because that's where constraints are checked.
* A benefit of doing constraint checking during request deserialization is that we can "short-circuit" on e.g. the `length` trait if we reach the maximum number of allowed items. But I can't see how `ConstrainAttempt` helps here, since it also takes in the already deserialized value.

david-perez: Yeah what you're describing is having `impl<T> FromStr for ConstraintAttempt<T>` do parsing _and_ deserializing at the same time. It's a viable approach (serde is similar in that the deserializer drives a visitor that does the parsing to map onto serde's data model), but it's not how are code is currently architected. Imagine you had a constrained map holding deeply nested data structures. `impl<T> FromStr for ConstrainAttempt<T>` is now not as straightforward as your implementation where it just delegates to `<T as Constrained>::Unconstrained::from_str`. It now has to parse the JSON tokens for the map elements and delegate to the nested data structures' deserializers, then compose the results. This is essentially what our `JsonParserGenerator` is doing. Its logic could be reused in `impl<T> FromStr for ConstraintAttempt<T>` of course, but it's a non-negligible deviation from our architecture, where we have deserializing and parsing as decoupled concerns.

I do think it's a valuable approach though, since it allows us to more easily short-circuit in the case of `@length`, whereas if we ever want to do that right now, we'd have to trickle down that intent through the `JsonParserGenerator`'s logic.

If we're going to fundamentally change the deserializer's architecture, I'd rather invest in trying out implementing serde's traits though.

hlbarber (approved):
LGTM

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
david-perez: It's calling the `set_*` methods; that is existing behavior I'm not changing.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
david-perez: This is not a `RustType.MaybeConstrained`, it's a `RustType.Opaque` (created by `UnconstrainedCollectionGenerator`). `MaybeConstrained` is only used in server builder members.

david-perez:
Changelog has been updated.

crisidev (approved):

PR #1343, opened 2022-04-26
Fix SDK assemble Gradle pathing issue

jdisanti:
## Motivation and Context

On M1 Macs, Gradle's working directory is the path of the Gradle daemon
rather than the project root. The SDK's `IntegrationTestDecorator` was
assuming that the working directory would always be the project root, so
assemble was failing on this setup.

This commit adds SDK-specific settings so that the integration test
directory can be configured in `smithy-build.json`.

### DIFF
--- a/aws/sdk/build.gradle.kts
+++ b/aws/sdk/build.gradle.kts
@@ -15,7 +15,7 @@ extra["moduleName"] = "software.amazon.smithy.rust.awssdk"
 tasks["jar"].enabled = false
 
 plugins {
-    id("software.amazon.smithy").version("0.5.3")
+    id("software.amazon.smithy").version("0.6.0")
 }
 
 val smithyVersion: String by project
@@ -105,7 +105,12 @@ fun generateSmithyBuild(services: AwsServices): String {
                         "moduleDescription": "${service.moduleDescription}",
                         ${service.examplesUri(project)?.let { """"examples": "$it",""" } ?: ""}
                         "moduleRepository": "https://github.com/awslabs/aws-sdk-rust",
-                        "license": "Apache-2.0"
+                        "license": "Apache-2.0",
+                        "customizationConfig": {
+                            "awsSdk": {
+                                "integrationTestPath": "${project.projectDir.resolve("integration-tests")}"
+                            }
+                        }
                         ${service.extraConfig ?: ""}
                     }
                 }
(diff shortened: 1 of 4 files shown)


aws/sdk/build.gradle.kts
```diff
@@ -105,7 +105,12 @@ fun generateSmithyBuild(services: AwsServices): String {
                         "moduleDescription": "${service.moduleDescription}",
                         ${service.examplesUri(project)?.let { """"examples": "$it",""" } ?: ""}
                         "moduleRepository": "https://github.com/awslabs/aws-sdk-rust",
-                        "license": "Apache-2.0"
+                        "license": "Apache-2.0",
+                        "customizationConfig": {
+                            "awsSdk": {
+                                "integrationTestPath": "${project.projectDir.resolve("integration-tests")}"
+                            }
+                        }
```
rcoh, line 113: cool idea!! I ran into this doing some refactorings, this is a nice way to handle it

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RustSettings.kt
```diff
@@ -81,7 +82,8 @@ class RustSettings(
     val runtimeConfig: RuntimeConfig,
     val codegenConfig: CodegenConfig,
     val license: String?,
-    val examplesUri: String? = null
+    val examplesUri: String? = null,
+    val customizationConfig: ObjectNode? = null
```
rcoh, line 86: I don't _love_ this name although I don't have a better one in mind. Maybe "pluginConfig"? Not sure. I just feel like `Customization` is sort of an internally used word that might not have clear meaning otherwise

rcoh (approved):
this is a great idea (and I think this also fixes some other issues)

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RustSettings.kt
jdisanti: I didn't want to do `pluginConfig` since smithy-rs itself is a plugin for Smithy, so it's more like `pluginOfAPluginConfig` if we go that route. I'm definitely open to suggestions as I'm struggling to think of a better name.

PR #1344, opened 2022-04-26
fix: Silence home directory warning in Lambda Extensions

ryansb:
## Motivation and Context

Similar to [aws-sdk-rust #307](https://github.com/awslabs/aws-sdk-rust/issues/307), suppress the home directory expansion warning when in Lambda. The fix #893 worked for Lambda functions, but not for Extensions.

Related commit: fbb4bc

## Description

This PR follows on to https://github.com/awslabs/smithy-rs/pull/893 which added `LAMBDA_TASK_ROOT` check. When running as a [Lambda Extension](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html) the environment variable for `LAMBDA_TASK_ROOT` is redacted, so it is not reliable for testing whether you are in Lambda.

To remedy this, I switched the check to use `AWS_LAMBDA_FUNCTION_NAME` so that the check returns correctly for both function handlers and extensions.

## Testing

Ran the full test suite, and validated that this suppresses the warning by rebuilding my local SDK and using it in a Lambda Extension.

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -38,5 +38,5 @@ author = "jdisanti"
 [[aws-sdk-rust]]
 message = "Suppress irrelevant `$HOME` expansion warning when running in a Lambda Extension"
 references = ["smithy-rs#1344"]
-meta = { "breaking" = false, "tada" = false, "bug" = false }
+meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "ryansb"
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -34,3 +34,9 @@ message = "Upgrade to Smithy 1.21.0"
 references = ["smithy-rs#1330"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Suppress irrelevant `$HOME` expansion warning when running in a Lambda Extension"
+references = ["smithy-rs#1344"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
+author = "ryansb"
--- a/aws/rust-runtime/aws-config/src/profile/parser/source.rs
+++ b/aws/rust-runtime/aws-config/src/profile/parser/source.rs
@@ -261,7 +261,7 @@ mod tests {
     #[traced_test]
     #[test]
     fn load_config_file_should_not_emit_warning_on_lambda() {
-        let env = Env::from_slice(&[("LAMBDA_TASK_ROOT", "/")]);
+        let env = Env::from_slice(&[("AWS_LAMBDA_FUNCTION_NAME", "someName")]);
         let fs = Fs::from_slice(&[]);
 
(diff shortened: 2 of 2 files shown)


aws/rust-runtime/aws-config/src/profile/parser/source.rs
```diff
@@ -184,7 +184,9 @@ fn expand_home(
 /// that we can check.
 fn check_is_likely_running_on_a_lambda(environment: &aws_types::os_shim_internal::Env) -> bool {
     // LAMBDA_TASK_ROOT – The path to your Lambda function code.
-    environment.get("LAMBDA_TASK_ROOT").is_ok()
+    // AWS_LAMBDA_RUNTIME_API - Double-check because LAMBDA_TASK_ROOT is redacted to any code running in Extensions
+    // https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html
+    environment.get("LAMBDA_TASK_ROOT").is_ok() || environment.get("AWS_LAMBDA_RUNTIME_API").is_ok()
```
jdisanti, line 189: From what I can tell in the docs, `AWS_LAMBDA_RUNTIME_API` is only available when using a custom runtime. Perhaps we should check `AWS_LAMBDA_FUNCTION_NAME` instead of either of these values. Do you know if `AWS_LAMBDA_FUNCTION_NAME` is set for extensions also?

jdisanti (commented):
Good find. Thank you for the contribution! I want to examine the set of environment variables we have available to us a little more to make sure we choose the right ones, but otherwise, it looks good to go.

aws/rust-runtime/aws-config/src/profile/parser/source.rs
ryansb: `AWS_LAMBDA_RUNTIME_API` is always set in a Lambda extension process, it is required for the communication between the extension and the runtime. 

I'll check on `AWS_LAMBDA_FUNCTION_NAME` and get back to you, if that works for both then we can reduce this back down to one check.

ryansb: Yep, `AWS_LAMBDA_FUNCTION_NAME` exists for both, so I've switched the check to use that.

CHANGELOG.next.toml
```diff
@@ -34,3 +34,9 @@ message = "Upgrade to Smithy 1.21.0"
 references = ["smithy-rs#1330"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Suppress irrelevant `$HOME` expansion warning when running in a Lambda Extension"
+references = ["smithy-rs#1344"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
```
jdisanti, line 41:
```suggestion
meta = { "breaking" = false, "tada" = false, "bug" = true }
```

jdisanti (approved):
Looks great! Thank you!

PR #1346, opened 2022-04-27
Fix local CI on M1 Macs

jdisanti:
## Motivation and Context
The `acquire-build-image` script was coded up with the assumption that
it would only work with x86_64 build images. This commit revises that
script and the `Dockerfile` to correctly work on ARM64 architectures as
well.

## Testing
- Ran `./ci generate-aws-sdk` locally


rcoh (approved):
TIL that docker doesn't abstract over architecture (which makes sense in hindsight)

PR #1348, opened 2022-04-27
add FAQ entry  on "why we don't use `serde`"

Velfi:


design/src/faq.md
```diff
@@ -1,10 +1,19 @@
...
 
-2. Versioning: It is expected that over time we may major-version-bump individual services. New updates will be pushed for _some_ AWS service nearly every day. Maintaining separate crates allows us to only increment versions for the relevant pieces that change.
+2. **Versioning:** It is expected that over time we may major-version-bump individual services. New updates will be pushed for _some_ AWS service nearly every day. Maintaining separate crates allows us to only increment versions for the relevant pieces that change.
 
 It is worth noting that this isn't a set-in-stone design decision. A parent crate may be even be created at some point!
+
+## Why don't the SDK service crates implement `serde::Serialize` or `serde::Deserialize` for any types?
+
+1. **Compilation time:** `serde` makes heavy use of [several crates](https://crates.io/crates/serde_derive/1.0.136/dependencies) *(`proc-macro2`, `quote`, and `syn`)* that are very expensive to compile. Several service crates are already quite large and adding a `serde` dependency would increase compile times beyond what we consider acceptable.
```
crisidev, line 16: I believe that before stating this we should have some measurements. What is considered acceptable?

Velfi: I just ran an unscientific experiment and these are the results:
with `serde`: 2m 02s, 1m 58s, 2m 12s *(375s total across all 3 trials)*
without `serde`: 1m 54s, 1m 36s, 1m 35s *(305s total across all 3 trials)*

For this experiment I added serde derives for serialization and deserialization to each operation and model in the S3 SDK. For each test, I ran `cargo build --release`. In between tests, I ran `cargo clean`. With `serde`, compilations take approx. 23% longer.

Velfi: I've added a new line with that "23% slower" result

rcoh (approved):

PR #1352, opened 2022-04-29
Add a debug log entry for retries

jdisanti:
## Motivation and Context
This PR logs a debug event when a retry is going to be performed so that it's easier to tell retry is working by looking at logs (prompted by https://github.com/awslabs/aws-sdk-rust/issues/524).

## Checklist

### DIFF
--- a/rust-runtime/aws-smithy-client/src/retry.rs
+++ b/rust-runtime/aws-smithy-client/src/retry.rs
@@ -314,7 +314,12 @@ impl RetryHandler {
             }
         };
 
-        tracing::debug!("retrying after {:?}", dur);
+        tracing::debug!(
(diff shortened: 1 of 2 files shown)


rust-runtime/aws-smithy-client/src/retry.rs
```diff
@@ -313,6 +313,8 @@ impl RetryHandler {
                 return None;
             }
         };
+
+        tracing::debug!("retrying after {:?}", dur);
```
rcoh, line 317: should we include more info here? retry kind,  number of retries, etc.?

rcoh (approved):
LGTM, but seems like we might as well add a little more info?

PR #1356, opened 2022-04-29
Add support for `credential_process` from profiles

jszwedko:
## Motivation and Context

Fixes: https://github.com/awslabs/aws-sdk-rust/issues/261

## Description

Starting to add support for `credential_process` in the default credentials chain. I'm just pushing this up as a draft to get early feedback before continuing implementation.

## Testing

I tested this manually by using patching it into http://github.com/vectordotdev/vector but will follow through with tests here.

## Checklist


aws/rust-runtime/aws-config/Cargo.toml
```diff
@@ -39,6 +39,7 @@ zeroize = "1"
 bytes = "1.1.0"
 http = "0.2.4"
 tower = { version = "0.4.8" }
+shlex = { version = "1.1.0" }
```
jszwedko, line 42: I borrowed this shell parsing dependency from what `rusoto` was doing, but open to other ideas here.

https://docs.rs/rusoto_credential/latest/src/rusoto_credential/profile.rs.html#417-427

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        future::ProvideCredentials::new(self.credentials())
+    }
+}
+
+impl CredentialProcessProvider {
+    /// Create new [`CredentialProcessProvider`]
+    pub fn new(credential_process: String) -> Self {
+        Self { credential_process }
+    }
```
jszwedko, line 32: I think it'd be good to refactor this to take a `Vec<String>` rather than parsing in here (`moving the parsing out into aws/rust-runtime/aws-config/src/profile/credentials/repr.rs`) but curious to hear thoughts.

Alternatively, the `new` here could return a `Result` and do the parsing up front.

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -0,0 +1,111 @@
...
+            "Error retrieving credentials from external process: provided command empty",
+        )
+    })?);
+    command.args(iter);
+    Ok(command)
+}
+
+#[cfg(test)]
+mod test {
+    // TODO
```
jszwedko, line 110: Will add.

aws/rust-runtime/aws-config/src/profile/credentials.rs
```diff
@@ -493,4 +501,5 @@ mod test {
     make_test!(retry_on_error);
     make_test!(invalid_config);
     make_test!(region_override);
+    make_test!(credentials_process);
```
jszwedko, line 504: Will implement.

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
```diff
@@ -209,13 +216,19 @@ mod static_credentials {
...
+    dbg!("checking base provider", &profile);
     // the profile must define either a `CredentialsSource` or a concrete set of access keys
     match profile.get(role::CREDENTIAL_SOURCE) {
         Some(source) => Ok(BaseProvider::NamedSource(source)),
-        None => web_identity_token_from_profile(profile)
+        None => credential_process_from_profile(profile)
+            .or_else(|| web_identity_token_from_profile(profile))
             .or_else(|| sso_from_profile(profile))
             .unwrap_or_else(|| Ok(BaseProvider::AccessKey(static_creds_from_profile(profile)?))),
     }
```
jszwedko, line 234: I'm not sure what the usual ordering is for the SDKs. I assume the Java one is pretty canonical but lacks mention of `credential_process`: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html

jszwedko (commented):
Hey all! I'll push this forward more next week but wanted to get it up for any early feedback.

jszwedko:
Ah, also, I integrated this into the profile file provider rather than creating a new one that is part of the default chain. I think it makes sense as part of the profile file provider (this is how rusoto modeled it too) rather than a separate provider that _also_ tries to load profile files, but open to opinions here.

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
jelford: I happen to be looking at some functionality in a similar area (I stumbled on your PR while checking for any existing efforts on MFA), and I've been cribbing from `botocore`. I'm sure someone from the team will be able to do better but I thought I might as well pass over my notes, since I was just in the area - hopefully this helps. I agree with the general sentiment that the ordering should be consistent between SDKs - so I would guess it makes sense to put `credentials_process` after SSO?

---

`botocore` looks like it has the ordering for profiles as](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L110):

1. env
2. assume role
3. [profile](https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L167):
    1. web
    4. sso
    5. credentials process
    6. config
4. legacy ec2
5. boto
6. container
7. instance metadata

I don't have a local checkout of the Java SDK, but I think [the equivalent section for profiles is here](https://github.com/aws/aws-sdk-java/blob/fd409dee8ae23fb8953e0bb4dbde65536a7e0514/aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/ProfilesConfigFile.java#L211), which goes in the order:

1. assume role
2. credentials process
3. static credentials (equiv. to boto's "config" in profile handling?)

jszwedko: Good find, thanks @jelford ! It seems sensible to follow `boto` here. I'll move `credential_process` down below `sso`.

aws/rust-runtime/aws-config/src/credential_process.rs
rcoh: I'd support having new return a result here

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -0,0 +1,111 @@
...
+        CredentialsError::provider_error(
+            "Error retrieving credentials from external process: could not parse provided value as command",
+        )
+    })?;
+    let mut iter = args.iter();
+    let mut command = Command::new(iter.next().ok_or_else(|| {
+        CredentialsError::provider_error(
+            "Error retrieving credentials from external process: provided command empty",
+        )
+    })?);
```
rcoh: I wonder if we should spawn this into a task...the trouble is that we'd need to then abstract over spawning. I'll think about it.

rcoh: I wonder if we should spawn this into a task...the trouble is that we'd need to then abstract over spawning. I'll think about it.

aws/rust-runtime/aws-config/src/profile/credentials.rs
rcoh: we'll need to figure out how to mock this—probably some sort of "fake command" interface will need to get added to our existing OS shim

aws/rust-runtime/aws-config/src/profile/credentials/exec.rs
```diff
@@ -100,6 +101,9 @@ impl ProviderChain {
                     })?
             }
             BaseProvider::AccessKey(key) => Arc::new(key.clone()),
+            BaseProvider::CredentialProcess(credential_process) => Arc::new(
+                CredentialProcessProvider::new(credential_process.to_string()),
```
rcoh, line 105: if `new` returned a result, we could return a profile file error here which is probably the right thing

rcoh, line 105: if `new` returned a result, we could return a profile file error here which is probably the right thing

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
rcoh: yup, lets follow boto

rcoh: yup, lets follow boto

jszwedko: Reordered in [14abfdd](https://github.com/awslabs/smithy-rs/pull/1356/commits/14abfdd7811afc6aaf65765956919c9fdb1989b9)

aws/rust-runtime/aws-config/Cargo.toml
jszwedko: Ended up refactoring this to take the Java SDK's approach of just passing the whole string through `sh -c` / `cmd.exe /C` in [14ad7a9](https://github.com/awslabs/smithy-rs/pull/1356/commits/14ad7a9084a5047d8863f6038e675663738cadd4)

aws/rust-runtime/aws-config/src/profile/credentials.rs
rcoh: also, should validate that the credentials process can be used as the source profile for assume role

rcoh: also, should validate that the credentials process can be used as the source profile for assume role

aws/rust-runtime/aws-config/src/credential_process.rs
jszwedko: Ended up refactoring this to take the Java SDK's approach of just passing the whole string through `sh -c` / `cmd.exe /C` in [14ad7a9](https://github.com/awslabs/smithy-rs/pull/1356/commits/14ad7a9084a5047d8863f6038e675663738cadd4) to avoid needing to do any parsing.

rcoh (commented):
overall this is exactly the right direction! I may end up asking you to not include it in the default chain..by default. We need to figure out if there are any security safeguards needed to start using this generally, but I'll get back to you on that ASAP!

Great work on this so far, this looks awesome!

aws/rust-runtime/aws-config/src/credential_process.rs
jszwedko: 👍 I'm open to thoughts here. I actually up refactoring this to take the Java SDK's approach of just passing the whole string through `sh -c` / `cmd.exe /C` in [14ad7a9](https://github.com/awslabs/smithy-rs/pull/1356/commits/14ad7a9084a5047d8863f6038e675663738cadd4) to avoid needing to do any parsing in the SDK itself.

aws/rust-runtime/aws-config/src/profile/credentials.rs
jszwedko: 👍 I ended up using `echo` for the test in which should be supported on all platforms, but happy to try to sub in command mocking if you prefer.

Ref: [a787b61](https://github.com/awslabs/smithy-rs/pull/1356/commits/a787b61f4da20699afb394e9e43a3305075e7672)

aws/rust-runtime/aws-config/src/profile/credentials/exec.rs
jszwedko: I mentioned this in https://github.com/awslabs/smithy-rs/pull/1356#discussion_r863130251 but I actually ended up pulling out the command parsing logic, following suit with the Java SDK of just passing through the commands to `sh -c` / `cmd.exe /C`.

jszwedko:
Thanks for taking a look @rcoh ! I pushed some more changes and think this is ready for full review.

aws/rust-runtime/aws-config/src/json_credentials.rs
```diff
@@ -177,7 +199,7 @@ pub(crate) fn parse_json_credentials(
 
 pub(crate) fn json_parse_loop<'a>(
     input: &'a [u8],
-    mut f: impl FnMut(Cow<'a, str>, Cow<'a, str>),
+    mut f: impl FnMut(Cow<'a, str>, &Token<'a>) -> Result<(), InvalidJsonCredentials>,
```
jszwedko, line 202: The changes to the JSON parsing to support IMDSv2, SSO, and the differing `credential_process` JSON formats were a bit more awkward than I was hoping. I started with changing this loop function to return all value types since the `credential_process` format has a number in it (`Version`) which spidered out a bit.

aws/rust-runtime/aws-config/src/json_credentials.rs
```diff
@@ -115,7 +118,8 @@ pub(crate) fn parse_json_credentials(
     let mut expiration = None;
     let mut message = None;
     json_parse_loop(credentials_response.as_bytes(), |key, value| {
-        match key {
+        dbg!("calling loop with", &key, &value);
```
Velfi, line 121: Don't forget to change this to a `tracing` call

CHANGELOG.next.toml
```diff
@@ -22,3 +22,14 @@ message = "Log a debug event when a retry is going to be peformed"
 references = ["smithy-rs#1352"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "jdisanti"
+
```
Velfi, line 25:
```suggestion

[[aws-sdk-rust]]
```
missing the header for this entry

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -0,0 +1,207 @@
...
+                DateTime::from_str(expiration.as_ref(), Format::DateTime).map_err(|err| {
+                    InvalidJsonCredentials::InvalidField {
+                        field: "Expiration",
+                        err: err.into(),
+                    }
+                })?,
+            )
+            .map_err(|_| {
+                InvalidJsonCredentials::Other(
+                    "credential expiration time cannot be represented by a SystemTime".into(),
```
Velfi, line 161: I'm a bit confused by this message. Is it saying "you gave us data but we couldn't build a `SystemTime` from it" or "you gave us a `SystemTime` but that's not valid for this"?

jdisanti (commented):
This looks great! Thanks a ton for putting in the work here.

I read through some internal documentation on this credential provider, which is the source of my redaction and process execution comments.

We should put `// Security: <reason>` comments around things like manual `Debug` implementations to avoid them getting converted back into derivations by a would be good intentioned, code simplifying, dev in the future 😃

mpetri:
any update on this functionality?

jdisanti:
> any update on this functionality?

I'll try and get this wrapped up and merged this week.

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let mut command = if cfg!(windows) {
+            let mut command = Command::new("cmd.exe");
+            command.args(&["/C", &self.command]);
+            command
+        } else {
+            let mut command = Command::new("sh");
+            command.args(&["-c", &self.command]);
+            command
+        };
```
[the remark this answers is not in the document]
jdisanti: Interestingly, this is the way the [Go V2 SDK](https://github.com/aws/aws-sdk-go-v2/blame/main/credentials/processcreds/provider.go#L90-L94) and [Java V2 SDKs](https://github.com/aws/aws-sdk-java-v2/blob/master/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ProcessCredentialsProvider.java#L74-L80) are doing it. Boto3 looks like it does [its own platform-based splitting](https://github.com/boto/botocore/blob/faa8e9636b4ce9b80204904c7cadc498b4c69a61/botocore/compat.py#L176), with its own implementation for Windows and using `shlex` for other platforms. I'll dig into this some more to figure out what we should do for Rust.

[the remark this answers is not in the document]
jdisanti: Not going to block on this since Go V2 and Java V2 are doing it this way.

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -0,0 +1,230 @@
...
+
+            _ => {}
+        };
+        Ok(())
+    })?;
+
+    match version {
+        Some(1) => {
+            let access_key_id =
+                access_key_id.ok_or(InvalidJsonCredentials::MissingField("AccessKeyId"))?;
```
rcoh, line 167: as a minor nit for readability, I'd probably handle the error cases & abort in the match then have the bulk of the code be un-nested

aws/rust-runtime/aws-config/test-data/profile-provider/credential_process/fs/home/.aws/config
```diff
@@ -0,0 +1,6 @@
+[default]
```
rcoh, line 1: can we add a test where the command is something like `exit 1`? And a test where it returns invalid JSON?

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -11,33 +11,95 @@ use aws_smithy_types::date_time::Format;
...
+impl<T> fmt::Debug for CommandWithSensitiveArgs<T>
+where
+    T: AsRef<str>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", format!("{}", self))
     }
 }
 
+impl<T> Clone for CommandWithSensitiveArgs<T>
```
rcoh, line 56: I think this is just `#[derive(Clone)]`?

rcoh (approved):
🚢

aws/rust-runtime/aws-config/src/credential_process.rs
jdisanti: It needs the `where T: Clone` clause, but maybe I can just put that bound on the struct itself and then derive...

jdisanti: If I take that approach, then each other trait `impl` needs to have a `T: Clone` bound added, so I don't think it's much of a savings.

jszwedko:
Thank you for taking this across the line @jdisanti ! Thanks to @rcoh @Velfi for the review too. Apologies that I didn't get to circle back to address the comments yet. I'm happy to see this merged in though 😄

PR #1357, opened 2022-04-30
Refactor the `sdk-sync` tool to sync commits in parallel

jdisanti:
## Motivation and Context
Each individual smithy-rs commits takes between 3 and 5 minutes to generate code for. This adds up when the number of commits between releases is in the 20s. This PR runs code generation in parallel as much as possible to reduce the total sync time.


rcoh (commented):
overall this seems like a great change! parallelizing this will save a ton of time. I might consider a refactor to reduce the amount of traits/mocking required—it seems like it adds a pretty significant amount of cognitive load on top of what is in actuality, a fairly simple process

Velfi (approved):

PR #1358, opened 2022-05-02
Point example links to example's in current branch so this will work …

bigOconstant:
## Motivation and Context

The example links are hard coded to main branch when they could (imo should) be pointed to the examples in the current branch. 

With this change example links points to the current examples folder. This will also allow this to work in other git apps such as gitea.

## Description

Changed hard coded example links to point to the examples in the current branch


bigOconstant:
@Velfi

Velfi (approved):

Velfi:
I set CI to run. I'll merge this tomorrow morning once it passes

PR #1361, opened 2022-05-03
Update `FsBuilder`-based `ByteStreams` to support file offsets

Velfi:
## Motivation and Context

This will help with [aws-sdk-rust#494](https://github.com/awslabs/aws-sdk-rust/issues/494)

## Description

The `aws_smithy_http::byte_stream::bytestream_util::FsBuilder` has been updated to allow for easier creation of
multi-part requests.

- `FsBuilder::offset` is a new method allowing users to specify an offset to start reading a file from.
- `FsBuilder::file_size` has been reworked into `FsBuilder::length` and is now used to specify the amount of data to read.

With these two methods, it's now simple to create a `ByteStream` that will read a single "chunk" of a file.

## Testing

I wrote several new tests

## Checklist


rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -609,17 +610,22 @@ mod tests {
...
             .buffer_size(16384)
-            // This isn't the right file length - one shouldn't do this in real code
-            .file_size(200)
+            .length(file_length)
             .build()
             .await?
             .into_inner();
 
-        // assert that the file length specified size is used as size hint
-        assert_eq!(body.size_hint().exact(), Some(200));
```
rcoh, line 622: because we loaded the real file, this doesn't test that the hinting actually works anymore

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -628,14 +634,167 @@ mod tests {
             .await
             .expect("should have some data")
             .expect("read should not fail");
-        // The size of one read should be equal to that of the buffer size
+        // The size of one read should be equal to that of the length
```
rcoh, line 637: I think `buffer size` was correct?

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -628,14 +634,167 @@ mod tests {
...
-            298890 - some_data.len()
+            file_length as usize - some_data.len()
         );
 
         Ok(())
     }
+
+    #[cfg(feature = "rt-tokio")]
+    #[tokio::test]
+    async fn doesnt_read_past_end_of_path_based_bytestreams_with_builder(
```
rcoh, line 650: maybe `respects_length`?

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -628,14 +634,167 @@ mod tests {
...
+
+        let mut file = NamedTempFile::new()?;
+        let line_0 = "Line 0\n";
+        let line_1 = "Line 1\n";
+
+        write!(file, "{}", line_0)?;
+        write!(file, "{}", line_1)?;
+
+        let body = ByteStream::read_from()
+            .path(&file)
```
rcoh, line 664: actually surprised this works—I would have expected that the file needed to be flushed

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -628,14 +634,167 @@ mod tests {
...
+        let data_str = String::from_utf8(data.to_vec())?;
+
+        assert_eq!(&data_str, line_0);
+
+        Ok(())
+    }
+
+    #[cfg(feature = "rt-tokio")]
+    #[tokio::test]
+    async fn read_from_offset_of_path_based_bytestreams_with_builder(
```
rcoh, line 680: `supports_offset`?

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -150,32 +171,54 @@ impl FsBuilder {
         let buffer_size = self.buffer_size;
 
         if let Some(path) = self.path {
-            let path_buf = path.to_path_buf();
-            let file_size = self.file_size.unwrap_or(
-                tokio::fs::metadata(path)
+            let offset = self.offset.unwrap_or(DEFAULT_OFFSET);
+            let length = self.length.unwrap_or({
+                let file_length = tokio::fs::metadata(&path)
```
rcoh, line 176: seems like this should be `unwrap_or_else`? I'm surprised clippy didn't catch this—this will end up always calling metatdata even when length is specified

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -150,32 +171,54 @@ impl FsBuilder {
...
+                    self.offset,
                 )))
             };
             Ok(ByteStream::new(SdkBody::retryable(body_loader)))
-        } else if let Some(file) = self.file {
-            let file_size = self.file_size.unwrap_or(
-                file.metadata()
+        } else if let Some(mut file) = self.file {
+            let offset = self.offset.unwrap_or(DEFAULT_OFFSET);
+            let length = self.length.unwrap_or({
```
rcoh, line 200: same here

rcoh: I guess _technically_ this works because the future is never polled?

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -150,32 +171,54 @@ impl FsBuilder {
         let buffer_size = self.buffer_size;
 
         if let Some(path) = self.path {
-            let path_buf = path.to_path_buf();
-            let file_size = self.file_size.unwrap_or(
-                tokio::fs::metadata(path)
+            let offset = self.offset.unwrap_or(DEFAULT_OFFSET);
```
rcoh, line 174: can you fix this to panic if path and file are both set?

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -150,32 +171,54 @@ impl FsBuilder {
...
+            let length = self.length.unwrap_or({
+                let file_length = tokio::fs::metadata(&path)
                     .await
                     .map_err(|err| Error(err.into()))?
-                    .len(),
-            );
+                    .len();
```
rcoh, line 179: should probably extract a little block to calculate late:

```rust
let file_metadata_sz = async { ... };


let length = self.length.unwrap_or(file_metdata_sz.await?);
```

rcoh: can you verify that this isn't called when length is set?

rcoh (commented):
Looks good! Some naming suggestions + small tweaks

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
Velfi: Yeah, I tried to change it and realized I'd have to do something about the `await` and never got back to it

rust-runtime/aws-smithy-http/src/byte_stream.rs
Velfi: I don't remember why I changed this in the first place lol

rust-runtime/aws-smithy-http/src/byte_stream.rs
Velfi: I'll add a `flush` just to be certain

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
Velfi: Instead, I'll make it impossible to construct one with both.

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
Velfi: I made a better version based on match

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
Velfi: nm, I couldn't accomplish this in a nice way that didn't mess with the current `ByteStream` APIs for `from_path` and `from_file`

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
Velfi: I switched this one to a match statement too

rust-runtime/aws-smithy-http/src/byte_stream.rs
Velfi: I wrote a new test just for the hinting and I think it's correct but I would appreciate your scrutiny

CHANGELOG.next.toml
```diff
@@ -22,3 +22,50 @@ message = "Log a debug event when a retry is going to be peformed"
...
+        // size of the final chunk.
+        file_size - (i * chunk_size)
+    } else {
+        chunk_size
+    };
+
+    let byte_stream = ByteStream::read_from()
+        .path(&file_path)
+        .offset(i * chunk_size)
+        .length(length)
```
jdisanti, line 57: I think the `length` function name change should get called out as its own separate breaking change entry. Are there any other breaking changes in this entry, or is that the only one?

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -139,43 +152,63 @@ impl FsBuilder {
...
+    /// Returns a [`ByteStream`](crate::byte_stream::ByteStream) from this builder.
     pub async fn build(self) -> Result<ByteStream, Error> {
+        if self.path.is_some() && self.file.is_some() {
+            panic!("The 'file' and 'path' options on an FsBuilder are mutually exclusive but both were set. Please set only one")
+        };
+
         let buffer_size = self.buffer_size;
+        let offset = self.offset.unwrap_or(DEFAULT_OFFSET);
+        let length = match self.length {
+            Some(length) => length,
```
jdisanti, line 172: Should `offset + length` be clamped to the file size if they're greater than it?

For example, if I have an 11 byte file, and a chunk size of 5, the last chunk would be 1 byte. If I passed in 5 as the length for that last chunk out of laziness, should that error, or should that just work?

There's also (maybe?) a use case of uploading a file in chunks that is still being written to by another process, so perhaps it should just let this fly without any manipulation.

Velfi: Yeah, I wasn't sure how strict to be so I decided to just be permissive

Velfi: I came up with a new API based on the poll

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -236,11 +301,290 @@ impl Body for PathBody {
...
+        write!(file, "{}", line_1)?;
+
+        // Ensure that the file was written to
+        file.flush().expect("flushing is OK");
+
+        let body = FsBuilder::new()
+            .path(&file)
+            // We're going to skip all file contents by setting an offset
+            // much larger than the file size
+            .offset(9000)
```
jdisanti, line 488: So if the length isn't set at all, we default to `Length::UpTo(actual_file_size)` behavior. Is this consistent?

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -139,56 +161,90 @@ impl FsBuilder {
...
-                    .len(),
-            );
+        let length = match self.length {
+            Some(Length::Exact(length)) => {
+                let file_length = self.get_file_size().await?;
+                if length > file_length.saturating_sub(offset) {
+                    return Err(Error(Box::new(io::Error::new(
+                        io::ErrorKind::UnexpectedEof,
+                        "value Length::Exact was larger than file size minus read offset",
+                    ))));
```
jdisanti, line 190: Looks like there isn't a unit test exercising this error.

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -236,11 +301,290 @@ impl Body for PathBody {
...
+        let data = body.collect().await?.into_bytes();
+        let data_str = String::from_utf8(data.to_vec())?;
+
+        assert_eq!(&data_str, line_1);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn fsbuilder_offset_and_length_work_together() -> Result<(), Box<dyn std::error::Error>> {
```
jdisanti, line 441: Nit: I think tests that `unwrap` are a lot easier to debug when things go wrong than ones that return an error. The latter doesn't give you a nice line number to work with (unless I'm missing some neat trick).

jdisanti (commented):
This looks pretty much good to go. Just one more question on what the correct behavior is for offsets, and then another unit test.

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
Velfi: In the case of not setting a length, the past and present behavior are the same. The entire file will be read.

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
Velfi: good callout, I'll create one

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
rcoh: yeah you'd need `context!()`

rust-runtime/aws-smithy-http/src/byte_stream/bytestream_util.rs
```diff
@@ -236,11 +304,316 @@ impl Body for PathBody {
...
+
+        FsBuilder::new()
+            .path(&file)
+            // We're going to skip all file contents by setting an offset
+            // much larger than the file size
+            .offset(9000)
+            .build()
+            .await
+            // this will panic
+            .unwrap();
```
jdisanti, line 529: There are a few other unwraps in this test, so it could potentially pass even if the code under test is broken. It would be better to do an `assert!(matches!(` with the expected error type here.

jdisanti (approved):
LGTM!

Issue #1363, opened 2022-05-04
Bug - with_capacity function or associated item not found in `ReaderStream<_>`


ZelCloud:
I've tried to update the aws sdk crates to version ```0.11.0```, and unforunately it seems there's an error in aws-smithy-http version ```0.41.0``` which is a dependency of the former. 

Using the latest rust v1.60.0 - 2021 standard.

```bash
   Compiling aws-smithy-http v0.41.0
error[E0599]: no function or associated item named `with_capacity` found for struct `ReaderStream` in the current scope
  --> C:\Users\%USERNAME%\.cargo\registry\src\github.com-1fcc6799dn9ec863\aws-smithy-http-0.41.0\src\byte_stream\bytestream_util.rs:47:48
   |
47 |             state: State::Loaded(ReaderStream::with_capacity(file, buffer_size)),
   |                                                ^^^^^^^^^^^^^ function or associated item not found in `ReaderStream<_>`

error[E0599]: no function or associated item named `with_capacity` found for struct `ReaderStream` in the current scope
   --> C:\Users\%USERNAME%\.cargo\registry\src\github.com-1fcc6799dn9ec863\aws-smithy-http-0.41.0\src\byte_stream\bytestream_util.rs:215:61
    |
215 | ...                   State::Loaded(ReaderStream::with_capacity(file, self.buffer_size));
    |                                                   ^^^^^^^^^^^^^ function or associated item not found in `ReaderStream<_>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `aws-smithy-http` due to 2 previous errors
```

rcoh:
ah—good catch. We are only requiring tokio-util `0.6` but `with_capacity` wasn't added until `0.6.8`. If you run `cargo update` it will fix the issue for you. I'll get a fix in on our side as well

rcoh:
(closing this was a mouse slip)

jdisanti:
It looks like this was fixed a while ago. We're now depending on `tokio-util` 0.7.

PR #1365, opened 2022-05-04
Fix several minimal version bugs and add it to CI

rcoh:
## Motivation and Context
- #1363 (and others)—when we fail to require the minimal version of a dependency, customers with an older semver compatible version in their lockfile will experience compilation failures

## Description
- fix several too-low versions
- add cargo minimal-versions to CI
- removed two `thiserror` usages
## Testing
- ran cargo minimal versions locally

## Checklist
no-changelog


Velfi (approved):

Velfi (approved):

tools/ci-build/scripts/check-rust-runtimes-and-tools
```diff
@@ -23,6 +23,7 @@ do
     cargo clippy --all-features
     cargo test --all-features
     cargo doc --no-deps --document-private-items --all-features
+    cargo minimal-versions check --all-features
```
jdisanti, line 26: Not blocking for this PR, but we need to split `check-rust-runtimes-and-tools` up since its now our long pole.

jdisanti (approved):

Issue #1170, opened 2022-02-09
Consider removing `axum_core` from the server's public API


david-perez:
We currently reexport [`axum_core::body::BoxBody`](https://github.com/awslabs/smithy-rs/blob/a4adea2f54018f1e90425673c2740df9e2fba408/rust-runtime/aws-smithy-http-server/src/body.rs#L15-L15) in our public API, our common body type for responses. Ideally we'd just use [`http_body::combinators::UnsyncBoxBody`](https://docs.rs/http-body/0.4.4/http_body/combinators/struct.UnsyncBoxBody.html), which is also what `axum_core::body::BoxBody` _is_ underneath.

However, we can't move to `UnsyncBoxBody` because we're using the `IntoResponse` trait internally, which leaks axum's `BoxBody` in its signature.

We should consider moving away from `axum_core` entirely and define our own `FromRequest` and `IntoResponse`, since we're only using the _definitions_ of these traits and not the implementations provided by axum. Two other advantages:
* (negligibly) faster compile times for service implementers.
* we can define our own `Operation` associated type in our equivalent `FromRequest`, and get rid of the ugly hack we put in place of wrapping every operation input type in an operation-specific struct (which was introduced because Smithy allows an operation input shape to be reused accross operations, and we need a way to distinguish internally for which operation we're deserializing a request).

crisidev:
<3

PR #1368, opened 2022-05-06
Remove axum-core dependency

82marbag:
This change removes the dependency on axum-core and brings in the
relevant resources we are using from that crate.

Issue: #1170

## Motivation and Context

Closes https://github.com/awslabs/smithy-rs/issues/1170

## Description

The crate axum-core has been removed from the dependencies.
IntoResponse and FromRequest have been kept.
RequestParts has been kept because we need ownership of `body` to consume it in places such as [here](https://github.com/awslabs/smithy-rs/blob/ae9a5ed50a4b14296f4ede85b65f5313f1398df3/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt#L588) and have `&mut` requests, but the http crate provides it as a stream and requests are not `Clone`able.

## Testing

`./gradlew :codegen-server-test:test` in the root directory
`cargo test` in `rust-runtime/aws-smithy-http-server`

## Checklist

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
@@ -38,7 +38,6 @@ class ServerOperationHandlerGenerator(
     private val runtimeConfig = codegenContext.runtimeConfig
     private val codegenScope = arrayOf(
         "AsyncTrait" to ServerCargoDependency.AsyncTrait.asType(),
-        "AxumCore" to ServerCargoDependency.AxumCore.asType(),
         "PinProjectLite" to ServerCargoDependency.PinProjectLite.asType(),
         "Tower" to ServerCargoDependency.Tower.asType(),
         "FuturesUtil" to ServerCargoDependency.FuturesUtil.asType(),
@@ -106,9 +105,9 @@ class ServerOperationHandlerGenerator(
                     """
                     type Sealed = #{ServerOperationHandler}::sealed::Hidden;
                     async fn call(self, req: #{http}::Request<B>) -> #{http}::Response<#{SmithyHttpServer}::body::BoxBody> {
-                        let mut req = #{AxumCore}::extract::RequestParts::new(req);
-                        use #{AxumCore}::extract::FromRequest;
-                        use #{AxumCore}::response::IntoResponse;
+                        let mut req = #{SmithyHttpServer}::request::RequestParts::new(req);
+                        use #{SmithyHttpServer}::request::FromRequest;
+                        use #{SmithyHttpServer}::response::IntoResponse;
                         let input_wrapper = match $inputWrapperName::from_request(&mut req).await {
                             Ok(v) => v,
                             Err(runtime_error) => {
--- a/rust-runtime/aws-smithy-http-server/src/body.rs
+++ b/rust-runtime/aws-smithy-http-server/src/body.rs
@@ -10,9 +10,37 @@ pub use http_body::Body as HttpBody;
 #[doc(hidden)]
 pub use hyper::body::Body;
 
-// `boxed` is used in the codegen of the implementation of the operation `Handler` trait.
+use bytes::Bytes;
+
+use crate::error::{BoxError, Error};
+
 #[doc(hidden)]
-pub use axum_core::body::{boxed, BoxBody};
+pub type BoxBody = http_body::combinators::UnsyncBoxBody<Bytes, Error>;
+
+// `boxed` is used in the codegen of the implementation of the operation `Handler` trait.
+/// Convert a [`http_body::Body`] into a [`BoxBody`].
+#[doc(hidden)]
+pub fn boxed<B>(body: B) -> BoxBody
+where
+    B: http_body::Body<Data = Bytes> + Send + 'static,
+    B::Error: Into<BoxError>,
+{
+    try_downcast(body).unwrap_or_else(|body| body.map_err(Error::new).boxed_unsync())
+}
+
+#[doc(hidden)]
+pub(crate) fn try_downcast<T, K>(k: K) -> Result<T, K>
+where
+    T: 'static,
+    K: Send + 'static,
+{
+    let mut k = Some(k);
+    if let Some(k) = <dyn std::any::Any>::downcast_mut::<Option<T>>(&mut k) {
+        Ok(k.take().unwrap())
+    } else {
+        Err(k.unwrap())
+    }
+}
 
 pub(crate) fn empty() -> BoxBody {
     boxed(http_body::Empty::new())
--- a/rust-runtime/aws-smithy-http-server/src/request.rs
+++ b/rust-runtime/aws-smithy-http-server/src/request.rs
@@ -1,0 +1,117 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// This code was copied and then modified from Tokio's Axum.
+
+/* Copyright (c) 2022 Tower Contributors
+ *
+ * Permission is hereby granted, free of charge, to any
+ * person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the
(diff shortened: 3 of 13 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationHandlerGenerator.kt
```diff
@@ -38,7 +38,6 @@ class ServerOperationHandlerGenerator(
     private val runtimeConfig = codegenContext.runtimeConfig
     private val codegenScope = arrayOf(
         "AsyncTrait" to ServerCargoDependency.AsyncTrait.asType(),
-        "AxumCore" to ServerCargoDependency.AxumCore.asType(),
```
crisidev, line 41: `AxumCore` should be also removed from `ServerCargoDependency.kt`.

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+// This code was copied and then modified from Tokio's Axum.
+
+/* Copyright (c) 2021 Tower Contributors
```
crisidev, line 8:
```suggestion
/* Copyright (c) 2022 Tower Contributors
```

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,204 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+// This code was copied and then modified from Tokio's Axum.
+
+/* Copyright (c) 2021 Tower Contributors
```
crisidev, line 8:
```suggestion
/* Copyright (c) 2022 Tower Contributors
```

rust-runtime/aws-smithy-http-server/src/body.rs
```diff
@@ -10,9 +10,34 @@ pub use http_body::Body as HttpBody;
 #[doc(hidden)]
 pub use hyper::body::Body;
 
+use bytes::Bytes;
+
+use crate::error::{BoxError, Error};
+
+pub type BoxBody = http_body::combinators::UnsyncBoxBody<Bytes, Error>;
```
crisidev, line 17: This type should be `#[doc(hidden)]` because it is not meant to be directly used by the user, but only by the code generator. We use `#[doc(hidden)]` to signal that is is public API that is not supported outside of the codegenerator.

rust-runtime/aws-smithy-http-server/src/body.rs
```diff
@@ -10,9 +10,34 @@ pub use http_body::Body as HttpBody;
...
+
 // `boxed` is used in the codegen of the implementation of the operation `Handler` trait.
-#[doc(hidden)]
-pub use axum_core::body::{boxed, BoxBody};
+/// Convert a [`http_body::Body`] into a [`BoxBody`].
+pub fn boxed<B>(body: B) -> BoxBody
```
crisidev, line 21: Same as above, `#[doc(hidden)]`

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Returns `None` if the body has been taken by another extractor.
+    pub fn body_mut(&mut self) -> Option<&mut B> {
+        self.body.as_mut()
+    }
+
+    /// Takes the body out of the request, leaving a `None` in its place.
+    pub fn take_body(&mut self) -> Option<B> {
+        self.body.take()
+    }
+}
```
crisidev, line 178: I am wondering if we should implement only the methods we use in the codegen instead of everything. @david-perez WDYT?

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,204 @@
...
+
+#[async_trait]
+pub trait FromRequest<B>: Sized {
+    /// If the extractor fails it'll use this "rejection" type. A rejection is
+    /// a kind of error that can be converted into a response.
+    type Rejection: IntoResponse;
+
+    /// Perform the extraction.
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection>;
+}
```
crisidev, line 49: Can we add some documentation on the public traits?

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -0,0 +1,63 @@
...
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+ * IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+use std::convert::Infallible;
+
+use crate::body::BoxBody;
+
+pub type Response<T = BoxBody> = http::Response<T>;
```
crisidev, line 39: I think this should be `#[doc(hidden)]` as it is only used by the codegenerator.

crisidev (commented):
Looks good, just some notes about hiding public types that are only used in the code generator.

rust-runtime/aws-smithy-http-server/src/request.rs
david-perez: The less code the better.

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,204 @@
...
+where
+    T: FromRequest<B>,
+    B: Send,
+{
+    type Rejection = Infallible;
+
+    async fn from_request(req: &mut RequestParts<B>) -> Result<Self, Self::Rejection> {
+        Ok(T::from_request(req).await)
+    }
+}
```
david-perez, line 204: Why do we need these? Shouldn't we only need operation inputs to implement `FromRequest`?

rust-runtime/aws-smithy-http-server/src/error.rs
```diff
@@ -43,6 +42,8 @@ pub struct Error {
     inner: BoxError,
 }
 
+pub type BoxError = Box<dyn std::error::Error + Send + Sync>;
```
david-perez, line 45: `std::error::Error` is `use`d in this module; might as well use it.

david-perez (commented):
This comment needs updating:

https://github.com/awslabs/smithy-rs/blob/f3a2cb3210e35a1b17c95c41393ce0c25bc89a4e/rust-runtime/aws-smithy-http-server/src/protocols.rs#L86-L89

rust-runtime/aws-smithy-http-server/src/error.rs
```diff
@@ -43,6 +42,8 @@ pub struct Error {
     inner: BoxError,
 }
 
+pub type BoxError = Box<dyn StdError + Send + Sync>;
```
crisidev, line 45: This should be `pub(crate)` as it was before when imported from Axum.

rust-runtime/aws-smithy-http-server/src/request.rs
```diff
@@ -0,0 +1,115 @@
...
+ * IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+use async_trait::async_trait;
+use http::{Extensions, HeaderMap, Request, Uri};
+
+use crate::response::IntoResponse;
+
+/// Types that can be created from requests.
```
crisidev, line 40: This documentation makes no sense to me :)

What about something like
```suggestion
/// Trait for extracting information from requests.
///
/// A type implementing the `FromRequest` trait is used to handle and extract information from an async handler taking in a `RequestParts` as input.
```

crisidev (approved):
Just a small suggestion about documentation.

LGTM!

rust-runtime/aws-smithy-http-server/src/request.rs
82marbag: Thanks!

PR #1369, opened 2022-05-06
Do not generate the operation registry when there are no operations

david-perez:
Do not generate the operation registry, nor the operation handler
traits, when there are no operations in the service closure.

## Testing

Without this commit, we codegen a crate for the `empty.smithy` service that
cannot be compiled.


crisidev (approved):

david-perez:
There's no harm in adding the test to the client too I guess.

rcoh (approved):
LGTM

rcoh (commented):
let's make this fail

david-perez:
I'll work on making it throw an exception instead of logging a warning.

LukeMathWalker:
I assume we can close this since the registry is gone? @david-perez

PR #1375, opened 2022-05-06
Enable Gradle parallel builds

david-perez:
It looks like this is a low-hanging optimization. I haven't been able to
observe a performance improvement in my (unscientific) experiments.

https://docs.gradle.org/current/userguide/performance.html#parallel_execution

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,6 +6,8 @@
 # Rust MSRV (entered into the generated README)
 rust.msrv=1.58.1
 
+org.gradle.parallel=true
+
 # Version number to use for the generated SDK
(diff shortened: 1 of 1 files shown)


82marbag (approved):

jdisanti:
I don't think this is going to produce much benefit. The parts that can run in parallel already are for the most part (for example, each Smithy model has code generated in parallel). Running compilation tasks in parallel may be beneficial, but dependency order prevents that. Running Junit tests in parallel would probably help a bit, but I don't think this change will actually accomplish that. Going to close this for now.

PR #1377, opened 2022-05-07
Fix `SPDX-License-Identifier` in header comments

jdisanti:
## Motivation and Context
@hyandell pointed out the `SPDX-License-Identifier` has an unnecessary trailing period in https://github.com/awslabs/aws-sdk-kotlin/issues/601, and I think we must have copied the license header from that repository at some point in the early days of `smithy-rs`. This PR removes the trailing period and updates our linter.


crisidev (approved):
Nice sed and replace :)

PR #1381, opened 2022-05-10
Add excluded headers list in the setting option. for verification …

alonlud:
…signed request

## Motivation and Context
I would like to verify an HTTP request which already been signed with the sigv4, Authorization header.
The header contains:
"SignedHeaders=amz-sdk-invocation-id;amz-sdk-request;amz-sdk-retry;content-length;content-type;host;**user-agent**;x-amz-date;x-amz-target, ..."

To correctly sign the request and verify the signature, I will also need to sign the User-Agent header.

## Description
Add bool flag in the setting option allow_signing_user_agent_header.
The flag is by default false. But can be set to the true outside of the sigv4 crate.
This allows verifying a signed request correctly as the sigv4 is able now to sign also the User-Agent header.


rcoh:
Can you tell me more about your use case? Usually the user agent header is unsigned because it can be altered by proxies

rcoh:
I think this is a useful thing to support but I think we may need a different approach—I suspect we'll want to either accept an explicit list of headers to sign or a function. I think the `User-Agent` will not be the only relevant header. I'm open to design suggestions but otherwise, someone from the Rust SDK team can dig into options

alonlud:
I agree with you, high chance that it would be more useful to change it to some kind of vector of HeaderNames which should not be part of the signing. something like this:

    pub headers_not_to_sign: Option<Vec<HeaderName>>

The default can be: Some([User-Agent])

I can modify the PR to implement this.
Please let me know if this is a valid suggestion

jdisanti:
That sounds reasonable to me.

alonlud:
I have updated the PR.
Let me know what you think.

jdisanti (approved):
LGTM! Thank you for the contribution!

PR #1382, opened 2022-05-10
Fix some bugs in the `sdk-sync` tool

jdisanti:
## Motivation and Context
This PR fixes two issues in `sdk-sync`:
- The temporary `__sdk_sync__models_` branch in `smithy-rs` wasn't getting cleaned up
- `sdk-sync` assumed there was only one commit for model changes rather than multiple


tools/sdk-sync/tests/mock_e2e_test.rs
```diff
@@ -1,605 +0,0 @@
-/*
- * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
- * SPDX-License-Identifier: Apache-2.0
- */
-
```
jdisanti, line 5: I removed this test since it seems to be more of a drag than a boost after `e2e_test.rs` was added. It takes a lot of time to update for changes and doesn't really catch a whole lot.

Velfi (approved):

PR #1395, opened 2022-05-17
RFC: Replacing Callback APIs for `ByteStream` and `SdkBody` with a body-wrapping API

Velfi:
## Motivation and Context

In the process of implementing the new checksum features required by recent S3 updates, I found that the callbacks API that I proposed in RFC0013 was less flexible than I had hoped. I considered extending it but that felt like the wrong answer. This RFC proposes a new body-wrapping API and shows how I'd implement a `ChecksumBody` that would wrap an `SdkBody` and provide both trailers and an accurate size hint for setting `Content-Length` headers.


Velfi:
I'm still not exactly sure how to correctly size the trailers but I'm looking into it and expect my current approach to work, with a little tweaking.

design/src/rfcs/rfc0016_body_callback_apis_v2.md
```diff
@@ -0,0 +1,389 @@
+RFC: Replacing Callback APIs for `ByteStream` and `SdkBody` with a body-wrapping API
+=================================================
+
+> Status: RFC
+
+Adding a body-wrapping API to `ByteStream` and `SdkBody` will enable developers using the SDK to implement things like checksum validations and 'read progress' callbacks. This RFC supersedes [RFC0013] and will partially replace the changes made for that RFC.
```
jdisanti, line 6: Seems like we should update the status line in RFC 13 to say something like "Superceded by X"

design/src/rfcs/rfc0016_body_callback_apis_v2.md
```diff
@@ -0,0 +1,389 @@
...
+        match &self.0 {
+            Inner::Sha256(callback) => callback.trailers(),
+            Inner::Crc32c(callback) => callback.trailers(),
+            Inner::Crc32(callback) => callback.trailers(),
+            Inner::Sha1(callback) => callback.trailers(),
+        }
+    }
+
+    // TODO I think I still need to account for the size of the trailer names.
+    //      Should I be doing that here?
```
jdisanti, line 235: I think so. I would change the interface on the individual checksum callbacks something like this:
```rust
/// Returns the size in bytes of the actual checksum value
fn checksum_size_hint(&self) -> usize;

/// Returns the checksum header name
fn checksum_header_name(&self) -> HeaderName;

/// Returns the checksum header value
fn checksum_value(&self) -> HeaderValue;
```

The `checksum_value` would only get called once when trailers are needed, but `checksum_header_name` could be called both by trailers and the size hint computation. I think it's safe to make assumptions about the interface here since these are private.

design/src/rfcs/rfc0016_body_callback_apis_v2.md
```diff
@@ -0,0 +1,389 @@
...
+                Poll::Ready(Some(Ok(bytes)))
+            }
+            Poll::Ready(None) => Poll::Ready(None),
+            Poll::Ready(Some(Err(e))) => Poll::Ready(Some(Err(e))),
+            Poll::Pending => Poll::Pending,
+        }
+    }
+}
+
+pub type Error = Box<dyn StdError + Send + Sync>;
```
jdisanti, line 310: SdkBody's `Error` type is public, so you could potentially reuse it directly: https://github.com/awslabs/smithy-rs/blob/e1e9a29d5db1330eb51c1e84e5c6c3acb9b4f65e/rust-runtime/aws-smithy-http/src/body.rs#L19

design/src/rfcs/rfc0016_body_callback_apis_v2.md
```diff
@@ -0,0 +1,389 @@
...
+            checksum_callback: ChecksumCallback::new(checksum_algorithm),
+            inner: body,
+        }
+    }
+
+    // TODO a more intelligent person could prob impl this with generics
+    fn poll_inner(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Bytes, Error>>> {
```
jdisanti, line 287: What would be gained by using generics for this function?

jdisanti (commented):
Overall, it seems like this approach should work well. Nice thinking!

design/src/rfcs/rfc0016_body_callback_apis_v2.md
Velfi: Then it would support other body types by default. I also considered parameterizing this on the `http_body::Body` trait so that it'd be more generic. If y'all aren't concerned then I won't be either.

After thinking about it, I think my understanding/question is:
"We make wrapper structs generic so that we can support other wrapped types in the future. Is it better to implement the wrappers with concrete types as necessary or should we always seek to parameterize on traits that wrapped types must implement?"

jdisanti: My thinking on this is that `SdkBody` isn't generic, so the thing that wraps `SdkBody` doesn't need to be generic. Given how infectious generics are in Rust, I'm inclined to index on using less of them rather than making things overly generic without a concrete use-case.

design/src/rfcs/rfc0016_body_callback_apis_v2.md
```diff
@@ -58,6 +58,9 @@ const CRC_32_C_HEADER_NAME: &str = "x-amz-checksum-crc32c";
 const SHA_1_HEADER_NAME: &str = "x-amz-checksum-sha1";
 const SHA_256_HEADER_NAME: &str = "x-amz-checksum-sha256";
 
+const WITH_OPTIONAL_WHITESPACE: bool = true;
+const TRAILER_SEPARATOR: &str = if WITH_OPTIONAL_WHITESPACE { ": " } else { ":" };
```
jdisanti, line 62: Is this configurable in the conversion of `http::Request` into the actual HTTP request headers?

design/src/rfcs/rfc0016_body_callback_apis_v2.md
```diff
@@ -74,21 +77,26 @@ impl Crc32Callback {
...
-        header_map.insert(key, value);
+        header_map.insert(Self::header_name(), self.header_value());
 
         Ok(Some(header_map))
     }
 
     // Size of the checksum in bytes
-    fn size() -> u64 {
+    fn size() -> usize {
         4
```
jdisanti, line 87: I think this number is not actually accurate (should be 8?) since you need the length of the base64 encoding of the final number. Do we know for certain that that encoded version of it will always be the same length? I _think_ it is due to the padding, but I'm not familiar enough off the top of my head to say with certainty.

design/src/rfcs/rfc0016_body_callback_apis_v2.md
```diff
@@ -108,21 +116,26 @@ impl Crc32cCallback {
 
     fn trailers(&self) -> Result<Option<HeaderMap<HeaderValue>>, BoxError> {
```
jdisanti, line 117: Seems like you could provide a default implementation for this function on the trait since it's the same for all four now.

design/src/rfcs/rfc0016_body_callback_apis_v2.md
Velfi: I think that makes a lot of sense. 👍

design/src/rfcs/rfc0016_body_callback_apis_v2.md
```diff
@@ -231,19 +254,24 @@ impl ChecksumCallback {
...
-        let checksum_size_in_bytes = match &self.0 {
-            Inner::Crc32(_) => Crc32Callback::size(),
-            Inner::Crc32c(_) => Crc32cCallback::size(),
-            Inner::Sha1(_) => Sha1Callback::size(),
-            Inner::Sha256(_) => Sha256Callback::size(),
+        let (trailer_name_size_in_bytes, checksum_size_in_bytes) = match &self.0 {
+            // We want to get the size of the actual `HeaderName` except those don't have a `.len()`
+            // method so we'd have to convert back to the original string. That's why we're getting
+            // `.len()` from the original strings here.
+            Inner::Crc32(_) => (CRC_32_HEADER_NAME.len(), Crc32Callback::size()),
```
jdisanti, line 262: It seems like these should use the `header_name` function on the callback. Perhaps this could be simplified more if a trait is added for callback, and `Inner` has functions `callback` and `callback_mut` that return `&dyn Callback` and `&mut dyn Callback` respectively. With the trait, you could keep the switching logic limited to the constructor and to these function impls in `Inner`, rather than in every place the callback is referenced.

design/src/rfcs/rfc0016_body_callback_apis_v2.md
Velfi: It's not configurable and I don't think it will need to be. I'm expecting that Hyper decides whether or not to include the OWS and that it never changes but I have to go check. I put this code in there so that I don't forget to verify this.

Velfi:
I'm closing this for now. I'm going to publish a checksum implementation RFC that doesn't specifically require us to define a generic body-wrapping API.

I still think that switching out the `BodyCallback` api for a wrapping one will be better in the long run and I may write an RFC to that effect in the future.

PR #1396, opened 2022-05-18
Remove individual devs from crate ownership in crates.io

jdisanti:
## Motivation and Context
Before this PR, our crates in crates.io have two owners: `awslabs/rust-sdk-owners`, and whoever published the crate for the very first time from the time when crate publication wasn't automated. This PR updates the publisher tool to remove any owners that are not `awslabs/rust-sdk-owners`, which should fix the crate ownership for all crates with the next release.


Velfi:
We generally don't leave comments in the code or publish info to our design docs on why these tools exist. Are we doing a disservice to future maintainers or will the need for these tools still be obvious?

jdisanti:
> We generally don't leave comments in the code or publish info to our design docs on why these tools exist

I don't think this assertion is really true. We have [READMEs](https://github.com/awslabs/smithy-rs/blob/main/tools/sdk-versioner/README.md) [and](https://github.com/awslabs/smithy-rs/blob/main/tools/sdk-sync/src/main.rs#L32-L33) [comments](https://github.com/awslabs/smithy-rs/blob/main/tools/publisher/src/subcommand/fix_manifests.rs#L6-L10) [in](https://github.com/awslabs/smithy-rs/blob/main/tools/ci-build/README.md) our [tools](https://github.com/awslabs/smithy-rs/blob/main/tools/api-linter/README.md), and several of the tools origins are talked about in [design](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0007_split_release_process.md) [docs](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0011_crates_io_alpha_publishing.md).

All of this can certainly be improved and expanded on, and a couple of the tools are missing docs outright since they were written in a time when we had zero tools and didn't think we would have so many later on. I also think a central summary of the tools would be useful. These are all things I want to work on as we get closer to closing out release automation.

I'm not really sure what you're alluding to for this specific PR as this is just augmenting the existing crate ownership check that is a part of the publish process. Was there something you had in mind?

Velfi:
@jdisanti I'm sorry, I should have chosen my words more carefully. I was just thinking about providing some central repository with info on what all this stuff does and why it exists. I think this PR is good.

PR #1397, opened 2022-05-18
Remove TODO about required `httpHeader`-bound collection shapes

david-perez:
We do support required `httpHeader`-bound collection shapes, as
evidenced by the test operation this commit adds, which is generated by
the server, that does not use `Option` for `required` member shapes.

This is because even though the deserialization always returns an
optional collection, we return `None` if no headers are found (as
opposed to an empty collection), so the call that sets the member field
on the operation structure does nothing; when `build()` gets called, a
`BuildError` is raised because the member is required.

## Testing

See added test.

### DIFF
--- a/codegen-test/model/misc.smithy
+++ b/codegen-test/model/misc.smithy
@@ -11,23 +11,26 @@ use aws.protocols#restJson1
 @title("MiscService")
 service MiscService {
     operations: [
-        OperationWithInnerRequiredShape,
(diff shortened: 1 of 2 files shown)


Velfi (approved):

rcoh (approved):

crisidev (approved):

82marbag:
Is the `set` to be removed here (because removed in smithy 2)?

PR #1398, opened 2022-05-18
Unknown enum variants removed from server

### LINKED ISSUE #1187: Server enums should not have `Unknown` variants
When generating enums for the server, we're generating an `Unknown` variant.

https://github.com/awslabs/smithy-rs/blob/548f6ed4307faae5bb98e431db96ae90a59e489d/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/EnumGenerator.kt#L154-L155

Server enums should not contain this variant. This also means we can no longer `impl From<&str> for MyEnum`, since [`From` is required to never fail](https://doc.rust-lang.org/std/convert/trait.From.html). We should instead implement [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html). This also means deserialization code will be affected, since deserializing a `&str` into an enum can now fail.

82marbag:
The server must have the most up to date variants and the unknown enum
variant should not be used. Clients are generated with it because they
might not have the most recent model and the server might return
an unknown variant to them.

Closes #1187

## Motivation and Context

Closes #1187

## Description

* Remove unknown in server
* Implement TryFrom and the relevant Errors for enums
* Refactor the code to use the server/client generators
* Added and refactored tests

## Testing

`./gradlew :codegen-server-test:test `

## Checklist


CHANGELOG.next.toml
```diff
@@ -11,6 +11,12 @@
 # meta = { "breaking" = false, "tada" = false, "bug" = false }
 # author = "rcoh"
 
+[[smithy-rs]]
+message = "Remove Unknown variant from enums in server"
+references = ["smithy-rs#1187"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
+author = "82marbag"
```
crisidev, line 18: We do not publish CHANGELOG entries for server only changes. If this PR does not change how the client SDK is generated, this should go away.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -276,10 +277,17 @@ class HttpBindingGenerator(
                         }
                     }
                     if (targetShape.hasTrait<EnumTrait>()) {
-                        rust(
-                            "Ok(#T::from(body_str))",
-                            symbolProvider.toSymbol(targetShape)
-                        )
+                        if (mode == CodegenMode.Server) {
```
crisidev, line 280: I am not a fan of all these if `CodegenMode.Server` / `CodegenMode.Client` :(

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/EnumGenerator.kt
```diff
@@ -151,8 +155,10 @@ class EnumGenerator(
         meta.render(writer)
         writer.rustBlock("enum $enumName") {
             sortedMembers.forEach { member -> member.render(writer) }
-            docs("$UnknownVariant contains new variants that have been added since this code was generated.")
-            write("$UnknownVariant(String)")
+            if (mode == CodegenMode.Client) {
+                docs("$UnknownVariant contains new variants that have been added since this code was generated.")
+                write("$UnknownVariant(String)")
+            }
```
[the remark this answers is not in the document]
82marbag: Yes, I understand. They are here because the client generates the same code as the server, plus some more. The one where the server is generating actual different code is in the server only. But not to copy/paste code around, I generate the common code in the client and the additional code in the client. Moving to ServerEnumGenerator would mean to copy the same code there anyway. I am open for suggestions

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
82marbag: It's because client and server only change for one line or two every time; otherwise we'd have to create a new class every time (or two) and copy the remaining code. But if that's the best approach for this, I'll go for it

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGenerator.kt
```diff
@@ -0,0 +1,61 @@
...
+    override fun renderFromForStr() {
+        val errorStruct = "${enumName}UnknownVariantError"
+        writer.write("##[derive(Debug, PartialEq, Eq, Hash)]")
+        writer.write("pub struct $errorStruct(String);")
+        writer.rustBlock("impl #T<&str> for $enumName", RuntimeType.TryFrom) {
+            write("type Error = $errorStruct;")
+            writer.rustBlock("fn try_from(s: &str) -> Result<Self, <$enumName as #T<&str>>::Error>", RuntimeType.TryFrom) {
+                writer.rustBlock("match s") {
+                    sortedMembers.forEach { member ->
+                        write("""${member.value.dq()} => Ok($enumName::${member.derivedName()}),""")
```
david-perez, line 38: `dq` is useful when your enclosing string is in double quotes:

```suggestion
                        write("${member.value.dq()} => Ok($enumName::${member.derivedName()}),")
```

When using `"""`, you can write double quotes directly:

```suggestion
                        write(""""${member.value}" => Ok($enumName::${member.derivedName()}),""")
```

But using `"""` _and_ `dq` is noisy. I notice this has been copied from `EnumGenerator`, we should make the improvement there too.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGenerator.kt
```diff
@@ -0,0 +1,61 @@
...
+                    write("_ => Err($errorStruct(s.to_owned()))")
+                }
+            }
+        }
+        writer.rustBlock("impl #T<$errorStruct> for #T", RuntimeType.From, ServerRuntimeType.RequestRejection(runtimeConfig)) {
+            writer.rustBlock("fn from(e: $errorStruct) -> Self") {
+                write("Self::EnumVariantNotFound(Box::new(e))")
+            }
+        }
+        writer.rustBlock("impl From<$errorStruct> for #T", RuntimeType.jsonDeserialize(runtimeConfig)) {
```
david-perez, line 49: Either write out the fully namespaced `std::convert::From`, or use `RuntimeType.From` like you did in line 44 which does the namespacing for you.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGenerator.kt
```diff
@@ -0,0 +1,61 @@
...
+                write("""Self::custom(format!("unknown variant {}", e))""")
+            }
+        }
+        writer.rustBlock("impl std::error::Error for $errorStruct") {}
+        writer.rustBlock("impl std::fmt::Display for $errorStruct") {
+            writer.rustBlock("fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result") {
+                write("self.0.fmt(f)")
+            }
+        }
+    }
```
david-perez, line 60: Elaborating on the "strive to write as much uninterrupted Rust as possible", this code is fine, but notice it's exactly the same as:

```suggestion
        writer.rustTemplate(
            """
            impl #{From}<$errorStruct> for #{RequestRejection} {
                fn from(e: $errorStruct) -> Self {
                    Self::EnumVariantNotFound(Box::new(e))
                }
            }
            
            impl #{From}<$errorStruct> for #{JsonDeserialize} {
                fn from(e: $errorStruct) -> Self {
                    Self::custom(format!("unknown variant {}", e))
                }
            }
            
            impl #{StdError} for $errorStruct { }
            
            impl std::fmt::Display for $errorStruct {
                fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
                    self.0.fmt(f)
                }
            }
            """,
            "From" to RuntimeType.From,
            "StdError" to RuntimeType.StdError,
            "RequestRejection" to ServerRuntimeType.RequestRejection(runtimeConfig),
            "JsonDeserialize" to RuntimeType.jsonDeserialize(runtimeConfig),
        )
```

which reads much nicer.

(Bonus points: we should also have a `RuntimeType.Display` and use it everywhere in the codebase where we've used `RuntimeType.stdfmt.member("Display")` or written out `Display` manually).

The `writer.*block` methods are useful when you have to loop over a collection, when you want to insert blocks conditionally, or when you need to put conditionals based on Kotlin variables inside your Rust blocks. For example, above was a good use of `rustBlock`, because you needed to iterate over `sortedMembers`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -858,12 +862,22 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
+                                        let v = <#{memberShape}>::try_from(#{PercentEncoding}::percent_decode_str(&v).decode_utf8()?.as_ref())?;
+                                        """,
+                                        *codegenScope,
+                                        "memberShape" to symbolProvider.toSymbol(memberShape),
+                                    )
+                                } else {
+                                    rustTemplate(
+                                        """
+                                        let v = <_>::from(#{PercentEncoding}::percent_decode_str(&v).decode_utf8()?.as_ref());
+                                        """.trimIndent(),
```
david-perez, line 877: nit: you'll sometimes see these `trimIndent` sprinkled over the codebase. They are added by IntelliJ when hitting Enter inside a string block. We decided some time ago that we would not insert these anymore when generating Rust code, since we run `cargo fmt` after generation, so they just add noise. If you touch a part of the codebase where you see one of these, feel free to remove it.

I guess we could also just grep them out in one commit and forget about them, but I haven't gotten around to doing that just yet. Feel free to take it on if you want.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -993,12 +1008,28 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
+                when (type.rustType()) {
+                    RustType.String ->
+                        rustTemplate(
+                            """
+                            let value = <#{T}>::from(#{PercentEncoding}::percent_decode_str(value).decode_utf8()?.as_ref());
+                            """,
+                            *codegenScope,
+                            "T" to type,
+                        )
+                    else -> // RustType.Opaque, the Enum
```
david-perez, line 1020: Add a `check` assertion to ensure it's `RustType.Opaque`, or use `RustType.Opaque` in this arm and then `throw` an exception in the `else` arm. Always code defensively to avoid introducing future bugs :)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -984,6 +998,7 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
     private fun generateParsePercentEncodedStrAsStringFn(binding: HttpBindingDescriptor): RuntimeType {
         val output = symbolProvider.toSymbol(binding.member)
         val fnName = generateParseStrFnName(binding)
+        val type = output.extractSymbolFromOption()
```
david-perez, line 1001: nit: this is technically a `Symbol`, not a `RustType`, so `symbol` would be a more appropriate variable name.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGeneratorTest.kt
```diff
@@ -0,0 +1,222 @@
...
+            assert_eq!(InstanceType::try_from("t2.nano").unwrap(), InstanceType::T2Nano);
+            // check no unknown
+            match instance {
+            InstanceType::T2Micro => (),
+            InstanceType::T2Nano => (),
+            }
+            """
+        )
+
+        writer.toString() shouldContain "#[non_exhaustive]"
```
david-perez, line 60: Now that the `Unknown` variant is no longer there, we also don't need to generate `#[non_exhaustive]` on server enums. The server always has the latest version of the model; this will allow users to exhaustively `match` on their enums.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGeneratorTest.kt
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Test
+    fun `named enums are implement eq and hash`() {
```
david-perez, line 64:
```suggestion
    fun `named enums implement Eq and Hash`() {
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGeneratorTest.kt
```diff
@@ -0,0 +1,222 @@
...
+            assert_eq!(FooEnum::Foo, FooEnum::Foo);
+            assert_ne!(FooEnum::Bar, FooEnum::Foo);
+            let mut hash_of_enums = std::collections::HashSet::new();
+            hash_of_enums.insert(FooEnum::Foo);
+            """.trimIndent()
+        )
+    }
+
+    @Test
+    fun `unnamed enums are implement eq and hash`() {
```
david-perez, line 94:
```suggestion
    fun `unnamed enums are implement Eq and Hash`() {
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGeneratorTest.kt
```diff
@@ -0,0 +1,222 @@
...
+            @enum([
+            {
+                value: "Foo",
+            },
+            {
+                value: "Bar",
+            }])
+            string FooEnum
+        """.asSmithyModel()
+        val shape: StringShape = model.lookup("test#FooEnum")
```
david-perez, line 106:
```suggestion
        val shape = model.lookup<StringShape>("test#FooEnum")
```

Just to be consistent.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGeneratorTest.kt
```diff
@@ -0,0 +1,222 @@
...
+        val trait = shape.expectTrait<EnumTrait>()
+        val writer = RustWriter.forModule("model")
+        val generator = ServerEnumGenerator(model, testSymbolProvider(model), writer, shape, trait, CodegenMode.Server, TestRuntimeConfig)
+        generator.render()
+        writer.compileAndTest(
+            """
+            assert_eq!(FooEnum::try_from("Foo").unwrap(), FooEnum::try_from("Foo").unwrap());
+            assert_ne!(FooEnum::try_from("Bar").unwrap(), FooEnum::try_from("Foo").unwrap());
+            let mut hash_of_enums = std::collections::HashSet::new();
```
david-perez, line 115:
```suggestion
            let mut hash_set_of_enums = std::collections::HashSet::new();
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGeneratorTest.kt
```diff
@@ -0,0 +1,222 @@
...
+import software.amazon.smithy.rust.codegen.rustlang.RustWriter
+import software.amazon.smithy.rust.codegen.smithy.CodegenMode
+import software.amazon.smithy.rust.codegen.testutil.TestRuntimeConfig
+import software.amazon.smithy.rust.codegen.testutil.asSmithyModel
+import software.amazon.smithy.rust.codegen.testutil.compileAndTest
+import software.amazon.smithy.rust.codegen.testutil.testSymbolProvider
+import software.amazon.smithy.rust.codegen.util.expectTrait
+import software.amazon.smithy.rust.codegen.util.lookup
+
+class ServerEnumGeneratorTest {
```
david-perez, line 21: All of these tests are copied from `EnumGeneratorTest`. I personally would remove them and only have tests for the server-specific bits (e.g. no `Unknown` variant, no `#[non_exhaustive]`, the `TryFrom` impl etc). `EnumGenerator` is effectively a dependency of `ServerEnumGenerator`. We do not have to unit test our dependencies.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
```diff
@@ -406,6 +406,14 @@ fun Symbol.isOptional(): Boolean = when (this.rustType()) {
     else -> false
 }
 
+/**
+ * Get the referenced symbol for T if [this] is an Option<T>, [this] otherwise
+ */
+fun Symbol.extractSymbolFromOption(): Symbol = when (this.isOptional()) {
```
david-perez, line 412: Use:

```
symbol.mapRustType { it.stripOuter<RustType.Option>() }
```

for this.

Symbol `references` should not be relied upon.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/EnumGenerator.kt
```diff
@@ -189,14 +197,16 @@ class EnumGenerator(
...
 
+    private fun renderFromStr() {
         writer.rust(
             """
             impl std::str::FromStr for $enumName {
                 type Err = std::convert::Infallible;
 
                 fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
-                    Ok($enumName::from(s))
+                    Ok($enumName::try_from(s).unwrap())
```
david-perez, line 209: `EnumGenerator` should use `from`, not `try_from`. This will cause the client to panic.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -238,14 +239,26 @@ class JsonParserGenerator(
...
+                    }
+                }
                 else -> rust("u.into_owned()")
             }
         }
     }
 
+    private fun convertsToEnumInServer(shape: StringShape): Boolean {
+        return mode == CodegenMode.Server && shape.hasTrait<EnumTrait>()
+    }
```
david-perez, line 256:
```suggestion
    private fun convertsToEnumInServer(shape: StringShape) =
        mode == CodegenMode.Server && shape.hasTrait<EnumTrait>()

```

codegen/src/test/kotlin/software/amazon/smithy/rust/testutil/Annotations.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.jupiter.params.support.AnnotationConsumer
+import software.amazon.smithy.rust.codegen.smithy.CodegenMode
+import java.util.stream.Stream
+
+@Target(AnnotationTarget.TYPE, AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.RUNTIME)
+@ArgumentsSource(CodegenModeProvider::class)
+annotation class CodegenModeSource
+
+class CodegenModeProvider : ArgumentsProvider, AnnotationConsumer<CodegenModeSource> {
```
david-perez, line 21: Where is this used?

rust-runtime/aws-smithy-http-server/src/rejection.rs
```diff
@@ -185,6 +185,8 @@ pub enum RequestRejection {
     // error, but it would be a lot of effort for comparatively low benefit.
     /// Used when consuming the input struct builder.
     Build(crate::Error),
+
+    EnumVariantNotFound(Box<dyn std::error::Error + Send + Sync>),
```
david-perez, line 189: Add documentation. Mention this is the only one that does not take in `crate::Error` because the inner type is code-generated, unlike the other ones.

david-perez (changes requested):

rust-runtime/aws-smithy-http-server/src/rejection.rs
```diff
@@ -185,6 +185,11 @@ pub enum RequestRejection {
     // error, but it would be a lot of effort for comparatively low benefit.
     /// Used when consuming the input struct builder.
     Build(crate::Error),
+
+    // Used by the server when the enum variant sent by a client is not known
+    // Unlike the rejections above, the inner type is code generated,
+    // with each enum having its own generated error type
```
david-perez, line 191:
```suggestion
    /// Used by the server when the enum variant sent by a client is not known.
    /// Unlike the rejections above, the inner type is code generated,
    /// with each enum having its own generated error type.
```

codegen/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/XmlBindingTraitParserGeneratorTest.kt
```diff
@@ -89,7 +89,7 @@ internal class XmlBindingTraitParserGeneratorTest {
     """.asSmithyModel()
 
     @Test
-    fun `generates valid parsers`() {
+    fun `generates valid parsers in client`() {
```
david-perez, line 92: Why this change?

david-perez (changes requested):
An empty `rust-runtime/.attach_pid18283` file crept into the patchset.

codegen/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/XmlBindingTraitParserGeneratorTest.kt
82marbag: In revision 0, I had two completely independent tests. Sorry this remained from that revision

david-perez (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/Instantiator.kt
```diff
@@ -252,7 +254,11 @@ class Instantiator(
             writer.rust("$data.to_string()")
         } else {
             val enumSymbol = symbolProvider.toSymbol(shape)
-            writer.rust("#T::from($data)", enumSymbol)
+            if (mode == CodegenMode.Server) {
+                writer.rust("""#T::try_from($data).expect("This is used in tests ONLY")""", enumSymbol)
+            } else {
+                writer.rust("#T::from($data)", enumSymbol)
+            }
```
Velfi, line 261: I think we should prefer `when` statements to `if` statements when differentiating between client and server codegen.
- It's very explicit about what cases are being covered
- I don't expect us to ever add a third mode, but writing things this way would force us to update things if we ever did
- It also saves one line of code

```suggestion
            when (mode) {
                 CodegenMode.Server -> writer.rust("""#T::try_from($data).expect("This is used in tests ONLY")""", enumSymbol)
                 CodegenMode.Client -> writer.rust("#T::from($data)", enumSymbol)
            }
```

Issue #1367, opened 2022-05-05
[WIP] Generate server implementation for pure Python business logic


crisidev:
## Motivation and Context
[Pyo3](https://github.com/PyO3/pyo3) will allow us to generate an idiomatic Python implementation which is exported from Rust as a shared library. 

The framework heavy-lifting will be handled completely in Rust, while the business logic will be implemented in pure Python.

### Example

```python
# Import the shared library application, context and input/output types.
from pokemon_service import PokemonService, SharedContext
from pokemon_service.model import PokemonDescriptionInput, PokemonDescriptionOutput, AuthError
from pokemon_service.http import Request
# import application specific code.
from .database import DatabasePool

# Define a context class to be shared between handlers
@dataclass
class Context:
    db = DatabasePool()
 
# Create the application and register context.   
app = PokemonService()
app.context(Context())

# Define a middleware function.
@app.middleware
async def auth_middleware(request: Request):
    if request.headers.get("a-token") != "i_am_so_secret":
        raise AuthError("Invalid a-token")

# Define the operation using the type imported from Rust.
@app.pokemon_description
async def pokemon_description(input: PokemonDescriptionInput, ctx: SharedContext) -> PokemonDescriptionOutput:
    description = await ctx.db.get_description(input.name)
    return PokemonDescriptionOutput(description)

# Start the application. 
app.run()
```

## Features
- [ ] Python subproject scaffolding.
- [ ] Code sharing between :codegen, :codegen-server and :codegen-server:python Gradle projects.
- [ ] Object model for input / outputs / errors.
- [ ] Operation wrappers to execute Python code inside Rust.
- [ ] Shared state/context support.
- [ ] Support for both asyncio and traditional Python code.
- [ ] Support spawning multiple interpreters using a shared socket.
- [ ] Handle Python exceptions inside infallible operations.
- [ ] Inject [tracing](https://docs.rs/tracing/latest/tracing/) into the Python logger system.

### Tentative
These features are hard and we are not sure if it will land in the first release.

- [ ] Support custom middlewares.
- [ ] Support type hinting.

PR #1399, opened 2022-05-18
Implement Python runtime crate and shared socket

crisidev:
## Motivation and Context
Continue on #1367 

## Description
* Add the Python runtime crate.
* Implement `SharedSocket` support.
* Implement `logging` to `tracing` support.


rcoh (approved):

rust-runtime/aws-smithy-http-server-python/rustfmt.toml
```diff
@@ -0,0 +1,4 @@
+edition = "2021"
+max_width = 120
+# Prevent carriage returns
+newline_style = "Unix"
```
jdisanti, line 4: I'm kind of surprised we haven't had any issues with the `aws-smithy-server` crate failing lints where rustfmt formatted it properly according to this config (see https://github.com/awslabs/smithy-rs/blob/a0539e20b069a7de021c84521d8f3c7ba098ad6d/tools/ci-build/scripts/check-style-and-lints#L14-L13) since the default `max_width` is 100. Or maybe we have and I just haven't noticed.

I'm completely onboard with making these settings apply to all of our Rust crates if you want. We should probably be consistent about it though to keep CI simpler.

No need to block on this--just be aware that CI may give you some grief while it remains inconsistent.

jdisanti (approved):

rust-runtime/aws-smithy-http-server-python/rustfmt.toml
crisidev: I am happy to remove this and make it consistent. I am a fan of 120 chars, but I am ok also with 100 :)

82marbag (approved):

rust-runtime/aws-smithy-http-server-python/rustfmt.toml
jdisanti: I'm also a fan of 120, so either way

rust-runtime/aws-smithy-http-server-python/README.md
```diff
@@ -0,0 +1,7 @@
+# aws-smithy-http-server-python
+
+Server libraries for smithy-rs generated servers, targeting pure Python business logic.
+
+<!-- anchor_start:footer -->
+This crate is part of the [AWS SDK for Rust](https://awslabs.github.io/aws-sdk-rust/) and the [smithy-rs](https://github.com/awslabs/smithy-rs) code generator. In most cases, it should not be used directly.
```
david-perez, line 6: Not part of the AWS Rust SDK. We have the same phrase in `aws-smithy-http-server`.

david-perez (approved):

PR #1404, opened 2022-05-23
Use md5 implementation of RustCrypto organization

petrosagg:
## Motivation and Context

I noticed that the S3 SDK was depending to the `md5` crate which is not part of the `RustCrypto` organization. The `RustCrypto` organization is the de-facto standard of cryptographic primitive implementations in Rust with an active community and already used by `aws-smithy-checksums` for its `sha1` and `sha2` checksums.

## Description

I changed the `md5` crate to the one provided by `RustCrypto`. 

## Testing

No behavior changes to be tested

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -28,3 +28,21 @@ author = "alonlud"
  references = ["smithy-rs#1390"]
  meta = { "breaking" = false, "tada" = true, "bug" = false }
  author = "Velfi"
+
+[[smithy-rs]]
+message = "Add ability to specify a different rust crate name than the one derived from the package name"
+references = ["smithy-rs#1404"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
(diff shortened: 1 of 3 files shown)


jdisanti:
It looks like the `md-5` crate renames itself to `md5`:
https://github.com/RustCrypto/hashes/blob/master/md5/Cargo.toml#L14-L15

Our codegen doesn't currently support that concept, so I think this will require adding a property to `CargoDependency` and updating either the `CargoTomlGenerator` to re-rename the crate, or the `RustWriter` to use the different name.

petrosagg:
@jdisanti I see, I pushed a change that should fix that, let's see what CI thinks about it

petrosagg:
@jdisanti oh, great! Thanks for looking into it, it would have been a deep rabbit hole to set up a local devenv for me, I've never worked with Java nor Kotlin before :)

Issue #1229, opened 2022-03-02
Problems with rendering of `httpResponseCode` binding


david-perez:
There are two problems with the code that handles this trait:

1. We're not falling back to the `http` trait if the targetted member is optional and no value is provided by the service implementer. See https://github.com/awslabs/smithy/issues/1098#issuecomment-1051019501.
1. It only works when the targetted member is optional.

PR #1405, opened 2022-05-24
Correctly generate http response codes

82marbag:
http response codes will be:
* @httpResponseCode if set
* @http's code, if @httpResponseCode is not set
* the default one, if neither of the above are set

@httpResponseCode honors @required and generates code that
doesn't use Option when applied.

Closes: #1229

## Description

* set http response status to using Option only when not `@required`
* default to `@http`'s code when the status is not set
* default to the builder's default if none is set
* add tests

## Testing

```
./gradlew :codegen-server-test:test && ./gradlew :codegen-test:test
```
## Checklist


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -552,24 +561,35 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
     }
 
     private fun serverRenderResponseCodeBinding(
```
david-perez, line 563: I'd create two methods, one to set the code from the HTTP binding, one to set it to a default code. (Instead of having a method that is checking that at least one of the params is not null, when only at most one is going to be not null)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -552,24 +561,35 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
-            // 1. We're not falling back to the `http` trait if no `output.$memberName` is `None`.
-            // 2. It only works when `output.$memberName` is of type `Option<i32>`.
+            binding?.let {
+                val memberName = symbolProvider.toMemberName(binding.member)
+                rust("let status = output.$memberName")
+                if (binding.member.isOptional) {
+                    rustTemplate(
+                        """
+                        .ok_or(#{ResponseRejection}::MissingHttpStatusCode)?
+                        """.trimIndent(),
```
david-perez, line 577:
```suggestion
                        """,
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -552,24 +561,35 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
+
         return writable {
-            val memberName = symbolProvider.toMemberName(binding.member)
-            // TODO(https://github.com/awslabs/smithy-rs/issues/1229): This code is problematic for two reasons:
-            // 1. We're not falling back to the `http` trait if no `output.$memberName` is `None`.
-            // 2. It only works when `output.$memberName` is of type `Option<i32>`.
+            binding?.let {
+                val memberName = symbolProvider.toMemberName(binding.member)
+                rust("let status = output.$memberName")
+                if (binding.member.isOptional) {
```
david-perez, line 573: Always go through the symbol provider to determine if you're working with an `Option` or not i.e. use `symbolProvider.toSymbol(binding.member).isOptional()` here.

This is because the fact that a shape is optional in the model does not mean that it will get rendered as an `Option` in Rust. An example: https://github.com/awslabs/smithy/issues/1179

codegen-test/model/misc.smithy
```diff
@@ -108,3 +112,67 @@ union AUnion {
...
+}
+
+@input
+structure ResponseCodeRequiredInput {}
+
+@output
+structure ResponseCodeRequiredOutput {
+    @required
+    @httpResponseCode
+    responseCode: HttpResponseCode,
```
david-perez, line 175:
```suggestion
    responseCode: Integer,
```

You can just use the shape in the prelude to not have to define `HttpResponseCode`.

codegen-test/model/misc.smithy
```diff
@@ -108,3 +112,67 @@ union AUnion {
...
+@input
+structure ResponseCodeDefaultInput {}
+
+@output
+structure ResponseCodeDefaultOutput {}
+
+/// This operation tests that @httpResponseCode is @required
+/// and is used over @http's code
+@http(method: "GET", uri: "/responseCodeRequired", code: 200)
+@httpRequestTests([
```
david-perez, line 147: This should be an `httpResponseTests` (add it to the array below). There are no response codes in HTTP requests.

codegen-test/model/misc.smithy
```diff
@@ -108,3 +112,67 @@ union AUnion {
     string: String,
     time: Timestamp,
 }
+
+/// This operation tests that the response code defaults to @http's code
+@http(method: "GET", uri: "/responseCodeDefault", code: 418)
+@httpRequestTests([
```
david-perez, line 118: nit: always put `httpRequestTests/httpResponseTests` above all other traits.

david-perez: But this should be an `httpResponseTests`.

codegen-test/model/misc.smithy
```diff
@@ -12,6 +14,8 @@ use aws.protocols#restJson1
 service MiscService {
     operations: [
         OperationWithInnerRequiredShape,
+        ResponseCodeRequired,
+        ResponseCodeDefault,
```
david-perez, line 18: We should also test that when neither `http.code` nor `httpResponseCode` are provided, we default to 200.

codegen-test/model/misc.smithy
david-perez: https://awslabs.github.io/smithy/1.0/spec/core/prelude-model.html

david-perez (changes requested):

codegen-test/model/misc.smithy
```diff
@@ -108,3 +113,69 @@ union AUnion {
     string: String,
     time: Timestamp,
 }
+
+/// This operation tests that the response code defaults to 200 when no other code is set
+@httpResponseTests([
+    {
+        id: "ResponseCodeDefaultTest",
```
david-perez, line 120: nit: there's no need to suffix these IDs with `Test`.

codegen-test/model/misc.smithy
```diff
@@ -108,3 +113,69 @@ union AUnion {
...
+}
+
+@input
+structure ResponseCodeHttpFallbackInput {}
+
+@output
+structure ResponseCodeHttpFallbackOutput {}
+
+/// This operation tests that @httpResponseCode is @required
+/// and is used over @http's code
```
david-perez, line 158: Not a good test, because `http.code` is `418`, so you wouldn't know if it's using `params.responseCode` or falling back to `http.code`.

david-perez (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -551,25 +561,47 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
+                    .ok_or(#{ResponseRejection}::MissingHttpStatusCode)?
+                    """.trimIndent(),
+                    *codegenScope,
+                )
+            }
             rustTemplate(
                 """
-                let status = output.$memberName
-                    .ok_or(#{ResponseRejection}::MissingHttpStatusCode)?;
+                ;
```
rcoh, line 598: sometimes cleaner ways to generate code that handles this, usually by wrapping in a match instead of chaining with OK or, but this is fine

rcoh (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
82marbag: `status` might not be an Option. I'll keep your suggestion in mind, thanks

PR #1407, opened 2022-05-25
Implement release-aware yank tooling

jdisanti:
## Motivation and Context
This is a prerequisite for partial releases and releases with unsynchronized version numbers. Once we start doing those, the new `[release]` section of `versions.toml` (introduced in #1400) will need to be used to yank a release. This PR reworks the existing yank tool to take a release tag or `versions.toml` file path to yank a release rather than the old crate category and repository approach.


tools/publisher/src/subcommand/yank_release.rs
```diff
@@ -0,0 +1,160 @@
...
+
+const MAX_CONCURRENCY: usize = 5;
+
+#[derive(Parser, Debug)]
+pub struct YankReleaseArgs {
+    /// The aws-sdk-rust release tag to yank
+    #[clap(long, required_unless_present = "versions-toml")]
+    release_tag: Option<String>,
+    /// Path to a `versions.toml` file with a `[release]` section to yank.
+    /// The `--release-tag` option is preferred to this, but this is provided as a fail safe.
```
rcoh, line 27: this is a little confusing—this is a _git_ tag but that's different that the release _version_

I think this can be fixed up with documentation and maybe naming. Something like `--github-tag` maybe? or `--git-tag`? 

As written I briefly thought it would yank all crates that had a specified cargo version

tools/publisher/src/subcommand/yank_release.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::str::FromStr;
+use std::sync::Arc;
+use tokio::sync::Semaphore;
+use tracing::info;
+
+const MAX_CONCURRENCY: usize = 5;
+
+#[derive(Parser, Debug)]
+pub struct YankReleaseArgs {
+    /// The aws-sdk-rust release tag to yank
```
rcoh, line 23: docs should say something like: "this will download `versions.toml` from GitHub at the tagged release and yank crates that had that version" or something

tools/publisher/src/subcommand/yank_release.rs
```diff
@@ -0,0 +1,160 @@
...
+    );
+
+    let mut tasks = Vec::new();
+    for (crate_name, crate_version) in release.crates {
+        let permit = semaphore.clone().acquire_owned().await.unwrap();
+        tasks.push(tokio::spawn(async move {
+            info!("Yanking `{}-{}`...", crate_name, crate_version);
+            let result = cargo::Yank::new(&crate_name, &crate_version).spawn().await;
+            drop(permit);
+            info!("Successfully yanked `{}-{}`", crate_name, crate_version);
```
rcoh, line 66: this log line isn't correct I think? if the yank fails, we'll still see success here? probably need to use `result`

rcoh (approved):
LGTM—some small docs / logging issues noted

tools/publisher/src/subcommand/yank_release.rs
jdisanti: Good catch!

PR #1408, opened 2022-05-25
Add flag to bypass confirmation prompt for crate publish

jdisanti:
## Motivation and Context
Crate publish automation needs a way to bypass the confirmation prompt.

### DIFF
--- a/tools/publisher/src/subcommand/publish.rs
+++ b/tools/publisher/src/subcommand/publish.rs
@@ -37,9 +37,18 @@ pub struct PublishArgs {
     /// Path containing the crates to publish. Crates will be discovered recursively
     #[clap(long)]
     location: PathBuf,
+
+    /// Don't prompt for confirmation before publishing
+    #[clap(short('y'))]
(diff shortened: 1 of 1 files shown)


tools/publisher/src/subcommand/publish.rs
```diff
@@ -114,7 +123,7 @@ async fn publish(handle: &PackageHandle, crate_path: &Path) -> Result<()> {
 
 async fn confirm_correct_tag(location: &Path) -> Result<()> {
     let repository = Repository::new(SDK_REPO_NAME, location)?;
-    let versions_manifest = VersionsManifest::from_file(location.join("versions.toml"))?;
+    let versions_manifest = VersionsManifest::from_file(location.join("../versions.toml"))?;
```
jdisanti, line 126: This was a bug. It was looking for `versions.toml` in `aws-sdk-rust/sdk` rather than in the repository root since the location is from the resolved publish root, which is the `sdk` directory.

rcoh (approved):

Issue #1394, opened 2022-05-17
Server fails with `httpPrefixHeaders` on `required` member shape


david-perez:
Minimal reproducer:

```smithy
$version: "1.0"

namespace com.amazonaws.simple

use aws.protocols#restJson1

@restJson1
@title("SimpleService")
service SimpleService {
    operations: [
        AnOperation,
    ],
}

@http(uri: "/operation", method: "GET")
operation AnOperation {
    input: AnOperationInputOutput,
    output: AnOperationInputOutput,
}

structure AnOperationInputOutput {
    @required
    @httpPrefixHeaders("X-Required-")
    requiredHeaderMap: HeaderMap,
}

map HeaderMap {
    key: String,
    value: String,
}
```

Yields:

```
error[E0308]: `?` operator has incompatible types
  --> simple/rust-server-codegen/src/operation_deser.rs:19:13
   |
19 | ...   crate::http_serde::deser_prefix_header_an_operation_an_operation_input_required_header_map(request.headers().ok_or(aws_smithy_http_server::rejection::RequestRejection::HeadersAlreadyExtracted)?)?
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `Option`, found struct `HashMap`
   |
   = note: `?` operator cannot convert from `HashMap<_, _>` to `Option<HashMap<_, _>>`
   = note: expected enum `Option<HashMap<_, _>>`
            found struct `HashMap<_, _>`
help: try wrapping the expression in `futures_util::__private::Some`
   |
19 |             futures_util::__private::Some(crate::http_serde::deser_prefix_header_an_operation_an_operation_input_required_header_map(request.headers().ok_or(aws_smithy_http_server::rejection::RequestRejection::HeadersAlreadyExtracted)?)?)
   |             ++++++++++++++++++++++++++++++                                                                                                                                                                                                   +

error[E0308]: mismatched types
  --> simple/rust-server-codegen/src/http_serde.rs:15:5
   |
4  |   ) -> std::result::Result<
   |  ______-
[...]

PR #1410, opened 2022-05-25
 Allow @required with @httpPrefixHeaders 

82marbag:
The server generates members with @required and @httpPrefixHeaders as if
they were always optional. This commit allows the correct generation with @required.

Closes: #1394 

## Motivation and Context

#1394 

## Testing

Same as in #1394 

```
@http(uri: "/requiredHttpPrefixHeaders", method: "GET")
operation RequiredHttpPrefixHeadersOperation {
    input: RequiredHttpPrefixHeadersInputOutput,
    output: RequiredHttpPrefixHeadersInputOutput,
}

structure RequiredHttpPrefixHeadersInputOutput {
    @httpPrefixHeaders("X-Required-")
    @required
    requiredHeaderMap: RequiredHttpPrefixHeadersMap,
}

map RequiredHttpPrefixHeadersMap {
    key: String,
    value: String,
}
```

## Checklist


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -146,9 +147,14 @@ class HttpBindingGenerator(
                 deserializeFromHeader(model.expectShape(target.value.target), binding.member)
             }
         }
+        val retType = if (outputT.isOptional()) {
+            "#T"
+        } else {
+            "std::option::Option<#T>"
+        }
```
david-perez, line 154: This is correct.

It's also equivalent to:

```
val retType = outputT.mapRustType { it.asOptional() }.rustType().render()
```

However, I think it's best to deal with symbols and have the `CodeWriter` always render them as strings when possible.

```
val returnTypeSymbol = outputT.mapRustType { it.asOptional() }
```

And then below do:

```
            writer.rustBlock(
                "pub fn $fnName(header_map: &#T::HeaderMap) -> std::result::Result<#T, #T::ParseError>",
                RuntimeType.http,
                returnTypeSymbol,
                headerUtil
            ) {
```

nit: now that we're working on this, I'd rename `outputT` as `outputSymbol`, since it's a `Symbol`, not a `RustType`.

david-perez (approved):
I'd also add a comment pointing to the fact that this deserialization function _never_ returns `None` (note the last line is `out.map(Some)`); when there are no headers it returns an empty map.

Which might prompt the reader to ask themselves: 
* why don't we return a non-`Option`al type here?
* why is `httpPrefixHeaders` not coupled with `required`?

We could add a link to https://github.com/awslabs/smithy/issues/1228

---

PR title needs updating.

crisidev (approved):

rcoh (approved):

PR #1417, opened 2022-05-26
Remove usage of percent_decode_str for decoding query string key-value pairs

hlbarber:
## Motivation and Context

* Removing duplicated work done by `percent_encoding::percent_decode_str`
* Prevent eager collection into `Vec` done by `serde_urlencoded::from_str`
* Fixes https://github.com/awslabs/smithy-rs/issues/1231

## Description

* Use `form_urlencoding::parse` over `serde_urlencoded::from_str`, this removes our dependency on `serde_urlencoded`.
* Remove `percent_encoding::percent_decode_str` where possible.
* Contract `generateParsePercentEncodedStrFn` and its children into one function and make it optionally apply percent encoding.
* Add `impl From<std::convert::Infallible> for RequestRejection` to remove codegen branch.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -981,21 +979,7 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
-        val target = model.expectShape(binding.member.target)
-        return when {
-            target.isStringShape -> generateParsePercentEncodedStrAsStringFn(binding)
-            target.isTimestampShape -> generateParsePercentEncodedStrAsTimestampFn(binding)
-            else -> generateParseStrAsPrimitiveFn(binding)
-        }
-    }
-
-    private fun generateParsePercentEncodedStrAsStringFn(binding: HttpBindingDescriptor): RuntimeType {
+    private fun generateParseFn(binding: HttpBindingDescriptor, percentDecoding: Boolean): RuntimeType {
```
hlbarber, line 982: Write two functions instead of taking a `Boolean` here?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -1005,85 +989,69 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
                 *codegenScope,
                 "O" to output,
             ) {
-                // `<_>::from()` is necessary to convert the `&str` into:
-                //     * the Rust enum in case the `string` shape has the `enum` trait; or
-                //     * `String` in case it doesn't.
-                when (symbol.rustType()) {
-                    RustType.String ->
```
hlbarber, line 1012: This is the branch the `From<std::convert::Infallible> for RequestRejection` removes. Worth it?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -1005,85 +989,69 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
+                            rustTemplate(
+                                """
+                                let value = #{DateTime}::from_str(value, #{format})?;
+                                """,
+                                *codegenScope,
+                                "format" to timestampFormatType,
+                            )
+                        }
                     }
+                    else -> rustTemplate(
```
hlbarber, line 1043: Note that `percent_decode_str` does not take effect on this branch - see https://github.com/awslabs/smithy-rs/pull/1417/files#diff-da9c47cae3f2f75fc01b3625ee430f56f2f495870a77302c3c148e98f7341bd4L994 and https://github.com/awslabs/smithy-rs/pull/1417/files#diff-da9c47cae3f2f75fc01b3625ee430f56f2f495870a77302c3c148e98f7341bd4L1072. This is the reason why I didn't opt to take `percent_decode_str` and the branch completely out of this function.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -1005,85 +989,69 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
-                            "T" to symbol,
-                        )
+                val target = model.expectShape(binding.member.target)
+
+                when {
+                    target.isStringShape -> {
+                        if (percentDecoding) {
+                            rustTemplate(
+                                """
+                                let value = #{PercentEncoding}::percent_decode_str(value).decode_utf8()?;
```
hlbarber, line 999: Taking this and https://github.com/awslabs/smithy-rs/pull/1417/files#diff-da9c47cae3f2f75fc01b3625ee430f56f2f495870a77302c3c148e98f7341bd4R1027 out from within the `when` cause `clippy` to complain in the cases when it's not generated.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCargoDependency.kt
```diff
@@ -17,9 +17,9 @@ import software.amazon.smithy.rust.codegen.smithy.RuntimeConfig
 object ServerCargoDependency {
     val AsyncTrait: CargoDependency = CargoDependency("async-trait", CratesIo("0.1"))
     val FuturesUtil: CargoDependency = CargoDependency("futures-util", CratesIo("0.3"))
+    val FormUrlEncoded: CargoDependency = CargoDependency("form_urlencoded", CratesIo("1"))
```
hlbarber, line 20: This matches the version picked by `serde_urlencoded = "0.7"`.

https://github.com/nox/serde_urlencoded/blob/master/Cargo.toml#L21

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -865,15 +864,15 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
                                 if (memberShape.hasTrait<EnumTrait>()) {
                                     rustTemplate(
                                         """
-                                        let v = <#{memberShape}>::try_from(#{PercentEncoding}::percent_decode_str(&v).decode_utf8()?.as_ref())?;
```
hlbarber, line 868: Although not mentioned in the issue, this seems like the same class of redundancy?

hlbarber: Same as https://github.com/awslabs/smithy-rs/pull/1417/files#diff-da9c47cae3f2f75fc01b3625ee430f56f2f495870a77302c3c148e98f7341bd4L876

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
hlbarber: There is another branch which I noted for removal - https://github.com/awslabs/smithy-rs/pull/1417/files#diff-da9c47cae3f2f75fc01b3625ee430f56f2f495870a77302c3c148e98f7341bd4R860. I opted out of removing it because I was less sure.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
david-perez: Either is fine.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -1005,85 +989,69 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
+                                let value = #{DateTime}::from_str(value, #{format})?;
+                                """,
+                                *codegenScope,
+                                "format" to timestampFormatType,
+                            )
+                        }
                     }
+                    else -> rustTemplate(
+                        """
+                        let value = std::str::FromStr::from_str(value)?;
```
david-perez, line 1045: Here we must be parsing a primitive.

rust-runtime/aws-smithy-http-server/src/rejection.rs
```diff
@@ -194,6 +194,18 @@ pub enum RequestRejection {
 
 impl std::error::Error for RequestRejection {}
 
+// Consider a conversion between `T` and `U` followed by a bubbling up of the conversion error
+// through `Result<_, RequestRejection>`. This [`From`] implementation accomodates the special case
+// where `T` and `U` are equal, in such cases `T`/`U` a enjoy `TryFrom<T>` with
+// `Err = Infallible`.
+impl From<std::convert::Infallible> for RequestRejection {
```
david-perez, line 201: Add a note that we don't need this converter when `!` stabilises.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
david-perez: So a good idea would be to add a

```
                            check(target is NumberShape || target is BooleanShape)
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -1005,85 +989,69 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
-                        )
+                val target = model.expectShape(binding.member.target)
+
+                when {
+                    target.isStringShape -> {
+                        if (percentDecoding) {
+                            rustTemplate(
+                                """
+                                let value = #{PercentEncoding}::percent_decode_str(value).decode_utf8()?;
+                                let value = #{T}::try_from(value.as_ref())?;
```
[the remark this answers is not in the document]
david-perez: No, my suggestion doesn't work in the case the string shape has `@enum` attached. We need to use `try_from` to convert the `String` into the generated enum. If the string shape does not have `@enum` attached, then the `try_from` is a no-op and it's leveraging the `impl From<std::convert::Infallible> for RequestRejection` that you added.

Might be worth adding a comment.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
hlbarber: I've taken this second branch out in https://github.com/awslabs/smithy-rs/pull/1417/commits/1477e678baa73c52510c12d9ccae639e650d2179

david-perez (approved):

PR #1420, opened 2022-05-27
Refactor `sdk-versioner` to operate off the versions manifest

jdisanti:
## Motivation and Context
This is an incremental step towards independent service crate versions for [RFC 12](https://awslabs.github.io/smithy-rs/design/rfcs/rfc0012_independent_crate_versioning.html). This PR refactors the `sdk-versioner` tool used to revise example crate manifests to point to the correct versions of the SDK crates to use the `versions.toml` rather than fixed version numbers.


tools/sdk-versioner/src/main.rs
```diff
@@ -263,23 +341,26 @@ mod tests {
...
             version = "0.1.0"
 
             [dependencies]
             aws-config = { version = "0.5.0", path = "/foo/asdf/aws-config" }
-            aws-sdk-s3 = { version = "0.5.0", path = "/foo/asdf/s3" }
-            aws-smithy-types = { version = "0.35.0", path = "/foo/asdf/aws-smithy-types" }
-            aws-smithy-http = { version = "0.35.0", path = "/foo/asdf/aws-smithy-http", features = ["test-util"] }
+            aws-sdk-s3 = { version = "0.13.0", path = "/foo/asdf/s3" }
+            aws-smithy-types = { version = "0.10.0", path = "/foo/asdf/aws-smithy-types" }
+            aws-smithy-http = { version = "0.9.0", path = "/foo/asdf/aws-smithy-http", features = ["test-util"] }
```
Velfi, line 363: It may be useful to add some major and patch versions in here just to increase the cases it's covering

aws/sdk/build.gradle.kts
```diff
@@ -369,9 +367,10 @@ tasks.register("finalizeSdk") {
         "relocateExamples",
         "generateIndexMd",
         "fixManifests",
+        "generateVersionManifest",
+        "fixExampleManifests",
         "hydrateReadme",
-        "relocateChangelog",
-        "generateVersionManifest"
```
rcoh, line 374: we are removing this because the versions manifest will start being updated differently?

tools/sdk-versioner/src/main.rs
```diff
@@ -18,56 +19,102 @@ use toml::value::{Table, Value};
...
+enum Args {
+    /// Revise crates to use paths in dependencies
+    UsePathDependencies {
+        /// Root SDK path the path dependencies will be based off of
+        #[clap(long)]
+        sdk_path: PathBuf,
+        /// Path(s) to recursively update Cargo.toml files in
+        #[clap()]
+        crate_paths: Vec<PathBuf>,
+    },
```
rcoh, line 32: is this mode actually still needed?

rcoh (approved):
LGTM—personally I might keep `path` and `version` as options and just alter behavior on whether or not they're provided `use-path-and-version-dependencies` is just kind of long? but not strong preference

tools/sdk-versioner/src/main.rs
jdisanti: It's useful for local development when working on examples.

aws/sdk/build.gradle.kts
jdisanti: It wasn't removed, just reordered. The versions manifest needs to get generated before the example manifests can be fixed now.

tools/sdk-versioner/src/main.rs
jdisanti: The version numbers themselves are opaque to the tool, so I'm not too worried about this. Just needs to test that they're coming from the `versions.toml` file.

PR #1423, opened 2022-05-30
Set Content-Length header in server responses

hlbarber:
## Motivation and Context

* https://httpwg.org/specs/rfc7230.html#header.content-length
* Closes https://github.com/awslabs/smithy-rs/issues/1146

## Description

* Add `Content-Length` header to non-streaming server responses
* Add a `Content-Length` checks to the existing `httpResponseTests`


hlbarber:
@david-perez 

> We should contribute a PR to awslabs/smithy adding an @httpResponseTests asserting Content-Length header has the expected value

Posted an issue inquiring about this, will follow with a PR shortly after if needed

https://github.com/awslabs/smithy/issues/1255

PR #1438, opened 2022-06-03
Use the `SymbolProvider` and `Model` directly from the Smithy plugins

david-perez:
Despite all the decorators and protocols that we have, there aren't any
`RustCodegenDecorator`s or `ProtocolGeneratorFactory`s that are
modifying the base symbol provider that the Smithy plugin provides. So
we can use directly the base symbol provider everywhere and remove the
methods from these interfaces to make the code a little bit simpler.

Likewise, there isn't a single `ProtocolGeneratorFactory` that
transforms the base model that the Smithy plugin provides.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerAwsJson.kt
```diff
@@ -36,8 +35,6 @@ class ServerAwsJsonFactory(private val version: AwsJsonVersion) : ProtocolGenera
     override fun buildProtocolGenerator(codegenContext: CodegenContext): ServerHttpBoundProtocolGenerator =
         ServerHttpBoundProtocolGenerator(codegenContext, protocol(codegenContext))
 
-    override fun transformModel(model: Model): Model = model
```
rcoh, line 39: I used this yesterday to augment the S3 model with some extra metadata—I think we should keep these, especially since the default implementation means that there isn't any cost to having it

david-perez: You used the `transformModel` from the decorators, not this one. I accidentally removed the invocation of the former, but I didn't mean to, that's why CI failed. I've now fixed that.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
```diff
@@ -12,14 +12,13 @@ import software.amazon.smithy.model.shapes.ServiceShape
...
 import software.amazon.smithy.rust.codegen.smithy.RustCrate
-import software.amazon.smithy.rust.codegen.smithy.RustSymbolProvider
 import software.amazon.smithy.rust.codegen.smithy.generators.LibRsCustomization
 import software.amazon.smithy.rust.codegen.smithy.generators.ManifestCustomizations
 import software.amazon.smithy.rust.codegen.smithy.generators.client.FluentClientDecorator
 import software.amazon.smithy.rust.codegen.smithy.generators.config.ConfigCustomization
 import software.amazon.smithy.rust.codegen.smithy.protocols.ProtocolMap
 import software.amazon.smithy.rust.codegen.util.deepMergeWith
-import java.util.ServiceLoader
+import java.util.*
```
82marbag, line 21: Why hasn't the linter complained about `*`?

82marbag (approved):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
david-perez: What linter?

rcoh: kotlin style check

rcoh (approved):
LGTM, thanks for the cleanup!

crisidev (approved):

PR #1439, opened 2022-06-03
Remove erroneous reference to the base `Symbol` when mapping Rust types

david-perez:
`.addReference(this)` adds a reference to the `Symbol` on which
`.mapRustType` was called. This is correct only when `f` is a function
that _wraps_ its input `RustType`; for example, when `f` wraps it in a
`Box` or constructs a `Vec`. However, the code is incorrect for an
arbitrary `f`; for example, when `f` _swaps_ the type.

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt
@@ -109,11 +109,15 @@ fun Symbol.makeOptional(): Symbol {
     }
 }
 
-/** Map the RustType of a symbol with [f] */
+/**
+ * Map the [RustType] of a symbol with [f].
+ *
+ * WARNING: This function does not set any `SymbolReference`s on the returned symbol. You will have to add those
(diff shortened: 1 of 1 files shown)


Velfi (approved):

rcoh:
do you think this should just use `symbol.toBuilder()` instead of making a new symbol? We probably want to preserve symbol references—those are use for nested dependencies. We could also make this method private and only use it in the mapping use cases you described

david-perez:
> do you think this should just use symbol.toBuilder() instead of making a new symbol? We probably want to preserve symbol references—those are use for nested dependencies.

That would also be incorrect for an arbitrary `f` right? `f` could map from e.g. `Vec<http::header::HeaderName>` to `Option<String>` --- the reference to `http::header::HeaderName` should not be preserved; if you do preserve it, and it's your only usage of a type from the `http` crate, you'll end up with a dependency on `http` that you don't need.

It's very unlikely that eliminating `SymbolReference`s upon usage of this function will cause us to stop depending on a crate that we do need. But if we do, the build will fail, which I think is a better outcome than silently depending on something we don't need.

rcoh (approved):

crisidev (approved):

PR #1443, opened 2022-06-06
Reintroduce `sdk-sync` parallelism by disabling Smithy parallelism

jdisanti:
## Motivation and Context
This reverts #1436 and configures the JVM so that Smithy codegen won't project services in parallel, and introduces a CLI arg to override the number of threads used.

### DIFF
--- a/tools/sdk-sync/src/sync/gen.rs
+++ b/tools/sdk-sync/src/sync/gen.rs
@@ -112,8 +112,6 @@ impl DefaultSdkGenerator {
     fn aws_sdk_assemble(&self) -> Result<()> {
         info!("Generating the SDK...");
         let mut command = Command::new("./gradlew");
-        command.arg("--no-daemon"); // Don't let Gradle continue running after the build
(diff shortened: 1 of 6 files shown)


tools/sdk-sync/src/sync/gen.rs
```diff
@@ -114,6 +114,9 @@ impl DefaultSdkGenerator {
         let mut command = Command::new("./gradlew");
         command.arg("--no-daemon"); // Don't let Gradle continue running after the build
         command.arg("--info"); // Increase logging verbosity for failure debugging
+
+        // Disable Smithy's codegen parallelism in favor of sdk-sync parallelism
+        command.arg("-Djava.util.concurrent.ForkJoinPool.common.parallelism=1");
```
rcoh, line 119: lets make this configurable? I think we _may_ want to have at least some parallelism here? (eg. this probably massively reduces the memory burden)

Whatever we can do to make this easy to tweak in the build job without redeploying the image would be 💯

rcoh (approved):

PR #1444, opened 2022-06-06
Add the ability to generate `pub(crate)` values

david-perez:
A small commit adding the ability to generate `pub(crate)` values.
Currently we don't need to, but this is used in #1342. It's also nicer
to use an enum instead of passing around a boolean.


david-perez:
This is primarily to make the diff of #1342 lighter.

crisidev (approved):
Nice!

rcoh (approved):

PR #1448, opened 2022-06-08
Configure Gradle parallelism and Kotlin compiler in `sdk-sync`

jdisanti:
## Motivation and Context
Each invocation of `gradlew aws:sdk:assemble` from `sdk-sync` was creating some form of unbounded parallelism that was causing issues with the release process. These changes restrict that a bit more, although something still seems to be doing things in parallel besides `sdk-sync` even after these changes.


rcoh (approved):

david-perez:
> was causing issues with the release process

What issues?

PR #1449, opened 2022-06-09
Docker hop limit

jmklix:
add note on settings needed when using sdk inside docker container

## Motivation and Context

https://github.com/awslabs/aws-sdk-rust/issues/540

## Description

Add comments to explain docker settings needed when using this sdk

## Checklist


rcoh (changes requested):
thanks for making this change!

I'm not sure that it makes sense in that list—that list is a list of where IMDS will search for configuration options. This is actually related to IMDSv1 vs. V2—there is a section at the top of this file that refers to this—lets add this there.

We should also note the root cause of this behavior:
> By default, the response to PUT requests has a response hop limit (time to live) of 1 at the IP protocol level. You can adjust the hop limit using the modify-instance-metadata-options command if you need to make it larger. For example, you might need a larger hop limit for backward compatibility with container services running on the instance. For more information, see [modify-instance-metadata-options](https://docs.aws.amazon.com/cli/latest/reference/ec2/modify-instance-metadata-options.html) in the AWS CLI Command Reference.

I think we should also clarify _where_ these settings need to be applied—as a customer just seeing those two settings listed doesn't give me context of what I should do with them. These changes actually don't have anything to do with docker—docker just adds an extra hop which causes timeouts when combined with the default IMDS hop-limit of 1

These settings need to be applied to the instance, either via `ModifyInstanceMetadata` or `RunInstances`.

So pulling all that together, what do you think about adding something like this at the top of the docs after the note about only supporting IMDSv2:

----

**Note**: When running in a Docker container, all network requests will incur an additional hop. When combined with the default IMDS hop limit of 1, this will cause requests to IMDS to timeout! To fix this issue, you'll need to set the following instance metadata settings :
```
amazonec2-metadata-token=required
amazonec2-metadata-token-response-hop-limit=2
```

On an instance that is already running, these can be set with [ModifyInstanceMetadataOptions](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyInstanceMetadataOptions.html). On a new instance, these can be set with the `MetadataOptions` field on [RunInstances](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html).

For more information about IMDSv2 vs. IMDSv1 see [this guide](https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/configuring-instance-metadata-service.html)

rcoh (changes requested):

PR #1451, opened 2022-06-09
add: RFC for supporting flexible checksums

Velfi:


design/src/SUMMARY.md
```diff
@@ -27,8 +27,9 @@
   - [RFC-0011: Publishing Alpha to Crates.io](./rfcs/rfc0011_crates_io_alpha_publishing.md)
   - [RFC-0012: Independent Crate Versioning](./rfcs/rfc0012_independent_crate_versioning.md)
   - [RFC-0013: Body Callback APIs](./rfcs/rfc0013_body_callback_apis.md)
-  - [RFC-0014: Fine-grained timeout configuration](./rfcs/rfc00014_timeout_config.md)
-  - [RFC-0015: How Cargo "features" should be used in the SDK and runtime crates](./rfcs/rfc00015_using_features_responsibly.md)
+  - [RFC-0014: Fine-grained timeout configuration](./rfcs/rfc0014_timeout_config.md)
+  - [RFC-0015: How Cargo "features" should be used in the SDK and runtime crates](./rfcs/rfc0015_using_features_responsibly.md)
+  - [RFC-0016: Supporting Flexible Checksums](./rfcs/rfc0016_flexible_checksum_support.md)
```
jdisanti, line 32: There are two places with links to the docs. The other is in `overview.md`.

design/src/rfcs/rfc0016_flexible_checksum_support.md
```diff
@@ -0,0 +1,1442 @@
...
+    /// - checksum when base64 encoded
+    /// - trailer name
+    /// - trailer separator
+    ///
+    /// This is necessary for calculating the true size of the request body for certain
+    /// content-encodings.
+    pub fn trailer_length(&self) -> u64 {
+        let trailer_name_size_in_bytes = self.checksum.header_name().as_str().len();
+        let base64_encoded_checksum_size_in_bytes =
+            base64::encoded_length(self.checksum.size() as usize);
```
jdisanti, line 550: Nice!

design/src/rfcs/rfc0016_flexible_checksum_support.md
```diff
@@ -0,0 +1,1442 @@
...
+```
+
+### Sigv4 Update
+
+When sending checksum-verified requests with a streaming body, we must update the usual signing process. Instead of signing the request based on the request's body, we must sign it with a magic string instead:
+
+```HTTP
+Authorization: <computed authorization header value using "STREAMING-UNSIGNED-PAYLOAD-TRAILER">
+x-amz-content-sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER
+```
```
jdisanti, line 1253: How does the signer know which algorithm header is used in order to create this magic string?

Velfi: It relies on pre-existing functionality, like how we currently send unsigned streaming requests.

Velfi: I'll update the language to reflect that we're setting this string to tell the signer that we have unsigned trailers following the body

design/src/rfcs/rfc0016_flexible_checksum_support.md
```diff
@@ -0,0 +1,1442 @@
...
+        if self.options.chunk_length.is_none() && self.options.stream_length.is_none() {
+            return None;
+        }
+
+        let mut length = 0;
+        let stream_length = self.options.stream_length.unwrap_or_default();
+        if stream_length != 0 {
+            if let Some(chunk_length) = self.options.chunk_length {
+                // I don't think we'll hit this case b/c we only ever send things in one chunk
+                assert!(chunk_length > MINIMUM_CHUNK_LENGTH);
```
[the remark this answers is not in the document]
Velfi: I carried over this assert when I implemented this part by reading the Go v2 SDK. Looking at the referenced Sigv4, it seems to be a bit unclear on whether or not there are limits. All this stuff works right now because we don't use the chunking at all. I brought it over in case we want to implement signed chunks. I'll write a test to ensure the chunking works.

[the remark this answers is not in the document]
Velfi: I removed this check because it's not  necessary

jdisanti (approved):
LGTM!

PR #1454, opened 2022-06-10
Fix issues with setting crate ownership during publish

jdisanti:
## Motivation and Context
PR #1396 added some code to remove individual devs as crate owners during publish to crates.io, but it turns out a team can't be the sole owner of a crate, so this failed during the last release. This PR adds the SDK bot user as an owner, and since this requires accepting ownership of the crate as the bot user, doesn't attempt to remove any other individuals if the bot needs to be added. This should fix ownership on all the crates after two releases.

Only an individual owner of a crate can invite another individual to own a crate. Thus, separate from this PR, all individual owners of the SDK/runtime crates need to run a script to invite the bot user before the next release.

Tracking who still needs to run this script below:
- [x] @jdisanti
- [x] @Velfi
- [x] @rcoh

## Testing
- Since crate publishing is idempotent, I was able to verify that this works on some of the crates that I own. For example, [`aws-hyper`](https://crates.io/crates/aws-hyper) was corrected by these changes.


tools/publisher/src/subcommand/publish.rs
```diff
@@ -195,22 +195,36 @@ async fn correct_owner(package: &Package) -> Result<()> {
...
-                    .spawn()
-                    .await
-                    .context("remove incorrect owner")?;
-                info!(
-                    "Removed incorrect owner `{}` from crate `{}`",
-                    incorrect_owner, package.handle
-                );
+            let mut added_individual = false;
+            for &crate_owner in CRATE_OWNERS {
+                if !owners.iter().any(|owner| owner == crate_owner) {
```
rcoh, line 200: can just use `owners.contains` here I think

rcoh (approved):

tools/publisher/src/subcommand/publish.rs
jdisanti: Would need a reference to a `String` type:
```
   --> src/subcommand/publish.rs:200:37
    |
200 |                 if !owners.contains(crate_owner) {
    |                                     ^^^^^^^^^^^ expected struct `std::string::String`, found `str`
    |
    = note: expected reference `&std::string::String`
               found reference `&str`
```

PR #1464, opened 2022-06-13
Pin the `clap` version used to prevent future deprecation warnings

jdisanti:
## Motivation and Context
The `clap` maintainers are planning to introduce more deprecations with every minor release (https://github.com/clap-rs/clap/issues/3822):
>  We do expect to be adding deprecations every minor release and we recommend not blocking CI on this as the toil adds backpressure on crates to not evolve their API.

Since we're disallowing warnings broadly at the environment variable level, this PR pins all the `clap` versions so that future minor releases won't break our CI.


Velfi (approved):

rust-runtime/aws-smithy-http-server/examples/pokemon_service/Cargo.toml
```diff
@@ -7,7 +7,7 @@ authors = ["Smithy-rs Server Team <smithy-rs-server@amazon.com>"]
 description = "A smithy Rust service to retrieve information about Pokémon."
 
 [dependencies]
-clap = { version = "3.2.1", features = ["derive"] }
+clap = { version = "~3.2.1", features = ["derive"] }
```
epage, line 10: This can now be reverted as deprecations are now off by default, see clap-rs/clap#3830 for details.  When you do start working through the deprecations as part of upgrading to 4.0, we've also improved the warnings for derive users.  See clap-rs/clap#3832 for examples.

PR #1471, opened 2022-06-16
Fix `modifyMtime` task when `generateSmithyBuild` task is up to date

david-perez:
When `generateSmithyBuild` task is up to date, the hashes of the build
are not calculated, so `modifyMtime` fails, because it expects them to
be registered in Gradle's project properties. This can happen if you run
a command like `./gradlew codegen-server-test:test` twice consecutively.

In this case, we should skip the `modifyMtime` task altogether.
Everything is up to date, so no codegen will run, and no artifacts will
change.


crisidev (approved):

buildSrc/src/main/kotlin/CodegenTestCommon.kt
```diff
@@ -187,19 +189,23 @@ fun Project.registerModifyMtimeTask() {
...
-                .filter { it.isFile }
-                .map {
-                    getChecksumForFile(it) to it
-                }
-                .forEach { (currentHash, currentFile) ->
-                    previousBuildHashes[currentHash]?.also { oldMtime ->
-                        println("Setting `mtime` of $currentFile back to `$oldMtime` because its hash `$currentHash` remained unchanged after a rebuild.")
-                        currentFile.setLastModified(oldMtime)
+            if (!project.extra.has(previousBuildHashesKey)) {
+                println("No hashes from a previous build exist because `generateSmithyBuild` is up to date, skipping `mtime` fixups")
```
rcoh, line 193: side note—I'm actually not sure the best thing to do here, but we might want to use some sort of logger invocation instead of println

rcoh (approved):

buildSrc/src/main/kotlin/CodegenTestCommon.kt
jdisanti: In Gradle, you should use SLF4J to log: https://docs.gradle.org/current/userguide/logging.html#sec:sending_your_own_log_messages

PR #1474, opened 2022-06-20
Make `make_operation` synchronous

david-perez:

### DIFF
--- a/aws/rust-runtime/aws-config/src/profile/credentials/exec.rs
+++ b/aws/rust-runtime/aws-config/src/profile/credentials/exec.rs
@@ -58,7 +58,6 @@ impl AssumeRoleProvider {
             .build()
             .expect("operation is valid")
             .make_operation(&config)
-            .await
             .expect("valid operation");
         let assume_role_creds = client_config
             .sts_client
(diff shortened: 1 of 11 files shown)


82marbag (approved):

david-perez:
Nevermind, Glacier has an operation customization that calculates [two headers](https://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html) and needs to `.collect().await` the body into a byte slab.

https://github.com/awslabs/smithy-rs/blob/2a9d29d46359c34556ff9bd88723512d1e1696bd/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/glacier/GlacierDecorator.kt#L16-L38

https://github.com/awslabs/smithy-rs/blob/2a9d29d46359c34556ff9bd88723512d1e1696bd/aws/rust-runtime/aws-inlineable/src/glacier_checksums.rs#L20-L64

Velfi:
We could potentially do this by moving some of what `make_operation` does into middleware. If this turns out to be important, write up a justification and we can discuss.

david-perez:
@Velfi Not very important; I simply found it odd that a method named `make_operation` would have to be async and decided to remove `async` to see what broke. The UI might make other people scratch their heads too in that why could preparing the operation possibly have to be async; the async part should only be relevant when sending it. But it's a minor papercut.

Moving it to middleware might be better yes. The implementation could also be smarter and use [multipart uploads](https://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html#checksum-calculations-upload-archive-using-mpu) instead of always buffering everything and sending a big request? I didn't see that bit taken into account but I might have missed it.

PR #1478, opened 2022-06-21
Add support for retrying individual call timeouts

rcoh:
## Motivation and Context
- https://github.com/awslabs/aws-sdk-rust/issues/558

## Description
When timeout support was added initially, the resulting error `SdkError::TimeoutError` was never added to the retry policy which prevented these errors from being retried. This was a bug—This commit rectifies the and adds two integration-level tests that ensure that timeouts are properly retried.

## Testing
- [x] integration test

## Checklist


Velfi (approved):
Thanks for cleaning up after me!

jdisanti (approved):

Issue #1157, opened 2022-02-04
Add support for server event streams


david-perez:
The server SDK supports _blob_ streaming, but we haven't formally added support nor tested event streams https://awslabs.github.io/smithy/1.0/spec/core/stream-traits.html#event-streams

Issue #1219, opened 2022-02-24
PR bot codegen diff is broken for large diffs


jdisanti:
Discovered in https://github.com/awslabs/smithy-rs/pull/1216, it seems like diff2html omits files from the diff when the diff is exceedingly large.

I did the following diagnosis work:
- Modified my local [`codegen-diff-revisions.py`](https://github.com/awslabs/smithy-rs/blob/main/tools/codegen-diff-revisions.py) script to not delete the temporary branches ([here](https://github.com/awslabs/smithy-rs/blob/main/tools/codegen-diff-revisions.py#L83-L86)).
- Ran the script for the changes in the PR linked above.
- Confirmed S3 and several other services had changes in the `git diff` between the two temporary branches.
- Confirmed the diff2html output did not include these files.

This issue tracks fixing the HTML output. We might need to build a tool or update an existing tool to generate a paginated HTML diff since SDK codegen diffs are always going to be huge.

PR #1479, opened 2022-06-21
Support server event streams

82marbag:
Support server event streams

Closes: #1157

## Motivation and Context

This PR adds support for server event streams. See #1157

## Description

This PR adds support for server event streams and updates the Pokemon model and service to use them

## Testing

* Run the Pokemon server and client
* `./gradlew :codegen-server-test:test `
* `./gradlew :codegen-test:test`

## Checklist


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -682,29 +679,29 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
+                    structuredHandler = structureShapeHandler
+                )
+                return writable {
+                    if (binding.member.isStreaming(model)) {
                         rustTemplate(
                             """
                             {
                                 let body = request.take_body().ok_or(#{RequestRejection}::BodyAlreadyExtracted)?;
-                                Some(body.into())
+                                let bytes = #{Hyper}::body::to_bytes(body).await?;
```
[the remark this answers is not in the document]
crisidev: Agree. What is the reasoning behind this?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/EventStreamDecorator.kt
```diff
@@ -5,89 +5,10 @@
...
-                        *codegenScope
-                    )
-                }
-            }
-            else -> emptySection
-        }
-    }
+interface EventStreamDecorator : RustCodegenDecorator {
+    fun applies(codegenContext: CodegenContext): Boolean =
+        codegenContext.serviceShape.hasEventStreamOperations(codegenContext.model)
```
jdisanti, line 13: I think you can make this a private function on `MergedEventStreamDecorator` and just call it in `configCustomizations` and `operationCustomizations`. There's no need for one of these to apply while another one doesn't.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt
```diff
@@ -25,7 +26,7 @@ val DECORATORS = listOf(
     RegionDecorator(),
     AwsEndpointDecorator(),
     UserAgentDecorator(),
-    EventStreamDecorator(listOf(SigV4SigningDecorator())),
+    MergedEventStreamDecorator(listOf(SigV4SigningDecorator(), NoOpEventStreamSigningDecorator())),
```
jdisanti, line 29: Shouldn't this only have the `SigV4SigningDecorator` in the list?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4SigningDecorator.kt
```diff
@@ -111,23 +112,29 @@ class SigV4SigningConfig(
...
             else -> emptySection
         }
     }
+
+    override fun inner(): Writable {
+        return writable {
+            rustTemplate(
+                """
+                /// Creates a new Event Stream `SignMessage` implementor.
+                pub fn new_event_stream_sigv4_signer(
```
jdisanti, line 127: Should we make this `pub(crate)`? I can't think of a reason a customer would need to call this.

Separately, would it be possible to have one `new_event_stream_signer` function that returns `impl SignMessage`?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
```diff
@@ -164,6 +169,19 @@ class HttpBoundProtocolPayloadGenerator(
         }
     }
 
+    private fun generateSigner(): Writable {
+        when (target) {
+            CodegenTarget.CLIENT -> {
+                if (codegenContext.serviceShape.hasTrait<SigV4Trait>()) {
+                    return writable { rust("_config.new_event_stream_signer(properties.clone())") }
+                }
+                return writable { rust("_config.new_event_stream_no_op_signer(properties.clone())") }
```
jdisanti, line 178: The benefit of having one `new_event_stream_signer` function on the config is that you won't need this if/else for each signing trait.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4SigningDecorator.kt
82marbag: It is possible, but I'd either make `MergedEventStreamDecorator` return the first that applies or take only one `EventStreamDecorator`; we can't have multiple `fn new_event_stream_signer` anyway and we'd have to be careful while coding (so maybe better to prevent it?)

jdisanti: Is there a good reason for having multiple?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
```diff
@@ -156,39 +167,67 @@ class HttpBoundProtocolPayloadGenerator(
...
+        serializerGenerator: StructuredDataSerializerGenerator,
+        outerName: String,
     ) {
         val memberName = symbolProvider.toMemberName(memberShape)
         val unionShape = model.expectShape(memberShape.target, UnionShape::class.java)
 
+        val contentType = when (target) {
+            CodegenTarget.CLIENT -> httpBindingResolver.requestContentType(operationShape)
+            CodegenTarget.SERVER -> httpBindingResolver.responseContentType(operationShape)
+        }
```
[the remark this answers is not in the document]
82marbag: I don't find it in the spec, is it a custom header used by some services? If so, I should remove it rather than adjusting it. On some services I see it's `application/vnd.amazon.event-stream` (notice the `-`)

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4SigningDecorator.kt
82marbag: I thought we could generate all that apply, but then I'll remove the list and use only the relevant one and keep `applies()` as it was before

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/transformers/RemoveEventStreamOperations.kt
```diff
@@ -21,7 +21,7 @@ object RemoveEventStreamOperations {
 
     fun transform(model: Model, settings: RustSettings): Model {
         // If Event Stream is allowed in build config, then don't remove the operations
-        if (settings.codegenConfig.eventStreamAllowList.contains(settings.moduleName)) {
+        if (true || settings.codegenConfig.eventStreamAllowList.contains(settings.moduleName)) {
```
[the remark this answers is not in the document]
82marbag: I haven't seen any error building and testing this change, I'll remove this process

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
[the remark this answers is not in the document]
82marbag: Discussed offline and created https://github.com/awslabs/smithy/issues/1287 to document them. Fixed in this PR

rust-runtime/aws-smithy-http/src/event_stream/receiver.rs
```diff
@@ -156,10 +156,19 @@ impl<T, E> Receiver<T, E> {
...
-                raw: RawMessage::Decoded(message),
-            }),
+            Err(err) => {
+                if matches!(
+                    err,
+                    aws_smithy_eventstream::error::Error::EventStreamError(_)
+                ) {
+                    // Terminate the stream
+                    self.buffer = RecvBuf::EosPartial(Default::default())
+                }
```
jdisanti, line 166: I think it would be better to keep `unmarshall` immutable on `self`, and if the buffer needs to be changed, the caller of `unmarshall` should set it if it returns `Err` at all. All errors would be terminal on this. The code can be rearranged so that `unmarshall` is only called once in `recv`.

`RecvBuf::EosPartial` doesn't seem like the right variant to use here since it specifically means: some data was received, it wasn't a full message. A new variant should be created instead.

rust-runtime/aws-smithy-http/src/event_stream/sender.rs
```diff
@@ -77,16 +117,12 @@ where
...
     }
 }
 
-impl<T, E> Stream for MessageStreamAdapter<T, E>
-where
-    E: StdError + Send + Sync + 'static,
-{
-    type Item = Result<Bytes, SdkError<E>>;
+impl<T> Stream for MessageStreamAdapter<T> {
+    type Item = Result<Bytes, SdkError<MessageStreamError>>;
```
jdisanti, line 125: The `SdkError` here needs to be generic over the modeled error type so that customers can match on the streaming errors. I don't think we need to have `MessageStreamError` since `SdkError` already encompasses its functionality.

jdisanti (commented):
I think you might have misunderstood event stream errors, and that's what most of my comments are about. Other than errors and the removal of the event stream allow list, I think this looks good. Both of these need to be addressed before approval though.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/transformers/EventStreamNormalizer.kt
```diff
@@ -37,3 +59,38 @@ object EventStreamNormalizer {
...
+            inputOutput.filter { it.getTrait<ErrorTrait>()!!.isServerError },
+            inputOutput.filter { it.getTrait<ErrorTrait>()!!.isClientError },
+        )
+    }
+    allErrors[operationShape.id.name] = errors
+}
+fun OperationShape.getStreamError(target: CodegenTarget): List<Shape> {
+    return allErrors[this.id.name]!!.getStreamError(target)
+}
+fun OperationShape.nonEventStreamErrors(): List<Shape> {
```
jdisanti: I think we should call them "operation errors" and "event stream errors" rather than "non event stream errors".

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/CombinedErrorGenerator.kt
```diff
@@ -43,6 +46,13 @@ fun OperationShape.errorSymbol(symbolProvider: SymbolProvider): RuntimeType {
     return RuntimeType("${symbol.name}Error", null, "crate::error")
 }
 
+fun OperationShape.eventStreamErrorSymbol(symbolProvider: SymbolProvider, target: CodegenTarget): RuntimeType {
+    val symbol = symbolProvider.toSymbol(this)
+    return when (target) {
+        CodegenTarget.CLIENT -> RuntimeType("${symbol.name}ClientEventStreamError", null, "crate::error")
+        CodegenTarget.SERVER -> RuntimeType("${symbol.name}ServerEventStreamError", null, "crate::error")
```
jdisanti, line 54: I don't think we need to differentiate between client/server in the error names. If you disagree, I want to understand better why this is required.

Separately, we should brainstorm some shorter names for these 😄

82marbag: Not doing so leads to the same problem of having all of them together (operation and event streams), that the client has the option to yield errors it's not supposed to. Merging them would generate like
```
        let payload = match input.kind {
            crate::error::CapturePokemonOperationEventStreamErrorKind::MasterBallUnsuccessful(inner) => ...
            crate::error::CapturePokemonOperationEventStreamErrorKind::Unhandled(_inner) => ...
          _ => ???
```
If they are merged together, at this point it's the same to having all of them under the operation errors. What do you think?
As for the names, any suggestion I'm happy to take (input/output as well) :)

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/transformers/EventStreamNormalizer.kt
```diff
@@ -20,13 +26,34 @@ import software.amazon.smithy.rust.codegen.util.isEventStream
...
+        if (shape is OperationShape && shape.isEventStream(model)) {
+            addStreamErrorsToOperationErrors(model, shape)
+        } else if (shape is UnionShape && shape.isEventStream()) {
             syntheticEquivalent(model, shape)
         } else {
             shape
         }
     }
 
+    private fun addStreamErrorsToOperationErrors(model: Model, operation: OperationShape): OperationShape {
```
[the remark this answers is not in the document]
82marbag: 1. True, thanks. Fixed it.
2. Also true, that it indicates fault, but I made it so that the client writes `yield Err( ClientEventStreamError::... )` and similarly for the server; and so the client `match (_) => Err( ServerEventStreamError::... )`. As above, I'm happy to change the names however you think they make more sense. As for "If an error is included in an input stream, then it is a client error. If it is included in an output stream, then it is a server error. If it is included in an input stream and an output stream, then it is both" I haven't seen this in the docs, I'm asking the team to confirm this

rust-runtime/aws-smithy-eventstream/src/error.rs
```diff
@@ -25,6 +25,7 @@ pub enum Error {
     TimestampValueTooLarge(DateTime),
     Marshalling(String),
     Unmarshalling(String),
+    EventStreamError(Box<dyn StdError + Send + Sync>),
```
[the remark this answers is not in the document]
82marbag: Removed

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt
```diff
@@ -179,6 +180,18 @@ class EventStreamUnmarshallerGenerator(
...
+                    .forEach {
+                        rustTemplate(
+                            """
+                            if let Some(error) = parsed.${it.memberName} {
+                                return Err(#{Error}::EventStreamError(Box::new(error)));
+                            }
+                            """,
+                            *codegenScope
+                        )
+                    }
```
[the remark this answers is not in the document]
82marbag: Fixed, thanks

codegen-server-test/model/pokemon.smithy
```diff
@@ -22,6 +22,64 @@ resource PokemonSpecies {
...
+@output
+structure CapturePokemonOperationEventsOutput {
+    @httpPayload
+    events: CapturePokemonEvents,
+}
+
+@streaming
+union AttemptCapturingPokemonEvent {
+    event: CapturingEvent,
+}
```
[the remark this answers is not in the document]
82marbag: Added

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/NoOpEventStreamSigningDecorator.kt
```diff
@@ -0,0 +1,76 @@
...
+            rustTemplate(
+                """
+                /// Creates a new Event Stream `SignMessage` implementor.
+                pub fn new_event_stream_signer(
+                    &self,
+                    _properties: #{SharedPropertyBag}
+                ) -> impl #{SignMessage} {
+                    #{NoOpSigner}{}
+                }
+                """,
```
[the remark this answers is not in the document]
82marbag: This is to have `_properties` and `properties` depending on whether it's used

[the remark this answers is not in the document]
jdisanti: You can just have it always be `_properties` to avoid the warning (we do this in a number of places). It can still be used.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/CombinedErrorGenerator.kt
82marbag: Implemented, thank you

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/NoOpEventStreamSigningDecorator.kt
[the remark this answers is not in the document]
82marbag: Ok. Done, thanks

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/CombinedErrorGenerator.kt
```diff
@@ -52,14 +60,29 @@ class CombinedErrorGenerator(
...
-        val errors = operationIndex.getErrors(operation)
-        val operationSymbol = symbolProvider.toSymbol(operation)
         val symbol = operation.errorSymbol(symbolProvider)
+        renderErrors(writer, operation.operationErrors(model).map { it.asStructureShape().get() }.toMutableList(), symbol)
+
+        if (operation.isEventStream(model)) {
+            operation.eventStreamErrors(model)
+                .forEach {
+                    val unionShape = it.key
+                    val unionErrors = it.value
```
jdisanti, line 74:
```suggestion
                .forEach { (unionShape, unionErrors) ->
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/error/TopLevelErrorGenerator.kt
```diff
@@ -73,27 +77,40 @@ class TopLevelErrorGenerator(codegenContext: CodegenContext, private val operati
...
-                            val errSymbol = symbolProvider.toSymbol(errorShape)
-                            rust("#TKind::${errSymbol.name}(inner) => Error::${errSymbol.name}(inner),", operationError)
+        val nonEventStreamErrors = operationShape.operationErrors(model).map { it.id }
+        val allErrors: List<Pair<RuntimeType, List<ShapeId>>> = listOf(
+            Pair(operationShape.errorSymbol(symbolProvider), nonEventStreamErrors),
+        ) + operationShape.eventStreamErrors(model)
+            .map { Pair(it.key.eventStreamErrorSymbol(symbolProvider), it.value.map { it.id }) }
+        allErrors.forEach { pair ->
+            val symbol = pair.first
+            val errors: List<ShapeId> = pair.second
```
jdisanti, line 87:
```suggestion
        allErrors.forEach { (symbol, errors) ->
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/transformers/EventStreamNormalizer.kt
```diff
@@ -37,3 +80,40 @@ object EventStreamNormalizer {
             .build()
     }
 }
+
+fun OperationShape.operationErrors(model: Model): List<Shape> {
```
jdisanti, line 84: What is the intent for this function? The `OperationIndex.getErrors` already returns all the errors for the operation, right? So why does it need to be filtered by the synthetic trait? Wouldn't all of them have it since the normalizer added it to all of them?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/transformers/EventStreamNormalizer.kt
```diff
@@ -37,3 +80,40 @@ object EventStreamNormalizer {
...
+    return Walker(model).walkShapes(shape)
+        .filter { it is UnionShape && it.isEventStream() }
+        .map { it.asUnionShape().get() }
+        .associateWith { unionShape ->
+            unionShape.expectTrait<SyntheticEventStreamUnionTrait>().errorMembers
+                .map { model.expectShape(it.target).asStructureShape().get() }
+        }
+}
+
+fun UnionShape.inputEventStreamErrors(): List<Shape> {
```
jdisanti, line 100: Should this be renamed to `eventStreamErrors` since it's the same logic for both inputs and outputs?

rust-runtime/aws-smithy-http/src/event_stream/sender.rs
```diff
@@ -3,77 +3,122 @@
...
+    pub fn meta(&self) -> &aws_smithy_types::Error {
+        &self.meta
+    }
+}
+
+impl StdError for MessageStreamError {}
+impl fmt::Display for MessageStreamError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self.kind {
+            MessageStreamErrorKind::Unhandled(inner) => std::fmt::Debug::fmt(inner, f),
```
jdisanti, line 91: Shouldn't this be formatting it with `Display` instead?

82marbag: Yes thanks

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/config/EventStreamSigningConfig.kt
```diff
@@ -0,0 +1,63 @@
...
+    open fun signer(): Writable = emptySection
+
+    override fun section(section: ServiceConfig): Writable {
+        return when (section) {
+            is ServiceConfig.ConfigImpl -> inner()
+            else -> emptySection
+        }
+    }
+
+    open fun inner(): Writable {
```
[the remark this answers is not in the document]
82marbag: Looks good, thanks!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
```diff
@@ -30,12 +35,35 @@ open class ServerCombinedErrorGenerator(
...
+                }
+        }
+    }
+
+    open fun renderErrors(
+        writer: RustWriter,
+        errors: List<StructureShape>,
+        errorSymbol: RuntimeType,
+        operationSymbol: Symbol
+    ) {
```
crisidev, line 66: It doesn't look like, but please let's be sure the changes on this function will not break the Python implementation. The `render()` function signature has changed, so that will have to be fixed.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestJson.kt
```diff
@@ -99,7 +99,7 @@ class RestJson(private val codegenContext: CodegenContext) : Protocol {
     private val jsonDeserModule = RustModule.private("json_deser")
 
     override val httpBindingResolver: HttpBindingResolver =
-        RestJsonHttpBindingResolver(codegenContext.model, ProtocolContentTypes.consistent("application/json"))
+        RestJsonHttpBindingResolver(codegenContext.model, ProtocolContentTypes("application/json", "application/json", "application/vnd.amazon.eventstream"))
```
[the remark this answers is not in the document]
crisidev: I think it does.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/EventStreamErrorMarshallerGenerator.kt
```diff
@@ -0,0 +1,168 @@
...
+import software.amazon.smithy.rust.codegen.smithy.generators.unknownVariantError
+import software.amazon.smithy.rust.codegen.smithy.rustType
+import software.amazon.smithy.rust.codegen.smithy.traits.SyntheticEventStreamUnionTrait
+import software.amazon.smithy.rust.codegen.smithy.transformers.eventStreamErrors
+import software.amazon.smithy.rust.codegen.util.dq
+import software.amazon.smithy.rust.codegen.util.expectTrait
+import software.amazon.smithy.rust.codegen.util.hasTrait
+import software.amazon.smithy.rust.codegen.util.toPascalCase
+
+class EventStreamErrorMarshallerGenerator(
```
crisidev, line 39: Let's add some documentation to this class and methods..

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/EventStreamMarshallerGenerator.kt
```diff
@@ -237,7 +243,7 @@ class EventStreamMarshallerGenerator(
         }
     }
 
-    private fun RustWriter.addStringHeader(name: String, valueExpr: String) {
+    fun RustWriter.addStringHeader(name: String, valueExpr: String) {
```
crisidev, line 246: Since this is called by a subclass, should be `protected`?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/EventStreamMarshallerGenerator.kt
```diff
@@ -175,7 +177,12 @@ class EventStreamMarshallerGenerator(
         else -> throw IllegalStateException("unsupported event stream header shape type: $target")
     }
 
-    private fun RustWriter.renderMarshallEventPayload(inputExpr: String, member: MemberShape, target: Shape) {
+    fun RustWriter.renderMarshallEventPayload(
```
crisidev, line 180: Since this is called by a subclass, should be protected?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/EventStreamMarshallerGenerator.kt
```diff
@@ -134,15 +134,17 @@ class EventStreamMarshallerGenerator(
...
-            renderMarshallEventPayload("inner", unionMember, eventStruct)
+            val serializerFn = serializerGenerator.payloadSerializer(unionMember)
+            renderMarshallEventPayload("inner", unionMember, eventStruct, serializerFn)
         } else {
             rust("Vec::new()")
         }
     }
 
-    private fun RustWriter.renderMarshallEventHeader(memberName: String, member: MemberShape, target: Shape) {
+    fun RustWriter.renderMarshallEventHeader(memberName: String, member: MemberShape, target: Shape) {
```
crisidev, line 147: Since this is called by a subclass, should be protected?

rust-runtime/aws-smithy-http-server/examples/pokemon_service/tests/simple_integration_test.rs
```diff
@@ -10,15 +10,42 @@
 use std::time::Duration;
 
 use crate::helpers::{client, PokemonService};
+use async_stream::stream;
```
crisidev, line 13: Be aware the this file is currently symlinked inside the Python folder. You will need to copy the original implementation there since event streams are not supported yet.

crisidev (approved):
LGTM. I have added some minor comments. This PR will break as soon as you merge in the Python server, so some other adjustments are needed.

One thing I suggest you to do is to throw a CodegenException if EventStreams are used in model built with the Python plugin.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGenerator.kt
82marbag: `render()` hasn't changed actually, if you were using `render()` it should keep working the same way

crisidev: Ok, great! We are using directly `render()`

rcoh (changes requested):
Wow! This is an incredible piece of engineering with a lot of cool ideas baked in. Really impressed with how quick this came together.

In the current state of the PR I can't really evaluate the changes, especially the changes around errors for event streams. Here's what I'd suggest:

- An RFC / design doc (doesn't need to be long) or an addition to the docs folder explaining different union shape structures and the event stream types we plan to generate
- As possible, split out the most complex pieces of code into standalone PRs with tests, eg. the changes to EventStream normalization can stand alone I think?

rust-runtime/aws-smithy-eventstream/src/frame.rs
```diff
@@ -27,7 +27,19 @@ pub type SignMessageError = Box<dyn StdError + Send + Sync + 'static>;
 pub trait SignMessage: fmt::Debug {
     fn sign(&mut self, message: Message) -> Result<Message, SignMessageError>;
 
-    fn sign_empty(&mut self) -> Result<Message, SignMessageError>;
+    fn sign_empty(&mut self) -> Option<Result<Message, SignMessageError>>;
```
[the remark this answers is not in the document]
82marbag: Added

design/src/smithy/event_streams.md
```diff
@@ -0,0 +1,302 @@
...
+
+The marshallers set header and payload in a [Message](https://github.com/awslabs/smithy-rs/blob/8f7e03ff8a84236955a65dba3d21c4bdbf17a9f4/rust-runtime/aws-smithy-eventstream/src/frame.rs#L368) structure: `Message` is a structure with a vector, the headers; and bytes, the payload.
+The headers are the values targeted by the `@eventHeader` trait and the payload by `@eventPayload`.
+
+At the end of the marshalling and signing processes, `MessageStreamAdapter` takes the `Message` built by the marshaller and signer
+and [writes](https://github.com/awslabs/smithy-rs/blob/8f7e03ff8a84236955a65dba3d21c4bdbf17a9f4/rust-runtime/aws-smithy-eventstream/src/frame.rs#L224) it as bytes into a `Vec<u8>`,
+in a format of a sequence of bytes: `<type, data>` where `type` indicates if the `data` is a bool, integer and so on for all types.
+
+Headers that are always sent are:
```
jdisanti, line 281: Would probably remove the word "always" here since some of them are only sent on certain conditions.

design/src/smithy/event_streams.md
```diff
@@ -0,0 +1,302 @@
...
+* `:message-type` (`event` or `exception`) to signal the kind of message being sent
+* `:content-type`, currently always set to `application/octet-stream`
```
jdisanti, line 283: I don't think this is correct? If sending a `Blob` as the `@eventPayload`, then yes. But if it's a `structure`, then it will have the content type of the protocol (e.g., `application/json` for `@restJson1`).

design/src/smithy/event_streams.md
```diff
@@ -0,0 +1,302 @@
...
+   1. Signal the end of the stream, if the event is `None`
+   2. Marshall either the message or the error, depending on the data coming from the stream
+   3. Sign the marshalled data and return it
+3. Otherwise, signal to poll back later
+
+Similarly, the [Receiver](https://github.com/awslabs/smithy-rs/blob/8f7e03ff8a84236955a65dba3d21c4bdbf17a9f4/rust-runtime/aws-smithy-http/src/event_stream/receiver.rs#L125) handles the receiving side of the stream.
+The `Receiver` has more logic to handle buffering data and possibly errors.
+
+Server and client serialize similarly. Serializing for `CapturePokemonOperation` on the server, with `serialize_capture_pokemon_operation_response`:
+1. Sets the `content-type` header to `application/vnd.amazon.eventstream`
```
jdisanti, line 255: Should clarify:
```suggestion
1. Sets the `content-type` HTTP header to `application/vnd.amazon.eventstream`
```

design/src/rfcs/rfc0017_event_streams_errors.md
```diff
@@ -0,0 +1,183 @@
+<!-- Give your RFC a descriptive name saying what it would accomplish or what feature it defines -->
+RFC: Errors for event streams
```
jdisanti, line 2: Some of the changes proposed here need some breaking changelog entries for the SDK.

jdisanti (commented):
This is looking pretty good! I think we're getting close to being able to merge. Thanks a ton for adding the documentation!

design/src/rfcs/rfc0017_event_streams_errors.md
```diff
@@ -0,0 +1,183 @@
...
+=============
+
+<!-- RFCs start with the "RFC" status and are then either "Implemented" or "Rejected".  -->
+> Status: RFC
+
+<!-- A great RFC will include a list of changes at the bottom so that the implementor can be sure they haven't missed anything -->
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+<!-- Insert a short paragraph explaining, at a high level, what this RFC is for -->
+This RFC defines how customers will use errors defined in `@streaming` unions (event streams).
```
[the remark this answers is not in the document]
82marbag: Yes

jdisanti (commented):
Very nice work! 🚀 

I checked out the changes locally and verified that the [Amazon Transcribe streaming example](https://github.com/awsdocs/aws-doc-sdk-examples/tree/main/rust_dev_preview/transcribestreaming) still works correctly. Also looked through the generated documentation to see how the errors look there. Everything looks good to me.

Overall, I'm happy with these changes. The only snag left is that the codegen diff doesn't show a diff for a full service thanks to #1219. I think if we get #1565 merged, and then update this with the latest main, the diff will be small enough that we can see a full service diffed. Seeing that would be a nice sanity check.

I think we're very close to getting this merged. @rcoh or @Velfi, feel free to approve on my behalf if I'm not available to review with the changelog and codegen diff updates in.

82marbag:
I tried #1565 here, but the diff of errors is still as large (20k lines). Instead, with and without #1565, I generated the whole SDK for both `main` and `eventstreams`, sorted all files and created a diff.
[This](https://pastebin.com/2VMSWr9x) is the diff without #1565. It's not large and it can be reviewed manually; the only changes are related to event streams (mainly additions). 
@rcoh suggested to run semver check as well and it's all PASS

jdisanti (approved):

rcoh (approved):

PR #1480, opened 2022-06-21
Add Japanese flavor text to the example pokemon service

booniepepper:
## Motivation and Context

The game was made in Japan... so why not add the Japanese flavor text?

## Description

Uses the original flavor text from the Pokemon Red/Blue era like the others. Text came from [ポケモンWiki](https://wiki.ポケモン.com/wiki/%E3%83%94%E3%82%AB%E3%83%81%E3%83%A5%E3%82%A6#.E3.83.9D.E3.82.B1.E3.83.A2.E3.83.B3.E3.81.9A.E3.81.8B.E3.82.93.E3.81.AE.E8.AA.AC.E6.98.8E.E6.96.87) and is the non-kanji version like what appears in a gameboy.

## Testing

`cargo test` in the rust-runtime directory

## Checklist


hiljusti:
Sorry for the noise - I think I satisfied all the tests now. Not sure what I can do for S3 credentials (if anything)

crisidev:
Some of the CI cannot run from forks, so we are good here, don't you worry!

PR #1484, opened 2022-06-22
Fix server routing matching URIs with prefixes

### LINKED ISSUE #1483: Routing incorrectly matches URIs with prefixes
I discovered this while testing integration with Lambda behind API Gateway, where the deployed API contains the API stage in the URI:

Both these cases fail:

```rust
#[test]
fn foo() {
    let get_server_statistics_request_spec = RequestSpec::new(
        Method::GET,
        UriSpec::new(PathAndQuerySpec::new(
            PathSpec::from_vector_unchecked(vec![PathSegment::Literal(String::from("stats"))]),
            QuerySpec::from_vector_unchecked(vec![]),
        )),
    );

    let misses = vec![
        (Method::GET, "/beta/stats"),
        (Method::GET, "/multiple/stages/in/path/stats"),
    ];
    for (method, uri) in &misses {
        assert_eq!(
            Match::No,
            get_server_statistics_request_spec.matches(&req(method, uri, None))
        );
    }
}
```

david-perez:
We are incorrectly matching URIs to the spec when the URI has a prefix
that is not in the spec.

This is because when generating the regex describing the spec, we're not
using the `^` anchor to match from the beginning of the text.

Fixes #1483.

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
+++ b/rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
@@ -98,7 +98,7 @@ impl From<&PathSpec> for Regex {
     fn from(uri_path_spec: &PathSpec) -> Self {
         let sep = "/";
         let re = if uri_path_spec.0.is_empty() {
-            String::from("/")
+            String::from(sep)
         } else {
             uri_path_spec
                 .0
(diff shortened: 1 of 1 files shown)


crisidev (approved):
Nice catch!

PR #1485, opened 2022-06-22
Bubble up unknown enum variant error when parsing XML in the server

### LINKED ISSUE #1477: Server RestXML (S3) compile error trying to init enum from string without using from_str
This is the problematic line - It generates code that tries to init the enum from a string, but using from() instead of from_str() -

https://github.com/awslabs/smithy-rs/blob/a91b813e43c01606d6ba5c01353adea6a1fe417d/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/XmlBindingTraitParserGenerator.kt#L663

I wanted to fix it but found that I need to add `use std::str::FromStr` and I can't find how to add this to the entire file, and adding it per function is tedious because there are many different function types.

This also highlights that #1161 is needed to keep from regressions for restXML.

david-perez:
XML deserialization of enums is currently broken.

This commit should have been done as part of #1398, but we only tested
that patch using the restJson1 protocol.

The converter from the unknown enum variant error into
`aws_smithy_json::deserialize::Error` has been removed from
`ServerEnumGenerator`, since it's protocol-specific logic. We instead
use `map_err` in the protocol-specific parsers before bubbling up using
`?`.

Fixes #1477.

## Testing

I have tested XML using the following diff on the `simple.smithy` model. But,
as mentioned in #1477, we should add the `rest_xml` model to our protocol test
suite, so as to not regress or introduce bugs in XML when making changes to the
server codegen. That should be done in a follow-up PR.

```diff
diff --git a/codegen-server-test/model/simple.smithy b/codegen-server-test/model/simple.smithy
index d4420787..bd543934 100644
--- a/codegen-server-test/model/simple.smithy
+++ b/codegen-server-test/model/simple.smithy
@@ -2,10 +2,12 @@ $version: "1.0"
 
 namespace com.amazonaws.simple
 
-use aws.protocols#restJson1
+use aws.protocols#restXml
+//use aws.protocols#restJson1
 use smithy.test#httpRequestTests
 use smithy.test#httpResponseTests
 
+@restXml
 @restJson1
 @title("SimpleService")
 @documentation("A simple service example, with a Service resource that can be registered and a readonly healthcheck")
@@ -120,12 +122,42 @@ structure StoreServiceBlobInput {
     @required
     @httpLabel
     id: ServiceId,
-    @required
-    @httpPayload
-    content: Blob,
+
+    enum: MyString,
+
+    string: String,
 }
 
 @documentation("Store a blob for a service id output structure")
 structure StoreServiceBlobOutput {
-
+    enum: MyString
 }
+
+@enum([
+    {
+        value: "t2.nano",
+        name: "T2_NANO",
+        documentation: """
+            T2 instances are Burstable Performance
+            Instances that provide a baseline level of CPU
+            performance with the ability to burst above the
+            baseline.""",
+        tags: ["ebsOnly"]
+    },
+    {
+        value: "t2.micro",
+        name: "T2_MICRO",
+        documentation: """
+            T2 instances are Burstable Performance
+            Instances that provide a baseline level of CPU
+            performance with the ability to burst above the
+            baseline.""",
+        tags: ["ebsOnly"]
+    },
+    {
+        value: "m256.mega",
+        name: "M256_MEGA",
+        deprecated: true
+    }
+])
+string MyString
```


rcoh (approved):

crisidev (approved):
Thanks for fixing this.

PR #1491, opened 2022-06-23
Rewrite `ServerOperationRegistryGenerator`

david-perez:
This code was written a long time ago, and made heavy use of string
interpolation instead of the more robust Rust templating primitives from
`RustWriter.kt`.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationRegistryGenerator.kt
```diff
@@ -108,146 +111,146 @@ class ServerOperationRegistryGenerator(
...
+            }
+
+            rustBlock("pub fn build(self) -> Result<$operationRegistryNameWithArguments, ${operationRegistryErrorName}>") {
+                withBlock("Ok( $operationRegistryName {", "})") {
+                    for (operationName in operationNames) {
+                        rust(
+                            """
+                            $operationName: match self.$operationName {
+                                Some(v) => v,
+                                None => return Err(${operationRegistryErrorName}::UninitializedField("operationName")),
```
crisidev, line 186: There is an error here:

```suggestion
                                None => return Err(${operationRegistryErrorName}::UninitializedField("$operationName")),
```

crisidev (changes requested):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationRegistryGenerator.kt
david-perez: Good catch!

crisidev (approved):

PR #1495, opened 2022-06-24
Decouple `aws-sdk-rust` release from `smithy-rs` release

jdisanti:
## Motivation and Context
This PR makes it possible to release smithy-rs separately from aws-sdk-rust, which allows for multiple SDK releases per code generator release so that we will be able to update SDK service models more frequently in the future without needlessly releasing smithy-rs.

I recommend reviewing each individual commit in the PR as that will make it easier to understand what's going on.

At a high level, this PR:
- Removes the concept of smoketest/full-sdk modes in `aws:sdk:assemble`. Now, instead, there is just a model path, and all the models from that path are used unless an explicit list is provided.
- Moves the changelog render tool out of `sdk-lints` and into its own independent tool. This new tool also has a `split` subcommand that is used to break SDK changelog entries away from smithy-rs changelog entries when a smithy-rs release is cut. The SDK release process then incorporates those split out entries the next time its code generator is updated. Entry duplication is avoided by tacking on a `since_commit` field for each SDK changelog entry when the logs are split.
- Refactors the tools to have one shared implementation for interacting with Git, where there used to be two different implementations.
- Adds a `release.yml` GitHub Action workflow that automatically releases smithy-rs upon workflow dispatch. This can be kicked off via GitHub CLI or through the website. It takes a `dry-run` input, which will produce release artifacts for inspection, but not cut the release and publish to crates.io if set.
- Updates `sdk-sync` to use models from aws-sdk-rust rather than from smithy-rs.

## Testing
- Ran dry run releases for both aws-sdk-rust and smithy-rs with these changes and verified the artifacts to be published looked correct
- Wrote the dry runs in such a way that they validate the secrets used so that when it comes time to do a real release, that aspect should be working
- Verified tool/build changes are backwards/forwards compatible enough that `sdk-sync` doesn't break when syncing older/newer commits. This resulted in some TODO comments that will be addressed in https://github.com/awslabs/smithy-rs/issues/1493


Velfi (approved):
A monumental effort! Good work.

tools/changelogger/src/render.rs
```diff
@@ -0,0 +1,576 @@
...
+pub struct RenderArgs {
+    /// Which set of changes to render
+    #[clap(long, action)]
+    pub change_set: ChangeSet,
+    /// Whether or not independent crate versions are being used (defaults to false)
+    #[clap(long, action)]
+    pub independent_versioning: bool,
+    /// Source changelog entries to render
+    #[clap(long, action)]
+    pub source: Vec<PathBuf>,
```
drganjoo, line 55: Can we make this a required parameter please - thanks

jdisanti: I think you mean have a minimum of one? Good call out. I'll take a look at what clap offers for this.

jdisanti: Looks like `required(true)` does the job.

Issue #1164, opened 2022-02-07
Server protocol tests extras


david-perez:
The client has suites like [`rest-json-extras.smithy`](https://github.com/awslabs/smithy-rs/blob/25c50a5f0b6ac6b76bd373ad09bcba4841e4d27c/codegen-test/model/rest-json-extras.smithy#L10) that contains extra protocol tests, some of which are upstreamed in the awslabs/smithy repo and will be hosted there in the next Smithy release.

Ideally the server should also pass these suites, and we should put there additional tests we come up with.

Fixed tests should also go there, instead of hot-patching them at runtime [like we currently do](https://github.com/awslabs/smithy-rs/blob/25c50a5f0b6ac6b76bd373ad09bcba4841e4d27c/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt#L679-L679). Advantages:
* We can more easily upstream the fixed test to awslabs/smithy repo since we write Smithy and not Kotlin.
* When we upgrade to the next Smithy release, the broken ones in `ExpectFailing` will fail because they now pass, so we will get alerted and remove the fixed tests from the "extra" suite.

david-perez:
A first test we should add to not regress on #1163 (but to the extras suite, not to `simple.smithy` like the diff below shows):

```diff
diff --git a/codegen-server-test/model/simple.smithy b/codegen-server-test/model/simple.smithy
index 70dee327..2feae1d2 100644
--- a/codegen-server-test/model/simple.smithy
+++ b/codegen-server-test/model/simple.smithy
@@ -63,7 +63,10 @@ resource Service {
         id: "RegisterServiceResponseTest",
         protocol: "aws.protocols#restJson1",
         params: { id: "1", name: "TestService" },
-        body: "{\"id\":\"1\",\"name\":\"TestService\"}",
+        headers: {
+            "Content-Type": "TestService",
+        },
+        body: "{\"id\":\"1\"}",
         code: 200,
     }
 ])
@@ -85,6 +88,9 @@ structure RegisterServiceInputRequest {
 structure RegisterServiceOutputResponse {
     @required
     id: ServiceId,
+
+    @required
+    @httpHeader("Content-Type")
     name: ServiceName,
 }
```

david-perez:
Another one: `httpQueryParams` duplicate query keys behavior: https://github.com/awslabs/smithy/issues/1071

[...]

PR #1497, opened 2022-06-24
Add server extra tests

82marbag:
## Motivation and Context

See: https://github.com/awslabs/smithy-rs/issues/1164

## Description

Tests:
* https://github.com/awslabs/smithy-rs/issues/1164#issuecomment-1031385029
* https://github.com/awslabs/smithy/pull/1243

## Checklist


codegen-test/model/rest-json-extras.smithy
```diff
@@ -322,9 +335,118 @@ structure EmptyStructWithContentOnWireOpOutput {
...
+            }
+        },
+        response: {
+            code: 406,
+            headers: {
+                "x-amzn-errortype": "NotAcceptableException"
+            }
+        },
+        tags: [ "accept" ],
+        appliesTo: "server",
```
david-perez, line 368: `httpMalformedRequestTests` only apply to servers, this line is not needed.

codegen-test/model/rest-json-extras.smithy
```diff
@@ -322,9 +335,118 @@ structure EmptyStructWithContentOnWireOpOutput {
...
+    }
+])
+
+@suppress(["UnstableTrait"])
```
david-perez, line 372: What is being suppressed here?

codegen-test/model/rest-json-extras.smithy
```diff
@@ -322,9 +335,118 @@ structure EmptyStructWithContentOnWireOpOutput {
...
+        params: { id: "1", name: "TestService" },
+        headers: {
+            "Content-Type": "TestService",
+        },
+        body: "{\"id\":\"1\"}",
+        code: 200,
+        appliesTo: "server",
+    }
+])
+operation FixedRegisterService {
```
david-perez, line 413: What are these tests testing that is not already being tested in the `rest_json` suite?

codegen-test/model/rest-json-extras.smithy
```diff
@@ -93,7 +95,8 @@ structure ExtraError {}
         body: "rawstring",
         params: { payload: "rawstring" },
         method: "POST",
-        protocol: "aws.protocols#restJson1"
+        protocol: "aws.protocols#restJson1",
+        appliesTo: "client",
```
david-perez, line 99: Why all these `appliesTo: "client"`s? Many of these should run / can be made to run correctly for the server implementation too.

david-perez (changes requested):

codegen-test/model/rest-json-extras.smithy
82marbag: For #1164, but will address that in another PR

codegen-test/model/rest-json-extras.smithy
82marbag: Removed the unnecessary ones, thanks

crisidev:
Is this PR still relevant?

codegen-test/model/rest-json-extras.smithy
```diff
@@ -78,7 +78,8 @@ service RestJsonExtras {
         code: 500,
         body: "",
         headers: { "X-Amzn-Errortype": "ExtraError" },
-        params: {}
+        params: {},
+        appliesTo: "client",
```
david-perez, line 82: This test fails for the server because [servers are expected to serialize `{}` when the output parameters are empty](https://github.com/awslabs/smithy/blob/69cc7ba87b62da026130876d3e306194f39759ee/smithy-aws-protocol-tests/model/restJson1/errors.smithy#L32-L36).

Hence why we have a [`serverStructureSerializer`](https://github.com/awslabs/smithy-rs/blob/6c21fb0eb377c7120a8179f4537ba99a4b50ba96/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/serialize/JsonSerializerGenerator.kt#L173-L173) that behaves like that. We should arguably modify the Kotlin docs for `serverStructureSerializer` calling this out.

Anyway, this test can remain for the client only, but we should duplicate in a new test that only applies to the server and that has `body: "{}"`, and copy over the documentation from `RestJsonGreetingWithErrors` to justify why we're doing so.

codegen-test/model/rest-json-extras.smithy
```diff
@@ -201,7 +203,8 @@ structure MapWithEnumKeyInputOutput {
         method: "POST",
         protocol: "aws.protocols#restJson1",
         body: "{\"map\":{\"enumvalue\":\"something\"}}",
-        params: { map: { "enumvalue": "something" } }
+        params: { map: { "enumvalue": "something" } },
+        appliesTo: "client",
```
david-perez, line 207:
```suggestion
        headers: { "Content-Type": "application/json" },
```

codegen-test/model/rest-json-extras.smithy
```diff
@@ -240,6 +243,7 @@ structure EscapedStringValuesInputOutput {
         protocol: "aws.protocols#restJson1",
         body: "{\"enum\":\"has\\\"quotes\",\"also\\\"has\\\"quotes\":\"test\"}",
         params: { enum: "has\"quotes", someString: "test" },
+        appliesTo: "client",
```
david-perez, line 246:
```suggestion
        headers: { "Content-Type": "application/json" },
```

codegen-test/model/rest-json-extras.smithy
```diff
@@ -113,7 +114,8 @@ structure StringPayloadInput {
     uri: "/primitive-document",
     method: "POST",
     body: "{}",
-    params: {}
+    params: {},
+    appliesTo: "client",
```
david-perez, line 118:
```suggestion
    headers: { "Content-Type": "application/json" },
```

This test fails for the server because it's an invalid restJson1 request. This should fix this test and the other ones where I've made the same suggestion.

Place `headers` below `body` though like in the awslabs/smithy tests; the GitHub UI does not allow me to make that suggestion.

codegen-test/model/rest-json-extras.smithy
```diff
@@ -302,7 +307,8 @@ operation CaseInsensitiveErrorOperation {
         code: 500,
         body: "{\"Message\": \"hello\"}",
         headers: { "X-Amzn-Errortype": "CaseInsensitiveError" },
-        params: { message: "hello" }
+        params: { message: "hello" },
+        appliesTo: "client",
```
david-perez, line 311: I would expand the documentation to justify why this is a client-only test; took me a while to figure out why. It turns out that smithy-rs clients should be able to parse error messages if the JSON keys them as `message`, `Message`, or `errorMessage`. The relevant parsing code is [here](https://github.com/awslabs/smithy-rs/blob/6c21fb0eb377c7120a8179f4537ba99a4b50ba96/rust-runtime/inlineable/src/json_errors.rs#L51-L51).

david-perez (changes requested):

codegen-test/model/rest-json-extras.smithy
82marbag: Added a comment with the link

david-perez (approved):
nit: i'd place documentation inside the `documentation` field instead of in a comment in the Smithy file, so that it gets rendered in the protocol test.

jdisanti (approved):

PR #1500, opened 2022-06-27
SERVER_CHANGELOG.md has been added for v0.44.0

drganjoo:
## Motivation and Context
SERVER_CHANGELOG.md  has been added (generated through the changelogger tool)

## Description
Server side SDK has it's own SERVER_CHANGELOG.md that will have the server side changes carried out in a revision.

## Checklist


jdisanti:
Let's talk about how we should go about this. My initial impression is that this shouldn't be a separate `.next.toml` since we can make changes that affect both the client and the server, and we shouldn't have to copy/paste entries into both files to represent those. Also, GitHub can only have one set of notes for a release, so these really should be combined.

I think we can just add a property to the current changelog toml format that indicates the change is for client, server, or both. Then we can just modify how we render the changelog so that it breaks out separate client/server sections.

codegen-server/SERVER_CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+<!-- Do not manually edit this file. Use the `changelogger` tool. -->
+v0.44.0 (June 23rd, 2022)
+=========================
+**Breaking Changes:**
+- ⚠ ([smithy-rs#1484](https://github.com/awslabs/smithy-rs/issues/1484)) Server routing for URIs with prefixes has been fixed and now URIs are matched from the beginning
+- ⚠ ([smithy-rs#1456](https://github.com/awslabs/smithy-rs/issues/1456)) Deprecated v1 arg format from sdk-versioner have been removed
+- ⚠ ([smithy-rs#1424](https://github.com/awslabs/smithy-rs/issues/1424)) An error is raised if the http `Accept` header does not match the content type specified in smithy file
+
+**New this release:**
+- 🎉 ([smithy-rs#1454](https://github.com/awslabs/smithy-rs/issues/1454)) Python can be used for writing business logic, executing on top of smithy-rs Rust runtime
```
crisidev, line 10: This PR is not related to Python, I think you made a mistake linking it. It should be https://github.com/awslabs/smithy-rs/pull/1403

On top of this, the wording of the description is not 100% correct as until https://github.com/awslabs/smithy-rs/pull/1476 is not merged, the Python system is not usable. 

I would reword it like: `Initial support to code-generate support to write the business logic in Python, executing on top of Rust. This PR introduce code-generation of operations and simple shapes (Structure and Enum) and add runtime support to start the Hyper server from Python, but does not generate an importable shared library yet.`

crisidev (changes requested):

crisidev:
Is this PR still relevant?

drganjoo:
This PR is no longer needed as changelog tool has now been updated to include server SDK changes.

PR #1501, opened 2022-06-27
Feature: add support for aws-chunked content encoding

Velfi:
## Motivation and Context

Supports the [flexible checksums](https://awslabs.github.io/smithy/1.0/spec/aws/aws-core.html#aws-protocols-httpchecksum-trait) feature. When sending streaming requests we need to include checksums as trailers and we need `aws-chunked` content encoding to do that.

## Description

This provides functionality specified in the [flexible checksums RFC](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0016_flexible_checksum_support.md)

## Testing

I wrote a test

## Checklist


aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,325 @@
...
+            AwsChunkedBodyState::WritingTrailers => {
+                return match this.inner.poll_trailers(cx) {
+                    Poll::Ready(Ok(trailers)) => {
+                        *this.state = AwsChunkedBodyState::Closed;
+                        let total_length_of_trailers_in_bytes =
+                            this.options.trailer_lens.iter().sum();
+
+                        Poll::Ready(Some(Ok(trailers_as_aws_chunked_bytes(
+                            total_length_of_trailers_in_bytes,
+                            trailers,
```
rcoh, line 230: is there any check that these trailers actually match this.options.trailer_lens?

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// Chunked-Body    = *chunk
+    ///                   last-chunk
+    ///                   chunked-trailer
+    ///                   CRLF
+    ///
+    /// chunk           = chunk-size CRLF chunk-data CRLF
+    /// chunk-size      = 1*HEXDIG
+    /// last-chunk      = 1*("0") CRLF
+    /// chunked-trailer = *( entity-header CRLF )
+    /// entity-header   = field-name ":" OWS field-value OWS
```
rcoh, line 86: should explain somewhere that this will always write the request into a single chunk

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,325 @@
...
+    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
+        tracing::trace!("polling AwsChunkedBody");
+        let mut this = self.project();
+
+        match *this.state {
+            AwsChunkedBodyState::WritingChunkSize => {
+                tracing::trace!("writing chunk size");
+                *this.state = AwsChunkedBodyState::WritingChunk;
+                // A chunk must be prefixed by chunk size in hexadecimal
+                let chunk_size = Bytes::from(format!("{:X?}\r\n", this.options.stream_length));
```
rcoh, line 205: you define the CRLF constant above but don't actually use it here

rcoh (changes requested):
looking good so far! the state machine is very clean, just a couple spots that could be cleaned up / clarified

CHANGELOG.next.toml
```diff
@@ -21,3 +21,9 @@ message = "Add a trailing slash to the URI `/latest/meta-data/iam/security-crede
 references = ["aws-sdk-rust#560", "smithy-rs#1487"]
 meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "rcoh"
+
+[[smithy-rs]]
```
jdisanti, line 25: This is a SDK feature rather than a Smithy feature.
```suggestion
[[aws-sdk-rust]]
```

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// Create a new [`AwsChunkedBodyOptions`][AwsChunkedBodyOptions]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Set stream length
+    pub fn with_stream_length(mut self, stream_length: u64) -> Self {
+        self.stream_length = stream_length;
+        self
+    }
```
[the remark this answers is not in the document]
Velfi: yeah size hint doesn't account for trailers

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,325 @@
...
+        self.state == AwsChunkedBodyState::Closed
+    }
+
+    fn size_hint(&self) -> SizeHint {
+        SizeHint::with_exact(
+            self.encoded_length()
+                .expect("Requests made with aws-chunked encoding must have known size")
+                as u64,
+        )
+    }
```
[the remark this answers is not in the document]
Velfi: We can only check that the encoded body is larger (due to chunk prefix, CRLF, trailers) than the size hint of the inner body but I can't think of how to test that other than by constructing a body that lies about its size.

aws/rust-runtime/aws-http/src/content_encoding.rs
Velfi: I added one

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,325 @@
...
+    pub fn with_stream_length(mut self, stream_length: u64) -> Self {
+        self.stream_length = stream_length;
+        self
+    }
+
+    /// Set a trailer len
+    pub fn with_trailer_len(mut self, trailer_len: u64) -> Self {
+        self.trailer_lens.push(trailer_len);
+        self
+    }
```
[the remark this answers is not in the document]
Velfi: I updated the API so you can input a vec when creating the Options

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,325 @@
...
+#[non_exhaustive]
+pub struct AwsChunkedBodyOptions {
+    /// The total size of the stream. Because we only support unsigned encoding
+    /// this implies that there will only be a single chunk containing the
+    /// underlying payload.
+    pub stream_length: u64,
+    /// The length of each trailer sent within an `AwsChunkedBody`. Necessary in
+    /// order to correctly calculate the total size of the body accurately.
+    pub trailer_lens: Vec<u64>,
+}
```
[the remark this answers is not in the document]
Velfi: I refactored the API, let me know what you think

CHANGELOG.next.toml
```diff
@@ -21,3 +21,9 @@ message = "Add a trailing slash to the URI `/latest/meta-data/iam/security-crede
...
+message = "Add support for aws-chunked content encoding"
```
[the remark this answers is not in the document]
Velfi: sure, good callout

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -276,6 +282,30 @@ where
...
+
+                let mut values = header_map.get_all(key).into_iter();
+
+                if let Some(value) = values.next() {
+                    total += value.len() + CRLF.len();
+                }
+
+                for value in values {
+                    total += value.len() + ",".len();
+                }
```
jdisanti, line 300: Does this match the actual serialization behavior for multiple values? I thought that hyper would add multiple of the same header name with a different value, but I'm not 100% sure.

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -276,6 +282,30 @@ where
     len
 }
 
+fn total_rendered_length_of_trailers(header_map: Option<&HeaderMap>) -> u64 {
```
jdisanti, line 285: Seems like we should have a unit test that serializes the header map with the `http` crate and compares its length against this function. Especially with multiple headers of the same name.

aws/rust-runtime/aws-http/src/content_encoding.rs
Velfi: I'm not sure if it matches what Hyper would do in this situation but it I did test it with S3 and it accepted it. I can test an upload the hyper-style headers if you like.

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,407 @@
...
+    Inner: Body<Data = Bytes, Error = aws_smithy_http::body::Error>,
+{
+    type Data = Bytes;
+    type Error = aws_smithy_http::body::Error;
+
+    fn poll_data(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
+        tracing::trace!("polling AwsChunkedBody (state = {:?})", self.state);
```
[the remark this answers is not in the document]
Velfi: TIL

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,407 @@
...
+        Self {
+            stream_length,
+            trailer_lengths,
+        }
+    }
+
+    /// The total size of the stream. Because we only support unsigned encoding
+    /// this implies that there will only be a single chunk containing the
+    /// underlying payload.
+    pub fn stream_length(&self) -> u64 {
```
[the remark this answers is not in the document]
Velfi: I think not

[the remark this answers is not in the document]
Velfi: I ended up removing this method b/c I don't think I need to ever look at that value outside of this module

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -0,0 +1,407 @@
...
+            },
+            AwsChunkedBodyState::WritingTrailers => {
+                return match this.inner.poll_trailers(cx) {
+                    Poll::Ready(Ok(trailers)) => {
+                        *this.state = AwsChunkedBodyState::Closed;
+                        let expected_total_trailer_length =
+                            total_rendered_length_of_trailers(trailers.as_ref());
+                        let actual_total_trailer_length = this.options.total_trailer_length();
+                        assert_eq!(expected_total_trailer_length, actual_total_trailer_length,
+                        "while writing trailers, the expected length of trailers ({expected_total_trailer_length}) differed from the actual length ({actual_total_trailer_length})");
```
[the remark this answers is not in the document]
Velfi: I decided to just make it an error. Let me know what you think

aws/rust-runtime/aws-http/src/content_encoding.rs
Velfi: I added a test that ensures our trailer serializer and trailer length calculator stay in sync

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -143,44 +137,57 @@ fn get_unsigned_chunk_bytes_length(payload_length: u64) -> u64 {
...
+/// - Trailer names with multiple values will be written out one line per value, with the name
+///   appearing on each line.
+fn total_rendered_length_of_trailers(trailer_map: Option<&HeaderMap>) -> u64 {
+    match trailer_map {
+        Some(trailer_map) => trailer_map
+            .iter()
+            .map(|(trailer_name, trailer_value)| {
+                trailer_name.as_str().len()
+                    + TRAILER_SEPARATOR.len()
+                    + trailer_value.to_str().unwrap().len()
```
jdisanti, line 185: The length header the trailer length gets incorporated into is in bytes, not characters, right? So the `to_str().unwrap()` is actually unnecessary and you can just count the bytes?

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -143,44 +137,57 @@ fn get_unsigned_chunk_bytes_length(payload_length: u64) -> u64 {
...
-                // New name, new value
-                Some(header_name) => {
-                    if already_wrote_first_trailer {
-                        // First trailer shouldn't have a preceding CRLF, but every trailer after it should
-                        trailers.write_str(CRLF).unwrap();
-                    } else {
-                        already_wrote_first_trailer = true;
-                    }
+        let mut current_header_name = None;
+        let mut trailers = BytesMut::new();
```
jdisanti, line 149: Not blocking: Since we know the length of buffer needed for trailers (`total_rendered_length_of_trailers + 1`), you might consider using [`reserve`](https://docs.rs/bytes/1.1.0/bytes/struct.BytesMut.html#method.reserve) to reduce allocations.

aws/rust-runtime/aws-http/src/content_encoding.rs
```diff
@@ -276,6 +282,28 @@ where
...
 }
 
+/// Errors related to `AwsChunkedBody`
+#[derive(Debug)]
+enum AwsChunkedBodyError {
+    /// Error that occurs when the sum of `trailer_lengths` set when creating an `AwsChunkedBody` is
+    /// not equal to the actual length of the trailers returned by the inner `http::Body`
+    /// implementor. These trailer lengths are necessary in order to correctly calculate the total
+    /// size of the body for setting the content length header.
+    ReportedTrailerLengthMismatch { actual: u64, expected: u64 },
```
jdisanti, line 292: I see @rcoh suggested making an error for this, but I think this is a case where we have a bug in our code, and there's little to no value for a customer in matching on this error. It's a case that simply should not happen if the code is correct, so an assertion was warranted.

On the flip side, with an assertion, it would silently allow the lengths to mismatch when compiled in release mode, which would likely result in a server-side error response. I think this is OK though, since a bug with this should not have been released in the first place if it was adequately tested.

We might want to proptest header names/values for the size prediction vs. the rendering to be absolutely certain they cannot mismatch. I think they can right now in the case where a header value has Unicode characters in it since the length predictor function is using string length instead of byte length.

jdisanti: Another thing we could consider is just getting rid of the length prediction entirely, just render the trailers to a `BytesMut` when the length is initially needed, and store that `BytesMut` in the `AwsChunkedBody` for later.

aws/rust-runtime/aws-http/src/content_encoding.rs
Velfi: I used `with_capacity`, thanks for suggesting

aws/rust-runtime/aws-http/src/content_encoding.rs
rcoh: @jdisanti only `debug_assert` doesn't fail in release mode

Velfi: I added a new error for when stream_length is set incorrectly

rcoh (approved):
LGTM!

PR #1502, opened 2022-06-27
Feature: checksum related body-wrappers

Velfi:
## Motivation and Context

smithy-rs#1263

## Description

add: an implementation of checksum related body-wrappers and tests

## Testing

unit tests are included

## Checklist


rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,444 @@
...
+use http::header::HeaderName;
+use http::{HeaderMap, HeaderValue};
+use http_body::{Body, SizeHint};
+use pin_project_lite::pin_project;
+
+use std::fmt::Display;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+// pin_project_lite can't handle triple-slash doc comments
```
jdisanti, line 22: Wait, but `///` doc comments are used below?

rust-runtime/aws-smithy-checksums/src/lib.rs
```diff
@@ -5,197 +5,251 @@
...
+pub trait Checksum: Send + Sync {
+    /// Given a slice of bytes, update this checksum's internal state.
+    fn update(&mut self, bytes: &[u8]) -> Result<(), BoxError>;
+    /// "Finalize" this checksum, returning the calculated value as `Bytes` or an error that
+    /// occurred during checksum calculation. To print this value in a human-readable hexadecimal
+    /// format, you can print it using Rust's builtin [formatter].
+    ///
+    /// _**NOTE:** typically in Rust, "finalizing" a checksum in Rust will take ownership of the
+    /// checksum struct. In this method, we clone the checksum's state before finalizing because
+    /// these checksums may be used in a situation where taking ownership is not possible._
```
jdisanti, line 28: Thank you for adding this note!

rust-runtime/aws-smithy-checksums/src/lib.rs
```diff
@@ -5,197 +5,251 @@
...
+        use sha2::Digest;
+        Ok(Bytes::copy_from_slice(
+            self.hasher.clone().finalize().as_slice(),
+        ))
+    }
 
-        header_map.insert(key, value);
+    // Size of the checksum in bytes
+    fn size() -> u64 {
+        32
```
jdisanti, line 177: Not a big deal, but I think all the hash `Digest` types have a `output_size` function you can use for this value (e.g., https://docs.rs/sha2/0.10.2/sha2/trait.Digest.html#tymethod.output_size).

rust-runtime/aws-smithy-types/src/base64.rs
```diff
@@ -151,9 +151,17 @@ fn decode_inner(inp: &str) -> Result<Vec<u8>, DecodeError> {
     Ok(ret)
 }
 
+/// Given the length of some data in bytes, return how many bytes it would take to base64 encode
+/// that data.
+pub fn encoded_length(length: u64) -> u64 {
+    (length + 2) / 3 * 4
```
jdisanti, line 157: I'm unsure about this. Was the `+ 2` intended to round up to a multiple of four? If so, I don't think it will accomplish that. Can you modify `test_base64_encoded_length` to be a `proptest` to show that this works more broadly?

rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,444 @@
...
+pin_project! {
+    /// A `ChecksumBody` will read and calculate a request body as it's being sent. Once the body has"
+    /// been completely read, it'll append a trailer with the calculated checksum."
+    pub struct ChecksumBody<InnerBody> {
+        #[pin]
+        inner: InnerBody,
+        #[pin]
+        checksum: Box<dyn HttpChecksum>,
```
jdisanti, line 30: `HttpChecksum` doesn't have any functions on it that have a `Pin` self type, so this shouldn't need pin projection, right?

rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,444 @@
...
+use pin_project_lite::pin_project;
+
+use std::fmt::Display;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+// pin_project_lite can't handle triple-slash doc comments
+pin_project! {
+    /// A `ChecksumBody` will read and calculate a request body as it's being sent. Once the body has"
+    /// been completely read, it'll append a trailer with the calculated checksum."
```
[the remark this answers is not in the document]
Velfi: Yeah this is a leftover from when I thought I needed to use doc attributes for docs in `pin_project` blocks, thanks for catching it.

rust-runtime/aws-smithy-checksums/src/body.rs
Velfi: Another leftover from when I thought I needed to use doc attributes for docs in `pin_project` blocks, thanks for catching.

rust-runtime/aws-smithy-checksums/src/lib.rs
Velfi: TIL, thanks for pointing that out!

rust-runtime/aws-smithy-types/src/base64.rs
Velfi: I don't remember where this impl came from but it does work. Adding a proptest is a good callout though, so I added one:
```rust
        #[test]
        fn encoded_length_is_correct(v in any::<Vec<u8>>()) {
            let correct = encode(v.as_slice()).len() as u64;
            let ours = encoded_length(v.len() as u64);
            assert_eq!(ours, correct);
        }
```

rust-runtime/aws-smithy-checksums/src/body.rs
Velfi: I'm pretty foggy on this actually. As far as I understand it, because I have to mutate both `ChecksumBody.inner` and `ChecksumBody.checksum` in the poll, I have to pin them. Perhaps @rcoh could explain it in a more correct and helpful way.

rcoh: Box is Unpin so I don't think you need to pin this

rust-runtime/aws-smithy-checksums/src/lib.rs
```diff
@@ -5,197 +5,254 @@
...
+    fn update(&mut self, bytes: &[u8]) -> Result<(), BoxError>;
+    /// "Finalize" this checksum, returning the calculated value as `Bytes` or an error that
+    /// occurred during checksum calculation. To print this value in a human-readable hexadecimal
+    /// format, you can print it using Rust's builtin [formatter].
+    ///
+    /// _**NOTE:** typically in Rust, "finalizing" a checksum in Rust will take ownership of the
+    /// checksum struct. In this method, we clone the checksum's state before finalizing because
+    /// these checksums may be used in a situation where taking ownership is not possible._
+    ///
+    /// [formatter]: https://doc.rust-lang.org/std/fmt/trait.UpperHex.html
```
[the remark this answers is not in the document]
Velfi: Yeah, I was taking a cue from the existing stuff out there but it's definitely not very "final" 😜

what about "get_state"?

rust-runtime/aws-smithy-checksums/src/http.rs
```diff
@@ -0,0 +1,201 @@
...
+pub const CHECKSUM_ALGORITHMS_IN_PRIORITY_ORDER: [&str; 4] =
+    [CRC_32_C_NAME, CRC_32_NAME, SHA_1_NAME, SHA_256_NAME];
+
+/// Checksum algorithms are use to validate the integrity of data. Structs that implement this trait
+/// can be used as checksum calculators. This trait requires Send + Sync because these checksums are
+/// often used in a threaded context.
+pub trait HttpChecksum: Checksum + Send + Sync {
+    /// Either return this checksum as a `HeaderMap` containing one HTTP header, or return an error
+    /// describing why checksum calculation failed.
+    fn headers(&self) -> Result<Option<HeaderMap<HeaderValue>>, BoxError>;
```
rcoh, line 93: can provide a default impl of this

rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,342 @@
...
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
+        let this = self.project();
+        match this.checksum {
+            Some(checksum) => {
+                let poll_res = this.body.poll_data(cx);
+                match &poll_res {
+                    Poll::Ready(Some(Ok(data))) => checksum.update(&data),
+                    _ => (),
+                }
```
rcoh, line 55: IMO, the intention here is a little clearer with `if let`

rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,342 @@
...
+impl Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        match self {
+            Error::ChecksumMismatch { expected, actual } => write!(
+                f,
+                "body checksum mismatch. expected body checksum to be {} but it was {}",
+                hex::encode(expected),
+                hex::encode(actual)
+            ),
+            Self::Taken => write!(f, "this checksum body was accessed after finalization. Once a checksum body is finalized, it should not be accessed again."),
```
rcoh, line 195: see above—we can remove this case by returning `None` from poll_trailers

rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,342 @@
...
+    ) -> Poll<Result<Option<HeaderMap>, Self::Error>> {
+        let this = self.project();
+        let poll_res = this.body.poll_trailers(cx);
+
+        if let Poll::Ready(Ok(maybe_inner_trailers)) = poll_res {
+            let checksum_headers = if let Some(checksum) = this.checksum.take() {
+                checksum.headers()
+            } else {
+                return Poll::Ready(Err(Box::new(Error::Taken)));
+            };
```
rcoh, line 75: let's return `Poll::Ready(Ok(None))` in this case—I think that's going to work with more downstream implementations

rcoh: we can also totally remove the error

rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,342 @@
...
+                )))),
+                None => Poll::Ready(Ok(Some(checksum_headers))),
+            };
+        }
+
+        poll_res
+    }
+
+    fn is_end_stream(&self) -> bool {
+        self.body.is_end_stream()
```
rcoh, line 90: this is actually wrong and **may** cause us to not send trailers:
From the docs:
```
An end of stream means that both poll_data and poll_trailers will return None.
```

rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    fn size_hint(&self) -> SizeHint {
+        self.body.size_hint()
```
rcoh, line 94: so we don't take the trailers into account in the size hint?

rust-runtime/aws-smithy-checksums/src/http.rs
```diff
@@ -0,0 +1,226 @@
...
+/// often used in a threaded context.
+pub trait HttpChecksum: Checksum + Send + Sync {
+    /// Either return this checksum as a `HeaderMap` containing one HTTP header, or return an error
+    /// describing why checksum calculation failed.
+    fn headers(self: Box<Self>) -> HeaderMap<HeaderValue> {
+        let mut header_map = HeaderMap::new();
+        header_map.insert(self.header_name(), self.header_value());
+
+        header_map
+    }
```
rcoh, line 100: is it weird that these are called headers but they are trailers?

rust-runtime/aws-smithy-checksums/src/http.rs
```diff
@@ -0,0 +1,226 @@
...
+        MD5_HEADER_NAME.clone()
+    }
+
+    fn header_value(self: Box<Self>) -> HeaderValue {
+        use md5::Digest;
+        // We clone the hasher because `Hasher::finalize` consumes `self`
+        let hash = self.hasher.finalize();
+        HeaderValue::from_str(&base64::encode(&hash[..]))
+            .expect("will always produce a valid header value from an MD5 checksum")
+    }
```
rcoh, line 186: this duplication makes me wonder if we got this API right

rust-runtime/aws-smithy-checksums/src/http.rs
```diff
@@ -0,0 +1,226 @@
...
+        let actual_size = HttpChecksum::size(&*checksum);
+        assert_eq!(expected_size, actual_size)
+    }
+
+    #[test]
+    fn test_trailer_length_of_sha256_checksum_body() {
+        let checksum = new_from_algorithm(SHA_256_NAME);
+        let expected_size = 66;
+        let actual_size = HttpChecksum::size(&*checksum);
+        assert_eq!(expected_size, actual_size)
```
rcoh, line 224: should we have some tests of actual checksums?

imagining bugs like typoing the checksum algorithm

rcoh, line 224: should we have some tests of actual checksums?

imagining bugs like typoing the checksum algorithm

rust-runtime/aws-smithy-checksums/src/body.rs
Velfi: I think we don't, but I kinda psyched myself out on this one after looking at `hyper::body::Body`'s impl of `HttpBody`. I have tests that read checksum bodies, shouldn't those fail if this is wrong?

rust-runtime/aws-smithy-checksums/src/http.rs
Velfi: If you typo the algorithm, `new_from_algorithm` will panic. Is that enough?

rust-runtime/aws-smithy-checksums/src/http.rs
Velfi: I can actually replace it with a default impl, I'll do that

rust-runtime/aws-smithy-checksums/src/body.rs
Velfi: I fixed this by checking if the inner `HttpChecksum` has been taken. That only happens once we're reading trailers and the inner body has emitted any trailers it contains.

rust-runtime/aws-smithy-checksums/src/http.rs
rcoh: sorry I meant "check actually header value and ensure it is what you expect"

Velfi: Good idea, I added tests for that

rust-runtime/aws-smithy-checksums/src/lib.rs
```diff
@@ -5,197 +5,220 @@
...
+pub mod body;
+pub mod http;
+
+/// Checksum algorithms are use to validate the integrity of data. Structs that implement this trait
+/// can be used as checksum calculators. This trait requires Send + Sync because these checksums are
+/// often used in a threaded context.
+pub trait Checksum: Send + Sync {
+    /// Given a slice of bytes, update this checksum's internal state.
+    fn update(&mut self, bytes: &[u8]);
+    /// "Finalize" this checksum, returning the calculated value as `Bytes` or an error that
```
rcoh, line 19: firstline

rust-runtime/aws-smithy-checksums/src/lib.rs
```diff
@@ -246,15 +269,28 @@ mod tests {
...
     #[test]
     fn test_sha256_checksum() {
-        let mut checksum_callback = Sha256Callback::default();
-        checksum_callback.update(TEST_DATA.as_bytes()).unwrap();
-        let checksum_callback_result = checksum_callback.trailers().unwrap().unwrap();
-        let encoded_checksum = checksum_callback_result.get(SHA_256_NAME).unwrap();
-        let decoded_checksum = header_value_as_checksum_string(encoded_checksum);
+        let mut checksum = Sha256::default();
+        checksum.update(TEST_DATA.as_bytes());
+        let checksum_result = Box::new(checksum).headers();
```
rcoh, line 274: as a side note, I don't think you need to box here because of autoderef

rcoh (commented):
this is looking great! it's nice how the default trait impl really cleaned things up. Only remaining thing is removing the panic on checksum algorithm

rust-runtime/aws-smithy-checksums/src/body.rs
```diff
@@ -0,0 +1,346 @@
...
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Option<HeaderMap<HeaderValue>>, Self::Error>> {
+        self.project().inner.poll_trailers(cx)
+    }
+
+    // Once the inner body returns true for is_end_stream, we still need to
+    // verify the checksum; Therefore, we always return false here.
+    fn is_end_stream(&self) -> bool {
+        false
```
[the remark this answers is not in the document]
Velfi: true, this is leftover from an earlier version and I forgot to update it. Good catch.

Issue #932, opened 2021-12-06
Change `pin_project` to `pin_project_lite`


jdisanti:
change pin_project to pin_project_lite (removes syn / proc macro dependency)

Velfi:
Reopening this as we still have pin-project in a few places.

rcoh:
done

PR #1503, opened 2022-06-27
Update: swap pin-project for pin-project-lite

Velfi:
## Motivation and Context

#932 

## Description

swap pin-project for pin-project-lite

## Testing

ran existing tests

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -45,7 +45,7 @@ single value. Each have been converted to struct variants and the inner value is
 instead of the `0` field.
 """
 references = ["smithy-rs#932"]
-meta = { "breaking" = true, "tada" = false, "bug" = true }
+meta = { "breaking" = true, "tada" = false, "bug" = false }
 author = "Velfi"
 
 [[aws-sdk-rust]]
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -21,3 +21,29 @@ message = "Add a trailing slash to the URI `/latest/meta-data/iam/security-crede
 references = ["aws-sdk-rust#560", "smithy-rs#1487"]
 meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "rcoh"
+
+[[smithy-rs]]
+message = """
+Replaced use of `pin-project` with equivalent `pin-project-lite`. For pinned enum tuple variants and tuple structs, this
+change requires that we switch to using enum struct variants and regular structs. Most of the structs and enums that
+were updated had only private fields/variants and so have the same public API. However, this change does affect the
+public API of `aws_smithy_http_tower::map_request::MapRequestFuture<F, E>`. The `Inner` and `Ready` variants contained a
+single value. Each have been converted to struct variants and the inner value is now accessible by the `inner` field
+instead of the `0` field.
+"""
+references = ["smithy-rs#932"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
+author = "Velfi"
+
+[[aws-sdk-rust]]
+message = """
+Replaced use of `pin-project` with equivalent `pin-project-lite`. For pinned enum tuple variants and tuple structs, this
+change requires that we switch to using enum struct variants and regular structs. Most of the structs and enums that
+were updated had only private fields/variants and so have the same public API. However, this change does affect the
+public API of `aws_smithy_http_tower::map_request::MapRequestFuture<F, E>`. The `Inner` and `Ready` variants contained a
+single value. Each have been converted to struct variants and the inner value is now accessible by the `inner` field
+instead of the `0` field.
+"""
+references = ["smithy-rs#932"]
+meta = { "breaking" = true, "tada" = false, "bug" = true }
+author = "Velfi"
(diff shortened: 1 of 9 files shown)


CHANGELOG.next.toml
```diff
@@ -22,6 +22,32 @@ references = ["aws-sdk-rust#560", "smithy-rs#1487"]
...
+message = """
+Replaced use of `pin-project` with equivalent `pin-project-lite`. For pinned enum tuple variants and tuple structs, this
+change requires that we switch to using enum struct variants and regular structs. Most of the structs and enums that
+were updated had only private fields/variants and so have the same public API. However, this change does affect the
+public API of `aws_smithy_http_tower::map_request::MapRequestFuture<F, E>`. The `Inner` and `Ready` variants contained a
+single value. Each have been converted to struct variants and the inner value is now accessible by the `inner` field
+instead of the `0` field.
+"""
+references = ["smithy-rs#932"]
+meta = { "breaking" = true, "tada" = false, "bug" = true }
```
jdisanti, line 48: This wasn't matching the other entry. Not sure which you intended.
```suggestion
meta = { "breaking" = true, "tada" = false, "bug" = false }
```

rust-runtime/aws-smithy-http/src/event_stream/input.rs
```diff
@@ -48,19 +48,20 @@ where
...
-    stream: Pin<Box<dyn Stream<Item = Result<T, BoxError>> + Send>>,
-    end_signal_sent: bool,
-    _phantom: PhantomData<E>,
+pin_project! {
+    /// Adapts a `Stream<SmithyMessageType>` to a signed `Stream<Bytes>` by using the provided
+    /// message marshaller and signer implementations.
+    ///
+    /// This will yield an `Err(SdkError::ConstructionFailure)` if a message can't be
+    /// marshalled into an Event Stream frame, (e.g., if the message payload was too large).
+    pub struct MessageStreamAdapter {
```
jdisanti, line 57: The `<T, E>` generic args disappeared? Make sure to test this crate with `--all-features` since the event stream parts aren't on by default.

rust-runtime/aws-smithy-http/src/event_stream/input.rs
```diff
@@ -48,19 +48,20 @@ where
...
+        marshaller: Box<dyn MarshallMessage<Input = T> + Send + Sync>,
+        signer: Box<dyn SignMessage + Send + Sync>,
+        #[pin]
+        stream: Pin<Box<dyn Stream<Item = Result<T, BoxError>> + Send>>,
```
jdisanti, line 61: Is pin projection actually needed here if the member is `Pin`? It seems like it shouldn't be, but I'm not 100% sure.

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -49,27 +50,40 @@ impl Debug for SdkBody {
...
-    /// `Taken`. This will return an Error when polled. Attempting to read data out of a `Taken`
-    /// Body is a bug.
-    Taken,
+pub type BoxBody = http_body::combinators::BoxBody<Bytes, Error>;
+
+pin_project! {
+    #[project = InnerProj]
+    enum Inner {
+        Once {
+            #[pin]
```
jdisanti, line 59: Do `Once` and `Dyn` actually need `#[pin]`? I understand why `Streaming` needs it, but I'm fuzzy on why these would need it.

rust-runtime/aws-smithy-http-tower/src/map_request.rs
```diff
@@ -97,10 +97,15 @@ where
...
 }
 
-#[pin_project(project = EnumProj)]
-pub enum MapRequestFuture<F, E> {
-    Inner(#[pin] F),
-    Ready(Option<E>),
+pin_project! {
+    #[project = EnumProj]
+    pub enum MapRequestFuture<F, E> {
+        Inner {
```
jdisanti, line 103: Not really related to your changes, but the name `Inner` is kind of confusing to me here. Maybe we should have called it `Pending` or something? Is there a broader idiomatic pattern for this with futures?

rust-runtime/aws-smithy-http/src/event_stream/input.rs
Velfi: I don't understand why it needs to be `Pin` but if I remove it then the compiler starts complaining about it not implementing some obtuse generic traits.

rust-runtime/aws-smithy-http/src/body.rs
Velfi: `Dyn` complains without it but `Once` seems to be ok when I remove it

rust-runtime/aws-smithy-http/src/event_stream/input.rs
rcoh: @jdisanti is correct—the key here is using `.as_mut()` on the stream instead of using pin projection. 

Interestingly you actually need to keep `pin_project!` on the root (or you need to manually implement `Unpin` (which is safe (!!))

rust-runtime/aws-smithy-http-tower/src/map_request.rs
Velfi: Looking at it, I would name the variants `Pending` and `Complete`. @rcoh what do you think?

PR #1513, opened 2022-06-30
Server SDK CHANGELOG.md file is created separately by the changelogger tool

drganjoo:
## Motivation and Context
. Server side SDK changes are to be written in a separate file codegen-server/CHANGELOG.md
. CHANGELOG.next.toml file will contain changes for both server side and client side SDK.
. Revision manifest will contain changes for both Server and Client side combined.

## Description
Currently, CHANGELOG.next.md contains only client side changes. A new attribute has been added to the meta tags to specify whether the change affects server side, client side or both.

```
# [[smithy-rs]]
message = "Fix typos in module documentation for generated crates"
references = ["smithy-rs#920"]
meta = { "breaking" = false, "tada" = false, "bug" = false, "sdk" = "client | server | both" }
author = "developer"
```

Manifest file for the revision will contain combined information:

**Client SDK Changes**
[release_notes for client]

**Server SDK Changes**
[release_notes for server sdk]

## Testing
. Using cargo test
. **Adding a few more tests** to ensure no existing functionality is broken and that the server side SDK changes are written properly

## Checklist


tools/changelogger/src/render.rs
```diff
@@ -29,7 +29,7 @@ pub const EXAMPLE_ENTRY: &str = r#"
 # [[smithy-rs]]
 # message = "Fix typos in module documentation for generated crates"
 # references = ["smithy-rs#920"]
-# meta = { "breaking" = false, "tada" = false, "bug" = false }
+# meta = { "breaking" = false, "tada" = false, "bug" = false, "sdk" = "client | server | both" }
```
jdisanti, line 32: Should we call it a `target` rather than `sdk` so that it's consistent with `CodegenTarget` in the code generator?

tools/changelogger/src/render.rs
```diff
@@ -58,6 +58,9 @@ pub struct RenderArgs {
     pub source_to_truncate: PathBuf,
     #[clap(long, action)]
     pub changelog_output: PathBuf,
+    /// where should smity_rs entries marked as server / both be written to
+    #[clap(long, action)]
+    pub server_changelog_output: Option<PathBuf>,
```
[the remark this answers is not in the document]
drganjoo: The other option would be to separate the two under different headings. But I think this would be better and maybe we can also include an emoji icon to indicate if it is a server change or a client.

tools/changelogger/src/render.rs
drganjoo: Oh yes it makes more sense to use target instead of sdk.

tools/changelogger/src/entry.rs
```diff
@@ -25,12 +25,16 @@ impl ChangelogEntries {
...
+        smithy_rs_sdk : SdkAffected,
         change_set: ChangeSet,
         previous_release_versions_manifest: Option<&Path>,
     ) -> Result<Vec<ChangelogEntry>> {
         match change_set {
             ChangeSet::AwsSdk => {
-                if let Some(manifest_path) = previous_release_versions_manifest {
+                if smithy_rs_sdk == SdkAffected::Server {
+                    Ok(vec![])
+                }
```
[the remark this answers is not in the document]
drganjoo: Changing them into their own structures would require a bit more time as right now we have common code in render_handauthored that is used for both AwsSdkRust hand authored and SmithyRs hand authored. I have refactored the code to keep entry.rs the same as before and has no change in it. render.rs checks the change_set and based on that calls two different functions. I will test it a bit more to confirm that AwsRustSdk entires cannot have a target attribute in their meta.

tools/changelogger/src/render.rs
```diff
@@ -232,9 +235,19 @@ fn render_entry(entry: &HandAuthoredEntry, mut out: &mut String) {
...
+        ", "
+    }
+    else {
+        ""
+    };
     if !references.is_empty() {
-        write!(meta, "({}) ", references.join(", ")).unwrap();
+        write!(meta, "{sep}{}", references.join(", ")).unwrap();
     }
+    meta.push_str(") ");
```
jdisanti, line 249: I think this can be simplified by just doing `references.insert(0, entry.meta.target)` so that you don't have to do manual comma delimitation.

tools/changelogger/src/render.rs
```diff
@@ -275,56 +288,125 @@ fn update_changelogs(
...
-            output_path.join(&release_metadata.manifest_name),
-            serde_json::to_string_pretty(&release_manifest)?,
+            &args.changelog_output,
+            old_client_contents
         )
-        .context("failed to write release manifest")?;
+        .context(format!("server changelog output could not be written (error: {e}). Failed to restore client changelog!"))?;
+
+        Err(anyhow::Error::msg(format!("Could not write server release notes. Error: {e}")))
+    }
```
jdisanti, line 340: This tool is basically always run in automation, so you can just `?` operator the error on the second `write_changelog_md` since that will fail the whole release GitHub Actions workflow, which means it won't matter that the client file was overwritten since that whole workspace gets thrown away.

If someone did run it manually, they can use git to restore files.

tools/smithy-rs-tool-common/src/changelog.rs
```diff
@@ -7,15 +7,64 @@
...
-use std::fmt;
+use std::fmt::{self, Display};
 use std::path::Path;
 use std::str::FromStr;
 
+#[derive(Copy, Clone, Debug, Serialize, PartialEq, Eq)]
+pub enum SdkAffected {
+    Client,
+    Server,
+    Both
```
jdisanti, line 18: Thinking about this more, we should probably call this `All` in case there's more than client and server in the future.

tools/smithy-rs-tool-common/src/changelog.rs
```diff
@@ -7,15 +7,64 @@
...
+}
+
+impl FromStr for SdkAffected {
+    type Err = anyhow::Error;
+
+    fn from_str(sdk: &str) -> std::result::Result<Self, Self::Err> {
+        match sdk.to_lowercase().as_str() {
+            "client" => Ok(SdkAffected::Client),
+            "server" => Ok(SdkAffected::Server),
+            "both" => Ok(SdkAffected::Both),
```
jdisanti, line 43:
```suggestion
            "all" => Ok(SdkAffected::All),
```

tools/changelogger/src/render.rs
```diff
@@ -58,6 +58,9 @@ pub struct RenderArgs {
     pub source_to_truncate: PathBuf,
     #[clap(long, action)]
     pub changelog_output: PathBuf,
+    /// where should smity_rs entries for server sdk 
+    #[clap(long, action)]
+    pub server_changelog_output: Option<PathBuf>,
```
jdisanti, line 63: I'm not sure I'm aligned yet on having multiple changelog files. What is your reasoning for this, and if we have more than just client and server in the future, how does this look?

drganjoo: At the moment, the server side SDK is missing some features that the clients are specifically waiting for.  Also, at this stage, we would be carrying out more frequent breaking changes than the client SDK is. Having a separate server changelog allows customers to find these changes / features more easily by just looking at the server side document. 

Moving forward, once the server SDK is more stable,  we can deprecate this changelog and start writing to the single changelog file.

tools/changelogger/tests/e2e_test.rs
```diff
@@ -205,6 +206,14 @@ fn render_smithy_rs_test() {
...
     )
     .unwrap();
+    fs::write(
+        &server_dest_path,
+        format!(
+            "{}\nv0.41.0 (Some date in the past)\n=========\n\nOld entry contents\n",
+            USE_UPDATE_CHANGELOGS
+        ),
+    )
+    .unwrap();
```
jdisanti, line 216: Not needed anymore?

jdisanti (approved):
Looks good! Just one comment on what appears to be dead code.

It looks like you'll need to run `cargo fmt` to get CI passing.

tools/changelogger/tests/e2e_test.rs
```diff
@@ -464,3 +468,93 @@ Old entry contents
...
+# [[aws-sdk-rust]]
+# message = "Fix typos in module documentation for generated crates"
+# references = ["smithy-rs#920"]
+# meta = { "breaking" = false, "tada" = false, "bug" = false }
+# author = "rcoh"
+#
+# [[smithy-rs]]
+# message = "Fix typos in module documentation for generated crates"
+# references = ["smithy-rs#920"]
+# meta = { "breaking" = false, "tada" = false, "bug" = false }
```
jdisanti, line 486: Be sure to update `EXAMPLE_ENTRY` so that `target` is included in its example.

tools/smithy-rs-tool-common/src/changelog.rs
```diff
@@ -173,16 +176,91 @@ pub struct SdkModelEntry {
...
+        |entry: &mut HandAuthoredEntry| -> Result<(), anyhow::Error> {
+            if entry.meta.target.is_some() {
+                bail!("cannot have target SDK affected as metadata")
+            }
+            Ok(())
+        },
+    )
+}
+
+fn deserialize_handauthored<'de, D, F>(
```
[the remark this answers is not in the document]
drganjoo: Deserialisation code has been removed and moved into parse_str now. The premise of implementing at the deserialisation stage was 1) to generate errors early in the pipeline 2) an incorrect value in a toml filed should result in an error at the same stage as the error for missing field does. e.g. a hand authored entry with a missing author field results in errors when `toml::from_str` is called. 3) the code should not rely on the caller to call a separate validation function, because if they forget to call it they would be working with incorrect data 4) calling a validation function should be an immutable and repeatable operation where as we need to overwrite missing meta.target values with SdkAffected::All.

The new code calls validation as part of `Changelog::parse_str`, so that point number 2,3 and 4 in the above paragraph can still be guaranteed. However, if someone calls `toml::from_str` to load the Changelog then `parse_str` is not called  and they will have to call validation separately.

tools/changelogger/src/render.rs
```diff
@@ -489,20 +503,23 @@ module = "aws-sdk-s3"
 version = "0.14.0"
 kind = "Feature"
 message = "Some new API to do X"
+references = ["smithy-rs#199"]
```
jdisanti, line 506: `[[aws-sdk-model]]` entries shouldn't have references. Looks like there are about four of these that should be removed.

tools/smithy-rs-tool-common/src/changelog.rs
```diff
@@ -177,12 +238,30 @@ impl Changelog {
...
 
     pub fn validate(&self) -> Result<(), Vec<String>> {
-        let mut errors = vec![];
-        for entry in self.aws_sdk_rust.iter().chain(self.smithy_rs.iter()) {
-            if let Err(e) = entry.validate() {
-                errors.push(format!("{}", e));
+        let validate_aws_handauthored = |entry: &HandAuthoredEntry| -> Result<()> {
+            entry.validate()?;
+            if entry.meta.target.is_some() {
+                bail!("aws-sdk-rust changelog entry cannot have an affected sdk");
```
jdisanti, line 244:
```suggestion
                bail!("aws-sdk-rust changelog entry cannot have an affected target");
```

tools/smithy-rs-tool-common/src/changelog.rs
```diff
@@ -177,12 +238,30 @@ impl Changelog {
...
             }
-        }
+            Ok(())
+        };
+
+        let validate_smithyrs_handauthored = |entry: &HandAuthoredEntry| -> Result<()> {
+            entry.validate()?;
+            if entry.meta.target.is_none() {
+                bail!("smithy-rs entry must have an affected sdk");
```
jdisanti, line 252:
```suggestion
                bail!("smithy-rs entry must have an affected target");
```

tools/changelogger/tests/e2e_test.rs
jdisanti: What is this for?

jdisanti (approved):
Just some minor things left. Otherwise, looks good.

jdisanti (approved):

PR #1520, opened 2022-06-30
Codegenerate Python application example and add explicit cast during JSON deserialization

crisidev:
## Motivation
See #1367.

## Description
* Explicitly call `from` complex types in JSON deserialization if the shape has been overriden by the symbol provider.
* Codegenerate Python application example.

## Checklist


crisidev:
Example: 

image

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
```diff
@@ -179,4 +188,69 @@ class PythonApplicationGenerator(
...
+/// app = App()
+/// app.context(Context())
+///
+${operationImplementationStubs(operations)}
+///
+/// app.run()
+/// ```
+///
+/// Any of operations above can be written as well prepending the `async` keyword and
+/// the Python application will automatically handle it and schedule it on the even loop for you.
```
hlbarber, line 227: event* loop

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
```diff
@@ -179,4 +188,69 @@ class PythonApplicationGenerator(
...
+    private fun renderPyApplicationRustDocs(writer: RustWriter) {
+        writer.rust(
+"""
+##[allow(clippy::tabs_in_doc_comments)]
+/// Main Python application, used to register operations and context and start multiple
+/// workers on the same shared socket.
+///
+/// Operations can be registrered using the application object as a decorator (`@app.operation_name`).
+///
+/// Here's a full example to get you started using coroutines:
```
hlbarber, line 201: Seems like `operationImplementationStubs` will print non-async signatures? So it might confuse people with "started using coroutines" here?

crisidev: Very good point!

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
crisidev: Will fix.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/ParserUtil.kt
```diff
@@ -0,0 +1,40 @@
...
+import software.amazon.smithy.rust.codegen.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.smithy.RustSymbolProvider
+import software.amazon.smithy.rust.codegen.smithy.rustType
+
+/*
+ * Utility class used to force casting a non primitive type into one overriden by a new symbol provider,
+ * by explicitly calling `from()`.
+ *
+ * For example we use this in the server Python implementation, where we override types like [Blob] and [DateTime]
+ * with wrappers compatibile with Python, without touching the original implementation coming from `aws-smithy-types`.
```
jdisanti, line 24:
```suggestion
 * with wrappers compatible with Python, without touching the original implementation coming from `aws-smithy-types`.
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/ParserUtil.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+class ParserUtil(private val symbolProvider: RustSymbolProvider, private val runtimeConfig: RuntimeConfig) {
```
jdisanti, line 26: I don't think the name `ParserUtil` is really conveying the intent of this class. Maybe call it something like, `TypeConversionGenerator`?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/ParserUtil.kt
```diff
@@ -0,0 +1,40 @@
...
+    fun convertViaFrom(shape: Shape): Writable =
```
jdisanti, line 27: How this is written now assumes the conversion will always be a `.something()` suffix, which may not always be the case. I think this would be a lot more flexible if it took a String or Writable as an argument that represents the value/expression being converted, and then places that in the correct place itself.

jdisanti (approved):
Overall, looks good to me. Just a few suggestions on `ParserUtil`.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/ParserUtil.kt
crisidev: Yeah, it better conveys the idea.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/ParserUtil.kt
crisidev: I am not sure I understand properly what you are suggesting here...

What do you mean with expect `.something()` to always exists?

I thought that passing a shape and adding the suffix to it in case it differs from the original shape was the most general way of achieving this.

Can you expand a little on your proposal?

hlbarber (approved):

PR #1521, opened 2022-06-30
forbid timezone in rfc3339 

82marbag:
According to [0], "Date time as defined by the date-time production
in RFC3339 section 5.6 with no UTC offset..."

This commit makes the parsing compliant.

[0] https://awslabs.github.io/smithy/1.0/spec/core/protocol-traits.html#timestampformat-trait

## Description

Comply with RFC3339 with no timezone offset. Format: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6

## Checklist


rust-runtime/aws-smithy-types/src/date_time/format.rs
```diff
@@ -377,6 +377,12 @@ pub(crate) mod rfc3339 {
     // Timezones not supported:
     // Not OK: 1985-04-12T23:20:50-02:00
     pub(crate) fn parse(s: &str) -> Result<DateTime, DateTimeParseError> {
+        let timezone_offset = "-02:00".len() - 1;
+        if matches!(s.chars().rev().nth(timezone_offset), Some('-') | Some('+')) {
+            return Err(DateTimeParseError::Invalid(
+                "invalid RFC-3339 date-time: timezone offset not supported".into(),
```
jdisanti, line 383: This will break the SDK since the `HttpCredentialProvider` uses this module to parse the credential expiration time, and the SDK JSON credentials format allows for timezone offsets. You can see the call here: https://github.com/awslabs/smithy-rs/blob/ea2ae7bfc6d8ba1daf7a4648a4027b7268f56283/aws/rust-runtime/aws-config/src/json_credentials.rs#L181

I'm happy our unit tests caught this 😄 

I don't have any good suggestions at the moment for working around this. @rcoh or @Velfi might have some ideas. It's probably a tradeoff between modifying the `aws-smithy-types` public API or using the `time` crate directly for parsing in `aws-config`.

david-perez: I'd expect things in this crate to follow the idiosyncrasies of the Smithy spec. It appears like the Rust docs already mention that this wasn't intended to be fully compliant for RFC3339 by calling out that timezones are not supported.

rust-runtime/aws-smithy-types/src/date_time/format.rs
```diff
@@ -377,6 +377,12 @@ pub(crate) mod rfc3339 {
     // Timezones not supported:
     // Not OK: 1985-04-12T23:20:50-02:00
     pub(crate) fn parse(s: &str) -> Result<DateTime, DateTimeParseError> {
+        let timezone_offset = "-02:00".len() - 1;
+        if matches!(s.chars().rev().nth(timezone_offset), Some('-') | Some('+')) {
```
david-perez, line 381: `s` is all ASCII right? Can't we index directly instead of consuming the iterator in linear time?

rust-runtime/aws-smithy-types/src/date_time/format.rs
```diff
@@ -626,6 +632,15 @@ mod tests {
         assert_eq!(e2, expected);
     }
 
+    #[test]
+    fn parse_rfc3339_timezone_forbidden() {
+        let dt = rfc3339::parse("1985-04-12T23:20:50-02:00");
+        assert_eq!(
+            true,
+            matches!(dt.unwrap_err(), DateTimeParseError::Invalid(_))
```
david-perez, line 640:
```suggestion
        assert!(
            matches!(dt.unwrap_err(), DateTimeParseError::Invalid(_))
```

rust-runtime/aws-smithy-types/src/date_time/format.rs
rcoh: let's use `time` directly in aws-config and keep smithy-types smithy compliant. It was just a convenience that I used Smithy to do it—I didn't realize that was slightly wrong

rcoh (commented):
This looks good—I support fixing aws-smithy-types. However we will need to go fix the `aws-config` crate.

82marbag:
@rcoh 

> However, is there another assertion that would work? eg. can we assert that it ends in Z?

Yes, that works as well. I've changed it to match on the ending char instead, thanks

rust-runtime/aws-smithy-types/src/date_time/format.rs
82marbag: Yes, done. Thanks

jdisanti:
Waiting for the `aws-config` changes to review this.

82marbag:
> Waiting for the `aws-config` changes to review this.

@jdisanti updated

rust-runtime/aws-smithy-types/src/date_time/format.rs
```diff
@@ -377,6 +377,11 @@ pub(crate) mod rfc3339 {
     // Timezones not supported:
     // Not OK: 1985-04-12T23:20:50-02:00
     pub(crate) fn parse(s: &str) -> Result<DateTime, DateTimeParseError> {
+        if !matches!(s.chars().last(), Some('Z')) {
+            return Err(DateTimeParseError::Invalid(
+                "invalid RFC-3339 date-time: timezone offset not supported".into(),
```
jdisanti, line 382: This error message should probably be rephrased since date times with a timezone offset actually are valid RFC-3339 timestamps:
```suggestion
                "Smithy does not support timezone offsets in RFC-3339 date times".into(),
```

jdisanti (approved):
LGTM!

rust-runtime/aws-smithy-types/src/date_time/format.rs
82marbag: You're right, thanks!

hlbarber (approved):

PR #1523, opened 2022-07-01
Fix RustWriter debug mode

hlbarber:
## Motivation and Context

Debug mode is useful for tracing the origin of generated code. Currently `./gradlew codegen-server-test:check` will fail with it enabled.

This is caused by the following `rustfmt` bug https://github.com/rust-lang/rustfmt/issues/5425 which causes valid rust code to break after `cargo fmt`. One instance of this pattern is generated by `withBlock("$memberName: self.$memberName", ",")` found [here](https://github.com/awslabs/smithy-rs/blob/46984ab91d027e6a4d1e312c125a850f5cb23cfd/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/BuilderGenerator.kt#L252).

## Description

To prevent such code being generated we disallow debug comment being introduced when the line is just a single `,`.

Note that this fix is not "tight" in the following sense:

1. This probably does not completely cover all cases where we can run into this `rustfmt` bug.
2. This prevents debug comments being attached to commas which wouldn't result in breakage.

Without a deeper inspection of the AST it's unlikely we'll be able to achieve 1 or 2. This PR is presented as a reasonable compromise until the `rustfmt` bug is fixed.

## Testing

1. Set [debugMode](https://github.com/awslabs/smithy-rs/blob/main/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt#L331) to `true`.
2. Rerun all tests.
3. Observe that all tests pass, except for the handful codegen tests which assert equality on string outputs.

This [commit](https://github.com/awslabs/smithy-rs/pull/1523/commits/caee9c29281b1831cbe297c4ca48c144c1c78b34) has been made and reverted to demonstrate this in CI.


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
```diff
@@ -359,7 +359,9 @@ class RustWriter private constructor(
     }
 
     override fun write(content: Any?, vararg args: Any?): RustWriter {
-        if (debugMode) {
+        // The second condition introduced to prevent the following bug
+        // https://github.com/rust-lang/rustfmt/issues/5425
+        if (debugMode && (content as? String?)?.let { it.trim() != "," } ?: false) {
```
hlbarber, line 364: Will this cast to `String` ever fail? Should we default to `true`?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
```diff
@@ -328,7 +328,7 @@ class RustWriter private constructor(
     private val commentCharacter: String = "//",
     private val printWarning: Boolean = true,
     /** Insert comments indicating where code was generated */
-    private val debugMode: Boolean = false,
+    private val debugMode: Boolean = true,
```
david-perez, line 331: This should be reverted, it's causing the unit tests to fail.

hlbarber: Yep, pushed to get a link to CI in the case where it's `true`. Reverting it now.

hlbarber:
As explained in https://github.com/rust-lang/rustfmt/issues/5425#issuecomment-1172649588, we're able to avoid the `rustfmt` behavior by running it with `trailing_comma=Never`. This might be a better temporarily fix.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
rcoh: the first argument to write always needs to be a string, I think. (in our usage at least)

rcoh (approved):
this is OK for now, please add a `// TODO(<issue>)` with a link to the issue

PR #1524, opened 2022-07-01
Make `ServerProtocolLoader` delegate to `ProtocolLoader`

david-perez:
While the set of supported protocols and their implementation is
different among servers and clients, the logic to load a specific
protocol given a Smithy model is identical. This commit makes
`ServerProtocolLoader` inherit from `ProtocolLoader` so that said logic
can be reused among clients and servers.


rcoh (approved):

rcoh:
could parameterize the `ProtocolLoader` to return `T: Protocol` then have `ServerProtocol` as a subclass and cleanup the other protocol-related PR

PR #1525, opened 2022-07-01
Make `ServerOperationRegistryGenerator` protocol-agnostic

david-perez:
`ServerOperationRegistryGenerator` is the only server generator that is
currently not protocol-agnostic, in the sense that it is the only
generator that contains protocol-specific logic, as opposed to
delegating to the `Protocol` interface like other generators do.

With this change, we should be good to implement a
`RustCodegenDecorator` loaded from the classpath that implements support
for any protocol, provided the decorator provides a class implementing
the `Protocol` interface.

This commit also contains some style changes that are making `ktlint`
fail. It seems like our `ktlint` config was recently broken and some
style violations slipped through the cracks in previous commits that
touched these files.

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/RestRequestSpecGenerator.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/RestRequestSpecGenerator.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.rust.codegen.smithy.generators.http
+
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.rust.codegen.rustlang.CargoDependency
+import software.amazon.smithy.rust.codegen.rustlang.Writable
+import software.amazon.smithy.rust.codegen.rustlang.asType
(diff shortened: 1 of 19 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/http/RestRequestSpecGenerator.kt
```diff
@@ -0,0 +1,94 @@
...
+import software.amazon.smithy.rust.codegen.rustlang.writable
+import software.amazon.smithy.rust.codegen.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.smithy.protocols.HttpBindingResolver
+
+/**
+ * [RestRequestSpecGenerator] generates a restJson1 or restXml specific `RequestSpec`. Both protocols are routed the same.
+ *
+ * This class has to live in the `codegen` subproject instead of in the `codegen-server` subproject because it is used
+ * by the implementations of the `serverRouterRequestSpec` of the [Protocol] interface, which is used by both subprojects
+ * (even though only the `codegen-server` subproject calls `serverRouterRequestSpec`).
```
rcoh, line 23: could potentially have `ServerProtocol` and try to downcast? that might have other issues I guess. This is fine to merge, but we may be want to think about other ways to handle this. (Eg. maybe refactoring out `CoreProtocol`, `ClientProtocol` and `ServerProtocol` or something)

rcoh (approved):

82marbag (approved):

PR #1527, opened 2022-07-04
Remove FromRequest and IntoResponse trait

hlbarber:
## Motivation and Context

`async fn` are currently not supported by Rust. To allow for such syntax today, the [async_trait](https://docs.rs/async-trait/latest/async_trait/) macro eagerly `Box::pin`s the `async` block. This incurs an unnecessary allocation.

One instance where `async_trait` is used is on the [FromRequest](https://docs.rs/axum/latest/axum/extract/trait.FromRequest.html) method (inherited from [axum](https://docs.rs/axum/latest/axum/)). The use case for `FromRequest` in `axum` does not exist in our case, it is therefore a candidate for removal.

While `IntoResponse` does not involve the same boxing as `FromRequest`, it also does not need to be a trait - and so we remove it for symmetry. 

## Description

* Remove `FromRequest` and replace it with a static method
* Remove `IntoResponse` and replace it with a static method


david-perez (approved):
Nice!

david-perez (approved):

Issue #1529, opened 2022-07-05
Use OpenSSL as an optional crypto provider


syedriko:
In a FIPS environment, there is a requirement that all crypto operations are delegated to a FIPS-certified crypto module. Some Linux distributions, such as RHEL, ship FIPS-certified OpenSSL. In order for the SDK to be usable in such environment, it should be possible to rely on OpenSSL for all crypto needs.
https://github.com/awslabs/smithy-rs/issues/261 seems to be related.

jdisanti:
Closing in favor of tracking issue #3563.

PR #1530, opened 2022-07-05
Use OpenSSL as an optional crypto provider (#1529)

syedriko:
A first stab at making OpenSSL an alternative crypto provider in aws-sigv4 and
aws-config, so the SDK can be consumed in a FIPS environment.

## Motivation and Context
In a FIPS environment, there is a requirement that all crypto operations are delegated to a FIPS-certified crypto module. Some Linux distributions, such as RHEL, ship FIPS-certified OpenSSL. In order for the SDK to be usable in such environment, it should be possible to rely on OpenSSL exclusively for all crypto needs.
https://github.com/awslabs/smithy-rs/issues/1529

## Description
I would like to start a conversation on how the SDK can be used in a FIPS environment. This PR makes it possible to replace Ring with OpenSSL as a crypto provider in aws-sigv4 and aws-config, but there are still rough edges as Ring types are exposed in the public API.

## Testing
- Tested with tools/ci-build/scripts/check-rust-runtimes-and-tools 
- Tested as part of https://github.com/vectordotdev/ against live AWS

## Checklist
- [x ] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates


Velfi:
Hey @syedriko, thanks for looking into this and submitting a PR. I have a philosophical issue with it.

I wrote an [RFC a little while ago](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0015_using_features_responsibly.md) _(that the team hasn't yet accepted)_ detailing how I think we should handle situations like this. You have code similar to this in a few places:

```rust
cfg_if::cfg_if! {
    if #[cfg(feature = "ring")] {
        use ring::digest::{self, digest};
        let mut ret: [u8; 20] = [0; 20];
        ret.clone_from_slice(digest(&digest::SHA1_FOR_LEGACY_USE_ONLY, buf).as_ref());
        ret
    } else if #[cfg(feature = "openssl")] {
        openssl::sha::sha1(buf)
    } else {
        compile_error!("Either feature \"openssl\" or \"ring\" must be enabled for this crate.");
    }
}
```

I'd prefer that we move away from writing the SDK in such a way that customers configure how it works by setting features. Writing things like this puts a burden on users to understand how configuration is set depending on what matrix of features they have enabled. This is further complicated by how features in Rust are additive. It's easy to unintentionally  activate features when configuring dependencies because of [feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification).

I think that a better alternative would be to make the signer/hash generator something that could be replaced through configuration, in the same way that we allow users to
- [replace the default credentials provider](https://docs.rs/aws-config/latest/aws_config/struct.ConfigLoader.html#method.credentials_provider)
- [set their own async sleep impl](https://docs.rs/aws-config/latest/aws_config/struct.ConfigLoader.html#method.sleep_impl)
- [set an http connector to use over the default hyper connector](https://docs.rs/aws-config/latest/aws_config/struct.ConfigLoader.html#method.http_connector) _(currently, this feature is only partially implemented)_

It's not my intention to shoot your contribution down; I do think there is a place for this code in the SDKs or a related crate. In my opinion, an RFC is the next step. Let me know what you think.

jdisanti:
Hey @syedriko, we appreciate this contribution, but it's been a year and there hasn't been any traction on it. I'm going to close it for now. If you decide you want to work on it again, feel free to open a new PR.

PR #1533, opened 2022-07-06
`aws-config`: Fix compilation error with `rustls` and `native-tls` disabled

joshtriplett:
The `ProviderConfig::with_tcp_connector` method uses
`aws_smithy_client::hyper_ext`, which only exists with the `client-hyper`
feature enabled. Add a feature enabling that, and enable it by default.

Introducing this feature does not cause breakage, because aws-config did not
previously compile with `default-features = false` and neither `rustls` nor
`native-tls` enabled.

## Testing

CI to make sure existing combinations of features don't fail; careful
inspection for the new feature. (Compiling the SDK overall without either
rustls or native-tls enabled is still a work in progress.)

## Checklist


Velfi (approved):
Thanks for fixing this!

Velfi:
@joshtriplett  Looking at the CI failures, you'll need to fix a few minor issues before we can merge this:
- Your changelog entry doesn't reference an issue so I [created one for you](https://github.com/awslabs/smithy-rs/issues/1541). You can reference it like so: `smithy-rs#1541`
- There are some unused imports when the `client-hyper` feature is inactive, these will have to be moved around or usages can be changed to be fully qualified. Running `cargo test --no-default-features` for `aws-config` should expose these
- We have a test in `aws-config/src/default_provider/credentials.rs` using `with_tcp_connector` that will need to be feature-gated

Let me know if you have trouble with any of that.

_P.S. I want to personally thank you for everything you've done for the Rust community_

joshtriplett:
@Velfi Done, I think. And thank you. :)

rcoh:
good catch! seems like this slipped through the cracks because the `test-util` feature (enabled in dev-dependencies) brings in Hyper from aws-smithy-client.

rcoh (approved):

PR #1535, opened 2022-07-06
aws-smithy-client: Support using rustls with either native roots or webpki roots

joshtriplett:
Add separate cargo features for rustls-native-roots and
rustls-webpki-roots. Make the existing `rustls` feature a compatibility
alias for `rustls-native-roots`, to make this a non-breaking change.

Upgrade hyper-rustls to 0.23.0 to improve support for this. (This also
avoids having two versions of hyper-rustls (and its dependencies) in
code that uses a custom hyper-rustls connector.)

This change makes it much easier to work with webpki-roots, without having to
fight the SDK or override connectors in multiple different places to avoid
depending on native roots.

## Testing

Confirmed that every combination of features builds and passes tests.

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -25,3 +25,13 @@ message = "Re-export aws_types::SdkConfig in aws_config"
 references = ["smithy-rs#1457"]
 meta = { "breaking" = false, "tada" = true, "bug" = false }
 author = "calavera"
+
+[[aws-sdk-rust]]
+message = "aws-smithy-client: Support using rustls with either native roots or webpki-roots"
+meta = { "breaking" = false, "tada" = true, "bug" = false }
+author = "joshtriplett"
+
+[[aws-sdk-rust]]
+message = "aws-smithy-client: Upgrade hyper-rustls to 0.23.0"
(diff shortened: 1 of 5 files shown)


Velfi:
Per [the comment](https://github.com/awslabs/aws-sdk-rust/issues/575#issuecomment-1176408885) I left on aws-sdk-rust#575, I want to hold off on merging this. I'm going to schedule a team meeting to discuss how we want to handle this and get back to you as soon as that happens.

FSMaxB:
@Velfi did this team meeting happen?

Velfi:
@FSMaxB The current state of things is that overriding connectors is the way to accomplish this. Things have improved since this PR was submitted. Namely, it's now much easier to set the connector once and have it be used everywhere.

For guidance on creating the desired connector and setting it, [see this discussion](https://github.com/awslabs/smithy-rs/discussions/3022).

_BEWARE: this discussion only applies to the upcoming version 0.57.0 and up._

jdisanti:
This isn't an approach we want to take for this, and it's been a year since it was submitted. I'm going to close this PR for now. If you're interested in working on it again in the future, then there should be an RFC written for your proposed approach. Thank you!

PR #1536, opened 2022-07-06
RFC: Logging in the Presence of Sensitive Data

hlbarber:
[Rendered](https://github.com/awslabs/smithy-rs/blob/harryb/logging-rfc/design/src/rfcs/rfc0018_logging_sensitive.md)


design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+- **Runtime crate**: A crate existing within the `rust-runtime/` folder.
+- **Service**: A trait defined in the [`tower-service` crate][tower_service::Service]. The lowest level of abstraction we deal with when making HTTP requests. Services act directly on data to transform and modify that data. A Service is what eventually turns a request into a response.
+- **Layer**: Layers are a higher-order abstraction over services that is used to compose multiple services together, creating a new service from that combination. Nothing prevents us from manually wrapping services within services, but Layers allow us to do it in a flexible and generic manner. Layers don't directly act on data but instead can wrap an existing service with additional functionality, creating a new service. Layers can be thought of as middleware. *NOTE: The use of [Layers can produce compiler errors] that are difficult to interpret and defining a layer requires a large amount of boilerplate code.*
+- **Potentially sensitive**: Data that _could_ be bound to a sensitive field of a structure, for example [HTTP Binding Traits](#http-binding-traits).
+
+## Background
+
+### HTTP Binding Traits
```
hlbarber, line 27: Are there any other ways that `@sensitive` fields can be bound/leak?

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+
+These two cases illustrate that `smithy-rs` can only prevent violation of the specification in a restricted scope - logs emitted from generated code and the runtime crates. A `smithy-rs` specific guideline should be available to the customer which outlines how to avoid violating the specification in areas outside of our control.
+
+### Routing
+
+The sensitivity and HTTP bindings are declared within specific structures/operations. For this reason, in the general case, it's unknowable whether or not any given part of a request is sensitive until we determine which operation is tasked with handling the request and hence which fields are bound. Implementation wise, this means that any `Layer` applied _before_ routing has taken place cannot log anything sensitive without performing routing logic itself.
+
+Note that:
+- We are not required to deserialize the entire request before we can make judgments on what data is sensitive or not - only which operation it has been routed to.
+- We are permitted to emit logs prior to routing when:
```
hlbarber, line 67: I think this allows us to do important logging in the `Router` and the remaining logging can be done within the logging layer.

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+### Code Generated Logging Layer
+
+Using the smithy model, for each operation, a logging `Layer` should be generated. Through the model, the code generation knows which fields are sensitive and which HTTP bindings exist, therefore the `Logging` layer can be careful crafted to avoid leaking sensitive data.
+
+This `Layer` should be applied to the [OperationHandler](https://github.com/awslabs/smithy-rs/blob/cd0563020abcde866a741fa123e3f2e18e1be1c9/rust-runtime/inlineable/src/server_operation_handler_trait.rs#L17-L21) directly after its construction in the generated `operation_registry.rs`. The `Layer` should preserve the associated types of the `OperationHandler` (`Response = Response<BoxBody>`, `Error = Infallible`) so cause minimal breakage.
+
+This is illustrated before as the introduction of `Logging{Operation}::new`.
+
+```rust
+let empty_operation = LoggingEmptyOperation::new(operation(registry.empty_operation));
```
hlbarber, line 142: One of the problems with putting it here is that any layer applied on top of this will fail to included in the `span!`. What we really want is for the logging layer to **always** be applied just after the routing has been performed.

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+        let fut = async {
+            let response = self.inner.call(request).await;
+
+            debug!(status_code = %Sensitive(request.status()), "sending response");
+
+            response
+        };
+
+        // Instrument the future with a span
+        let span = span!(%method, %uri, headers = request.headers(), "received request");
```
hlbarber, line 221: This might be a lot of data to include in a span? We might want a way to reduce the amount placed in here and instead log an `debug!(?headers)` for example.

The benefit of having the `span` is that every log within this will also have access to this data.

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+
+        self.inner.call(request)
+    }
+}
+```
+
+A middleware layer must be code generated (much in the same way as the logging layer) which is dedicated to inserting the `Sensitivity` struct into the extensions of each incoming request.
+
+```rust
+impl<B, S> Service<Request<B>> for SensitivityInserter<S>
```
hlbarber, line 288: This also runs into the same problem as . We want `Request<B>` to get the extension as soon as routing has been done and we know which operation it's being passed to.

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+    });
+```
+
+A more subtle violation of the specification may occur when the customer enables verbose logging - a third-party dependency might simply log data marked as sensitive, for example `tokio` or `hyper`.
+
+These two cases illustrate that `smithy-rs` can only prevent violation of the specification in a restricted scope - logs emitted from generated code and the runtime crates. A `smithy-rs` specific guideline should be available to the customer which outlines how to avoid violating the specification in areas outside of our control.
+
+### Routing
+
+The sensitivity and HTTP bindings are declared within specific structures/operations. For this reason, in the general case, it's unknowable whether or not any given part of a request is sensitive until we determine which operation is tasked with handling the request and hence which fields are bound. Implementation wise, this means that any `Layer` applied _before_ routing has taken place cannot log anything sensitive without performing routing logic itself.
```
hlbarber, line 63: > in the general case

in noted because in the case where there are absolutely no `@sensitive` traits used in the model we can log freely anywhere.

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+
+> Sensitive data MUST NOT be exposed in things like exception messages or log output. Application of this trait SHOULD NOT affect wire logging (i.e., logging of all data transmitted to and from servers or clients).
+
+This RFC is concerned with solving the problem of honouring this specification.
+
+Progress has been made towards this goal in the form of the [Sensitive Trait PR](https://github.com/awslabs/smithy-rs/pull/229), which uses code generation to remove sensitive fields from `Debug` implementations.
+
+The problem remains open due to the existence of HTTP binding traits and a lack of clearly defined user guidelines which customers may follow to honour the specification.
+
+This RFC proposes a new logging `Layer` to be generated and applied to each `OperationHandler` `Service`, and internal and external developer guidelines on how to avoid violating the specification.
```
crisidev, line 15: I would give some details about the fact that our users also have no understanding of the routing decisions before they happen and this can be very confusing and hard to debug.

Remember this applies only to `rest*` protocols. AwsJson1.0 and 1.1 can be already enhanced with logging information about the URI (always /), the method (always POST) and the operation (x-amzn-target header) without caring about the sensitive trait since all the fields are part of the payload, so routing info are safe.

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+
+## Terminology
+
+- **Model**: A [Smithy Model](https://awslabs.github.io/smithy/1.0/spec/core/model.html), usually pertaining to the one in use by the customer.
+- **Runtime crate**: A crate existing within the `rust-runtime/` folder.
```
crisidev, line 20:
```suggestion
- **Runtime crate**: A crate existing within the `rust-runtime/` folder, used to implement shared functionalities that do not have to be code-generated.
```

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+- **Layer**: Layers are a higher-order abstraction over services that is used to compose multiple services together, creating a new service from that combination. Nothing prevents us from manually wrapping services within services, but Layers allow us to do it in a flexible and generic manner. Layers don't directly act on data but instead can wrap an existing service with additional functionality, creating a new service. Layers can be thought of as middleware. *NOTE: The use of [Layers can produce compiler errors] that are difficult to interpret and defining a layer requires a large amount of boilerplate code.*
+- **Potentially sensitive**: Data that _could_ be bound to a sensitive field of a structure, for example [HTTP Binding Traits](#http-binding-traits).
+
+## Background
+
+### HTTP Binding Traits
+
+Smithy provides various [HTTP binding traits](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html). These allow protocols to configure a HTTP request by way of binding fields to parts of the request. For this reason sensitive data might be unintentionally leaked through logging of a bound request.
```
crisidev, line 29: This is true only for `rest*` protocols in our current implementation, AwsJson 1.0 and 1.1 do not support HTTP binding traits.

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+| [httpPrefixHeaders](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpprefixheaders-trait) | Headers             |
+| [httpLabel](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httplabel-trait)                 | URI                 |
+| [httpQuery](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpquery-trait)                 | Query Parameters    |
+| [httpResponseCode](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpresponsecode-trait)   | Status Code
+
+Each of these configurable parts must therefore be logged cautiously.
+
+### Scope and Guidelines
+
+Making the logging of sensitive data forbidden an type theoretic invariant is unfeasible. With the current API, the customer will always have an opportunity to log a request containing sensitive data before it enters the `Service<Request<B>>` that we provide to them.
```
crisidev, line 44:
```suggestion
Making the logging of sensitive data forbidden and type theoretic invariant is unfeasible. With the current API, the customer will always have an opportunity to log a request containing sensitive data before it enters the `Service<Request<B>>` that we provide to them.
```

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+
+### Runtime Crates
+
+The crates existing in `rust-runtime` are not code generated - their source code is agnostic to the specific model in use. For this reason, if such a crate wanted to log potentially sensitive data then there must be a way to conditionally toggle that log without manipulation of the source code. Any proposed solution must acknowledge this concern.
+
+## Proposal
+
+This proposal serves to honor the sensitivity specification via code generation of a logging `Layer`s which is aware of the sensitivity, together with a developer contract disallowing logging potentially sensitive data in the runtime crates. An internal and external guideline should be provided in addition to the `Layer`s.
+
+All data known to be sensitive should be replaced with `"_redacted_"` when logged. Implementation wise this means that [tracing::Event](https://docs.rs/tracing/latest/tracing/struct.Event.html)s and [tracing::Span](https://docs.rs/tracing/latest/tracing/struct.Span.html)s of the form `debug!(field = "sensitive data")` and `span!(..., field = "sensitive data")` must become `debug!(field = "_redacted_")` and `span!(..., field = "_redacted_")`.
```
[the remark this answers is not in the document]
crisidev: `_redacted_` or `**********` work for me. I probably prefer the latter, but not a big deal :)

design/src/rfcs/rfc0017_logging_sensitive.md
crisidev: This is a fair callout. I think I am in favour of have as many infos as possible. What do you think people?

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+- Applicable to _all_ middleware.
+- Good separation of concerns:
+  - Does not pollute the API of the middleware
+  - No specific logic required within middleware.
+
+#### Disadvantages
+
+- Complex implementation.
+- Not necessarily fast.
+- `tracing::Layer`s seem to only support filtering entire `Event`s, rather than more fine grained removal of fields.
```
crisidev, line 376: Would this middleware injected by us or should the customer add it to their middleware chain?

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+
+A guideline should be made available to customers to outline the following:
+
+- The [HTTP bindings traits](#http-binding-traits) and why they are of concern.
+- Warn against the two potential leaks described in [Scope and Guidelines](#scope-and-guidelines):
+  - Sensitive data leaking from third-party dependencies.
+  - Sensitive data leaking from middleware applied to the `Router`.
+- How to use the `Sensitive` struct and the `debug-logging` feature flag described in [Debug Logging](#debug-logging).
+
+## Alternative Proposals
```
crisidev, line 247: I think I lean towards the main solution you proposed as the code-generation allows us to precisely tailor the logging on a per-operation basis, while using runtime switches feels more complicated and we could probably loose control without noticing in the future.

crisidev (commented):
Nice work Harry, I put some comments and thoughts.

design/src/rfcs/rfc0017_logging_sensitive.md
hlbarber: I think it'd have to be applied by the customer - the design of `tracing` prohibits anyone other than the customers app from constructing and initializing the tracing `Subscriber`.

hlbarber: This could act as an alternative to `debug-logging` feature flag - the customer just doesn't apply the `tracing::Layer`.

design/src/rfcs/rfc0017_logging_sensitive.md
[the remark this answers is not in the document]
hlbarber: This marker should perhaps break the URL/Header spec in order to **really** signal that this is a piece of redacted data. For example, in the current proposal a logged URI could just be a normal `http::amazon.com/invetory/_redacted_/` URI?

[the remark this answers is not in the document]
hlbarber: URI values:
https://www.rfc-editor.org/rfc/rfc3986#section-2.2

Header values:
https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6

So perhaps `{redacted}` or `<redacted>` or something involving `{}<>` would be good?

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,376 @@
...
+
+## Proposal
+
+This proposal serves to honor the sensitivity specification via code generation of a logging `Layer`s which is aware of the sensitivity, together with a developer contract disallowing logging potentially sensitive data in the runtime crates. An internal and external guideline should be provided in addition to the `Layer`s.
+
+All data known to be sensitive should be replaced with `"_redacted_"` when logged. Implementation wise this means that [tracing::Event](https://docs.rs/tracing/latest/tracing/struct.Event.html)s and [tracing::Span](https://docs.rs/tracing/latest/tracing/struct.Span.html)s of the form `debug!(field = "sensitive data")` and `span!(..., field = "sensitive data")` must become `debug!(field = "_redacted_")` and `span!(..., field = "_redacted_")`.
+
+### Debug Logging
+
+Developers might want to observe sensitive data for debugging purposes. It should be possible to opt-out of the redactions by enabling a feature flag `debug-logging` (which is disabled by default).
```
jdisanti, line 83: To make it clearer that `debug-logging` isn't something that should be enabled in production, should it be called something like `unredacted-logging`?

crisidev: I like this name John!

design/src/rfcs/rfc0017_logging_sensitive.md
hlbarber: I think this is something quite serious and should be addressed. I'm leaving a comment in the RFC and adding it to the checklist.

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,403 @@
...
+    debug!(%data, "logging here");
+} else {
+    debug!(data = "{redacted}", "logging here");
+}
+```
+
+the following wrapper should be provided
+
+```rust
+pub struct Sensitive<T>(T);
```
jdisanti, line 102: Where should this struct live?

jdisanti (commented):
Great work on this RFC! It's very well written. I have just a couple questions.

design/src/rfcs/rfc0017_logging_sensitive.md
hlbarber: Good question, there's conversation to be had here. I've raised it https://github.com/awslabs/smithy-rs/pull/1550#discussion_r923658147, but for the purpose of the RFC, perhaps it's specific enough to say that it'll live within the runtime crates.

hlbarber: https://github.com/awslabs/smithy-rs/pull/1536/commits/55490389a1f66e3efad0b4db8f6a6babbc08fde9

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,403 @@
...
+| [httpLabel](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httplabel-trait)                 | URI              |
+| [httpPayload](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httppayload-trait)             | Payload          |
+| [httpQuery](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpquery-trait)                 | Query Parameters |
+| [httpResponseCode](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpresponsecode-trait)   | Status Code      |
+
+Each of these configurable parts must therefore be logged cautiously.
+
+### Scope and Guidelines
+
+It is unfeasible to make the logging of sensitive data forbidden a type theoretic invariant. With the current API, the customer will always have an opportunity to log a request containing sensitive data before it enters the `Service<Request<B>>` that we provide to them.
```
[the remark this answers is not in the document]
hlbarber: https://github.com/awslabs/smithy-rs/pull/1536/commits/9ec9af84b4106e6ebc7da0e3540a09241186ef5f

design/src/rfcs/rfc0017_logging_sensitive.md
```diff
@@ -0,0 +1,407 @@
...
+        fut.instrument(span)
+    }
+}
+```
+
+### HTTP Debug/Display Wrappers
+
+The `Service::call` path, seen in [Code Generated Logging Middleware](#code-generated-logging-middleware), is latency-sensitive. Careful implementation is required to avoid excess allocations during redaction of sensitive data. Wrapping [Uri](https://docs.rs/http/latest/http/uri/struct.Uri.html) and [HeaderMap](https://docs.rs/http/latest/http/header/struct.HeaderMap.html) then providing a new [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html)/[Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) implementation which skips over the sensitive data is preferable over allocating a new `String`/`HeaderMap` and then mutating it.
+
+These wrappers should be provided alongside the `Sensitive` struct described in [Debug Logging](#debug-logging). If they are implemented on top of `Sensitive`, they will inherit the same behavior - allowing redactions to be toggled using `undredacted-logging` feature flag.
```
jdisanti, line 221:
```suggestion
These wrappers should be provided alongside the `Sensitive` struct described in [Debug Logging](#debug-logging). If they are implemented on top of `Sensitive`, they will inherit the same behavior - allowing redactions to be toggled using `unredacted-logging` feature flag.
```

jdisanti (approved):
LGTM!

hlbarber:
@drganjoo Mentioned a good point, echoing it here for posterity.

> There might be a want for `{redacted}` to instead be some sort of hash of the underlying value. This would obey the spec but allow for developers to identify equivalent values.

The [current implementation](https://github.com/awslabs/smithy-rs/pull/1550) of this RFC allows for this to be added in the future without much pain.

crisidev (approved):

design/src/rfcs/rfc0018_logging_sensitive.md
```diff
@@ -0,0 +1,407 @@
...
+    });
+```
+
+A more subtle violation of the specification may occur when the customer enables verbose logging - a third-party dependency might simply log data marked as sensitive, for example `tokio` or `hyper`.
+
+These two cases illustrate that `smithy-rs` can only prevent violation of the specification in a restricted scope - logs emitted from generated code and the runtime crates. A `smithy-rs` specific guideline should be available to the customer which outlines how to avoid violating the specification in areas outside of our control.
+
+### Routing
+
+The sensitivity and HTTP bindings are declared within specific structures/operations. For this reason, in the general case, it's unknowable whether or not any given part of a request is sensitive until we determine which operation is tasked with handling the request and hence which fields are bound. Implementation wise, this means that any middleware applied _before_ routing has taken place cannot log anything sensitive without performing routing logic itself.
```
david-perez, line 66: > cannot log anything sensitive without performing

`cannot log anything potentially sensitive without performing`

design/src/rfcs/rfc0018_logging_sensitive.md
```diff
@@ -0,0 +1,407 @@
...
+  - they contain no potentially sensitive data, or
+  - the request failed to route, in which case it's not subject to the constraints of an operation.
+
+### Runtime Crates
+
+The crates existing in `rust-runtime` are not code generated - their source code is agnostic to the specific model in use. For this reason, if such a crate wanted to log potentially sensitive data then there must be a way to conditionally toggle that log without manipulation of the source code. Any proposed solution must acknowledge this concern.
+
+## Proposal
+
+This proposal serves to honor the sensitivity specification via code generation of a logging middleware which is aware of the sensitivity, together with a developer contract disallowing logging potentially sensitive data in the runtime crates. An internal and external guideline should be provided in addition to the middleware.
```
david-perez, line 81: > internal and external guideline

Why would there be a need for two guidelines?

design/src/rfcs/rfc0018_logging_sensitive.md
```diff
@@ -0,0 +1,407 @@
...
+
+In which case the branch above becomes
+
+```rust
+debug!(sensitive_data = %Sensitive(data));
+```
+
+### Code Generated Logging Middleware
+
+Using the smithy model, for each operation, a logging middleware should be generated. Through the model, the code generation knows which fields are sensitive and which HTTP bindings exist, therefore the logging middleware can be careful crafted to avoid leaking sensitive data.
```
david-perez, line 139: carefully crafted

design/src/rfcs/rfc0018_logging_sensitive.md
```diff
@@ -0,0 +1,407 @@
...
+
+This is a general problem, not specific to this proposal. For example, [Use Request Extensions](#use-request-extensions) must also solve this problem.
+
+Fortunately, this problem is separable from the actual implementation of the logging middleware and we can get immediate benefit by application of it in the suboptimal position described above.
+
+### Logging within the Router
+
+There is need for logging within the `Router` implementation - this is a crucial area of business logic. As mentioned in the [Routing](#routing) section, we are permitted to log potentially sensitive data in cases where requests fail to get routed to an operation.
```
david-perez, line 249: Will we be able to correlate the log statement of a request that failed to be routed successfully (where we log potentially sensitive data) with previous log statements pertaining to the same request (where we are not allowed to log anything potentially sensitive)?

design/src/rfcs/rfc0018_logging_sensitive.md
david-perez: I read the sections for the guidelines below and still don't understand why we need two guidelines. Aren't all the bullet points in those sections useful for both internal and external developers?

design/src/rfcs/rfc0018_logging_sensitive.md
hlbarber: I don't think there is a general solution we can enforce to provide this correlation. The customer is always free to apply middleware to the outside of the `Router` and log in it anyway they want.

For example:
```rust
let svc = router
    .map_request(|request| {
            if request.len() > 5 { info!("hello world"); }
            request
    });
```

What the customer could do is to attach a unique ID to a span which is open during a requests lifetime. The unique ID would adjoined to all the logging messages and so would provide the correlation. You'd then get a log like

```
span(id = 342432)
    -> info(message = "hello world")
    -> error(message = "failed to route")
```

@drganjoo was eluding to this the other day. https://github.com/awslabs/smithy-rs/pull/1536#issuecomment-1192766074 relates to it.

design/src/rfcs/rfc0018_logging_sensitive.md
hlbarber: Good point.

To clarify, by external developers I mean someone who has a model and wants to use `smithy-rs` to construct a server. In this case the external developer would be composing/initializing `Subscribers`, whereas the internal developer wouldn't be. This means internal developers, while just using `tracing` crate, aren't concerned with:

> - Sensitive data leaking from third-party dependencies.
> - Sensitive data leaking from middleware applied to the `Router`.

There might also be a slight difference in emphasis about certain things:

- The internal developer needs to very familiar with the concept of "potentially sensitive data" because, outside of codegen, they cannot make assumptions about the model.
- The external developer knows the model they're using and therefore knows exactly which data is sensitive.

But yeah, these differences are probably not large enough to warrant a separate guide.

PR #1543, opened 2022-07-08
Use endpoint config from aws-sdk-rust

jdisanti:
## Motivation and Context
This PR makes aws-sdk-rust the source of truth for endpoints and default configuration so that those can both be updated via automation separately from smithy-rs releases.

## Testing
- [x] Generated the full SDK from aws-sdk-rust with the changes from https://github.com/awslabs/aws-sdk-rust/pull/578 and ran `cargo test` on the resulting SDK

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkSettings.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkSettings.kt
@@ -7,6 +7,8 @@ package software.amazon.smithy.rustsdk
 import software.amazon.smithy.model.node.ObjectNode
 import software.amazon.smithy.rust.codegen.smithy.CoreRustSettings
 import software.amazon.smithy.rust.codegen.util.orNull
+import java.nio.file.Path
+import java.nio.file.Paths
 
 /**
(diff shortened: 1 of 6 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkSettings.kt
```diff
@@ -17,6 +19,20 @@ class SdkSettings private constructor(private val awsSdk: ObjectNode?) {
             SdkSettings(coreRustSettings.customizationConfig?.getObjectMember("awsSdk")?.orNull())
     }
 
+    /** Path to the `sdk-default-configuration.json` config file */
+    val defaultsConfigPath: Path get() =
```
rcoh, line 23: should probably note in comments that this is unused currently

rcoh (approved):

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkSettings.kt
jdisanti: I think this kind of comment is highly likely to get out of date and cause confusion.

PR #1545, opened 2022-07-11
RFC: Customizable client operations

jdisanti:
## Motivation and Context
This RFC refines a solution that was briefly outlined in https://github.com/awslabs/aws-sdk-rust/issues/537 to the problem of easily customizing the HTTP requests that the SDK will make.

[Rendered view](https://github.com/awslabs/smithy-rs/blob/jdisanti-rfc-cust-op/design/src/rfcs/rfc0017_customizable_client_operations.md)

Shout out to @rcoh and @Velfi for all the ideas on how we could approach this!

### DIFF
--- a/design/src/rfcs/rfc0017_customizable_client_operations.md
+++ b/design/src/rfcs/rfc0017_customizable_client_operations.md
@@ -11,7 +11,7 @@ client has to be used to create an operation, and then the HTTP request augmente
 that operation type. For example:
 
 ```rust
-let input = SomeOperation::builder().some_value(5).build()?;
+let input = SomeOperationInput::builder().some_value(5).build()?;
 
 let operation = {
     let op = input.make_operation(&service_config).await?;
--- a/design/src/rfcs/rfc0017_customizable_client_operations.md
+++ b/design/src/rfcs/rfc0017_customizable_client_operations.md
@@ -1,0 +1,185 @@
+RFC: Customizable Client Operations
+===================================
+
+> Status: RFC
+
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+SDK customers occasionally need to add additional HTTP headers to requests, and currently,
+the SDK has no easy way to accomplish this. At time of writing, the lower level Smithy
+client has to be used to create an operation, and then the HTTP request augmented on
+that operation type. For example:
+
+```rust
+let input = SomeOperation::builder().some_value(5).build()?;
+
+let operation = {
+    let op = input.make_operation(&service_config).await?;
+    let (request, response) = op.into_request_response();
+
+    let request = request.augment(|req, _props| {
+        req.headers_mut().insert(
+            HeaderName::from_static("x-some-header"),
+            HeaderValue::from_static("some-value")
+        );
+        Result::<_, Infallible>::Ok(req)
+    })?;
+
+    Operation::from_parts(request, response)
+};
+
+let response = smithy_client.call(operation).await?;
+```
+
+This approach is both difficult to discover and implement since it requires acquiring
+a Smithy client rather than the generated fluent client, and it's anything but ergonomic.
+
+This RFC proposes an easier way to augment requests that is compatible with the fluent
+client.
+
+Terminology
+-----------
+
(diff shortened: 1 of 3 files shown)


design/src/rfcs/rfc0017_customizable_client_operations.md
```diff
@@ -0,0 +1,185 @@
...
+
+```rust
+pub struct CustomizableOperation<O, R> {
+    handle: Arc<Handle>,
+    operation: Operation<O, R>,
+}
+
+impl<O, R> CustomizableOperation<O, R> {
+    // Allows for customizing the operation's request
+    fn map_request<E>(
```
rcoh, line 80: we could also add `mutate_request` for the common case where we don't need to take ownership and the changes are infallible

fn mutate_request(self, f: impl FnOnce(&mut Request<SdkBody)) -> Self { ... }

design/src/rfcs/rfc0017_customizable_client_operations.md
```diff
@@ -0,0 +1,185 @@
...
+        HeaderValue::from_static("some-value")
+    );
+let response = operation.send().await?;
+```
+
+### Eliminating `async` from `make_operation`
+
+To avoid requiring an `await` after the call to `customize`, the `make_operation` function
+generated on input structs needs to be made synchronous. The `make_operation` function
+originally was synchronous, but was made async during the implementation of the Glacier
```
rcoh, line 145: I think I'd probaby keep it async and require an await after customize. I think keeping make_operation async is fairly important from the perspective of keeping the design forwards compatible

design/src/rfcs/rfc0017_customizable_client_operations.md
```diff
@@ -0,0 +1,185 @@
...
+
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+SDK customers occasionally need to add additional HTTP headers to requests, and currently,
+the SDK has no easy way to accomplish this. At time of writing, the lower level Smithy
+client has to be used to create an operation, and then the HTTP request augmented on
+that operation type. For example:
+
+```rust
+let input = SomeOperation::builder().some_value(5).build()?;
```
Velfi, line 14:
```suggestion
let input = SomeOperationInput::builder().some_value(5).build()?;
```

I think it's easy for people to get confused about operations vs inputs and so we should clarify that this is an input for an operation and not the operation itself.

jdisanti: Yeah, I intended this to be the input. Good catch.

design/src/rfcs/rfc0017_customizable_client_operations.md
jdisanti: Good idea!

rcoh (approved):
LGTM...galaxy brain thought: with this change can we make `make_operation` private...?
I think not quite because of how we use STS in credentials

design/src/rfcs/rfc0017_customizable_client_operations.md
```diff
@@ -0,0 +1,198 @@
...
+        f: impl FnOnce(Request<SdkBody>) -> Result<Request<SdkBody>, E>,
+    ) -> Result<Self, E> {
+        let (request, response) = self.operation.into_request_response();
+        let request = request.augment(|req, _props| f(req))?;
+        self.operation = Operation::from_parts(request, response);
+        Ok(self)
+    }
+
+    // Convenience for `map_request` where infallible direct mutation of request is acceptable
+    fn mutate_request<E>(
```
DanielBauman88, line 91: This function doesn't use the type parameter <E> which was declared.

I think this is causing me to be unable to compile due to

```
154 |             .mutate_request(|req| {
    |              ^^^^^^^^^^^^^^ cannot infer type for type parameter `E` declared on the associated function `mutate_request`
```

code:

```
            .mutate_request(|req| {
                if let Some(request_id) = request_id {
                    req.headers_mut().insert(
                        HeaderName::from_static("x-amzn-requestid"),
                        HeaderValue::from_static(request_id),
                    );
                }
            })
```

jdisanti: This was fixed in https://github.com/awslabs/smithy-rs/pull/1835, which will be included in the next SDK release.

PR #1548, opened 2022-07-12
Feature: make retry strategy exponential backoff multiplier configurable

Velfi:
_NOTE: I plan to allow for user-defined retry strategies in a separate PR._

## Motivation and Context

aws-sdk-rust#567

## Description
Updated SDK Client retry behavior to allow for a configurable backoff multiplier.

Previously, the backoff multiplier (named `r` in the code) was set to 2 seconds. This is not an ideal default for services like DynamoDB that expect clients to quickly retry failed request attempts. Now, users can set quicker (or slower) multipliers according to their needs.

```rust
#[tokio::main]
async fn main() -> Result<(), aws_sdk_dynamodb::Error> {
    let retry_config = aws_smithy_types::retry::RetryConfigBuilder::new()
        .max_attempts(4)
        .backoff_multiplier(Duration::from_millis(20));

    let shared_config = aws_config::from_env()
        .retry_config(retry_config)
        .load()
        .await;

    let client = aws_sdk_dynamodb::Client::new(&shared_config);

    // Given the 20ms backoff multiplier, and assuming this request fails 3 times before succeeding,
    // the first retry would take place between 0-20ms after the initial request,
    // the second retry would take place between 0-40ms after the first retry,
    // and the third retry would take place between 0-80ms after the second retry.
    let request = client
        .put_item()
        .table_name("users")
        .item("username", "Velfi")
        .item("account_type", "Developer")
        .send().await?;

    Ok(())
}
```

**For a request that gets retried 3 times, when base is 1 and initial_backoff is 2 seconds:**
- the first retry will occur after 0 to 2 seconds
- the second retry will occur after 0 to 4 seconds
- the third retry will occur after 0 to 8 seconds

**For a request that gets retried 3 times, when base is 1 and initial_backoff is 30 milliseconds:**
- the first retry will occur after 0 to 30 milliseconds
- the second retry will occur after 0 to 60 milliseconds
- the third retry will occur after 0 to 120 milliseconds

## Testing

I wrote some test for the backoff calculation and ran existing retry tests

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -39,9 +39,9 @@ author = "calavera"
 
 [[smithy-rs]]
 message = """
-Updated the smithy client's retry behavior to allow for a configurable initial_backoff. Previously, the backoff multiplier
+Updated the smithy client's retry behavior to allow for a configurable initial backoff. Previously, the initial backoff
 (named `r` in the code) was set to 2 seconds. This is not an ideal default for services that expect clients to quickly
-retry failed request attempts. Now, users can set quicker (or slower) multipliers according to their needs.
+retry failed request attempts. Now, users can set quicker (or slower) backoffs according to their needs.
 """
 references = ["aws-sdk-rust#567"]
 meta = { "breaking" = false, "tada" = true, "bug" = false }
@@ -49,9 +49,9 @@ author = "Velfi"
 
 [[aws-sdk-rust]]
 message = """
-Updated SDK Client retry behavior to allow for a configurable backoff multiplier. Previously, the backoff multiplier
+Updated SDK Client retry behavior to allow for a configurable initial backoff. Previously, the initial backoff
 (named `r` in the code) was set to 2 seconds. This is not an ideal default for services like DynamoDB that expect
-clients to quickly retry failed request attempts. Now, users can set quicker (or slower) multipliers according to their
+clients to quickly retry failed request attempts. Now, users can set quicker (or slower) backoffs according to their
 needs.
 
 ```rust
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -25,3 +25,53 @@ message = "Re-export aws_types::SdkConfig in aws_config"
 references = ["smithy-rs#1457"]
 meta = { "breaking" = false, "tada" = true, "bug" = false }
 author = "calavera"
+
+[[smithy-rs]]
+message = """
+Updated the smithy client's retry behavior to allow for a configurable backoff_multiplier. Previously, the backoff multiplier
+(named `r` in the code) was set to 2 seconds. This is not an ideal default for services that expect clients to quickly
+retry failed request attempts. Now, users can set quicker (or slower) multipliers according to their needs.
+"""
+references = ["aws-sdk-rust#567"]
+meta = { "breaking" = false, "tada" = true, "bug" = false }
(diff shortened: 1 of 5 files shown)


rust-runtime/aws-smithy-client/src/retry.rs
```diff
@@ -57,6 +58,7 @@ pub struct Config {
     no_retry_increment: usize,
     timeout_retry_cost: usize,
     max_attempts: u32,
+    backoff_multiplier: Duration,
```
rcoh, line 61: so my first reaction is that a duration being a multiple is a little confusing. Is this more like the initial delay?

rust-runtime/aws-smithy-client/src/retry.rs
```diff
@@ -532,4 +565,34 @@ mod test {
         assert!(no_retry.is_none());
         assert_eq!(policy.retry_quota(), 480);
     }
+
+    #[test]
+    fn calculate_exponential_backoff_n_is_one() {
+        let n = 1.0;
+
+        for (expected, iter) in [(n, 0u32), (2.0, 1), (4.0, 2)] {
```
rcoh, line 573: this was surprisingly hard to load into my brain—does `.enumerate()` work here?

rust-runtime/aws-smithy-client/src/retry.rs
```diff
@@ -532,4 +565,34 @@ mod test {
         assert!(no_retry.is_none());
         assert_eq!(policy.retry_quota(), 480);
     }
+
+    #[test]
+    fn calculate_exponential_backoff_n_is_one() {
+        let n = 1.0;
```
rcoh, line 571: can you add a comment explaining what `n` is here?

rcoh (commented):
this looks great! the one change I'd make is considering renaming `with_backoff_multiplier` to `initial_backoff`—it's what we call it in the interior code and I think it's a bit more grokable. But totally open to discuss this

rust-runtime/aws-smithy-client/src/retry.rs
Velfi: Your suggestion to rename this one is a good one. It is a thing that we multiply the base exponential backoff duration by but I agree calling it initial_backoff will aid users in constructing a more helpful mental model of how it works.

rust-runtime/aws-smithy-client/src/retry.rs
Velfi: yeah it does, I'll switch it

CHANGELOG.next.toml
```diff
@@ -36,3 +36,53 @@ message = "Re-export aws_types::SdkConfig in aws_config"
...
+(named `r` in the code) was set to 2 seconds. This is not an ideal default for services that expect clients to quickly
+retry failed request attempts. Now, users can set quicker (or slower) multipliers according to their needs.
+"""
+references = ["aws-sdk-rust#567"]
+meta = { "breaking" = false, "tada" = true, "bug" = false }
+author = "Velfi"
+
+[[aws-sdk-rust]]
+message = """
+Updated SDK Client retry behavior to allow for a configurable backoff multiplier. Previously, the backoff multiplier
```
rcoh, line 52: backoff multiplier -> initial delay

rcoh (approved):
Nice! LGTM

Issue #1364, opened 2022-05-04
Simplify conversion from HTTP body error type to `RequestRejection`


david-perez:
As mentioned in this comment:

https://github.com/awslabs/smithy-rs/blob/ae9a5ed50a4b14296f4ede85b65f5313f1398df3/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt#L328

This bound, that "trickles up" all the way to the `OperationRegistry` from the function that parses requests for a given operation:

```rust
aws_smithy_http_server::rejection::RequestRejection: From<<B as aws_smithy_http_server::body::HttpBody>::Error> {
```

Is only needed when we convert the body [`to_bytes`](https://docs.rs/hyper/latest/hyper/body/fn.to_bytes.html) which is only required in the (very likely) scenario that the modeled operation input has data bound to the HTTP body.

However:
1. the bound is generated _unconditionally_; and
2. we could simplify things substantially if we converted to a `RequestRejection` _in-place_ when the conversion fails. This would make the bounds go away, and remove the need for having to add converters like the below for every body type we want to support.

https://github.com/awslabs/smithy-rs/blob/ae9a5ed50a4b14296f4ede85b65f5313f1398df3/rust-runtime/aws-smithy-http-server/src/rejection.rs#L222-L226

Issue #1125, opened 2022-01-27
Make `aws_smithy_http::byte_stream::ByteStream` not expose `SdkBody`


david-perez:
As mentioned here https://github.com/awslabs/smithy-rs/pull/1023#discussion_r784751006

> We have decided that we will use `aws_smithy_http::byte_stream::ByteStream`,
> making all methods referencing `SdkBody` `#[doc(hidden)]` (`new`, `into_inner`,
> `From<SdkBody>` impl), and we will add generic methods that don’t reference
> `SdkBody`.

This change is to allow server users to return `ByteStream`s in their output shapes, for streaming responses. The server so far has not used `SdkBody`, with all its code working against [`http_body::Body`](https://docs.rs/http-body/latest/http_body/trait.Body.html).

david-perez:
There's some prior work [here](https://github.com/awslabs/smithy-rs/commit/cd7c27406fd07ff5cbda77e17480144a27618473), making `ByteStream` and `FsBuilder` generic over the body type, but `FsBuilder::build` has in its implementation:

```rust
Ok(ByteStream::new(SdkBody::retryable(body_loader)))
```

And it wasn't rewritten successfully with the same semantics with an arbitrary body `B`.

david-perez:
We also want `ByteStream` to be constructible from an arbitrary `Stream`, which might be an easier thing to do first.

PR #1551, opened 2022-07-16
feat: support for server `lambda_http::Request`

hugobast:
## Motivation and Context

This PR replaces: https://github.com/awslabs/smithy-rs/pull/1338

Adds support for accepting `lambda_http::Request`.

## Testing

Pokemon Service is updated with a lambda_http run example.

## Checklist


rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -24,6 +24,7 @@ futures-util = { version = "0.3", default-features = false }
 http = "0.2"
 http-body = "0.4"
 hyper = { version = "0.14", features = ["server", "http1", "http2", "tcp", "stream"] }
+lambda_http = { git = "https://github.com/awslabs/aws-lambda-rust-runtime" }
```
hugobast, line 27: A release isn't cut yet on this crate. This PR necessitates the latest changes from https://github.com/awslabs/aws-lambda-rust-runtime/pull/491

rust-runtime/aws-smithy-http-server/examples/pokemon_service/Cargo.toml
```diff
@@ -9,6 +9,7 @@ description = "A smithy Rust service to retrieve information about Pokémon."
 [dependencies]
 clap = { version = "~3.2.1", features = ["derive"] }
 hyper = {version = "0.14", features = ["server"] }
+lambda_http = { git = "https://github.com/awslabs/aws-lambda-rust-runtime" }
```
hugobast, line 12: https://github.com/awslabs/smithy-rs/pull/1551/commits/ab87d3a6fe47f2d280670c0de76869fa8fe4f0d7#r922617338

crisidev (approved):
LGTM. Do you mind to fix the conflict so we can merge.

rust-runtime/aws-smithy-http-server/src/rejection.rs
```diff
@@ -248,3 +248,6 @@ convert_to_request_rejection!(std::str::Utf8Error, InvalidUtf8);
 // tests use `[crate::body::Body]` as their body type when constructing requests (and almost
 // everyone will run a Hyper-based server in their services).
 convert_to_request_rejection!(hyper::Error, HttpBody);
+
+// In order to use `Router<lambda_http::Body>` this line works around [Simplify conversion from HTTP body error type to RequestRejection](https://github.com/awslabs/smithy-rs/issues/1364)
+convert_to_request_rejection!(Box<dyn std::error::Error + Send + Sync>, HttpBody);
```
david-perez, line 253:
```suggestion
convert_to_request_rejection!(lambda_http::Error, HttpBody);
```

You want to refer to the type declaration in the dependent crate right? The dependency on `lambda_http` in `Cargo.toml` is not used otherwise.

david-perez (commented):
Sorry it took me so long to take a look at this, I was on holiday.

> I believe the last thing that stands in the way is https://github.com/awslabs/smithy-rs/issues/1364

#1364 is unrelated to this change. What this change does is to allow smithy-rs to take in `lambda_http:Request`s as input, by adding a converter from Lambda's body error type into a smithy-rs `RequestRejection`. This will get returned when we fail to buffer non-streaming requests as bytes. You might want to update the PR description / commit message to better reflect the intent of this patch.

> I could use some help devising a test; the way I've tested this is by changing the pokemon service to use lambda_http::run in this fashion and witnessing that the code compiles:

It'd be nice if we could showcase a working Lambda in the repo as an example project. #1338 includes a nice example you can copy over. You'll also need to copy over the stage stripping part from the request URI path. Do you want to tackle this in a separate PR?

> I am unsure what to put under the CHANGELOG for this change; please advise.

A simple message along the lines of "Add support for accepting input `lambda_http::Request`s in the server" will do.

---

Thank you!

hugobast:
> Sorry it took me so long to take a look at this, I was on holiday.

I'm currently taking a week off too, I'll get to this on Monday. Thanks!

hugobast:
👋 Back dusting this off and see if I can finish the last leg

> https://github.com/awslabs/smithy-rs/issues/1364 is unrelated to this change. What this change does is to allow smithy-rs to take in lambda_http:Requests as input, by adding a converter from Lambda's body error type into a smithy-rs RequestRejection. This will get returned when we fail to buffer non-streaming requests as bytes. You might want to update the PR description / commit message to better reflect the intent of this patch.

I probably had the wrong thing to paste here; I believe I intended to link to this comment: https://github.com/awslabs/smithy-rs/pull/1338#issuecomment-1117643522 which had a comment about a workaround to #1364. Sorry for the confusion.

> You'll also need to copy over the stage stripping part from the request URI path. Do you want to tackle this in a separate PR?

I'd like to get this all done in this PR, so I'll update accordingly,

Thanks for the feedback!

rust-runtime/aws-smithy-http-server/src/rejection.rs
```diff
@@ -262,3 +262,6 @@ convert_to_request_rejection!(std::str::Utf8Error, InvalidUtf8);
 // tests use `[crate::body::Body]` as their body type when constructing requests (and almost
 // everyone will run a Hyper-based server in their services).
 convert_to_request_rejection!(hyper::Error, HttpBody);
+
+// In order to use `Router<lambda_http::Body>` this line works around [Simplify conversion from HTTP body error type to RequestRejection](https://github.com/awslabs/smithy-rs/issues/1364)
```
david-perez, line 266:
```suggestion
// Required in order to accept Lambda HTTP requests using `Router<lambda_http::Body>`.
```

david-perez (changes requested):
Nice!

Do you want to copy over Eduardo's `cargo lambda` config (see `rust-runtime/aws-smithy-http-server/examples/{Makefile, README.md}` [in his PR](https://github.com/awslabs/smithy-rs/pull/1338/files#diff-ff2570520b8f25e4ad20978444bfe85ec5c7fc27b56937ec5730c9164273a4d6)). It would run the new Lambda executable example.

Can you also update the `CHANGELOG.next.toml` saying that there is now a canonical and easy way to run smithy-rs on Lambda, linking to the example? We have users doing the old request-response mapping + HTTP URI path stage stripping in their applications, and they'll be happy to know they can get rid of them. Call out in the changelog that there are now converters between `SdkBody` and `lambda_http::Body`.

david-perez:
In your Lambdas, how are you handling usage of [the `Lambda` context](https://docs.rs/lambda_http/latest/lambda_http/ext/trait.RequestExt.html#tymethod.lambda_context) in your operation handlers? It's useful for X-Ray integration, logging a request ID etc.

Our users currently have to retrieve it out of the `lambda_http::Request` and insert it into the HTTP extensions with their state. It's not a great UX. I was wondering if you had any ideas to make it better.

hugobast:
New information

> You'll also need to copy over the stage stripping part from the request URI path. Do you want to tackle this in a separate PR?

I had to do some research yesterday since stage stripping wasn't giving me the expected result. It turns out that as implemented in https://github.com/awslabs/smithy-rs/pull/1338 we were not stripping anything.

```
let raw_path = request.raw_http_path()
/// raw_path = "" (empty)
```

The routing worked at a specific version (sha: https://github.com/eduardomourar/smithy-rs/commit/e05b302b058e21cc4d2f4d4acd63b3dd6b734972) because of how the server initializes RequestSpec regexes. 

Example:

```
uri_path_regex: /orders/123/shipments$
```

It's subtle but it lacks the start of line assert `^` which is present in current versions of smithy-rs. That old regex will match a path like this `/prod/orders/123/shipments` and more relevant for me it will match a versioned path like this `/prod/2022-01-01/orders/123/shipments`.

The new regexes look like this:

```
uri_path_regex: ^/orders/123/shipments$
```

After stage stripping it will still not match the versioned path that I have `/2022-01-01/orders/123/shipments`.

I'm not sure what to do about stripping `/2022-01-01` if any of this sparks a thought for anyone reading please let me know.

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -250,6 +250,22 @@ impl From<hyper::Body> for SdkBody {
...
+        };
+
+        SdkBody {
+            inner: Inner::Streaming { inner: hyper_body },
+            rebuild: None,
+            callbacks: Vec::new(),
+        }
+    }
+}
+
```
rcoh, line 268: we definitely don't want an un-gated dependency on `lambda_http` in `aws-smithy-http`. Can we keep these converters isolated? I'd probably suggest refactoring the way this works so that there is a "labmda" module that can be conditionally enabled, preferably in the server crate

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -445,6 +445,12 @@ impl From<hyper::Body> for ByteStream {
     }
 }
 
+impl From<lambda_http::Body> for ByteStream {
+    fn from(input: lambda_http::Body) -> Self {
+        ByteStream::new(SdkBody::from(input))
+    }
+}
```
rcoh, line 452: this shouldn't be needed since `lambda_http::Body` can be converted into `BoxBody` right?

rust-runtime/aws-smithy-http-server/src/routing/into_make_lambda_service.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+impl<S> Service<Request> for IntoMakeLambdaService<S>
+where
+    S: Service<Request>,
+{
+    type Error = S::Error;
+    type Response = S::Response;
+    type Future = S::Future;
```
rcoh, line 34: seems like this is where we might want to do the body conversion?

rcoh (changes requested):
we'll need to figure out how to make this work without adding an unconditional lambda dependency to aws-smithy-http

rust-runtime/aws-smithy-http/src/body.rs
hugobast: I thought this might be the case, I added the code to get the feedback, I'll see about doing a lambda module.

rust-runtime/aws-smithy-http-server/src/routing/into_make_lambda_service.rs
hugobast: Changes were made in order to avoid doing a body conversion from a `lambda_http::Body` to a `hyper::Body` it's unfortunate that we would have to do a body conversion here anyway for `ByteStream`?

rust-runtime/aws-smithy-http/src/byte_stream.rs
david-perez: It's to support the event streaming case. We have an `B: Into<aws_smithy_http::byte_stream::ByteStream>` bound on server operation handlers that do event streams, because eventually there's an

```
crate::http_serde::deser_payload_capture_pokemon_operation_capture_pokemon_operation_input_events(
    &mut body.into().into_inner(),
)?,
```

We convert the request body type (remember in the server we support all body types so long as it implements `HttpBody`) into `SdkBody` before sending it to the deserializer, which just works with `SdkBody`. All of this would go away if we tackled https://github.com/awslabs/smithy-rs/issues/1125 and made `ByteStream` work with all body types.

rust-runtime/aws-smithy-http-server/src/routing/into_make_lambda_service.rs
david-perez: Indeed, it's unfortunate. There's an issue to make `ByteStream` work with all body types: https://github.com/awslabs/smithy-rs/issues/1125 but we haven't been able to complete it.

hugobast:
> I'm not sure what to do about stripping /2022-01-01 if any of this sparks a thought for anyone reading please let me know.

Stepped back a bit and I believe this is a "userland" concern, customers (like me/us) can strip this out in a layer.

rust-runtime/aws-smithy-http-server/src/routing/into_make_lambda_service.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use http::uri;
+use lambda_http::{Request, RequestExt};
+use std::{
+    fmt::Debug,
+    task::{Context, Poll},
+};
+use tower::Service;
+
+/// A [`MakeService`] that produces AWS Lambda compliant services.
```
hlbarber, line 14: I don't think this qualifies as a [MakeService](https://docs.rs/tower/latest/tower/make/trait.MakeService.html#implementors). A `MakeService` is roughly a `Service` which given a request produces a `Service` as it's `Response`.

`hyper` requires you provide a ~`MakeService` to `Builder::serve` however `http_lambda::run` requires you provide a ~`Service<http::Request, Response = http::Response>`.

hugobast:
@david-perez 

> In your Lambdas, how are you handling usage of [the Lambda context](https://docs.rs/lambda_http/latest/lambda_http/ext/trait.RequestExt.html#tymethod.lambda_context) in your operation handlers? It's useful for X-Ray integration, logging a request ID etc.

> Our users currently have to retrieve it out of the lambda_http::Request and insert it into the HTTP extensions with their state. It's not a great UX. I was wondering if you had any ideas to make it better.

I've asked a team member about this; I don't think we do anything with the Lambda context yet. I wasn't involved with our X-Ray integration, but I know we have one, and it works.

Here's what I know we have made an ext trait like this one

```
pub trait RequestExt {
    fn account_id(&self) -> Option<&str>;
    fn request_id(&self) -> Option<&str>;
}
```

And `Request<T>` implements `RequestExt`.

rust-runtime/aws-smithy-http-server/src/routing/into_make_lambda_service.rs
hlbarber: `struct` name has changed but looks like the docs still refer to it as a `MakeService`.

hugobast: Thanks, fixed in: https://github.com/awslabs/smithy-rs/pull/1551/commits/32762e5044ed17fb721726b5e0b3ba8d1e37053d

david-perez (approved):

david-perez:
> @david-perez thanks for the review, will work through it today. How are other users dealing with having versioning in their path?

Is the version an API Gateway stage? If so, they are stripping it using a similar function than `convert_event` from this PR, applying it as middleware. The Smithy model does not include the stage in the URI, since it's deployed to multiple stages.

david-perez:
> Maybe this is related, for things like extensions on Request, we've made a RequestExt trait like this:

At what point in the request lifecycle do you get to use the `account_id` and `request_id` methods? Do you thread those strings to the operation handler functions? If yes, how?

hugobast:
> At what point in the request lifecycle do you get to use the account_id and request_id methods? Do you thread those strings to the operation handler functions? If yes, how?

Oh right, no, we don't have those in our operation handlers; sorry to mislead, they are only available in middlewares (layers/services)

hugobast:
@david-perez I made a new set of changes (and rebased) to get the two jobs that were failing to pass

- https://github.com/awslabs/smithy-rs/runs/7903857949?check_suite_focus=true
- https://github.com/awslabs/smithy-rs/runs/7903858148?check_suite_focus=true

rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/lambda.rs
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// This program is exported as a binary named `pokemon_service`.
```
david-perez, line 6:
```suggestion
// This program is exported as a binary named `pokemon-service-lambda`.
```

CHANGELOG.next.toml
```diff
@@ -107,3 +107,13 @@ Servers now allow requests' ACCEPT header values to be:
...
 meta = { "breaking" = true, "tada" = false, "bug" = false, "target" = "server" }
 author = "82marbag"
+
+[[smithy-rs]]
+message = """
+There is a canonical and easier way to run smithy-rs on Lambda [see example].
+
+[see example]: https://github.com/awslabs/smithy-rs/blob/031376bf96fa9d0262e15336a0e20a99d43112cc/rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/lambda.rs
+"""
+references = ["smithy-rs#1551]
```
david-perez, line 117:
```suggestion
references = ["smithy-rs#1551"]
```

CHANGELOG.next.toml
```diff
@@ -107,3 +107,13 @@ Servers now allow requests' ACCEPT header values to be:
 references = ["smithy-rs#1544"]
 meta = { "breaking" = true, "tada" = false, "bug" = false, "target" = "server" }
 author = "82marbag"
+
+[[smithy-rs]]
+message = """
+There is a canonical and easier way to run smithy-rs on Lambda [see example].
+
+[see example]: https://github.com/awslabs/smithy-rs/blob/031376bf96fa9d0262e15336a0e20a99d43112cc/rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/lambda.rs
```
david-perez, line 115:
```suggestion
[see example]: https://github.com/awslabs/smithy-rs/blob/main/rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/lambda.rs
```

david-perez:
@hugobast Thanks for your contribution.

> Changes were made in order to avoid doing a body conversion from a lambda_http::Body to a hyper::Body it's unfortunate that we would have to do a body conversion here anyway for ByteStream?

We will try to address this, perhaps via #1125, so that `convert_event` does not have to make the body type conversion.

PR #1563, opened 2022-07-21
aws-smithy-http-server crates to be published as part of release cycle

drganjoo:
## Motivation and Context

aws-smithy-http-server runtime crate should be automatically published on each release cycle.

## Description

On every release cycle of smithy-rs, aws-smithy-* crates are published but aws-smithy-http-server crate is skipped. The server is now stable enough that it should be published along with the rest of the smithy-rs crates. 

## Testing

1. changed version in the Cargo.toml to 0.46.0 
2. changed dependency paths to versions for aws-smithy-* crates
3. ran cargo publish --dry-run --allow-dirty to verify that changes work

## Checklist


rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -10,8 +10,7 @@ categories = ["asynchronous", "web-programming", "api-bindings"]
 description = """
 Server runtime for Smithy Rust Server Framework.
 """
-# until this is not stable, it is not publishable.
-publish = false
+publish = true
```
crisidev, line 13: This should be done also for `aws-smithy-http-server-python`

crisidev (changes requested):

rust-runtime/aws-smithy-http-server/Cargo.toml
drganjoo: Done. Have changed the Cargo.toml for python to be auto published as well.

jdisanti: If you want it to be included in the automated smithy-rs releases, it will need to be added to this list as well: https://github.com/awslabs/smithy-rs/blob/b77c66cbf72e8a6083af87cf125fd7d84eaa69a4/buildSrc/src/main/kotlin/CrateSet.kt#L34

crisidev (approved):

jdisanti (approved):

Issue #356, opened 2021-05-07
Generate `#[deprecated]` based on the smithy deprecated trait


rcoh:

tanadeau:
I'd like to try this as my first issue.

PR #1570, opened 2022-07-25
Basic support of @deprecated trait in Smithy model

weihanglo:
## Motivation and Context

#356

Basic support of [@deprecated] trait in Smithy model.

#### Same feature parity between Smithy and Rust 

The feature parity of [`@deprecated`] trait in Smithy model is exact the same as [`#[deprecated]`][rust-deprecated] attribute in Rust. Both have a since field to indicate when it has been deprecated, with an additional note/message field.

#### Service shapes are not supported

This patch mostly takes the implementation of documentation trait as a reference. Everywhere [`documentShape`] being called would get its counterpart of `deprecatedShape`, which is in exact our implementation of generating `#[deprecated]` attribute from `@deprecated` trait. 

As a result, some important shapes haven't yet got support with documentation trait are also not going to get `@deprecated` feature. Namely, service, resource, and operation trait are not supported.

#### Support `deprecated` property in `@enum` trait

The other feature that is not actually a `@deprecated` trait but would start generating `#[deprecated]` attribute is [`@enum`] trait, which has an optional `deprecated` property is the same as deprecated trait in terms of functionality. This part resides in EnumGenerator.kt.

## Testing

The simplest test is play around with `@deprecated` trait with different models.

[`@deprecated`]: https://awslabs.github.io/smithy/1.0/spec/core/documentation-traits.html#deprecated-trait
[rust-deprecated]: https://doc.rust-lang.org/reference/attributes/diagnostics.html?highlight=must_use#the-deprecated-attribute
[`documentShape`]: https://github.com/awslabs/smithy-rs/blob/5edd9d265dd00cb1a2bdbf3cee72463b293a876a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt#L213-L218
[`@enum`]: https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html#enum-trait


82marbag:
We should warn, not error, on deprecated

weihanglo:
> We should warn, not error out, on deprecated

Reasonable to me. 

However, `#[deprecated]` is [at warn level as of 1.62.1](https://doc.rust-lang.org/1.62.1/rustc/lints/listing/warn-by-default.html#deprecated) I suspect the failure was from clippy or something else turning on `-D warnings` or `#![deny(warnings)]`. Are we going to add `#![allow(deprecated)]` for each crate? I am fine with that.

82marbag:
> > We should warn, not error out, on deprecated
> 
> Reasonable to me.
> 
> However, `#[deprecated]` is [at warn level as of 1.62.1](https://doc.rust-lang.org/1.62.1/rustc/lints/listing/warn-by-default.html#deprecated). I suspect the failure was from clippy or something else turning on `-D warnings` or `#![deny(warnings)]`. Are we going to add `#![allow(deprecated)]` for each crate? I am fine with that.

We are on 1.58 and we always remain behind in terms of Rust versions. Let's allow it and add a note to remove it when possible

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
```diff
@@ -375,6 +375,28 @@ sealed class Attribute {
...
+                builder.append("deprecated")
+                if (note.isBlank().and(since.isBlank())) {
+                    // No-op. Rustc would emit a default message.
+                } else if (note.isNotBlank().and(since.isBlank())) {
+                    builder.append("(note = \"${note}\")")
+                } else if (note.isBlank().and(since.isNotBlank())) {
+                    builder.append("(since = \"${since}\")")
+                } else {
+                    builder.append("(note = \"${note}\", since = \"${since}\")")
+                }
```
crisidev, line 396: Escaped quotes can be handled using the `dp()` function:

```kotlin
import software.amazon.smithy.rust.codegen.util.dq

} else if (note.isNotBlank().and(since.isBlank())) {
    builder.append("(note = "${note.qt()}")")
} else if (note.isBlank().and(since.isNotBlank())) {
    builder.append("(since = ${since.dq()})")
} else {
    builder.append("(note = ${note.dq()}, since = ${since.dq()})")
}
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
```diff
@@ -375,6 +375,28 @@ sealed class Attribute {
...
             }
         }
+
+        companion object {
+            /**
+             * Renders a
+             * [`#[deprecated]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute)
+             * attribute.
+             */
+            fun deprecated(note: String = "", since: String = ""): Custom {
```
crisidev, line 385: Why don't you use optional strings here as default, it's a better semantic.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerCombinedErrorGeneratorTest.kt
```diff
@@ -88,6 +95,8 @@ class ServerCombinedErrorGeneratorTest {
                 """
             )
 
+            println("file:///${project.baseDir}/src/lib.rs")
+            println("file:///${project.baseDir}/src/error.rs")
```
crisidev, line 99: Leftover to be removed.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
```diff
@@ -269,6 +271,20 @@ fun <T : AbstractCodeWriter<T>> T.docs(text: String, vararg args: Any, newlinePr
...
 }
 
+/**
+ * Generates a `#[deprecated]` attribute for [shape].
+ */
+fun RustWriter.deprecatedShape(shape: Shape): RustWriter {
+    val deprecatedTrait = shape.getTrait<DeprecatedTrait>() ?: return this
+
+    val note = deprecatedTrait.message.orElse("")
+    val since = deprecatedTrait.since.orElse("")
```
crisidev, line 281: See the comment about optional strings in the file above.

crisidev (changes requested):
We need to add a changelog entry for this PR.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustWriter.kt
weihanglo: I cannot find any comment about that. Could you provide a link to those lines?

82marbag: Or you can use `let` with a nullable. I believe @crisidev refers to https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

crisidev: I am referring to Kotlin nullability: https://kotlinlang.org/docs/null-safety.html#checking-for-null-in-conditions

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
```diff
@@ -375,6 +375,32 @@ sealed class Attribute {
...
+                if (note.isBlank() && since.isBlank()) {
+                    // No-op. Rustc would emit a default message.
+                } else if (note.isNotBlank() && since.isBlank()) {
+                    builder.append("(note = ${note.dq()})")
+                } else if (note.isBlank() && since.isNotBlank()) {
+                    builder.append("(since = ${since.dq()})")
+                } else {
+                    builder.append("(note = ${note.dq()}, since = ${since.dq()})")
+                }
+                return Custom(builder.toString())
```
crisidev, line 401: Probably my comment was not clear in the review before. Now that you have nullable values, you can just use them and Kotlin will figure out the real type if they are not `null`. Something like this should work:

```suggestion
                if (note == null && since.null) {
                    // No-op. Rustc would emit a default message.
                } else if (note != null && since == null) {
                    builder.append("(note = ${note.dq()})")
                } else if (not == null && note != null) {
                    builder.append("(since = ${since.dq()})")
                } else {
                    builder.append("(note = ${note.dq()}, since = ${since.dq()})")
                }
                return Custom(builder.toString())
```

crisidev (changes requested):

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustTypes.kt
weihanglo: Addressed. Thanks for your suggestion!

Velfi (approved):

crisidev (approved):
LGTM!

PR #1575, opened 2022-07-26
Add/alternative runtime tests

Velfi:
## Motivation and Context

This is part of my hack week project.

## Description

This PR adds integration tests demonstrating running the SDK on async runtimes other than tokio.

## Testing

This PR is all tests


aws/sdk/integration-tests/s3/tests/alternative-async-runtime.rs
```diff
@@ -0,0 +1,159 @@
...
+        })
+    }
+}
+
+#[test]
+fn test_smol_runtime_timeouts() {
+    if let Err(err) = smol::block_on(async { timeout_test(Arc::new(SmolSleep)).await }) {
+        println!("{err}");
+        panic!();
+    }
```
jdisanti, line 38: Simpler to just `unwrap`/`expect` the `block_on` call rather than handling the error and manually panicking.

jdisanti (approved):

aws/sdk/integration-tests/s3/tests/alternative-async-runtime.rs
Velfi: I wanted to print the nice display version of the errors instead of the debug version. I can change it if you don't think that's worth it.

jdisanti: `panic!("{err}");`?

It would reduce the lines of code quite a bit without it though 🤷

PR #1578, opened 2022-07-26
Change detailed logs in CredentialsProviderChain from info to debug

lkts:
## Motivation and Context
Those logs are written quite often and in current state there can be ~ 4 log lines for each credential retrieval operation: three for providers that didn't provide credentials and one final for provider that does.

Given that some providers in chain not providing credentials is an expected situation, having those logs on INFO level is quite noisy.

## Description
This change changes level of commonly written logs from info to debug.

## Testing
-

## Checklist


Velfi (approved):

jdisanti:
> ^ I am not sure what should i put there, what's "tada" f.e.?

We usually use tada (:tada:) for new features or things that a lot of people have been asking for. If you don't think this fits any of tada/bug/breaking, then you can just have all three be `false`.

lkts:
I haven't used GitHub in a while and i don't know how to deal with two commits that are from `main` but show up here.

jdisanti (approved):

PR #1581, opened 2022-07-27
Fix `codegen-server` and `codegen-server:python` projects when crate name contains hyphens

david-perez:
Rust crate names are allowed to contain hyphens, while Rust identifiers
are not [0]. However, the Rust and Python server projects are currently
broken because we use the user-provided crate name verbatim
(`moduleName` in `settings`) in places where Rust expects valid
identifiers:

* The Rust server's generated operation registry Rust docs use the crate
  name verbatim in import statements.
* The Python server uses the crate name verbatim to generate a shared
  library, but library target names in `Cargo.toml` cannot contain
  hyphens.

We never caught this bug because we don't generate any crates with
hyphens in their names.

This commit:

* fixes the above usages of verbatim crate names where identifiers are
  expected by calling `toSnakeCase()`,
* modifies the unit test for the server's operation registry generator
  to generate a crate with hyphens in its name,
* changes the Pokémon service server SDK crate names from
  `pokemon_service_sdk` to `pokemon-service-server-sdk`, so that we
  exercise generation of a crate with hyphens in its name, and to better
  convey that the generated SDK is server-specific; and
* changes the Pokémon service client SDK crate name from
  `pokemon_service_client` to `pokemon-service-client`, and changes the
  Pokémon service crate name itself from `pokemon_service` to
  `pokemon-service`, for symmetry.

[0]: https://doc.rust-lang.org/reference/identifiers.html


82marbag (approved):

rcoh (approved):

crisidev (commented):
LGTM

PR #1585, opened 2022-07-27
update: `From<hyper::Body>` for `ByteStream` now delegates to `SdkBody::from`

Velfi:
This makes it easier to create streaming bodies without pulling in runtime crates.
Originally, `ByteStream::from::<hyper::Body>` would internally create a new `SdkBody` by calling `SdkBody::from_dyn` and boxing the `hyper::Body`. This PR updates it to instead call `SdkBody::from::<hyper::Body>`.


jdisanti (commented):
Would be nice to get more context in the PR description. This probably deserves a changelog entry too.

jdisanti (approved):

PR #1588, opened 2022-07-27
Fix the SDK canary

jdisanti:
## Motivation and Context
This PR updates the `canary-runner` to work with `versions.toml` rather than getting versions from the `aws-config` crate for https://github.com/awslabs/aws-sdk-rust/issues/594

## Testing
- [x] [Successfully ran the canary with these changes](https://github.com/awslabs/aws-sdk-rust/runs/7549720790?check_suite_focus=true)


tools/ci-cdk/canary-runner/src/build_bundle.rs
```diff
@@ -0,0 +1,462 @@
...
+        for notable in NOTABLE_SDK_RELEASE_TAGS.iter() {
+            if notable < release_tag {
+                enabled.push(notable.as_str().into());
+            }
+        }
+    }
+    enabled
+}
+
+fn generate_crate_manifest(crate_source: CrateSource) -> Result<String> {
```
Velfi, line 121: It is perhaps overkill, but we could use cargo's stuff for this in the future https://docs.rs/cargo/0.63.1/cargo/core/manifest/struct.Manifest.html

tools/ci-cdk/canary-runner/src/build_bundle.rs
```diff
@@ -0,0 +1,462 @@
...
+        };
+        let bin_path = target_path.join("bootstrap");
+        let bin_hash = sha1_file(&bin_path)?;
+        let bundle_path = target_path.join(format!("canary-lambda-{bin_hash}.zip"));
+
+        let zip_file = fs::File::create(&bundle_path).context(here!())?;
+        let mut zip = zip::ZipWriter::new(zip_file);
+        zip.start_file(
+            "bootstrap",
+            zip::write::FileOptions::default().unix_permissions(0o755),
```
Velfi, line 234: We'll never run these canaries on a non-linux system right?

tools/publisher/src/subcommand/yank_release.rs
```diff
@@ -77,33 +77,12 @@ pub async fn subcommand_yank_release(
...
 async fn acquire_release_from_tag(tag: &str) -> Result<Release> {
-    validate_tag(tag)?;
-
-    let manifest_url = format!(
-        "https://raw.githubusercontent.com/awslabs/aws-sdk-rust/{}/versions.toml",
-        tag
-    );
-    info!("Downloading versions.toml from {}", manifest_url);
-    let manifest_contents = reqwest::get(manifest_url)
+    let tag = ReleaseTag::from_str(tag).map_err(|err| anyhow!("invalid release tag: {}", err))?;
```
Velfi, line 81: I think you can just use anyhow's `context` for this right?

tools/smithy-rs-tool-common/src/release_tag.rs
```diff
@@ -0,0 +1,153 @@
...
+use anyhow::bail;
+use lazy_static::lazy_static;
+use regex::Regex;
+use semver::Version;
+use std::cmp::Ordering;
+use std::fmt;
+use std::str::FromStr;
+
+lazy_static! {
+    static ref VERSION_TAG: Regex = Regex::new(r"^v(\d+)\.(\d+)\.(\d+)$").unwrap();
```
Velfi, line 15: may we never make an RC tagged release 🤞

tools/smithy-rs-tool-common/src/release_tag.rs
```diff
@@ -0,0 +1,153 @@
...
+impl FromStr for ReleaseTag {
+    type Err = anyhow::Error;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        if let Some(caps) = VERSION_TAG.captures(value) {
+            Ok(ReleaseTag::Version(VersionReleaseTag {
+                version: Version::new(
+                    caps.get(1).expect("validated by regex").as_str().parse()?,
+                    caps.get(2).expect("validated by regex").as_str().parse()?,
+                    caps.get(3).expect("validated by regex").as_str().parse()?,
```
Velfi, line 62: In case we ever use a release tag with a diff format, we should probably put an explanatory error here so future us have the context to address the issue.

Velfi (approved):
I had a few comments for possible improvements but they're not required.

tools/ci-cdk/canary-runner/src/build_bundle.rs
jdisanti: Depending on `cargo` would definitely be overkill, but there are some smaller crates that just operate on the Cargo manifest file. The one I've used before ends up serializing all the optional properties that aren't set though, so I'm hesitant to go that route.

tools/ci-cdk/canary-runner/src/build_bundle.rs
jdisanti: That's correct.

tools/smithy-rs-tool-common/src/release_tag.rs
jdisanti: There will be no more versioned release tags. They're all date-based from here on out.

tools/smithy-rs-tool-common/src/release_tag.rs
jdisanti: I think this is covered by line 72.

tools/ci-cdk/canary-lambda/additional-ci
```diff
@@ -12,5 +12,10 @@ if [[ "${GITHUB_ACTIONS}" == "true" ]]; then
...
 fi
 
-./build-bundle --sdk-path "${SDK_PATH}" --manifest-only
-cargo clippy
+# The canary-lambda doesn't have a Cargo.toml in a fresh checkout of smithy-rs, so generate one before checks
+pushd ../canary-runner
+cargo run -- build-bundle --manifest-only --canary-path ../canary-lambda --sdk-path "${SDK_PATH}"
+popd
+
+cargo test --all-features
```
jdisanti, line 20: Can't actually do this since the tests hit actual AWS.
```suggestion
```

PR #1592, opened 2022-07-28
Correctly load client or server specific decorators from classpath

david-perez:
The current approach that attempts to downcast never worked; all Rust
decorators were being loaded, and the cast was doing nothing, because
the generic type parameter is erased at runtime.

Attempting to downcast a generic class `C<T>` to `C<U>` where `U: T` is
not possible to do in Kotlin (and presumably all JVM-based languages)
_at runtime_. Not even when using reified type parameters of inline
functions. See https://kotlinlang.org/docs/generics.html#type-erasure
for details.

This commit thus goes for another approach, suggested in the linked
Stack Overflow question [0]: add a method to the loaded classes that
signals at runtime the generic type parameter (`ClientCodegenContext` or
`ServerCodegenContext`) they can work with, in order to filter them.

This commit also simplifies the way the Python server project loads the Python
server-specific decorators, by deleting the combined decorator
`PythonServerCodegenDecorator`, which was being loaded from the classpath, and
instead directly using `CombinedCodegenDecorator` and passing it the Python
server-specific decorators in the `extras` parameter.

[0]: https://stackoverflow.com/questions/5451734/loading-generic-service-implementations-via-java-util-serviceloader


82marbag (approved):

david-perez:
@Velfi 

```
failures:

---- test_async_std_runtime_timeouts stdout ----
thread 'test_async_std_runtime_timeouts' panicked at 'actual = 506.378614ms, expected = 500ms', sdk/s3/tests/alternative-async-runtime.rs:120:5


failures:
    test_async_std_runtime_timeouts
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
```diff
@@ -68,6 +69,8 @@ interface RustCodegenDecorator<C : CoreCodegenContext> {
         currentProtocols
 
     fun transformModel(service: ServiceShape, model: Model): Model = model
+
+    fun supportsCodegenContext(clazz: Class<*>): Boolean
```
jdisanti, line 73: If you want to make this slightly simpler, you can add a `val codegenContextClass: Class<*>` after `val order`, and then you can provide a default implementation of `supportsCodegenContext` with the actual check in it. Then implementations only have to set the class value rather than implementing the function.

Should we use a stricter generic bound than `Class<*>`? It will always implement `CoreCodegenContext`, right?

david-perez: > If you want to make this slightly simpler, you can add a val codegenContextClass: Class<*> after val order, and then you can provide a default implementation of supportsCodegenContext with the actual check in it. Then implementations only have to set the class value rather than implementing the function.

That would make `val codegenContextClass: Class<*>` part of the public interface, at which point `supportsCodegenContext` is no longer needed and `fromClasspath` can just access the property.

> Should we use a stricter generic bound than Class<*>? It will always implement CoreCodegenContext, right?

Yes. But adding a stricter bound would entail making `RustCodegenDecorator` generic over another type parameter, so that we can write `U: CoreCodegenContext`, `supportsCodegenContext(clazz: Class<U>)`. Not worth it. The bound is already enforced in the signature of the callsite anyway.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
```diff
@@ -147,26 +153,29 @@ open class CombinedCodegenDecorator<C : CoreCodegenContext>(decorators: List<Rus
...
+                // parameters of inline functions. See https://kotlinlang.org/docs/generics.html#type-erasure for details.
                 .filter {
-                    try {
-                        it as RustCodegenDecorator<T>
-                        true
-                    } catch (e: ClassCastException) {
-                        false
-                    }
+                    val clazz = T::class.java
+                    it.supportsCodegenContext(clazz)
```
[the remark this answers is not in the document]
david-perez: No. I thought about it and concluded I don't know how to add one. It would entail writing a test that somehow managed to modify the classpath of the JVM it's running under. Is that possible? In any case, it would then need to assert something on the received `CombinedCodegenDecorator`, but you can't reach into the object to look at the decorators that have been combined...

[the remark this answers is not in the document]
jdisanti: Yeah, that's a fair point. I think what you can do is split the `fromClasspath` function into two halves: 1. `fromClasses` that is `internal` and has all the logic in it, and 2. `fromClasspath` which invokes the `ServiceLoader` and calls `fromClasses`. Then you can easily unit test the complex part by directly calling `fromClasses` in a unit test with a list of classes.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
jdisanti: You should be able to restrict it using `Class<out CoreCodegenContext>` without introducing another generic.

david-perez: Smart! https://github.com/awslabs/smithy-rs/pull/1592/commits/e3f498c2b204e28cf5667ee7f3fc9f0ee1a6101b

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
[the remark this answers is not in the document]
david-perez: Unfortunately public API inline functions are not allowed to use non-public-API declarations. See https://kotlinlang.org/docs/inline-functions.html#restrictions-for-public-api-inline-functions

But it's a small price to pay to make this unit testable. https://github.com/awslabs/smithy-rs/pull/1592/commits/a13783cff7540927c668f6590a3c3098dd1054fd

jdisanti (approved):

PR #1594, opened 2022-07-28
update: add 10ms margin of error to timeout duration assertion

Velfi:
The test is currently flaky. This gives it a little more space.


aws/sdk/integration-tests/s3/tests/alternative-async-runtime.rs
```diff
@@ -117,7 +116,17 @@ async fn timeout_test(sleep_impl: Arc<dyn AsyncSleep>) -> Result<(), Box<dyn std
         .unwrap_err();
 
     assert_eq!(format!("{:?}", err), "TimeoutError(RequestTimeoutError { kind: \"API call (all attempts including retries)\", duration: 500ms })");
-    assert_elapsed!(now, std::time::Duration::from_secs_f32(0.5));
```
jdisanti, line 120: We should add an overload to `assert_elapsed!` that takes a configurable margin of error. Maybe something like:
```rust
#[macro_export]
macro_rules! assert_elapsed {
    ($start:expr, $dur:expr) => {
        assert_elapsed!($start, $dur, margin_of_error_millis = 5);
    };
    ($start:expr, $dur:expr, margin_of_error_millis = $margin_of_error:expr) => {{
        let elapsed = $start.elapsed();
        // type ascription improves compiler error when wrong type is passed
        let lower: std::time::Duration = $dur;

        // Handles ms rounding
        assert!(
            elapsed >= lower && elapsed <= lower + std::time::Duration::from_millis($margin_of_error),
            "actual = {:?}, expected = {:?}",
            elapsed,
            lower
        );
    }};
}
```

Velfi: Wow, that's so cool. I didn't know you could overload macros like that!

jdisanti (approved):

PR #1595, opened 2022-07-28
Remove excess cargo test passes from additional-ci

hlbarber:
## Motivation and Context

https://github.com/awslabs/smithy-rs/pull/1550#discussion_r930987565

## Description

- Add `--exclude-all-features` to `cargo hack test --all-features` in `additional-ci`.
- Remove excess `cargo test --all-features`.

## Testing

Observe CI still has same coverage.

### DIFF
--- a/aws/rust-runtime/aws-config/additional-ci
+++ b/aws/rust-runtime/aws-config/additional-ci
@@ -17,7 +17,7 @@ cargo tree -d --edges normal --all-features
 # `aws-config` is not part of the `rust-runtime` workspace. As a result it is not run through
 # `check-rust-runtimes-and-tools` and does not get `cargo test --all-features` prior to this point. As a result we do
 # not apply `--exclude-all-features` here.
-echo "### Testing each feature in isolation"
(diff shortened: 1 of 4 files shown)


aws/rust-runtime/aws-config/additional-ci
```diff
@@ -14,11 +14,8 @@ cargo "+${RUST_NIGHTLY_VERSION:-nightly}" api-linter --all-features --config api
 echo "### Checking for duplicate dependency versions in the normal dependency graph with all features enabled"
 cargo tree -d --edges normal --all-features
 
-echo "### Testing with all features enabled"
-cargo test --all-features
```
jdisanti, line 18: This one needs to say because `aws-config` is not part of the `aws/rust-runtime` workspace, so its tests are not being run by `check-rust-runtimes-and-tools` by default. Definitely deserves a comment explaining that though so we don't accidentally remove it in the future.

jdisanti (approved):

Issue #1598, opened 2022-07-28
Smithy config customizations are being loaded in `AwsCodegenDecorator`


Velfi:
The retry config, timeout config, and sleep impl customizations are not AWS-specific and should be loaded by `RequiredCustomizations` in smithy codegen.

Velfi:
implemented in #1599

jdisanti:
Released in https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2022-08-31

PR #1599, opened 2022-07-28
fix: move smithy config customization into the smithy code generator

Velfi:
## Motivation and Context

#1598 

## Description

Previously, the config customizations that added functionality related to retry configs, timeout configs, and the async sleep impl were defined in the smithy codegen module but were being loaded in the AWS codegen module. They have now been updated to be loaded during smithy codegen. The affected classes are all defined in the `software.amazon.smithy.rust.codegen.smithy.customizations` module of smithy codegen. This change does **not** affect the generated code.

These classes have been removed:
- `RetryConfigDecorator`
- `SleepImplDecorator`
- `TimeoutConfigDecorator`

These classes have been renamed:
- `RetryConfigProviderConfig` is now `RetryConfigProviderCustomization`
- `PubUseRetryConfig` is now `PubUseRetryConfigGenerator`
- `SleepImplProviderConfig` is now `SleepImplProviderCustomization`
- `TimeoutConfigProviderConfig` is now `TimeoutConfigProviderCustomization`

## Testing

Ran existing tests

## Checklist


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -138,15 +138,14 @@ private class AwsFluentClientExtensions(types: Types) {
                     C: #{SmithyConnector}<Error = E> + Send + 'static,
                     E: Into<#{ConnectorError}>,
                 {
-                    let retry_config = conf.retry_config.as_ref().cloned().unwrap_or_default();
-                    let timeout_config = conf.timeout_config.as_ref().cloned().unwrap_or_default();
-                    let sleep_impl = conf.sleep_impl.clone();
+                    let retry_config = conf.retry_config().cloned().unwrap_or_default();
```
jdisanti, line 141: Not blocking since it was already doing this before: it seems like `timeout_config` and `retry_config` could both be `pub(crate)` and then we could avoid the clones since we already have ownership of the `Config` struct.

jdisanti (approved):

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
Velfi: We can't do that b/c we need to use the `Config` struct to build the service client. Only part of it is used to create the inner smithy client.

PR #1603, opened 2022-07-29
Improve manual config experience for SDK retries and timeouts

jdisanti:
## Motivation and Context
This PR addresses https://github.com/awslabs/aws-sdk-rust/issues/586 by:
- Removing the Rust `Default` implementation for `aws_config::RetryConfig` and renaming `RetryConfig::new` to `RetryConfig::standard`
- Adding integration tests for each of the possible retry/timeout configurations with/without an async sleep implementation
- Improving the documentation around manual retry/timeout configuration
- Adding a panic when an invalid configuration is detected

## Checklist


CHANGELOG.next.toml
```diff
@@ -225,3 +225,45 @@ message = "Support @deprecated trait for aggregate shapes"
...
+the SDK, then you are NOT affected by this change. If you use `SdkConfig::builder()` to
+configure the SDK, then you ARE affected by this change and should set the retry config
+on that builder.
+"""
+references = ["smithy-rs#1603", "aws-sdk-rust#586"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
+author = "jdisanti"
+
+[[smithy-rs]]
+message = "Clients now default to retry being disabled if no `RetryConfig` is given to the `Config`"
```
jdisanti, line 254: I'm unsure about this change, which is why I have this PR set to Draft. I think we need the SDK client behavior to diverge from the non-SDK client behavior here, since the SDK can rely on `aws-config` to get default retry config, and non-SDK clients can't.

aws/sdk/integration-tests/s3/tests/alternative-async-runtime.rs
```diff
@@ -130,6 +130,7 @@ async fn retry_test(sleep_impl: Arc<dyn AsyncSleep>) -> Result<(), Box<dyn std::
         .credentials_provider(aws_types::credentials::SharedCredentialsProvider::new(
             credentials,
         ))
+        .retry_config(RetryConfig::standard())
```
rcoh, line 133: need to figure out how this plays with `aws-smithy-client::retry::Standard.config`

jdisanti: Yeah, that's a good callout. I think we need to remove the retry policy from the client's generic args, since it really should be configured by the `RetryConfig` that feeds into `Config` rather than by the client builder. That way, the `RetryMode` in `RetryConfig` would set the policy instead.

If we want to retain the ability to customize the retry policy implementation completely for non-SDK Smithy clients, we could potentially add a `Custom` variant to `RetryMode`, but I don't think we want that for the SDK.

What are your thoughts on this?

CHANGELOG.next.toml
jdisanti: From discussion with @rcoh, we should default to retry being on for manually constructed SDK clients, and panic if the sleep impl isn't provided.

aws/sdk/integration-tests/s3/tests/alternative-async-runtime.rs
jdisanti: From discussion with @rcoh, it seems like removing the retry generic may be the best long term approach for simplifying this, but isn't currently possible due to `tower::retry::Policy` not being `dyn`-safe yet. This potential simplification should be punted to an RFC.

jdisanti:
Won't be able to fix the rest of CI until https://github.com/awslabs/smithy-rs/pull/1654 is merged, but I think this is otherwise ready for review.

rust-runtime/aws-smithy-client/src/builder.rs
```diff
@@ -308,54 +332,92 @@ mod tests {
...
-            .build();
+            .middleware(tower::layer::util::Identity::new());
+        builder.set_sleep_impl(None);
 
-        assert!(!logs_contain(TIMEOUTS_WITHOUT_SLEEP_MSG));
-        assert!(!logs_contain(RETRIES_WITHOUT_SLEEP_MSG));
-        assert!(!logs_contain(RECOMMENDATION_MSG));
+        let result = panic::catch_unwind(AssertUnwindSafe(move || {
+            let _ = builder.build();
+        }));
```
Velfi, line 353: Is this preferred to using `#[should_panic = "the panic message"]`?

CHANGELOG.next.toml
```diff
@@ -208,3 +208,68 @@ fn f1(n: aws_smithy_types::Number) {
...
+message = """
+Client creation now panics if retries or timeouts are enabled without an async sleep
+implementation set on the SDK config.
+
+If you're using the Tokio runtime and have the `rt-tokio` feature enabled (which is enabled by default),
+then you shouldn't notice this change at all.
+
+Otherwise, if using something other than Tokio as the async runtime, the `AsyncSleep` trait must be implemented,
+and that implementation given to the config builder via the `sleep_impl` method. Alternatively, retry can be
+explicitly turned off by setting `max_attempts` to 1, which will result in successful client creation without an
```
Velfi, line 246: I saw elsewhere that you had created a `RetryConfig::disabled`. Shouldn't we encourage people to use that instead?

Velfi (approved):
I had 2 minor comments but otherwise this looks good.

P.S. we can probably delete `TriState` now; I think we were only using it for sleep impls

rust-runtime/aws-smithy-client/src/builder.rs
jdisanti: I wanted to ensure the panic originated from the `build` method, which the `catch_unwind` allows me to do.

CHANGELOG.next.toml
jdisanti: Good call!

jdisanti:
It looks like `TriState` is still used by timeouts.

jdisanti:
Posted a PR to fix the examples: https://github.com/awsdocs/aws-doc-sdk-examples/pull/3526

Issue #1596, opened 2022-07-28
Running `./gradlew :aws:sdk:cargoTest` fails with error


Velfi:
Also failing:
- `./gradlew tasks`
  ```
  Execution failed for task ':tasks'.
  > Could not create task ':jacocoMerge'.
    > Project#afterEvaluate(Action) on root project 'software.amazon.smithy.rust.codegen.smithy-rs' cannot be        executed in the current context.
  ``` 
- `./gradlew aws:sdk:cargoClippy`, `./gradlew aws:sdk:cargoCheck`
  ```
  Could not determine the dependencies of task ':aws:sdk:cargoClippy'.
  > Task with path 'modifyMtime' not found in project ':aws:sdk'.
  ```

I believe the relevant code is [here](https://github.com/awslabs/smithy-rs/blob/852c2ec260428776bd3e1a2137f903580fb912dc/buildSrc/src/main/kotlin/CodegenTestCommon.kt#L49) but I'm not sure what needs to change.

david-perez:
The two problems are unrelated. #1607 fixes the second one. I don't know how to fix the `./gradlew tasks` one.

aajtodd:
OBE, this is working now (fixed somewhere along the way in hyper1 [implementation](https://github.com/smithy-lang/smithy-rs/commit/bdec1a232a5dfdba16bafd1f637c99a920a30734))

PR #1607, opened 2022-08-01
Don't depend on the `modifyMtime` task if it has not been registered

david-perez:
When registering the Cargo commands tasks. The `aws:sdk` is the only
Gradle subproject that depends on the Cargo commands tasks, but that
does not register the `modifyMtime` task.

Partially addresses #1596.


Velfi (approved):

Velfi:
Looks good to me. You're good to merge once you fix that trailing comma issue.

PR #1609, opened 2022-08-01
Allow to build and test on MacOX M1/x86  chips

crisidev:
## Motivation and Context
This PR aims to fix building and testing the Python codegenerator by addressing a couple of issues:

1) MacOS produces `dylib` instead of `so` libraries, the `Makefile` has been changed to account for it.
2) Add `.cargo/config.toml` to the example folder with MacOS specific linker configurations.
3) Update `PyApp` class to ensure it is not picked during fork. With Python 3.8, the default mechanism to create a new Process has changed from "fork" to "spawn" on MacOS and "spawn" requires pickling, which is not yet supported by PyO3.
4) Update README.


weihanglo (approved):
It works! The change looks reasonable. Thanks.

weihanglo (approved):

PR #1619, opened 2022-08-04
Add `awslabs/smithy-rs-server` as a crate owner in the `publisher` tool

david-perez:
This commit adds the SDK and server groups to smithy-rs runtime crates,
but only the SDK group on `AwsRuntime` and `AwsSdk` crates.

With this change, the `publisher` tool won't fail when publishing the
`aws-smithy-http-server` crate.

## Motivation and Context

https://github.com/awslabs/smithy-rs/runs/7663884426?check_suite_focus=true

## Testing

**I have not tested this change**. The `correct_owner` function has side-effects.


crisidev (approved):

jdisanti:
I think we need to refactor the publisher so that it places both groups on smithy-rs runtime crates, but only the SDK group on the `AwsRuntime` and `AwsSdk` crates. The [`PackageCategory`](https://github.com/awslabs/smithy-rs/blob/main/tools/smithy-rs-tool-common/src/package.rs#L26) enum can be used to tell which crate is which.

Alternatively, we can just remove the groups entirely and only give the bot access, since really, it's the only thing that actually needs access for releases.

david-perez:
> I think we need to refactor the publisher so that it places both groups on smithy-rs runtime crates, but only the SDK group on the AwsRuntime and AwsSdk crates. The [PackageCategory](https://github.com/awslabs/smithy-rs/blob/main/tools/smithy-rs-tool-common/src/package.rs#L26) enum can be used to tell which crate is which.

Ah! Much better. Didn't know we had `PackageCategory`. I've done the change in https://github.com/awslabs/smithy-rs/pull/1619/commits/b07709c977635e8d81c7fb19d21786b641a4ddb2.

tools/publisher/src/package.rs
```diff
@@ -86,6 +87,26 @@ impl Package {
...
+    pub fn expected_owners(&self) -> HashSet<String> {
+        let mut ret = HashSet::new();
+
+        // Crate ownership for SDK crates. Crates.io requires that at least one owner
+        // is an individual rather than a team, so we use the automation user for that.
+        ret.insert(String::from(RUST_SDK_CI_OWNER));
+
+        if self.category.is_sdk() {
+            ret.insert(String::from(RUST_SDK_OWNER));
+        } else if self.handle.name == "aws-smithy-http-server" {
```
jdisanti, line 101: This check really should live in the package category code since it's already missing the `aws-smithy-http-server-python` crate. We probably shouldn't add a new category since that will require changes in a lot of places, but we could add a `is_server` function (and potentially store some data on the package category to make that determination).

david-perez: Didn't realize we were already publishing `aws-smithy-http-server-python` automatically too. I've addressed this in https://github.com/awslabs/smithy-rs/pull/1619/commits/6cf564d60db39ec36832df3c67ff48012deeace4.

> This check really should live in the package category code

The problem is the determination of whether a crate should be owned by the server group is not based on package category, but on the crate name. If we put the check there we'd need to store data like the crate name, or data derived from it, on `PackageCategory`. I'm not sure that's the right modeling. It feels more intuitive to model `PackageCategory` and crate owner as orthogonal concepts, and map from the former to the latter in this `expected_owners` function.

jdisanti (approved):

PR #1620, opened 2022-08-04
RFC: Service Builder Improvements

hlbarber:
[Rendered](https://github.com/awslabs/smithy-rs/blob/harryb/builder-rfc/design/src/rfcs/rfc0020_service_builder.md)
[Toy Implementation](https://github.com/hlbarber/service-builder/pull/1)


design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,286 @@
...
+        Operation0,
+    ]
+}
+```
+
+We have purposely omitted details from the model that are unimportant to describing the proposal. We also omit distracting details from the Rust snippets.
+
+### Handlers
+
+An core concept in the current service builder is the `Handler` trait:
```
david-perez, line 46:
```suggestion
A core concept in the current service builder is the `Handler` trait:
```

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,286 @@
...
+        let extension = Extension(request.extensions().get::<T>().clone());
+        let output = self(input, extension).await;
+
+        let response = /* Create `http::Response` from `output: Operation0Output` */
+        response
+    }
+}
+```
+
+Creating `Operation0Input` from a `http::Request` and `http::Response` from a `Operation0Output` involves the HTTP binding traits and protocol aware deserialization. The structure [RuntimeError](https://github.com/awslabs/smithy-rs/blob/458eeb63b95e6e1e26de0858457adbc0b39cbe4e/rust-runtime/aws-smithy-http-server/src/runtime_error.rs#L53-L5) enumerates internal error cases such as serialization/deserialization failures, `extensions().get::<T>()` failures, etc. We omit error handling in the snippets above, but, in full, they also involve protocol aware conversions from the `RuntimeError` and `http::Response`. The reader should make note of the influence of the model/protocol on the different sections of this procedure.
```
david-perez, line 93:
```suggestion
Creating `Operation0Input` from a `http::Request` and `http::Response` from a `Operation0Output` involves the [HTTP binding traits](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html) and protocol aware deserialization. The structure [RuntimeError](https://github.com/awslabs/smithy-rs/blob/458eeb63b95e6e1e26de0858457adbc0b39cbe4e/rust-runtime/aws-smithy-http-server/src/runtime_error.rs#L53-L5) enumerates error cases such as serialization/deserialization failures, `extensions().get::<T>()` failures, etc. We omit error handling in the snippets above, but, in full, they also involve protocol aware conversions from the `RuntimeError` and `http::Response`. The reader should make note of the influence of the model/protocol on the different sections of this procedure.
```

Despite not being `pub`, `RuntimeError` is not internal in that it gets exposed to the clients over the wire.

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,286 @@
...
+
+The `request.extensions().get::<T>()` present in the `Fun: FnOnce(Operation0Input, Extension<S>) -> Fut` implementation is the current approach to injecting state into handlers. The customer is required to apply a [AddExtensionLayer](https://docs.rs/tower-http/latest/tower_http/add_extension/struct.AddExtensionLayer.html) to the output of the service builder so that, when the request reaches the handler, the `extensions().get::<T>()` will succeed.
```
david-perez, line 95: I'd call out in this paragraph that otherwise, the application panics, to better support your proposal.

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,286 @@
...
+        Poll::Ready(Ok(()))
+    }
+
+    async fn call(&mut self, req: Request<B>) -> Result<Self::Response, Self::Error> {
+        self.handler.call(req).await.map(Ok)
+    }
+}
+```
+
+The service build uses both `Handler` and `OperationHandler`
```
david-perez, line 123:
```suggestion
The service build uses both `Handler` and `OperationHandler`.
```

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,286 @@
...
+                    Ok(ok) => ok.oneshot().await,
+                    Err(err) => /* Convert routing error into http::Response */
+                }
+            }
+        }
+    }
+}
+```
+
+Along side the protocol specific constructors, `Router` includes a `layer` method. This provides a way for the customer to apply a `tower::Layer` to all routes. For every protocol, `Router::layer`, has the approximately the same behavior:
```
david-perez, line 271:
```suggestion
Along side the protocol specific constructors, `Router` includes a `layer` method. This provides a way for the customer to apply a `tower::Layer` to all routes. For every protocol, `Router::layer` has the approximately the same behavior:
```

david-perez (commented):
So far Background section looks good.

design/src/rfcs/rfc0020_service_builder.md
david-perez: My comment only applies to the second paragraph; I must have mistakenly selected both in the UI when writing it.

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,286 @@
+# RFC: Service Builder Improvements
+
+> Status: RFC
+
+One might characterize `smithy-rs` as a tool for transforming a [Smithy service](https://awslabs.github.io/smithy/1.0/spec/core/model.html#service) into a [tower::Service](https://docs.rs/tower-service/latest/tower_service/trait.Service.html) builder. A Smithy model defines semantics and constrains the behavior of the generated service, but not completely - closures must be passed to the builder to before the `tower::Service` is fully specified. This builder structure is the primary API surface we provide to the customer, as a result, it is important that it meets their needs.
```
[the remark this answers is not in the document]
hlbarber: It'd be nice to have a term which encompasses everything that implements `FnOnce`.

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,286 @@
...
+
+operation Operation1 {
+    input: Input1,
+    output: Output1
+}
+
+service Service {
+    operations: [
+        Operation0,
+        Operation0,
```
82marbag, line 37: `Operation1`

design/src/rfcs/rfc0020_service_builder.md
hlbarber: I've worded this too roughly. What I'm trying to get at is that there are protocol specific failures which occur outside of the handler which can get captured by `RuntimeError`. What I'm going to argue for later is that, perhaps if we are going to have protocol specific routers, then perhaps it makes sense to have protocol specific errors (which all enjoy something akin to a `IntoResponse` trait).

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -283,4 +280,269 @@ let new_routes = old_routes
...
+
+// Model aware conversion between `tower::Service<{Operation}Input, Response = {Operation}Output, Error = Infallible>`
+// and `tower::Service<http::Request, Response = http::Response, Error = Infallible>`
+let operation0 = make_operation0(svc);
+
+// Middleware can be applied at this point
+let operation0 = http_layer.layer(op1_svc);
+
+OperationRegistryBuilder::default()
+    .operation0(operation0)
```
david-perez: What is preventing the user from passing in an arbitrary `tower::Service` for `operation0` here? The current operation registry ensures that the provided handler takes in and returns the appropriate input and output types.

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -283,4 +280,269 @@ let new_routes = old_routes
...
+
+In `smithy-rs` a customer is only able to apply a layer to either the `aws_smithy_http::routing::Router` or every route via the [layer method](#router) described above.
+
+# Proposal
+
+The proposal is presented as a series of compatible transforms to the existing service builder, each paired with a motivation. Most of these can be independently implemented, but in the case where there exists an interdependency it is stated.
+
+Although presented as a mutation to the existing service builder, the actual implementation should exist as a entirely separate builder - reusing code generation from the old builder but exposes a new Rust API. Preserving the old API surface will prevent breakage and make it easier to perform comparative benchmarks and testing.
+
+## Remove two-step build procedure
```
david-perez, line 384: I assume this transformation will be superseded by the yet-to-be-drafted "Type erasure with the name of the Smithy service" (?) They are both independent, but since they're also both breaking changes, I'd just release one of them. If this is the case I'd note it down somewhere in the RFC.

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -283,4 +280,269 @@ let new_routes = old_routes
...
+    .nest("/teams", team_routes);
+```
+
+In `smithy-rs` a customer is only able to apply a layer to either the `aws_smithy_http::routing::Router` or every route via the [layer method](#router) described above.
+
+# Proposal
+
+The proposal is presented as a series of compatible transforms to the existing service builder, each paired with a motivation. Most of these can be independently implemented, but in the case where there exists an interdependency it is stated.
+
+Although presented as a mutation to the existing service builder, the actual implementation should exist as a entirely separate builder - reusing code generation from the old builder but exposes a new Rust API. Preserving the old API surface will prevent breakage and make it easier to perform comparative benchmarks and testing.
```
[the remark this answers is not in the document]
david-perez: Is this goal achievable for _all_ proposed transformations? E.g. How do we expose an entirely separate `OperationRegistryBuilder` without changing the type's name(space)?

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -283,4 +280,269 @@ let new_routes = old_routes
...
+
+```rust
+async fn handler(Extension(state): Extension</* State */>) -> /* Return Type */ {}
+
+let api_routes = Router::new()
+    .nest("/users", user_routes.layer(Extension(/* state */)))
+    .nest("/teams", team_routes);
+```
+
+In `smithy-rs` a customer is only able to apply a layer to either the `aws_smithy_http::routing::Router` or every route via the [layer method](#router) described above.
```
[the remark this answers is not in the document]
hlbarber: The `Router` itself is a `Service` so the customer can wrap it in middleware.

design/src/rfcs/rfc0020_service_builder.md
hlbarber: Nothing, it's possible that the customer can pass a `tower::Service<http::Request, Response = http::Response>` here which completely ignores the model. I'm unsure if this is any worse than allowing the customer to apply arbitrary `Layer`s.

Approach C has been added which is more attractive from this POV.

design/src/rfcs/rfc0020_service_builder.md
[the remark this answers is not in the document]
david-perez: But currently that's the only place one can apply a layer in smithy-rs?

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,719 @@
...
+    pub fn from_service(inner: S) -> Self {
+        Self {
+            inner,
+            layer: Identity
+        }
+    }
+}
+
+impl<F> Operation0<ServiceFn<F>, Identity> {
+    pub fn from_handler(inner: F) -> Self {
```
david-perez, line 642: Perhaps the omission is justified by the sentence above that said you're not going to deal with the whole handlers with and without state issue (or answered by the item in approach B "- `from_handler`, which takes an async `Operation0Input -> Operation0Output`."), but: in both this approach and approach B, what are the bounds on `F`? Is the idea to still require `F` implement `Handler`, which we will implement in codegen like we do now, for functions with and without state?

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,719 @@
...
+        Operation0 {
+            inner: self.inner,
+            layer: Stack::new(self.layer, /* Construct auth middleware */)
+        }
+
+    }
+}
+
+impl<Op1, Op2> OperationRegistryBuilder<Op1, Op2> {
+    pub fn operation0<S, L>(self, operation: Operation0<S, L>) -> OperationRegistryBuilder<<L as Layer<Service0Operation0<S>>::Service, Op2>
```
david-perez, line 672: I like approach C.

Perhaps I'm asking too much, but wouldn't it be nice if we could make `operation0` take in _both_ an `Operation0` struct, and a function/closure just as it does now? So `registry.operation0(fun)` is just syntax sugar for `registry.operation0(Operation0::from_handler(fun))`. Maybe by making `operation0` take a generic param implementing a trait that we auto-implement in codegen for both `Operation0` and the correct function types (with and without state)? Essentially an `UberHandler`.

design/src/rfcs/rfc0020_service_builder.md
david-perez: Actually, is this RFC going to address this extensions "type unsafety" issue? I thought you wanted to fix it but I don't see a relevant heading under the Proposal section.

design/src/rfcs/rfc0020_service_builder.md
hlbarber: I think this is probably definitely doable and desirable. We could have `operation0` take a `Into<Operation<S, L>>` and implement `From<F>` where `F: Fn(/* with or without state */)` perhaps.

One thing I'm struggling with is that there's a lot of moving parts in this section and it's hard to maintain a narrative without side stepping. I'll make sure to get this in though.

design/src/rfcs/rfc0020_service_builder.md
hlbarber: Will add a section about :+1:

design/src/rfcs/rfc0020_service_builder.md
hlbarber: On second thought I don't think it's possible. It's possible to `impl Handler` for `F: Fn(X) -> ...` and `F: Fn(X, Y) -> ...` because it's sugar for `Fn<(X, )>` and `Fn<(X, Y)>` - no struct can be both so we can't run into conflicts. This isn't the case for `F: Fn` and `S: Service<Req>` - structs could presumably be both.

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,719 @@
...
+
+impl OperationRegistryBuilder<Op0, In0, Op1, In1>
+where
+    Op0: Handler<B, In0, Operation0Input>,
+    Op1: Handler<B, In1, Operation1Input>,
+{
+    pub fn build(self) -> OperationRegistry<Op0, In0, Op1, In1> {
+        OperationRegistry {
+            operation0: self.operation0,
+            operation1: self.operation1,
```
weihanglo, line 429: I am a bit confused. Doesn't we get a type mismatch with `self.operation0` (an `Option`) and `OperationRegistry.operation0`?

Let me guess. Are you going to propose a solution in type-level state machine?

weihanglo (commented):
Thank you for putting into this!

I favour approach C as well in terms of middleware constructors.

design/src/rfcs/rfc0020_service_builder.md
hlbarber: Yep, that's exactly what the intention is here. I should mention that we get rid of `Option`s in addition to simply changing the signature of the setter methods.

design/src/rfcs/rfc0020_service_builder.md
hlbarber: On second thought I'm not sure this is possible because a `S: Service<Input, Response = Output>` could also be an async `Input -> Output`, in which case there would be conflicting conversion implementations.

design/src/rfcs/rfc0020_service_builder.md
hlbarber: Yep, we should continue with `Handler` pattern except the bounds change now. https://github.com/hlbarber/service-builder/pull/1/files#diff-67196527c762aa94fd3d640b4495ff46ab8e9a0760f426b9314b0e00b3f0846b

jdisanti (approved):

weihanglo (approved):

david-perez (approved):
Very happy that we're doing this huge improvement.

Overall the document's content is very well written.

Consider running it through `mdlint` or `prettier` before merging to aid reading and maintaining the source. Consider hard-wrapping lines too.

---

I also take it that these remaining 3 ideas we discussed are being punted to another RFC (which is for the best, since this one is long):

1. Type-safe state extraction.
2. Allowing a `build_unchecked()` on the service builder for the case where the user does not want to specify a handler for all operations (think the case of a multi-Lambda service).
3. Exploration of the `ServiceBuilderExt` idea to allow us and third-parties to vend curated stacks of middleware.

In fact, I don't think any of these warrant an RFC. Just the implementation PR, if they turn out to be possible, would be fine.

design/src/rfcs/rfc0020_service_builder.md
```diff
@@ -0,0 +1,863 @@
...
+        }
+    }
+}
+```
+
+The customer will now get a compile time error rather than a runtime error when they fail to specify a handler.
```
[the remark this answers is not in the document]
hlbarber: I think `required` would definitely be a good candidate for compile time checking.

The decision is not as as clear cut as the service builder case though:

- We already have generics on the service builder which allows us to type check.
- "missing handler" is the entire failure criteria so we eliminate fallibility completely.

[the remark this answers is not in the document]
david-perez: Thinking more about it, I think it's not worth introducing API asymmetry for `required` with respect to the other constraint traits. Maybe it'd be weird to have one constraint trait checked at compile time and the others at runtime. And modifying the deserializers to make them aware of when they've set a value would be a hassle.

hlbarber:
> I also take it that these remaining 3 ideas we discussed are being punted to another RFC (which is for the best, since this one is long):
> 
> 1. Type-safe state extraction.
> 2. Allowing a `build_unchecked()` on the service builder for the case where the user does not want to specify a handler for all operations (think the case of a multi-Lambda service).
> 3. Exploration of the `ServiceBuilderExt` idea to allow us and third-parties to vend curated stacks of middleware.
> 
> In fact, I don't think any of these warrant an RFC. Just the implementation PR, if they turn out to be possible, would be fine.

Just to add some extra context/thoughts to these:

1. We now have a great case study on how to do this via https://github.com/tokio-rs/axum/pull/1155. I think it's best that we complete the RFC implementation "extractor-less" and then put up a PR to add the type extractor machinery separately. The implementation seems straight forward, but there's a question of whether it's worth adding an extra set of generics for a marginal(?) improvement to ergonomics.
2. I'm confident we can get something like this added.
3. Similar to (1). In this case we might be brushing against the lack of GATs and other nightly features.

PR #1621, opened 2022-08-08
Add codegen version to generated package

unexge:
Closes https://github.com/awslabs/smithy-rs/issues/1612

## Description

This PR adds smithy-rs's version to generated `Cargo.toml`s. This information later can be fetched via `cargo metadata` (or [cargo_metadata](https://crates.io/crates/cargo_metadata) crate).

## Testing
It can be tested by generating any package, for example:
```bash
$ cd rust-runtime/aws-smithy-http-server/examples
$ make # generates both client and server for Pokemon service

$ rg -B1 -u codegen-version # version can be found in generated `Cargo.toml`s
pokemon-service-client/Cargo.toml
9-[package.metadata.smithy]
10:codegen-version = "0.47.0-0198d26096eb1af510ce24766c921ffc5e4c191e"

pokemon-service-server-sdk/Cargo.toml
9-[package.metadata.smithy]
10:codegen-version = "0.47.0-0198d26096eb1af510ce24766c921ffc5e4c191e"

# it can also be fetched via `cargo metadata`
$ cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "pokemon-service-server-sdk") | .metadata.smithy."codegen-version"'
0.47.0-0198d26096eb1af510ce24766c921ffc5e4c191e
```

## Checklist

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/Version.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/Version.kt
@@ -7,22 +7,17 @@ package software.amazon.smithy.rust.codegen.smithy
 
 import software.amazon.smithy.codegen.core.CodegenException
 
-class Version(private val content: String) {
-    fun fullVersion(): String = content
+// generated as part of the build in the "{smithy_rs_version}\n{git_commit_hash}" format,
+// see codegen/build.gradle.kts
+private const val VERSION_FILENAME = "runtime-crate-version.txt"
 
-    fun crateVersion(): String {
-        val parts = content.split("-")
-        if (parts.size < 2) {
-            return ""
-        }
-        return parts.first()
-    }
+class Version(private val content: String) {
+    // Returns full version in the "{smithy_rs_version}-{git_commit_hash}" format
+    fun fullVersion(): String = content.lines().joinToString("-")
+
+    fun crateVersion(): String = content.lines().first()
 
     companion object {
-        // generated as part of the build in the "{smithy_rs_version}-{git_commit_hash}" format,
-        // see codegen/build.gradle.kts
-        private const val VERSION_FILENAME = "runtime-crate-version.txt"
-
         fun fullVersion(): String =
             fromDefaultResource().fullVersion()
 
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/Version.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/Version.kt
@@ -11,7 +11,7 @@ class Version {
     companion object {
         // generated as part of the build in the "{smithy_rs_version}-{git_commit_hash}" format,
         // see codegen/build.gradle.kts
-        private const val VERSION_FILENAME = "smithy-version.txt"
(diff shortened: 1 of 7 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/CargoTomlGenerator.kt
```diff
@@ -74,3 +80,6 @@ class CargoTomlGenerator(
         writer.writeWithNoFormatting(TomlWriter().write(cargoToml))
     }
 }
+
+fun smithyCodegenVersion(): String =
+    try { Version.crateVersion() } catch (ex: Exception) { "unknown" }
```
82marbag, line 85: Not needed because it throws in `defaultRuntimeCrateVersion`

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/Version.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class Version(private val content: String) {
+    fun fullVersion(): String = content
+
+    fun crateVersion(): String {
+        val parts = content.split("-")
+        if (parts.size < 2) {
+            return ""
+        }
+        return parts.first()
```
82marbag, line 18: If you want to go for "the file contains one line", this may not always work; some versions don't follow the format `vx.y.z`, like `release-2022-08-04`. In this case you should remove the last element, instead of picking the first one

82marbag (approved):

codegen/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/VersionTest.kt
```diff
@@ -0,0 +1,65 @@
...
+            Arguments.of(
+                "0.0.0",
+                "0.0.0",
+                "0.0.0",
+            ),
+            Arguments.of(
+                "",
+                "",
+                "",
+            ),
```
jdisanti, line 62: Are these valid use-cases? It seems like it should fail in these cases.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/CargoTomlGenerator.kt
```diff
@@ -63,6 +64,11 @@ class CargoTomlGenerator(
                 "edition" to "2021",
                 "license" to settings.license,
                 "repository" to settings.moduleRepository,
+                "metadata" to listOfNotNull(
+                    "smithy" to listOfNotNull(
+                        "codegen-version" to smithyCodegenVersion(),
```
jdisanti, line 69: Nit: You can call `Version.fullVersion()` directly rather than wrapping it in a function.

jdisanti (approved):
Couple comments you may want to address, but nothing blocking from the client side.

codegen/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/VersionTest.kt
unexge: Yeah good point, I've updated test cases and the implementation.

PR #1623, opened 2022-08-08
Update Smithy version dependency.

DavidOgunsAWS:
## Motivation and Context

* New major version of Smithy supporting IDLv2. AWS SDK code generator will soon need to support models in this version.

## Description

* Update the version dependency and a generated protocol test.

## Testing

* Built upcoming version of [Smithy](https://github.com/awslabs/smithy) from IDL2 branch locally to VERSION `1.23.0` using `./gradlew clean build publishToMavenLocal smithy-cli:runtime`
* Made the changes in this PR to update the version of Smithy referenced in dependencies and ran: `./gradlew --refresh-dependencies clean codegen-test:test aws:sdk:assemble`

To success.
## Checklist


hlbarber (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerHttpSensitivityGenerator.kt
```diff
@@ -132,7 +132,9 @@ class ServerHttpSensitivityGenerator(
                     it.direction == RelationshipDirection.DIRECTED
                 }
                 .filter {
-                    it.hasTrait<SensitiveTrait>()
+                    it.hasTrait<SensitiveTrait>() ||
```
hlbarber, line 135: NIT: Perhaps this can be pulled out into it's own function.

david-perez:
@DavidOgunsAWS You just have to remove those two tests from the list of tests that we expect to fail, and CI should pass. The tests were fixed upstream by @82marbag in https://github.com/awslabs/smithy/pull/1319, so with this PR, they now pass. I noted this in https://github.com/awslabs/smithy-rs/pull/1590#pullrequestreview-1054482811.

CHANGELOG.next.toml
```diff
@@ -47,8 +48,39 @@ references = ["smithy-rs#1598"]
 meta = { "breaking" = true, "tada" = true, "bug" = false, "target" = "client" }
 author = "Velfi"
 
+[[smithy-rs]]
+message = "Update Smithy dependency to 1.23.0. Models using version 2.0 of the IDL are now supported."
+references = ["smithy-rs#1623"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
```
jdisanti, line 54:
```suggestion
meta = { "breaking" = false, "tada" = true, "bug" = false }
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/EventStreamTestTools.kt
```diff
@@ -98,7 +98,7 @@ private fun fillInBaseModel(
         MessageWithNoHeaderPayloadTraits: MessageWithNoHeaderPayloadTraits,
         SomeError: SomeError,
     }
-    structure TestStreamInputOutput { @required value: TestStream }
+    structure TestStreamInputOutput { @httpPayload @required value: TestStream }
```
jdisanti, line 101: What was this change for?

codegen-test/model/rest-xml-extras.smithy
```diff
@@ -61,7 +61,7 @@ structure CreateFooInput {}
         </PrimitiveIntDocument>
        """,
        bodyMediaType: "application/xml",
-       params: { value: 1 }
+       params: { value: 1, requiredValue: 0 }
```
jdisanti, line 64: The test description is: "Primitive ints should not be serialized when they are unset", so adding `requiredValue` to the params seems to break what is being tested. Are these tests still valid with Smithy IDL 2?

jdisanti (commented):
Mostly looks good to me. Just a couple questions.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/EventStreamTestTools.kt
DavidOgunsAWS: [This](https://github.com/awslabs/smithy/pull/1076) and possibly another breakfix for model handling. We needed to make the same fix for Kotlin code generation as well. Only http protocol test instantiations of this model require the trait, but the non-http protocol test instantiations ignore it so it's safe to add like this.

codegen-test/model/rest-xml-extras.smithy
DavidOgunsAWS: [Primitives](https://github.com/awslabs/smithy/pull/1286) are now just the type with a default value backing it.

And default values are [always serialized.](https://awslabs.github.io/smithy/2.0/spec/type-refinement-traits.html#default-value-serialization)

jdisanti: I think the model for this test needs to be revised because it seems like the test isn't testing what it intended to test anymore. It seems like the [`requiredValue`](https://github.com/awslabs/smithy-rs/blob/2be10ed9dd9348336b239ecd5b1467ad221d36c0/codegen-test/model/rest-xml-extras.smithy#L75) should get a default set on it rather than providing a value for it in the test case.

jdisanti (approved):

codegen-test/model/rest-xml-extras.smithy
DavidOgunsAWS: The change I made was on the basis of understanding that old behavior around primitives and `@required` were kind of a wild west -- apparently not just for nullability, but also serialization. `@default` in 2.0 is where the behavior starts to be defined and predicted so I upgraded the test model to 2.0 and used that trait instead.

I don't believe there is anything further to be tested.

Issue #1562, opened 2022-07-21
Client response streaming minimum speed limit/timeout


jdisanti:
Clients should have a way of timing out if the response stream started but then slowed significantly or stopped entirely without closing the connection part way through. For example, if a call to S3 GetObject was made, and some bytes of the object were received in the streaming response body, but the server stopped responding with more for a significant period of time, that should result in a timeout.

Internal tracking: P66130286

jdisanti:
@Velfi wrote a prototype for this in https://github.com/awslabs/smithy-rs/pull/1627

PR #1627, opened 2022-08-10
add: naive implementation of a minimum-throughput body with tests

Velfi:
## Motivation and Context

#1562 

## Description

Exactly what it says on the tin

## Testing

tests are included

## Checklist

### DIFF
--- a/aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
+++ b/aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
@@ -1,0 +1,68 @@
+use aws_smithy_async::future::fn_stream::FnStream;
+use aws_smithy_async::rt::sleep::AsyncSleep;
+use aws_smithy_http::body::SdkBody;
+use aws_smithy_http::byte_stream::ByteStream;
+use aws_smithy_http::minimum_throughput::MinimumThroughputBody;
+use std::convert::Infallible;
+use std::time::Duration;
+
+// This test is flaky. The error message will end with something like "0.999 B/s was observed"
+#[should_panic = "minimum throughput was specified at 2 B/s, but throughput of"]
+#[tokio::test]
+async fn test_throughput_timeout_happens_for_slow_stream() {
+    tracing_subscriber::fmt::init();
+    tracing::info!("tracing is working");
+    // let test_start_time = std::time::Instant::now();
+    // Have to return results b/c `hyper::body::Body::wrap_stream` expects them
+    let stream: FnStream<Result<String, Infallible>, _> = FnStream::new(|tx| {
+        let async_sleep = aws_smithy_async::rt::sleep::TokioSleep::new();
+        Box::pin(async move {
+            for i in 0..10 {
+                // Will send slightly less that 1 byte per second because ASCII digits have a size
+                // of 1 byte and we sleep for 1 second after every digit we send.
+                tx.send(Ok(format!("{}", i))).await.expect("failed to send");
+                async_sleep.sleep(Duration::from_secs(1)).await;
+            }
+        })
+    });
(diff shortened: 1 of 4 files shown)


aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
```diff
@@ -0,0 +1,68 @@
...
+    // let test_start_time = std::time::Instant::now();
+    // Have to return results b/c `hyper::body::Body::wrap_stream` expects them
+    let stream: FnStream<Result<String, Infallible>, _> = FnStream::new(|tx| {
+        let async_sleep = aws_smithy_async::rt::sleep::TokioSleep::new();
+        Box::pin(async move {
+            for i in 0..10 {
+                // Will send slightly less that 1 byte per second because ASCII digits have a size
+                // of 1 byte and we sleep for 1 second after every digit we send.
+                tx.send(Ok(format!("{}", i))).await.expect("failed to send");
+                async_sleep.sleep(Duration::from_secs(1)).await;
```
jdisanti, line 24: Ideally, we should be able to fake time keeping in the `MinimumThroughputBody` so that we don't need to sleep in tests. Essentially, be able to replace the source of `Instant::now()` inside of `MinimumThroughputBody`.

rust-runtime/aws-smithy-http/src/minimum_throughput.rs
```diff
@@ -0,0 +1,123 @@
...
+            let mut total_bytes_read = *first_bytes;
+
+            while let Some((_, bytes_read)) = logs.next() {
+                total_bytes_read += bytes_read;
+            }
+
+            let minimum_bytes_per_second =
+                this.minimum_throughput.0 as f64 / this.minimum_throughput.1.as_secs_f64();
+            let actual_bytes_per_second =
+                total_bytes_read as f64 / time_elapsed_since_first_poll.as_secs_f64();
```
jdisanti, line 63: I think to be accurate for longer downloads, and to avoid consuming excess memory, you'll need to have a sliding window over the `throughput_logs`. It's probably as simple as popping the front of the vec until the first entry is no longer older than the configured duration (before doing this calculation).

Additionally, we probably shouldn't do this comparison at all unless the configured duration has elapsed already. Otherwise, there's a chance that a slow-to-start download would always result in failure since it would only have one data point.

aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
```diff
@@ -0,0 +1,68 @@
...
+    });
+    let res = body.collect().await;
+
+    match res {
+        Ok(_) => panic!("Expected an error due to slow stream but no error occurred."),
+        Err(e) => panic!("{}", e),
+    }
+}
+
+#[tokio::test]
```
jdisanti, line 44: I'm feeling mathy today, bear with me...

I think a good test to add may be a shrinking sine wave. Something like [`(abs(sin(x))+0.5) * max(0, (-0.05x+1))`](https://www.wolframalpha.com/input?i=%28abs%28sin%28x%29%29%2B0.5%29*max%280%2C+%28-0.05x%2B1%29%29), which may require a little bit of tweaking.

With that function, it should start off happy despite some significant drop offs in speed periodically, but eventually (after 20 seconds or so) get to a point where it needs to timeout.

Another good test is a straight line going up starting at 1, something like [`x+1`](https://www.wolframalpha.com/input?i=x%2B1), just to make sure it doesn't break on slow starts.

And then, another good one would be a full minute of good throughput, then 0.5 seconds of zero throughput, followed by good throughput again.

jdisanti (commented):
Great work! This is a good first step! I don't think there's a ton of work to make this mergeable. The tricky part will be to figure out how this should be configured, and what good defaults are.

jdisanti:
Closing this until we pick up #1562 again.

PR #1634, opened 2022-08-15
Add TLS support to Pokemon server

unexge:
## Motivation and Context
This PR adds TLS support to Pokemon server. We can refer Pokemon server from our docs to show our users how to configure TLS for their own services. It also addresses https://github.com/awslabs/smithy-rs/issues/1542 because without TLS and necessary configuration, by default clients don't negotiate their way up to http2.

## Description
This PR adds TLS support to Pokemon server using [tls-listener](https://crates.io/crates/tls-listener) and [tokio-rustls](https://crates.io/crates/tokio-rustls) (and [rustls-pemfile](https://crates.io/crates/rustls-pemfile) to parse PEM files). It can be enabled via `--use-tls` and can be configured with `--tls-cert-path` and `--tls-key-path` flags.

## Testing
It can be tested by creating test certificates and making requests via cURL to make sure it works as expected.

```bash
$ cd rust-runtime/aws-smithy-http-server/examples/pokemon-service
# to trust local CA, you can also skip this step but then you need to use `-k|--insecure` with cURL because certificate will be unknown
$ mkcert --install
# Create test certificates for localhost, it will create "./localhost.pem" and "./localhost-key.pem" in same directory
$ mkcert localhost 
# Run Pokemon service with TLS enabled using generated certificates
$ cargo run -- --use-tls --tls-cert-path ./localhost.pem --tls-key-path ./localhost-key.pem
```

```bash
# In different terminal make request to Pokemon service with https
$ curl -D- -H "Accept: application/json" https://localhost:13734/pokemon-species/pikachu
HTTP/2 200
content-type: application/json
content-length: 573
date: Mon, 15 Aug 2022 11:38:54 GMT

{"flavorTextEntries":[{"flavorText":"When several of these Pokémon gather, their electricity could build and cause lightning storms.","language":"en"},{"flavorText":"Cuando varios de estos Pokémon se juntan, su energía puede causar fuertes tormentas.","language":"es"},{"flavorText":"Quando vari Pokémon di questo tipo si radunano, la loro energia può causare forti tempeste.","language":"it"},{"flavorText":"ほっぺたの りょうがわに ちいさい でんきぶくろを もつ。ピンチのときに ほうでんする。","language":"jp"}],"name":"pikachu"}
```


weihanglo (commented):
I find the PR really great! However, in my macbook if there is a Unknown CA or corrupted TLS data, the server just goes down. Do we need to handle this?

weihanglo:
BTW, `curl -k` can bypass the unknown CA issue.

unexge:
Good catch! I was using `mkcert -install` to have trusted local CA and missed that error. I'm not sure how to handle unknown CAs or corrupted TLS data but the server shouldn't stop at least

weihanglo (approved):
LGTM now. Thanks for fixing the crashing issue 👍🏾

david-perez (changes requested):
Content looks good. However, I'd prefer it if we put this TLS example in a separate executable [using `[[bin]]`](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#configuring-a-target). I expect most service owners to use a separate process (possibly in another host) to do TLS termination. This way we keep the main example for most users copy-pasteable. That executable could be documented with your investigation regarding h2 negotiation you've written in the PR description, which is also valuable.

Bonus points: it'd be nice if the new executable could have integration tests with the Pokémon Smithy client using HTTPS, perhaps asserting that HTTP2 is used too?

rust-runtime/aws-smithy-http-server/examples/pokemon-service/Cargo.toml
```diff
@@ -16,6 +21,8 @@ tower = "0.4"
 tower-http = { version = "0.3", features = ["trace"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+
+# This dependencies only required for `pokemon-service-tls`
```
david-perez, line 25:
```suggestion
# These dependencies are only required for `pokemon-service-tls`.
```

rust-runtime/aws-smithy-http-server/examples/pokemon-service/Cargo.toml
```diff
@@ -31,6 +38,9 @@ home = "0.5"
 serial_test = "0.7.0"
 wrk-api-bench = "0.0.7"
 
+# This dependencies only required for testing `pokemon-service-tls`
```
david-perez, line 41:
```suggestion
# These dependencies are only required for testing `pokemon-service-tls`.
```

rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/bin/pokemon-service-tls.rs
```diff
@@ -0,0 +1,152 @@
...
+// then you can run TLS server via:
+// `$ cargo run --bin pokemon-service-tls -- --tls-cert-path ./localhost.pem --tls-key-path ./localhost-key.pem`
+// and test it:
+// ```bash
+// $ curl -k -D- -H "Accept: application/json" https://localhost:13734/pokemon-species/pikachu
+// HTTP/2 200
+// # ...
+// ```
+// note that by default created certificates will be unknown and you should use `-k|--insecure`
+// flag while making requests with cURL or you can run `$ mkcert -install` to trust certificates created by `mkcert`.
```
david-perez, line 23:
```suggestion
// flag while making requests with cURL or you can run `mkcert -install` to trust certificates created by `mkcert`.
```

rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/bin/pokemon-service-tls.rs
```diff
@@ -0,0 +1,152 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// This program is exported as a binary named `pokemon-service-tls`.
+// It uses `tls-listener`, `tokio-rustls` (and `rustls-pemfile` to parse PEM files)
+// to serve TLS connections. It also enables h2 ALPN protocol,
+// without this clients by default don't upgrade to http2.
+//
+// You can use `mkcert` to create certificates for testing:
```
david-perez, line 11: I'd link somewhere to https://github.com/FiloSottile/mkcert.

rust-runtime/aws-smithy-http-server/examples/pokemon-service/tests/helpers.rs
```diff
@@ -33,11 +79,50 @@ pub fn client() -> Client<
...
+                .with_no_client_auth(),
+        )
+        .https_only()
+        .enable_http2()
+        .build();
+
+    let base_url = PokemonServiceVariant::Https.base_url();
+    let raw_client = Builder::new()
+        .connector(DynConnector::new(Adapter::builder().build(connector)))
+        .middleware_fn(move |mut req| {
```
david-perez, line 123: This closure could now be extracted to its own function, since both the HTTP client and HTTPs client builders use it.

rust-runtime/aws-smithy-http-server/examples/pokemon-service/tests/helpers.rs
```diff
@@ -33,15 +80,53 @@ pub fn client() -> Client<
...
 }
+
+fn rewrite_base_url(base_url: String) -> impl Fn(Request) -> Request + Clone {
+    move |mut req| {
+        let http_req = req.http_mut();
+        let uri = format!("{base_url}{}", http_req.uri().path());
+        *http_req.uri_mut() = uri.parse().unwrap();
+        req
+    }
+}
```
david-perez, line 132:
```suggestion
fn rewrite_base_url(base_url: String) -> Request {
    let http_req = req.http_mut();
    let uri = format!("{base_url}{}", http_req.uri().path());
    *http_req.uri_mut() = uri.parse().unwrap();
    req
}
```

This should work too.

david-perez (approved):

PR #1635, opened 2022-08-15
Support granular control of specifying runtime crate versions

weihanglo:
## Motivation and Context

Related: https://github.com/awslabs/smithy-rs/issues/1416

Support a more granular control on specifying runtime crate versions in smithy-build.json runtime config.

The current behaviour is not ideal. You need to specify `{ version: "DEFAULT" }` to get the default version. You can only specify the same version for all runtime crates, which is somehow inconvenient.

## Description

This PR mainly deals with two things:

- Make default default. You don't need to specify anything to become default.
- Give users the ability to specify different versions for different runtime crates.

**BREAKING**: after this PR, in `smity-build.json` the path
`rust-codegen.runtimeConfig.version` no longer exists. Instead, a new
field `versions` comes in. It's an objct mapping a runtime crate name
to a version string. There is also a special key `DEFAULT`, which is
presetted as detected runtime version but open to override. Crates
without version specified would be set as the same version as which
associated to key `DEFAULT`.

An example `smithy-build.json` might look like:

```json
{
  "version": "1.0",
  "projections": {
    "simple": {
      "plugins": {
        "rust-codegen": {
          "runtimeConfig": {
            "versions": {
               "DEFAULT": "0.48", // Set your own default version.
                                  // Leave blank for codegen auto-detecting a version.
               "aws-smithy-http": "0.47", // Can specify a different version.
               "aws-smithy-types": "0.46",
               "aws-smithy-json": "0.45",
            }
          },
          "service": ...
        }
      }
    }
  }
}
```

## Testing

Tweak the value in [`buildSrc/src/main/kotlin/CodegenTestCommon.kt`](https://github.com/awslabs/smithy-rs/blob/374a1c5df7314b3192e16d5acbc01b74a260dd80/buildSrc/src/main/kotlin/CodegenTestCommon.kt#L25-L27) and see if your runtime crate version changed.

## Checklist


weihanglo:
Open question: 

- Shall we support more customizations such like setting relative path for each runtime crate?
- Shall we be compatible with the old `version` field without breaking everyone's build script?

CHANGELOG.next.toml
```diff
@@ -16,3 +16,9 @@ message = "Add codegen version to generated package metadata"
 references = ["smithy-rs#1612"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "unexge"
+
+[[smithy-rs]]
+message = "Support granular control of specifying runtime crate versions"
+references = ["smithy-rs#1416"]
+meta = { "breaking" = true, "tada" = true, "bug" = false }
```
jdisanti, line 23: For breaking changes, add instructions to the message on what changes need to be made when upgrading. You can switch the message to triple quotes if it needs to be multi-line.

CHANGELOG.next.toml
```diff
@@ -16,3 +16,9 @@ message = "Add codegen version to generated package metadata"
 references = ["smithy-rs#1612"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "unexge"
+
+[[smithy-rs]]
+message = "Support granular control of specifying runtime crate versions"
+references = ["smithy-rs#1416"]
```
jdisanti, line 22:
```suggestion
references = ["smithy-rs#1635", "smithy-rs#1416"]
```

CHANGELOG.next.toml
weihanglo: Updated. Please review again. Thanks!

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -23,26 +24,33 @@ import software.amazon.smithy.rust.codegen.rustlang.asType
...
- * This can be configured via the `runtimeConfig.version` field in smithy-build.json
+ * This can be configured via the `runtimeConfig.versions` field in smithy-build.json
  */
-data class RuntimeCrateLocation(val path: String?, val version: String?) {
+data class RuntimeCrateLocation(val path: String?, val versions: CrateVersionMap) {
     init {
-        check(path != null || version != null) {
-            "path ($path) or version ($version) must not be null"
+        check(path != null || versions.map.isNotEmpty()) {
+            "path ($path) or versions ($versions) must not be null or empty"
```
david-perez, line 37:
```suggestion
            "path ($path) must not be null or or versions ($versions) must not be empty"
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -23,26 +24,33 @@ import software.amazon.smithy.rust.codegen.rustlang.asType
...
 
-fun RuntimeCrateLocation.crateLocation(): DependencyLocation = when (this.path) {
-    null -> CratesIo(this.version!!)
-    else -> Local(this.path, this.version)
+fun RuntimeCrateLocation.crateLocation(crateName: String?): DependencyLocation {
+    val version = crateName.let { versions.map[crateName] } ?: versions.map[DEFAULT_KEY]
+    return when (this.path) {
+        // CratesIo needs an exact version. However, for local crate we do not
+        // provide a detected version unless user explicitly sets via `DEFAULT` key.
+        null -> CratesIo(version ?: defaultRuntimeCrateVersion())
```
david-perez, line 51: I don't think this has been documented in the changelog (i.e., that when you don't specify anything in `versions`, then the latest smithy-rs version number is used for all crates).

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -23,26 +24,33 @@ import software.amazon.smithy.rust.codegen.rustlang.asType
...
 }
 
-fun RuntimeCrateLocation.crateLocation(): DependencyLocation = when (this.path) {
-    null -> CratesIo(this.version!!)
-    else -> Local(this.path, this.version)
+fun RuntimeCrateLocation.crateLocation(crateName: String?): DependencyLocation {
+    val version = crateName.let { versions.map[crateName] } ?: versions.map[DEFAULT_KEY]
+    return when (this.path) {
+        // CratesIo needs an exact version. However, for local crate we do not
+        // provide a detected version unless user explicitly sets via `DEFAULT` key.
```
david-perez, line 50:
```suggestion
        // CratesIo needs an exact version. However, for runtime crates we do not
        // provide a detected version unless the user explicitly sets one via the `versions` map.
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -53,6 +61,14 @@ fun defaultRuntimeCrateVersion(): String {
     }
 }
 
+/**
+ * A mapping for crate name to user specified version.
+ */
+@JvmInline
+value class CrateVersionMap(
```
david-perez, line 68: TIL about [`value class`](https://kotlinlang.org/docs/inline-classes.html). Very interesting.

This is the first time we introduce them in the codebase. I wonder why we're not using them? They seem to be better at encoding "transparent" newtypes, like here, something for which we rely on `data class`.

cc someone who actually knows Kotlin @rcoh.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -53,6 +61,14 @@ fun defaultRuntimeCrateVersion(): String {
     }
 }
 
+/**
+ * A mapping for crate name to user specified version.
```
david-perez, line 65:
```suggestion
 * A mapping from crate name to a user-specified version.
```

CHANGELOG.next.toml
```diff
@@ -16,3 +16,47 @@ message = "Add codegen version to generated package metadata"
...
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "unexge"
+
+[[smithy-rs]]
+message = """
+Support granular control of specifying runtime crate versions.
+
+For code generation, the field `runtimeConfig.version` in smithy-build.json has been removed.
+The new field `runtimeConfig.verisions` is an object that keys are runtime crate names (e.g. aws-smithy-http),
+and values are their versions user specifying.
```
david-perez, line 26:
```suggestion
The new field `runtimeConfig.versions` is an object whose keys are runtime crate names (e.g. `aws-smithy-http`),
and values are user-specified versions.
```

CHANGELOG.next.toml
```diff
@@ -16,3 +16,47 @@ message = "Add codegen version to generated package metadata"
...
+
+If you previous set `version = "DEFAULT"`, then you're lucky.
+The new behaviour automatically detects and injects the version for you by checking the version of the code generator.
+You don't need to add any new configuration.
```
david-perez, line 30: Is this true? I don't recognize it being implemented.

CHANGELOG.next.toml
```diff
@@ -16,3 +16,47 @@ message = "Add codegen version to generated package metadata"
...
+
+If you specified a certain version such as `version = "0.47.0", you can migrate to a special reserved key `DEFAULT`.
```
david-perez, line 32: But they couldn't specify an explicit version before, right?

CHANGELOG.next.toml
```diff
@@ -16,3 +16,47 @@ message = "Add codegen version to generated package metadata"
...
+{
+  "runtimeConfig": {
+      "versions": {
+          "DEFAULT": "0.47.0"
+      }
+  }
+}
+```
+
+Then all runtime crate are set with version 0.47.0 by default unless specified versions by crate name. For example,
```
david-perez, line 45:
```suggestion
Then all runtime crates are set with version 0.47.0 by default unless overridden by specific crates. For example,
```

david-perez (changes requested):
I think the implementation is correct, but it would help if there'd be unit tests for these main scenarios (correct me if I'm wrong if this is not what's meant to happen):

1. If user specifies `relativePath` in `runtimeConfig`, then that always takes precedence over `versions`.
1. User does not specify the `versions` object. The version number of the code-generator should be used as the version for all runtime crates.
1. User specifies `versions` key but leaves the object empty. The version number of the code-generator should be used as the version for all runtime crates.
1. User specifies `versions` object, setting `DEFAULT`. That version is used for all runtime crates.
1. User specifies `versions` object, setting explicit version numbers for some runtime crates. Then the rest of the runtime crates use the code-generator's version as their version.
1. User specifies `versions` object, setting `DEFAULT` _and_ setting version numbers for some runtime crates. Then the specified version in `DEFAULT` is used for all runtime crates, except for those where the user specified a value for in the map.
1. If the user specifies a version in the `versions` map for a crate whose name does not correspond to a runtime crate, we fail.

I think all unit tests can start with `RuntimeConfig.fromNode()`, and then you assert that the correct `RuntimeConfig` object was parsed.

david-perez:
I'm thinking that it would also be cool if we allowed the user to specify concrete versions for third-party crates e.g. the `http` crate. That would allow them to mitigate a CVE-like situation by rolling back that crate version without us having to publish a new version of smithy-rs.

We can probably tackle it in another PR.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
rcoh: no reason not to use them, TIL about them TBH, they seem handy!

We _often_ end up with more than one field which is why `data class` is usually useful but for single-member types this seems great!

rcoh:
> I'm thinking that it would also be cool if we allowed the user to specify concrete versions for third-party crates e.g. the http crate. That would allow them to mitigate a CVE-like situation by rolling back that crate version without us having to publish a new version of smithy-rs.

@jdisanti and I discussed this recently—I think it would be a great idea and it would also allow us to unify dependency versions across the entire codebase

CHANGELOG.next.toml
weihanglo: I think they could

https://github.com/awslabs/smithy-rs/blob/0f47f69ddf76a41515ac2c198391cf080cd92b30/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt#L70-L80

CHANGELOG.next.toml
weihanglo: Hmm… I can't remember what I was writing about. Anyway, fixed with b31e0b4f

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
weihanglo: Fixed with b31e0b4

weihanglo:
Tests added. Thanks @david-perez for the suggestion.

> 7. If the user specifies a version in the `versions` map for a crate whose name does not correspond to a runtime crate, we fail.

This has yet been implemented. Shall we add the check?

CHANGELOG.next.toml
david-perez: Ah, right. They could specify a concrete version, that's the `configuredVersion` branch.

codegen/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypesTest.kt
```diff
@@ -0,0 +1,172 @@
...
+        @JvmStatic
+        fun runtimeConfigProvider() = listOf(
+            Arguments.of(
+                """{}""",
+                RuntimeCrateLocation(null, CrateVersionMap(mapOf())),
+            ),
+            Arguments.of(
+                """{
+                "relativePath": "/path"
+                }""",
```
david-perez, line 56:
```suggestion
                """
                {
                    "relativePath": "/path"
                }
                """,
```

nit: if a `"""` string is multiline, I think the style we're following is to put the `"""` in its own line.

david-perez:
> > 7. If the user specifies a version in the `versions` map for a crate whose name does not correspond to a runtime crate, we fail.

> This has yet been implemented. Shall we add the check?

We can leave it if you want. I was thinking that it would be a UX improvement (think a frustrated user who misspelled a crate name in the `versions` map and wondering why they're not seeing the override in the code generated crate).

But delivering this feature would entail us having to maintain a list of the crates that we're publishing and keeping it up to date when there are additions or removals. Unless we automate this check in CI, we're likely going to forget.

weihanglo:
All review commented addressed. Thank you!

> > > 7. If the user specifies a version in the `versions` map for a crate whose name does not correspond to a runtime crate, we fail.
> 
> > This has yet been implemented. Shall we add the check?
> 
> We can leave it if you want. I was thinking that it would be a UX improvement (think a frustrated user who misspelled a crate name in the `versions` map and wondering why they're not seeing the override in the code generated crate).
> 
> But delivering this feature would entail us having to maintain a list of the crates that we're publishing and keeping it up to date when there are additions or removals. Unless we automate this check in CI, we're likely going to forget.

That's exactly what I worry about. We could have a separate PR to do that if we want.

jdisanti (approved):

PR #1636, opened 2022-08-16
Override codegen version commit hash in PR diffs

unexge:
## Motivation and Context
After https://github.com/awslabs/smithy-rs/pull/1621 we started inserting codegen versions to generated packages. Those versions also include git commit hash and that causes PR Bot to generate diffs for every PR even when there are no changes in generated packages ([example](https://d2luzm2xt3nokh.cloudfront.net/codegen-diff/fd98756160ebbbc2f964c804a80587d0a81f99c3/84a8d19621dfebd98e4aaa5fb9840146c20f6f1b/diff-aws-sdk.html)). In order to prevent that we want to have a fixed commit hash while generating diffs for generated packages.

## Description
This PR allows overriding commit hash used in codegen versions via `SMITHY_RS_VERSION_COMMIT_HASH_OVERRIDE` env variable. We provide that value in PR Bot to fix commit hash used in PR Bot diffs.

## Testing
You can provide that env variable while generating packages to see whether codegen version is changed.
```bash
$ cd rust-runtime/aws-smithy-http-server/examples
$ SMITHY_RS_VERSION_COMMIT_HASH_OVERRIDE=test make
$ rg -B1 -u codegen-version
pokemon-service-client/Cargo.toml
9-[package.metadata.smithy]
10:codegen-version = "0.47.0-test"

pokemon-service-server-sdk/Cargo.toml
9-[package.metadata.smithy]
10:codegen-version = "0.47.0-test"
```

### DIFF
--- a/.github/workflows/pull-request-bot.yml
+++ b/.github/workflows/pull-request-bot.yml
@@ -39,6 +39,8 @@ jobs:
     env:
       AWS_REGION: us-west-2
       S3_BUCKET_NAME: ${{ secrets.SMITHY_RS_PULL_REQUEST_CDN_S3_BUCKET_NAME }}
+      # Override version commit hash to prevent unnecessary diffs
+      SMITHY_RS_VERSION_COMMIT_HASH_OVERRIDE: ci
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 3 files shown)


crisidev (approved):

jdisanti (approved):

jdisanti:
> I had to move env variable to shell script because env variables in GitHub action is not propagated to script. Could you please review it again?

This is because the diff script is run inside the build Docker image. Another way you could approach this is have the execution of gradle from within the script set the environment variable.

crisidev (approved):

PR #1646, opened 2022-08-18
Allow * in ACCEPT header

82marbag:
## Motivation and Context
Allow `*/*`, `type/*` and `type/subtype`. This addresses https://github.com/awslabs/smithy-rs/issues/1544 and will comply with https://github.com/awslabs/smithy/issues/1332.

## Testing
`cargo test` in the Pokemon service and unit tests

## Checklist


Velfi (approved):

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -39,33 +39,75 @@ pub fn check_content_type<B>(
     }
 }
 
+pub fn accept_header_classifier<B>(req: &RequestParts<B>, content_type: &'static str) -> bool {
+    if req.headers().is_none() {
+        return true;
+    }
+    let headers = req.headers().unwrap();
+    if !headers.contains_key(http::header::ACCEPT) {
+        return true;
```
unexge, line 48: Shouldn't we reject requests if there is no `Accept` header present?

82marbag: From [here](https://httpwg.org/specs/rfc7231.html#header.accept),

> A request without any Accept header field implies that the user agent will accept any media type in response.

We're complying with this behavior here (same in the other implementations of smithy)

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -39,33 +39,75 @@ pub fn check_content_type<B>(
...
+                    let subtype = content_type.subtype();
+                    match (mim.type_(), mim.subtype()) {
+                        (t, s) if t == typ && s == subtype => true,
+                        (t, mime::STAR) if t == typ => true,
+                        (mime::STAR, mime::STAR) => true,
+                        _ => false,
+                    }
+                })
+                .unwrap_or(false)
+        })
```
crisidev, line 76: I think this piece of code deserve some extensive documentation. It's pretty hard to read, there is a lot going on between `map`, `flat_map`, multiple splits and so on.

Is the unwrap after the `nth(0)` safe?

crisidev (changes requested):

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -39,33 +39,75 @@ pub fn check_content_type<B>(
...
+    if req.headers().is_none() {
+        return true;
+    }
+    let headers = req.headers().unwrap();
+    if !headers.contains_key(http::header::ACCEPT) {
+        return true;
+    }
+    let content_type = content_type
+        .parse::<mime::Mime>()
+        .expect("BUG: content_type is not valid");
```
crisidev, line 52:
```suggestion
        .expect("BUG: MIME parsing failed, content_type is not valid");
```

rust-runtime/aws-smithy-http-server/src/protocols.rs
weihanglo: > Is the unwrap after the nth(0) safe?

It should return an empty string for an empty string, so safe.

crisidev (approved):

PR #1648, opened 2022-08-18
Protocol tests for ACCEPT header with *

82marbag:
## Motivation and Context

Add protocol tests for https://github.com/awslabs/smithy-rs/pull/1646

## Testing

` ./gradlew :codegen-server-test:test && ./gradlew :codegen-test:test`

## Checklist


codegen-test/model/misc.smithy
```diff
@@ -199,3 +200,42 @@ structure ResponseCodeRequiredOutput {
...
+])
+operation AcceptHeaderStarService {
+    input: AcceptHeaderStarServiceInput,
+    output: AcceptHeaderStarServiceOutput,
+}
+
+@output
+structure AcceptHeaderStarServiceOutput {}
+@input
+structure AcceptHeaderStarServiceInput {}
```
david-perez, line 241: I think we should dispense of superfluous modeling in the protocol tests, even if they are best practices, to keep the files minimal so readers can easily focus on the modeling aspects of the test that are strictly relevant to what it is testing.

So, e.g. not using `@input/@output`, using a single `AcceptHeaderStarServiceInputOutput` structure, not using `@idempotent` etc.

david-perez (approved):
Can you link to [the Smithy PR](https://github.com/awslabs/smithy/pull/1365) in a TODO somewhere in the model? So that we don't forget to remove this test when we upgrade to the next Smithy version.

codegen-test/model/misc.smithy
```diff
@@ -18,6 +18,7 @@ service MiscService {
         ResponseCodeRequiredOperation,
         ResponseCodeHttpFallbackOperation,
         ResponseCodeDefaultOperation,
+        AcceptHeaderStarService,
```
jdisanti, line 21: Shouldn't this be added to `codegen-server-test`?

82marbag: It's just a symlink:
```
$ ls -l codegen-server-test/model/misc.smithy
codegen-server-test/model/misc.smithy -> ../../codegen-test/model/misc.smithy
```

jdisanti (approved):

82marbag:
It was called out in https://github.com/http-rs/http-types/issues/349 that values can take the form `"x, y"` so splitting by comma isn't always the correct approach. However for us I don't think we will have this issue. We will move to an implementation by a library, but for now, I'll leave it like this.

PR #1651, opened 2022-08-22
Use . delimiter in OperationExtension

hlbarber:
## Motivation and Context

`#` is a special character which is rejected in many protocols.

## Description

- Change `#` in `OperationExtension` to `.`
- Add basic unit tests.

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/extension.rs
+++ b/rust-runtime/aws-smithy-http-server/src/extension.rs
@@ -69,7 +69,7 @@ pub struct OperationExtension {
 }
 
 /// An error occurred when parsing an absolute operation shape ID.
-#[derive(Debug, Clone, Error)]
+#[derive(Debug, Clone, Error, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum ParseError {
     #[error(". was not found - missing namespace")]
@@ -208,6 +208,9 @@ mod tests {
     #[test]
     fn ext_reject() {
         let value = "CompleteSnapshot";
-        assert!(OperationExtension::new(value).is_err())
+        assert_eq!(
+            OperationExtension::new(value).unwrap_err(),
+            ParseError::MissingNamespace
+        )
     }
 }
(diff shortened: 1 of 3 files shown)


rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -188,3 +190,24 @@ where
...
+
+        assert_eq!(ext.absolute(), value);
+        assert_eq!(ext.namespace(), "com.amazonaws.ebs");
+        assert_eq!(ext.name(), "CompleteSnapshot");
+    }
+
+    #[test]
+    fn ext_reject() {
+        let value = "CompleteSnapshot";
+        assert!(OperationExtension::new(value).is_err())
```
weihanglo, line 211: Could we assert the erro type as well?

rust-runtime/aws-smithy-http-server/src/extension.rs
```diff
@@ -70,15 +72,15 @@ pub struct OperationExtension {
...
     MissingNamespace,
 }
 
 impl OperationExtension {
-    /// Creates a new [`OperationExtension`] from the absolute shape ID of the operation.
+    /// Creates a new [`OperationExtension`] from the absolute shape ID of the operation with `#` symbol replaced with a `.`.
     pub fn new(absolute_operation_id: &'static str) -> Result<Self, ParseError> {
         let (namespace, name) = absolute_operation_id
-            .split_once('#')
+            .rsplit_once('.')
```
weihanglo, line 83: `rsplit` seems to work well,l though I worry that the caller accidentally pass a wrong shape ID, such as `com.amazonaws.ebs`. And then we interpret it as namespace `com.amazonaws` and name `ebs`.

It is unlikely to happen IMO but could we find a syntax less ambiguous?

weihanglo (commented):
Out of curious, what issue you bumped into making you post this patch?

rust-runtime/aws-smithy-http-server/src/extension.rs
hlbarber: @crisidev Any good ideas?

crisidev: The only thing I can think about it? Is it possible to have a service that is dottet? as in com.amazon.ebs.My.Service? I don't know..

With the old code it would be com.amazon.ebs#My.Service

crisidev (approved):

PR #1656, opened 2022-08-23
Fix aws-sigv4 canonical request formatting fallibility

ysaito1001:
This PR attempts to address [issue 711](https://github.com/awslabs/smithy-rs/issues/711) where signing a request can cause `panic!` if a `CanonicalRequest` contains invalid UTF-8 in the header value.

The PR intentionally starts with a simple solution that makes [a given test](https://github.com/awslabs/smithy-rs/issues/711#issuecomment-1222622313) pass before we commit ourselves to start refactoring `CanonicalRequest` to support a special formatting function, which will probably be more expensive to do so. If the solution described in this PR works, that is great but if not, we can certainly switch to the refactoring ideas mentioned in the open issue.

The rationale behind the code changes is to make validating UTF-8 in the header value be part of an invariant of `CanonicalRequest`. That is, by the time `CanonicalRequest` has been successfully constructed by `CanonicalRequest::from`, we can guarantee that the type always contains valid UTF-8 in the header value. That way, we don't need to wait to catch invalid UTF-8 in the header value until we call `to_string` on `CanonicalRequest` ([here](https://github.com/awslabs/smithy-rs/blob/9bfd1b2c0a485c029557aa83af866660d1b15673/aws/rust-runtime/aws-sigv4/src/http_request/sign.rs#L188)).
That being said, the proposed approach may be inefficient because of eager checking of valid UTF-8 in the header value within `normalize_header_value`, especially when a `CanonicalRequest` does not need to be signed (in which case I assume `to_string` on the `CanonicalRequest` does not need to be called in the execution path).

This PR comes with a test provided in the open issue as well as a small test for `normalize_header_value` to exercise the part that has been updated.

I would appreciate if the reviewers with more background knowledge could assess the code changes from different perspectives.

### DIFF
--- a/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
+++ b/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
@@ -196,7 +196,7 @@ impl<'a> CanonicalRequest<'a> {
             // Using append instead of insert means this will not clobber headers that have the same lowercased name
             canonical_headers.append(
                 HeaderName::from_str(&name.as_str().to_lowercase())?,
-                normalize_header_value(value),
+                normalize_header_value(value)?,
             );
         }
 
@@ -373,11 +373,12 @@ fn trim_spaces_from_byte_string(bytes: &[u8]) -> &[u8] {
     &bytes[starting_index..ending_index]
 }
 
-/// Works just like [trim_all] but acts on HeaderValues instead of bytes
-fn normalize_header_value(header_value: &HeaderValue) -> HeaderValue {
+/// Works just like [trim_all] but acts on HeaderValues instead of bytes.
+/// Will ensure that the underlying bytes are valid UTF-8.
+fn normalize_header_value(header_value: &HeaderValue) -> Result<HeaderValue, Error> {
     let trimmed_value = trim_all(header_value.as_bytes());
-    // This can't fail because we started with a valid HeaderValue and then only trimmed spaces
-    HeaderValue::from_bytes(&trimmed_value).unwrap()
+    let trimmed_value_as_utf8_str = std::str::from_utf8(&trimmed_value).map_err(Error::from)?;
+    HeaderValue::from_str(trimmed_value_as_utf8_str).map_err(Error::from)
 }
 
 #[derive(Debug, PartialEq, Default)]
@@ -738,9 +739,9 @@ mod tests {
         }
 
         #[test]
-        fn test_normalize_header_value_doesnt_panic(v in (".*")) {
+        fn test_normalize_header_value_works_on_valid_header_value(v in (".*")) {
             if let Ok(header_value) = HeaderValue::from_maybe_shared(v) {
-                let _ = normalize_header_value(&header_value);
+                assert!(normalize_header_value(&header_value).is_ok());
             }
         }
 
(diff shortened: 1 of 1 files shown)


rcoh:
this looks great! do you mind making a few small improvements:
- since `Display` is still infallible, we should add proptesting or fuzzing to the input to the signer to ensure that no matter what the input is, we can't cause a runtime panic
- will you add a test (and probably make a small update to the error) to ensure that the error that we emit gives a clear message about why signing failed? I think the current error probably won't indicate anything about the problem being a header

ysaito1001:
Thanks @rcoh ! I took a stab at incorporating the feedback.

> we should add proptesting or fuzzing to the input to the signer to ensure that no matter what the input is, we can't cause a runtime panic

I converted `test_signing_utf8_headers` to a proptest (renamed it and moved it to `sign.rs`) to exercise a wide range of inputs for request header values. As far as I can see, besides the the passed-in `Formatter` returning an error, the `Display` trait implementation for `CanonicalRequest` can only fail because of `std::str::from_utf8` for `HeaderValue`. So I specifically targeted the proptest at that use case. However, what you requested seems like exercising a higher entity than `HeaderValue`. To be on the same page, if the proptest is not aligned with what you thought, could you tell me the name of system under test, i.e. a type or a function to test, so I can test it further.

> will you add a test (and probably make a small update to the error) to ensure that the error that we emit gives a clear message about why signing failed?

I have added an error type `InvalidHeaderError` in the `canonical_request` module (since I was not sure whether I put it in the right place, please let me know if you want me to move `InvalidHeaderError` to its own `error` module or somewhere else). I have added a test in `sign.rs` to verify that if signing failed we should expect an `InvalidHeaderError`. I have used `downcast_ref` to compare the actual error type against the expected one but did not attempt to print out a diagnostic message and use it for verification as comparing against a raw string for verification can make the test brittle (please let me know if you think otherwise, I am open to suggestions). FYI, if we were to print an error when signing failed, it would look something like:
```
invalid UTF-8 contained in header value: invalid utf-8 sequence of 1 bytes from index 0
```

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
```diff
@@ -373,11 +409,16 @@ fn trim_spaces_from_byte_string(bytes: &[u8]) -> &[u8] {
...
+fn normalize_header_value(header_value: &HeaderValue) -> Result<HeaderValue, Error> {
     let trimmed_value = trim_all(header_value.as_bytes());
-    // This can't fail because we started with a valid HeaderValue and then only trimmed spaces
-    HeaderValue::from_bytes(&trimmed_value).unwrap()
+    let trimmed_value_as_utf8_str =
+        std::str::from_utf8(&trimmed_value).map_err(|e| InvalidHeaderError {
+            source: Error::from(e),
+            kind: InvalidHeaderErrorKind::InvalidUtf8,
+        })?;
+    HeaderValue::from_str(trimmed_value_as_utf8_str).map_err(Error::from)
```
jdisanti, line 421: Since `Error` is just `Box<dyn Error + Send + Sync + 'static`, I think you can just convert `from_utf8`'s error into it directly to avoid creating `InvalidHeaderError`.

Having the top-level error type be `Box<dyn Error>` is definitely not ideal though. I think that will get addressed through the error refactor though. There's a [RFC for errors](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0022_error_context_and_compatibility.md) if you want more context on that.

jdisanti (approved):
This looks good! I think can merge it as is, although it would be nice to improve the overall error situation in this crate.

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
ysaito1001: Thanks for taking a look!

> Since `Error` is just `Box<dyn Error + Send + Sync + 'static`, I think you can just convert `from_utf8`'s error into it directly to avoid creating `InvalidHeaderError`.

You are right. I try to recall why I did this way and remembered that it was an attempt to incorporate a suggestion by @rcoh
> I think the current error probably won't indicate anything about the problem being a header

You can see that a commit prior to the Russel's comment 1bf2bac did convert `from_uti8`'s error into `Error` directly.

I'm completely fine with removing my half-baked attempt `InvalidHeaderError` and converting `from_uti8`'s error into `Error` directly. Now that I'm going through the mentioned RFC, I think what's described there is a much better way to incorporate the Russell's suggestion moving forward.

What do you think?

jdisanti: Agreed. I think we should make a smaller change in this PR, and then fix all the errors in the error refactor.

jdisanti (approved):

jdisanti (approved):

Issue #1613, opened 2022-08-03
Multiple smithy-rs releases without releasing SDK may lose changelog entries


jdisanti:
If smithy-rs is released multiple times in between aws-sdk-rust releases, then changelog entries for aws-sdk-rust will be lost since the smithy-rs release will completely replace the `SDK_CHANGELOG.next.json` file during the `changelogger split` subcommand.

Steps to fix this:
- [ ] Move canary `generate-matrix` octorust logic to list release tags into `smithy-rs-tool-common` behind a feature flag
- [ ] In `changelogger split`:
    - [ ] Pull release tags for awslabs/aws-sdk-rust to find the latest release
    - [ ] Load the `versions.toml` at that release to get the smithy-rs revision
    - [ ] Load the old `SDK_CHANGELOG.next.json` file in smithy-rs
    - [ ] Filter the old entries by [using git](https://stackoverflow.com/questions/3005392/how-can-i-tell-if-one-commit-is-a-descendant-of-another-commit) to determine which `since_commit` entries are no longer relevant
    - [ ] Merge retained old entries with new entries
- [ ] Update existing `changelogger split` tests
- [ ] Add new tests to exercise filter/merge logic

jdisanti:
An easier approach that should work well enough:
- [ ] Add simple counts to each JSON changelog entry, starting at 0
- [ ] Each release increments the counts, and entries that make it to 5 are removed
That will allow 5 smithy-rs releases per SDK release, and if 5 turns out to not be enough, we can simply increase that number as needed. That many smithy-rs releases before a SDK release seems highly unlikely though.

jdisanti:
This is now fixed.

PR #1661, opened 2022-08-23
Retain old SDK changelog entries for multiple smithy-rs releases

jdisanti:
## Motivation and Context
This addresses #1613.


tools/changelogger/src/split.rs
```diff
@@ -8,8 +8,9 @@ use clap::Parser;
 use smithy_rs_tool_common::changelog::Changelog;
 use smithy_rs_tool_common::git::{find_git_repository_root, Git, GitCLI};
 use std::path::{Path, PathBuf};
-use std::{env, fs};
+use std::{env, fs, mem};
 
+const MAX_ENTRY_AGE: usize = 5;
```
Velfi, line 13: Can you add a comment about why we chose 5 and what it means?

Velfi (approved):

PR #1662, opened 2022-08-24
Upgrade min versions of tokio, tracing-subscriber, and criterion

jdisanti:
## Motivation and Context
The [minimum versions audit in aws-sdk-rust](https://github.com/awslabs/aws-sdk-rust/runs/7963423906?check_suite_focus=true) revealed that some of our minimum dependency versions need to be upgraded.

This PR upgrades the minimum versions of tokio, tracing-subscriber, and criterion to address the advisories.

For choosing what dependency versions to use, I used the following logic:
- Is the crate with the dependency consumed as a library, and is the dependency not a `dev-dependency`? Use the minimum secure version
- Otherwise? Use the latest version

This resulted in the tools and the Python server taking the latest dependency versions, while everything else took the minimum secure versions.


crisidev (commented):
Why is tokio version different (`1.20.1` or `1.8.4`)? Can't we update everything to `1.20.1`?

weihanglo:
> Why is tokio version different (`1.20.1` or `1.8.4`)? Can't we update everything to `1.20.1`?

I believe @jdisanti was following this rule _"Use the minimum secure version"_, but it did confused me that pokemon example service uses a newer version of tokio while the underlying rust-runtime pinned in an older one. Is that how customer use smithy-rs in real life?

jdisanti:
Libraries should strive to use the lowest possible version to maximize compatibility with other libraries that may have lower version requirements to compile. It is OK to have the library ask for 1.8.4 while the thing that depends on the library requests 1.20.1. In that case, Cargo will automatically choose 1.20.1 since that's compatible with the library's semver requirements.

Velfi (approved):

jdisanti:
@crisidev, @weihanglo - Any other concerns?

PR #1665, opened 2022-08-24
Add "invalid xml body root" check exemption for S3's `GetObjectAttributes`

Velfi:
## Motivation and Context

aws-sdk-rust#609

## Description

It is now possible to exempt specific operations from XML body root checking. To do this, add the `AllowInvalidXmlRoot`
trait to the output struct of the operation you want to exempt. Previously, S3's `GetObjectAttributes` would always fail because they return an XML object with a root named `GetObjectAttributesResponse` instead of `GetObjectAttributesOutput` as specified by their model.

_They'll still be returning the incorrect XML, we just won't hold them to their model once this is merged._

## Testing

I have included an integration test

## Checklist

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
@@ -6,8 +6,13 @@
 package software.amazon.smithy.rustsdk.customize.s3
 
 import software.amazon.smithy.aws.traits.protocols.RestXmlTrait
+import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.model.shapes.Shape
 import software.amazon.smithy.model.shapes.ShapeId
+import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.model.transform.ModelTransformer
 import software.amazon.smithy.rust.codegen.rustlang.CargoDependency
 import software.amazon.smithy.rust.codegen.rustlang.RustModule
 import software.amazon.smithy.rust.codegen.rustlang.Writable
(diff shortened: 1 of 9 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
```diff
@@ -23,17 +28,24 @@ import software.amazon.smithy.rust.codegen.smithy.customize.RustCodegenDecorator
...
  * Top level decorator for S3
  */
 class S3Decorator : RustCodegenDecorator<ClientCodegenContext> {
-    override val name: String = "S3ExtendedError"
+    override val name: String = "S3"
     override val order: Byte = 0
+    private val logger: Logger = Logger.getLogger(javaClass.name)
+    private val invalidXmlRootAllowList = setOf(
+        // API returns GetObjectAttributes_Response_ instead of Output
+        ShapeId.from("com.amazonaws.s3#GetObjectAttributesOutput"),
```
jdisanti, line 47: I took a brief stroll through botocore to see if there were any other obvious operations that have the wrong shape, and we may want to check `s3#GetBucketLocation`, `s3#ListObjects`, `s3#ListObjectVersions`, and `ec2#GetConsoleOutput` (this last one is using the ec2Query protocol, which still has XML responses). Not saying these ones meet the criteria, but that I see [special code in botocore parsing these XML responses in botocore](https://github.com/boto/botocore/blob/develop/botocore/handlers.py). If it's easy to check these, then we should, but probably not worth spending a ton of time on.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestXml.kt
```diff
@@ -110,3 +113,12 @@ open class RestXml(private val coreCodegenContext: CoreCodegenContext) : Protoco
 
     override fun serverRouterRuntimeConstructor() = "new_rest_xml_router"
 }
+
+/**
+ * Indicates that a service is expected to send XML where the root element name does not match the modeled member name.
+ */
+class AllowInvalidXmlRoot : AnnotationTrait(ID, Node.objectNode()) {
```
jdisanti, line 120: When I was cross checking the S3 documentation with the S3 model to try and find other operations that may be broken (which unfortunately turned out to be a pointless endeavor&mdash;they're both wrong), I noticed that some S3 operations use the `@xmlName` trait to correct the wrong root. For example, `GetBucketLocationOutput` has `@xmlName("LocationConstraint")` to fix its root. Should the S3 customization just add that trait to the outputs for these cases rather than the codegen needing to handle a separate synthetic trait?

jdisanti (commented):
Good work on this! Most of my comments are minor, and feel free to punt on the ones that aren't minor to a follow up so that we can get a release out with this fixed.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestXml.kt
```diff
@@ -110,3 +113,12 @@ open class RestXml(private val coreCodegenContext: CoreCodegenContext) : Protoco
...
+    companion object {
+        val ID: ShapeId = ShapeId.from("aws.api.internal#allowInvalidXmlRoot")
```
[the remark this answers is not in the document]
Velfi: Ah, good catch!

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/parse/RestXmlParserGenerator.kt
```diff
@@ -18,13 +26,43 @@ class RestXmlParserGenerator(
...
+                        #{tracing}::trace!("legacy API returned invalid root, expected $shapeName but got {:?}", start_el)
+                        """,
+                        "tracing" to CargoDependency.Tracing.asType(),
+                    )
+                }
+            } else {
+                writable {
+                    rustTemplate(
+                        """
+                        // Zelda B
```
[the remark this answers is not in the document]
Velfi: Yeah 😝

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
Velfi: I've broken this work out into a separate issue: #1668

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/RestXml.kt
Velfi: I've broken this work out into a separate issue: #1668

jdisanti (approved):

PR #1666, opened 2022-08-24
Add protocol specific routers

hlbarber:
## Motivation and Context

See **Protocol specific Routers** and **Protocol specific Errors** section of https://github.com/awslabs/smithy-rs/pull/1620.

## Description

- Add protocol specific routers
- Add `IntoResponse<Protocol>` to allow for multiple implementations of `IntoResponse`.
- Remove `UnknownOperation` from `RuntimeError`.
- Convert directly from routing errors to a `http::Response` via `IntoResponse`.
- Refactor existing `Router` to use protocol specific routers internally.

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/response.rs
+++ b/rust-runtime/aws-smithy-http-server/src/response.rs
@@ -36,3 +36,7 @@ use crate::body::BoxBody;
 
 #[doc(hidden)]
 pub type Response<T = BoxBody> = http::Response<T>;
+
+pub trait IntoResponse<Protocol> {
+    fn into_response(self) -> http::Response<BoxBody>;
+}
--- a/rust-runtime/aws-smithy-http-server/src/response.rs
+++ b/rust-runtime/aws-smithy-http-server/src/response.rs
@@ -37,6 +37,8 @@ use crate::body::BoxBody;
 #[doc(hidden)]
 pub type Response<T = BoxBody> = http::Response<T>;
 
+/// A protocol aware function taking `self` to [`http::Response`].
 pub trait IntoResponse<Protocol> {
+    /// Performs a conversion into a [`http::Response`].
     fn into_response(self) -> http::Response<BoxBody>;
 }
--- a/rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
+++ b/rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
@@ -1,3 +1,8 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 use std::convert::Infallible;
 
 use http::header::ToStrError;
--- a/rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
+++ b/rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
@@ -10,7 +10,7 @@ use thiserror::Error;
 use tower::{Layer, Service};
 
 use crate::{
-    body::BoxBody,
+    body::{empty, BoxBody},
     extension::RuntimeErrorExtension,
     protocols::{AwsJson10, AwsJson11},
     response::IntoResponse,
@@ -49,8 +49,8 @@ impl IntoResponse<AwsJson10> for Error {
                 .extension(RuntimeErrorExtension::new(
                     super::UNKNOWN_OPERATION_EXCEPTION.to_string(),
                 ))
-                .body(crate::body::to_boxed(""))
-                .expect("invalid HTTP response"),
+                .body(empty())
+                .expect("invalid HTTP response for AWS JSON routing error; please file a bug report under https://github.com/awslabs/smithy-rs/issues"),
         }
     }
 }
@@ -65,8 +65,8 @@ impl IntoResponse<AwsJson11> for Error {
                 .extension(RuntimeErrorExtension::new(
                     super::UNKNOWN_OPERATION_EXCEPTION.to_string(),
                 ))
-                .body(crate::body::to_boxed(""))
-                .expect("invalid HTTP response"),
+                .body(empty())
+                .expect("invalid HTTP response for AWS JSON routing error; please file a bug report under https://github.com/awslabs/smithy-rs/issues"),
         }
     }
 }
--- a/rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
+++ b/rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
@@ -23,7 +23,7 @@ pub enum Error {
     /// Relative URI was not "/".
     NotRootUrl,
     /// Method was not `POST`.
-    MethodDisallowed,
+    MethodNotAllowed,
     /// Missing the `x-amz-target` header.
     MissingHeader,
     /// Unable to parse header into UTF-8.
@@ -35,7 +35,7 @@ pub enum Error {
 impl IntoResponse<AwsJson10> for Error {
     fn into_response(self) -> http::Response<BoxBody> {
         match self {
-            Error::MethodDisallowed => super::method_disallowed(),
+            Error::MethodNotAllowed => super::method_disallowed(),
             _ => http::Response::builder()
                 .status(http::StatusCode::NOT_FOUND)
                 .header("Content-Type", "application/x-amz-json-1.0")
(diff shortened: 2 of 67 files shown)


rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
```diff
@@ -0,0 +1,128 @@
...
+    protocols::{AwsJson10, AwsJson11},
+    response::IntoResponse,
+    routing::{tiny_map::TinyMap, Route},
+};
+
+use super::Router;
+
+pub enum Error {
+    NotRootUrl,
+    MethodDisallowed,
```
david-perez, line 18:
```suggestion
    MethodNotAllowed,
```

To match HTTP 405.

rust-runtime/aws-smithy-http-server/src/routing/routers/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    response::IntoResponse,
+};
+
+const UNKNOWN_OPERATION_EXCEPTION: &str = "UnknownOperationException";
+
+fn method_disallowed() -> http::Response<BoxBody> {
+    http::Response::builder()
+        .status(http::StatusCode::METHOD_NOT_ALLOWED)
+        .body(empty())
+        .expect("valid HTTP response")
```
david-perez, line 28:
```suggestion
        .expect("invalid HTTP response")
```

Like in the other places.

I'd also be consistent with the other places where we know that if it panics, it's a bug on us, and direct the user to file a report. See https://github.com/awslabs/smithy-rs/blob/a1f2e3edeb0f639029e30da6c1b090ee492bdb3d/rust-runtime/aws-smithy-http-server/src/runtime_error.rs#L92-L92

david-perez: Actually, I'm also thinking whether in these places we can dispense with the panic by doing:

```rust
let mut response = Response::new(empty()); // Or `Response::default()`.
*response.status_mut() = StatusCode::METHOD_NOT_ALLOWED;
```

rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl IntoResponse<AwsJson10> for Error {
+    fn into_response(self) -> http::Response<BoxBody> {
+        match self {
+            Error::MethodDisallowed => super::method_disallowed(),
+            _ => http::Response::builder()
+                .status(http::StatusCode::NOT_FOUND)
+                .header("Content-Type", "application/x-amz-json-1.0")
+                .extension(RuntimeErrorExtension::new(
+                    super::UNKNOWN_OPERATION_EXCEPTION.to_string(),
```
david-perez, line 32: You don't have to change it, but my preference for some time has been to use `to_owned()`, because of https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441/6

Curious to learn what you prefer and why.

rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+}
+
+impl IntoResponse<AwsJson11> for Error {
+    fn into_response(self) -> http::Response<BoxBody> {
+        match self {
+            Error::MethodDisallowed => super::method_disallowed(),
+            _ => http::Response::builder()
+                .status(http::StatusCode::NOT_FOUND)
```
david-perez, line 45: Would be nice to debug log the concrete error?

rust-runtime/aws-smithy-http-server/src/routing/routers/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    pub fn map<RNew, F: FnOnce(R) -> RNew>(self, f: F) -> RoutingService<RNew, P> {
+        RoutingService {
+            router: f(self.router),
+            _protocol: PhantomData,
+        }
+    }
+}
+
+type EitherOneshotReady<S, B> = Either<
```
david-perez, line 83: Love how the types and bounds work out beautifully in this file below this line.

rust-runtime/aws-smithy-http-server/src/routing/routers/rest.rs
```diff
@@ -0,0 +1,129 @@
...
+        request_spec::{Match, RequestSpec},
+        Route,
+    },
+};
+
+use super::Router;
+
+pub enum Error {
+    NotFound,
+    MethodDisallowed,
```
david-perez, line 20:
```suggestion
    MethodNotAllowed,
```

david-perez (approved):

rust-runtime/aws-smithy-http-server/src/response.rs
```diff
@@ -36,3 +36,7 @@ use crate::body::BoxBody;
 
 #[doc(hidden)]
 pub type Response<T = BoxBody> = http::Response<T>;
+
+pub trait IntoResponse<Protocol> {
```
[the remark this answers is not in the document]
hlbarber: There's already a `#[doc(hidden)]` at the module level in `lib.rs`. `Response` is double hidden, which threw me off too - should I remove that?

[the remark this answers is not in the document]
david-perez: > There's already a #[doc(hidden)] at the module level in lib.rs

Ah, missed that. Thought it wasn't because `Response` was doc hidden.

Ok then, no need to double hide `Response` --- although, I'm thinking that middleware authors will need access to it right?

rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
```diff
@@ -0,0 +1,128 @@
...
+    MissingHeader,
+    InvalidHeader(ToStrError),
+    NotFound,
+}
+
+impl IntoResponse<AwsJson10> for Error {
+    fn into_response(self) -> http::Response<BoxBody> {
+        match self {
+            Error::MethodDisallowed => super::method_disallowed(),
+            _ => http::Response::builder()
```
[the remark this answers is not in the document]
david-perez: We should only do things that are in the spec of these protocols. I don't even know if implementations of these protocols return 405 when the URI matches the route but the verb doesn't... These are good questions to be asking in the awslabs/smithy issue tracker, and push for better protocol test coverage.

rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
hlbarber: I prefer `to_string` over `into_owned` generally because the reader immediately knows that the output type is `String` when they read `s.to_string()` whereas with `s.to_owned()` you can't know without prior knowledge of `s`'s type. I guess if this was an inline string literal that wouldn't be a problem.

rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
hlbarber: Shall we implement `Error` on the router errors and write a general `error!` in the `RouterService::call`?

rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
david-perez: Yeah, fair point, it's only applicable for string literals defined in place.

rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
david-perez: Not `error!`, use debug. Services receive invalid requests from misbehaving/scanning clients all the time, and teams have alarms triggering on logs at the error level.

rust-runtime/aws-smithy-http-server/src/response.rs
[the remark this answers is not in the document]
hlbarber: I agree that library authors probably want to see `IntoResponse` and `FromRequest` eventually. But was thinking of keeping them hidden until everything in the RFC has been implemented before removing the `#[doc(hidden)]`. I'm not 100% sure on whether `IntoResponse` needs a second parameterization over the operation, which would be a breaking change.

rust-runtime/aws-smithy-http-server/src/routing/routers/mod.rs
hlbarber: Switched the `method_disallowed` builder to the `Response` mutation. In the cases where headers are set we'd be switching out the explicit `.expect` for a hidden panic in `HeaderValue::from_static`.

rust-runtime/aws-smithy-http-server/src/routing/routers/aws_json.rs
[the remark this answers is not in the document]
hlbarber: https://github.com/awslabs/smithy/issues/1368

PR #1637, opened 2022-08-16
Add initial RFC document (endpoints 2.0)

rcoh:
## Motivation and Context
Design for upcoming implementation of endpoints 2.0
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

[Rendered](https://github.com/awslabs/smithy-rs/blob/endpoints-20-rfc/design/src/rfcs/rfc0024_endpoints_20.md)

PR #1667, opened 2022-08-25
Endpoints 2.0 Standard Library functions

rcoh:
## Motivation and Context
For #1637, the language relies on a set of standard library functions underpinning the generated rules. This adds a subset of those functions. 

## Description
- `substring`
- `aws.parseArn`
- `parseUrl`
- `isValidHostLabel`

## Testing
- [x] unit and proptests of individual functions
Detailed and complete testing of these functions will be performed when we generate the endpoint ruleset test suite in a subsequent PR
## Checklist
- no changelog

### DIFF
--- a/rust-runtime/inlineable/src/endpoint_lib/arn.rs
+++ b/rust-runtime/inlineable/src/endpoint_lib/arn.rs
@@ -1,0 +1,138 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+use crate::endpoint_lib::diagnostic::DiagnosticCollector;
+use std::borrow::Cow;
+use std::error::Error;
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug, Eq, PartialEq)]
+pub(crate) struct Arn<'a> {
+    partition: &'a str,
+    service: &'a str,
+    region: &'a str,
+    account_id: &'a str,
+    resource_id: Vec<&'a str>,
+}
+
+#[allow(unused)]
+impl<'a> Arn<'a> {
+    pub(crate) fn partition(&self) -> &'a str {
+        self.partition
+    }
+    pub(crate) fn service(&self) -> &'a str {
+        self.service
+    }
+    pub(crate) fn region(&self) -> &'a str {
+        self.region
+    }
+    pub(crate) fn account_id(&self) -> &'a str {
+        self.account_id
+    }
+    pub(crate) fn resource_id(&self) -> &Vec<&'a str> {
+        &self.resource_id
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub(crate) struct InvalidArn {
+    message: Cow<'static, str>,
+}
+
+impl InvalidArn {
+    fn from_static(message: &'static str) -> InvalidArn {
+        Self {
+            message: Cow::Borrowed(message),
+        }
+    }
+}
+impl Display for InvalidArn {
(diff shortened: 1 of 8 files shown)


rust-runtime/inlineable/src/endpoint_lib/arn.rs
```diff
@@ -0,0 +1,138 @@
...
+        write!(f, "{}", self.message)
+    }
+}
+impl Error for InvalidArn {}
+
+impl<'a> Arn<'a> {
+    pub(crate) fn parse(arn: &'a str) -> Result<Self, InvalidArn> {
+        let mut split = arn.splitn(6, ':');
+        let invalid_format =
+            || InvalidArn::from_static("ARN must have 6 components delimited by `:`");
```
Velfi, line 62: Sometimes I read closures as logical ORs and it breaks my brain to not see the first part of the statement

rust-runtime/inlineable/src/endpoint_lib/arn.rs
```diff
@@ -0,0 +1,138 @@
...
+impl Display for InvalidArn {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.message)
+    }
+}
+impl Error for InvalidArn {}
+
+impl<'a> Arn<'a> {
+    pub(crate) fn parse(arn: &'a str) -> Result<Self, InvalidArn> {
+        let mut split = arn.splitn(6, ':');
```
Velfi, line 60: There are ARNs that don't fit this pattern. Should we be more clear that this is only for parsing a specific subset of ARNs?

This isn't yet public but a comment would be nice just so future us don't get confused about that.

https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-paths

rust-runtime/inlineable/src/endpoint_lib/parse_url.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+    if !["http", "https"].contains(&url.scheme()) {
+        e.report_error(format!(
+            "URL scheme must be HTTP or HTTPS (found {})",
+            url.scheme()
+        ));
+        return None;
+    }
+    Some(Url { url, uri, raw })
+}
```
Velfi, line 68: It's def overkill but it could be cool to use something like [`miette`](https://docs.rs/miette/latest/miette/) for these error messages

rust-runtime/inlineable/src/endpoint_lib/substring.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // substring doesn't support unicode, it always returns none
+    #[test]
+    fn substring_unicode() {
+        assert_eq!(
+            substring("a🐱b", 0, 2, false, &mut DiagnosticCollector::new()),
+            None
+        );
+    }
```
Velfi, line 86: Could you also add a check that the collector receives the "not ASCII" error message?

rust-runtime/inlineable/src/endpoint_lib/arn.rs
rcoh: do you have an example of an ARN you're talking about? The path at the end gets split up into the resource, and those are still 6-part ARNs

Velfi: Ah, Ok. I didn't realize the 5-colon ARNs like `arn:aws:s3:::my_corporate_bucket/Development/*` had an implicit sixth one

rcoh: there is a `::` in the middle (it's not implicit, just empty)

Issue #1676, opened 2022-08-26
When `runtimeConfig` is _not_ specified, the default behavior is still not using the latest runtime crate versions on crates.io


david-perez:
After #1635, we made it so that if the `runtimeConfig` key is specified, you now don't need to specify `version: "DEFAULT"` to make the code generated crate depend on the runtime crates on crates.io. Instead, now users have more flexibility: by default it will use the latest crates on crates.io, but they can also use the `versions` (note the plural) key to specify an explicit version for a runtime crate, or a path dependency.

However, if the `runtimeConfig` key is _not_ specified, then we go through the else branch here

https://github.com/awslabs/smithy-rs/blob/30d985ad9d39849bb40705d244413eae2764ba7f/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt#L92-L78

and `RuntimeCrateLocation` is thus set to `RuntimeCrateLocation.Path("../")`, a path dependency. So the default behavior is still not using the latest runtime crate versions on crates.io in this case.

82marbag:
Can this be closed?

david-perez:
Yes. Closing.

PR #1678, opened 2022-08-26
missing `RuntimeConfig` node should behave the same as `{}`

weihanglo:
## Motivation and Context

Address #1676. #1635 forgot to handle a case that the whole `RuntimeConfig` is missing.

## Description

If seeing a empty node for where `RuntimeConfig` should be, we just treat it as `{}` empty object. They now share the same behaviour that by default using runtime crates with the versions from crates.io.

## Checklist

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
@@ -75,21 +75,18 @@ data class RuntimeConfig(
         /**
          * Load a `RuntimeConfig` from an [ObjectNode] (JSON)
          */
-        fun fromNode(node: Optional<ObjectNode>): RuntimeConfig {
-            return if (node.isPresent) {
-                val crateVersionMap = node.get().getObjectMember("versions").orElse(Node.objectNode()).members.entries.let { members ->
-                    val map = members.associate { it.key.toString() to it.value.expectStringNode().value }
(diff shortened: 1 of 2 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/RuntimeTypes.kt
```diff
@@ -75,21 +75,18 @@ data class RuntimeConfig(
...
-                val path = node.get().getStringMember("relativePath").orNull()?.value
-                val runtimeCrateLocation = RuntimeCrateLocation(path = path, versions = crateVersionMap)
-                RuntimeConfig(
-                    node.get().getStringMemberOrDefault("cratePrefix", "aws-smithy"),
-                    runtimeCrateLocation = runtimeCrateLocation,
-                )
-            } else {
-                RuntimeConfig()
+        fun fromNode(node_: Optional<ObjectNode>): RuntimeConfig {
+            val node = node_.orElse(Node.objectNode())
```
jdisanti, line 79:
```suggestion
        fun fromNode(maybeNode: Optional<ObjectNode>): RuntimeConfig {
            val node = maybeNode.orElse(Node.objectNode())
```

jdisanti (approved):
Looks good. Will be interesting to see if this breaks the SDK CI or not--we probably don't explicitly set a path dependency for it.

jdisanti (approved):

david-perez (approved):
Thank you.

PR #1682, opened 2022-08-30
Add partition function

rcoh:
## Motivation and Context
Adds the `aws.partition` function for endpoints 2.0

## Description
Adds support for the partition function

## Testing
- [x] basic testing. More thorough testing will come from generated endpoint tests

## Checklist

### DIFF
--- a/rust-runtime/inlineable/Cargo.toml
+++ b/rust-runtime/inlineable/Cargo.toml
@@ -23,11 +23,11 @@ repository = "https://github.com/awslabs/smithy-rs"
 "pin-project-lite" = "0.2"
 "tower" = { version = "0.4.11", default_features = false }
 "async-trait" = "0.1"
+regex = "1"
 "url" = "2.2.2"
 
 [dev-dependencies]
 proptest = "1"
-regex = "1"
 
 [package.metadata.docs.rs]
 all-features = true
--- a/rust-runtime/inlineable/src/endpoint_lib/partition.rs
+++ b/rust-runtime/inlineable/src/endpoint_lib/partition.rs
@@ -1,0 +1,573 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Partition function to determine a partition for a given region
+//!
+//! This function supports adding regions dynamically, parsing a JSON file, and builder construction.
+//!
+//! If, at a future point, this interface stabilizes it is a good candidate for extraction into a
+//! shared crate.
+use crate::endpoint_lib::diagnostic::DiagnosticCollector;
+use aws_smithy_json::deserialize::Error;
+use regex::Regex;
+use std::borrow::Cow;
+use std::collections::HashMap;
+
+/// Determine the AWS partition metadata for a given region
+pub(crate) struct PartitionResolver {
+    partitions: Vec<PartitionMetadata>,
+}
+
+impl PartitionResolver {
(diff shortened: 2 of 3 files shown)


rust-runtime/inlineable/src/endpoint_lib/partition.rs
```diff
@@ -0,0 +1,577 @@
...
+
+static DEFAULT_OVERRIDE: &PartitionOutputOverride = &PartitionOutputOverride {
+    name: None,
+    dns_suffix: None,
+    dualstack_dns_suffix: None,
+    supports_fips: None,
+    supports_dualstack: None,
+};
+
+/// Merge the base output and the override output, dealing with Cow's
```
Velfi, line 41:
```suggestion
/// Merge the base output and the override output, dealing with `Cow`s
```

rust-runtime/inlineable/src/endpoint_lib/partition.rs
```diff
@@ -0,0 +1,577 @@
...
+    pub(crate) fn add_partition(&mut self, partition: PartitionMetadata) {
+        self.partitions.push(partition);
+    }
+
+    /// Resolve a partition for a given region
+    ///
+    /// 1. Enumerate each partition in the `partitions` array, and determine if the identifier to be resolved matches an
+    ///    explicit
+    ///    region listed in the `regions` array for a given partition. If identifier matches, proceed to step 4, otherwise
+    ///    continue to step 2.
```
Velfi, line 65:
```suggestion
    /// 1. Enumerate each partition in the `partitions` array, and determine if the identifier to be resolved matches an
    ///    explicit region listed in the `regions` array for a given partition. If identifier matches, proceed to step 4, 
    ///    otherwise continue to step 2.
```

rust-runtime/inlineable/src/endpoint_lib/partition.rs
```diff
@@ -0,0 +1,577 @@
...
+    /// 2. Enumerate each partition in the `partitions` array, use the regular expression `regionRegex` to determine if the
+    ///    identifier matches the regular expression. If the identifier matches, proceed to step 4, otherwise continue to step
+    /// 3. If no partition is matched after exhausting step 1 and step 2, then fallback to matching the identifier to the
+    ///    partition where `id == "aws"`, and proceed to step 4. If no `aws` partition is present, return `None`
+    /// 4. After matching the identifier to a partition using one of the previous steps, the partition function should return a
+    ///    typed data structure containing the fields in `outputs` in the matched partition. **Important:** If a specific region
+    ///    was matched, the properties associated with that region **MUST** be merged with the `outputs` field.
+    pub(crate) fn resolve_partition(
```
Velfi, line 73: Consider adding an example to these docs

rust-runtime/inlineable/Cargo.toml
```diff
@@ -23,11 +23,11 @@ repository = "https://github.com/awslabs/smithy-rs"
 "pin-project-lite" = "0.2"
 "tower" = { version = "0.4.11", default_features = false }
 "async-trait" = "0.1"
+regex = "1"
```
jdisanti, line 26: Hmmm... It looks like we should be using a minimum of 1.5.5: https://rustsec.org/advisories/RUSTSEC-2022-0013.html

I guess we must have a dependency that forces upgrade to that version since the minimal version audit hasn't caught this.

Would also need to be fixed here: https://github.com/awslabs/smithy-rs/blob/462aef63945883d9567b2e8cd2be324016f8111e/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/CargoDependency.kt#L205

PR #1683, opened 2022-08-30
Endpoint params

rcoh:
## Motivation and Context
This adds generation of the `EndpointParams` struct from endpoint rules. This is the first of a series of PRs to the `feature/endpoints` branch. This uses a `source` dependency to avoid the need to manually publishToMavenLocal. Prior to merging to mainline, we will need to finalize the Ruleset changes in a smithy release and remove the source override.

## Description
- add endpoint params generator
- add tests that uses the test cases provided by endpoints 2.0

## Testing
- standard compile-and-test unit tests
## Checklist

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/endpoints/EndpointParamsGenerator.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/endpoints/EndpointParamsGenerator.kt
@@ -1,0 +1,282 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.rust.codegen.smithy.endpoints
+
+import software.amazon.smithy.rulesengine.language.EndpointRuleset
+import software.amazon.smithy.rulesengine.language.eval.Value
+import software.amazon.smithy.rust.codegen.rustlang.Attribute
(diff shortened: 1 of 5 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/endpoints/EndpointParamsGenerator.kt
```diff
@@ -0,0 +1,282 @@
...
+                }
+                rust("Ok(#W)", params)
+            }
+            endpointRules.parameters.toList().forEach { parameter ->
+                val name = parameter.memberName()
+                val type = parameter.symbol().mapRustType { t -> t.stripOuter<RustType.Option>() }
+                rustTemplate(
+                    """
+                    /// Sets the value for $name #{extraDocs:W}
+                    pub fn $name(mut self, value: #{type}) -> Self {
```
jdisanti, line 256: Should this setter also use `impl Into<#{nonOptionalType}>`?

jdisanti (approved):
LGTM! Just one minor comment.

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/endpoints/EndpointParamsGenerator.kt
rcoh: ah—I didn't do that because it makes editors get super confused and not show good hints.

PR #1693, opened 2022-08-31
Add new service builder codegen

hlbarber:
## Motivation and Context

https://github.com/awslabs/smithy-rs/pull/1679

## Description

- Add `ServerProtocol` interface to allow for server side protocol specific methods.
- Make public the structs merged in https://github.com/awslabs/smithy-rs/pull/1679.
- Add `ServerOperationGenerator`, which generates a ZST and implements `OperationShape` on it.
- Add `ServerServiceGeneratorV2`, which generates the service newtype around a router and a service builder.
- Add `hidden` argument to `RustModule` which allows modules to be marked with `#[doc(hidden)]`.
- Add `BuildModifier` trait to provide a common interface for extending service builders.
- Add `Upgradable` trait to simplifying bounds when upgrading from an `Operation` to a HTTP service.
- Add `FromRequest`, `FromParts`, and `IntoResponse` implementations.
- Make `RoutingService` accept general body types `B` for the inner services `http::Response<B>`.
- Use new service builder in protocol tests.

## Notes

- All new generated code is placed in `operation_shape.rs` and `service.rs`. These are both marked as `#[doc(hidden)]` for now.
- The generated diff should be examined extensively. 
- Follow up tasks:
    - https://github.com/awslabs/smithy-rs/issues/1707
    - https://github.com/awslabs/smithy-rs/issues/1706
    - https://github.com/awslabs/smithy-rs/issues/1705
    - https://github.com/awslabs/smithy-rs/issues/1703


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -0,0 +1,206 @@
...
+    companion object {
+        // Upgrades the core protocol to a `ServerProtocol`.
+        fun fromCoreProtocol(coreCodegenContext: CoreCodegenContext, protocol: Protocol): ServerProtocol {
+            val serverProtocol = when (protocol) {
+                is AwsJson -> ServerAwsJsonProtocol(coreCodegenContext, protocol)
+                is RestJson -> ServerRestJsonProtocol(coreCodegenContext, protocol)
+                is RestXml -> ServerRestXmlProtocol(coreCodegenContext, protocol)
+                else -> throw IllegalStateException("unsupported protocol")
+            }
+            return serverProtocol
```
david-perez, line 64:
```suggestion
        // Upgrades the core protocol to a `ServerProtocol`.
        fun fromCoreProtocol(coreCodegenContext: CoreCodegenContext, protocol: Protocol): ServerProtocol =
            when (protocol) {
                is AwsJson -> ServerAwsJsonProtocol(coreCodegenContext, protocol)
                is RestJson -> ServerRestJsonProtocol(coreCodegenContext, protocol)
                is RestXml -> ServerRestXmlProtocol(coreCodegenContext, protocol)
                else -> throw IllegalStateException("unsupported protocol: $protocol")
            }
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -0,0 +1,206 @@
...
+        val name = when (coreProtocol.version) {
+            is AwsJsonVersion.Json10 -> {
+                "AwsJson10"
+            }
+            is AwsJsonVersion.Json11 -> {
+                "AwsJson11"
+            }
+            else -> throw IllegalStateException()
+        }
+        return RuntimeType(name, ServerCargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::protocols")
```
david-perez, line 93: Use `ServerCargoDependency.SmithyHttpServer(runtimeConfig).asType().namespace` here and in the other protocols.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -0,0 +1,206 @@
...
+                is RestJson -> ServerRestJsonProtocol(coreCodegenContext, protocol)
+                is RestXml -> ServerRestXmlProtocol(coreCodegenContext, protocol)
+                else -> throw IllegalStateException("unsupported protocol")
+            }
+            return serverProtocol
+        }
+    }
+}
+
+class ServerAwsJsonProtocol(
```
david-perez, line 69: So are these meant to replace the existing server protocol classes? Like, the `ServerAwsJson` class in `ServerAwsJson.kt` etc.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -0,0 +1,206 @@
...
+                "AwsJson11"
+            }
+            else -> throw IllegalStateException()
+        }
+        return RuntimeType(name, ServerCargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::protocols")
+    }
+
+    override fun routerType(): RuntimeType {
+        return RuntimeType("AwsJsonRouter", ServerCargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::routing::routers::aws_json")
+    }
```
david-perez, line 98:
```suggestion
    override fun routerType() = RuntimeType("AwsJsonRouter", ...
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -0,0 +1,206 @@
...
+
+    override fun routerConstruction(service: ServiceShape, operationValues: Iterable<Writable>, model: Model): Writable = writable {
+        val allOperationShapes = allOperationShapes(service, model)
+
+        // TODO(restore): This causes a panic: "symbol visitor should not be invoked in service shapes"
```
david-perez, line 103: I guess it's because we never had a need to generate a Rust type for a `service` shape. Now only the server does. We should track these in an issue to work on creating our own bottom-most symbol provider, tailored for the server codegen needs, now that we're starting to see how they need to deviate.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -0,0 +1,206 @@
...
+        // val serviceName = symbolProvider.toSymbol(service).name
+        val serviceName = service.id.name
+        val pairs = writable {
+            for ((operation, operationValue) in allOperationShapes.zip(operationValues)) {
+                val operationName = symbolProvider.toSymbol(operation).name
+                val key = "$serviceName.$operationName".dq()
```
david-perez, line 109: You can inline this where it's used and remove the `.dq()`; you can use double-quotes inside triple double-quotes.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -0,0 +1,206 @@
...
+                    "OperationValue" to operationValue,
+                    *codegenScope,
+                )
+            }
+        }
+        rustTemplate(
+            """
+            #{Router}::from_iter([#{Pairs:W}])
+            """,
+            "Router" to routerType(), "Pairs" to pairs,
```
david-perez, line 126:
```suggestion
            "Router" to routerType(),
            "Pairs" to pairs,
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
david-perez: I prefer this composition approach over the inheritance one we used over there though. Can you them into these ones, and split in several files?

And the factories should now return these ones e.g. `ServerRestJsonFactory`'s `protocol` method should return `ServerRestJson` etc...

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
david-perez: You can define one-line functions without creating a `{}` block like this. There are several places in the diff, especially in this file, where you could use this to reduce verbosity a bit.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -207,9 +209,31 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
                         )
                 }
             }
+
```
david-perez, line 212: nit: here and in line 223, these are empty lines, when we want them to be indented 12 spaces, so that the indent of the until the start of the block is constant.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -207,9 +209,31 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
...
+            impl<B> #{SmithyHttpServer}::request::FromRequest<#{Marker}, B> for #{I}
+            where
+                B: #{SmithyHttpServer}::body::HttpBody + Send,
+                B: 'static,
+                ${streamingBodyTraitBounds(operationShape)}
+                B::Data: Send,
+                #{RequestRejection} : From<<B as #{SmithyHttpServer}::body::HttpBody>::Error>
+            {
+                type Rejection = #{RuntimeError};
+                type Future = std::pin::Pin<Box<dyn std::future::Future<Output = Result<Self, Self::Rejection>> + Send>>;
```
david-perez, line 222: I guess we should define `Pin`, `Box`, `Future` in `RuntimeTypes.kt`

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/RustModule.kt
```diff
@@ -5,23 +5,27 @@
 
 package software.amazon.smithy.rust.codegen.rustlang
 
-data class RustModule(val name: String, val rustMetadata: RustMetadata, val documentation: String?) {
+data class RustModule(val name: String, val rustMetadata: RustMetadata, val documentation: String?, val hidden: Boolean?) {
```
david-perez, line 8: Whether the module is doc hidden or not should be indicated via the input `RustMetadata`. Example:

```
RustMetadata(additionalAttributes = listOf(Attribute.DocHidden))
```

And then line 14 takes care of rendering the doc hidden attribute.

rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/main.rs
```diff
@@ -47,10 +52,21 @@ pub async fn main() {
         // implementation.
         .into();
 
+    // New builder
```
david-perez, line 55: I think it's best we don't modify the public Pokémon Service example until we're ready to move people to the new builder and deprecate the old one.

We should instead create a Kotlin unit test that generates a simple model and uses the new builder instead. It could also assert that the new builder is generated as doc hidden in `lib.rs`.

rust-runtime/aws-smithy-http-server/src/build_modifier.rs
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use crate::operation::Operation;
+
+pub trait BuildModifier<Protocol, Op, S, L> {
```
david-perez, line 8: Missing Rust docs?

rust-runtime/aws-smithy-http-server/src/runtime_error.rs
```diff
@@ -54,6 +57,12 @@ pub struct RuntimeError {
     pub kind: RuntimeErrorKind,
 }
 
+impl<P> IntoResponse<P> for RuntimeError {
```
david-perez, line 60: Nice. Is there any blocker now to making `RuntimeError` generic over the protocol, and implement protocol-specific implementations of `IntoResponse` for them? It's not blocked on moving to the new builder right?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
david-perez: I'm not sure how to best merge them since inheritance has the advantage that we just override the method of the `Protocol` class that needs to return a different generator or that is subtly difference, and everything works correctly.

Maybe we need to use a combination of both: the truly server-specific methods go into this `ServerProtocol` interface, but then the classes that implement this interface should also inherit from `Protocol` to override things like e.g. `structuredDataSerializer`?

david-perez (commented):
Codegen looks good, just a few nits.

I guess the main two things to merge this are:
* A unit test instead of modifying the Pokémon service to use the new service builder; and
* Figuring out how best to merge the server protocol classes so we don't end up with duplicates, one implementing the `ServerProtocol` interface this PR introduces, and the others inheriting from `Protocol`...

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
```diff
@@ -66,6 +67,28 @@ open class ServerServiceGenerator(
...
+        }
+
+        // TODO(Temporary): Remove, this is temporary.
+        rustCrate.withModule(
+            RustModule.public("service", hidden = true),
+        ) { writer ->
+            val serverProtocol = ServerProtocol.fromCoreProtocol(coreCodegenContext, protocol)
+            ServerServiceGeneratorV2(
+                coreCodegenContext,
+                coreCodegenContext.serviceShape,
```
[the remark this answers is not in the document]
hlbarber: This was done in anticipation of having multiple services generated at the same time, in which case `CodegenContext` would not store a single `serviceShape`. As I understand it now, this is not really a goal. I'll revert it.

rust-runtime/aws-smithy-http-server/src/runtime_error.rs
hlbarber: After this is merged I think we could replace

```rust
#[derive(Debug)]
pub struct RuntimeError {
    pub protocol: Protocol,
    pub kind: RuntimeErrorKind,
}

impl<P> IntoResponse<P> for RuntimeError {
    fn into_response(self) -> http::Response<crate::body::BoxBody> {
        self.into_response()
    }
}

impl RuntimeError {
    pub fn into_response(self) -> Response {
        /* runtime match on enum */
```

with

```rust
#[derive(Debug)]
pub struct RuntimeError {
    pub kind: RuntimeErrorKind,
}

impl IntoResponse<RestJson1> for RuntimeError {
    fn into_response(self) -> http::Response<crate::body::BoxBody> {
        /* protocol specific implementation */
    }
}
```

I think this should still be superseded by https://github.com/awslabs/smithy-rs/issues/1703 where we do away with `RuntimeError` completely and have protocol specific enums and impls.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
hlbarber: Side note: I hope that this associated type will not exist as a `Pin<Box<_>>` for long - when we are landing https://github.com/awslabs/smithy-rs/issues/1706 we should be able to give this a concrete type.

rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/main.rs
```diff
@@ -47,10 +52,21 @@ pub async fn main() {
...
+    let get_pokemon_species = GetPokemonSpecies::from_handler(get_pokemon_species);
```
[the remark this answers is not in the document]
hlbarber: I think this would give an error saying that `Op` and `Exts` are unconstrained.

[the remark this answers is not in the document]
hlbarber: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d42bc87814ba44f09e5df4464f084fbd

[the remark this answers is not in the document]
hlbarber: This is a bit closer but `Ext` is still unconstrained unfortunately.

```rust
impl<Op, H, Exts> From<H> for Operation<IntoService<Op, H>>
where
    Op: OperationShape,
    H: Handler<Op, Exts>,
{
    fn from(h: H) -> Self {
        todo!()
    }
}
```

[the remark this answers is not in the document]
hlbarber: If `Exts` was an associated type on `Handler` then this would work and worth experimenting with. It might have some knock on effects elsewhere though.

Velfi (approved):
I only looked at the changes that affect the SDK. LGTM 😅

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -0,0 +1,376 @@
...
+                        router: self.router.map(|s| s.layer(layer))
+                    }
+                }
+            }
+
+            impl<B, RespB, S> #{Tower}::Service<#{Http}::Request<B>> for $serviceName<S>
+            where
+                S: #{Tower}::Service<#{Http}::Request<B>, Response = #{Http}::Response<RespB>> + Clone,
+                RespB: #{HttpBody}::Body<Data = #{Bytes}::Bytes> + Send + 'static,
+                RespB::Error: Into<#{Tower}::BoxError>
```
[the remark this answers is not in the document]
hlbarber: I've just inlined the alias. We have `type BoxError` in nearly all the runtime crates - it feels odd to me to `pub` when the others aren't.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
hlbarber: I've extended

```kotlin
    fun Protocol(runtimeConfig: RuntimeConfig) =
        RuntimeType("Protocol", ServerCargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::protocols")

```
in `codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRuntimeType.kt` with
```kotlin
    fun Protocol(name: String, runtimeConfig: RuntimeConfig) =
        RuntimeType(name, ServerCargoDependency.SmithyHttpServer(runtimeConfig), "${runtimeConfig.crateSrcPrefix}_http_server::protocols")

    fun Protocol(runtimeConfig: RuntimeConfig) = Protocol("Protocol", runtimeConfig)
```

Unresolve if you're not happy with this.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -0,0 +1,206 @@
...
+
+    // Returns the Rust marker struct enjoying `OperationShape`.
+    fun markerStruct(): RuntimeType
+
+    // Returns the Rust router type.
+    fun routerType(): RuntimeType
+
+    // Returns the construction of the `routerType` given a `ServiceShape`, a collection of operation values
+    // (`self.operation_name`, ...), and the `Model`.
+    // TODO(Decouple): Perhaps this should lean on a Rust interface.
```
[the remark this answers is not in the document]
hlbarber: I had a fleeting thought that perhaps there should be some sort of associated type `RoutingInfo` (which is `String` for `AwsJsonRouter` and `RequestSpec` for `RestRouter`.) on some trait which provides a constructor.

I don't think this gives us any benefits though - I'm going to delete this TODO.

rust-runtime/aws-smithy-http-server/src/build_modifier.rs
hlbarber: Removed this completely, will add it back in a corollary PR.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
hlbarber: Will track this effort/investigation here https://github.com/awslabs/smithy-rs/issues/1698

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
david-perez: https://github.com/awslabs/smithy-rs/issues/1724

david-perez (approved):

hlbarber:
Closes https://github.com/awslabs/smithy-rs/issues/1705

PR #1695, opened 2022-09-01
Add wiring to generate parameters in operations

rcoh:
## Motivation and Context

- #1637 
- Wire up the `params` in `operation` generator

## Description
Code generate params into `make_operation` generator. As a stop gap, we currently extract the region from this parameter and use it to drive the custom AWS resolver. The AWS resolver will be removed once we have endpoint rules for all services.

As a TLDR, this adds the following code to the body of make_operation:
```rust
        let _endpoint_params = crate::endpoint_resolver::Params::builder()
            /* region is a builtIn, other parameters will be injected here in the future */
            .set_region(_config.region.as_ref().map(|r| r.as_ref()))
            .build();

```
## Testing
- [x] codegen integration tests
- [x] added new codegen integration test
## Checklist

### DIFF
--- a/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
+++ b/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
@@ -12,6 +12,7 @@ import software.amazon.smithy.model.shapes.ServiceShape
 import software.amazon.smithy.model.shapes.ShapeId
 import software.amazon.smithy.rust.codegen.smithy.CoreCodegenContext
 import software.amazon.smithy.rust.codegen.smithy.RustCrate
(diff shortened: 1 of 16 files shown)


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/customize/RustCodegenDecorator.kt
```diff
@@ -70,6 +71,8 @@ interface RustCodegenDecorator<C : CoreCodegenContext> {
     fun transformModel(service: ServiceShape, model: Model): Model = model
 
     fun supportsCodegenContext(clazz: Class<out CoreCodegenContext>): Boolean
+
+    fun builtInResolvers(codegenContext: C): List<RulesEngineBuiltInResolver> = listOf()
```
jdisanti, line 75: Should be called something like `builtInEndpointResolvers`?

jdisanti (approved):

PR #1708, opened 2022-09-05
Test `httpRequestTests` against actual Services

weihanglo:
## Motivation and Context

https://github.com/awslabs/smithy-rs/issues/1212

## Description

Before this PR, we constructs `Request` manually to test against [`@httpRequestTests`] of protocol compliances tests. After this PR, we construct a tower service and turn it into a Router. 

That makes our test cases 

- exercise against a larger surface from routing to parameter validation. 
- avoids calling non-public method to exercise tests
- test against HTTP `method` field

[`@httpRequestTests`]: https://awslabs.github.io/smithy/1.0/spec/http-protocol-compliance-tests.html?highlight=httprequesttest#httprequesttests-trait

The implementation is quite simple. We add a helper module `renderTestHelper` which contains `fn validate_request(…)` sharing the logic of building a `Router` and exercise against each `@httpRequestTests`. We then put the assertion logic inside each operation handler. Some type coercions are needed inside `fn create_operation_registry_builder(…)` since we want a builder with blanket operations at first, and then set the actual operation handler for each test case.

## Testing

Run `./gradlew codegen-server-test:test` and see if all protocol tests are passed.

## Checklist


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -114,7 +136,83 @@ class ServerProtocolTestGenerator(
...
+
+                fn create_operation_registry_builder() -> RegistryBuilder {
+                    crate::operation_registry::$operationRegistryBuilderName::default()
+                        ${operations.mapIndexed { idx, operationShape ->
+                    val (inputT, outputT) = operationInputOutputTypes[operationShape]!!
+                    ".${operationNames[idx]}((|_| Box::pin(async { todo!() })) as F<$inputT, $outputT> )"
+                }.joinToString("\n")}
+                }
+
+                pub(crate) async fn validate_request(
```
david-perez, line 190:
```suggestion
                pub(crate) async fn build_router_and_make_request(
```

Request validation is usually associated with rejecting a request if it doesn't satisfy constraint traits.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -114,7 +136,83 @@ class ServerProtocolTestGenerator(
...
+                        .build()
+                        .expect("unable to build operation registry")
+                        .into();
+                    _ = router.into_make_service()
+                        .call(())
+                        .await
+                        .expect("unable to get a router")
+                        .call(http_request)
+                        .await
+                        .expect("unable to make an HTTP request");
```
david-perez, line 204:
```suggestion
                    let mut router: #{Router} = f(create_operation_registry_builder())
                        .build()
                        .expect("unable to build operation registry")
                        .into();
                    _ = router
                        .call(http_request)
                        .await
                        .expect("unable to make an HTTP request");
```

There's no need to convert into a `MakeService` just to call the service once :).

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -141,23 +239,25 @@ class ServerProtocolTestGenerator(
...
             )
             writer.withModule(testModuleName, moduleMeta) {
-                renderAllTestCases(allTests)
+                renderAllTestCases(operationShape, operationSymbol, allTests)
             }
         }
     }
 
-    private fun RustWriter.renderAllTestCases(allTests: List<TestCase>) {
+    private fun RustWriter.renderAllTestCases(operationShape: OperationShape, operationSymbol: Symbol, allTests: List<TestCase>) {
```
david-perez, line 247: These methods do not need to take in `operationSymbol`: they can recompute it from the `operationShape` using the `symbolProvider`.

In general methods should not take in things which they can recompute from the state (class properties) and input parameters.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -372,20 +481,45 @@ class ServerProtocolTestGenerator(
...
 
-    private fun checkParams(httpRequestTestCase: HttpRequestTestCase, rustWriter: RustWriter) {
-        rustWriter.writeInline("let expected = ")
-        instantiator.render(rustWriter, inputShape, httpRequestTestCase.params)
-        rustWriter.write(";")
+    private fun checkRequest(operationShape: OperationShape, operationSymbol: Symbol, httpRequestTestCase: HttpRequestTestCase, rustWriter: RustWriter) {
+        val inputShape = operationShape.inputShape(coreCodegenContext.model)
+        val outputShape = operationShape.outputShape(coreCodegenContext.model)
 
         val operationName = "${operationSymbol.name}${ServerHttpBoundProtocolGenerator.OPERATION_INPUT_WRAPPER_SUFFIX}"
```
david-perez, line 488: `operationName` is now unused.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -372,20 +481,45 @@ class ServerProtocolTestGenerator(
...
-        rustWriter.rustTemplate(
+        val (inputT, outputT) = operationInputOutputTypes[operationShape]!!
+        val helperNamespace = "crate::operation::protocol_test_helper"
```
david-perez, line 490: Consider extracting the string `protocol_test_helper`, since it's used too when defining the module. You also don't need to prefix with `crate::operation::`, since the module is defined in the same file as this chunk of code.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -372,20 +481,45 @@ class ServerProtocolTestGenerator(
...
+                &|builder| {
+                    builder.${operationShape.toName()}((|input| Box::pin(async move {
             """,
-            *codegenScope,
-        )
 
+            "})) as $helperNamespace::F<$inputT, $outputT>)}).await",
+
+        ) {
+            // Construct expected request
```
david-perez, line 502:
```suggestion
            // Construct expected request.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -372,20 +481,45 @@ class ServerProtocolTestGenerator(
...
+            rustWriter.writeInline("let expected = ")
+            instantiator.render(rustWriter, inputShape, httpRequestTestCase.params)
+            rustWriter.write(";")
```
david-perez, line 505:
```suggestion
            rustWriter.withBlock("let expected = ", ";") {
                instantiator.render(this, inputShape, httpRequestTestCase.params)
            }
```

This pattern is handy to not have to keep track of the trailing `;`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -522,21 +658,24 @@ class ServerProtocolTestGenerator(
...
-        )
-        rustWriter.writeWithNoFormatting(
-            """
-            assert_eq!(operation_extension.absolute(), format!("{}.{}", "${operationShape.id.namespace}", "${operationSymbol.name}"));
-            """.trimIndent(),
-        )
-    }
+    // We can't check that the `OperationExtension` is set in the response, because it is set in the implementation
+    // of the operation `Handler` trait, a code path that does not get exercised when we don't have a request to
+    // invoke it with (like in the case of an `httpResponseTest` test case).
```
david-perez, line 663: We could perform this check _after_ invoking the router with the request right? That is, above we construct:

```
let res = router
    .call(http_request)
    .await
    .expect("unable to make an HTTP request");

// Check that the extension is set on the response here.
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/Instantiator.kt
```diff
@@ -68,13 +69,19 @@ class Instantiator(
         val streaming: Boolean,
         // Whether we are instantiating with a Builder, in which case all setters take Option
         val builder: Boolean,
+        // If a given required field is missing, try to provide a default value.
```
david-perez, line 72:
```suggestion
        // Fill out `required` fields with a default value.
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/Instantiator.kt
```diff
@@ -267,16 +283,60 @@ class Instantiator(
...
         }
     }
+
+    /**
+     * Fill default values for missing required value of a shape.
+     */
+    private fun fillDefaultValue(shape: Shape): Node = when (shape) {
+        // Compound Shapes
+        is StructureShape -> Node.objectNode()
+        is UnionShape -> Node.objectNode()
```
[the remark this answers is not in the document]
weihanglo: Sure. Make sense! 

However, we still need to variant name for rendering, so the logic of getting the first variant would duplicate in two places.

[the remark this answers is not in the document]
weihanglo: Just realised that recursion doesn't happen here but `render`. One place to recurse is sufficient, no?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
weihanglo: Addressed with <https://github.com/awslabs/smithy-rs/pull/1708/commits/440dbae47be6be45119bf028112aa6b42cecd076>, though [this particular case](https://github.com/awslabs/smithy/blob/f598b87c51af5943686e38706847a5091fe718da/smithy-aws-protocol-tests/model/restJson1/http-labels.smithy#L254-L266) failed as it cannot find any extensions from the HTTP response. The HTTP status Code is a 404.

```
[rest_json/rust-server-codegen/src/operation.rs:877] http_response = Response {
    status: 404,
    version: HTTP/1.1,
    headers: {
        "content-type": "application/json",
        "x-amzn-errortype": "UnknownOperationException",
    },
    body: UnsyncBoxBody,
}
```

Haven't dug into details, but it might be a routing bug or something else.

weihanglo:
Most of the issues have been resolved. I marked minor refactoring comments as `resolve` and left other major issues unresolved. Thank you for your review, @david-perez!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
hlbarber: Can we take out the `OperationExtension` assertion entirely? 

1. It's inclusion is not part of the Smithy spec
2. It's not added in the deserialization error path
3. It's cannot be added in the routing failure error path
4. The panic from the `todo!`s ([and later `InternalFailure` + 500](https://github.com/awslabs/smithy-rs/pull/1709#issuecomment-1239186367)) signal that tests have routed to an unexpected operation.

david-perez: It's a harmless thing to assert. If the request was processed successfully (which all of the `httpRequestTests` should be), we expect the `OperationExtension` to be there.

hlbarber: I guess it signals that we **have** entered a specific operation.

An alternative, if we want to assert that a specific operation is hit then perhaps we can pass a sending end of a channel to the handler closure and then `assert!` we receive a signal at the end of the test.

In both approaches we should log the `http_response` so we get some insight into the exact error.

david-perez: > we should log the http_response so we get some insight into the exact error when deserialization/routing fails.

Yeah we should `debug!()` the response and the request.

weihanglo: With the latest two commits 2a5768c1b9671c43b911127a1cd1eac4c67f31cd 334269383fcfdfc42f332f10ae866a20929ab32d and #1719 from Harry, most of the issues of broken protocol tests have been fixed. Before https://github.com/awslabs/smithy/pull/1392 gets merged, we can proceed with temporary fixes.

david-perez (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/Instantiator.kt
```diff
@@ -222,14 +229,23 @@ class Instantiator(
      */
     private fun renderUnion(writer: RustWriter, shape: UnionShape, data: ObjectNode, ctx: Ctx) {
         val unionSymbol = symbolProvider.toSymbol(shape)
-        check(data.members.size == 1)
-        val variant = data.members.iterator().next()
-        val memberName = variant.key.value
+
+        val variant = if (ctx.defaultsForRequiredFields && data.members.isEmpty()) {
```
Velfi, line 233: It'd be nice to destructure here, maybe something like this?
```suggestion
        val (memberName, defaultValue) = if (ctx.defaultsForRequiredFields && data.members.isEmpty()) {
```

jdisanti (approved):
Client changes look good to me!

weihanglo:
Since this PR somehow blocks #1693 I'll merge it as is and do some follow-ups. Thank you all for the reviews!

PR #1709, opened 2022-09-05
Allow for unchecked service builders

hlbarber:
## Motivation and Context

Closes https://github.com/awslabs/smithy-rs/issues/1705

## Description

- Add a `PanicService` which implements `Service` and does nothing but panic.
- Add `DummyOperation` which implements `Upgradable`, yielding `UnreachableService`.
- Add `unchecked_builder` which sets all builder generics to `DummyOperation`.

## Notes

Depends on https://github.com/awslabs/smithy-rs/pull/1693

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
@@ -292,6 +292,23 @@ class ServerServiceGeneratorV2(
         }
     }
 
+    // Returns a `Writable` comma delimited sequence of `DummyOperation`.
+    private fun dummyGenerics(): Writable = writable {
+        for (index in 1..allOperationShapes.size) {
+            rustTemplate("#{SmithyHttpServer}::operation::DummyOperation,", *codegenScope)
+        }
+    }
+
+    // Returns a `Writable` comma delimited sequence of `builder_field: DummyOperation`.
+    private fun dummyFields(): Writable = writable {
+        for (fieldName in builderFieldNames()) {
+            rustTemplate(
+                "$fieldName: #{SmithyHttpServer}::operation::DummyOperation,",
+                *codegenScope,
+            )
+        }
+    }
+
     // Returns a `Writable` containing the service struct definition and its implementations.
     private fun struct(): Writable = writable {
         // Generate struct documentation.
@@ -312,6 +329,17 @@ class ServerServiceGeneratorV2(
                 pub fn builder() -> $builderName<#{NotSetGenerics:W}> {
                     $builderName {
                         #{NotSetFields:W}
+                        modifier: #{SmithyHttpServer}::build_modifier::Identity,
+                        _exts: std::marker::PhantomData
+                    }
+                }
+
+                /// Constructs a unchecked builder for [`$serviceName`].
+                ///
+                /// This will not enforce that all operations are set, however if an unset operation is used at runtime it will cause a panic.
+                pub fn unchecked_builder() -> $builderName<#{DummyGenerics:W}> {
+                    $builderName {
+                        #{DummyFields:W}
                         modifier: #{SmithyHttpServer}::build_modifier::Identity,
                         _exts: std::marker::PhantomData
                     }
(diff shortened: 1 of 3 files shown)


82marbag:
This is blocked on https://github.com/awslabs/smithy-rs/pull/1693 right?

hlbarber:
> This is blocked on #1693 right?

Yep, pointed to that branch for now. Will rebase after it's merged. Thanks for the label.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -316,6 +333,18 @@ class ServerServiceGeneratorV2(
                         _exts: std::marker::PhantomData
                     }
                 }
+
+                /// Constructs a unchecked builder for [`$serviceName`].
```
david-perez, line 337:
```suggestion
                /// Constructs an unchecked builder for [`$serviceName`].
```

rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/main.rs
```diff
@@ -63,6 +63,9 @@ pub async fn main() {
         .health_check_operation(health_check_operation)
         .build();
 
+    // Unchecked build does not type check whether or not an operation is set but will panic at runtime.
```
david-perez, line 66: We can showcase this in the lambda binary better. But for now this new service builder will not be announced, so a Kotlin unit test is better for now.

rust-runtime/aws-smithy-http-server/src/operation/upgrade.rs
```diff
@@ -264,4 +266,38 @@ where
...
+    type Response = http::Response<BoxBody>;
+    type Error = Infallible;
+    type Future = Ready<Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, _req: Request<B>) -> Self::Future {
+        panic!("this operation was not set")
```
david-perez, line 293: Will the user be able to identify to which operation the request was sent that they did not set?

david-perez (commented):
I'm thinking that panicking is perhaps too drastic a measure to take when receiving a request that was routed to an operation for which the user did not register a handler. Panicking is nice in test stages since it probably means the user forgot to register the operation handler, but consider what would happen in production. Anyone could take down a service if they happen to know how to (or just did some fuzz attack) hit the operation that that service did not register a handler for!

So perhaps we should just log an `error!()`.

hlbarber:
@david-perez I can add an alternative for the lambda folks which returns a safe fallback response perhaps?

```rust
let service = PokemonService::fallback_builder(|| Response::builder()./* create default response */).build();
```

Then perhaps rename this one to

```rust
let service = PokemonService::panic_builder().build();
```
?

rust-runtime/aws-smithy-http-server/src/operation/upgrade.rs
hlbarber: I'll add a constructor to the `DummyOperation` to pass an operation name to this `call`.

david-perez:
I don't think there's a use case for `panic_builder` nor `fallback_builder`. We can introduce them if there's user interest. For now I think logging the operation name as an `error!()` adequately fulfills the multi-lambda setup use case, which is what we set out to address with this patch.

hlbarber:
As is, the `build` procedure involves taking a `Op{N}`, using it's `Upgradable` implementation to promote it to a HTTP `tower::Service`, then boxing it (`Route::new`) and collecting it into a `RouterService`.

I don't think it's possible to get a request to route, emit a `error!(%operation)` but not respond with a fallback `http::Response` or `panic!` without changes to the routing mechanisms.

An approach that could get us this behavior is:

1. Change `Upgradable` to output an `Option` HTTP `tower::Service`, the `DummyOperation` would always return `None`.
2. Change the routers to collect `(RoutingInfo, Option<Route>)` instead of `(RoutingInfo, Route)`.
3. When a request is routed to a `None` then report `error!`.

hlbarber:
Summary of clarification: We should return an [InternalFailure](https://github.com/awslabs/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/resources/software/amazon/smithy/typescript/codegen/framework-errors.smithy#L5) and log an error.

hlbarber:
Closed, incorporated it into https://github.com/awslabs/smithy-rs/pull/1693 instead

Issue #976, opened 2021-12-14
remove doc[hidden] aliases from codegen


rcoh:
To pass information through between the operations and the fluent builders, pub types are created: https://github.com/awslabs/smithy-rs/blob/25def3a61287a126157d500829d5aff689e26ec4/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/protocol/ProtocolGenerator.kt#L246-L251

These are doc(hidden) and public. We should remove them, the have confused customers.

PR #1710, opened 2022-09-06
Remove the need for operation type aliasing in codegen

Velfi:
## Motivation and Context

#976 

## Description

This update reorganizes codegen in a small way so that we don't need to generate confusing operation and retry policy type aliases that we hide in the docs. Because they were already `#[doc(hidden)]`, I chose not to mention this change in the SDK changelog.

other changes:
- rename: `FluentClientGenerics.sendBounds` params to be more accurate
- update: `FlexibleClientGenerics.sendBounds` impl for readability
- update: type of `FluentClientGenerator` input param `retryPolicyType` to be `Any` with a default of `RustType.Unit`
- update: `PaginatorGenerator` to take `retryPolicy` as an input
- chore: fix some spelling and grammar issues
- remove: redundant `nextTokenEmpty` function from `PaginatorGenerator`
- add: `writable` property to `RustType` and `RuntimeType` that returns them in `Writable` form

## Testing

ran existing tests

## Checklist


codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
```diff
@@ -48,13 +48,15 @@ class PaginatorGenerator private constructor(
     service: ServiceShape,
     operation: OperationShape,
     private val generics: FluentClientGenerics,
+    retryPolicyType: Any = RustType.Unit,
```
Velfi, line 51: I'd make this a union type if I could but Kotlin doesn't support those and I couldn't find another way to say that it could be a `Symbol`, `RuntimeType`, `RustType`, `String`, or `GenericsGenerator`.

If y'all can think of a nice solution to that, I'll update the PR. The biggest thing is that I want to use the same `Templateable` type everywhere we template stuff and I can overload some `fun`s that only take one arg but other `fun`s take multiple args for writable things in no set order like my `rustTypeParameters` function and I don't want to change that to no longer be variadic e.g. I want to support this use case:

```kotlin
rustTemplate(
    "some_fn::<#{type_params:W}>();",
    "type_params" to rustTypeParameters(
        symbolProvider.toSymbol(operation),
        RustType.Unit,
        runtimeConfig.smithyHttp().member("body::SdkBody"),
        GenericsGenerator(GenericTypeArg("A"), GenericTypeArg("B")),
    )
)
```

CHANGELOG.next.toml
```diff
@@ -169,3 +169,11 @@ let response = client.some_operation()
 references = ["smithy-rs#1647", "smithy-rs#1112"]
 meta = { "breaking" = false, "tada" = true, "bug" = false, "target" = "client"}
 author = "Velfi"
+
+[[smithy-rs]]
+message = """
+Removed the need to generate operation output and retry aliases in codegen.
+"""
+references = ["smithy-rs#967"]
```
jdisanti, line 177:
```suggestion
references = ["smithy-rs#976", "smithy-rs#1710"]
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -74,7 +74,7 @@ private class AwsClientGenerics(private val types: Types) : FluentClientGenerics
     override val bounds = writable { }
 
     /** Bounds for generated `send()` functions */
-    override fun sendBounds(input: Symbol, output: Symbol, error: RuntimeType): Writable = writable { }
+    override fun sendBounds(operation: Symbol, operationOutput: Symbol, operationError: RuntimeType, retryPolicy: Any): Writable = writable { }
```
jdisanti, line 77: It's hard to tell from the diff, but it looks like only `RustType`s are ever passed into these `Any` args. What's forcing them to be `Any`?

Velfi: For whitelabel clients, the passed-in type is `RustType.Unit`

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
Velfi: For this specific arg, it's going to be a `RuntimeType` for all AWS clients and it's going to be `RustType.Unit` for "native" smithy clients.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
jdisanti: `RustType.Unit` is a `RustType` though?

Velfi: Ah, sorry, my brain rewrote your original comment to be about `RuntimeType`. `RuntimeType` and `RustType` are separate right?

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/PaginatorGenerator.kt
rcoh: a couple options: 
- take `Writeable` as an argument and make the caller deal with it
- make unit a `RuntimeType`: `val Unit = RuntimeType(namespace = "()")`
- make a type alias `Formattable = Any` but is clearly documented

jdisanti (approved):

PR #1711, opened 2022-09-07
Add `client` to `codegen-client` package namespace

jdisanti:
## Motivation and Context
I didn't want to include this in #1697 since it would make that diff a lot larger and more difficult to merge in. This adds `client` to the package names in the `codegen-client` module.


Velfi (approved):

hlbarber (approved):

david-perez:
Perhaps we should ignore this commit in blame? https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

jdisanti:
> Perhaps we should ignore this commit in blame? https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

I like this idea! Looks like I'll have to do this after merging this since the final squashed commit hash isn't known yet.

> Are we settling then on the convention that the namespace containing client or server is enough to distinguish to which project a class belongs to? In the server project we've also been prefixing most of our classes with Server to tell at a glance, so that readers don't have to inspect the classpath.

I wasn't intending to establish anything one way or another here, but I think this should be fine if you want to do it. The only downside is that it will be harder to find the class you want in an IDE when searching by name since you'll have multiple to choose from. I'm hoping we'll have less name collisions as we refactor away from an inheritance approach to a composition approach.

PR #1716, opened 2022-09-07
Add a join method for `Iterable<Writable>`, `Sequence<Writable>`, and `Array<Writable>`

hlbarber:
## Motivation and Context

For `Iterable<String>`, `Sequence<String>`, and `Array<String>` Kotlin provides a [joinToString](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/join-to-string.html) function which takes a separator and returns the concatenated `String` - it'd be nice if such an API existed for `Writable` too.

## Description

- Add `join` method for `Iterable<Writable>`, `Sequence<Writable>`, and `Array<Writable>`.
- Change `writable` definition to `writeInline` to prevent excess newlines from being introduced.

## Notes

We may want to add prefix/postfix arguments in a similar fashion to `joinToString`.

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
@@ -27,9 +27,13 @@ fun Writable.isEmpty(): Boolean {
 }
 
 /**
- * Helper allowing a `Iterable<Writable>` to be joined together using a separator.
+ * Helper allowing a `Iterable<Writable>` to be joined together using a `String` separator.
  */
 fun Iterable<Writable>.join(separator: String = ", ") = join(writable(separator))
+
+/**
+ * Helper allowing a `Iterable<Writable>` to be joined together using a `Writable` separator.
+ */
 fun Iterable<Writable>.join(separator: Writable = writable { }): Writable {
     val iter = this.iterator()
     return writable {
@@ -41,9 +45,25 @@ fun Iterable<Writable>.join(separator: Writable = writable { }): Writable {
         }
     }
 }
+
+/**
+ * Helper allowing a `Sequence<Writable>` to be joined together using a `String` separator.
+ */
 fun Sequence<Writable>.join(separator: String = ", ") = asIterable().join(separator)
+
+/**
+ * Helper allowing a `Sequence<Writable>` to be joined together using a `Writable` separator.
+ */
 fun Sequence<Writable>.join(separator: Writable = writable { }) = asIterable().join(separator)
+
+/**
+ * Helper allowing a `Array<Writable>` to be joined together using a `String` separator.
+ */
 fun Array<Writable>.join(separator: String = ", ") = asIterable().join(separator)
+
+/**
+ * Helper allowing a `Array<Writable>` to be joined together using a `Writable` separator.
+ */
 fun Array<Writable>.join(separator: Writable = writable { }) = asIterable().join(separator)
 
 /**
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
@@ -18,7 +18,7 @@ typealias Writable = RustWriter.() -> Unit
  */
 fun writable(w: Writable): Writable = w
 
-fun writable(w: String): Writable = writable { rust(w) }
+fun writable(w: String): Writable = writable { writeInline(w) }
 
 fun Writable.isEmpty(): Boolean {
     val writer = RustWriter.root()
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
@@ -27,6 +27,26 @@ fun Writable.isEmpty(): Boolean {
 }
 
 /**
+ * Helper allowing a `Iterable<Writable>` to be joined together using a separator.
+ */
+fun Iterable<Writable>.join(separator: String = ", ") = join(writable(separator))
+fun Iterable<Writable>.join(separator: Writable = writable { }): Writable {
+    val iter = this.iterator()
+    return writable {
+        iter.forEach { value ->
+            value()
+            if (iter.hasNext()) {
+                separator()
+            }
+        }
+    }
+}
+fun Sequence<Writable>.join(separator: String = ", ") = asIterable().join(separator)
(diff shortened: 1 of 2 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
```diff
@@ -18,14 +18,34 @@ typealias Writable = RustWriter.() -> Unit
  */
 fun writable(w: Writable): Writable = w
 
-fun writable(w: String): Writable = writable { rust(w) }
+fun writable(w: String): Writable = writable { writeInline(w) }
```
hlbarber, line 21: This is the most controversial change - it makes the asserts easier in the tests but might have knock-on effects.

rcoh: I doubt this method is used in a lot of places tbh, the codegen diff should show any changes

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
```diff
@@ -18,14 +18,54 @@ typealias Writable = RustWriter.() -> Unit
...
+                separator()
+            }
+        }
+    }
+}
+
+/**
+ * Helper allowing a `Sequence<Writable>` to be joined together using a `String` separator.
+ */
+fun Sequence<Writable>.join(separator: String = ", ") = asIterable().join(separator)
```
rcoh, line 52: this seems like it would be very confusing which one the compiler picked—sequence vs. iterable. One will have `,` but the other will have no separator

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
```diff
@@ -18,14 +18,54 @@ typealias Writable = RustWriter.() -> Unit
...
+
+/**
+ * Helper allowing a `Sequence<Writable>` to be joined together using a `Writable` separator.
+ */
+fun Sequence<Writable>.join(separator: Writable = writable { }) = asIterable().join(separator)
```
rcoh, line 57: hmm same here, I think this resolution is ambiguous?

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/rustlang/WritableTest.kt
```diff
@@ -75,4 +75,36 @@ internal class RustTypeParametersTest {
 
         writer.toString() shouldContain "'<()>'"
     }
+
```
rcoh, line 78: there are no tests of the ambiguous behavior when `join` is called with no args.

IMO, I'd remove the default arguments, I think they just will make things confusing

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/rustlang/WritableTest.kt
```diff
@@ -75,4 +75,36 @@ internal class RustTypeParametersTest {
...
+    @Test
+    fun `join iterable`() {
+        val writer = RustWriter.forModule("model")
+        listOf(writable("A"), writable("B"), writable("C")).join("-")(writer)
+        listOf(writable("D"), writable("E"), writable("F")).join(writable("+"))(writer)
+        writer.toString() shouldContain "A-B-CD+E+F"
```
rcoh, line 84: I'd probably make this test a bit more realistic, eg. make a `vec![...]` out of a series of writeables.

rcoh (changes requested):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
hlbarber: whoops, yeah - this should be `writable(", ")`.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/rustlang/WritableTest.kt
hlbarber: The default separator was added to mirror https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/join-to-string.html. Will remove it.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
Velfi: I added a similar templating fn (`rustInlineTemplate`) in my recent PR for the same reason (easy asserts)

Velfi (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
hlbarber: @Velfi shall I undo this and use that in my asserts instead?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/rustlang/Writable.kt
```diff
@@ -18,14 +18,54 @@ typealias Writable = RustWriter.() -> Unit
...
+    val iter = this.iterator()
+    return writable {
+        iter.forEach { value ->
+            value()
+            if (iter.hasNext()) {
+                separator()
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
hlbarber: On a related note, I was actually thinking about using the `when` pattern you've used in `rustTypeParameters` rather than using overload resolution to allow for `String` and `Writable` argument.

The reason being, if we wanted to add an `prefix`/`suffix` arguments, in addition to separator, we'd have to write a huge number of different function signatures to allow them to also switch between `Writable` and `String` independently. In contrast, we can just `when` for each argument in one function.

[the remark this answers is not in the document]
hlbarber: I was even thinking of a `ToWritable` interface, which `String`, `RuntimeTypes`, and `Writable` etc satisfies.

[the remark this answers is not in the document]
hlbarber: > I think it could possibly be improved by using this.

https://github.com/awslabs/smithy-rs/compare/harryb/iterable-writable...harryb/shrink-rust-type-params

rcoh (approved):

Issue #1715, opened 2022-09-07
No retry for native Smithy clients


jdisanti:
The native (non-AWS SDK) clients generated by smithy-rs for currently use `()` as the retry classifier, which results in no retry.

jdisanti:
This has been fixed and released.

PR #1717, opened 2022-09-07
Fix native Smithy client retry and retry response errors

jdisanti:
## Motivation and Context
This PR addresses  https://github.com/awslabs/aws-sdk-rust/issues/303 and #1715.

## Checklist


jdisanti:
Need to decide what to do about the "retry policy" to "response classifier" rename. The "response classifier" name is a lot more accurate, but the rename goes deep since the operation `retry_policy` and `with_retry_policy` functions will also need to be renamed, and examples fixed:
https://github.com/awsdocs/aws-doc-sdk-examples/blob/3b8c70b319882a40a03aedfcd5383b6851c9dd76/rust_dev_preview/dynamodb/src/bin/movies.rs#L302-L319

rcoh:
> That example also mentions "retry every second until the table is ready" in its comments, but I highly doubt that's what's actually happening. It's probably using exponential backoff.

It's using `RetryKind::Explicit(Duration::from_secs(1))`, so that is accurate at least.

I'd recommend doing the rename, adding new methods, deprecating the old ones, and deleting them in a future release to simplify coordination

aws/rust-runtime/aws-http/src/retry.rs
```diff
@@ -39,41 +28,38 @@ const THROTTLING_ERRORS: &[&str] = &[
...
-                    RetryKind::Error(ErrorKind::TransientError)
-                } else if let Some(ek) = err.is_other() {
-                    RetryKind::Error(ek)
-                } else {
-                    RetryKind::UnretryableFailure
-                };
-            }
-            Err(_) => return RetryKind::UnretryableFailure,
+    fn classify(&self, result: Result<&T, &SdkError<E>>) -> RetryKind {
+        let (err, response) = match DefaultResponseClassifier::try_extract_err_response(result) {
```
Velfi, line 60: Could you add a comment here and on line 63 about how this is the basic retry classification and how the AWS-specific retry logic is the fallback? I was a bit confused as to why all logic wasn't encompassed by the `try_extract_err_response` function.

Velfi:
@jdisanti Will you be submitting an examples update PR as well?

CHANGELOG.next.toml
```diff
@@ -186,3 +186,21 @@ message = "Smithy IDL v2 mixins are now supported"
 references = ["smithy-rs#1680"]
 meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "all"}
 author = "ogudavid"
+
+[[smithy-rs]]
+message = "Generated clients now retry transient errors without replacing the retry policy."
```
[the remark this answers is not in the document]
jdisanti: White label/adhoc AWS clients already had retry. This is specifically fixing for Smithy clients. I think this should be clear by it being a `smithy-rs` entry.

aws/rust-runtime/aws-http/src/retry.rs
```diff
@@ -39,41 +28,38 @@ const THROTTLING_ERRORS: &[&str] = &[
...
+/// 1. The `x-amz-retry-after` header is checked
+/// 2. The modeled error retry mode is checked
+/// 3. The code is checked against a predetermined list of throttling errors & transient error codes
+/// 4. The status code is checked against a predetermined list of status codes
+#[non_exhaustive]
+#[derive(Clone, Debug)]
+pub struct AwsResponseClassifier;
+
+impl AwsResponseClassifier {
+    /// Create an `AwsResponseClassifier` with the default set of known error & status codes
```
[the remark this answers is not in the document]
jdisanti: If there is a way to link to the private constants with rustdoc, then I could add this pretty easily. I'm reluctant to duplicate the list in the documentation since it will get out of date.

jdisanti:
> @jdisanti Will you be submitting an examples update PR as well?

Yeah.

jdisanti:
Examples fixed in https://github.com/awsdocs/aws-doc-sdk-examples/pull/3593

aws/rust-runtime/aws-http/src/retry.rs
```diff
@@ -39,41 +28,42 @@ const THROTTLING_ERRORS: &[&str] = &[
 ];
 const TRANSIENT_ERRORS: &[&str] = &["RequestTimeout", "RequestTimeoutException"];
 
-impl AwsErrorRetryPolicy {
-    /// Create an `AwsErrorRetryPolicy` with the default set of known error & status codes
+/// Implementation of [`ClassifyRetry`] that models AWS error codes.
```
Velfi, line 31: Consider a more descriptive verb as an alternative to "models"

aws/sdk/integration-tests/dynamodb/tests/movies.rs
```diff
@@ -228,37 +171,13 @@ async fn movies_it() {
...
-        .call(
-            movies_in_year(table_name, 2222)
-                .make_operation(&conf)
-                .await
-                .expect("valid request"),
-        )
-        .await
-        .expect("query should succeed");
+    let films_2222 = movies_in_year(&client, table_name, 2222).await;
     // this isn't back to the future, there are no movies from 2022
```
Velfi, line 177:
```suggestion
    // this isn't "Back To The Future", there are no movies from 2222
```

rust-runtime/aws-smithy-http-tower/src/lib.rs
```diff
@@ -102,7 +103,10 @@ mod tests {
         });
 
         let mut svc = ServiceBuilder::new()
-            .layer(ParseResponseLayer::<TestParseResponse, ()>::new())
+            .layer(ParseResponseLayer::<
+                TestParseResponse,
```
Velfi, line 107: Consider calling this "TestResponseHandler" or, instead, changing the verbiage for "response handler" to "response parser"

Velfi (approved):

rust-runtime/aws-smithy-http-tower/src/lib.rs
jdisanti: I think the name is actually fine considering it implements `ParseStrictResponse`. I think `handler` might actually be the out of place name that should potentially get fixed in other places.

PR #1719, opened 2022-09-08
Escape routing regex in path segment literals

hlbarber:
## Motivation and Context

When constructing the regex to match routes we should escape path segments deemed as literals.

This is covered by the  [HttpRequestWithRegexLiteral protocol test](https://github.com/awslabs/smithy/blob/f598b87c51af5943686e38706847a5091fe718da/smithy-aws-protocol-tests/model/restJson1/http-labels.smithy#L254-L266)

This has caused problems [here](https://github.com/awslabs/smithy-rs/pull/1708#discussion_r965846689).

## Description

Escape `PathSegment::Literal`s so that they don't contribute to the routing regex.

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
+++ b/rust-runtime/aws-smithy-http-server/src/routing/request_spec.rs
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
LGTM. Thanks for the quick fix!

PR #1722, opened 2022-09-08
chore: upgrade dokka gradle plugin

eduardomourar:
## Motivation and Context

Due to an issue between dokka and gradle (describe [here](https://github.com/Kotlin/dokka/issues/1834)), the build is failing on my machine.

## Description

The dokka gradle plugin has been upgraded to v1.4.30 as suggested in that GitHub issue and the changes in the generated documentation should be minimal.

## Testing

I have been able to generate the documentation successfully.

## Checklist


jdisanti:
Hey there! Thank you for the contribution!

As far as I'm aware, the smithy-rs maintainers don't make use of the Dokka documentation for any of the Kotlin modules, which is probably why the newer codegen-server modules don't have it. As such, I'm leaning more towards removing it entirely from the build scripts. I want to understand better where you're coming from. What led you to try building these docs, and do you find the output useful for that goal?

eduardomourar:
I do check the generated documentation from time to time to visually browse the code (although having them published like the [smithy javadoc](https://awslabs.github.io/smithy/javadoc/lastest/) would be really nice). In our particular case, we are publishing the smithy rust modules (more specifically `codegen-core`, `codegen-client`, `codegen-server` and `rust-runtime`) to a private Maven registry and this was causing issues while preparing the package in GitHub actions (as well as in my local machine).

My suggestion here is to keep it and let it be generate by default at the root project in order to simplify configuration the subprojects.

build.gradle.kts
```diff
@@ -16,14 +16,15 @@ buildscript {
 
 plugins {
     kotlin("jvm") version "1.3.72" apply false
-    id("org.jetbrains.dokka") version "0.10.0"
+    id("org.jetbrains.dokka") version "1.4.30"
```
jdisanti, line 19: Why this older version of Dokka rather than the latest?

.gitignore
```diff
@@ -27,6 +27,8 @@ hs_err_pid*
 # Gradle
 .gradle
 build/
+**/bin/main/
+**/bin/test/
```
jdisanti, line 31: What is this for?

build.gradle.kts
```diff
@@ -16,14 +16,15 @@ buildscript {
...
 }
 
 allprojects {
     repositories {
         mavenLocal()
         mavenCentral()
         google()
+        maven("https://maven.pkg.jetbrains.space/public/p/kotlinx-html/maven") // https://github.com/Kotlin/kotlinx.html/issues/173
```
jdisanti, line 27: Seems like this might not be necessary with a newer Dokka version.

build.gradle.kts
eduardomourar: I will confirm if that works as expected with v1.7.10 and will push the change accordingly

.gitignore
eduardomourar: A lot of files are put into the bin folder in each subproject whenever I run gradle build (I believe that is a VSCode thing as shown here: https://github.com/microsoft/vscode-gradle/blob/develop/.gitignore)

build.gradle.kts
eduardomourar: it works well, although the documentation layout is significantly different.

jdisanti (approved):
Looks good! Thank you!

Issue #1698, opened 2022-09-02
Separate Protocol class into CoreProtocol, ServerProtocol, and ClientProtocol


hlbarber:
At the time of writing, there is a single `Protocol` class which holds both client and server methods, despite it living inside the `codegen` folder. The `codegen` folder is supposed to be server agnostic.

We should separate `Protocol` interface into `CoreProtocol`, `ServerProtocol`, and `ClientProtocol` interfaces.

- `CoreProtocol` holds methods common to both the server and client.
- `ServerProtocol` holds an instance of `CoreProtocol` and lives in the `codegen-server`.
- `ClientProtocol` should work similarly to `ServerProtocol` - encapsulating all client specific methods and living in an appropriate space.

As a result, the classes implementing `Protocol` should also be separated. For example:

- `class CoreRestJsonProtocol: CoreProtocol`
- `class ServerRestJsonProtocol: ServerProtocol`
-  `class ClientRestJsonProtocol: ClientProtocol`

and methods should be added to `Server{Protocol}Protocol` and `Client{Protocol}Protocol` to retrieve `Core{Protocol}Protocol`.

hlbarber:
Progress towards this goal: https://github.com/awslabs/smithy-rs/pull/1697

hlbarber:
The `ServerProtocol` class has been introduced in https://github.com/awslabs/smithy-rs/pull/1693. Further work needs to be done to move existing methods from the implementations of `Protocol` onto the implementations of `ServerProtocol`.

hlbarber:
> and methods should be added to Server{Protocol}Protocol and Client{Protocol}Protocol to retrieve Core{Protocol}Protocol.

Rather than have this be the case, it's probably more ergonomic to have `ServerProtocol : ClientProtocol` and `Server{Protocol}Protocol : Client{Protocol}Protocol`.

Changes have been made in https://github.com/awslabs/smithy-rs/pull/1693 to do this for the server side.

PR #1723, opened 2022-09-09
Unsupported content type

82marbag:
## Motivation and Context

Add validation for the Content-Type header and pass (remove from the failing list) the relevant protocol tests.

Closes: https://github.com/awslabs/smithy-rs/issues/1663

## Testing

`./gradlew :codegen-server-test:test && ./gradlew :codegen-test:test`

## Checklist


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -699,6 +724,17 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
         val err = if (StructureGenerator.fallibleBuilder(inputShape, symbolProvider)) {
             "?"
         } else ""
+        if (noInputs) {
+            val emptyBody = if (parser != null) { "body.is_empty()" } else { "true" }
+            rustTemplate(
+                """
+                #{SmithyHttpServer}::protocols::content_type_header_empty_body_no_modeled_input(
```
david-perez, line 731: I think we should split this code. There should be two functions, each handling the case of whether the body is empty or not.

david-perez (changes requested):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -180,6 +184,30 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
                     *codegenScope,
                 )
             }
+            val inputMembers = operationShape
+                .inputShape(model)
+                .members()
```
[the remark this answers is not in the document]
82marbag: Thanks. The `break` in the loop was doing the same, but this is cleaner

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -20,35 +20,76 @@ pub struct AwsJson10;
...
+pub fn content_type_header_empty_body_no_modeled_input<B>(
     req: &RequestParts<B>,
-    expected_mime: &'static mime::Mime,
+    empty_body: bool,
+    protocol: Protocol,
 ) -> Result<(), MissingContentTypeReason> {
-    let found_mime = req
-        .headers()
-        .ok_or(MissingContentTypeReason::HeadersTakenByAnotherExtractor)?
+    if req.headers().is_none() || (protocol != Protocol::RestJson1 && protocol != Protocol::RestXml) {
```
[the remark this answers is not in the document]
82marbag: No, it's just that a client should always send an empty JSON object payload in AwsJson 1.0, so with `{}` in the body this would fail (this only checks whether the body is empty)

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -20,35 +20,76 @@ pub struct AwsJson10;
...
+        Err(MissingContentTypeReason::UnexpectedMimeType {
+            expected_mime: None,
+            found_mime: Some(found_mime),
+        })
+    } else {
+        Ok(())
+    }
+}
+
+pub fn content_type_header_classifier<B>(
```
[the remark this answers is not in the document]
82marbag: Added docs explaining what they do, but neither panic

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
82marbag: But when the body is not empty, it's just `Ok(())`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -667,16 +695,13 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
         Attribute.AllowUnusedMut.render(this)
         rust("let mut input = #T::default();", inputShape.builderSymbol(symbolProvider))
         val parser = structuredDataParser.serverInputParser(operationShape)
+        val noInputs = model.expectShape(operationShape.inputShape).expectTrait<SyntheticInputTrait>().originalId == null
```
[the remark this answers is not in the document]
82marbag: This is on purpose because of [this test](https://github.com/awslabs/smithy/blob/f8968b9462db03e858b20938c7e82c0c5ac39ac0/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy#L199). It's the first example you give, but has a non-empty body and a content-type

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/Protocol.kt
```diff
@@ -90,6 +90,13 @@ interface Protocol {
      * protocol.
      */
     fun serverRouterRuntimeConstructor(): String
+
+    /**
+     * In some protocols, such as RestJson,
+     * when there is no modeled body input, content type must not be set and the body must be empty
+     * Returns a writable to perform this check
+     */
+    fun serverContentTypeCheckNoModeledInput(): Boolean = false
```
jdisanti, line 99: I don't want to block approval on this, but now that the prerequisite work has been done in #1698, it would be better to have this in the `ServerProtocol` implementations instead.

jdisanti (approved):

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -20,35 +20,81 @@ pub struct AwsJson10;
...
         .parse::<mime::Mime>()
         .map_err(MissingContentTypeReason::MimeParseError)?;
-    if &found_mime == expected_mime {
-        Ok(())
+    // There is a content-type header
+    // If there is an implied content type, they must match
+    if let Some(expected_content_type) = expected_content_type {
+        let content_type = expected_content_type
+            .parse::<mime::Mime>()
+            /* `expected_content_type` comes from the codegen */
```
david-perez, line 82:
```suggestion
            // `expected_content_type` comes from the codegen.
```

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/Protocol.kt
```diff
@@ -90,6 +90,13 @@ interface Protocol {
      * protocol.
      */
     fun serverRouterRuntimeConstructor(): String
+
+    /**
+     * In some protocols, such as RestJson,
+     * when there is no modeled body input, content type must not be set and the body must be empty
+     * Returns a writable to perform this check
```
david-perez, line 97:
```suggestion
     * In some protocols, such as restJson1,
     * when there is no modeled body input, content type must not be set and the body must be empty.
     * Returns a boolean indicating whether to perform this check.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -696,6 +723,19 @@ private class ServerHttpBoundProtocolTraitImplGenerator(
         serverRenderUriPathParser(this, operationShape)
         serverRenderQueryStringParser(this, operationShape)
 
+        if (noInputs && protocol.serverContentTypeCheckNoModeledInput()) {
+            val emptyBody = if (parser != null) { "body.is_empty()" } else { "true" }
+            rustTemplate(
+                """
+                if $emptyBody {
```
david-perez, line 730: Clippy does not flag this `if true`?

I'd rewrite this to avoid it. You can use `conditionalBlock("if body.is_empty() {", "}", conditional = parser != null) { ... }` for this purpose.

david-perez (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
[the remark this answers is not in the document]
david-perez: Ah I see. Interesting. What's the protocol test that asserts that body must be empty when there is no modeled input? I can't seem to find it right now.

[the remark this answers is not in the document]
82marbag: [This one](https://github.com/awslabs/smithy/blob/603c0f2de0dc54c7ed5cfa0d1e2f1c63d521eed2/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-content-type.smithy#L12) (which corresponds to the second example you give)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
82marbag: Thanks!

codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/Protocol.kt
hlbarber: After this is merged I'll move this in https://github.com/awslabs/smithy-rs/pull/1731

PR #1730, opened 2022-09-13
Move most utils and synthetic traits into core

jdisanti:
## Motivation and Context
As part of refactoring more code into the core module, this PR:
- Moves most of the `util` package from `codegen-client` into `codegen-core`
- Moves the synthetic traits into `codegen-core`
- Consolidates duplicated `RustTypesTest` files
- Corrects `lang` -> `rustlang` package name in `codegen-client` tests
- Moves `letIf` from `SymbolVisitor` into its own file in `codegen-core`

The `Synthetics` util was left in `codegen-client` since its unit test requires `testutil`, and `testutil` requires dozens more files from `codegen-client`. Pulling it in would easily quadruple the size of this PR.


hlbarber:
It looks like `Writable.kt` and a few other modules have not been moved from the `rustlang` folder in `client` to `core`. Is this intended?

Velfi (approved):

jdisanti:
> Should we move everything in the rustlang folder into core?

Yes, we should. Not in this PR though :smile: I'm trying to keep the changes smaller and easier to review. I attempted to move `rustlang`, but that ended up requiring a lot of `smithy` as well, so I decided to just stick with `util` for this PR.

hlbarber (approved):

crisidev (approved):
Looks good!

Issue #1582, opened 2022-07-27
Errors Tracking


Velfi:
This issue tracks the improvements we want to make to errors in smithy-rs and the SDKs during Q3

- [ ] [Error types should not implement source _and_ print inner error in Display](https://github.com/awslabs/smithy-rs/issues/1345)
- [ ] [Better error context](https://github.com/awslabs/smithy-rs/issues/1568)

jdisanti:
Closing this tracking issue in favor of #1692.

PR #1733, opened 2022-09-14
Various fixes and updates for `aws-smithy-http` including errors and `impl futures::Stream for SdkBody`

Velfi:
## Motivation and Context

I started this because of #1582 but I added more changes.

## Description

This fixes several small issues I had with `aws-smithy-http` including:
add: futures::Stream impl for SdkBody
update: use futures-lite instead of futures-core
remove: callbacks API for SdkBody and ByteStream
remove: Error::source impl for byte_stream::Error
update: hyper version

## Testing

wrote 2 tests and ran existing tests

## Checklist


aws/rust-runtime/aws-config/Cargo.toml
```diff
@@ -26,7 +26,7 @@ aws-types = { path = "../../sdk/build/aws-sdk/sdk/aws-types" }
 time = { version = "0.3.4", features = ["parsing"] }
 tokio = { version = "1.8.4", features = ["sync"] }
 tracing = { version = "0.1" }
-hyper = { version = "0.14.12", default-features = false }
+hyper = { version = "0.14.20", default-features = false }
```
jdisanti, line 29: Why was the hyper upgrade necessary?

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -323,13 +264,47 @@ impl http_body::Body for SdkBody {
...
         }
     }
 }
 
+const SIZE_HINT_32_BIT_PANIC_MESSAGE: &str = r#"
+You're running a 32-bit system and this stream's length is too large to be represented with a usize.
+Please limit stream length to less than 4.294Gb or run this program on a 64-bit computer architecture.
+"#;
+
+impl Stream for SdkBody {
```
jdisanti, line 279: The reasoning for not doing this up until now was that [`AsyncIterator`](https://doc.rust-lang.org/std/async_iter/trait.AsyncIterator.html) was close to becoming stabilized in `std`, and this will be obsolete as soon as that happens. This is definitely worth discussing further since it looks like that will take longer due to discussion of generic keywords. Maybe we should create another runtime crate similar to `aws-smithy-types-convert` that adds in useful conversions to the current set of async traits that are widely used until they become `std`.

There are a couple of crates that need to get revisited as we get closer to a stable release:
- https://github.com/awslabs/smithy-rs/blob/cbbbec4a4892e301ebde3af5cc852aa48d82e608/rust-runtime/aws-smithy-async/external-types.toml#L2
- https://github.com/awslabs/smithy-rs/blob/cbbbec4a4892e301ebde3af5cc852aa48d82e608/rust-runtime/aws-smithy-http/external-types.toml#L23

aws/rust-runtime/aws-config/Cargo.toml
Velfi: It wasn't, I just saw that the server folks were using `.20` so I figured I'd update our stuff too. I can remove this if you like.

jdisanti: Follow the RFC on this: https://awslabs.github.io/smithy-rs/design/rfcs/rfc0021_dependency_versions.html

PR #1734, opened 2022-09-14
[Python] Allow to run pure Python request middlewares inside a Tower service

crisidev:
## Motivation and Context
* Customers want to be able to implement simple middleware directly in Python. This PR aims to add the initial support for it.
* Improve the idiomatic experience of logging by exposing a handler compatible with Python's stardard library `logging` module.

### Middleware
A middleware is defined as a sync or async Python function that can return multiple values, following these rules:

* Middleware not returning will let the execution continue without changing the original request.
* Middleware returning a modified Request will update the original request before continuing the execution.
* Middleware returning a Response will immediately terminate the request handling and return the response constructed from Python.
* Middleware raising MiddlewareException will immediately terminate the request handling and return a protocol specific error, with the option of setting the HTTP return code.
* Middleware raising any other exception will immediately terminate the request handling and return a protocol specific error, with HTTP status code 500.

Middlewares are registered into the Python application and executed in order of registration.

Example:

```python
from sdk import App
from sdk.middleware import Request, MiddlewareException

app = App()

@app.request_middleware
def inject_header(request: Request):
    request.set_header("x-amzn-answer", "42")
    return request

@app.request_middleare
def check_header(request: Request):
    if request.get_header("x-amzn-answer") != "42":
        raise MiddlewareException("Wrong answer", 404)

@app.request_middleware
def dump_headers(request: Request):
    logging.debug(f"Request headers after middlewares: {request.headers()}")
```

**NOTE: this PR only adds support for request middlewares, which are executed before the operation handler. Response middlewares, executed after the operation are tracked here: https://github.com/awslabs/smithy-rs/issues/1754.**

### Logging
To improve the idiomatic experience, now logging need to be configured from the Python side by using the standard `logging` module. This allows customers to opt-out of our `tracing` based logging implementation and use their own and logging level is now driven by Python.

```python
import logging
from sdk.logging import TracingHandler

logging.basicConfig(level=logging.INFO, handlers=[TracingHandler.handle()])
```

## Checklist


rust-runtime/aws-smithy-http-server-python/src/server.rs
```diff
@@ -73,17 +77,15 @@ pub trait PyApp: Clone + pyo3::IntoPy<PyObject> {
                     .getattr(py, "pid")
                     .map(|pid| pid.extract(py).unwrap_or(-1))
                     .unwrap_or(-1);
-                tracing::debug!("Terminating worker {idx}, PID: {pid}");
+                println!("Terminating worker {idx}, PID: {pid}");
```
crisidev, line 80: These tracing statements have been moved to `println!` since at this stage we do not have the tracing subscriber initialized.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerOperationHandlerGenerator.kt
```diff
@@ -90,16 +99,14 @@ class PythonServerOperationHandlerGenerator(
...
-                        output.extract::<$output>()
-                    })
+                #{pyo3}::Python::with_gil(|py| {
+                    let pyhandler: &#{pyo3}::types::PyFunction = handler.extract(py)?;
+                    let output = if handler.args == 1 {
+                        pyhandler.call1((input,))?
+                    } else {
+                        pyhandler.call1((input, state.0))?
+                    };
+                    output.extract::<$output>()
```
crisidev, line 109: The reason for removing `tokio::task::block_in_place` is that we are still bound to the mutex on the GIL that is taken inside the `Python::with_gil` closure. 

No matter what we do, this synchronous mutex will not allow tokio to continue the execution.

Removing `block_in_place` yields a significant performance improvement since it removes the overhead of scheduling new tasks.

crisidev: Other frameworks using the same approach as Smithy-rs Python are also doing the same: https://github.com/sansyrox/robyn/commit/4ef01e60262e1455fa564b06af9b1017deecba1c

rust-runtime/aws-smithy-http-server-python/src/middleware/layer.rs
```diff
@@ -0,0 +1,248 @@
...
+    type Response = Response<BoxBody>;
+    type Error = S::Error;
+    type Future = ResponseFuture<S>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let inner = self.inner.clone();
```
unexge, line 112: Note about making this clone less expensive (Arc?)

hlbarber (commented):
Is it the case that this will be less expressive than a general `tower::Layer`, even after the response side middleware is also added?

A general `tower::Layer`, or more concretely the `struct Middleware<S> { inner: S, ... }` it applies around a `S`, has the ability to hold onto state from the request side to the response side.

For example, I might pop a request header, hold onto it while the `S` does some handling, and then insert it as a response header. Is this a class of behavior we want to support?

rust-runtime/aws-smithy-http-server-python/src/middleware/layer.rs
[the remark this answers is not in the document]
crisidev: Interesting, I did not know this.

crisidev:
> Is it the case that this approach will be less expressive than a general `tower::Layer`, even after the response side middleware is also implemented?
> 
> A general `tower::Layer`, or more concretely the `struct Middleware<S> { inner: S, ... }` it applies around a `S`, has the ability to hold onto state from the request side to the response side.
> 
> For example, I might pop a request header, hold onto it while the `S` does some handling, and then insert it as a response header. Is this a class of behavior we want to support?
> 
> You can see this kind of middleware in Django too https://docs.djangoproject.com/en/4.1/topics/http/middleware/#writing-your-own-middleware
> 
> If this were possible to support then request/response middleware would then become a subclass of this more general middleware.

This is an initial implementation of a subset of the features I would like to see available in middleware handling, but there are certain features that will be either extremely hard or not possible to do with PyO3: for example there is no way to store generics inside a Python compatible structure. I will handoff this to @unexge after this initial support for middleware is done. I think after it we can sit down together and discuss possible improvement and a more general implementation. 

FYI, I would like to see most if not all the middleware we vend to customers to be implemented directly in Rust for performance and safety reasons, but I think it would be nice for customers to be able to implement simple middleware directly in Python.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerOperationHandlerGenerator.kt
```diff
@@ -69,11 +69,20 @@ class PythonServerOperationHandlerGenerator(
...
+                    let span = #{tracing}::span!(
+                        #{tracing}::Level::TRACE, "python",
+                        pid = #{tracing}::field::Empty,
+                        module = #{tracing}::field::Empty,
+                        filename = #{tracing}::field::Empty,
+                        lineno = #{tracing}::field::Empty
+                    );
+                    let guard = span.enter();
                     let result = if handler.is_coroutine {
                         #{PyCoroutine:W}
```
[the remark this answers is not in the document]
crisidev: What I did here is totally wrong and doesn't work as expected. The await point doesn't really bother from my tests, but the span needs to be created inside the tracing python wrapper to have the right info like filename, line number, etc..

Next commit will remove this from here and move it in another place where it will also resolve the holding the span over the await point.

rust-runtime/aws-smithy-http-server-python/src/middleware/request.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl PyRequest {
+    /// Create a new Python-compatible [Request] structure from the Rust side.
+    ///
+    /// This is done by cloning the headers, method, URI and HTTP version to let them be owned by Python.
+    pub fn new(request: &Request<Body>) -> Self {
+        Self {
+            method: request.method().to_string(),
+            uri: request.uri().to_string(),
+            headers: request
```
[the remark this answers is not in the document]
crisidev: Done

rust-runtime/aws-smithy-http-server-python/src/middleware/layer.rs
crisidev: Done

unexge (approved):

PR #1736, opened 2022-09-15
Allow third-party devs to write custom plugins for service builders

82marbag:
## Motivation and Context

With the new service and service builder types, it is verbose to add layers to add middleware. Third-party developers can't write an ergonomic middleware to add to services.

## Description

With this change, third-parties can write middleware that add a function to service builder,
which smithy service owners can incorporate by calling that function.
Having implemented a trait that adds a custom printing facility to every (any) operation, a third-party owner provides a `.print()` plugin to a service builder:
```
let app = pokemon_service_server_sdk::service::PokemonService::builder()
    ... all other operations ...
    .empty_operation(empty_operation)
    .print()
    .build();
```
Multiple plugins can be added to provide some functionality: `.print().auth()...`

## Checklist


rust-runtime/aws-smithy-http-server/examples/pokemon-service/tests/helpers.rs
```diff
@@ -34,7 +34,7 @@ impl PokemonServiceVariant {
         };
 
         // Give PokémonService some time to start up.
-        time::sleep(Duration::from_millis(500)).await;
+        time::sleep(Duration::from_millis(1000)).await;
```
hlbarber, line 37: Is this required?

rust-runtime/aws-smithy-http-server/src/operation/upgrade.rs
```diff
@@ -273,13 +279,13 @@ pub struct MissingOperation;
 /// This _does_ implement [`Upgradable`] but produces a [`Service`] which always returns an internal failure message.
 pub struct FailOnMissingOperation;
 
-impl<P, Op, Exts, B> Upgradable<P, Op, Exts, B> for FailOnMissingOperation
+impl<P, Op, Exts, B, Plugin> Upgradable<P, Op, Exts, B, Plugin> for FailOnMissingOperation
```
hlbarber, line 282: Maybe we keep `Plugin` here as `Pl` to be consistent with the others?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -213,6 +224,17 @@ class ServerServiceGeneratorV2(
                     }
                 }
             }
+
+            impl<$builderGenerics, NewPlugin> #{SmithyHttpServer}::plugin::Pluggable<NewPlugin> for $builderName<$builderGenerics> {
```
hlbarber, line 228: Having a `,` here in `$builderGenerics, NewPlugin` might cause problems when `$builderGenerics` is empty? I'm not entirely sure. It might be worth generating this using `.joinToString()` so we don't get a comma in when `$builderGenerics` is empty.

hlbarber (approved):
:tada:

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
hlbarber: Same below with `$builderGenericsNoPlugin, aws_smithy_http_server::plugin::PluginStack<$pluginName, NewPlugin>`.

I'm not sure this is worth it though - there might be a ton of other instances where this breaks in the case of no operations.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -64,6 +64,7 @@ class ServerServiceGeneratorV2(
 
     /** A `Writable` block containing all the `Handler` and `Operation` setters for the builder. */
     private fun builderSetters(): Writable = writable {
+        val pluginType = listOf("Plugin")
```
david-perez, line 67: I'd use `Pl` for the generic type name like in the other places.

david-perez (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
82marbag: It breaks the operation registry too

hlbarber: Related: https://github.com/awslabs/smithy-rs/pull/1369

PR #1737, opened 2022-09-15
Update CODEOWNERS

jdisanti:
## Motivation and Context
Needs fixing due to the creation of `codegen-core`.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,6 @@
 *                                               @awslabs/rust-sdk-owners
-/codegen/                                       @awslabs/rust-sdk-owners @crisidev @david-perez
+/buildSrc/                                      @awslabs/rust-sdk-owners @crisidev @david-perez
+/codegen-core/                                  @awslabs/rust-sdk-owners @crisidev @david-perez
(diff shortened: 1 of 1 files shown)


Velfi (approved):

crisidev (approved):

david-perez:
`@awslabs/smithy-rs-server` should be added to `codegen-core`.

We're also missing `codegen-client` and `codegen-client-test`, which should be scoped to `awslabs/rust-sdk-owners`.

jdisanti:
> @awslabs/smithy-rs-server should be added to codegen-core.

Good point. Will fix.

> We're also missing codegen-client and codegen-client-test, which should be scoped to awslabs/rust-sdk-owners.

The wildcard at the top is handling this.

david-perez:
Wildcard should be `@awslabs/rust-sdk-owners && @awslabs/smithy-rs-server`. Currently that wildcard is making it so that changes to e.g. `design`, `tools`, `buildSrc` don't require approval from `@awslabs/smithy-rs-server`, when these directories contain things relating to both projects.

PR #1739, opened 2022-09-15
Add `@awslabs/smithy-rs-server` as owners to `codegen-core`

jdisanti:


jdisanti:
I don't think there is a way to do a logical and. I think we would need to write a GitHub Action to enforce reviews from both groups on `codegen-core`.

jdisanti:
> Perhaps adding two lines with the same directory, each scoped to different teams?

I don't think that's a feature. As far as I can tell, GitHub doesn't support any mechanism for requiring two groups to approve.

> We could also create a separate GitHub team containing the union of both teams.

This would be equivalent to listing the two teams on the folder in `CODEOWNERS`. Only one person from the larger group would have to approve.

david-perez:
Ah yeah you're right.

🤷 I guess this is the best we can do unless we write a GitHub action.

Can you also address https://github.com/awslabs/smithy-rs/pull/1737#issuecomment-1248518829 (allowing `@awslabs/smithy-rs-server` to approve on changes to shared directories like `buildSrc`, `design`, `tools`).

david-perez (approved):

crisidev (approved):

Issue #256, opened 2021-03-17
Default request timeouts


rcoh:
- [x] Rename API timeouts to operation timeouts and move them into `TimeoutConfig`
- [x] Remove currently unused TCP/HTTP timeout config structs
- [x] Add `connect_timeout` to `TimeoutConfig` and default it to 3100 milliseconds
- [x] Implement support for connect timeout
- [x] Integration test all supported timeouts

jdisanti:
Breaking out the per-operation override capability out into #1732 since that requires a good bit of design work and refactoring.

jdisanti:
Defaults should be the following for now:
```
connectTimeoutInMillis: 3100
tlsNegotiationTimeoutInMillis: 3100
```
Taken from the current `sdk-default-configuration.json` standard mode: https://github.com/awslabs/aws-sdk-rust/blob/a58ba1a21f383b26e09ebdd4cc07d57a87e30388/aws-models/sdk-default-configuration.json#L12-L16

jdisanti:
It looks like only the API timeouts have been implemented, so these defaults can't be set yet. Will need to implement the TCP and HTTP timeouts.

PR #1740, opened 2022-09-16
Improve timeout config ergonomics and add SDK default timeouts

jdisanti:
## Motivation and Context
This is work towards #256 for SDK production readiness. This PR:
- Overhauls timeout configuration so that it more closely resembles retry configuration
- Removes configs for timeouts that are not implemented and have no plan to be implemented
- Removes non-standard env var/profile file timeout configuration values so that they won't be a roadblock for standardizing them across SDKs at a later date
- Improves re-exports for runtime types so that customers are less likely to need to take direct dependencies on them
- Refactors Smithy client building:
  - Smithy clients can no longer have their retry/timeout configs changed after creation. Previously, this would have allowed them to get into an invalid state since the `sleep_impl` validation takes place inside of the builder's `build()` method.
  - Tracking standard vs. non-standard retry policy is now handled better in the builder implementation
  - Made the builder's setter functions consistent with our other builders
- Moves the code generator for `CustomizableOperation` into its own file
- Makes it possible to code generate non-root modules
- Renames the `crate::customizable_operation` module to `crate::operation::customize`
- Re-exports types that can be used with operation customization inside of `crate::operation::customize`

## Checklist
- [ ] I have updated the AWS SDK examples


rust-runtime/aws-smithy-client/src/http_connector.rs
```diff
@@ -55,26 +56,99 @@ impl HttpConnector {
...
-/// HttpSettings for HTTP Connectors
+/// Builder for [`HttpSettings`].
 #[non_exhaustive]
 #[derive(Default, Debug)]
-pub struct HttpSettings {
-    /// Timeout configuration used when making HTTP connections
-    pub http_timeout_config: timeout::Http,
-    /// Timeout configuration used when creating TCP connections
-    pub tcp_timeout_config: timeout::Tcp,
+pub struct HttpSettingsBuilder {
```
Velfi, line 62: We should consider renaming this to `ConnectorSettings` because we use the `connector` terminology in several other places and I believe it to be more specific than the current name.

rust-runtime/aws-smithy-client/src/hyper_ext.rs
```diff
@@ -197,63 +205,88 @@ impl Builder {
...
+            Some(duration) => ConnectTimeout::new(
                 connector,
-                sleep
+                sleep_impl
                     .clone()
-                    .expect("a sleep impl must be provided to use timeouts"),
+                    .expect("a sleep impl must be provided in order to have a connect timeout"),
                 duration,
             ),
-            // Some day, we could provide a default timeout if none is set. Today is not that day.
```
Velfi, line 210: I guess today is finally that day 😆

rust-runtime/aws-smithy-types/src/timeout/config.rs
```diff
@@ -89,30 +120,154 @@ impl Config {
...
+/// # }
+/// ```
+#[non_exhaustive]
+#[derive(Clone, PartialEq, Debug)]
+pub struct TimeoutConfig {
+    connect_timeout: Option<Duration>,
+    read_timeout: Option<Duration>,
+    operation_timeout: Option<Duration>,
+    operation_attempt_timeout: Option<Duration>,
+}
```
Velfi, line 206: I don't have any suggestions here, but is it weird when the builder and the built struct are structurally identical? It feels weird. It feels good to uphold the pattern of having a separate builder for forwards compat reasons, so idk.

rust-runtime/aws-smithy-client/src/builder.rs
```diff
@@ -78,9 +85,8 @@ impl<M, R> Builder<(), M, R> {
     pub fn connector<C>(self, connector: C) -> Builder<C, M, R> {
         Builder {
             connector,
-            standard_retry_config: self.standard_retry_config,
-            retry_policy: self.retry_policy,
             middleware: self.middleware,
+            retry_config: self.retry_config,
```
Velfi, line 89: ❤️ ty John

rust-runtime/aws-smithy-client/src/builder.rs
```diff
@@ -10,6 +10,14 @@ use aws_smithy_http::result::ConnectorError;
 use aws_smithy_types::timeout::TimeoutConfig;
 use std::sync::Arc;
 
+#[derive(Clone, Debug)]
+enum BuilderRetryConfig<R> {
+    // Since generic specialization isn't a thing yet, we have to keep a copy of the retry
+    // config around when using standard retry so that it can be properly validated.
+    Standard { has_retry: bool, policy: R },
+    Overridden(R),
```
Velfi, line 18: I think adding some extra info on what gets lost when overriding the retry config would be helpful to understanding how this all works/gets used

rust-runtime/aws-smithy-client/src/builder.rs
```diff
@@ -260,38 +265,39 @@ impl<C, M, R> Builder<C, M, R> {
...
                 implementation set automatically.";
-            if self
-                .standard_retry_config
-                .map(|src| src.has_retry())
-                .unwrap_or(false)
-            {
-                panic!("Retries require a `sleep_impl`, but none was passed into the builder. {ADDITIONAL_HELP}");
-            } else if self.timeout_config.has_timeouts() {
+            if let BuilderRetryConfig::Standard { has_retry, .. } = self.retry_config {
+                if has_retry {
```
Velfi, line 285: I'm wondering if we should make this concept more generic and rename it to something like `depends_on_sleep_impl` or `requires_sleep_impl` because timeouts also have similar language. Perhaps we could even make trait (although that could be going too far.)

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/rustlang/RustModule.kt
```diff
@@ -24,6 +24,7 @@ data class RustModule(val name: String, val rustMetadata: RustMetadata, val docu
             default(name, visibility = Visibility.PRIVATE, documentation = documentation)
 
         val Config = public("config", documentation = "Configuration for the service.")
-        val Error = public("error", documentation = "Errors that can occur when calling the service.")
+        val Error = public("error", documentation = "All error types that operations can return.")
+        val Operation = public("operation", documentation = "All operations that this crate can perform.")
```
Velfi, line 28: ❤️

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/CodegenDelegator.kt
```diff
@@ -139,6 +138,22 @@ open class RustCrate(
         return this
     }
 
+    /**
+     * Create a new non-root module directly.
+     */
+    fun withNonRootModule(
```
Velfi, line 144: consider adding an example of how this will create a module given some input

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/CodegenDelegator.kt
```diff
@@ -153,12 +168,12 @@ open class RustCrate(
  * Allowlist of modules that will be exposed publicly in generated crates
  */
 val DefaultPublicModules = setOf(
-    RustModule.public("error", documentation = "All error types that operations can return."),
-    RustModule.public("operation", documentation = "All operations that this crate can perform."),
+    RustModule.Error,
+    RustModule.Operation,
     RustModule.public("model", documentation = "Data structures used by operation inputs/outputs."),
     RustModule.public("input", documentation = "Input structures for operations."),
     RustModule.public("output", documentation = "Output structures for operations."),
```
Velfi, line 175: How come these three aren't defined as a `val` like `Error`, `Operation`, and `Config`?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
```diff
@@ -23,8 +26,17 @@ class CustomizableOperationGenerator(
...
+    companion object {
+        const val CUSTOMIZE_MODULE = "crate::operation::customize"
+    }
+
     fun render(crate: RustCrate) {
-        crate.withModule(FluentClientGenerator.customizableOperationModule) { writer ->
+        crate.withModule(RustModule.Operation) { writer ->
+            writer.docs("Operation customization and supporting types")
+            writer.rust("pub mod customize;")
+        }
```
Velfi, line 40: It would be cool if this could be handled automatically by `withNonRootModule` somehow but I guess that would open us up to annoying codegen write order issues.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/ResiliencyConfigCustomization.kt
```diff
@@ -221,13 +225,14 @@ class ResiliencyReExportCustomization(private val runtimeConfig: RuntimeConfig)
...
                 pub use #{sleep}::{AsyncSleep, Sleep};
-                pub use #{types_retry}::{RetryConfig, RetryConfigBuilder};
-                /// Config structs required by [`RetryConfig`]
+                /// Retry configuration
                 pub mod retry {
-                    pub use #{types_retry}::RetryMode;
-                    pub use #{types_retry}::RetryKind;
+                    pub use #{types_retry}::{RetryConfig, RetryConfigBuilder, RetryMode};
+                }
+                /// Timeout configuration
```
Velfi, line 232: Should we add a note about these being reëxports so people understand that configs from separate crates are compatible?

Velfi (approved):
I left some comments but this otherwise looks great. Thanks for simplifying all this so much. I love to see changes that delete more code than they add 😄

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -177,21 +178,17 @@ private class AwsFluentClientExtensions(types: Types) {
...
-                    builder.set_retry_config(retry_config.into());
-                    builder.set_timeout_config(timeout_config);
-                    // the builder maintains a try-state. To avoid suppressing the warning when sleep is unset,
-                    // only set it if we actually have a sleep impl.
-                    if let Some(sleep_impl) = sleep_impl {
-                        builder.set_sleep_impl(Some(sleep_impl));
-                    }
+                        .middleware(#{DynMiddleware}::new(#{Middleware}::new()))
+                        .retry_config(retry_config.into())
+                        .timeout_config(timeout_config);
```
jdisanti, line 190: Note for myself: I started integration testing the connect timeout, and that revealed that the `timeout_config` here gets discarded since `dyn_https()` creates the connector and has no consideration of the `timeout_config` value at all.

This is another one of those weird situations with the `aws_smithy_client::Builder` similar to `retry_config` where you can configure a default implementation, or you can replace that implementation wholesale. I think we need to redesign this builder so that these conflicting builder methods are no longer possible. If you replace an implementation, then it shouldn't be possible to set configuration for the default.

rust-runtime/aws-smithy-types/src/timeout/config.rs
jdisanti: I added the builder both to make `TimeoutConfig` consistent with `RetryConfig` as well as for forwards compatibility.

rust-runtime/aws-smithy-client/src/builder.rs
jdisanti: Ended up needing to refactor this even more. It's now completely decomposed so that it's not possible to set something on the builder that another part of the builder makes use of. For example, previously the builder took both a `timeout_config` and a `connector`, but the connector needs pieces from the `timeout_config`, and those same parts used by the connector are not used where the `timeout_config` is actually used. This made it really easy to end up in a difficult to debug misconfiguration.

It should at least be easier to use correctly now, but a bit less convenient to use.

rust-runtime/aws-smithy-client/src/builder.rs
jdisanti: I think the latest refactoring of the builder has improved this. Let me know if you think some additional comments should be added somewhere.

rust-runtime/aws-smithy-client/src/builder.rs
jdisanti: It is now more generic and called `MaybeRequiresSleep` :smile:

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/CodegenDelegator.kt
jdisanti: I only centralized the ones that are referenced in multiple places.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
jdisanti: Certainly an area for improvement. I think we can improve on this incrementally.

hlbarber (approved):
Server side looks good.

aws/rust-runtime/aws-config/src/ecs.rs
```diff
@@ -61,11 +61,15 @@ use tower::{Service, ServiceExt};
...
 use crate::http_credential_provider::HttpCredentialProvider;
 use crate::provider_config::ProviderConfig;
+use aws_smithy_client::http_connector::ConnectorSettings;
 use aws_types::os_shim_internal::Env;
 use http::header::InvalidHeaderValue;
 use std::time::Duration;
 use tokio::sync::OnceCell;
 
+const DEFAULT_READ_TIMEOUT: Duration = Duration::from_secs(5);
+const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_secs(2);
```
Velfi, line 71: It'd be nice if all these defaults were documented in one place in our FAQ or something. Or perhaps they could be imported from a crate of constants?

Velfi: On second thought, I think defining these settings in the module where they are immediately relevant is the right choice, disregard my above comment.

rust-runtime/aws-smithy-client/src/builder.rs
```diff
@@ -67,6 +80,67 @@ where
...
 
+#[cfg(any(feature = "rustls", feature = "native-tls"))]
+use crate::erase::DynConnector;
+#[cfg(any(feature = "rustls", feature = "native-tls"))]
+use crate::http_connector::ConnectorSettings;
+#[cfg(any(feature = "rustls", feature = "native-tls"))]
+use crate::hyper_ext::Adapter as HyperAdapter;
+
+#[cfg(feature = "rustls")]
+impl<M, R> Builder<(), M, R> {
```
Velfi, line 91: This is so clever and cool; Love it.

Velfi (approved):

rust-runtime/aws-smithy-client/src/builder.rs
Velfi: I think it's good now. Your solution is both clever and explicit.

jdisanti:
Opened https://github.com/awslabs/smithy-rs/issues/1771 to track fixing the examples.

PR #1746, opened 2022-09-19
Refactor `redactIfNecessary` function

david-perez:
It's useful to call `redactIfNecessary` on arbitrary shape types, not
only member shapes.

It is now an extension function on `Shape`, with a specialization on
`MemberShape`, since note that the `sensitive` trait cannot be applied
to member shapes [0, 1], so we traverse to its target directly.

Ultimately this is just to make the diff of #1342 lighter.

[0]: https://awslabs.github.io/smithy/1.0/spec/core/documentation-traits.html#sensitive-trait
[1]: https://awslabs.github.io/smithy/1.0/spec/core/documentation-traits.html#sensitive-trait

## Testing

No diff in generated artifacts.


crisidev (approved):
Makes sense. LGTM

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt
```diff
@@ -81,6 +82,16 @@ fun ServiceShape.hasEventStreamOperations(model: Model): Boolean = operations.an
     model.expectShape(id, OperationShape::class.java).isEventStream(model)
 }
 
+fun MemberShape.redactIfNecessary(model: Model, safeToPrint: String) =
+    model.expectShape(this.target).redactIfNecessary(safeToPrint)
+
+fun Shape.redactIfNecessary(safeToPrint: String) =
```
jdisanti, line 88: I think this could be error prone. For example, if someone calls `Shape.redactIfNecessary` on a `MemberShape`, which is possible since a `MemberShape` is a shape, then that might result in the wrong code generated behavior, right? Perhaps we should just have one function that always takes a model, and do any necessary switching logic based on the type in that.

david-perez: Yeah, it crossed my mind that this could be error prone.

> if someone calls Shape.redactIfNecessary on a MemberShape, which is possible since a MemberShape is a shape, then that might result in the wrong code generated behavior, right?

Yes. I think there's an argument to be made that it's a mistake _on behalf of the caller_, since `sensitive` cannot be applied to member shapes. I only introduced `MemberShape.redactIfNecessary` as syntax sugar for the caller to not have to traverse to the member's target, but I don't know if I prefer to lay that responsibility on the caller, or to have a single function that always takes in `model` like you suggest. All 3 ways are reasonable.

david-perez: https://github.com/awslabs/smithy-rs/pull/1746/commits/d7e8572612bd9738740e42ac5aafcfebad66ce15

PR #1748, opened 2022-09-19
Don't error on duplicate pagination token

jdisanti:
## Motivation and Context
Some AWS operations, such as CloudWatch's GetLogEvents, respond with duplicate next page tokens so that infinite polling can be done (for use cases such as tailing a log). The SDK was previously erroring out in the paginator for this use case (see https://github.com/awslabs/aws-sdk-rust/issues/620).

This PR adds a `stop_on_duplicate_token` option to the paginators, and removes the error. This solution is the same as [what the Go V2 SDK is doing](https://github.com/aws/aws-sdk-go-v2/blob/main/service/cloudwatchlogs/api_op_GetLogEvents.go#L175-L177).

## Checklist


jdisanti:
I currently have `stop_on_duplicate_token` defaulting to `true`, since it felt wrong to default to infinite looping. Not sure this is correct though. Will cross check what other SDKs are doing.

jdisanti:
It looks like Java V2 uses a customization to mark operations as ending pagination by returning a duplicate token. Go V2 requires the customer to specify a value for its `StopOnDuplicateToken`. The Rust version of this requires a default with the current fluent paginator builder design, so given the above, I'm going to keep it defaulting to `true`.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/PaginatorGenerator.kt
```diff
@@ -139,13 +140,21 @@ class PaginatorGenerator private constructor(
...
+                        stop_on_duplicate_token: true,
                     }
                 }
 
                 #{page_size_setter:W}
 
                 #{items_fn:W}
 
+                /// Stop paginating when the service returns the same pagination token twice in a row.
+                ///
```
Velfi, line 152: We should include docs on when this would be helpful to set to false:
> When running status-checking operations or operations that you want to continually send in order to "watch" for the latest data (like AWS CloudWatch's `GetLogEvents`), set this to false. That way, you can continually call the paginator in order to receive new data as it becomes available.

Velfi:
> It looks like Java V2 uses a customization to mark operations as ending pagination by returning a duplicate token. Go V2 requires the customer to specify a value for its `StopOnDuplicateToken`. The Rust version of this requires a default with the current fluent paginator builder design, so given the above, I'm going to keep it defaulting to `true`.

I think this is the right approach to take. Infinite looping is something that people shouldn't be able to accidentally trigger unless the docs are very clear about when it happens. I do think we should add some info on why infinite looping is sometimes helpful and guide people to pass `false` when relevant to their use case.

Velfi (approved):

Velfi (approved):

PR #1753, opened 2022-09-20
Expand protocol layouts

82marbag:
## Motivation and Context

Routes and protocols currently live in the same module. To allow their further extension with new protocols and to reduce complexity of their current implementation, we should split protocols into their own module.

Part of [RFC 20](https://awslabs.github.io/smithy-rs/design/rfcs/rfc0020_service_builder.html)

## Description
New module structure:
```
aws_smithy_http_server::proto::rest::rest_json_1
aws_smithy_http_server::proto::rest::rest_xml
aws_smithy_http_server::proto::aws_json::aws_json_10
aws_smithy_http_server::proto::aws_json::aws_json_11
```

## Testing

`make && cargo test` in the Pokemon service folder

## Checklist


rust-runtime/aws-smithy-http-server/src/proto/mod.rs
```diff
@@ -0,0 +1,7 @@
+/*
```
hlbarber, line 1: Thinking out loud: Should we flatten this a bit and take `rest_json_1`/`rest_xml` out of `rest` and put them at this level? The common structs would still live in `rest` - the `rest` folder wouldn't contain a protocol ZST, but will contain a `router` module etc.

hlbarber: Couple this with https://github.com/awslabs/smithy-rs/pull/1753#discussion_r976310182 and everything might be more predictable for codegen - `proto/{protocol}/router.rs` always contains the `Router` and `Error` for every protocol.

hlbarber:
```
/proto/rest/router/RestRouter
/proto/rest/router/Error
/proto/rest_json_1/RestJson1
/proto/rest_json_1/router/RestRouter (re-exported)
/proto/rest_json_1/router/Error (re-exported)
/proto/rest_xml/RestXml
/proto/rest_xml/router/RestRouter (re-exported)
/proto/rest_xml/router/Error (re-exported)
/proto/aws_json/...
/proto/aws_json_1/...
```

rust-runtime/aws-smithy-http-server/src/lib.rs
```diff
@@ -31,13 +31,16 @@ pub mod routing;
 #[doc(hidden)]
 pub mod runtime_error;
 
+pub mod routers;
```
hlbarber, line 34: I think this should be `#[doc(hidden)]` for now.

hlbarber (approved):

PR #1755, opened 2022-09-20
Updates and fixes to HttpSensitivityGenerator

hlbarber:
## Motivation and Context

* Closes https://github.com/awslabs/smithy-rs/issues/1702
* Prepares for the logging middleware to be converted into a `Plugin`.

## Description

- Simplify sensitivity detection logic by assuming bindings only exist at top level input/output.
- Fix greedy label logic - they can now exist alongside of normal labels.
- Fix `httpPrefixHeader` false positive.
- Export `RequestFmt` and `ResponseFmt` types alongside the closures.
- More extensive use of `Writable` to improve readability.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerHttpSensitivityGenerator.kt
```diff
@@ -18,334 +15,435 @@ import software.amazon.smithy.model.traits.HttpLabelTrait
...
-        writer.rustBlockTemplate("|name: &#{Http}::header::HeaderName|", *codegenScope) {
-            rust("##[allow(unused_variables)]")
-            rust("let name = name.as_str();")
+    /** Returns the closure used during construction. */
+    internal fun closure(): Writable {
+        val nameMatch = if (headerKeys.isEmpty()) writable {
+            rust("false")
+        } else writable {
+            val matches = headerKeys.joinToString("|") { it.dq() }
+            rust("matches!(name, $matches)")
```
82marbag, line 163: Is `name` always used as `let name = name.as_str();`? If so, you could move `.as_str()` in `matches` and avoid the `allow_unused` (it's being added also when not needed)

hlbarber: I think the closure might complain that `name` is unused entirely in this case.

hlbarber: Worked, thanks! https://github.com/awslabs/smithy-rs/pull/1755/commits/b46b63e6894de5d07f7ede504048a7c2b1f81964

hlbarber:
Putting this back into draft to add more test coverage and documentation.

82marbag (approved):

PR #1762, opened 2022-09-22
Move more things into `codegen-core`

david-perez:
This moves most things from `codegen-client` that should be in
`codegen-core` into `codegen-core`.

This is a continuation of the efforts started in #1697, #1730.

## Testing

No generated diff should be observed.


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/RemoveEventStreamOperations.kt
```diff
@@ -3,13 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package software.amazon.smithy.rust.codegen.client.smithy.transformers
+package software.amazon.smithy.rust.codegen.core.smithy.transformers
```
jdisanti, line 6: I'm surprised to see `RemoveEventStreamOperations` used by the server. Was this intentional? It requires allow listing service module names through the `smithy-build.json`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/ClientProtocolLoader.kt
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.rust.codegen.core.smithy.protocols
```
jdisanti, line 6: This file should be renamed to `ProtocolLoader.kt`

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/GenericsGenerator.kt
```diff
@@ -3,12 +3,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package software.amazon.smithy.rust.codegen.client.smithy.generators
+package software.amazon.smithy.rust.codegen.core.smithy.generators
```
jdisanti, line 6: I think this class belongs in `rustlang` with the code that uses it.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustReservedWords.kt
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package software.amazon.smithy.rust.codegen.client.rustlang
+package software.amazon.smithy.rust.codegen.core.rustlang
```
jdisanti, line 6: I think `RustReservedWords` should stay in client for now. It needs to be broken up into multiple composed classes since right now it has client-specific logic in it to handle the `Unknown` enum variants.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
```diff
@@ -3,21 +3,22 @@
...
+import software.amazon.smithy.rust.codegen.core.rustlang.docs
+import software.amazon.smithy.rust.codegen.core.rustlang.documentShape
+import software.amazon.smithy.rust.codegen.core.rustlang.rust
+import software.amazon.smithy.rust.codegen.core.rustlang.rustBlock
+import software.amazon.smithy.rust.codegen.core.smithy.CodegenTarget
+import software.amazon.smithy.rust.codegen.core.smithy.expectRustMetadata
+import software.amazon.smithy.rust.codegen.core.smithy.renamedFrom
 import software.amazon.smithy.rust.codegen.core.util.toSnakeCase
 
 fun CodegenTarget.renderUnknownVariant() = when (this) {
```
jdisanti, line 24: This function shouldn't get copied into core.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CodegenTarget.kt
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package software.amazon.smithy.rust.codegen.client.smithy.generators
+package software.amazon.smithy.rust.codegen.core.smithy
```
jdisanti, line 6: I don't think core should have the concept of `CodegenTarget`. Anything that depends on should be kept in client so that it can be individually refactored into core with the target if/else cases removed.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreCodegenContext.kt
```diff
@@ -3,12 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package software.amazon.smithy.rust.codegen.client.smithy
+package software.amazon.smithy.rust.codegen.core.smithy
```
jdisanti, line 6: I think we should rename the `Core-` prefix out of these now that they're in the actual core module.

jdisanti (commented):
Nice work! Love seeing this happen.

It looks like none of the unit tests were moved? Most of them require `testutil` to move, which is not so easy. But it looks like you have enough critical mass of classes moved that `testutil` should actually move over successfully now. I would definitely prefer the tests move along with the core classes.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/RemoveEventStreamOperations.kt
david-perez: I was surprised too. I've opened https://github.com/awslabs/smithy-rs/pull/1764; once that's merged, I'll keep the transformer in `codegen-client` in this PR.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/GenericsGenerator.kt
david-perez: Done in [Move GenericsGenerator into rustlang and rename it to RustGenerics](https://github.com/awslabs/smithy-rs/pull/1762/commits/d6385d96267cd8e564fbc2037c318560afe64ad6). I also renamed it to `RustGenerics`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/ClientProtocolLoader.kt
david-perez: Good catch! Done in [Rename protocol loader in codegen-core from ClientProtocolLoader …](https://github.com/awslabs/smithy-rs/pull/1762/commits/0876167a0ed066c4db281e23ee9fa797ce98c191).

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreCodegenContext.kt
david-perez: Agree. Done in [Rename CoreCodegenContext as CodegenContext, since it's now in th…](https://github.com/awslabs/smithy-rs/pull/1762/commits/90d250dbb5ef6781050f623c14d58e124d819d28).

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CodegenTarget.kt
david-perez: I agree with this sentiment in the _general case_, but:

1. Untangling the branching on `CodegenTarget` is diffcult; and
2. I'm actually not sure that removing the branching _entirely_ is beneficial. In certain _particular_ core generators, all you need is a quick and dirty `if (target == ...)` to get the job done. In some cases I think that introducing inheritance/composition/customization injections to get rid of the notion of `CodegenTarget` can be detrimental to code maintainability.

`CodegenTarget` is used by > 14 generators that this PR places in `codegen-core`. Keeping them in `codegen-client` entails bringing more things back. It's too ingrained. I think putting `CodegenTarget` in `codegen-core` is the sweet spot here.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
david-perez: See my comment here https://github.com/awslabs/smithy-rs/pull/1762#discussion_r978609884.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustReservedWords.kt
david-perez: I agree we should refactor that bit out. However, I can't keep it in `codegen-client`, since `BuilderGenerator`, that is `codegen-core` depends on it, and we can't have `core` depending on `client`. We should eventually put `BuilderGenerator` in `codegen-client` anyway, since in #1342 the server project has its own builder generators. However, if I do so now in this PR, several things break, because the extension functions defined in `BuilderGenerator.kt` (`StructureShape.builderSymbol`, `RuntimeConfig.operationBuildError()`, `MemberShape.setterName()`...) are ingrained throughout `codegen-core`. Untangling all these is related to the `CodegenTarget` and reducing branching cleanup, that we should tackle separately.

david-perez: I opened https://github.com/awslabs/smithy-rs/issues/1766 to track this.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CodegenTarget.kt
david-perez: I opened https://github.com/awslabs/smithy-rs/issues/1765 to track this.

hlbarber (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/RemoveEventStreamOperations.kt
david-perez: #1764 was merged; I've moved `RemoveEventStreamOperations` back to the client in https://github.com/awslabs/smithy-rs/pull/1762/commits/67dc987505c9aadad62b9e23136071e425a99399.

jdisanti (approved):

PR #1763, opened 2022-09-23
Cleanup pokemon.smithy

hlbarber:
## Motivation and Context

* The `pokemon.smithy` is used as an example for customers and should encode best practices.
* Having `Operation` suffix on operation names collides with the `_operation` suffix of https://github.com/awslabs/smithy-rs/blob/e5c8cf3061a21f4d33423fd7df642d9c5fb70729/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt#L132

## Description

- Add missing full-stops.
- Remove `Operation` suffix from operations.
- Don't re-use an `@output` struct.

### DIFF
--- a/codegen-client-test/build.gradle.kts
+++ b/codegen-client-test/build.gradle.kts
@@ -88,7 +88,7 @@ val allCodegenTests = "../codegen-core/common-test-models".let { commonModels ->
             """.trimIndent(),
             imports = listOf("$commonModels/naming-obstacle-course-structs.smithy"),
         ),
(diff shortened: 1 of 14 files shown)


82marbag (approved):

crisidev (changes requested):
You will have to change the Python service implementation as well since operation names have changed. We codegenerate decorators with the operation name and I expect this CR to fail.

crisidev:
FYI this is the upstream issue tracking the full reconciliation of the Python and Rust model: https://github.com/awslabs/smithy-rs/issues/1508

crisidev (approved):

PR #1764, opened 2022-09-23
Remove `RemoveEventStreamOperations` model transformer from `ServerCodegenVisitor`

david-perez:
This is a holdover from when the server subproject was started. We've
never utilized this model transformer, nor will we have any use for it
now, since event stream operations are supported in the server since #1479.

The `eventStreamAllowList` codegen config setting has been moved from
`CoreCodegenConfig` into `ClientCodegenConfig`.

See https://github.com/awslabs/smithy-rs/pull/1762#discussion_r978151466.

## Testing

No diff should be observed.

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/TestHelpers.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/TestHelpers.kt
@@ -7,6 +7,7 @@ package software.amazon.smithy.rust.codegen.client.testutil
 
 import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.knowledge.NullableIndex
+import software.amazon.smithy.model.node.ObjectNode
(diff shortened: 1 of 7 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/TestHelpers.kt
```diff
@@ -41,6 +44,32 @@ val TestSymbolVisitorConfig = SymbolVisitorConfig(
     nullabilityCheckMode = NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1,
 )
 
+fun clientTestRustSettings(
```
david-perez, line 47: After #1762, whenever we move the unit tests from `codegen-client` to `codegen-core`, this function will stay in `codegen-client`'s unit tests and the one below will be moved to `codegen-core`.

jdisanti (approved):

PR #1773, opened 2022-09-27
Simplify Python Pokemon service

unexge:
I've changed `FastWriteCounter` with a simpler implementation and added a `main` function to use as entry point.


rust-runtime/aws-smithy-http-server-python/src/server.rs
```diff
@@ -426,7 +426,11 @@ event_loop.add_signal_handler(signal.SIGINT,
         // Forcing the multiprocessing start method to fork is a workaround for it.
         // https://github.com/pytest-dev/pytest-flask/issues/104#issuecomment-577908228
         #[cfg(target_os = "macos")]
-        mp.call_method1("set_start_method", ("fork",))?;
+        mp.call_method(
+            "set_start_method",
+            ("fork",),
+            Some(vec![("force", true)].into_py_dict(py)),
```
unexge, line 432: without this line it fails with `context has already been set` see https://github.com/pytorch/pytorch/issues/3492

crisidev (approved):

PR #1776, opened 2022-09-27
Move core tests from `codegen-client` into `codegen-core`

jdisanti:
## Motivation and Context
This is a continuation of the refactor of core functionality from `codegen-client` into `codegen-core`. This PR specifically attempts to move all the unit tests in `codegen-client` that are testing functionality that is currently in `codegen-core`.


Velfi (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/ClientProtocolLoader.kt
```diff
@@ -13,21 +13,92 @@ import software.amazon.smithy.aws.traits.protocols.RestJson1Trait
...
+    override fun protocol(codegenContext: ClientCodegenContext): Protocol = Ec2QueryProtocol(codegenContext)
+
+    override fun buildProtocolGenerator(codegenContext: ClientCodegenContext): HttpBoundProtocolGenerator =
+        HttpBoundProtocolGenerator(codegenContext, protocol(codegenContext))
+
+    override fun support(): ProtocolSupport = CLIENT_PROTOCOL_SUPPORT
+}
+
+class ClientRestXmlFactory(
+    private val generator: (CodegenContext) -> Protocol = { RestXml(it) },
```
david-perez, line 96: I'd like to take this example to ask a more general question about the way we're writing interfaces. This is an example where the interface could be tightened by requiring `ClientCodegenContext`, even if it doesn't need any of the specialized properties that it offers. Oftentimes we write interfaces that are as general as possible (requiring the bare minimum in a signature conveys valuable information to readers) and other times we tighten them to prevent misuse (it's clear we don't want the server project using this). Should we lean more towards the former or the latter?

I prefer the former, especially since _eventually_ the plan is for `codegen-server` to not depend on `codegen-client`, so this particular misuse scenario will become impossible.

Analogously (and in contrast), I think that _in general_ we should keep classes used _exclusively_ by one of the subprojects within the subproject itself, even if those classes _could_ be used by the other subproject to do something minimally useful (like we did with the `RemoveEventStreamOperations` model transformer, keeping it in `codegen-client`). I say _in general_ because there are clear exceptions: for example, `codegen.core.rustlang.RustGenerics` is right now only used by the client, but it's clear there's nothing client-specific about it and that the server generators could benefit from it.

So I guess the rule of thumb would be:
1. keep a class in the specific subproject if it's exclusively used in that subproject (unless you foresee that it's something broadly useful and clearly subproject-agnostic, in which case it should go into `codegen-core`); and
2. _within_ a subproject, keep signatures as general as possible.

Just wanted to write down these thoughts to see whether others agree or not.

david-perez (approved):
Very nice. Thank you. We should now work towards `codegen-server` _not_ depending on `codegen-client`.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/ClientProtocolLoader.kt
jdisanti: > 2. within a subproject, keep signatures as general as possible.

This has the downside of it being harder to reference client/server-specific properties in that more general code at a later date, as the general type may have to be changed in multiple layers to get it to the piece of code that needs it. The upside is that it's easier to move code into core, but this doesn't seem like it's going to be a frequent occurrence to me.

PR #1779, opened 2022-09-28
Fix HttpSensitivityGenerator string length calculation

hlbarber:
## Motivation and Context

There is no guarantee that Kotlins `String::length` is equivalent to Rusts `String::length`.

## Description

Add and use `toRustLen()` in `HttpSensitivityGenerator.kt`.


Velfi (approved):
Good catch

david-perez (commented):
The HTTP RFCs require that URIs and headers have to be ASCII-encoded, so the length calculation should be the same.

Smithy should also enforce this. See e.g. https://github.com/awslabs/smithy/issues/975

hlbarber:
Closing this in light of https://github.com/awslabs/smithy-rs/pull/1779#pullrequestreview-1124202615

PR #1782, opened 2022-09-29
Add "Anatomy of a Service" documentation

hlbarber:
## Motivation and Context

The new service builder API, described in [Service Builder Improvements](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0020_service_builder.md), is built on a variety of interfaces. Proper documentation should be provided to onboard new contributors on how they stack.

## Description

[Rendered](https://github.com/awslabs/smithy-rs/blob/harryb/anatomy-service/design/src/server/anatomy.md)

- Tweaks to Rust documentation.
- Add `anatomy.md` and various diagrams.

## Notes

- This API is implemented but is currently behind `#[doc(hidden)]` while it awaits customer feedback.
- This will not be made public (won't be indexed it in the [book](https://awslabs.github.io/smithy-rs/design/overview.html)) until the new service builder is too.
- https://github.com/awslabs/smithy-rs/issues/1778 if this is acted upon we should switch the GitHub links accordingly.


design/src/docs/anatomy.md
```diff
@@ -0,0 +1,601 @@
...
+    type Output;
+
+    /// Applies a plugin to the service builder.
+    fn apply(self, plugin: NewPlugin) -> Self::Output;
+}
+```
+
+As seen in the `Pluggable` documentation, third-parties can use extension traits over `Pluggable` to extend the API of builders.
+
+## Accessing Unmodelled Data
```
hlbarber, line 571: This section is quite brief. The full story amounts to pepper the code with `Exts: FromParts<P>` in a predictable way and can be read quite easily from the source code.

Should I go into more detail? Should I include the paragraph above?

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,601 @@
...
+pub trait Pluggable<NewPlugin> {
+    type Output;
+
+    /// Applies a plugin to the service builder.
+    fn apply(self, plugin: NewPlugin) -> Self::Output;
+}
+```
+
+As seen in the `Pluggable` documentation, third-parties can use extension traits over `Pluggable` to extend the API of builders.
+
```
hlbarber, line 570: I think it's probably a good idea to add the implementation of `Pluggable` on the `PokemonServiceBuilder`.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,601 @@
...
+```rust
+/// The Pokémon Service allows you to retrieve information about Pokémon species.
+#[derive(Clone)]
+pub struct PokemonService<S> {
+    router: RoutingService<RestRouter<S>, AwsRestJson1>,
+}
+```
+
+## Plugins
```
hlbarber, line 500: There should be a standalone tutorial on actually writing a `Plugin`. And when there is we should link to it here.

Add a TODO.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,601 @@
+# The Anatomy of a Service
+
+What is [Smithy](https://awslabs.github.io/smithy/2.0/index.html)? At a high-level, it's a grammar for specifying services modulo business logic. A [Smithy Service](https://awslabs.github.io/smithy/2.0/spec/service-types.html#service) specifies a set of function signatures in the form of [Operations](https://awslabs.github.io/smithy/2.0/spec/service-types.html#operation) to which encapsulate business logic. A Smithy implementation should, for each Smithy Service, provide a builder, which accepts functions conforming to said signatures, and returns a service subject to the semantics specified by the model.
+
+This is an dissection of the various mechanisms at work inside Smithy Rust in order to provide such a builder. This is intended for new contributors or users who are interested in the internal details.
```
hlbarber, line 5: This should be changed to emphasize that this is not a description of the logic/implementation details of the Kotlin code but rather a description of how the Rust composes together and it's relation to the Smithy model.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,601 @@
...
+
+/// [AWS JSON 1.1 Protocol](https://awslabs.github.io/smithy/2.0/aws/protocols/aws-json-1_1-protocol.html).
+pub struct AwsJson11;
+```
+
+## Upgrading a Model Service
+
+Notice that we can "upgrade" a model service to a HTTP service using `FromRequest` and `IntoResponse` described in the prior section:
+
+![Upgrade Data Flow Diagram](imgs/upgrade-dfd.png)
```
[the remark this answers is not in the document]
hlbarber: There's also the question of where should we host them?

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+As mentioned, `L` is applied _after_ the `Operation<S, L>` has been "upgraded" to a HTTP service. The procedure of upgrading a model service to a HTTP service is described in the [Upgrading a Model Service](#upgrading-a-model-service) section below.
+
+## Serialization and Deserialization
+
+A [Smithy protocol](https://awslabs.github.io/smithy/2.0/spec/protocol-traits.html#serialization-and-protocol-traits) specifies the serialization/deserialization scheme - how a HTTP request is transformed into a modelled input and a modelled output to a HTTP response. The is formalized using the [`FromRequest`](https://github.com/awslabs/smithy-rs/blob/4c5cbc39384f0d949d7693eb87b5853fe72629cd/rust-runtime/aws-smithy-http-server/src/request.rs#L156-L164) and [`IntoResponse`](https://github.com/awslabs/smithy-rs/blob/4c5cbc39384f0d949d7693eb87b5853fe72629cd/rust-runtime/aws-smithy-http-server/src/response.rs#L40-L44) traits:
+
+```rust
+/// Provides a protocol aware extraction from a [`Request`]. This consumes the
+/// [`Request`], in contrast to [`FromParts`].
+pub trait FromRequest<Protocol>: Sized {
```
hlbarber, line 171: These traits, and others all have a loose counterpart in `axum` e.g. https://docs.rs/axum/latest/axum/extract/trait.FromRequest.html because

1. We used to used `axum` extensively
2. These traits are reasonably natural and occur across all frameworks ([Rocket](https://api.rocket.rs/v0.4/rocket/request/trait.FromRequest.html, [Actix](https://docs.rs/actix-web/latest/actix_web/trait.FromRequest.html), )

Should we mention this?

design/src/docs/anatomy.md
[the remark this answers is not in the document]
hlbarber: `mermaid` looks incredible - this is a great idea.

design/src/docs/anatomy.md
LukeMathWalker: I don't think it's necessary, it's a fairly common approach in the Rust web ecosystem (as you stated).

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+            empty_operation: self.empty_operation,
+            /* ... */,
+
+            plugin: PluginStack::new(self.plugin, plugin),
+        }
+    }
+}
+```
+
+Here `PluginStack` works in a similar way to [`tower::layer::util::Stack`](https://docs.rs/tower/latest/tower/layer/util/struct.Stack.html) - allowing users to append a new plugin rather than replacing the currently set one.
```
LukeMathWalker, line 655: It's not immediately clear to me when a user should opt for a `Plugin` instead of attaching a `Layer` to the overall `Router`. 
A concrete example of common functionality that could only be accomplished with a `Plugin` instead of a `Layer` would ⭐

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+pub trait Router {
+    type Service;
+    type Error;
+
+    /// Matches a [`http::Request`] to a target [`Service`].
+    fn match_route(&self, request: &http::Request) -> Result<Self::Service, Self::Error>;
+}
+```
+
+which provides the ability to lookup an inner HTTP service from a collection using a `&http::Request`.
```
david-perez, line 323: nit: "lookup" gives me an O(1) immediacy sense that is not necessarily true. "determine" might be a better verb?

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+}
+
+impl<R, P> Service<http::Request> for RoutingService<R, P>
+where
+    R: Router<B>,
+    R::Service: Service<http::Request, Response = http::Response>,
+    R::Error: IntoResponse<P> + Error,
+{
+    type Response = http::Response;
+    type Error = Box<dyn Error>;
```
david-perez, line 343: I'd call out again that we're simplifying the implementation in favor of a more pedagogical text, lest a reader think we're actually boxing every time we return a routing error.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+}
+```
+
+The `RouterService` is the final piece necessary to form a functioning composition - it is used to aggregate together the HTTP services, created via the upgrade procedure, into a single HTTP service which can be presented to the customer.
+
+![RouterService](imgs/router.png)
+
+## Builders
+
+The service builder is the primary public API, it can be generated for every [Smithy Service](https://awslabs.github.io/smithy/2.0/spec/service-types.html). At a high-level, the purpose of a service builder is to take a function for each Smithy Operation, whose signature is constrained by the Smithy model, and produce a single HTTP service.
```
david-perez, line 365: Or closure, or model service, as explained in the beginning.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+    capture_pokemon_operation: Op1,
+    empty_operation: Op2,
+    get_pokemon_species: Op3,
+    get_server_statistics: Op4,
+    get_storage: Op5,
+    health_check_operation: Op6,
+}
+```
+
+The builder has one generic variable for each [Smithy Operation](https://awslabs.github.io/smithy/2.0/spec/service-types.html#operation) in the [Smithy Service](https://awslabs.github.io/smithy/2.0/spec/service-types.html#service). Each setter switches the type of the `Op{N}` type parameter:
```
david-perez, line 381:
```suggestion
The builder has one generic type parameter for each [Smithy Operation](https://awslabs.github.io/smithy/2.0/spec/service-types.html#operation) in the [Smithy Service](https://awslabs.github.io/smithy/2.0/spec/service-types.html#service). Each setter switches the type of the `Op{N}` type parameter:
```

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+            get_pokemon_species: value,
+            get_server_statistics: self.get_server_statistics,
+            get_storage: self.get_storage,
+            health_check_operation: self.health_check_operation,
+        }
+    }
+
+    /// Sets the [`GetPokemonSpecies`](crate::operation_shape::GetPokemonSpecies) operation.
+    ///
+    /// This should be a closure satisfying the [`Handler`](aws_smithy_http_server::operation::Handler) trait.
```
david-perez, line 413: Should we change these docs to highlight this can also be a function? And to mention `async`.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+        Op6,
+    >
+    where
+        H: Handler<GetPokemonSpecies>,
+    {
+        self.get_pokemon_species_operation(GetPokemonSpecies::from_handler(value))
+    }
+```
+
+To finalize the build and construct the complete service, `PokemonService`, each builder has a `build` method whose constraints list out all the requirements for composition:
```
david-perez, line 433:
```suggestion
To finalize the build and construct the complete service, `PokemonService`, each builder has a `build` method whose bounds list out all the requirements for composition:
```

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+            do_nothing: FailOnMissingOperation,
+            get_pokemon_species: FailOnMissingOperation,
+            get_server_statistics: FailOnMissingOperation,
+            capture_pokemon: FailOnMissingOperation,
+            get_storage: FailOnMissingOperation,
+        }
+    }
+```
+
+The `builder` constructor provides a `PokemonServiceBuilder` where `build` cannot be called until all operations are set because `MissingOperation` purposefully doesn't implement `Upgradable`. In contrast, the `unchecked_builder` which sets all `Op{N}` to `FailOnMissingOperation` can be immediately built, however any unset operations are upgraded into a service which always returns status code 500, as noted in [Upgrading a Model Service](#upgrading-a-model-service).
```
david-perez, line 498: I'd omit `unchecked_builder` entirely. It's just going to make the reader raise their eyebrows as to its utility if we don't mention here the multi-Lambda based set up use case, which is not relevant to this text.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+        let layer = Stack::new(UpgradeLayer::new(), layer);
+        layer.layer(inner)
+    }
+```
+
+![Upgradable Diagram](imgs/upgradable-plugin.png)
+
+An example `Plugin` implementation can be found in [aws-smithy-http-server/examples/pokemon-service/src/plugin.rs](https://github.com/awslabs/smithy-rs/blob/main/rust-runtime/aws-smithy-http-server/examples/pokemon-service/src/plugin.rs).
+
+All builders implement the [`Pluggable`](https://github.com/awslabs/smithy-rs/blob/4c5cbc39384f0d949d7693eb87b5853fe72629cd/rust-runtime/aws-smithy-http-server/src/plugin.rs#L8-L29) trait, which allows them to apply plugins to service builders:
```
david-perez, line 587:
```suggestion
The service builder implements the [`Pluggable`](https://github.com/awslabs/smithy-rs/blob/4c5cbc39384f0d949d7693eb87b5853fe72629cd/rust-runtime/aws-smithy-http-server/src/plugin.rs#L8-L29) trait, which allows service owners to apply plugins:
```

design/src/docs/anatomy.md
david-perez, line 655:
```suggestion
Here `PluginStack` works in a similar way to [`tower::layer::util::Stack`](https://docs.rs/tower/latest/tower/layer/util/struct.Stack.html) - allowing users to compose plugins rather than replacing the currently set one.
```

compose/stack/layer might be better verbs than append.

david-perez: That's a very good observation! Plugins allow us to modify the "upgrade procedure" from model services to HTTP services. There we can "squish in" middlewares that work at the model service layer, whereas the ones applied to the overall Router need to operate at the HTTP service layer, _before_ routing has taken place. You'll notice this difference if you look closely at the diagrams.

You can pull the thread for the primary motivation for introducing plugins in item 2 in the [Service Builder Improvements RFC](https://awslabs.github.io/smithy-rs/design/rfcs/rfc0020_service_builder.html#operations-as-middleware-constructors), to perform operation-specific logging behavior, but there are other use cases where we want to modify behavior at the model service layer.

david-perez (approved):
Excellent documentation. I read it in the headspace of a newcomer and in my opinion everything is very well-explained.

design/src/docs/anatomy.md
hlbarber: I'm on the fence here - if we don't have `unchecked_builder` and `FailOnMissingOperation` then we don't have an example of an alternative `Upgradable::upgrade` path which weakens the motivation for the `Upgradable` trait entirely - which might raise eyebrows.

design/src/docs/anatomy.md
hlbarber: I'm tempted to leave it as is because a [`Service`](https://docs.rs/tower-service/latest/tower_service/trait.Service.html) is already described as a

>An asynchronous function from a Request to a Response.

And closures are anonymous functions, so this is "true enough".

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,687 @@
...
+
+The [`Handler`](https://github.com/awslabs/smithy-rs/blob/4c5cbc39384f0d949d7693eb87b5853fe72629cd/rust-runtime/aws-smithy-http-server/src/operation/handler.rs#L21-L29) and [`OperationService`](https://github.com/awslabs/smithy-rs/blob/4c5cbc39384f0d949d7693eb87b5853fe72629cd/rust-runtime/aws-smithy-http-server/src/operation/operation_service.rs#L15-L29) both serve a similar purpose - they provide a common interface for converting to a model service `S`. The `Handler<GetPokemonSpecies>` trait covers all closures taking `GetPokemonSpeciesInput` and asynchronously returning a `Result<GetPokemonSpeciesOutput, GetPokemonSpeciesError>` - they are converted to a model service by a `IntoService`. The `OperationService<GetPokemonSpecies>` trait covers all `tower::Service`s with request `GetPokemonSpeciesInput`, response `GetPokemonSpeciesOutput` and error `GetPokemonSpeciesOutput` - they are converted to a model service by a `Normalize` (this is a very small conversion which flattens request tuples).
```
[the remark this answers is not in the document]
hlbarber: There's some detailed documentation in the `operation.rs` module. I'll link to that so they can see the details if they want.

design/src/docs/anatomy.md
hlbarber: How's this?
https://github.com/awslabs/smithy-rs/pull/1782/commits/8ff688e219a6dd41fc73cf65646116d0e657f52f

There's an unfortunate conflict, [which exists for now](https://github.com/awslabs/smithy-rs/issues/1707), between the [`Router`](https://github.com/awslabs/smithy-rs/blob/89a6e5894aebb81efa1305c41188d7ded3b3c43b/rust-runtime/aws-smithy-http-server/src/routing/mod.rs#L64-L67) struct, which is essentially an enum of the [protocol specific routers](https://github.com/awslabs/smithy-rs/blob/89a6e5894aebb81efa1305c41188d7ded3b3c43b/rust-runtime/aws-smithy-http-server/src/routing/mod.rs#L69-L82) and the [`Router`](https://github.com/awslabs/smithy-rs/blob/89a6e5894aebb81efa1305c41188d7ded3b3c43b/rust-runtime/aws-smithy-http-server/src/routers/mod.rs#L40-L47) trait which is an interface for routing requests to services. This makes things more confusing for people today.

design/src/docs/anatomy.md
```diff
@@ -547,7 +547,44 @@ pub struct PokemonService<S> {
...
+Although this provides a reasonably "complete" API, it can be cumbersome in some use cases. Suppose a customer wants to log the operation name when a request is routed to said operation. Writing a `Layer`, `NameLogger`, to log a operation name is simple, however with the current API the customer is forced to do the following
+
+```rust
+let get_pokemon_species = GetPokemonSpecies::from_handler(/* handler */).layer(NameLogger::new("GetPokemonSpecies"));
+let get_storage = GetStorage::from_handler(/* handler */).layer(NameLogger::new("GetStorage"));
+let do_nothing = DoNothing::from_handler(/* handler */).layer(NameLogger::new("DoNothing"));
+/* Repeat for every route... */
+```
+
+Note that `PokemonService::layer` cannot be used here because it applies a _single_ layer uniformly across all `Route`s stored in the `Router`.
```
LukeMathWalker, line 569: If I understood the machinery correctly, this is not necessarily the problem per se.
Our problem is that the layer we are applying does not access to the `Operation` struct, therefore it has no way to ergonomically retrieve the data that we want in this example (i.e. the name of the operation).

hlbarber: Yeah, by the time `PokemonService` exists we've already wrapped them in a `Route` and collected them into some `Router` implementation.

We currently don't hold any additional information other than the model service `S` and the layer `L` in the `Operation`, so if the `Layer` did have access to the specific `Operation` that had been bundled into the `Route` it still wouldn't be able to make any judgements about the name. Furthermore, suppose a `FailOnMissing` had been the source of the `Route`, there's no operation name for it.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,755 @@
...
+@title("Pokémon Service")
+@restJson1
+service PokemonService {
+    version: "2021-12-01",
+    resources: [PokemonSpecies, Storage],
+    operations: [
+        GetServerStatistics,
+        DoNothing,
+        CapturePokemon,
+        CheckHealth
```
jjant, line 20: The `GetPokemonSpecies` operation is not listed here, but is used throughout the examples, that's a bit confusing.

design/src/docs/anatomy.md
```diff
@@ -0,0 +1,755 @@
...
+```rust
+/// Retrieve information about a Pokémon species.
+pub struct GetPokemonSpecies;
+
+impl OperationShape for GetPokemonSpecies {
+    const NAME: &'static str = "com.aws.example#GetPokemonSpecies";
+
+    type Input = GetPokemonSpeciesInput;
+    type Output = GetPokemonSpeciesOutput;
+    type Error = GetPokemonSpeciesError;
```
jjant, line 91: This type, `GetPokemonSpeciesError`, doesn't appear in the smithy file, I assume it's a generated (rust) enum of the possible errors, but maybe it could use an introduction?

design/src/docs/anatomy.md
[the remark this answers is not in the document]
hlbarber: How're these?

https://github.com/awslabs/smithy-rs/pull/1782/commits/e40aff985957469a0de6e171050318384fc24c81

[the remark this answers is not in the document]
LukeMathWalker: Looking sharp!

crisidev (approved):
Really good stuff here Harry!

design/src/server/anatomy.md
```diff
@@ -0,0 +1,843 @@
...
+        }
+    }
+}
+```
+
+Here `PluginStack` works in a similar way to [`tower::layer::util::Stack`](https://docs.rs/tower/latest/tower/layer/util/struct.Stack.html) - allowing users to stack a new plugin rather than replacing the currently set one.
+
+## Accessing Unmodelled Data
+
+An additional omitted detail is that we provide an "escape hatch" allowing `Handler`s and `OperationService`s to accept data that isn't modelled. In addition to accepting `Op::Input` they can accept additional arguments which implement the [`FromParts`](https://github.com/awslabs/smithy-rs/blob/4c5cbc39384f0d949d7693eb87b5853fe72629cd/rust-runtime/aws-smithy-http-server/src/request.rs#L114-L121) trait:
```
jjant, line 815: Do we have any example of using `FromParts`? If so, it would probably be useful to link it here.

jjant (approved):
Great work @hlbarber!

PR #1803, opened 2022-10-03
Do not generate public empty modules

LukeMathWalker:
## Motivation and Context
When experimenting with `smithy-rs` I noticed that the generated server crate exposes two public empty modules - `operation` and `operation_handler`. 

## Description
Make sure that `operation` and `operational_handler` are private modules in the generated server crate, while keeping `operation` public for the generated client crate.

## Testing
I verified the generated server and client code via `./gradlew codegen-client-test:test` and `./gradlew codegen-server-test:test`.

## Checklist

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
@@ -10,6 +10,7 @@ import software.amazon.smithy.rust.codegen.core.rustlang.GenericTypeArg
 import software.amazon.smithy.rust.codegen.core.rustlang.RustGenerics
 import software.amazon.smithy.rust.codegen.core.rustlang.RustModule
 import software.amazon.smithy.rust.codegen.core.rustlang.RustWriter
+import software.amazon.smithy.rust.codegen.core.rustlang.Visibility
 import software.amazon.smithy.rust.codegen.core.rustlang.asType
 import software.amazon.smithy.rust.codegen.core.rustlang.docs
 import software.amazon.smithy.rust.codegen.core.rustlang.rust
@@ -34,7 +35,7 @@ class CustomizableOperationGenerator(
     private val smithyTypes = CargoDependency.SmithyTypes(runtimeConfig).asType()
 
     fun render(crate: RustCrate) {
-        crate.withModule(RustModule.Operation) { writer ->
+        crate.withModule(RustModule.operationModule(Visibility.PUBLIC)) { writer ->
             writer.docs("Operation customization and supporting types")
             writer.rust("pub mod customize;")
         }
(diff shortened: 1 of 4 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustModule.kt
```diff
@@ -25,6 +25,12 @@ data class RustModule(val name: String, val rustMetadata: RustMetadata, val docu
 
         val Config = public("config", documentation = "Configuration for the service.")
         val Error = public("error", documentation = "All error types that operations can return.")
-        val Operation = public("operation", documentation = "All operations that this crate can perform.")
+
+        /**
+         * Helper method to generate the `operation` Rust module.
+         * Its visibility depends on the generation context (client or server).
+         */
+        fun operationModule(visibility: Visibility): RustModule =
```
jjant, line 33: I don't think you need `Module` in the name here (it's already used as `RustModule. ...`).

LukeMathWalker: I added the suffix to stress that it's not a "general purpose" constructor (like `public` or `private` or `default`), but it might have been too heavy handed. I'll get rid of it.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
```diff
@@ -34,7 +35,7 @@ class CustomizableOperationGenerator(
     private val smithyTypes = CargoDependency.SmithyTypes(runtimeConfig).asType()
 
     fun render(crate: RustCrate) {
-        crate.withModule(RustModule.Operation) { writer ->
+        crate.withModule(RustModule.operation(Visibility.PUBLIC)) { writer ->
```
LukeMathWalker, line 38: I would have expected this to be enough to get the module to show up in `lib.rs` as `pub mod operation`, but apparently it wasn't - I had to add https://github.com/awslabs/smithy-rs/pull/1803/files#diff-f7da198414ffe3cefc591164d3b1bb41d67e22812d7ba3a86b8abd6d798cec49R87
I'd be curious to understand why.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CodegenDelegator.kt
```diff
@@ -175,7 +175,6 @@ open class RustCrate(
  */
 val DefaultPublicModules = setOf(
```
jdisanti, line 176: I think a better approach is to move `DefaultPublicModules` into the client and server modules, since it doesn't seem like there's a good reason to share it in `codegen-core`. This would also allow us to tailor the module docs to the client/servers separately.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
jdisanti: From reading the code, it definitely seems like it should emit the `mod` statement when it's missing from `DefaultPublicModules`.

PR #1807, opened 2022-10-03
Lambda support for Python servers

unexge:
This PR exposes `app.run_lambda()` function to allow Python servers to run on Lambda. Closes https://github.com/awslabs/smithy-rs/issues/1750.

## Why we use custom Rust runtime
We could also use builtin Python runtime, I decided to use custom Rust runtime because there was some performance overheads.

If you use Python runtime: 
- Receive requests with Python (Python)
- Pass requests to Rust (Rust)
- You basically receive a JSON object in API Gateway schema and you need to convert it to Hyper Request, this is handled by `lambda_http` crate but it is not designed to be reusable and requires some modification/copying (Rust)
- Call Python handler to get result (Python)
- Convert that result to Http response (Rust)
- Pass that result to Python  (Python)
- Finally return response to Lambda runtime to complete (Python)

If you use Rust runtime:
- Receive Hyper Request with Rust (Conversion from JSON object to Hyper Request is already handled by `lambda_http` crate) (Rust)
- Call Python handler to get result (Python)
- Convert that result to Http response (Rust)
- Finally return response to Lambda runtime to complete (Rust)

## Testing

It can be tested with a `Dockerfile` similar to this:
```Dockerfile
FROM public.ecr.aws/lambda/python:3.8-x86_64

COPY app.py ${LAMBDA_TASK_ROOT}
# Shared library should be on the same folder with application code, so it can be imported
COPY lib_pokemon_service_server_sdk.so ${LAMBDA_TASK_ROOT}

RUN ln -s ${LAMBDA_TASK_ROOT}/app.py /app.py

# override `ENTRYPOINT` to not call Python runtime, so we can use Rust runtime
ENTRYPOINT [ "/var/lang/bin/python3.8" ] 
CMD [ "/app.py" ]
```

### DIFF
--- a/rust-runtime/aws-smithy-http-server-python/Cargo.toml
+++ b/rust-runtime/aws-smithy-http-server-python/Cargo.toml
@@ -22,6 +22,7 @@ bytes = "1.2"
 futures = "0.3"
 http = "0.2"
 hyper = { version = "0.14.20", features = ["server", "http1", "http2", "tcp", "stream"] }
+lambda_http = "0.6.0"
 num_cpus = "1.13.1"
 parking_lot = "0.12.1"
 pin-project-lite = "0.2"
--- a/rust-runtime/aws-smithy-http-server-python/README.md
+++ b/rust-runtime/aws-smithy-http-server-python/README.md
@@ -47,8 +47,19 @@ COPY app.py ${LAMBDA_TASK_ROOT}
 # When you build your Server SDK for your service you get a shared library
 # that is importable in Python. You need to copy that shared library to same folder
 # with your application code, so it can be imported by your application.
+# Note that you need to build your library for Linux,
+# if you are on a different platform you can consult to
+# https://pyo3.rs/latest/building_and_distribution.html#cross-compiling
+# for cross compiling.
 COPY lib_pokemon_service_server_sdk.so ${LAMBDA_TASK_ROOT}
 
+# You can install your application's dependencies using file `requirements.txt`
+# from your project folder, if you have any.
+# COPY requirements.txt  .
+# RUN  pip3 install -r requirements.txt --target "${LAMBDA_TASK_ROOT}"
+
+# Create a symlink for your application's entrypoint,
+# so we can use `/app.py` to refer it
 RUN ln -s ${LAMBDA_TASK_ROOT}/app.py /app.py
 
 # By default `public.ecr.aws/lambda/python` images comes with Python runtime,
--- a/rust-runtime/aws-smithy-http-server-python/README.md
+++ b/rust-runtime/aws-smithy-http-server-python/README.md
@@ -59,7 +59,5 @@ CMD [ "/app.py" ]
 ```
 
 <!-- anchor_start:footer -->
-
 This crate is part of the [AWS SDK for Rust](https://awslabs.github.io/aws-sdk-rust/) and the [smithy-rs](https://github.com/awslabs/smithy-rs) code generator. In most cases, it should not be used directly.
-
 <!-- anchor_end:footer -->
--- a/rust-runtime/aws-smithy-http-server-python/README.md
+++ b/rust-runtime/aws-smithy-http-server-python/README.md
@@ -2,6 +2,64 @@
 
 Server libraries for smithy-rs generated servers, targeting pure Python business logic.
 
+## Running servers on AWS Lambda
+
+`aws-smithy-http-server-python` supports running your services on [AWS Lambda](https://aws.amazon.com/lambda/).
+
+You need to use `run_lambda` method instead of `run` method to start
+the [custom runtime](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html)
+instead of the [Hyper](https://hyper.rs/) HTTP server.
+
+In your `app.py`:
+
+```diff
+from libpokemon_service_server_sdk import App
+from libpokemon_service_server_sdk.error import ResourceNotFoundException
+
+# ...
+
+# Get the number of requests served by this server.
+@app.get_server_statistics
+def get_server_statistics(
+    _: GetServerStatisticsInput, context: Context
+) -> GetServerStatisticsOutput:
+    calls_count = context.get_calls_count()
+    logging.debug("The service handled %d requests", calls_count)
+    return GetServerStatisticsOutput(calls_count=calls_count)
+
+# ...
+
+-app.run()
++app.run_lambda()
+```
+
+`aws-smithy-http-server-python` comes with a
+[custom runtime](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html)
+so you should run your service without any provided runtimes.
+You can achieve that with a `Dockerfile` similar to this:
+
+```dockerfile
+# You can use any image that has your desired Python version
+FROM public.ecr.aws/lambda/python:3.8-x86_64
+
+# Copy your application code to `LAMBDA_TASK_ROOT`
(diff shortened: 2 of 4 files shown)


rust-runtime/aws-smithy-http-server-python/src/server.rs
```diff
@@ -460,4 +455,40 @@ event_loop.add_signal_handler(signal.SIGINT,
...
+        rt.block_on(async move {
+            let handler = LambdaHandler::new(app);
+            let lambda = lambda_http::run(handler);
+            tracing::debug!("Starting Lambda server");
+            if let Err(err) = lambda.await {
+                tracing::error!("server error: {}", err);
+            }
+        });
+        Ok(())
+    }
```
crisidev, line 479: It's not really a server :)

```suggestion
    fn run_lambda_handler(&mut self, py: Python) -> PyResult<()> {
        let event_loop = self.configure_python_event_loop(py)?;
        let app = self.build_and_configure_router(py, event_loop)?;
        let rt = runtime::Builder::new_multi_thread()
            .enable_all()
            .build()
            .expect("Unable to start a new Tokio runtime for this process");
        rt.block_on(async move {
            let handler = LambdaHandler::new(app);
            let lambda = lambda_http::run(handler);
            tracing::debug!("Starting Lambda handler");
            if let Err(err) = lambda.await {
                tracing::error!("server error: {}", err);
            }
        });
        Ok(())
    }
```

rust-runtime/aws-smithy-http-server-python/README.md
```diff
@@ -2,6 +2,62 @@
...
+    _: GetServerStatisticsInput, context: Context
+) -> GetServerStatisticsOutput:
+    calls_count = context.get_calls_count()
+    logging.debug("The service handled %d requests", calls_count)
+    return GetServerStatisticsOutput(calls_count=calls_count)
+
+# ...
+
+-app.run()
++app.run_lambda()
```
crisidev, line 33: What do you think about renaming the entrypoint of the standard server from `run()` to `run_server()` ?

rust-runtime/aws-smithy-http-server-python/Cargo.toml
```diff
@@ -22,6 +22,7 @@ bytes = "1.2"
 futures = "0.3"
 http = "0.2"
 hyper = { version = "0.14.20", features = ["server", "http1", "http2", "tcp", "stream"] }
+lambda_http = "0.6.0"
```
crisidev, line 25: Not a blocker for this PR, but I would probably create an issue to feature-gate lambda in the future so that we do not impose this dependency on all customers. I am not even sure if its worth it thou.. Thoughts?

rust-runtime/aws-smithy-http-server-python/README.md
unexge: I think `run()` is clearer. I don't know if it is idiomatic but, if we decide to unify `run_lambda` and any possible `run_*` versions into a big `run` function with keyword arguments like `run(lambda=True, foo="bar", ...)` that would allow us to do without renaming it again.

rust-runtime/aws-smithy-http-server-python/Cargo.toml
unexge: Yes good point. But I'm not sure if we currently support to generate SDKs with specific features on dependencies?

rust-runtime/aws-smithy-http-server-python/README.md
crisidev: Ok, makes sense!

crisidev (approved):
This looks good and I am very happy to see we can use this from the pure python lambda context! Are you planning to add an example of the Python code here?

unexge:
> This looks good and I am very happy to see we can use this from the pure python lambda context! Are you planning to add an example of the Python code here?

By example code you mean like having Lambda example of Pokemon service? If so, I think there is no need for that because only code change is changing `app.run()` to `app.run_lambda()`. What do you think?

rust-runtime/aws-smithy-http-server-python/src/server.rs
```diff
@@ -460,4 +455,40 @@ event_loop.add_signal_handler(signal.SIGINT,
...
+        let event_loop = self.configure_python_event_loop(py)?;
+        let app = self.build_and_configure_router(py, event_loop)?;
+        let rt = runtime::Builder::new_multi_thread()
+            .enable_all()
+            .build()
+            .expect("Unable to start a new Tokio runtime for this process");
+        rt.block_on(async move {
+            let handler = LambdaHandler::new(app);
+            let lambda = lambda_http::run(handler);
+            tracing::debug!("Starting Lambda handler");
```
hlbarber, line 473: I think the general consensus in Rust ecosystem is to use lowercase log messages.

https://docs.rs/env_logger/latest/env_logger/#example
https://github.com/rust-lang/rust/issues/442

rust-runtime/aws-smithy-http-server-python/src/server.rs
```diff
@@ -460,4 +455,40 @@ event_loop.add_signal_handler(signal.SIGINT,
...
+    ///
+    /// Unlike the `run_server`, `run_lambda_handler` does not spawns other processes,
+    /// it starts the Lambda handler on the current process.
+    fn run_lambda_handler(&mut self, py: Python) -> PyResult<()> {
+        let event_loop = self.configure_python_event_loop(py)?;
+        let app = self.build_and_configure_router(py, event_loop)?;
+        let rt = runtime::Builder::new_multi_thread()
+            .enable_all()
+            .build()
+            .expect("Unable to start a new Tokio runtime for this process");
```
hlbarber, line 469: https://github.com/rust-lang/api-guidelines/issues/79#issue-232073011

rust-runtime/aws-smithy-http-server-python/src/server.rs
```diff
@@ -460,4 +455,40 @@ event_loop.add_signal_handler(signal.SIGINT,
...
+        rt.block_on(async move {
+            let handler = LambdaHandler::new(app);
+            let lambda = lambda_http::run(handler);
+            tracing::debug!("Starting Lambda handler");
+            if let Err(err) = lambda.await {
+                tracing::error!("Lambda handler error: {}", err);
```
hlbarber, line 475: Perhaps this should be:

```rust
tracing::error!(error = %err, "unable to start lambda handler");
```

rust-runtime/aws-smithy-http-server-python/src/server.rs
```diff
@@ -460,4 +455,40 @@ event_loop.add_signal_handler(signal.SIGINT,
...
+    // Builds the router and adds necessary layers to it.
+    fn build_and_configure_router(
+        &mut self,
+        py: Python,
+        event_loop: &pyo3::PyAny,
+    ) -> pyo3::PyResult<Router> {
+        let app = self.build_router(event_loop)?;
+        // Create the `PyState` object from the Python context object.
+        let context = self.context().clone().unwrap_or_else(|| py.None());
+        tracing::debug!("Add middlewares to Rust Python router");
```
hlbarber, line 490: Same as above.

hlbarber (approved):

crisidev (approved):

rust-runtime/aws-smithy-http-server-python/src/server.rs
unexge: I've changed messages introduced in this PR and created https://github.com/awslabs/smithy-rs/issues/1838 for the rest

crisidev (approved):

PR #1811, opened 2022-10-05
Support 32bits targets

### LINKED ISSUE #1810: Support compiling aws-smithy-client on 32bit CPU architecture
https://github.com/awslabs/smithy-rs/blob/6322569ea7e1f0a43e380b859c1cfabfcab92a5d/rust-runtime/aws-smithy-client/src/never.rs#L30

This module is using an `AtomicU64`, which is preventing the client to be compilable on 32bit system. I propose to change this to use an `AtomicUsize`, but I am not sure if there are any implication or explicit decision to not support 32bit architectures.

LukeMathWalker:
## Motivation and Context
Fixes #1810 

## Description
We replace all `AtomicU64` instances with `AtomicUsize`.
This is a breaking change for `aws-smithy-client` since `NeverService::num_call` is exposed via its public API. 

There are some minor tweaks in the test suite of the Python server to make sure that the test suite compiles on a MacOS machine.

## Checklist


rust-runtime/aws-smithy-http-server-python/src/socket.rs
```diff
@@ -73,6 +73,14 @@ impl PySocket {
...
 
 #[cfg(test)]
+// `is_listener` on `Socket` is only available on certain platforms.
+// In particular, this fails to compile on MacOS.
+#[cfg(any(
+    target_os = "android",
+    target_os = "freebsd",
+    target_os = "fuchsia",
+    target_os = "linux",
+))]
```
crisidev, line 83: Good catch.

crisidev (approved):

crisidev:
I did not want to approve this, since it is still WIP, but I clicked the wrong button :(

crisidev:
How do I remove the approval? Anyway, not a big deal. We need @jdisanti approval here.

LukeMathWalker:
I put it as WIP because I am researching the CI side of things, but I'll probably just open a separate PR - that's likely to take longer and be somewhat hairy (QEMU 😢 ).

CHANGELOG.next.toml
```diff
@@ -11,6 +11,22 @@
...
+meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "server"}
+author = "LukeMathWalker"
+
+[[smithy-rs]]
+message = """
+Replace all usages of `AtomicU64` with `AtomicUsize` to support 32bit targets.
+"""
+references = ["smithy-rs#1811"]
+meta = { "breaking" = true, "tada" = false, "bug" = false, "target" = "client"}
+author = "LukeMathWalker"
```
LukeMathWalker, line 28: It's breaking only for the client - let me know if there are better ways to convey it.

LukeMathWalker:
I assumed the intention was to have it public as a utility to be used by customer for testing their own code.
If that's the case, it'd probably still makes sense to have it behind a dedicated feature flag (e.g. `test_utils`).

crisidev:
I think Jon can answer this question.. For the time being I would just fix the issue with 32bit targets and address the possibility of moving everything under `#[cfg(test]` flag. Same goes for improving the CI.

jjant (approved):

crisidev:
> I see @jdisanti. In the interest of unblocking the customers on `powerpc` - should we move forward with this PR and track the fate of `NeverService` in a separate issue?

I think it is fair to track `NeverService` in a separate issue.

crisidev (approved):

PR #1812, opened 2022-10-05
Add a new CI job to check that smithy-rs compiles on 32bit Linux-based targets

LukeMathWalker:
## Motivation and Context
Ensure we do not regress on #1810 

## Description
This initial PR makes sure that `smithy-rs` compiles on a few 32bits Linux-based targets.
We do not attempt to run tests on those targets as it would require a more complicated setup via emulation, which (from what I could gather) would be slow to execute and impact our overall CI completion time.
This puts our confidence in correct execution on 32bit targets a bit lower than our confidence when it comes to Windows.

We successfully reproduced the failure experienced by our customers on `powerpc` (see [run](https://github.com/awslabs/smithy-rs/actions/runs/3190707127/jobs/5206196058#logs)).

This PR depends on the changes in #1811, so it should be merged afterwards.


rust-runtime/aws-smithy-client/src/builder.rs
```diff
@@ -135,7 +135,7 @@ impl<M, R> Builder<(), M, R> {
         let with_https = |b: Builder<_, M, R>| b.rustls_connector(connector_settings);
         // If we are compiling this function & rustls is not enabled, then native-tls MUST be enabled
         #[cfg(not(feature = "rustls"))]
-        let with_https = |b: Builder<_, M, R>| b.native_tls_connector();
+        let with_https = |b: Builder<_, M, R>| b.native_tls_connector(connector_settings);
```
LukeMathWalker, line 138: A little bug that sneaked through since we use `--all-features` for testing, I assume.

.github/workflows/ci-pr.yml
```diff
@@ -5,6 +5,7 @@
 
 name: CI
 on:
+  workflow_dispatch:
```
jdisanti, line 8: What is this for?

rust-runtime/aws-smithy-client/src/builder.rs
jdisanti: Hmmm... Looks like we forgot to add `cargo hack` to `aws-smithy-client`'s `additional-ci` script. Could you add that so this doesn't happen again?

Example: https://github.com/awslabs/smithy-rs/blob/0fb5ccb81446b76cb20dd26c0e39e0cbc3965b8d/rust-runtime/aws-smithy-eventstream/additional-ci#L15

.github/workflows/ci-pr.yml
LukeMathWalker: It gives you the possibility to trigger the workflow from GitHub's UI.
See `ci-main`, which already has it:
image

rust-runtime/aws-smithy-client/src/builder.rs
LukeMathWalker: Done!

crisidev (approved):

rust-runtime/aws-smithy-client/additional-ci
```diff
@@ -10,3 +10,6 @@ set -e
 
 echo "### Checking for duplicate dependency versions in the normal dependency graph with all features enabled"
 cargo tree -d --edges normal --all-features
+
+"### Testing every combination of features (excluding --all-features)"
```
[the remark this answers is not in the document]
LukeMathWalker: Great catch, thank you ❤️

LukeMathWalker:
It looks like `aws-smithy-client` has a few more feature-related bugs to squash.

.github/workflows/ci-pr.yml
jdisanti: But why would you need this for `ci-pr.yml`? It runs every time a PR is updated.

And if it is run via workflow dispatch, how does it get values needed by the PR bot:
```
      issue_number: ${{ github.event.number }}
      base_revision: ${{ github.event.pull_request.base.sha }}
      head_revision: ${{ github.event.pull_request.head.sha }}
```

jdisanti (approved):

PR #1813, opened 2022-10-05
Add Server section to the Book

hlbarber:
## Motivation and Context

This is a _partial_ solution to https://github.com/awslabs/smithy-rs/issues/1777.

## Description

- Move all server documentation to a `Server` chapter in the book.
- Provide an `overview.md` for the chapter.
- Fix some lints in the documentation.

## Notes

- This is partial solution in the sense that we do _not_ go through the existing documentation and separate out client/core documentation.
- The following work remains:
    - Flesh out the `server/overview.md`, can we create a narrative here?
    - Go through the existing documentation and divide it into core/client.
    - Remove Rust SDK as the primary subject.


82marbag (approved):

design/src/server/pokemon_service.md
```diff
@@ -1,4 +1,4 @@
-## Code generating the Pokémon Service
+# Code Generating the Pokémon Service
```
jjant, line 1: Nitpick: In `design/src/server/code_generation.md` we use `Generating xyz Code`, but here we use `Code Generating the xyz`.

design/src/server/overview.md
```diff
@@ -0,0 +1,7 @@
+# Smithy Server
+
+Smithy Rust provides the ability to generate a server whose stubs are provided by the customer.
```
Velfi, line 3: What are stubs?

Velfi (approved):

PR #1819, opened 2022-10-06
Add RFC for Error Context and Compatibility

jdisanti:
This RFC proposes a pattern for writing Rust errors to provide consistent error context and forwards/backwards compatibility.

[Rendered](https://github.com/awslabs/smithy-rs/blob/jdisanti-error-rfc/design/src/rfcs/rfc0022_error_context_and_compatibility.md)


design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+inaccessible to the reporter).
+
+This situation might improve if the nightly `request_value`/`request_ref`/`provide` functions on
+`std::error::Error` are stabilized, since then contextual information needed for including things
+such as a backtrace could still be retrieved through the opaque error new-type.
+
+This RFC proposes that `Box<dyn Error + Send + Sync + 'static>` is used to refer to errors from libraries.
+An allocation to store this can be avoided if the error source is private, in which case, a reference to it
+can be returned from the `Error::source` implementation. Otherwise, if it's directly exposed as a public enum
+field, then it must be boxed.
```
jdisanti, line 209: I'm not confident that I got this right. Perhaps we should take a harder stance and make a rule to not directly expose the source error (outside of the `Error::source` call) so that we can replace it with an opaque error type later when that becomes more practical. The downside is that every enum variant that has a source error will need to have a separate struct in the same namespace in order to hide it.

Would love to get more input on this.

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+> Applies to: Generated clients and shared rust-runtime crates
+
+This RFC proposes a pattern for writing Rust errors to provide consistent
+error context AND forwards/backwards compatibility. The goal is to strike
+a balance between these four goals:
+
+1. Errors are forwards compatible, and changes to errors are backwards compatible
+2. Errors are ergonomic
+3. Error messages are easy to debug
+4. Errors adhere to Rust's `Error` trait
```
Velfi, line 15: did you mean:
```suggestion
4. Errors implement Rust's `Error` trait
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+> Status: RFC
+>
+> Applies to: Generated clients and shared rust-runtime crates
+
+This RFC proposes a pattern for writing Rust errors to provide consistent
+error context AND forwards/backwards compatibility. The goal is to strike
+a balance between these four goals:
+
+1. Errors are forwards compatible, and changes to errors are backwards compatible
+2. Errors are ergonomic
```
Velfi, line 13: I think it would be helpful to be more specific here somehow. What does it mean for an error to be ergonomic? Does it have helpful methods? Is it easy to match on? Is it easy to figure out how I can use it to understand the problem that necessitated it?

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+1. Using `#[non_exhaustive]` on the enum allows new errors to be added in the future.
+2. Breaking out different error types allows for more useful error messages, potentially with error-specific context.
+   Customers can match on these different error variants to change their program flow, although it's not immediately
+   obvious if such use cases exist for this error.
+4. The error cause is available through the `Error::source()` impl for variants that have a cause.
+
+However, there are also a number of things that could be improved:
+1. All tuple/struct enum members are public, and `InvalidUri` is an error from the `http` crate.
+   Exposing a type from another crate can potentially lock the GA SDK into a specific crate version
+   if breaking changes are ever made to the exposed types.
```
Velfi, line 71: Not to mention, this makes it more annoying to use something like Surf which uses a different HTTP library

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+2. `DnsLookupFailed` is missing `#[non_exhaustive]`, so new members can never be added to it.
+3. Use of enum tuples, even with `#[non_exhaustive]`, adds friction to evolving the API since the
+   tuple members cannot be named.
```
Velfi, line 74: This is suuuuuuuch a good callout and one that didn't even occur to me

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+4. Printing the source error in the `Display` impl leads to error repetition by reporters
+   that examine the full source chain.
+5. The `source()` impl has a `_` matcher, which means implementers may forget to propagate
+   a source when adding a new variant.
```
Velfi, line 78:
```suggestion
5. The `source()` impl has a `_` match arm, which means future implementers could forget to propagate
   a source when adding new error variants.
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+6. The error source can be downcasted to `InvalidUri` type from `http` in customer code. This is
+   a leaky abstraction where customers can start to rely on the underlying library the SDK uses
+   in its implementation, and if that library is replaced/changed, it can silently break the
+   customer's application. _Note:_ later in the RFC, we'll see that fixing this issue is not practical.
```
Velfi, line 82:
```suggestion
   customer's application. _Note:_ later in the RFC, I'll demonstrate why fixing this issue is not practical.
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+  debuggable thanks to the flexible messaging
+
+What could be improved:
+- It needlessly implements `Clone`, which may prevent it from holding
+  an error source in the future since errors are often not `Clone`.
+- In the future, if more error variants are needed, a private inner error
+  kind enum could be added to change messaging, but there's no way to expose
+  information specific to the new error variant to the customer.
+- Programmatic access to the error `Location` may be desired, but
+  this can be trivially added in the future without a breaking change.
```
Velfi, line 121: Added how? By making the field public?

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+    ObjectAlreadyInActiveTierError(ObjectAlreadyInActiveTierError),
+    ObjectNotInActiveTierError(ObjectNotInActiveTierError),
+    Unhandled(Box<dyn Error + Send + Sync + 'static>),
+}
+```
+
+Each error variant gets its own struct, which can hold error-specific contextual information.
+Both the error enum and the details on each variant are extensible.
+
+The code generated errors are already aligned with the goals for this RFC.
```
Velfi, line 146:
```suggestion
The code generated errors are already aligned with the goals of this RFC.
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+create an opaque error wrapping new-type that removes the ability to downcast to the other library's error.
+This, however, can be circumvented via unsafe code, and also breaks the ability for error reporters to
+properly display the error (for example, if the error has backtrace information, that would be
+inaccessible to the reporter).
+
+This situation might improve if the nightly `request_value`/`request_ref`/`provide` functions on
+`std::error::Error` are stabilized, since then contextual information needed for including things
+such as a backtrace could still be retrieved through the opaque error new-type.
+
+This RFC proposes that `Box<dyn Error + Send + Sync + 'static>` is used to refer to errors from libraries.
```
Velfi, line 206:
```suggestion
This RFC proposes that `Box<dyn Error + Send + Sync + 'static>` is used to refer to errors from external libraries.
```

Velfi: Maybe even change "to refer to" to "to type-erase"

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+An allocation to store this can be avoided if the error source is private, in which case, a reference to it
+can be returned from the `Error::source` implementation. Otherwise, if it's directly exposed as a public enum
+field, then it must be boxed.
+
+Ideally, errors should contain enough useful information on them that downcasting the underlying cause
+is not necessary. Customers needing to downcast the error source should be a last resort, and with the
```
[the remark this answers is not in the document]
Velfi: Maybe we could use an error type review checklist with questions like "is this error helpful without downcasting" that PR reviewers can use

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+understanding that the type could change at a later date with no compile-time guarantees.
+
+Error Proposal
+--------------
+
+Taking a customer's perspective, there are two broad categories of errors:
+
+1. **Actionable:** Errors that can/should influence program flow
+2. **Informative:** Errors that inform that something went wrong, but where
+   it's not useful to match on the error to change program flow
```
Velfi, line 222: It might be helpful to somehow relate this to the more specific cases of:
- You got a `Result::Err(e)` and your program flow depends on `e`
- You got a `Result::Err(_)` and your program flow depends on `Result::Err`

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+
+This RFC proposes that a consistent pattern be introduced to cover these two use cases for
+all errors in the public API for the Rust runtime crates and generated client crates.
+
+### Actionable error pattern
+
+Actionable errors must be represented as enums. If an error variant has an error source or additional
+contextual information, it must use a struct that is either embedded inside the enum, or separate where it makes
+sense to have private fields. For example:
```
Velfi, line 231: I tried to make this more concise but I may have gone too far

```suggestion
Actionable errors are represented as enums. Struct-like enum variants may contain an error source or additional
context. When private fields are required, a tuple-like enum variant is used to wrap a single struct. For example:
```

Velfi: btw, the "struct-like" and "tuple-like" language is lifted directly from https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html?highlight=struct%20variant#destructuring-enums

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+pub enum Error {
+    // Good
+    #[non_exhaustive]
+    VariantA,
+
+    // Good
+    #[non_exhaustive]
+    VariantB {
+        some_additional_info: u32,
+        source: AnotherError // AnotherError is from this crate
```
Velfi, line 243:
```suggestion
        source: AnotherError // This field is public, but AnotherError is from this crate
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+
+### Actionable error pattern
+
+Actionable errors must be represented as enums. If an error variant has an error source or additional
+contextual information, it must use a struct that is either embedded inside the enum, or separate where it makes
+sense to have private fields. For example:
+
+```rust
+pub enum Error {
+    // Good
```
Velfi, line 235: We couldn't make this a tuple or struct-like enum later, right?
```suggestion
    // Good, but not extensible
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+    VariantA,
+
+    // Good
+    #[non_exhaustive]
+    VariantB {
+        some_additional_info: u32,
+        source: AnotherError // AnotherError is from this crate
+    },
+
+    // Good: This is exhaustive and uses a tuple, but its sole member is an extensible struct
```
Velfi, line 246:
```suggestion
    // Good: This is exhaustive and uses a tuple, but its sole member is an extensible struct with private fields
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+    VariantF {
+        source: http::uri::InvalidUri
+    },
+
+    // Acceptable: Another library's error type may be in the box, and can be
+    // downcasted to, but the customer is then doing this downcasting at their
+    // own risk with no type safety guarantees that it will continue to work.
+    //
+    // Ideally, there would be a way to do opaque errors while playing
+    // nicely with fancy error reporters.
```
Velfi, line 265: I think you may be able to be more specific here but I'm not sure. I think I just have a bias against using the word "ideally". I'd rather see a definite statement.

When? Once an RFC is merged?
How? By extending it?

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+    #[non_exhaustive]
+    VariantG {
+        source: Box<dyn Error + Send + Sync + 'static>,
+    }
+}
+
+pub struct VariantC { some_private_field: u32 }
```
Velfi, line 272: You know how you can define associated types and `const`s in traits? It'd be cool if you could define a struct inside an enum with using a struct enum variant, just to show that they're intentionally linked together forever.

Velfi:
```rust
pub enum MyError {
    struct A {
        some_private_field: String,
    }
    
    VariantA(A)
}
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+
+impl Error for InformativeError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        Some(&self.source)
+    }
+}
+```
+
+In general, informative errors should be referenced by variants in actionable errors since they cannot be converted
+to adaptive errors at a later date without a breaking change. This is not a hard rule, however. Use your best judgement
```
Velfi, line 332: Adaptive errors? Is that a typo or a new term?

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,370 @@
...
+for the situation.
+
+### Displaying full error context
+
+In places where errors are logged rather than returned to the customer, the full error source chain
```
Velfi, line 337:
```suggestion
In code where errors are logged rather than returned to the customer, the full error source chain
```

design/src/rfcs/rfc0022_error_context_and_compatibility.md
jdisanti: To summarize an offline discussion with @Velfi: We agreed the error source should only be available through `Error::source`, and not via a public enum field. This means that each enum struct needs to be an actual struct so that it can have private members. To avoid obscuring the docs with a bunch of error structs, each module that has errors will need to have those errors moved into a new `error` submodule so that the error structs are grouped together and can't be confused with other non-error structs in the documentation.

design/src/rfcs/rfc0022_error_context_and_compatibility.md
jdisanti: Implementing `Error` is a given. My intent here was to communicate that they should adhere to the design/standards/best practices established by the `Error` trait. For example, the optional `source()` function should be implemented where possible.

design/src/rfcs/rfc0022_error_context_and_compatibility.md
jdisanti: Looks like you're right. Making it a tuple or struct later leads to compiler errors on existing match statements.

design/src/rfcs/rfc0022_error_context_and_compatibility.md
jdisanti: That would be nice.

design/src/rfcs/rfc0022_error_context_and_compatibility.md
jdisanti: It's a typo.

Velfi (approved):

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,402 @@
...
+```
+
+What this error does well:
+- The members are private, so `#[non_exhaustive]` isn't even necessary
+- The error is completely opaque (maximizing compatibility) while still being
+  debuggable thanks to the flexible messaging
+
+What could be improved:
+- It needlessly implements `Clone`, which may prevent it from holding
+  an error source in the future since errors are often not `Clone`.
```
jdisanti, line 119: The change checklist should track removing this `Clone` impl

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,402 @@
...
+```
+
+In general, informative errors should be referenced by variants in actionable errors since they cannot be converted
+to actionable errors at a later date without a breaking change. This is not a hard rule, however. Use your best judgement
+for the situation.
+
+### Displaying full error context
+
+In code where errors are logged rather than returned to the customer, the full error source chain
+must be displayed. This will be made easy by placing a `DisplayErrorContext` struct in `aws-smithy-types` that
```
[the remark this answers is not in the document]
jdisanti: Hadn't considered including that in `DisplayErrorContext`, but it seems like a good idea (it could follow the full error message in parenthesis or something).

design/src/rfcs/rfc0022_error_context_and_compatibility.md
```diff
@@ -0,0 +1,405 @@
...
+is used as a wrapper to get the better error formatting:
```
rcoh, line 355: one thing that I wish we were able to better support was retroactively adding context to an existing error—I feel like that's a very common pattern that would improve the errors we're able to give to customers & reduce the need to get debug logs when resolving issues

rcoh (commented):
Overall, I like the direction here. I feel like there are two big pieces missing from the RFC:
1. This RFC doesn't ever talk about the experience that customers have actually dealing with these errors or ways we can make the errors easy (or hard) for customers to interact with. I'd love to see code examples that showed what it's like to interact with these errors.

My hope is that by digging into this more directly, we'll be able to settle on a key set of utilities (`is_foo()` maybe?) that will make errors easy for customers to interact with while maintaining all of our backwards compatibility
2. Other than careful code review, I think we need a mechanism to drive consistency here, otherwise I suspect even after a large effort to _remove_ divergence, things will diverge again.

PR #625, opened 2021-08-02
Cleanups in smithy_http::Operation

rcoh:
Misc cleanups in PropertyBag & comments.

No changelog needed.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1821, opened 2022-10-06
Env vars and profile to configure IMDS retry and timeouts

kevinpark1217:
## Motivation and Context

This is a PR for the [issue #625](https://github.com/awslabs/aws-sdk-rust/issues/625) in `aws-sdk-rust` repo.

## Description

Adds the ability to configure retry logic using profile and environment variables.
https://docs.aws.amazon.com/sdkref/latest/guide/feature-ec2-instance-metadata.html

## Checklist

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -137,3 +137,24 @@ message = "Fix regression where `connect_timeout` and `read_timeout` fields are 
 references = ["smithy-rs#1822"]
 meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "kevinpark1217"
+
+[[aws-sdk-rust]]
+message = """
+It is now possible to configure IMDS retry and timeout with environment variables and profile keys.
+See [Amazon EC2 instance metadata](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ec2-instance-metadata.html) doc for more information.
+
+New environment variables:
+```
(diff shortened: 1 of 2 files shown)


kevinpark1217:
I can also update, so that `EndpointSource` struct also relies on the parsing logic of `Environment`. But I would like initial feedbacks first.

kevinpark1217:
@Velfi It's ready for review 😄

kevinpark1217:
Hey @Velfi, has there been any internal discussion on supporting these environment variables?

I still think it is valuable in order to configure applications that relies on AWS SDK under the hood. Customers can deploy third-party products into their own "on-prem" AWS account where they don't have direct access to the source code and limited configuration.

kevinpark1217:
Also the example code does not work because, the retry and timeout is set as while building the `Client` struct, which is a layer below the `ImdsCredentialsProvider`.

There is just so many layers of configuration just to reach IMDS Client configuration that it's cumbersome to configure IMDS retry and timeout through code.

jdisanti:
The example should work after the next SDK release since the bug that was preventing the config from applying has been fixed.

kevinpark1217:
@jdisanti Sorry, that is not what I meant.

Other than the bug fix that prevented setting timeouts in code, I also wanted the ability to implement them using environment variables and profile because IMDS client is hidden away in layers of credential provider structs.

But, I think I may have found another way that works for me. Please check out #1867.

With #1867 , then I can live without the new environment variables.

PR #1824, opened 2022-10-06
Remove `handleRustBoxing` config from `SymbolVisitor`

jdisanti:
Not really sure why the `handleRustBoxing` config value existed since it is always true. Happy to close this if there's a good reason.


david-perez (approved):
Yes. I've wondered the same in the past too.

Velfi (approved):

crisidev (approved):

PR #1827, opened 2022-10-07
Pascal case operation names

### LINKED ISSUE #1826: Operation input and output struct names are pascal cased, but operation error struct names are not
As of writing, this diff atop e78da55:

```smithy
diff --git a/codegen-core/common-test-models/naming-obstacle-course-ops.smithy b/codegen-core/common-test-models/naming-obstacle-course-ops.smithy
index 087d99b75..f54b27e76 100644
--- a/codegen-core/common-test-models/naming-obstacle-course-ops.smithy
+++ b/codegen-core/common-test-models/naming-obstacle-course-ops.smithy
@@ -5,6 +5,7 @@ use smithy.test#httpRequestTests
 use aws.protocols#awsJson1_1
 use aws.api#service
+use smithy.framework#ValidationException

 /// Confounds model generation machinery with lots of problematic names
 @awsJson1_1
@@ -41,17 +42,20 @@ service Config {
     }
 ])
 operation ReservedWordsAsMembers {
-    input: ReservedWords
+    input: ReservedWords,
+    errors: [ValidationException],
 }

 // tests that module names are properly escaped
 operation Match {
     input: ReservedWords
+    errors: [ValidationException],
 }

 // Should generate a PascalCased `RpcEchoInput` struct.
 operation RPCEcho {
     input: ReservedWords
+    errors: [ValidationException],
 }

 structure ReservedWords {
```

makes `codegen-server-test:build -P modules="naming_test_ops" -P cargoCommands="test"` yield:

```
error[E0412]: cannot find type `RpcEchoError` in module `crate::error`
   --> naming_test_ops/rust-server-codegen/src/operation_shape.rs:130:32
    |
130 |     type Error = crate::error::RpcEchoError;
    |                                ^^^^^^^^^^^^ help: an enum with a similar name exists: `RPCEchoError`
    |
   ::: naming_test_ops/rust-server-codegen/src/error.rs:326:1
    |
326 | pub enum RPCEchoError {
    | --------------------- similarly named enum `RPCEchoError` defined here
```

This is due to the fact that the operation error struct name is generated as:

https://github.com/awslabs/smithy-rs/blob/e78da5598a2d94d23fd18e30492a20ce3603f844/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/error/CombinedErrorGenerator.kt#L74

So it's _not_ pascal cased; `RPCEchoError` is generated. However, `ServerOperationGenerator` pascal cases the operation name here:

https://github.com/awslabs/smithy-rs/blob/e78da5598a2d94d23fd18e30492a20ce3603f844/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationGenerator.kt#L33-L33

Note that we _do_ pascal case operation inputs and outputs, since these are regular `Structure` shapes, and so `SymbolVisitor` pascal cases them here:

https://github.com/awslabs/smithy-rs/blob/e78da5598a2d94d23fd18e30492a20ce3603f844/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt#L310

My preference to fix this bug would be to just pascal case operation names; that's a breaking change for the clients, since it uses the operation names as-is in `operation.rs`. I'm assuming we don't rename operation names for a reason?

david-perez:
And hence pascal case operation error struct names too, like we're
currently doing with operation input and output struct names.

Fixes #1826.

## Checklist

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt
@@ -284,16 +284,10 @@ open class SymbolVisitor(
         TODO("Not yet implemented: https://github.com/awslabs/smithy-rs/issues/312")
     }
 
-    override fun operationShape(shape: OperationShape): Symbol {
-        return symbolBuilder(
-            RustType.Opaque(
-                shape.contextName(serviceShape)
-                    .replaceFirstChar { it.uppercase() },
-            ),
-        )
+    override fun operationShape(shape: OperationShape) =
+        symbolBuilder(RustType.Opaque(shape.contextName(serviceShape).toPascalCase()))
             .locatedIn(Operations)
             .build()
-    }
 
     override fun resourceShape(shape: ResourceShape?): Symbol {
         TODO("Not yet implemented: resources are not supported")
--- a/codegen-core/common-test-models/naming-obstacle-course-ops.smithy
+++ b/codegen-core/common-test-models/naming-obstacle-course-ops.smithy
@@ -5,6 +5,7 @@ use smithy.test#httpRequestTests
 use smithy.test#httpResponseTests
 use aws.protocols#awsJson1_1
 use aws.api#service
+use smithy.framework#ValidationException
 
 /// Confounds model generation machinery with lots of problematic names
 @awsJson1_1
@@ -41,17 +42,20 @@ service Config {
     }
 ])
 operation ReservedWordsAsMembers {
-    input: ReservedWords
+    input: ReservedWords,
+    errors: [ValidationException],
 }
 
 // tests that module names are properly escaped
 operation Match {
     input: ReservedWords
+    errors: [ValidationException],
 }
 
 // Should generate a PascalCased `RpcEchoInput` struct.
 operation RPCEcho {
     input: ReservedWords
+    errors: [ValidationException],
 }
 
 structure ReservedWords {
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
@@ -404,7 +404,6 @@ class HttpBindingGenerator(
     /**
      * Generate a unique name for the deserializer function for a given [operationShape] and HTTP binding.
      */
-    // Rename here technically not required, operations and members cannot be renamed.
     private fun fnName(operationShape: OperationShape, binding: HttpBindingDescriptor) =
         "${operationShape.id.getName(service).toSnakeCase()}_${binding.member.container.name.toSnakeCase()}_${binding.memberName.toSnakeCase()}"
 
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/OperationNormalizer.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/OperationNormalizer.kt
@@ -32,7 +32,6 @@ import kotlin.streams.toList
  */
 object OperationNormalizer {
     // Functions to construct synthetic shape IDs—Don't rely on these in external code.
-    // Rename safety: Operations cannot be renamed
     // In order to ensure that the fully qualified smithy id of a synthetic shape can never conflict with an existing shape,
     // the `synthetic` namespace is appended.
     private fun OperationShape.syntheticInputId() =
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationGenerator.kt
@@ -30,7 +30,7 @@ class ServerOperationGenerator(
     private val symbolProvider = codegenContext.symbolProvider
     private val model = codegenContext.model
 
-    private val operationName = symbolProvider.toSymbol(operation).name.toPascalCase()
+    private val operationName = symbolProvider.toSymbol(operation).name
     private val operationId = operation.id
 
     /** Returns `std::convert::Infallible` if the model provides no errors. */


david-perez:
Let's see what the "damage" to the SDK diff is.

Changelog entries missing.

david-perez:
It looks like the following 34 operations in the AWS Rust SDK are affected:

```rust
AddClientIDToOpenIDConnectProvider
CreateOpenIDConnectProvider
CreateSAMLProvider
CreateVirtualMFADevice
DeactivateMFADevice
DeleteOpenIDConnectProvider
DeleteSAMLProvider
DeleteSSHPublicKey
DeleteVirtualMFADevice
EnableMFADevice
EnableVolumeIO
GetOpenIDConnectProvider
GetSAMLProvider
GetSSHPublicKey
ListMFADeviceTags
ListMFADevices
ListOpenIDConnectProviderTags
ListOpenIDConnectProviders
ListSAMLProviderTags
ListSAMLProviders
ListSSHPublicKeys
ListVirtualMFADevices
RemoveClientIDFromOpenIDConnectProvider
ResyncMFADevice
TagMFADevice
TagOpenIDConnectProvider
TagSAMLProvider
UntagMFADevice
UntagOpenIDConnectProvider
UntagSAMLProvider
UpdateOpenIDConnectProviderThumbprint
UpdateSAMLProvider
UpdateSSHPublicKey
UploadSSHPublicKey
```

Even if the breaking change is large, we have to do something. We're being inconsistent right now. For example, we're currently generating:

* [`GetSAMLProvider`](https://docs.rs/aws-sdk-iam/0.19.0/aws_sdk_iam/operation/struct.GetSAMLProvider.html) in `operation.rs`,
* [`GetSamlProviderInput`](https://docs.rs/aws-sdk-iam/0.19.0/aws_sdk_iam/input/struct.GetSamlProviderInput.html) in `input.rs`,
* [`GetSamlProviderOutput`](https://docs.rs/aws-sdk-iam/0.19.0/aws_sdk_iam/output/struct.GetSamlProviderOutput.html) in `output.rs`; and
* [`GetSAMLProviderError`](https://docs.rs/aws-sdk-iam/0.19.0/aws_sdk_iam/error/struct.GetSAMLProviderError.html) in `error.rs`

in `aws-sdk-iam-0.19.0`.

There's an argument to be made that we should leave operation names for input/output/error structs as they are in the input model, because it's likely that we would deviate from the official API docs (e.g. the official API documentation for this particular operation is `GetSAMLProvider`: https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetSAMLProvider.html) 

@awslabs/rust-sdk-owners thoughts?

jdisanti:
Definitely agree we need consistency here. However, after examining the other SDKs and the SDK documentation, I think we should go in the opposite direction: preserve the operation name rather than convert it to Pascal case. The reason being is the overwhelming majority of the other SDKs take this approach, and maintaining the case from the model will make it consistent with the API reference docs.

What we could do is change the input/output shape creation to use `operationName + suffix` instead of calling `toPascalCase`.

Does this work for the server side?

jdisanti:
The breaking change will also need to get called out in the SDK changelog.

david-perez:
Agree, we should preserve the case.

jdisanti:
I think we should use the modeled name for every name where both Smithy models and Rust convention call for `PascalCase`/`PascalACRONYMCase`. So given that, the only one we would need to call `toPascalCase()` on would be the enum variants, since those are all modeled as `UPPER_SNAKE_CASE`.

We may also want to add a sanity check that we're outputting names in the correct Rust convention. For example, if someone calls their operation `foo_bar` in the Smithy model, we should correct that to `FooBar` in codegen. But if they provide the name `FOOBar`, then we should just keep that as is.

david-perez:
The [Smithy spec's ABNF](https://awslabs.github.io/smithy/2.0/spec/model.html#grammar-token-smithy-Identifier) says that shape names are `Identifier`s:

```
Identifier =
    IdentifierStart *IdentifierChars

IdentifierStart =
    *"_" ALPHA

IdentifierChars =
    ALPHA / DIGIT / "_"
```

`@enum` names in IDL v1 [follow the same rules](https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html#enum-trait), i.e. the `^[a-zA-Z_]+[a-zA-Z_0-9]*$` regular expression.

[Rust identifiers](https://doc.rust-lang.org/reference/identifiers.html)' grammar is a strict superset.

However, Rust API guidelines say [types and enum variants should be `UpperCamelCase`](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) (aka PascalCase).

To convert shape names to type and enum variant names, we can't simply:

1. uppercase the first character,
2. uppercase every character after an underscore; and
3. remove underscores,

as you suggest, as that could introduce conflicts (e.g. distinct `__my_shape` and `_my__shape` shapes would both convert to `MyShape`). Note that we currently have this bug with `toPascalCase()` too. We need either a smarter renaming strategy, or retain model names and `#[allow(non_camel_case_types)]`.

jdisanti:
I think Smithy is opinionated by default on naming conventions, right? You would need to suppress warnings to encounter these edge cases if I recall correctly.

rcoh:
the original failure to rename operations was an oversight (not an intentional choice). 

Personally I don't think we should necessarily preserve what's modeled—IIRC it's not totally consistent across all the models and can lead to weird results.

My preference would be:
1. Move operations to PascalCase. This has the nice property of adhering to Rust's [naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html):
> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

2. Add a detector for the `_` conflict case. We can either fix it with the reserved words or hard bail.

Personally, I'm not bothered at all by casing differences with the official docs but happy to be convinced

david-perez:
If we're going to rename, the devil is in the details: what renaming strategy do we choose? `toPascalCase` gives rise to _additional_ conflicts, e.g. two distinct Smithy operations `GetSAMLProvider` and `GetSamlPROVider` would both be renamed as `GetSamlProvider`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt
```diff
@@ -284,16 +284,10 @@ open class SymbolVisitor(
...
 
-    override fun operationShape(shape: OperationShape): Symbol {
-        return symbolBuilder(
-            RustType.Opaque(
-                shape.contextName(serviceShape)
-                    .replaceFirstChar { it.uppercase() },
-            ),
-        )
+    override fun operationShape(shape: OperationShape) =
+        symbolBuilder(RustType.Opaque(shape.contextName(serviceShape).toPascalCase()))
```
hlbarber, line 288: Why does this need a `toPascalCase`?

hlbarber (approved):

david-perez:
We now lean towards using Smithy identifiers as-is in generated Rust code. However, this is blocked by https://github.com/awslabs/smithy-rs/issues/2340.

jmklix:
we've decided to begin closing old PRs that we don't plan to work on in the near future. Please let us know if you still think this PR is needed

PR #1828, opened 2022-10-07
Factor out instrumentation utilities to common crate

jjant:
## Checklist


jdisanti:
I want to discuss more where these should live, since adding a new crate seems like a big decision. There are a number of parts in CI and the SDK generation process that need to get updated with new runtime crates, and we also have some informal standards in the runtime crates that we should standardize, such as [these additional lints](https://github.com/awslabs/smithy-rs/issues/763).

It seems like we should have a general RFC outlining how to add new runtime crates, and then a separate RFC to justify each new addition.

Velfi:

jdisanti:
I think at this point, this PR will take a considerable amount of work to merge with the latest. I'm going to opt to close it, but feel free to re-open if you disagree.

PR #1830, opened 2022-10-07
Add marker trait for numeric shapes that should be serialized even when equal to 0

Velfi:
## Motivation and Context

aws-sdk-rust#630

## Description

update: serializer generation to serialize zero values for shapes with the `SerializeZeroValues` trait
fix: bug that occurs when serializing a `"com.amazonaws.s3#ScanRange"` starting at zero
add: `SerializeZeroValues` smithy trait
update: unify names of `serializerUtil` vars in serializer codegen
update: CHANGELOG.next.toml with release notes

## Testing

I verified that this change fixes the bug and included an integration test.

## Checklist

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
@@ -7,6 +7,11 @@ package software.amazon.smithy.rustsdk.customize.s3
 
 import software.amazon.smithy.aws.traits.protocols.RestXmlTrait
 import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.DoubleShape
+import software.amazon.smithy.model.shapes.FloatShape
+import software.amazon.smithy.model.shapes.IntegerShape
+import software.amazon.smithy.model.shapes.LongShape
+import software.amazon.smithy.model.shapes.NumberShape
 import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.model.shapes.ServiceShape
 import software.amazon.smithy.model.shapes.Shape
@@ -32,6 +37,8 @@ import software.amazon.smithy.rust.codegen.core.smithy.generators.LibRsSection
 import software.amazon.smithy.rust.codegen.core.smithy.protocols.ProtocolMap
 import software.amazon.smithy.rust.codegen.core.smithy.protocols.RestXml
 import software.amazon.smithy.rust.codegen.core.smithy.traits.AllowInvalidXmlRoot
+import software.amazon.smithy.rust.codegen.core.smithy.traits.SerializeZeroValues
+import software.amazon.smithy.rust.codegen.core.util.UNREACHABLE
 import software.amazon.smithy.rust.codegen.core.util.letIf
 import software.amazon.smithy.rustsdk.AwsRuntimeType
 import java.util.logging.Logger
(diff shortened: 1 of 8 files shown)


david-perez:
Why are we not serializing zero values? I was able to trace the introduction of this behavior to https://github.com/awslabs/smithy-rs/issues/439. We're enforcing it via tests in `restXml`:

https://github.com/awslabs/smithy-rs/blob/e78da5598a2d94d23fd18e30492a20ce3603f844/codegen-client-test/model/rest-xml-extras.smithy#L38-L63

And `restJson1`:

https://github.com/awslabs/smithy-rs/blob/e78da5598a2d94d23fd18e30492a20ce3603f844/codegen-core/common-test-models/rest-json-extras.smithy#L117-L126

These tests are part of our "extras" suites; I can't find mention of this behavior in the official AWS specs.

To me this seems _really_ odd. The server is also reusing the serializers, so it's also not serializing zero values in responses.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
```diff
@@ -48,6 +55,12 @@ class S3Decorator : RustCodegenDecorator<ClientProtocolGenerator, ClientCodegenC
         ShapeId.from("com.amazonaws.s3#GetObjectAttributesOutput"),
     )
 
+    private val serializeZeroValuesAllowList = setOf(
+        // Omitting this when zero would change the meaning of the range.
+        // https://docs.aws.amazon.com/AmazonS3/latest/API/API_ScanRange.html
+        ShapeId.from("com.amazonaws.s3#Start"),
```
Licht-T, line 61: I believe `End` should also accept zero.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_ScanRange.html#AmazonS3-Type-ScanRange-End
```bash
[~] aws s3api select-object-content \
--bucket ${BUCKET_NAME_OMMITED} \
--key data-1m.json \
--expression 'SELECT * FROM s3object s' \
--expression-type SQL \
--input-serialization '{"JSON": {"Type": "LINES"}, "CompressionType": "NONE"}' \
--output-serialization '{"JSON": {}}' \
--scan-range Start=0,End=0 \
0.json
[~] echo $?
0
[~] cat 0.json
{"a":"foo","b":1,"c":"aaa"}
```

Licht-T (changes requested):

Licht-T:
I briefly looked into the `smithy-rs` code base and found that it sometimes uses zero as an optional value, which means not-inputted. Also, I took a look at other SDKs, and they are handling a not-inputted as `null`. 
- [Rust SDK](https://github.com/awslabs/aws-sdk-rust/blob/b06cd25e2ade91951d4ce52ced908c49ac280401/sdk/s3/src/xml_ser.rs#L1114-L1121)
- [Java SDK](https://github.com/aws/aws-sdk-java/blob/fd409dee8ae23fb8953e0bb4dbde65536a7e0514/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/transform/RequestXmlFactory.java#L176-L177)

I think this is the structural problem, and, in the long term, `smithy-rs` should handle a not-inputted as `Option`.

jdisanti:
I think the larger fix that needs to happen is input tracking. Any value provided by a customer must be serialized and sent, regardless if it's zero value or not. It's hard to say why omitting the zero value was implemented since the commit doesn't provide any reasoning. I also didn't see any information about it in the Smithy specs.

@david-perez - What's the expected behavior for server side serialization? Is it compatible with input tracking (output tracking in the server case)?

david-perez:
No problems server-side on implementing input/output tracking.

Omitting serialization of values (even if they're the default for a type) is even called out as something we should **not** do in the server [in the spec](https://awslabs.github.io/smithy/2.0/spec/type-refinement-traits.html#default-value-serialization):

> Authoritative model consumers like servers SHOULD always serialize default values to remove any ambiguity about the value of the most up to default value.

Velfi:
I spoke with @rcoh about this and he thinks it should be solved by boxing primitive fields for S3 models. I can look into this after I complete my current task

Velfi:
Closing in favor of an alternate solution

PR #1837, opened 2022-10-11
Add PluginExt extension trait and FilterByOperationName plugin

hlbarber:
## Motivation and Context

Customers may not want to apply a `Plugin` to every operation in their service. They might want to filter application by operation name.

## Description

- Refactor different `Plugin`s to live in their own folders.
- Add `PluginExt` extension trait, which has methods `stack` and `filter_by_operation_name`.
- Add `FilterByOperationName` `Plugin`.

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/operation/mod.rs
+++ b/rust-runtime/aws-smithy-http-server/src/operation/mod.rs
@@ -193,8 +193,11 @@ pub use upgrade::*;
 ///
 /// The `L` is held and applied lazily during [`Upgradable::upgrade`].
 pub struct Operation<S, L = Identity> {
-    inner: S,
-    layer: L,
+    /// The inner [`Service`](tower::Service) representing the logic of the operation.
+    pub inner: S,
+    /// The [`Layer`](tower::Layer) applied to the HTTP [`Service`](tower::Service) after `S` has been wrapped in
+    /// [`Upgrade`].
+    pub layer: L,
 }
 
 impl<S, L> Operation<S, L> {
--- a/rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
+++ b/rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
@@ -1,0 +1,42 @@
+use tower::util::Either;
+
+use crate::operation::{Operation, OperationShape};
+
+use super::Plugin;
+
+pub struct FilterByOperationName<Inner, F> {
+    inner: Inner,
+    predicate: F,
+}
+
+impl<Inner, F> FilterByOperationName<Inner, F> {
+    /// Creates a new [`FilterByOperationName`].
+    pub fn new(inner: Inner, predicate: F) -> Self {
+        Self { inner, predicate }
+    }
+}
+
+impl<P, Op, S, L, Inner, F> Plugin<P, Op, S, L> for FilterByOperationName<Inner, F>
+where
+    F: Fn(&'static str) -> bool,
+    Inner: Plugin<P, Op, S, L>,
+    Op: OperationShape,
+{
+    type Service = Either<Inner::Service, S>;
+    type Layer = Either<Inner::Layer, L>;
+
+    fn map(&self, input: Operation<S, L>) -> Operation<Self::Service, Self::Layer> {
+        if (self.predicate)(Op::NAME) {
+            let Operation { inner, layer } = self.inner.map(input);
+            Operation {
+                inner: Either::A(inner),
+                layer: Either::A(layer),
+            }
+        } else {
+            Operation {
+                inner: Either::B(input.inner),
+                layer: Either::B(input.layer),
+            }
+        }
+    }
+}
--- a/rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
+++ b/rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
@@ -9,6 +9,9 @@ use crate::operation::{Operation, OperationShape};
 
 use super::Plugin;
 
+/// A [`Plugin`] used to filter [`Plugin::map`] application using a predicate over the [`OperationShape::NAME`].
+///
+/// See [`PluginExt::filter_by_operation_name`](super::PluginExt::filter_by_operation_name) for more information.
 pub struct FilterByOperationName<Inner, F> {
     inner: Inner,
     predicate: F,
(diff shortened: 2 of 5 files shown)


rust-runtime/aws-smithy-http-server/src/operation/mod.rs
```diff
@@ -193,8 +193,11 @@ pub use upgrade::*;
 ///
 /// The `L` is held and applied lazily during [`Upgradable::upgrade`].
 pub struct Operation<S, L = Identity> {
-    inner: S,
-    layer: L,
+    /// The inner [`Service`](tower::Service) representing the logic of the operation.
```
hlbarber, line 196: These have become public in order to allow `Plugin` developers access to them.

Is this too severe? Maybe we should document that the average customer should use the `from_handler`/`from_service` methods.

rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
```diff
@@ -0,0 +1,42 @@
...
+use super::Plugin;
+
+pub struct FilterByOperationName<Inner, F> {
+    inner: Inner,
+    predicate: F,
+}
+
+impl<Inner, F> FilterByOperationName<Inner, F> {
+    /// Creates a new [`FilterByOperationName`].
+    pub fn new(inner: Inner, predicate: F) -> Self {
```
hlbarber, line 14: `pub(crate)` and force them to use the extension trait methods?

rust-runtime/aws-smithy-http-server/src/plugin/mod.rs
```diff
@@ -46,41 +54,39 @@ pub trait Plugin<Protocol, Op, S, L> {
...
+    /// ```rust
+    /// # use aws_smithy_http_server::{plugin::{Plugin, PluginExt}, operation::{Operation, OperationShape}};
+    /// # struct Pl;
+    /// # struct Op;
+    /// # impl OperationShape for Op { const NAME: &'static str = ""; type Input = (); type Output = (); type Error = (); }
+    /// # impl Plugin<(), Op, (), ()> for Pl { type Service = (); type Layer = (); fn map(&self, input: Operation<(), ()>) -> Operation<(), ()> { input }}
+    /// # let plugin = Pl;
+    /// # let operation = Operation { inner: (), layer: () };
+    /// // Prevents `plugin` from being applied to the `CheckHealth` operation.
+    /// let filtered_plugin = plugin.filter_by_operation_name(|name| name != "com.aws.example#CheckHealth");
```
LukeMathWalker, line 81: I imagine we'd like customers to prefer using the `NAME` associated constant on the relevant `OperationShape` to perform this kind of operations rather than hard-coding a static string.
Although I realise that makes writing the example somewhat more cumbersome.

rust-runtime/aws-smithy-http-server/src/plugin/identity.rs
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use crate::operation::Operation;
+
+use super::Plugin;
+
+/// An [`Plugin`] that maps an `input` [`Operation`] to itself.
+pub struct IdentityPlugin;
```
LukeMathWalker, line 11: Not related to this PR, but throwing it out as a discussion point: would `NoOpPlugin` be a better name for this?
I intuitively get what `identity` stands for here (math training, yay) but it feels like `no-op` would be clearer to a larger audience.

rust-runtime/aws-smithy-http-server/src/operation/mod.rs
LukeMathWalker: If we want customers to be able to write their own plugins, along the same lines of what we are doing here with the filtering, I don't see many other options.

rust-runtime/aws-smithy-http-server/src/plugin/identity.rs
jjant: I disagree, `NoOp` could be understood to mean that the plugin turns the operation into a no-op.

hlbarber: I'm tempted to hide behind some existing precedents and ignore the confusion of the masses.

https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/struct.Identity.html
https://docs.rs/tower/0.4.8/tower/layer/util/struct.Identity.html
https://doc.rust-lang.org/std/convert/fn.identity.html

rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
```diff
@@ -0,0 +1,50 @@
...
+        Self { inner, predicate }
+    }
+}
+
+impl<P, Op, S, L, Inner, F> Plugin<P, Op, S, L> for FilterByOperationName<Inner, F>
+where
+    F: Fn(&'static str) -> bool,
```
jjant, line 29: Instead of matching on a string, wouldn't it be better (i.e, more type-safe) to generate an `enum` of available operations and use that to filter?
i.e,
```rust
// Current 
let filtered_plugin = plugin.filter_by_operation_name(|name| name != "com.aws.example#CheckHealth");

// vs

let filtered_plugin = plugin.filter_by_operation(|op| op != Operation::CheckHealth);
```

Granted, the "magic strings" issue is partially solved if you use things like `CheckHealth::NAME`.

rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
LukeMathWalker: I think that's good call 👌🏻

rust-runtime/aws-smithy-http-server/src/plugin/identity.rs
LukeMathWalker: I think `std` is a good enough precedent here 👌🏻

rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
hlbarber: Good idea
https://github.com/awslabs/smithy-rs/issues/1839

jjant (approved):

LukeMathWalker (approved):

PR #1840, opened 2022-10-11
Add boxed method to service builder

hlbarber:
## Motivation and Context

As a result of the [Service Parameterized Routers](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0020_service_builder.md#service-parameterized-routers) change, the `{ServiceName::layer}` modifies the type of the service. This has the benefit of avoiding the heap but means that types can build up with several applications of `layer`.

For users who wish to erase the types via `Route::new` (`Box::new` in disguise), we provide a `boxed` method. This can be seen as an opt-in reversal to the [Service Parameterized Routers](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0020_service_builder.md#service-parameterized-routers) change.

## Description

- Add `boxed` method to the service builder.
- Add reasonable default `S` to the service builder.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -321,6 +321,21 @@ class ServerServiceGeneratorV2(
                         router: self.router.map(|s| s.layer(layer))
                     }
                 }
+
+                /// Applies [`Route::new`](#{SmithyHttpServer}::routing::Route::new) to all routes.
+                ///
+                /// This has the effect of erasing all types accumulated via [`layer`].
+                pub fn boxed<B>(self) -> $serviceName<#{SmithyHttpServer}::routing::Route<B>>
```
jjant, line 328: Couldn't this method go in an extension trait in a runtime crate?
Bringing this up because I feel that less generated code is usually better.

crisidev: I actually believe the opposite.. generated code is easier to manage and release than runtime code..

hlbarber: If this was a standalone function I think it might fit better in the `rust-runtime` crates but given that this is a small method on a generated struct I think it makes sense to leave it here for now. If we did this via extension trait the customer would have to import the trait to use the method which may be confusing.

It is possible though: we'd need to add a base trait with the `layer` method on it, impl that onto `PokemonService` via codegen, and then include this `boxed` method extension to that trait.

Maybe we should draft out a heuristic to aid in this kind of decision making.

crisidev: Yeah, we should figure out guidelines to decide whether to put stuff in runtime vs codegen!

LukeMathWalker: In this case I'd lean towards putting this method in the code generated `struct`. It's a better developer experience than having to import a single-use extension trait just to call it.

LukeMathWalker (approved):

PR #1841, opened 2022-10-12
Simplify implementation of rendering prefix headers

david-perez:
The implementation that binds the target of
`@httpPrefixHeaders`-annotated member shapes is overly complex in that
it accounts for the possibility of it targeting a list shape. However,
`@httpPrefixHeaders` can only be applied to structure members targeting
map shapes:

* https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpprefixheaders-trait
* https://awslabs.github.io/smithy/2.0/spec/http-bindings.html#httpprefixheaders-trait

The existing implementation likely came to be by copy-pasting the
`RustWriter.renderHeaders` extension function.

I've also taken the liberty of reformatting and renaming variables for
clarity and consistency.


jdisanti (approved):
Nice!

PR #1842, opened 2022-10-12
Add RFC for improving SDK credential caching through type safety

jdisanti:
## Motivation and Context
This PR adds an RFC that outlines some possible solutions to https://github.com/awslabs/aws-sdk-rust/issues/629 and recommends we move forward with one of them.

[Rendered](https://github.com/awslabs/smithy-rs/blob/jdisanti-rfc-cred-cache-type-safety/design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md)


Velfi (approved):
I like A

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
```diff
@@ -0,0 +1,260 @@
...
+--------------
+
+Option C should be taken since it offers the greatest type safety while still maintaining
+enough flexibility.
+
+Changes Checklist
+-----------------
+
+Option C:
+- [ ] Implement `CredentialsCache` with its `Lazy` variant and builder
```
rcoh, line 250: I assume AssumeRoleProvider won't be cached anymore?

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
```diff
@@ -0,0 +1,260 @@
...
+
+The `credentials_provider` method on `ConfigLoader` would only take `CredentialsCache` as an argument
+so that the SDK could not be configured without credentials caching, or if opting out of caching becomes
+a use case, then a `CredentialsCache::NoCache` variant could be made.
+
+Like Option A, a convenience method can be added to make using the default cache easier:
+
+```rust
+let sdk_config = aws_config::from_env()
+    .credentials_with_default_cache(ImdsCredentialsProvider::builder().build())
```
rcoh, line 197: probably wouldn't make the default name longer than custom name—maybe `cache_credentials`? or just `credentials` and document it?

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
```diff
@@ -0,0 +1,260 @@
...
+```rust
+pub struct LazyCache {
+    credentials_provider: Arc<dyn ProvideCredentials>,
+    // ...
+}
+
+pub struct EagerCache {
+    credentials_provider: Arc<dyn ProvideCredentials>,
+    // ...
+}
```
rcoh, line 127: I like this design, but I'm struck by the fact that we might be composing unnecessarily–what if we separate the cache from the inner credentials provider in this part of the design?

You would then be able to construct a "cache strategy" separate from the credentials provider. When the default chain was actually loaded, you could assemble the concrete impl. 
This also happens to address the fact that under the hood, an eager the cache and a lazy cache are the same code, we just pre-call provide credentials on the eager cache

Here's what I'm thinking:

```rust
/// Signal-struct to denote the cache mode (no logic)
pub struct CredentialsCache {
  inner: CredentialsCacheInner
}
enum CredentialsCacheInner {
  Lazy,
  Eager,
  NoCaching
}
impl CredentialsCache {
  pub fn lazy() -> Self { .. }
  pub fn eager() -> Self { .. }
  pub(crate) fn instantiate(provider: Box<dyn ProvideCredentials>, conf: ProviderConfig) -> ConcreteCredentialsCache { .. // private function used to actually make the chain }
}

async fn main() {
  // custom provider, default cache
  aws_config::load_from_env().credentials(ImdsProvider::new()).load(); // uses default cache
  // default provide , but no caching
  aws_config::load_from_env()
    .credentials_cache(CredentialsCache::no_cache())
    .load(); 

```

rcoh (approved):

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
jdisanti: I really like how this removes the need to think about caching when changing the credentials provider while still resolving the core issue. It's only slightly different from Option C, so I've opted to set it as the selected design that the doc focuses on, and move the rest of the alternatives to an appendix.

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
```diff
@@ -0,0 +1,328 @@
...
+    NoCaching,
+}
+pub struct CredentialsCache {
+    inner: Inner,
+}
+
+impl CredentialsCache {
+    // Methods prefixed with `default_` just use the default cache settings
+    pub fn default_lazy() -> Self { /* ... */ }
+    pub fn default_eager() -> Self { /* ... */ }
```
Velfi, line 54: Should we specify that these shouldn't change implementation based on features? Currently, I think several things depend on their being an HTTP client and I know we have some features that enable that client.

We already have the features RFC, so maybe mentioning this would just be redundant.

Velfi (approved):

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
```diff
@@ -0,0 +1,328 @@
...
+
+    // Unprefixed methods return a builder that can take customizations
+    pub fn lazy() -> LazyBuilder { /* ... */ }
+    pub fn eager() -> EagerBuilder { /* ... */ }
+
+    // Later, when custom implementations are supported
+    pub fn custom(cache_impl: Box<dyn SomeCacheTrait>) -> Self { /* ... */ }
```
ysaito1001, line 61: I wonder if there's a slight inconsistency in mappings between the method name and what it returns. `lazy` and `eager` return their respective builders but `custom` returns `Self`, not a builder. How do you feel us having
```
pub fn lazy() -> Self { /* ... */ }
pub fn eager() -> Self { /* ... */ }

pub fn lazy_builder() -> LazyBuilder { /* ... */ }
pub fn eager_builder() -> EagerBuilder { /* ... */ }

pub fn custom(cache_impl: Box<dyn SomeCacheTrait>) -> Self { /* ... */ }
```
That way, `lazy`, `eager`, and `custom`, which correspond to the variant names of `Inner`, now all return `Self`. Also, it's clear to the client that `lazy_builder` and `eager_builder` both return builders.

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
```diff
@@ -0,0 +1,328 @@
...
+
+- [ ] Implement `CredentialsCache` with its `Lazy` variant and builder
+- [ ] Add `credentials_cache` method to `ConfigLoader`
+- [ ] Refactor `ConfigLoader` to take `CredentialsCache` instead of `impl ProvideCredentials + 'static`
+- [ ] Refactor `SharedCredentialsProvider` to take a cache implementation in addition to an `impl ProvideCredentials + 'static`
+- [ ] Rename `LazyCachingCredentialsProvider` -> `LazyCredentialsCache`
+- [ ] Refactor the SDK `Config` code generator to be consistent with `ConfigLoader`
+- [ ] Write changelog upgrade instructions
+- [ ] Fix examples (if there are any for configuring caching)
+
```
ysaito1001, line 119: Related to [this](https://github.com/awslabs/smithy-rs/pull/1842#discussion_r1013228538) comment, we may also need
- [ ] Remove a field [cache](https://github.com/awslabs/smithy-rs/blob/fba88df3eeb6c5ea8be2ea2df57232366c016b37/aws/rust-runtime/aws-config/src/sts/assume_role.rs#L47) from `AssumeRoleProvider`

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
```diff
@@ -0,0 +1,328 @@
...
+    pub fn default_eager() -> Self { /* ... */ }
+
+    // Unprefixed methods return a builder that can take customizations
+    pub fn lazy() -> LazyBuilder { /* ... */ }
+    pub fn eager() -> EagerBuilder { /* ... */ }
+
+    // Later, when custom implementations are supported
+    pub fn custom(cache_impl: Box<dyn SomeCacheTrait>) -> Self { /* ... */ }
+
+    pub(crate) fn create_cache(
```
ysaito1001, line 63: Given this method returns `SharedCredentialsProvider`, maybe we should name this differently? I don't have a name that's spot on, but here is a list of names I'm considering:
- `back_credentials_provider`
- `cache_aware_credentials_provider`
- `enable_cache_for`

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
```diff
@@ -0,0 +1,328 @@
...
+The `credentials_cache` will default to `CredentialsCache::default_lazy()` if not provided.
+
+Changes Checklist
+-----------------
+
+- [ ] Implement `CredentialsCache` with its `Lazy` variant and builder
+- [ ] Add `credentials_cache` method to `ConfigLoader`
+- [ ] Refactor `ConfigLoader` to take `CredentialsCache` instead of `impl ProvideCredentials + 'static`
+- [ ] Refactor `SharedCredentialsProvider` to take a cache implementation in addition to an `impl ProvideCredentials + 'static`
+- [ ] Rename `LazyCachingCredentialsProvider` -> `LazyCredentialsCache`
```
ysaito1001, line 115: After renaming, does `LazyCredentialsCache` still impl `ProvideCredentials`? If so, can the user can write something like this
```
let sdk_config = aws_config::from_env()
    .credentials_cache(CredentialsCache::default_eager())
    .credentials_provider(LazyCredentialsCache::builder().load(/* some concrete impl for ProvideCredentials */).build())
    .load()
    .await;
```
assuming both methods, `credentials_provider` and `credentials_cache`, exist in `ConfigLoader`?

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
jdisanti: This method is `pub(crate)`, so the name from the RFC is just a suggestion and doesn't need to match the implementation.

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
jdisanti: The cache will no longer be a provider.

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
ysaito1001: Thanks. I missed this comment from Option C, which I assume is still applicable to the selected design:
```
// ^ Note: SharedCredentialsProvider would get renamed to SharedCredentialsCache.
// This code is using the old name to make it clearer that it already exists,
// and the rename is called out in the change checklist.
```
So the method is returning a type of a cache for which the name `create_cache` makes sense.

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
jdisanti: I think we're probably OK here given the other RFC.

design/src/rfcs/rfc0023_sdk_credential_cache_type_safety.md
jdisanti: I like it.

ysaito1001 (approved):

PR #1846, opened 2022-10-13
Python: Migrate to new service builder

unexge:
`OperationRegistryBuilder` will be deprecated soon but we're still using it in generated Python SDKs. This PR migrates codegen and runtime crates to the new service builder.


hlbarber (approved):
This looks good to me.

If nothing here breaks the Python customer facing API then I'm guessing this is fine to merge as is?

unexge:
> This looks good to me.
> 
> If nothing here breaks the Python customer facing API then I'm guessing this is fine to merge as is?

I don't expect customers to manually implement `PyApp` trait so it should be fine as long as they have matching versions of codegen and runtime crate. @crisidev what do you think about it?

crisidev:
> I don't expect customers to manually implement `PyApp` trait so it should be fine as long as they have matching versions of codegen and runtime crate. @crisidev what do you think about it?

Yep, I agree.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
```diff
@@ -193,25 +203,23 @@ class PythonApplicationGenerator(
                         """
                         let ${name}_locals = #{pyo3_asyncio}::TaskLocals::new(event_loop);
                         let handler = self.handlers.get("$name").expect("Python handler for operation `$name` not found").clone();
-                        let router = router.$name(move |input, state| {
-                            #{pyo3_asyncio}::tokio::scope(${name}_locals, crate::operation_handler::$name(input, state, handler))
+                        let builder = builder.$name(move |input, state| {
+                            #{pyo3_asyncio}::tokio::scope(${name}_locals.clone(), crate::operation_handler::$name(input, state, handler.clone()))
```
crisidev, line 207: Both `locals` and `handler` are already cloned / ownable. Am I missing anything here?

unexge: New service builder expects closures to be `Fn`, and without cloning those variables for every execution we basically consume them and that turns this closure to `FnOne` as far as I understand. I got this error if I don't clone those variables:
```
error[E0525]: expected a closure that implements the `Fn` trait, but this closure only implements `FnOnce`
   --> pokemon-service-server-sdk/src/python_server_application.rs:122:44
    |
122 |         let builder = builder.check_health(move |input, state| {
    |                               ------------ ^^^^^^^^^^^^^^^^^^^ this closure implements `FnOnce`, not `Fn`
    |                               |
    |                               the requirement to implement `Fn` derives from here
123 |             pyo3_asyncio::tokio::scope(
124 |                 check_health_locals,
    |                 ------------------- closure is `FnOnce` because it moves the variable `check_health_locals` out of its environment
```

crisidev: Ok. Regarding `locals`, I guess you can't just put `let ${name}_locals = #{pyo3_asyncio}::TaskLocals::new(event_loop);` inside the closure because you will need to move `event_loop`..

LGTM. I just wanted to make sure we are not cloning unnecessary stuff. This happens only once per process during startup, so we are good.

crisidev (approved):

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
unexge: I've tried to move `locals` inside closure but it fails to compile because `event_loop` is not `Sync`:
```
error[E0277]: `UnsafeCell<PyObject>` cannot be shared between threads safely
   --> pokemon-service-server-sdk/src/python_server_application.rs:184:30
    |
184 |             .service(builder.build());
    |                              ^^^^^ `UnsafeCell<PyObject>` cannot be shared between threads safely
    |
    = help: within `PyAny`, the trait `Sync` is not implemented for `UnsafeCell<PyObject>`
    = note: required because it appears within the type `PyAny`
    = note: required because of the requirements on the impl of `std::marker::Send` for `&PyAny`
    = note: required because it appears within the type `[closure@pokemon-service-server-sdk/src/python_server_application.rs:121:44: 127:10]`
    = note: required because it appears within the type `aws_smithy_http_server::operation::IntoService<CheckHealth, [closure@pokemon-service-server-sdk/src/python_server_application.rs:121:44: 127:10]>`
    = note: required because it appears within the type `Upgrade<AwsRestJson1, CheckHealth, (Extension<Py<PyAny>>,), aws_smithy_http_server::body::Body, aws_smithy_http_server::operation::IntoService<CheckHealth, [closure@pokemon-service-server-sdk/src/python_server_application.rs:121:44: 127:10]>>`
```

PR #1847, opened 2022-10-13
Support PowerPC architecture for sigv4 signature

crisidev:
## Motivation and Context
`aws-sigv4` crate does not compile on PowerPC because of `ring`. This PR adds support PowerPC by using `hmac` and `sha2` instead of `ring`.

I have also improved the CI to run as many tests as we can in the runtime crates. Some tests as been disabled under 32bit architectures, see #1857 .

## Testing
Unit tests are passing using `cross`. CI will run all the tests as well.

- [x] I have updated CHANGELOG.next.toml if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates


CHANGELOG.next.toml
```diff
@@ -154,4 +154,10 @@ author = "hlbarber"
...
 references = ["smithy-rs#1817"]
 meta = { "breaking" = false, "tada" = false, "bug" = true }
-author = "ethyi"
\ No newline at end of file
+author = "ethyi"
+
+[[aws-sdk-rust]]
+message = "Support Sigv4 signature generation on PowerPC."
+references = ["smithy-rs#1847"]
+meta = { "breaking" = false, "tada" = false, "bug" = true }
```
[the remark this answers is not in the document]
jdisanti: It is technically a breaking change, but has a very small blast radius, and this crate is not in a good position to be even remotely stable. The right thing to do is to call it out as a separate `[[aws-sdk-rust]]` breaking entry, but it's possible it won't actually break anyone.

This line should definitely get removed though: https://github.com/awslabs/smithy-rs/blob/cbbbec4a4892e301ebde3af5cc852aa48d82e608/aws/rust-runtime/aws-sigv4/external-types.toml#L9-L10

.github/workflows/ci.yml
```diff
@@ -161,7 +161,7 @@ jobs:
         # We also exclude all first-party crates that have a non-optional dependency on `ring`.
```
LukeMathWalker, line 161: This comment should be removed at this point.

.github/workflows/ci.yml
```diff
@@ -205,6 +205,12 @@ jobs:
         use-cross: true
         command: build
         args: --target ${{ matrix.target }} --manifest-path "aws/rust-runtime/Cargo.toml" ${{ matrix.aws_excludes }} --workspace
+    - name: Test AWS rust-runtime crates
```
LukeMathWalker, line 208: Should we run tests for `rust-runtime` crates as well at this point?

aws/rust-runtime/aws-sigv4/Cargo.toml
```diff
@@ -23,10 +23,17 @@ http = { version = "0.2", optional = true }
...
 time = "0.3.5"
 tracing = "0.1"
 
+[target.'cfg(not(any(target_arch = "powerpc", target_arch = "powerpc64")))'.dependencies]
+ring = "0.16"
+
+# ring does not compile on powerpc (https://github.com/briansmith/ring/issues/389) so we use hmac and sha2.
+[target.'cfg(any(target_arch = "powerpc", target_arch = "powerpc64"))'.dependencies]
+hmac = "0.12"
+sha2 = "0.10"
```
LukeMathWalker, line 35: Considering that the implementation using `hmac` and `sha2` can run on all platforms, do you think it makes sense for us to keep `ring` around and juggle the burden of conditional compilation here?

.github/workflows/ci.yml
crisidev: Yeah, I am working on it!

aws/rust-runtime/aws-sigv4/Cargo.toml
crisidev: AFAIK ring is much more performant on x86 than hmac + sha2. I'll make some benches to ensure we select the right dependency.

jjant: If we keep Ring for non-powerpc, you might want to add the reason in the comment here.

crisidev: I have added some benchmarks and ring is 3x slower than hmac+sha2. This PR is still in the works. I am talking with Jon to understand if we want to migrate all the architectures away from ring.

CHANGELOG.next.toml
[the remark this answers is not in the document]
crisidev: Done.

LukeMathWalker (approved):

.github/workflows/ci.yml
```diff
@@ -181,30 +194,61 @@ jobs:
...
+        passthrough = ["OPENSSL_DIR"]
         EOF
-    - name: Build rust-runtime crates
+    - name: Build Smithy-rs rust-runtime crates
       uses: actions-rs/cargo@v1
       with:
         use-cross: true
         command: build
-        args: --target ${{ matrix.target }} --manifest-path "rust-runtime/Cargo.toml" --exclude aws-smithy-http-server-python --workspace ${{ matrix.non_aws_features }}
+        args: -vv --target ${{ matrix.target }} --manifest-path "rust-runtime/Cargo.toml" ${{ matrix.build_smithy_rs_exclude }} --workspace ${{ matrix.build_smithy_rs_features }}
```
Velfi, line 233: Is the `-vv` still needed?

.github/workflows/ci.yml
```diff
@@ -181,30 +194,61 @@ jobs:
...
     - name: Build AWS rust-runtime crates
       uses: actions-rs/cargo@v1
       with:
         use-cross: true
         command: build
-        args: --target ${{ matrix.target }} --manifest-path "aws/rust-runtime/Cargo.toml" ${{ matrix.aws_excludes }} --workspace
+        args: -vv --target ${{ matrix.target }} --manifest-path "aws/rust-runtime/Cargo.toml" ${{ matrix.build_aws_exclude }} --workspace
```
Velfi, line 239: Same here about the `vv`

Velfi (approved):

.github/workflows/ci.yml
LukeMathWalker: Without `-vv` you get pretty much no logs if a pre-build script fails. 
I'd lean towards leaving it in - I only check CI logs when it fails and there I do appreciate having _something_ to look at.

PR #1856, opened 2022-10-14
Check that release workflow is run on main

jjant:
## Motivation and Context
We've recently had problems with a release by selecting an older commit/branch. This PR adds a check to prevent that particular mistake from happening.
We've also had difficulty determining if a particular release workflow run was in "dry run" mode. This PR increases visibility in that respect.

## Description
- Updates the name of the release workflow to `Release smithy-rs - Dry run` or `Release smithy-rs - Production run` depending on `inputs.dry_run`
- Release workflow now checks that it was run on `main` and fails if not.

## Testing
I verified that the main branch check works on this branch: https://github.com/awslabs/smithy-rs/actions/runs/3250387459/jobs/5334004431 (it fails as expected).
From that run you can see that the workflow title was updated successfully as well.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,7 @@ on:
 jobs:
   main-branch-check:
     name: Check that workflow is running in main
+    runs-on: ubuntu-latest
     steps:
     - name: Main branch check
       if: ${{ github.ref_name != 'main' }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,7 @@ jobs:
   main-branch-check:
     name: Check that workflow is running in main
     if: ${{ github.ref_name != "main" }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -13,6 +13,7 @@ env:
   rust_version: 1.61.0
 
 name: Release smithy-rs
+run_name: ${{ github.workflow }} - ${{ inputs.dry_run && "Dry run" || "Production run" }}
```
jjant, line 16: See [`run-name`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name).

`inputs.dry_run && "Dry run" || "Production run"` is a poor man's `inputs.dry_run ? "Dry run" : "Production run"`   (the latter is not supported in GH actions).

hlbarber (approved):

PR #1859, opened 2022-10-17
[RFC - Server] Refining the service builder API

LukeMathWalker:
## Motivation and Context
This PR adds an RFC that explores how we could evolve the new service builder API to improve developer experience, in particular for Rust beginners.

[Rendered RFC](https://github.com/awslabs/smithy-rs/blob/refine-builder/design/src/rfcs/rfc0022_refine_builder.md)

The API proposed in this RFC has been manually implemented for the Pokemon service. You can find the code [here](https://github.com/LukeMathWalker/builder-experiments).


design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+We can now try again to stray away from the happy path by forgetting to register a handler for the `CheckHealth` operation.  
+The code compiles just fine this time, but the application fails when launched via `cargo run`:
+
+```text
+<timestamp> ERROR pokemon_service: You must specify a handler for all operations attached to the `Pokemon` service.
+We are missing handlers for the following operations:
+- com.aws.example#CheckHealth
+
+Use the dedicated methods on `PokemonServiceBuilder` to register the missing handlers:
+- PokemonServiceBuilder::check_health
```
jjant, line 192: Beginners to rust may not know that this is telling them to use `builder.check_health()`, maybe we should clarify that.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
+RFC: Evolving the new service builder API
+========================
+
+> Status: Accepted
```
jjant, line 4: Is this accepted already? 😄

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+```
+
+All operation fields are now `Option`-wrapped.  
+We introduce a new `MissingOperationsError` error to hold the names of the missing operations and their respective setter methods:
+
+```rust
+#[derive(Debug)]
+pub struct MissingOperationsError {
+    service_name: &'static str,
+    operation_names2setter_methods: HashMap<&'static str, &'static str>,
```
jjant, line 175: We can probably just use a `TinyMap` or even `SmallVec/TinyVec` here, right?

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+    do_nothing: Option<Op2>,
+    get_pokemon_species: Option<Op3>,
+    get_server_statistics: Option<Op4>,
+    capture_pokemon: Option<Op5>,
+    get_storage: Option<Op6>,
+    #[allow(unused_parens)]
+    _exts: std::marker::PhantomData<(Exts1, Exts2, Exts3, Exts4, Exts5, Exts6)>,
+    plugin: Pl,
+}
+```
```
jjant, line 166: Maybe a diff view would make the distinction clearer here (tho you lose syntax highlighting AFAIK).

```diff
pub struct PokemonServiceBuilder<
    Op1,
    Op2,
    Op3,
    Op4,
    Op5,
    Op6,
    Exts1 = (),
    Exts2 = (),
    Exts3 = (),
    Exts4 = (),
    Exts5 = (),
    Exts6 = (),
    Pl = aws_smithy_http_server::plugin::IdentityPlugin,
  > {
-   check_health: Op1,
+   check_health: Option<Op1>,
    ...
    #[allow(unused_parens)]
    _exts: std::marker::PhantomData<(Exts1, Exts2, Exts3, Exts4, Exts5, Exts6)>,
    plugin: Pl,
}
```

design/src/rfcs/rfc0022_refine_builder.md
LukeMathWalker: It will if it gets merged, otherwise that doesn't matter 😛

design/src/rfcs/rfc0022_refine_builder.md
LukeMathWalker: Possibly! I kept the code simple in the RFC, but we can of course refine the internal implementation details when actually moving forward with the implementation.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+We no longer store the raw handlers inside `PokemonServiceBuilder`.
+We eagerly upgrade the operation handlers to a `Route` instance when they are registered with the builder. The existing API performs the upgrade when `build` is called, forcing `PokemonServiceBuilder` to store the raw handlers and keep two generic parameters around (`OpX` and `ExtsX`) for each operation.  
+
+There is one downside to this alternative approach - all plugins must be specified upfront, when creating an instance of the builder.
+Plugins cannot be modified through `PokemonServiceBuilder`'s API if we want to guarantee that all operation handlers are upgraded to a `Route` using the same plugins.
+
+```rust
+impl PokemonService<()> {
+    /// Constructs a builder for [`PokemonService`].
+    pub fn builder<Body, Plugin>(plugin: Plugin) -> PokemonServiceBuilder<Body, Plugin> {
```
hlbarber, line 484: We could also do a two-phase builder which keeps the user in the method chain, something like:

```rust
let app = PokemonService::builder() // : PokemonServicePluginBuilder<IdentityPlugin>
    .print()
    /* ... */
    .finalize_plugins() // PokemonServiceBuilder<B, PluginStack<IdentityPlugin, ..., PrintPlugin>>
    .get_pokemon_species(/* handler */)
    ...
    .build();
```

hlbarber, line 484: I think it's worth noting that now `Plugin` might be a complicated type given that it must be fully specified up front. Do we need to type erase it?

hlbarber (commented):
Looking good so far.

I think we should include a section on what happens to `FailOnMissingOperation` etc. There are various use cases where we want to opt out of the normal `Operation<S, L>` upgrade upgrade procedure.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+) -> PokemonServiceBuilder<Op1, Op2, Op3, Op4, Op5, Op6> {
+    builder
+}
+```
+
+That compiles, at last.  
+Let's try to register an operation handler now:
+
+```rust
+fn partial_setup<Op1, Op2, Op3, Op4, Op5, Op6>(
```
david-perez, line 428: Far from ideal, but for simple use cases, you could get away with a macro that performs this initialization.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+}
+```
+
+This pattern needs to be revisited if we want to move forward with this RFC, since new plugins cannot be registered after the builder has been instantiated.  
+My recommendation would be to implement `Pluggable` for `PluginStack`, providing the same pattern ahead of the creation of the builder:
+
+```rust
+// Currently you'd have to go for `PluginStack::new(IdentityPlugin, IdentityPlugin)`,
+// but that can be smoothed out even if this RFC isn't approved.
+let plugin_stack = PluginStack::default()
```
hlbarber, line 533: I think it'd also be a good idea to make a `PluginBuilder` here - `PluginStack` is kinda confusing term to be a primary API.

This would mirror how `tower` has a [`ServiceBuilder`](https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html) _and_ [`Stack`](https://docs.rs/tower/latest/tower/layer/util/struct.Stack.html).

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+        where
+            Self: Sized,
+    {
+        self.apply(PrintPlugin)
+    }
+}
+```
+
+This pattern needs to be revisited if we want to move forward with this RFC, since new plugins cannot be registered after the builder has been instantiated.  
+My recommendation would be to implement `Pluggable` for `PluginStack`, providing the same pattern ahead of the creation of the builder:
```
hlbarber, line 528: I think this is a great bit of luck that this migrates so cleanly. There's no harm in doing this kind of impl even without this RFC being accepted.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+```text
+error[E0277]: the trait bound `MissingOperation: Upgradable<AwsRestJson1, CheckHealth, (), _, IdentityPlugin>` is not satisfied
+  --> pokemon-service/src/bin/pokemon-service.rs:38:10
+   |
+38 |         .build();
+   |          ^^^^^ the trait `Upgradable<AwsRestJson1, CheckHealth, (), _, IdentityPlugin>` is not implemented for `MissingOperation`
+   |
+   = help: the following other types implement trait `Upgradable<Protocol, Operation, Exts, B, Plugin>`:
+             FailOnMissingOperation
+             Operation<S, L>
```
david-perez, line 75: I actually think this is one of the nicer error messages you can get. It has enough of the right words `MissingOperation`, `CheckHealth` for it to likely to not cause much confusion.

The error messages you get when passing in a handler that does not satisfy the type signature (not `async`, wrong input / output types) are worse, although those (and perhaps this one too?) can be palliated with a debug macro à la [`axum`'s `debug_handler`](https://docs.rs/axum-macros/latest/axum_macros/attr.debug_handler.html).

I expect this error to not be frequent. Users will likely start off from a template; the new service builder needs to generate Rust docs with an out-of-the-box compilable program, like in https://github.com/awslabs/smithy-rs/pull/1498#issuecomment-1280720925.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+```
+
+In practice, this is unlikely to have any impact on developers using `smithy-rs`.  
+We have seen how cumbersome it is to break the startup logic into different functions using the current service builder API. Developers are most likely specifying all plugins and routes in the same function even if the current API allows them to intersperse route registrations and plugin registrations: they would simply have to re-order their registration statements to adopt the API proposed in this RFC.
+
+### Alternatives: boxed trait objects 
+
+Trait objects are another common approach to achieve type erasure: instead of storing the raw handler in `PokemonServiceBuilder`, we box it and cast it to a `dyn Upgradable<...>`.  
+Using trait objects would allow us to delay upgrading to `Route` until `build` is called, mimicking the behaviour of the current API and allowing developers to register plugins at any point before calling `build`.  
+Using trait objects would unfortunately have a non-zero runtime impact: we would be introducing an extra layer of indirection (double-boxing the handler), affecting the performance of our application at runtime.
```
david-perez, line 505: I think this is worth exploring. Following a vtable has negligible performance impact for our use cases. If you can flatten the type that the router ends up holding to just a few boxes, it's likely ok.

david-perez (commented):
I don't think the missing handler error message is persuasive; there are sharper edges elsewhere. The other two identified shortcomings (branching and `partial_setup`) are very compelling and I agree we should do something to remediate them.

design/src/rfcs/rfc0022_refine_builder.md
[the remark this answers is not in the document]
LukeMathWalker: I should have called this out in the RFC, but I am indeed not convinced this is the right way forward for the same reason you mentioned - it can be confusing and lead to surprising runtime behaviour.
Especially considering how existing web frameworks in Rust ignore _at what point_ a middleware is registered and apply it to all routes at the same scope/level.

design/src/rfcs/rfc0022_refine_builder.md
LukeMathWalker: > The error messages you get when passing in a handler that does not satisfy the type signature (not async, wrong input / output types) are worse, although those (and perhaps this one too?) can be palliated with a debug macro à la [axum's debug_handler](https://docs.rs/axum-macros/latest/axum_macros/attr.debug_handler.html), which I think we direly need anyway.

I don't disagree that those errors are worse, but I don't think that the builder case can be fixed using a debug macro. Or, at least, not in the general case nor easily.
The macro would need to wrap the entire scope where the builder is being used, therefore ruling out refactorings that break the startup logic across multiple functions. We would also have to understand branches in the code we are wrapping to make sure that the all paths lead to a correctly instantiated builder - it gets nasty quite quickly.

design/src/rfcs/rfc0022_refine_builder.md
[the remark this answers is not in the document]
david-perez: If the intention is that the plugin would only affect the operations registered so far, I think it's ok.

We can also consider providing a way to apply plugins (i.e. layers to the operation before it's HTTP-upgraded) more fine-grained, to _specific_ operations via a method on `Operation`, instead of the current way in #1837. But this is a one-way door, we can add this functionality later if there's user interest.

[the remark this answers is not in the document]
hlbarber: > We can also consider providing a way to apply plugins (i.e. layers to the operation before it's HTTP-upgraded) more fine-grained, to specific operations via a method on Operation, instead of the current way in https://github.com/awslabs/smithy-rs/pull/1837.

I think just adding a blanket impl

```rust
impl<P, Op, Pl, S, L> Pluggable<Pl> for Operation<S, L> where Pl: Plugin<P, Op, S, L> {
    type Output = Operation<Pl::Service, Pl::Layer>;

   fn apply(self, new_plugin: Pl) -> Self::Output {
       new_plugin.map(self)
   }
}
```

to the `aws-smithy-http-server`

would allow you to

```rust
op.print()
```

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+> Applies to: Server
+
+[RFC 20] introduced a new service builder API. 
+It supports fine-grained configuration at multiple levels (per-handler middlewares, router middlewares, plugins) while trying to prevent some misconfiguration issues at compile-time (i.e. missing operation handlers).  
+There is consensus that the new API is an improvement over the pre-existing `OperationRegistryBuilder`/`OperationRegistry`, which is now on its way to deprecation in one of the next releases.  
+
+This RFC builds on top of [RFC 20] to explore an alternative API design prior to its stabilisation.  
+The API proposed in this RFC has been manually implemented for the Pokemon service. You can find the code [here](https://github.com/LukeMathWalker/builder-experiments).
+
+## Overview
```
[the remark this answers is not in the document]
LukeMathWalker: I've added a brief explanation in https://github.com/awslabs/smithy-rs/pull/1859/commits/c9be4e8f9d5de690b44f2fb2243aca4d5d57f253 - wdyt?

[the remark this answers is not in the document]
hlbarber: LGTM

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+    get_pokemon_species: Option<Route<Body>>,
+    get_server_statistics: Option<Route<Body>>,
+    capture_pokemon: Option<Route<Body>>,
+    get_storage: Option<Route<Body>>,
+    plugin: Plugin,
+}
+```
+
+We no longer store the raw handlers inside `PokemonServiceBuilder`.
+We eagerly upgrade the operation handlers to a `Route` instance when they are registered with the builder. The existing API performs the upgrade when `build` is called, forcing `PokemonServiceBuilder` to store the raw handlers and keep two generic parameters around (`OpX` and `ExtsX`) for each operation.
```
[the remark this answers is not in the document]
LukeMathWalker: Added one in https://github.com/awslabs/smithy-rs/pull/1859/commits/d8f012685a01599b4303f725e52b4186920e0a6c!

design/src/rfcs/rfc0022_refine_builder.md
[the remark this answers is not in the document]
LukeMathWalker: I have captured both as an alternative in https://github.com/awslabs/smithy-rs/pull/1859/commits/90ae7aa8a02b3e39b767c8e9379e244d47b7dd98
I've added as well a sketch of an API for attaching plugins to groups of routes using a nested/scoped builder.

design/src/rfcs/rfc0022_refine_builder.md
LukeMathWalker: Yeah, this could work as well, with a `PokemonService::builder_without_plugins()` for the case where the developer doesn't want to register any plugin and `PokemonService::builder().finalize_plugins()` would be ugly.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,556 @@
...
+    get_storage: Option<Route<Body>>,
+    plugin: Plugin,
+}
+```
+
+We no longer store the raw handlers inside `PokemonServiceBuilder`.
+We eagerly upgrade the operation handlers to a `Route` instance when they are registered with the builder. The existing API performs the upgrade when `build` is called, forcing `PokemonServiceBuilder` to store the raw handlers and keep two generic parameters around (`OpX` and `ExtsX`) for each operation.  
+
+There is one downside to this alternative approach - all plugins must be specified upfront, when creating an instance of the builder.
+Plugins cannot be modified through `PokemonServiceBuilder`'s API if we want to guarantee that all operation handlers are upgraded to a `Route` using the same plugins.
```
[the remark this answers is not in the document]
LukeMathWalker: Resolving this as the conversation on this topic is continuing in https://github.com/awslabs/smithy-rs/pull/1859#discussion_r996890519

design/src/rfcs/rfc0022_refine_builder.md
LukeMathWalker: Captured this in https://github.com/awslabs/smithy-rs/pull/1859/commits/a9aab96c5cba854b830632637204e953c8eaf852 👍🏻

design/src/rfcs/rfc0022_refine_builder.md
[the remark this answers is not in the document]
LukeMathWalker: > I don't think this is the case here - we could Box<Box<Box<dyn Upgradable<...>>>> but the actual upgrade procedure would still result in a singly nested handler (Route::new)

Is this the case though?

`Route::new` desugars to a `BoxCloneService::new` call, which introduces one layer of vtable indirection.
The argument to `BoxCloneService::new` is the parameter passed to `Route::new` - if that were to be behind multiple layers of boxing, I am under the impression that we would have to resolve those indirections at runtime when invoking the service.

It's a different situation when we are talking about boxed `Plugin`s - the call to `.map` would suffer from the indirection, but that only happens at startup time.

Have I got this right or am I missing something (more than likely, tbh)?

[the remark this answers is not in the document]
hlbarber: Currently, `Upgradable` looks like this:

```rust
/// Provides an interface to convert a representation of an operation to a HTTP [`Service`](tower::Service) with
/// canonical associated types.
pub trait Upgradable<Protocol, Operation, Exts, B, Plugin> {
    type Service: Service<http::Request<B>, Response = http::Response<BoxBody>>;

    /// Performs an upgrade from a representation of an operation to a HTTP [`Service`](tower::Service).
    fn upgrade(self, plugin: &Plugin) -> Self::Service;
}
```

If we wanted to erase the type we'd have to remove the associated type which would vary across each operation (else we'd have to specify it in the `dyn Upgradable<Service = ....>`). So we'd have to move the `Route::new` into the `Ugpradable`, like this:

```rust
pub trait Upgradable<Protocol, Operation, Exts, B, Plugin> {
    fn upgrade(self, plugin: &Plugin) -> Route<B>;
}
```

I think is acceptable tweak - all the gnarly `Send + Clone + ...` bounds would be pushed down into the `Upgradable for Operation` impl too, rather than hanging around on the service builder which might be worth while anyway (I recall you mentioning this before).

The `.boxed()` method on the service builder would then take the `OpX` and turn it into a `Box<dyn Upgradable<...>>`. The `Box` here would `Box` the `dyn Upgradable` and if we call `.boxed()` multiple times it'd become `Box<Box<Box<dyn Upgradable>>>` but this is `Box`ing the actual `Upgrade` trait. The `Route::new` is still only called once and hence the indirection at runtime only happens once as far as I can tell?

LukeMathWalker: I've significantly expanded this section exploring two different approaches for lazy type erasure. 
I don't believe it's actually practical moving forward with either of them - they do improve slightly on the current status quo but they carry a significant complexity budget in terms of implementation and ongoing maintenance.
We spent some time we @hlbarber to play around with ideas and they get quite complex quite quickly.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,791 @@
...
+    }
+}
+```
+
+This constraint guarantees that all operation handlers are upgraded to a `Route` using the same set of plugins.
+
+Having to specify all plugins upfront is unlikely to have a negative impact on developers currently using `smithy-rs`.
+We have seen how cumbersome it is to break the startup logic into different functions using the current service builder API. Developers are most likely specifying all plugins and routes in the same function even if the current API allows them to intersperse route registrations and plugin registrations: they would simply have to re-order their registration statements to adopt the API proposed in this RFC.
+
+### Alternatives: allow new plugins to be registered after builder creation
```
hlbarber, line 525: I think we should add a final alternative here:

- Don't solve the function signature problem at all, the customer just has to `PokemonService::builder()....build()` within one function.
- Solve the conditional branch problem with `Either<A, B>` types.

hlbarber (approved):
The problems described here are very real. The solution space is very subtle and the RFC is a great exploration of it. Here's a summary of my thoughts:

Clear wins in this exploration are:

- `PluginBuilder` API.
- [`Upgradable` associated type removal](https://github.com/awslabs/smithy-rs/pull/1891).

We should implement these regardless.

The main proposal provides the customer with a super ergonomic solution to both the conditional branching and error message problem. This is given to the customer by default - no intervention is required on their part.

The function signature problem seems more difficult to solve generally. The main proposal prevents the customer from having to work with the `O(n)` type parameters but the `Plugin` itself is now required to be known up front. The customer now only has to deal with 1 nasty type parameter when passing a builder through a function signature, but is it still too much of an obstruction? With that said, alternative approach 1 and 2 provide strictly less ergonomic and the `Either` proposal doesn't even attempt to solve it.

Fundamentally, if we want `Plugin<P, Op, S, L>` to be applied to operations we cannot do it after type erasure of `P`, `Op`, `S`, `L`. Any solution which preserves the current `Plugin` architecture and type erases any of these parameters will prevent `Plugin`s being applied afterward. You can see this with the main proposal and the alternative approach 2.

The only feasible solution which solves the conditional branching problem, provides an infallible builder, and prevents eager `Plugin` is to provide `Either<A, B>: Upgradable` in my opinion. This does not solve the function signature problem or the error problem and we still stuck with `O(n)` type parameters.

In conclusion:

- Both lazy type erasure via `Box<dyn Upgradable>` and `Route::new` can be used to hold onto the compile time safety of `build` but are far worse ergonomics for the customer and still require `Plugin`s to be known up front - this probably not a good trade-off. 
- If we submit to the inevitability of the function signature problem then I think the `Either` proposal is the main contestant to the proposal.

design/src/rfcs/rfc0022_refine_builder.md
```diff
@@ -0,0 +1,791 @@
...
+            plugin,
+        }
+    }
+}
+```
+
+This constraint guarantees that all operation handlers are upgraded to a `Route` using the same set of plugins.
+
+Having to specify all plugins upfront is unlikely to have a negative impact on developers currently using `smithy-rs`.
+We have seen how cumbersome it is to break the startup logic into different functions using the current service builder API. Developers are most likely specifying all plugins and routes in the same function even if the current API allows them to intersperse route registrations and plugin registrations: they would simply have to re-order their registration statements to adopt the API proposed in this RFC.
```
david-perez, line 523: > Developers are most likely specifying all plugins and routes in the same function

This seems to go counter to the [Refactoring into smaller functions -> Prepare for some type juggling!](https://github.com/awslabs/smithy-rs/blob/refine-builder/design/src/rfcs/rfc0022_refine_builder.md#refactoring-into-smaller-functions---prepare-for-some-type-juggling) section that motivates the proposal for reducing type parameters.

david-perez (commented):
What's the plan for solving the same set of problems when building a plugin stack (conditionally applying a plugin and partial initialization of a plugin stack in a function)?

design/src/rfcs/rfc0022_refine_builder.md
LukeMathWalker: I have not worded this correctly. 
What I meant: since it's so complicated to refactor into smaller functions (given the problems shown in the sections you linked) we can be fairly confident that all the code in the wild is currently sticking to having everything in a single function to avoid the pain (even if they don't necessarily _want_ to).

design/src/rfcs/rfc0022_refine_builder.md
david-perez: > Don't solve the function signature problem at all, the customer just has to `PokemonService::builder()....build()` within one function.

I think the RFC should mention using macros instead of functions to workaround this limitation.

LukeMathWalker: Macros that we provide or macros written on an ad-hoc basis by the customer?

LukeMathWalker: I'd argue that writing a macro is a skill outside the comfort zone of most Rust users, who primarily consume them. Sketching a very simple `macro_rules!` is somewhat more accessible, but not much.

LukeMathWalker:
Based on our standup discussion from a few days ago:

- This RFC does not attempt to solve the same set of problems (branching, refactoring into functions) for building a `Plugin` stack. We should explore the ergonomics of `Plugin` stack building in a separate RFC and/or provide dedicated documentation for common patterns that we expect developers to use;
- We will move forward with the implementation of this RFC before releasing the new builder API to minimise disruption (e.g. breaking changes) for users.

I'll merge this in and get cracking.

PR #1861, opened 2022-10-17
Add rust-toolchain override file

jjant:
## Description
Adds the [`rust-toolchain.toml`](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) file that compels rustup to override the local toolchain to use Rust 1.62.0 (our current MSRV).

Developers might want to run `rustup override unset` in their local smithy-rs repo to have the `rust-toolchain.toml` file be the source of truth for their local Rust version (instead of any overrides they might have set manually before).


hlbarber:
Does it make sense to try and centralize the specification of the MSRV? CI could read from this file etc?

jdisanti:
> Does it make sense to try and centralize the specification of the MSRV? CI could read from this file etc?

You might be able to get half way there, but it's not going to work for GitHub Actions. I don't think it's too difficult to grep for the current MSRV to set it correctly everywhere.

PR #1862, opened 2022-10-17
Use pin-utils crate instead of vendoring the `pin_mut` macro

jjant:
## Description
Remove our vendored copy of `pin_utils::pin_mut` in favor of using the crate (it's tiny, so it shouldn't affect compile times).


jdisanti (approved):
I think this should allow us to resolve https://github.com/awslabs/smithy-rs/issues/1111

crisidev (approved):

jdisanti:
It may be possible to simplify this by using pin projection since the `Inner` struct already has `pin_project!` applied to it.

Issue #1854, opened 2022-10-13
`changelogger` created an empty changelog entry for `release-2022-10-13`


jdisanti:
The SDK release on 2022-10-13 went out with an empty changelog for some reason: https://github.com/awslabs/aws-sdk-rust/commit/db33636adc8c171f7669918782e637fabcfbbeaf

- [x] Diagnose and fix root cause of the missing entries
- [ ] ~Add check to fail release on empty changelog~
- [x] Fix author case sensitivity issue (i.e., `LukeMathWalker` is getting converted to `lukemathwalker` before comparison with `smithy-rs-maintainers.txt`)

drganjoo:
So one difference I can think of is that the "Release Smithy-Rs" workflow (a non-dry run) had to be cancelled midway. Maybe this could have caused it?

jdisanti:
The root cause was determined to be that the release workflow was run on a different branch than main, which won't work since it needs to add a changelog update commit. This was addressed by https://github.com/awslabs/smithy-rs/pull/1856

> So one difference I can think of is that the "Release Smithy-Rs" workflow (a non-dry run) had to be cancelled midway. Maybe this could have caused it?

Canceling a release should be fine so long as it hasn't reached the "Push smithy-rs changes" step yet. Otherwise, it is not safe to cancel.

jdisanti:
On second thought, I'm hesitant to add a check to fail on empty changelog since that is a valid use case for aws-sdk-rust right now, if we were to update the models without updating the code generator, since there's currently no mechanism to add model updates to the changelog.

PR #1866, opened 2022-10-18
Fix case sensitivity bug in `changelogger`

jdisanti:
## Motivation and Context
This fixes a case sensitivity issue with the author check in `changelogger` discovered in the last release (#1854).

### DIFF
--- a/tools/changelogger/src/render.rs
+++ b/tools/changelogger/src/render.rs
@@ -6,6 +6,7 @@
 use crate::entry::{ChangeSet, ChangelogEntries, ChangelogEntry};
 use anyhow::{bail, Context, Result};
 use clap::Parser;
+use once_cell::sync::Lazy;
 use ordinal::Ordinal;
 use serde::Serialize;
 use smithy_rs_tool_common::changelog::{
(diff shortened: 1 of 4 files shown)


tools/changelogger/src/render.rs
```diff
@@ -181,8 +188,8 @@ fn render_entry(entry: &HandAuthoredEntry, mut out: &mut String) {
         .map(|t| t.to_string())
         .chain(entry.references.iter().map(to_md_link))
         .collect::<Vec<_>>();
-    if !maintainers().contains(&entry.author.to_ascii_lowercase().as_str()) {
-        references.push(format!("@{}", entry.author.to_ascii_lowercase()));
+    if !is_maintainer(&entry.author) {
+        references.push(format!("@{}", entry.author));
     };
```
jjant, line 193: Why were we converting author names to lowercase here? 
Wouldn't it make the code simpler (and maybe even "more correct") if we make this comparison case-sensitive?

Velfi: I think the casing doesn't matter because my GitHub name is capitalized but it looks like these are all equivalent:
- @veLFi
- @Velfi
- @velfi

raw markdown:
```
- @veLFi
- @Velfi
- @velfi
```

Velfi (approved):

jjant (approved):

PR #1868, opened 2022-10-18
Add ModelLayer parameter to Operation struct

hlbarber:
## Motivation and Context

There is an asymmetry with applying a `Layer` before and after the operation upgrade procedure. This can be seen in the difference between [C](https://github.com/awslabs/smithy-rs/blob/harryb/middleware-docs/design/src/server/middleware.md#c-operation-specific-http-middleware) and [D](https://github.com/awslabs/smithy-rs/blob/harryb/middleware-docs/design/src/server/middleware.md#d-operation-specific-model-middleware) in the [middleware documentation](https://github.com/awslabs/smithy-rs/blob/harryb/middleware-docs/design/src/server/middleware.md).

This can be rectified by adding a new type parameter to `Operation`.

## Description

- Add `ModelLayer` type parameter to `Operation` struct.
- Add `Operation::model_layer` method to `Operation`.
- Rename `Operation::layer` method to `Operation::http_layer`.
- Change `Plugin` trait to map `(ModelLayer, HttpLayer)`.


Velfi:

PR #1871, opened 2022-10-18
Python: map Python middlewares to Tower layers

unexge:
Map pure Python middlewares to Tower layers so they can have same flexibility as pure Rust middlewares.

Pure Python middlewares will look like this:
```python
@app.middleware
async def middleware(
    request: Request, 
    next: Callable[[Request], Awaitable[Response]]
) -> Response:
    # Do something with the `request`
    request.headers["X-Request-Id"] = ...

    # Call the handler or the next middleware in the stack
    response = await next(request)

    # Do something with the `response`
    response.headers["X-Server"] = ...

    return response
```
so they will have total control over the execution flow. They can modify the request, return a response without calling the handler, apply timeout to the handler, modify the response returned by the handler etc.

## TODOs
- [ ] ~~Provide a way to access to the `Context`~~
- [x] Modifiers for the `Request`
- [x] Modifiers for the `Response`

/cc @chadac @teruokun @PC-LoadLetter


codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
```diff
@@ -212,13 +209,22 @@ class PythonApplicationGenerator(
...
-                        .boxed_clone()
-                        .layer(
-                            #{SmithyPython}::PyMiddlewareLayer::<#{Protocol}>::new(self.middlewares.clone(), middleware_locals),
-                        )
-                        .service(builder.build());
+                    let mut service = #{tower}::util::BoxCloneService::new(builder.build());
+
+                    {
+                        use #{tower}::Layer;
+                        tracing::debug!("adding middlewares to rust python router");
```
LukeMathWalker, line 216: I would put this (and the log statement below) at `trace` level rather than `debug`.
It's almost never going to be useful, so worth keeping it at the chattiest level.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
```diff
@@ -248,11 +254,17 @@ class PythonApplicationGenerator(
...
                 }
-                /// Register a request middleware function that will be run inside a Tower layer, without cloning the body.
+                /// Register a Python function to be executed inside a Tower middleware layer.
                 ##[pyo3(text_signature = "(${'$'}self, func)")]
-                pub fn request_middleware(&mut self, py: #{pyo3}::Python, func: #{pyo3}::PyObject) -> #{pyo3}::PyResult<()> {
-                    use #{SmithyPython}::PyApp;
-                    self.register_middleware(py, func, #{SmithyPython}::PyMiddlewareType::Request)
+                pub fn middleware(&mut self, py: #{pyo3}::Python, func: #{pyo3}::PyObject) -> #{pyo3}::PyResult<()> {
+                    let handler = #{SmithyPython}::PyMiddlewareHandler::new(py, func)?;
+                    tracing::info!(
```
LukeMathWalker, line 261: Same here - lower it to `trace`.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
```diff
@@ -248,11 +254,17 @@ class PythonApplicationGenerator(
...
+                        "registering middleware function `{}`, coroutine: {}",
+                        handler.name,
+                        handler.is_coroutine,
```
LukeMathWalker, line 264: I'd generally recommend capturing `handler.name` and `handler.is_coroutine` as fields on the log statement rather than interpolating them as a string.
It makes it easier when it comes to searching through the logs.
See [here](https://docs.rs/tracing/latest/tracing/#recording-fields) for guidance.

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
```diff
@@ -229,7 +244,9 @@ async def stream_pokemon_radio(_: StreamPokemonRadioInput, context: Context):
 ###########################################################
 # Run the server.
 ###########################################################
-def main():
+def main() -> None:
     app.run(workers=1)
```
LukeMathWalker, line 248: Do we need a separate `main` function? 🤔

rust-runtime/aws-smithy-http-server-python/src/error.rs
```diff
@@ -63,7 +63,12 @@ impl PyMiddlewareException {
 
 impl From<PyErr> for PyMiddlewareException {
     fn from(other: PyErr) -> Self {
-        Self::newpy(other.to_string(), None)
+        // Try to extract `PyMiddlewareException` from `PyErr` and use that if succeed
+        let middleware_err = Python::with_gil(|py| other.to_object(py).extract::<Self>(py));
+        match middleware_err {
+            Ok(err) => err,
+            Err(_) => Self::newpy(other.to_string(), None),
```
LukeMathWalker, line 70: Here we probably want a log statement capturing the details of the exception that we failed to extract since it didn't match the one we expected.

rust-runtime/aws-smithy-http-server-python/src/middleware/error.rs
```diff
@@ -0,0 +1,27 @@
...
+pub enum PyMiddlewareError {
+    /// Returned when `next` is called multiple times.
+    #[error("next already called")]
+    NextAlreadyCalled,
+    /// Returned when request is accessed after `next` is called.
+    #[error("request is gone")]
+    RequestGone,
+    /// Returned when response is called after it is returned.
+    #[error("response is gone")]
+    ResponseGone,
```
LukeMathWalker, line 20: The comments are great - it feels like they should actually be used as error messages (by rephrasing them accordingly).

rust-runtime/aws-smithy-http-server-python/src/middleware/handler.rs
```diff
@@ -3,21 +3,64 @@
...
+}
+
+#[pyo3::pymethods]
+impl PyNext {
+    // Calls the inner Tower service with the `Request` that is passed from Python.
+    // It returns a coroutine to be awaited on the Python to complete the call.
+    // Note that it takes wrapped objects from both `PyRequest` and `PyNext`,
+    // so after calling `next`, consumer can't access to the `Request` or
+    // can't call the `next` again, this basically emulates consuming `self` and `Request`,
+    // but since we are crossing Python boundary we can't express it in natural Rust terms.
```
LukeMathWalker, line 45:
```suggestion
    // Calls the inner Tower service with the `Request` that is passed from Python.
    // It returns a coroutine to be awaited on the Python side to complete the call.
    // Note that it takes wrapped objects from both `PyRequest` and `PyNext`,
    // so after calling `next`, consumer can't access to the `Request` or
    // can't call the `next` again, this basically emulates consuming `self` and `Request`,
    // but since we are crossing the Python boundary we can't express it in natural Rust terms.
```

rust-runtime/aws-smithy-http-server-python/src/middleware/layer.rs
```diff
@@ -50,175 +50,79 @@ where
...
+        let handler_name = handler.name.clone();
+        let next = BoxService::new(inner.map_err(|err| err.into()));
+        let locals = self.locals.clone();
+        let into_response = self.into_response;
+
+        Box::pin(
+            handler
+                .call(req, next, locals)
+                .or_else(move |err| async move {
+                    tracing::error!(err = %err, "middleware '{handler_name}' failed");
```
LukeMathWalker, line 122: By only logging the display representation of the error (`%err`) I don't think you are capturing the traceback information.

rust-runtime/aws-smithy-http-server-python/src/middleware/pytests/layer.rs
```diff
@@ -0,0 +1,409 @@
+/*
```
LukeMathWalker, line 1: It feels like there is a lot of duplication in these tests, which results in long tests and a somewhat obscure purpose (i.e. what is each of them testing? It gets a bit lost).
If you extract the common bits into helper functions (e.g. the `mock::spawn_with` bits), the tests should become a bit terser and easier to digest.

rust-runtime/aws-smithy-http-server-python/src/error.rs
unexge: The error on the `Err(_)` variant will always be `'TypeError' object cannot be converted to 'MiddlewareException'` here because it is a conversion error and the real error will be logged at https://github.com/awslabs/smithy-rs/blob/e12136563d734e3c1ea40d96a1705d3b1f5d9cbe/rust-runtime/aws-smithy-http-server-python/src/middleware/layer.rs#L124

rust-runtime/aws-smithy-http-server-python/src/middleware/layer.rs
unexge: Yes good point, I've added `RichPyErr` to add more detail to `Debug` implementation of `PyErr` https://github.com/awslabs/smithy-rs/commit/6db19b31a3d8da5b3e0b58337fb42ebd0def89a3

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
unexge: I think it is not required but having a explicitly defined [entry point](https://packaging.python.org/en/latest/specifications/entry-points/) is kinda useful when you want to setup your service via a supervisor or something

rust-runtime/aws-smithy-http-server-python/examples/libpokemon_service_server_sdk.pyi
```diff
@@ -0,0 +1,17 @@
+# NOTE: This is manually created to surpass some mypy errors and it is incomplete,
```
crisidev, line 1: I think you need the SPDX header eveywhere.

rust-runtime/aws-smithy-http-server-python/examples/mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy]
```
crisidev, line 1: SPDX header.

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
```diff
@@ -124,49 +133,51 @@ def get_random_radio_stream(self) -> str:
...
+# to get `Response` from the handler or by constructing `Response` by itself.
+# It can also modify the `Request` before calling `next` or it can also modify
+# the `Response` returned by the handler.
+# It can also raise an `MiddlewareException` with custom error message and HTTP status code,
+# any other raised exceptions will cause an internal server error response to be returned.
+
+# Next is either the next middleware in the stack or the handler.
+Next = Callable[[Request], Awaitable[Response]]
+
+# This middleware checks the `Conten-Type` from the request header,
```
crisidev, line 148:
```suggestion
# This middleware checks the `Content-Type` from the request header,
```

rust-runtime/aws-smithy-http-server-python/src/middleware/header_map.rs
```diff
@@ -0,0 +1,159 @@
...
+    pyclass, PyErr, PyResult,
+};
+
+use crate::{mutable_mapping_pymethods, util::collection::PyMutableMapping};
+
+/// Python-compatible [HeaderMap] object.
+#[pyclass(mapping)]
+#[derive(Clone, Debug)]
+pub struct PyHeaderMap {
+    inner: Arc<Mutex<HeaderMap>>,
```
crisidev, line 21: Could a `RwLock` be cheaper for reads here?

crisidev (approved):
LGTM

rust-runtime/aws-smithy-http-server-python/src/util/collection.rs
```diff
@@ -0,0 +1,339 @@
...
+///
+/// [collections.abc.MutableMapping]: https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping
+pub trait PyMutableMapping {
+    type Key;
+    type Value;
+
+    fn len(&self) -> PyResult<usize>;
+    fn contains(&self, key: Self::Key) -> PyResult<bool>;
+    fn keys(&self) -> PyResult<Vec<Self::Key>>;
+    fn values(&self) -> PyResult<Vec<Self::Value>>;
```
LukeMathWalker, line 37: I'm wondering if this is a footgun.
In Python you'd expect iterating over an existing collection to be "free", while our implementation triggers a deep clone of the entire map.

LukeMathWalker: I.e. as a Python developer, I wouldn't expect this:
```python
for elem in map:
    assert elem in elems
```
to trigger a deep clone of all elements in the map.

crisidev: I think this is a good first step, but me and Burak already talked about understanding what is needed to prevent the deep clone and borrow the map so that it becomes cheaper. I am not even sure if this is possible with the current PyO3 support.

LukeMathWalker: Btw, if you are already tracking this issue feel free to merge and we can revisit it later. But perhaps worth calling it out in the docs for those methods.

unexge: 👍 I've added comments to the related methods: https://github.com/awslabs/smithy-rs/pull/1871/commits/c8bd36852b3be76914b5637f855a1e9e7a972571.


> This would be less of an issue If both the header names and their values happened to be behind smart pointers.

We are converting them to `String`s while crossing to Python: https://github.com/awslabs/smithy-rs/blob/c8bd36852b3be76914b5637f855a1e9e7a972571/rust-runtime/aws-smithy-http-server-python/src/middleware/header_map.rs#L56-L66

PR #1877, opened 2022-10-19
Unpin xmlparser.

dvc94ch:
## Motivation and Context
Dependencies shouldn't be pinned. It causes dependency resolution failures as cargo selects `0.13.6` and then finds the constraint `=0.13.3` which it fails to resolve. Pinning dependencies is always a bad idea. Alternatively someone could improve the cargo version constraint solver.


Velfi:
@dvc94ch We originally added this pin because of an issue with escaping characters in references. I'm going to add a test to this PR before merging it to make sure we don't experience a regression.

Issue #1860, opened 2022-10-17
Honor the `default` trait


david-perez:
Neither clients or servers are honoring [the `default` trait](https://awslabs.github.io/smithy/2.0/spec/type-refinement-traits.html#smithy-api-default-trait) from IDL v2. This trait affects serialization behavior of clients and servers.

Note that as far as I know, [awslabs/smithy does not yet provide protocol tests](https://github.com/awslabs/smithy/issues/1452) exercising this trait, so we may have to write our own and contribute them upstream.

jdisanti:
I think only the server needs to implement this. The clients aren't supposed to send the default value since changing the default is not a backwards breaking change.

82marbag:
PR failing pending more info on https://github.com/awslabs/smithy/issues/1456

jdisanti:
I believe the client does this correctly now.

PR #1879, opened 2022-10-19
Support @default

82marbag:
## Motivation and Context

The `@default` trait is not supported.

## Description

Add support for the `@default` trait.

Issue: #1860
Reference for the trait:
https://awslabs.github.io/smithy/2.0/spec/type-refinement-traits.html

I have updated the simple model to use version 2. If it is better to split in two and keep the current at version 1, I will change this.

## Checklist


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -258,6 +275,42 @@ class BuilderGenerator(
                 withBlock("$memberName: self.$memberName", ",") {
                     // Write the modifier
                     when {
+                        !memberSymbol.isOptional() && member.hasTrait<DefaultTrait>() -> {
```
jdisanti, line 278: Clients shouldn't set defaults, so this code should be in the server module.

codegen-core/common-test-models/simple.smithy
```diff
@@ -132,3 +132,22 @@ structure StoreServiceBlobInput {
 structure StoreServiceBlobOutput {
 
 }
+
+@http(method: "POST", uri: "/default-value")
```
jdisanti, line 136: This should go into a Kotlin unit test for the BuilderGenerator so that it can be more thoroughly tested.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -258,6 +275,42 @@ class BuilderGenerator(
...
+                                        .entries
+                                        .filter { entry -> entry.value == value }
+                                        .map { entry -> symbolProvider.toEnumVariantName(EnumDefinition.builder().name(entry.key).value(entry.value.toString()).build())!! }
+                                        .first()
+                                    val symbol = symbolProvider.toSymbol(target)
+                                    rust(".unwrap_or($symbol::${variant.name})")
+                                }
+                                is IntegerShape, is FloatShape -> rust(".unwrap_or(${node.expectNumberNode().value})")
+                                is BooleanShape -> rust(".unwrap_or(${node.expectBooleanNode().value})")
+                                is StringShape -> rust(".unwrap_or(String::from(${node.expectStringNode().value.dq()}))")
```
jdisanti, line 301: Recommend using `unwrap_or_else` to avoid unnecessary allocations.

82marbag: Thanks!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
82marbag: Changed, thank you

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
```diff
@@ -105,9 +105,8 @@ fun StructureShape.renderWithModelBuilder(
     model: Model,
     symbolProvider: RustSymbolProvider,
     writer: RustWriter,
-    forWhom: CodegenTarget = CodegenTarget.CLIENT,
```
jdisanti, line 108: This change is going to result in the errors being generated incorrectly for server tests.

Perhaps we should move this function into `codegen-client` and make an equivalent in `codegen-server`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -268,4 +271,8 @@ class BuilderGenerator(
             }
         }
     }
+
+    open fun renderDefaultBuilder(writer: RustWriter, member: MemberShape) {
```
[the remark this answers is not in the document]
82marbag: We'll have a ServerBuilderGenerator when builders of builders is merged. Do we still want to refactor this or can we merge this with the new generator (and add one for the client)?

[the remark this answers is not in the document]
jdisanti: @david-perez and @hlbarber - Can you weigh in on this? I don't think I have enough context to make this call.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -258,6 +260,7 @@ class BuilderGenerator(
                 withBlock("$memberName: self.$memberName", ",") {
                     // Write the modifier
                     when {
+                        !memberSymbol.isOptional() && member.hasTrait<DefaultTrait>() -> renderDefaultBuilder(writer, member)
```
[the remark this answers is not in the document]
jdisanti: My apologies, I was wrong on this. I talked to the Smithy folks, and the client should actually set a default value if a value is not present. If a value is given by the default trait, then it should use that. Otherwise, it should use the zero value.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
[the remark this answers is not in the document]
82marbag: I believe this is not relevant anymore since it's the same code for both client and server

aws/sdk/integration-tests/ec2/tests/paginators.rs
```diff
@@ -19,7 +19,7 @@ fn stub_config() -> Config {
 /// EC2 replies with `<nextToken></nextToken>` which our XML parser parses as empty string and not "none"
 #[tokio::test]
 async fn paginators_handle_empty_tokens() {
-    let request= "Action=DescribeSpotPriceHistory&Version=2016-11-15&AvailabilityZone=eu-north-1a&InstanceType.1=g5.48xlarge&ProductDescription.1=Linux%2FUNIX";
+    let request= "Action=DescribeSpotPriceHistory&Version=2016-11-15&AvailabilityZone=eu-north-1a&DryRun=false&InstanceType.1=g5.48xlarge&MaxResults=0&ProductDescription.1=Linux%2FUNIX";
```
jdisanti, line 22: Apologies again, I should have been clearer. For clients, the default should be set for responses, but not for requests. From [the spec](https://awslabs.github.io/smithy/2.0/spec/type-refinement-traits.html#default-trait):
> To allow servers to change default values if necessary, clients SHOULD NOT serialize default values unless the member is explicitly set to the default value or marked with the [default trait](https://awslabs.github.io/smithy/2.0/spec/type-refinement-traits.html#default-trait). This implies that clients SHOULD implement a kind of "presence tracking" of defaulted members so that the member is only serialized if it is explicitly set to the default value.

There shouldn't be any changes to these tests.

We need to implement presence tracking anyway for nullability support: https://github.com/awslabs/smithy-rs/issues/1767

I think you have a couple of options for moving forward:
1. Feature flag handling the default trait for clients and have it disabled for now until presence tracking is implemented. File an issue to track finishing it.
2. Implement presence tracking in this PR.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -60,7 +76,7 @@ class OperationBuildError(private val runtimeConfig: RuntimeConfig) {
 // Setter names will never hit a reserved word and therefore never need escaping.
 fun MemberShape.setterName() = "set_${this.memberName.toSnakeCase()}"
 
-class BuilderGenerator(
+open class BuilderGenerator(
```
jdisanti, line 79:
```suggestion
class BuilderGenerator(
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
```diff
@@ -444,3 +463,48 @@ open class ServerCodegenVisitor(
...
+                    }
+                    .first()
+                val symbol = symbolProvider.toSymbol(target)
+                rust(".unwrap_or($symbol::${variant.name})")
+            }
+            is IntegerShape -> rust(".unwrap_or(${node.expectNumberNode().value})")
+            is FloatShape -> rust(".unwrap_or(${node.expectNumberNode().value.toFloat()})")
+            is BooleanShape -> rust(".unwrap_or(${node.expectBooleanNode().value})")
+            is StringShape -> rust(".unwrap_or_else(|| String::from(${node.expectStringNode().value.dq()}))")
+            is ListShape, is MapShape -> rust(".unwrap_or_default()")
```
david-perez, line 497: This is incorrect. In the case of list and map shapes, the default trait's value should be a document. We need to parse said document node and construct the equivalent value in Rust.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
```diff
@@ -444,3 +463,48 @@ open class ServerCodegenVisitor(
...
+            is DocumentShape -> {
+                when (node) {
+                    is NullNode -> rust(".unwrap_or_default()")
+                    is BooleanNode -> rust(".unwrap_or(${node.value})")
+                    is StringNode -> rust(".unwrap_or_else(|| String::from(${node.value.dq()}))")
+                    is NumberNode -> rust(".unwrap_or(${node.value})")
+                    else -> rust(".unwrap_or_default()")
+                }
+            }
```
david-perez, line 506: This can be removed entirely. The `default` trait cannot be applied to Smithy document shapes.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
```diff
@@ -444,3 +463,48 @@ open class ServerCodegenVisitor(
...
+            else -> rust(".unwrap_or_default()")
```
david-perez, line 507: Crash code generation here, if we reach this, it's unexpected.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderConstraintViolations.kt
```diff
@@ -69,8 +69,11 @@ class ServerBuilderConstraintViolations(
         writer.rustBlock("pub${ if (visibility == Visibility.PUBCRATE) " (crate) " else "" } enum $constraintViolationSymbolName") {
             renderConstraintViolations(writer)
         }
-        renderImplDisplayConstraintViolation(writer)
-        writer.rust("impl #T for ConstraintViolation { }", RuntimeType.StdError)
+
+        if (all.isNotEmpty()) {
```
david-perez, line 73: Actually, this makes me realize the current implementation is wrong. It should `check` in the beginning of `render` that `all` is not empty and crash if it is. We don't want to render the constraint violation struct at all if no members are constrained.

david-perez: The rest of the changes in this file should not be needed then.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -524,8 +527,12 @@ class ServerBuilderGenerator(
                             }
                         }
                     }
-                    serverBuilderConstraintViolations.forMember(member)?.also {
-                        rust(".ok_or(ConstraintViolation::${it.name()})?")
+                    if (member.hasNonNullDefault()) {
```
david-perez, line 530: We need to inject the writable _before_ we check constraints, so that default values undergo validation too.

At least until Smithy validates the constraints and default values at build time.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorWithoutPublicConstrainedTypes.kt
```diff
@@ -160,6 +164,10 @@ class ServerBuilderGeneratorWithoutPublicConstrainedTypes(
                 withBlock("$memberName: self.$memberName", ",") {
                     serverBuilderConstraintViolations.forMember(member)?.also {
                         rust(".ok_or(ConstraintViolation::${it.name()})?")
+                    } ?: run {
```
david-perez, line 167: Same here, we need to inject the writable before constraint checking.

We also need unit tests that exercise this builder type.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedUnionGenerator.kt
```diff
@@ -146,12 +146,13 @@ class UnconstrainedUnionGenerator(
 
             if (shape.isReachableFromOperationInput()) {
                 rustBlock("impl $constraintViolationName") {
+                    val constraintViolations = constraintViolations()
                     rustBlockTemplate(
-                        "pub(crate) fn as_validation_exception_field(self, path: #{String}) -> crate::model::ValidationExceptionField",
+                        "pub(crate) fn as_validation_exception_field(self, ${if (constraintViolations.isEmpty()) "_" else ""}path: #{String}) -> crate::model::ValidationExceptionField",
```
david-perez, line 151: I don't understand this change. `UnconstrainedUnionGenerator` must surely only be called when at least one variant has a constraint, so this should never be empty.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorTest.kt
```diff
@@ -0,0 +1,78 @@
...
+class ServerBuilderGeneratorTest {
+    @Test
+    fun `generate default values`() {
+        val model =
+            """
+            namespace com.test
+            use aws.protocols#restJson1
+            use smithy.framework#ValidationException
+
+            structure MyStruct {
```
david-perez, line 33: We need _a lot_ more testing.

We need to test all shape combinations (including documents and collections), with and without `@required`, with `@default(null)`  etc.

_Ideally_, we should also test binding of the members to all parts of the HTTP message (headers, query string, uri path, payload etc). But that's not a strict requirement, since it shouldn't affect our builder logic and this is a unit test after all. Still, it should be easy to do.

I suggest you synthesize the model programmatically so that you don't have to manually write out each test case.

The tests should assert both cases when the user sets a value and the default does not get enacted, and when they don't and the default does get enacted.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+            @default(42)
+            integer PrimitiveInteger
+            """.asSmithyModel(smithyVersion = "2")
+        val provider = serverTestSymbolProvider(model)
+        val writer = RustWriter.forModule("model")
+        writer.rust("##![allow(deprecated)]")
+        val struct = model.lookup<StructureShape>("com.test#MyStruct")
+
+        val (_, testDir) = generatePluginContext(model)
```
david-perez, line 52: This test is using the old testing machinery. Use the modern way of writing unit tests (i.e. starting from a project `val project = TestWorkspace.testProject(symbolProvider)`.

david-perez (changes requested):
The implementation is a good start, but needs more work and we especially need _a lot_ more testing to be confident that it is correct.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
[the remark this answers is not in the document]
david-perez: This is no longer relevant. Now that the server has its own builders, we can place the implementation of default there until the client implements presence tracking.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
82marbag: From the [docs](https://smithy.io/2.0/spec/type-refinement-traits.html#default-trait) I read:

> The following shapes have restrictions on their default values:
> ...
> list: can only be set to an empty list.
> map: can only be set to an empty map.

Where is the different default?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
82marbag: (as above) [here](https://smithy.io/2.0/spec/type-refinement-traits.html#default-trait) I read:

> document: can be set to null, `true, false, string, numbers, an empty list, or an empty map.

david-perez: You're right! My bad, I only read the trait selector; I didn't now `document` shapes were simple types.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
david-perez: You're right! My bad, I hadn't read that section. I had assumed that since the value for the `@default` trait was a document, it could be arbitrarily complex.

david-perez (changes requested):

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorTest.kt
```diff
@@ -54,4 +123,195 @@ class ServerBuilderGeneratorTest {
...
+            TestConfig(defaultValues, requiredTrait = false, nullDefault = false, optional = false),
+            TestConfig(customValues, requiredTrait = true, nullDefault = false, optional = false),
+            TestConfig(defaultValues, requiredTrait = false, nullDefault = true, optional = true),
+        ).flatMap { builderWriters.stream().map { builderWriter -> Arguments.of(it, builderWriter) } }
+    }
+
+    data class TestConfig(
+        val values: Map<String, String?>,
+        val requiredTrait: Boolean,
+        val nullDefault: Boolean,
```
david-perez, line 314: So when `nullDefault` is `true`, that means the test should be being generated without setting any values, and relying on the modeled default values, right? If yes, this probably needs a better name, like `relyOnModeledDefaultValues`, at first I thought it had to do with `@default(null)`.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorTest.kt
```diff
@@ -54,4 +123,195 @@ class ServerBuilderGeneratorTest {
...
+        return Stream.of(
+            TestConfig(customValues, requiredTrait = true, nullDefault = true, optional = false),
+            TestConfig(defaultValues, requiredTrait = false, nullDefault = false, optional = false),
+            TestConfig(customValues, requiredTrait = true, nullDefault = false, optional = false),
+            TestConfig(defaultValues, requiredTrait = false, nullDefault = true, optional = true),
+        ).flatMap { builderWriters.stream().map { builderWriter -> Arguments.of(it, builderWriter) } }
+    }
+
+    data class TestConfig(
+        val values: Map<String, String?>,
```
david-perez, line 312: These are the asserted values right?

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorTest.kt
```diff
@@ -54,4 +123,195 @@ class ServerBuilderGeneratorTest {
...
+    fun `default values are generated and builders respect default and overrides`(testConfig: TestConfig, setupFiles: (w: RustWriter, m: Model, s: RustSymbolProvider) -> Unit) {
+        val valuesMap = defaultValues.mapValues { if (testConfig.nullDefault) null else it.value }
+        val model = generateModel(testConfig.requiredTrait, valuesMap)
+        val symbolProvider = serverTestSymbolProvider(model)
+        val project = TestWorkspace.testProject(symbolProvider)
+        project.withModule(RustModule.public("model")) {
+            setupFiles(this, model, symbolProvider)
+
+            val rustValues = setupValuesForTest(testConfig.values)
+            val setters = if (testConfig.requiredTrait) structSetters(rustValues) else writable { }
```
david-perez, line 138: This seems odd. It's only setting the values on the builder when the members are `@required`. What we want is to not set them when the members are `@default`, and test that the built structure uses the modeled default values.

`@required` should have no bearing on our assertions / generated code. Note [the spec](https://smithy.io/2.0/spec/aggregate-types.html#structure-member-optionality) says both `@default` and `@required` should generate non-`Option`al structure fields, so we're only testing that case out of an abundance of caution.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorTest.kt
```diff
@@ -54,4 +123,195 @@ class ServerBuilderGeneratorTest {
...
+            TestConfig(customValues, requiredTrait = true, nullDefault = false, optional = false),
+            TestConfig(defaultValues, requiredTrait = false, nullDefault = true, optional = true),
+        ).flatMap { builderWriters.stream().map { builderWriter -> Arguments.of(it, builderWriter) } }
+    }
+
+    data class TestConfig(
+        val values: Map<String, String?>,
+        val requiredTrait: Boolean,
+        val nullDefault: Boolean,
+        val optional: Boolean,
```
david-perez, line 315: Not sure what the role of `optional` is here. Whether the builder is fallible or not is entirely determined by whether one or more members is `required` or `default`, so we shouldn't need to specify this; we can derive it from the rest of the state.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorTest.kt
```diff
@@ -54,4 +123,195 @@ class ServerBuilderGeneratorTest {
...
+        val builderWriters = listOf(::writeServerBuilderGenerator, ::writeServerBuilderGeneratorWithoutPublicConstrainedTypes)
+        return Stream.of(
+            TestConfig(customValues, requiredTrait = true, nullDefault = true, optional = false),
+            TestConfig(defaultValues, requiredTrait = false, nullDefault = false, optional = false),
+            TestConfig(customValues, requiredTrait = true, nullDefault = false, optional = false),
+            TestConfig(defaultValues, requiredTrait = false, nullDefault = true, optional = true),
+        ).flatMap { builderWriters.stream().map { builderWriter -> Arguments.of(it, builderWriter) } }
+    }
+
+    data class TestConfig(
```
david-perez, line 311: This is the most important piece of information a reader needs to understand what the test is testing. Each field needs documentation on how it impacts the generated code and the assertions we make.

david-perez (changes requested):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderConstraintViolations.kt
```diff
@@ -190,12 +206,16 @@ enum class ConstraintViolationKind {
 
     // An unconstrained type was provided for a field targeting a constrained shape, but it failed to convert into the constrained type.
     CONSTRAINED_SHAPE_FAILURE,
+
+    // A default was provided but did not pass constraint checks
+    CONSTRAINED_DEFAULT_FAILURE,
```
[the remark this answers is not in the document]
david-perez: Opened https://github.com/awslabs/smithy-rs/issues/2102 for the latter problem.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorTest.kt
```diff
@@ -54,4 +123,195 @@ class ServerBuilderGeneratorTest {
...
+            document DocumentNumberPosInt
+            document DocumentNumberFloat
+            document DocumentList
+            document DocumentMap
+            """
+        return model.asSmithyModel(smithyVersion = "2")
+    }
+
+    private fun localParameters(): Stream<Arguments> {
+        val builderWriters = listOf(::writeServerBuilderGenerator, ::writeServerBuilderGeneratorWithoutPublicConstrainedTypes)
```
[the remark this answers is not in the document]
82marbag: I can't because the stream is consumed after the first use, that's why it's constructed again

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -257,6 +290,9 @@ class ServerBuilderGenerator(
             "fn build_enforcing_all_constraints(self) -> #{ReturnType:W}",
             "ReturnType" to buildFnReturnType(isBuilderFallible, structureSymbol),
         ) {
+            if (members.mapNotNull { serverBuilderConstraintViolations.forDefaultMember(it) }.isNotEmpty()) {
+                Attribute.Custom("allow(clippy::useless_conversion)").render(this)
```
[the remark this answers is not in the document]
82marbag: I made codegen a bit more complex to remove this and left a comment where needed

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -496,60 +537,228 @@ class ServerBuilderGenerator(
...
 
                 withBlock("$memberName: self.$memberName", ",") {
-                    // Write the modifier(s).
-                    serverBuilderConstraintViolations.builderConstraintViolationForMember(member)?.also { constraintViolation ->
-                        val hasBox = builderMemberSymbol(member)
-                            .mapRustType { it.stripOuter<RustType.Option>() }
-                            .isRustBoxed()
-                        if (hasBox) {
-                            rustTemplate(
+                    val defaultWrap: (inner: String) -> String = { inner ->
```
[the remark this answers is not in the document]
82marbag: Talked over this offline: we decided to revert back to the strategy of relying on java/kotlin checks (generation time) for constrained default values. The current path would lead to complexity and maintainability issues

david-perez:
We need a changelog entry.

david-perez (approved):

PR #1882, opened 2022-10-19
Upgrade `cargo-check-external-types` to 0.1.5

jdisanti:
## Motivation and Context
Version 0.1.5 fixes a bug where external types exposed in an associated type in a trait impl weren't detected.


Velfi (approved):

rcoh (approved):

aws/rust-runtime/aws-inlineable/external-types.toml
```diff
@@ -1,6 +1,10 @@
 allowed_external_types = [
+    "aws_endpoint::*",
+    "aws_http::*",
+    "aws_sig_auth::*",
```
ysaito1001, line 4: Noticed that `allowed_external_types` in `aws/sdk/sdk-external-types.toml` contains `"aws_smithy_async::*"`. Does `allowed_external_types` in this file need to contain `"aws_smithy_async::*"` as well?

jdisanti: This is a good observation.

Essentially, `aws-inlineable` and the generated SDK crates are one in the same since the inlineables end up in the generated crates. So there's definitely an argument to be made for them to just have one config file across both. I think making `aws-inlineable/external-types.toml` a symlink to `sdk-external-types.toml` is probably the right balance here so that it's still easy to run the tool against `aws-inlineable` locally since it's consistent with the rest.

jdisanti: Looking at it closer, I think we should keep them as separate files for now since `sdk-external-types.toml` has more types in it with TODO comments on them. If we consolidate now, it will make it easier to introduce more external types into the public API that will need to be fixed later.

PR #1887, opened 2022-10-20
Pin `derive_arbitrary` version

jdisanti:
## Motivation and Context
Version 1.2.0 of `derive_arbitrary` requires Rust 1.63, so we need to pin the version to the previous release since our MSRV is 1.62.


david-perez:
We need to update [`CODEOWNERS`](https://github.com/awslabs/smithy-rs/blob/main/CODEOWNERS). Server team is currently not being added as reviewer to this PR because we're only marked as owners of `aws-smithy-http-server`, but in reality any change to _most_ crates under `rust-runtime/` (like `aws-smithy-eventstream`) is depended upon by server SDKs.

Velfi (approved):

PR #1897, opened 2022-10-24
Forbid list/map as paylods

82marbag:
...and add tracking to the issue in unions in paylods

## Motivation and Context

Partially address https://github.com/awslabs/smithy-rs/issues/1896

## Checklist


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
```diff
@@ -272,6 +272,8 @@ class HttpBoundProtocolPayloadGenerator(
                                 """,
                             )
                             is StructureShape -> rust("#T()", serializerGenerator.unsetStructure(targetShape))
+                            is UnionShape -> throw CodegenException("Currently unsupported. Tracking issue: https://github.com/awslabs/smithy-rs/issues/1896")
+                            else -> throw CodegenException("List and Map are unsupported payloads")
```
david-perez, line 276: This arm could get hit by other shape types too if Smithy decides to support them in the future.

```suggestion
                            else -> throw CodegenException("`httpPayload` on member shapes targeting shapes of type ${targetShape.type()} is unsupported")
```

PR #1899, opened 2022-10-24
Add JSON serialization input/output extension points

hlbarber:
## Motivation and Context

Third-party plugins might want to provide customizations to the JSON input/output structures.

## Description

- Add `JsonSection.InputStruct` and `JsonSection.OutputStruct` classes.
- Add entry points for respective customizations.

## Notes

- We call both `OutputStruct` and `ServerError` for errors, do we want this behavior? Or should we call only `ServerError` for errors?

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerAwsJson.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerAwsJson.kt
@@ -71,7 +71,7 @@ class ServerAwsJsonError(private val awsJsonVersion: AwsJsonVersion) : JsonCusto
                 rust("""${section.jsonObject}.key("__type").string("${escape(typeId)}");""")
             }
         }
-        else -> writable {}
+        else -> emptySection
     }
 }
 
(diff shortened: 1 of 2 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerAwsJson.kt
```diff
@@ -71,6 +71,7 @@ class ServerAwsJsonError(private val awsJsonVersion: AwsJsonVersion) : JsonCusto
                 rust("""${section.jsonObject}.key("__type").string("${escape(typeId)}");""")
             }
         }
+        else -> writable {}
```
david-perez, line 74: You can use `emptySection` from `NamedSectionGenerator`.

david-perez (approved):

david-perez:
> Or should we call only `ServerError` for errors?

This sounds better no? `OutputStruct` for operation outputs, `ServerError` for operation errors.

hlbarber:
@david-perez How's this? https://github.com/awslabs/smithy-rs/pull/1899/commits/e7835f69c9bfc02b694dfb14e01c50861ddaac61

jdisanti (approved):

PR #1909, opened 2022-10-26
Re-export runtime types in server generated crates

### LINKED ISSUE #1506: Re-export runtime type in server generated crates
We can re-export certain types coming from `aws-smithy-http-server` crate inside the generated crate.

So far we agreed we can re-export `aws_smithy_http_server::Router` and we are debating over `aws_smithy_http_server::Extension`.

82marbag:
Re-export commonly used types from aws-smithy-http-server so that customers don't have to depend on that crate.

We did not agree to a full list yet. This is a possible starting point. We could also either add all types from the server runtime crate.

## Motivation and Context

Closes #1506 

## Description

Add a `server` module that re-exports some of the types from the runtime crate.

## Testing

Generated the Pokemon service.

## Checklist


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
```diff
@@ -94,6 +94,17 @@ open class ServerServiceGenerator(
         }
 
         renderExtras(operations)
+
+        rustCrate.withModule(
+            RustModule.public(
+                "server",
```
hlbarber, line 100: Alternatively here we could

```rust
pub use aws_smithy_http_server as server;
```

Which would allow us to preserve the root crate docs but would mean that we re-export _everything_ from the `aws-smithy-http-server` crate.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerRuntimeTypesReExportsGenerator.kt
```diff
@@ -0,0 +1,40 @@
...
+            """
+            pub use #{SmithyHttpServer}::Extension;
+            pub use #{Router};
+            pub use #{SmithyHttpServer}::operation::OperationShape;
+            pub use #{SmithyHttpServer}::operation::Operation;
+
+            ##[doc(hidden)]
+            pub use #{SmithyHttpServer}::instrumentation;
+            ##[doc(hidden)]
+            pub use #{SmithyHttpServer}::proto;
```
LukeMathWalker, line 35: What's the point of doing a #[doc(hidden)] re-export? I might be missing something obvious 🤔

hlbarber: It means that https://github.com/awslabs/smithy-rs/pull/1886 would only have to remove the `#[doc(hidden)]` rather than making decisions on what to include in this module or not.

This is a non-issue if we do https://github.com/awslabs/smithy-rs/pull/1909#discussion_r1005782072 though.

LukeMathWalker: > If #[doc(hidden)] is transitive and re-exporting something #[doc(hidden)] means the re-export is also hidden then this isn't necessary - I'm unsure if this is the case or not.

This was my expectation, hence the question.
I tested it and I found that:
- Re-exporting a `#[doc(hidden)]` module gives you a `#[doc(hidden)]` module, so no need to mark the rexport as `#[doc(hidden)]`;
- Any `#[doc(hidden)]` item inside the re-exported module remains `#[doc(hidden)]`.

LukeMathWalker: Code:
```rust
pub use dep::hello as hey;
pub use dep::my_hidden_module as hey_from_hidden;
```
Dependency code:
```rust
pub mod hello {
    #[doc(hidden)]
    struct MyHiddenStruct;
}

#[doc(hidden)]
pub mod my_hidden_module {
    struct MyVisibleStruct;
}
```
Screenshot of the docs:
image
image

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
hlbarber: The conclusion of a discussion about this was: no, we don't want to do a blanket re-export for now. But we should ensure that we have a 1:1 mapping of codegen module : codegen module to ensure that, in the future, if we reconsider it will be a non-breaking change.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerRuntimeTypesReExportsGenerator.kt
hlbarber: Good to know, we can drop the `#[doc(hidden)]` from these.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
crisidev: Isn't `server` a confusing name? What about calling the module `aws_smithy_http_server` to better signal where we are re-exporting from? Not a blocker TBH.

crisidev: Now that I think about this more, there will be probably naming collisions.. Forget this comment.

82marbag: If we want to hide where re-exports are from and allow us to re-export also from other runtime crates possibly, probably we shouldn't signal the name of the crate

LukeMathWalker: It feels like `server` is clear for context and it buys us a bit of flexibility, as @82marbag was saying.

crisidev: Allright! Thanks for the comments.

crisidev (approved):

PR #1911, opened 2022-10-26
Slow down crate publish more

jdisanti:
## Motivation and Context
Throttling still occurs every 18 crates or so when publishing the SDK, although this doesn't cause anything to fail since it gets retried with a 30 second delay. This PR doubles the amount of time the publisher tool waits in between crates so that throttling occurs less frequently. Overall, this should reduce the total time since 18 seconds + 30 second retry timeout is greater than 36 seconds with no throttling.

### DIFF
--- a/tools/publisher/src/subcommand/publish.rs
+++ b/tools/publisher/src/subcommand/publish.rs
@@ -66,15 +66,15 @@ pub async fn subcommand_publish(
             if !is_published(&package.handle).await? {
                 publish(&package.handle, &package.crate_path).await?;
 
+                // Keep things slow to avoid getting throttled by crates.io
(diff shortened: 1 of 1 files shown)


weihanglo:
Just FYI, there is a new unstable flag `-Zpublish-timeout` that may be helpful.

https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#publish-timeout

jdisanti:
> Just FYI, there is a new unstable flag -Zpublish-timeout that may be helpful.

At some point, that will enable us to get rid of the eventual consistency check, but I think we'll still have to do some additional sleeping to ensure we're not publishing too quickly.

Velfi (approved):

PR #1916, opened 2022-10-27
Fix repeated options when creating an issue

jjant:
## Description
After https://github.com/awslabs/smithy-rs/pull/1913 our issue options look like this:

image

Note the repeated option to view the security policy and the link on the bottom to create a blank issue (we already have that option).

The first happens because GH already links to the security policy if you have a `SECURITY.md` file.
The second is because we've added [this template](https://github.com/awslabs/smithy-rs/pull/1913/files#diff-5d7416ebfc56443c702e82a607f1ed7656f51a9138720a45d0781c84c4d53439), but kept `blank_issues_enable: true` in the config.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,1 @@
-blank_issues_enabled: true
-contact_links:
-- name: Report a security vulnerability
-  url: https://github.com/awslabs/smithy-rs/security/policy
-  about: Please review our security policy for more details
+blank_issues_enabled: false


Velfi (approved):

weihanglo (approved):
Seems good to me.

Just curious. Our SECURITY.md comes from [here](https://github.com/awslabs/.github/blob/master/SECURITY.md), right?

jjant:
> Seems good to me.
> 
> Just curious. Our SECURITY.md comes from [here](https://github.com/awslabs/.github/blob/master/SECURITY.md), right?

@weihanglo It's not super obvious but [it looks like it does](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file#about-default-community-health-files).

> You can **add default community health files to a public repository called .github**, in the root of the repository or in the docs or .github folders.

PR #1934, opened 2022-11-01
Revamp errors in `aws-config`

jdisanti:
## Motivation and Context
This PR revamps errors in `aws-config` to adhere to [RFC-0022: Error Context and Compatibility](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0022_error_context_and_compatibility.md).

This PR is part of a series that will fully implement the RFC.

## Checklist


aws/rust-runtime/aws-config/src/imds/client.rs
```diff
@@ -237,7 +232,9 @@ impl Client {
         &self,
         path: &str,
     ) -> Result<Operation<ImdsGetResponseHandler, ImdsResponseRetryClassifier>, ImdsError> {
-        let mut base_uri: Uri = path.parse().map_err(|_| ImdsError::InvalidPath)?;
+        let mut base_uri: Uri = path.parse().map_err(|_| {
+            ImdsError::unexpected("IMDS path was not a valid URI. Hint: does it begin with `/`?")
```
Velfi, line 236: I think having a hint is good but it's not clear if the hint is saying that it should or should not begin with `/`

Velfi:
```suggestion
            ImdsError::unexpected("IMDS path was not a valid URI. Hint: valid URIs begin with `/`")
```

aws/rust-runtime/aws-config/src/imds/client/error.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Returns the raw response from IMDS
+    pub fn response(&self) -> &http::Response<SdkBody> {
+        &self.raw
+    }
+}
+
+/// Error context for [`ImdsError::IoError`]
+#[derive(Debug)]
+pub struct IoError {
+    source: Box<dyn Error + Send + Sync + 'static>,
```
Velfi, line 44: We use this so much in so many places in the runtime crates that I wonder if we should type alias this and enshrine it somewhere.

Velfi (approved):

aws/rust-runtime/aws-config/src/ecs.rs
```diff
@@ -237,11 +237,11 @@ impl Display for EcsConfigurationErr {
...
 }
 
-impl Error for EcsConfigurationErr {
+impl Error for EcsConfigurationError {
     fn source(&self) -> Option<&(dyn Error + 'static)> {
         match &self {
-            EcsConfigurationErr::InvalidRelativeUri { err, .. } => Some(err),
-            EcsConfigurationErr::InvalidFullUri { err, .. } => Some(err),
+            EcsConfigurationError::InvalidRelativeUri { err, .. } => Some(err),
+            EcsConfigurationError::InvalidFullUri { err, .. } => Some(err),
```
jdisanti, line 244: This `source` method shouldn't use the catch all matcher per the RFC.

PR #1936, opened 2022-11-03
RFC: Client Crate Organization

jdisanti:
## Motivation and Context
This RFC proposes some organizational changes to the generated SDK and client crates to make navigating the documentation easier while also facilitating organized re-export of runtime crate types.

[Rendered](https://github.com/awslabs/smithy-rs/blob/jdisanti-rfc-client-crate-org/design/src/rfcs/rfc0026_client_crate_organization.md)


jdisanti:
This is very much a work in progress. I put some ideas down on paper, but I'm not entirely happy with it and would love suggestions.

There is a TODO in the errors section that especially needs attention. I'm wondering if, instead of what the RFC is currently proposing, we should put all code generated types inside of `crate::model` with some submodules to split up inputs/outputs/errors/etc.

design/src/rfcs/rfc0023_client_crate_organization.md
```diff
@@ -0,0 +1,378 @@
...
+├-- Credentials
+├-- Endpoint
+├-- Error
+├-- ErrorExt (for some services)
+├-- PKG_VERSION
+└── Region
+```
+
+The `AppName` is infrequently set, and should be moved into `crate::config`.
+The `Error` and `ErrorExt` types should be moved into `crate::error` since
```
rcoh, line 112: totally agree about `ErrorExt`. Depending on where you land on errors, I think keeping `Error` in the root so you can do without any duplicate `error::Error` is probably worthwhile.

The other stuff in the root like `Endpoint` should be moved. `PKG_VERSION` could be moved into a `meta` module. It's debatable if `Region` needs to be re-exported at all now that we have aws-config.

```
async fn do_a_thing() -> Result<String, aws_sdk_s3::Error> -> { ... }
```

design/src/rfcs/rfc0023_client_crate_organization.md
```diff
@@ -0,0 +1,378 @@
...
+    |   ├-- Operation
+    |   └-- RetryKind
+    └-- parsers (private/doc hidden)
+        └-- <One struct per operation>
+```
+
+Empty Modules
+-------------
+
+The `lens` and `http_body_checksum` modules have nothing inside them,
```
rcoh, line 283: if lens isn't private that's a bug 🤭

design/src/rfcs/rfc0023_client_crate_organization.md
```diff
@@ -0,0 +1,378 @@
...
+
+> `http_body_checksum`: Functions for modifying requests and responses for the purposes of checksum validation
+
+These modules hold private functions that are used by other generated code, and should just be made
+private or `#[doc(hidden)]` if necessary.
+
+New Organization
+----------------
+
+All combined, the following is the new organization:
```
rcoh, line 296:
```suggestion
All combined, the following is the new publicly visible organization:
```

rcoh (approved):
overall I really like this! I think we should do some more line-by-line nitpicking at a later date

design/src/rfcs/rfc0023_client_crate_organization.md
jdisanti: Your reason for keeping `Error` in the top-level makes sense to me.

I think `Region` should still be re-exported since it's referenced by `Config` and the config's `Builder`. It can be moved into `crate::config` though.

jdisanti:
The changes being proposed in this RFC will also impact server crate organization unless we do additional work to keep the server the same as it is now. Adding the `needs-server-review` tag to get some eyes on this.

design/src/rfcs/rfc0024_client_crate_organization.md
```diff
@@ -0,0 +1,444 @@
...
+├── Endpoint
+├── Error
+├── ErrorExt (for some services)
+├── PKG_VERSION
+└── Region
+```
+
+The `AppName` is infrequently set, and should be moved into `crate::config`. The `aws-config` crate
+drastically reduces the usefulness of `Credentials`, `Endpoint`, and `Region`, so these can move
+into `crate::config`. `ErrorExt` should be moved into `crate::error`, but `Error` should stay in
```
Velfi, line 115:
```suggestion
The `AppName` is infrequently set, and should be moved into `crate::config`. We currently encourage users to import `Credentials`, `Endpoint`, and `Region` from the `aws-config` crate, so service crate reexports of these types can move into `crate::config`. `ErrorExt` should be moved into `crate::error`, but `Error` should stay in
```

Velfi (commented):
Looking at this RFC, it's making me wonder again whether we should not also be feature-gating each operation in order to shorten compile times. We could generate the feature flags for all ops and then include them all by default. Users wanting short compile times would disable default features and include only what they use.

jdisanti:
Decided to move the "combine `Error`/`ErrorKind`" section into its own RFC: https://github.com/awslabs/smithy-rs/pull/2075

jdisanti:
The RFC has been updated to organize code generated types by operation.

Velfi (approved):

design/src/rfcs/rfc0026_client_crate_organization.md
```diff
@@ -0,0 +1,396 @@
...
+├── error
+|   ├── DisplayErrorContext (*)
+|   ├── ErrorExt (for some services)
+|   └── SdkError (*)
+├── meta
+|   └── PKG_VERSION
+├── middleware
+|   └── DefaultMiddleware
+├── operation
+|   ├── customize
```
jdisanti, line 337: From an offline discussion: `customize` should be moved into the `client` module to better avoid name collisions.

PR #1938, opened 2022-11-03
Remove our `base64` implementation in favour of `base64-simd`

jjant:
## Motivation and Context
Fixes (https://github.com/awslabs/smithy-rs/issues/1277), that is, the following protocol tests now pass:
```
RestJsonBodyMalformedBlobInvalidBase64_case1
RestJsonBodyMalformedBlobInvalidBase64_case2
RestJsonHeaderMalformedStringInvalidBase64MediaType_case1
```

## Description
- `base64-simd` requires Rust 1.62.1 to build, so I had to upgrade it here.

## Performance
`base64-simd` is around 30% to 500% faster than our previous hand-rolled implementation (on my M1 MBP at least).

image


rcoh (approved):
LGTM, in the future please split out MSRV upgrade PRs in case they need to get reverted

crisidev (approved):

PR #1942, opened 2022-11-03
RFC: RequestId for services

82marbag:
## Motivation and Context
This RFC proposes to give customers `RequestId`, a unique identifier for requests.

[Rendered document](https://github.com/awslabs/smithy-rs/blob/070f83d0bcf55089d9d0588daa8bc529e0554faa/design/src/rfcs/rfc0023_request_id.md)

Thanks to @hlbarber !

-----

RFC revision 1:
* Do not open a span automatically
* Have `ServerRequestId` and `ClientRequestId`

-----

RFC revision 2:
* Support updates to the generated ID
* Remove: generate a client ID if none is provided
* Make the client ID an `Option` to make it clear if it was set


design/src/rfcs/rfc0023_request_id.md
```diff
@@ -0,0 +1,103 @@
+RFC: RequestID in business logic handlers
+=============
+
+> Status: RFC
```
jdisanti, line 4: Does this apply to clients, servers, or both?

82marbag: Servers only. I'll make it clear in the RFC

design/src/rfcs/rfc0023_request_id.md
```diff
@@ -0,0 +1,105 @@
...
+2. RequestId format:
+
+Common formats for RequestIDs are:
+
+* UUID: a random string, represented in hex, of 128 bits from IETF RFC 4122: `7c038a43-e499-4162-8e70-2d4d38595930`
+* The hash of a sequence such as `date+thread+server`: `734678902ea938783a7200d7b2c0b487`
+* A verbose description: `current_ms+hostname+increasing_id`
+
+For privacy reasons, any format that provides service details should be avoided. A random string is preferred.
+The proposed format is to use UUID, version 4.
```
LukeMathWalker, line 64: Have we considered v7?
It is pretty much equivalent in terms of uniqueness guarantees, but it provides better performance if a service decides to store them.

design/src/rfcs/rfc0023_request_id.md
```diff
@@ -0,0 +1,105 @@
...
+
+A `Service` that inserts a RequestId in the extensions:
+```rust
+fn call(&mut self, mut req: http::Request<R>) -> Self::Future {
+    let req_id = RequestId::new();
+    let span = span!(request_id = %req_id);
+    request.extensions_mut().insert(req_id);
+    self.inner.call(request).instrument(span)
+}
+```
```
LukeMathWalker, line 98: I'd advice against coupling the insertion of `RequestId` into the extensions map with the creation of a root span.
Most services will want to customise the root span and we don't want to introduce an additional unnecessary span in their telemetry data.

We could provide a basic `TelemetryService` that could be used as default and takes care of injecting the `RequestId`, but it should be a separate service.

design/src/rfcs/rfc0023_request_id.md
82marbag: I'll take a look, but it seems to me like the RFC is still in draft for v7

LukeMathWalker: Yes, it is still in draft. 
Do we consider the format to be part of the public API of the request ID (i.e. is it exposed to users)? Or do we simply want to users to treat is as an opaque new type that can be converted into a `String` (therefore allowing us to evolve the internal implementation at a later date)?

design/src/rfcs/rfc0023_request_id.md
LukeMathWalker: Gotcha - I read this snippet as "this is what the service will look like" instead of "this is what users could do with it". 
If the latter, I'd expect them to retrieve an already-generated `RequestId` from the extensions rather than doing the work of creating and storing it.

design/src/rfcs/rfc0023_request_id.md
82marbag: The idea would be to provide an opaque type. We can change it as long as we're in preview, but we need it to be stable; service owners can write tooling to use that ID. We should choose one and stick to it unless we really need to

design/src/rfcs/rfc0023_request_id.md
LukeMathWalker: I'd lean against that - if they end up using both layers by mistake, things will get confusing since they will get two different ids generated and used in different parts of the app.
That won't happen if `RequestIdService` and `TelemetryService` are orthogonal (i.e. you can use the first or both, but the latter requires the first to get a valid `RequestId`).

hlbarber: The downside of vending them orthogonal is that `TelemetryService` is an additional and fallible type map lookup. There's also a question here of whether or not a `span!` with a lone `RequestId` is a problem?

If we do decide to do it we should perhaps look to extend `aws-smithy-http-server::instrumentation::service::InstrumentOperation`.

LukeMathWalker: > There's also a question here of whether or not a span! with a lone RequestId is a problem?

I'd say that it depends on the span level. At the `info` level, I think it's a problem - it's most likely not going to be filtered out in production and it's going to add 1 or 2 telemetry data points for every request - that can be costly and add to the overall telemetry noise if it's not providing enough information on its own.

> The downside of vending them orthogonal is that TelemetryService is an additional and fallible type map lookup. 
> If we do decide to do it we should perhaps look to extend aws-smithy-http-server::instrumentation::service::InstrumentOperation.

I think that it's OK to couple generation and logging if we allow for the insertion of custom properties into the span - then the user can benefit from out default plus whatever they care about (e.g. something like [`root_span!`](https://docs.rs/tracing-actix-web/latest/tracing_actix_web/macro.root_span.html) is one way of doing it).
`InstrumentOperation` seems like a good  spot to build on top of.

hlbarber: There’s a reasonable chunk of design work to be done around the `InstrumentOperation` API, at the time of writing,  just wraps the `http::Request` and `http::Response` parts with sensitive wrappers, opens a request span and emits an response event - no user configuration possible.

Perhaps we want it closer to what `TraceLayer` does, it’s unclear to me whether we reuse `TraceLayer` or not because the closures provided to each `TraceLayer` are ~ `|request: http::Request| Span/Event` / `|response: http::Response| Span/Event` / etc and we don’t really want to provide the customer with the entire `http::Request` - we instead want to provide them with the `http::Request`/`http::Response` parts in wrappers which redact sensitive information.

Inserting(?) and recording a `RequestId` is a similar level of configuration which could be added to the `InstrumentOperation`.

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,132 @@
...
+-----------
+
+- **RequestID**: a service-wide unique identifier
+- **UUID**: a universally unique identifier
+
+RequestID is an element that uniquely identifies a client request. RequestID is used by services to map all logs, events and
+specific data to a single operation. This RFC discusses whether and how smithy-rs can make that value available to customers.
+
+Services use a RequestID to collect logs related to the same request and see its flow through the various operations,
+help clients debug requests by sharing this value and, in some cases, use this value to perform their business logic. RequestID is unique across a service at least within a certain timeframe.
```
LukeMathWalker, line 20: Should the server-generated request ID be echoed back to the client as a header on the returned response?

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,132 @@
...
+    fn call(&mut self, mut req: http::Request<R>) -> Self::Future {
+        request.extensions_mut().insert(ServiceRequestId::new());
+        self.inner.call(req)
+    }
+}
+```
+
+For client request IDs, the process will be, in order:
+* If a header is found matching one of the possible ones, use it
+* If a way to generate the ID is provided, use it; also, provide a default generator (UUIDv4)
```
LukeMathWalker, line 94: This is a bit surprising to me - why would we want to generate a client request id if none was provided?

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,132 @@
...
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = S::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: http::Request<R>) -> Self::Future {
+        request.extensions_mut().insert(ServiceRequestId::new());
```
LukeMathWalker, line 86:
```suggestion
        request.extensions_mut().insert(ServerRequestId::new());
```

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,132 @@
...
+Common formats for RequestIDs are:
+
+* UUID: a random string, represented in hex, of 128 bits from IETF RFC 4122: `7c038a43-e499-4162-8e70-2d4d38595930`
+* The hash of a sequence such as `date+thread+server`: `734678902ea938783a7200d7b2c0b487`
+* A verbose description: `current_ms+hostname+increasing_id`
+
+For privacy reasons, any format that provides service details should be avoided. A random string is preferred.
+The proposed format is to use UUID, version 4.
+
+AWS Lambda sends a RequestID in the request context; there is no need to modify it.
```
LukeMathWalker, line 69: What do we mean by this?
Do we consider the Lambda-provided request id to be a server or a client request id? If the former, how does this affect the implementation of the `ServerRequestIdProvider` below?

design/src/rfcs/rfc0024_request_id.md
82marbag: Yes: this is a feature we will add with IntoParts (for adding headers to responses)

design/src/rfcs/rfc0023_request_id.md
82marbag: ([source](https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format#page-11))

> UUIDv8's uniqueness will be implementation-specific and SHOULD NOT be assumed. [...] To be clear: UUIDv8 is not a replacement for UUIDv4 where all 122 extra bits are filled with random data.


([source](https://docs.rs/uuid/latest/uuid/)) 

> If you just want to generate unique identifiers then consider version 4 (v4) UUIDs. If you want to use UUIDs as database keys or need to sort them then consider version 7 (v7) UUIDs. Other versions should generally be avoided unless there’s an existing need for them.

We might not need this at the moment

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,136 @@
...
+1. Implementing `FromParts` for `Extension<RequestId>` gives customers the ability to write their handlers:
+
+```rust
+pub async fn handler(
+    input: input::Input,
+    request_id: Extension<RequestId>,
+) -> ...
+```
+
+`RequestId` will be injected into the extensions by a layer.
```
LukeMathWalker, line 55: It should probably be `ServerRequestId` instead of `RequestId` in all mentions in this block.

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,136 @@
...
+This layer can also be used to open a span that will log the request ID: subsequent logs will be in the scope of that span.
+
+2. RequestId format:
```
LukeMathWalker, line 58:
```suggestion
2. ServerRequestId format:
```

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,136 @@
...
+* The client could repeatedly send the same RequestID
+* The client could send no RequestID
+* The client could send a malformed or malicious RequestID
+
+To minimise the attack surface and provide a uniform experience to customers, servers should generate the value.
+However, services should be free to read the ID sent by clients in HTTP headers: it is common for services to
+read the request ID a client sends, record it and send it back upon success. A client may want to send the same value to multiple services.
+Services should still decide to have their own unique request ID per actual call.
+
+RequestIDs are not to be used by multiple services, but only within a service.
```
david-perez, line 37:
```suggestion
RequestIDs are not to be used by multiple services, but only within a single service.
```

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,136 @@
...
+        }
+        req.extensions_mut().insert(None);
+        self.inner.call(req)
+    }
+}
+```
+
+The string representation of a generated ID will be valid for this regex:
+* For `ServerRequestId`: `/^[A-Za-z0-9_-]{,48}$/`
+* For `ClientRequestId`: `/^[ -~]$/` (printable ASCII characters)
```
david-perez, line 126: It's more complicated. You can just link to [the spec](https://httpwg.org/specs/rfc9110.html#rfc.section.5.5).

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,136 @@
...
+RequestID is an element that uniquely identifies a client request. RequestID is used by services to map all logs, events and
+specific data to a single operation. This RFC discusses whether and how smithy-rs can make that value available to customers.
+
+Services use a RequestID to collect logs related to the same request and see its flow through the various operations,
+help clients debug requests by sharing this value and, in some cases, use this value to perform their business logic. RequestID is unique across a service at least within a certain timeframe.
+
+This value for the purposes above must be set by the service. Clients could send any value not conforming to the valid, chosen format,
+avoid sending one altogether or exploit how some service is using that value
+(like in [1](https://en.wikipedia.org/wiki/Shellshock_(software_bug)) and
+[2](https://cwiki.apache.org/confluence/display/WW/S2-045)).
```
jjant, line 25: The sentence about the client sending the Request Id seems to be pretty much the same as the list directly below?

82marbag:
> We'll need to provide different extractor implementations. A lambda fronted by API Gateway will have the request id [here](https://docs.rs/aws_lambda_events/latest/x86_64-apple-darwin/aws_lambda_events/apigw/struct.ApiGatewayProxyRequestContext.html#structfield.request_id).
Alternatively (and my preference), I would make our {Client,Server}RequestId s only work in non-serverless environments; I'm unsure how we could make them error out if we're in a Lambda.
In Lambda, we should provide an extractor for the entire Context object, which hosts the request id.

I think they can both still work in non-serverless environments; `ServerRequestId` will be a generated token and `ClientRequestId` will be None. The extractor for Lambda could be separate and used when the service knows it's in Lambda

design/src/rfcs/rfc0024_request_id.md
82marbag: The proposal here is to constrain the values, not to allow anything that could be sent over the wire

david-perez:
`ClientRequestId` and `ServerRequestId` will work in non-serverless environments fine.

In an AWS Lambda:
* `ClientRequestId` will continue working the same way (extracting the request id from the header, if available).
* I'm leaning towards _not_ offering a `ServerRequestId` extractor, but rather extractors for the different `Context` objects, depending on what API Gateway type fronts the Lambda (like [ApiGatewayProxyRequestContext](https://docs.rs/aws_lambda_events/latest/x86_64-apple-darwin/aws_lambda_events/apigw/struct.ApiGatewayProxyRequestContext.html)).

82marbag:
>      I'm leaning towards _not_ offering a `ServerRequestId` extractor, but rather extractors for the different `Context` objects, depending on what API Gateway type fronts the Lambda (like [ApiGatewayProxyRequestContext](https://docs.rs/aws_lambda_events/latest/x86_64-apple-darwin/aws_lambda_events/apigw/struct.ApiGatewayProxyRequestContext.html)).

This may mean we are limited by whatever format Lambda uses; if they change it, we automatically change it, and a service can't migrate from EC2 to Lambda without possibly incurring in side effects

82marbag:
Agreed to keep separate Lambda context and the proposal in this RFC

design/src/rfcs/rfc0024_request_id.md
david-perez: But the `ClientRequestId` is provided by the client. They can send anything over the wire. The server shouldn't reject the request just because it doesn't comply with this regex.

david-perez (approved):

LukeMathWalker (approved):

design/src/rfcs/rfc0024_request_id.md
drganjoo: I am **not in favor** of doing this. 1/ The other day we discussed the Postel's law "Be conservative in what you send; be liberal in what you accept.". The specs do not require us to send this header. 2/ Up to release 0.51, smithy-rs was a pure http / smithy compliant server, however always sending back a custom header breaks that compliance 3/ How would a service owner disable this feature as they might not want this header to be sent? 4/ If a client wishes to trace their requests on the server, it should be their responsibility to send a proper `x-amz-request-id`. 5/ I think it would be better to have this as a middleware layer which service owners can opt into, and they should be able to specify the header they want sent.

design/src/rfcs/rfc0024_request_id.md
```diff
@@ -0,0 +1,136 @@
...
+RequestIDs are not to be used by multiple services, but only within a service.
+
+<!-- Explain how users will use this new feature and, if necessary, how this compares to the current user experience -->
+The user experience if this RFC is implemented
+----------------------------------------------
+
+The proposal is to implement a `RequestId` type and make it available to middleware and business logic handlers, through [FromParts](../server/from-parts.md) and as a `Service`.
+To aid customers already relying on clients' request IDs, there will be two types: `ClientRequestId` and `ServerRequestId`.
+
+1. Implementing `FromParts` for `Extension<RequestId>` gives customers the ability to write their handlers:
```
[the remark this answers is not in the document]
drganjoo: > Implementing `FromParts` for `Extension<RequestId>` gives customers the ability to write their handlers:

As a service owner would I have the ability to choose if I want the `ClientRequestId` or `ServerRequestId` to be passed to my handlers? or would I have some way of accessing the `ClientRequestId`? I am asking this because as a service owner I might have provided my own AWS SDK Client, which might be generating the correct IDs and I might be interested in those IDs rather than the server generated ones.

design/src/rfcs/rfc0024_request_id.md
drganjoo: Sorry I misunderstood that it will always be sent. Since service owners can decide if they want to opt-in or not, this is perfectly fine to go ahead and do this then.

PR #1945, opened 2022-11-04
Rework enum for forward-compatibility

### LINKED ISSUE #627: rework enums to ensure that equality remains valid
We should make it so that the `Unknown` variant itself cannot be directly instantiated. This ensures that all paths go through `from_str` so that we never end up with the case of a `unknown` value that matches an existing variant.

The easiest way to do this is probably add `#[non_exhaustive]` on the Unknown variant + docs.

ysaito1001:
## Motivation and Context
Fixes #627 

## Description
This PR updates the way we code-generate `enum`s, and it aims for removing a foot-gun that might break users' code in a subtle way when they upgrade SDK from one version to another. It is [discussed](https://github.com/awslabs/smithy-rs/issues/627#issuecomment-995923092) in the issue above, and the PR implements the proposed approach. Furthermore, to ensure that the users are aware of our forward-compatible guideline in writing an `match` expression against an `enum`, we also generate [`rustdoc`](https://github.com/awslabs/smithy-rs/blob/38d371b7ded9fab5540bd54acb3c1143a656112a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt#L251-L306) along side with it.

## Testing
- Added a new unit test [`matching on enum should be forward-compatible`](https://github.com/awslabs/smithy-rs/blob/38d371b7ded9fab5540bd54acb3c1143a656112a/codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGeneratorTest.kt#L379-L428) in `EnumGeneratorTest`.
- Ran all tests in CI.
 
We needed to update the inner class `EnumGeneratorTests` to use `TestWorkspace` instead of a plain `RustWriter` to ensure it rendered what was written in `RuntimeType.forInlineFcn` during testing.

## Checklist


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolMetadataProvider.kt
```diff
@@ -120,9 +123,10 @@ class BaseSymbolMetadataProvider(
     override fun enumMeta(stringShape: StringShape): RustMetadata {
         return containerDefault.withDerives(
             RuntimeType.std.member("hash::Hash"),
-        ).withDerives( // enums can be eq because they can only contain strings
+        ).withDerives(
+            // enums can be eq because the inner data also implements Eq
```
rcoh, line 127: probably more useful to say "ints and strings" — they can be eq because they can never contain non-Eq data because of the smithy spec

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -108,6 +111,10 @@ open class EnumGenerator(
             // pub enum Blah { V1, V2, .. }
             renderEnum()
             writer.insertTrailingNewline()
+            if (target == CodegenTarget.CLIENT) {
+                renderUnknownVariantValue()
+            }
```
rcoh, line 116: we should probably add codegenContext.clientOnly { /* ... lambda ... */ } as a helper

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -198,14 +209,28 @@ open class EnumGenerator(
         }
     }
 
+    private fun renderUnknownVariantValue() {
+        // No doc or note comes with this inner opaque struct.
+        // We do want to mark it as #[allow(missing_docs)] to suppress the missing docs lint.
+        writer.docWithNote(null, null)
```
rcoh, line 215: remove prbably

rcoh: we probably want docs right? explaining that it shouldn't be used?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -225,3 +250,61 @@ open class EnumGenerator(
...
+        version of SDK where the variant `$enumName::NewFeature` is defined.
+        Specifically, when `${enumName.lowercase()}` represents `NewFeature`,
+        the execution path will hit the second last match arm as before by virtue of
+        calling `as_str` on `$enumName::NewFeature` also yielding `"NewFeature"`.
+        """.trimIndent(),
+    )
+    docs("")
+    docs(
+        """
+        It is worth pointing out that explicitly matching on the `$unknownVariant` variant should
```
rcoh, line 304:
```suggestion
        Explicitly matching on the `$unknownVariant` variant should
```

rcoh (commented):
Looks good so far!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
ysaito1001: Yes we want docs, but it's more for the whole enum as opposed to just the `Unknown` variant. What the code is really trying to achieve here is to add `#[allow(missing_docs)]` for the `Unknown` variant, but calling `writer.docWithNote(null, null)` might have been roundabout. Replaced by `rust("##[allow(missing_docs)]")` in 99a645f.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
ysaito1001: Thanks for the suggestion! Addressed in 83a15e0.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolMetadataProvider.kt
ysaito1001: Good point, addressed in 2055e04.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -198,14 +206,28 @@ open class EnumGenerator(
         }
     }
 
+    private fun unknownVariantValue(): RuntimeType {
+        return RuntimeType.forInlineFun(UnknownVariantValue, RustModule.Model) {
+            rust("##[allow(missing_docs)]")
```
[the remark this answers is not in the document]
ysaito1001: Good point. I tried removing `pub` from `pub struct UnknownVariantValue(String);` (so that we don't have to add `#[allow(missing_docs)]`) but got
```
error[E0446]: private type `UnknownVariantValue` in public interface
    |
    |     Unknown(crate::model::UnknownVariantValue),
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private type
```
It is therefore marked as `pub`. I wonder if we should add some docs instead of `#[allow(missing_docs)]`?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -198,14 +206,28 @@ open class EnumGenerator(
         }
     }
 
+    private fun unknownVariantValue(): RuntimeType {
+        return RuntimeType.forInlineFun(UnknownVariantValue, RustModule.Model) {
```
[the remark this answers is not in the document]
ysaito1001: Thanks for catching! We definitely want to avoid that. I'll think about a way around it, but let's discuss.

CHANGELOG.next.toml
```diff
@@ -92,3 +92,8 @@ references = ["smithy-rs#1935"]
 meta = { "breaking" = true, "tada" = false, "bug" = false }
 author = "jdisanti"
 
+[[smithy-rs]]
```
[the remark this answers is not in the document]
ysaito1001: Got it. I'll include an example.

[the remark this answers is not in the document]
ysaito1001: Thanks for the suggestion! I'll include an example when adding `[[aws-sdk-rust]]`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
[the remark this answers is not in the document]
ysaito1001: It is now rendered in `types.rs` as of 6fd1c57.

CHANGELOG.next.toml
```diff
@@ -103,3 +103,36 @@ message = "Upgrade to Smithy 1.26.2"
 references = ["smithy-rs#1972"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "rcoh"
+
+[[smithy-rs]]
+message = "Generate enums that guide the users to write match expressions in a forward-compatible way."
```
jdisanti, line 108: The details in the `aws-sdk-rust` entry should also be in the `smithy-rs` entry since there are people using smithy-rs to generate clients for their own services.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustModule.kt
```diff
@@ -29,6 +29,10 @@ data class RustModule(val name: String, val rustMetadata: RustMetadata, val docu
         val Model = public("model", documentation = "Data structures used by operation inputs/outputs.")
         val Input = public("input", documentation = "Input structures for operations.")
         val Output = public("output", documentation = "Output structures for operations.")
+        val Types = public(
+            "types",
+            documentation = "Re-exported types from supporting crates and synthesized types from the service.",
```
jdisanti, line 34: We may want to workshop this a bit, but I don't have any good ideas off the top of my head.

jdisanti (approved):
Looks good! I just have a couple documentation comments.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustModule.kt
ysaito1001: Absolutely. I'm still struggling and the second part after "and" is far from spot on.

CHANGELOG.next.toml
ysaito1001: That makes sense. Duplicated the `message` into the `[[smithy-rs]]` section in 04e1a22.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustModule.kt
ysaito1001: In this PR, we will keep the name of a module `types` and have added the description `Data primitives referenced by other data types`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CodegenTarget.kt
```diff
@@ -11,3 +11,21 @@ package software.amazon.smithy.rust.codegen.core.smithy
 enum class CodegenTarget {
     CLIENT, SERVER
 }
+
+/**
+ * Convenience extension to execute thunk if the target is for CodegenTarget.CLIENT
+ */
+fun <B> CodegenTarget.ifClient(thunk: () -> B): B? = if (this == CodegenTarget.CLIENT) {
```
jdisanti, line 18: Nice! These are useful.

Nit: These can be implemented directly on the enum class, and then they won't need to be separately imported where they're used. IIRC, the syntax is:

```kotlin
enum class SomeEnum {
    SOMETHING, ANOTHER_THING;

    fun something() { /* ... */ }
}
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -198,14 +208,36 @@ open class EnumGenerator(
...
     }
 
+    private fun unknownVariantValue(): RuntimeType {
+        return RuntimeType.forInlineFun(UnknownVariantValue, RustModule.Types) {
+            docs(
+                """
+                Opaque struct used as inner data for the `Unknown` variant defined in enums in
+                the crate
+
+                While this is not used by users directly, it is marked as `pub` because it is
```
jdisanti, line 218:
```suggestion
                While this is not intended to be used directly, it is marked as `pub` because it is
```

jdisanti (approved):
Looks good! Just two minor things.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CodegenTarget.kt
ysaito1001: Thanks for the suggestion. Addressed in 7b719e1.

PR #1946, opened 2022-11-04
Rename rfc0022_refine_builder.md to rfc0023_refine_builder.md

LukeMathWalker:


hlbarber (approved):
https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/overview.md
and
https://github.com/awslabs/smithy-rs/blob/main/design/src/SUMMARY.md
need updating too I think

LukeMathWalker:
Good catch - I added it.

PR #1949, opened 2022-11-04
Use `escape` on the body in `@httpRequestTest` protocol test generation

hlbarber:
## Motivation and Context

If a `@httpRequestTest` has a body assertion containing a `#` character it will cause the `rustTemplate` to break due to a collision with templating syntax.

## Description

Add `escape` to the body bytes in `@httpRequestTest` generation.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -532,9 +532,11 @@ class ServerProtocolTestGenerator(
                 // corresponding Unicode code point. That is the "form feed" 0x0c character. When printing it,
                 // it gets written as "\f", which is an invalid Rust escape sequence: https://static.rust-lang.org/doc/master/reference.html#literals
                 // So we need to write the corresponding Rust Unicode escape sequence to make the program compile.
-                "#{SmithyHttpServer}::body::Body::from(#{Bytes}::from_static(${
-                body.replace("\u000c", "\\u{000c}").dq()
-                }.as_bytes()))"
+                //
+                // We also replace `#` with `##` to avoid interactions with templating in the case where the body contains `#`.
+                val sanitizedBody = body.replace("\u000c", "\\u{000c}").replace("#", "##").dq()
```
rcoh, line 537: use escape

hlbarber: https://github.com/awslabs/smithy-rs/pull/1949/commits/300eaf4c1ffa3ef816853a33a33b05663affcad5

david-perez (approved):

Issue #1957, opened 2022-11-07
Add S3 customization to prevent accidental data deletion for `AbortMultipartUpload` requests with no ID


Velfi:
The `AbortMultiPartUpload` API is ambiguous and will delete an object if no ID is passed, rather than returning an error. Add a customization that will fail the request when no ID is set.

rcoh:
we currently can do this for `{httpLabel}`: https://github.com/awslabs/smithy-rs/blob/b4f294c2770c58148c43411803330c26e369ec14/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/RequestBindingGenerator.kt#L246

I'd suggest adding a private trait `Mandatory` or something and then using the same technique to fail the request at serialization time

jdisanti:
The fix for this was included in [release-2022-12-14](https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2022-12-14).

PR #1973, opened 2022-11-09
Always write required query param keys, even if value is unset or empty

Velfi:
## Motivation and Context

#1957 

## Description

This fixes a bug where sending an `AbortMultipartUpload` request with no upload ID would cause accidental data loss. The fix works by adding a new trait `Mandatory` to the `UploadId` shape. The presence of this trait causes the `RequestBindingGenerator` to add code branches that guard against a missing ID.

## Testing

I wrote two tests

## Checklist


rcoh (commented):
this is great! super clean approach. My only question is whether we want to apply it to leaf shapes or member shapes. We _may_ want to double check that it actually works for both types of shapes if we don't have an assertion that the shape isn't a member shape

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
```diff
@@ -47,6 +51,10 @@ class S3Decorator : RustCodegenDecorator<ClientProtocolGenerator, ClientCodegenC
         // API returns GetObjectAttributes_Response_ instead of Output
         ShapeId.from("com.amazonaws.s3#GetObjectAttributesOutput"),
     )
+    private val mandatoryShapesList = setOf(
+        // Must be included or else S3 interprets the request as a get, put, or delete object request.
+        ShapeId.from("com.amazonaws.s3#MultipartUploadId"),
+    )
```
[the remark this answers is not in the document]
Velfi: I updated it to attach `Mandatory` to all required members named `UploadId`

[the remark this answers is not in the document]
Velfi: I spoke with Chase about this and the actual solution the SDK org decided on was to always serialize required query params, regardless of whether a value was set. I'll update the PR to take that approach.

aws/sdk/integration-tests/s3/tests/mandatory-query-param.rs
```diff
@@ -0,0 +1,45 @@
...
+    let client = aws_sdk_s3::Client::from_conf_conn(conf, conn.clone());
+
+    client
+        .abort_multipart_upload()
+        .bucket("a-bucket")
+        .key("a-key")
+        .send()
+        .await
+        .unwrap();
+}
```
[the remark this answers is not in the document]
Velfi: I updated these tests to work differently. After talking to Chase, I implemented it so that, for required query params, the key will always be written, regardless of if a value was set. This allows S3 to differentiate the request correctly and reject it for a missing/empty upload ID.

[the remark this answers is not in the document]
Velfi: The updated tests now check that the uploadId key is in the sent request

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/RequestBindingGenerator.kt
```diff
@@ -201,23 +203,43 @@ class RequestBindingGenerator(
...
-                        )
-                    }
+                    paramList(outerTarget, field, param, writer, memberShape)
+                }
+                if (memberShape.hasTrait<RequiredTrait>()) {
+                    rust(
+                        """
+                        else {
+                            query.push_kv(${param.locationName.dq()}, "");
+                        }
```
jdisanti, line 216: IIRC, the goal is to have this behavior match labels, which throw a validation error:
https://github.com/awslabs/smithy-rs/blob/7f40bbcb2c2c1ed4f8c5096490ad161b60e884a0/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/RequestBindingGenerator.kt#L250-L258

aws/sdk/integration-tests/s3/tests/mandatory-query-param.rs
```diff
@@ -0,0 +1,97 @@
...
+        .body(SdkBody::empty())
+        .unwrap()
+}
+
+fn empty_ok_response() -> http::Response<&'static str> {
+    http::Response::builder().status(200).body("").unwrap()
+}
+
+#[tokio::test]
+async fn test_mandatory_query_param_is_unset() {
```
ysaito1001, line 31: Could I suggest that the name of the test (and the other test as well) include `upload_id`, i.e.
`test_mandatory_query_param_upload_id_is_unset` or something to that effect so that the reader knows which mandatory query param we're talking about, unless these test functions are designed to be generic and may contain not just `upload_id` in the future?

ysaito1001, line 31: Could I suggest that the name of the test (and the other test as well) include `upload_id`, i.e.
`test_mandatory_query_param_upload_id_is_unset` or something to that effect so that the reader knows which mandatory query param we're talking about, unless these test functions are designed to be generic and may contain not just `upload_id` in the future?

ysaito1001 (commented):
Learning from the changes and they look great! Left a minor, opinionated suggestion for the names of the unit tests.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/RequestBindingGenerator.kt
```diff
@@ -201,23 +203,43 @@ class RequestBindingGenerator(
...
-                        val target = model.expectShape(targetId)
-                        rust(
-                            "query.push_kv(${param.locationName.dq()}, ${
-                            paramFmtFun(writer, target, memberShape, innerField)
-                            });",
-                        )
-                    }
+                    paramList(outerTarget, field, param, writer, memberShape)
+                }
+                if (memberShape.hasTrait<RequiredTrait>()) {
```
david-perez, line 211:
```suggestion
                if (memberShape.isRequired) {
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/RequestBindingGenerator.kt
```diff
@@ -201,23 +203,43 @@ class RequestBindingGenerator(
...
+                    rust(
+                        """
+                        else {
+                            query.push_kv(${param.locationName.dq()}, "");
```
david-perez, line 215:
```suggestion
                            query.push_kv("${param.locationName}", "");

```

aws/sdk/integration-tests/s3/tests/mandatory-query-param.rs
Velfi: We changed this to require any query param marked as required so I'm keeping it generic. However, I did update the names of the tests to be more clear.

CHANGELOG.next.toml
```diff
@@ -197,6 +197,12 @@ references = ["smithy-rs#1935"]
 meta = { "breaking" = true, "tada" = false, "bug" = false }
 author = "jdisanti"
 
+[[aws-sdk-rust]]
```
jdisanti, line 200: Would be good to add a broader `[[smithy-rs]]` entry for the overall required query param handling change as well.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/RequestBindingGenerator.kt
```diff
@@ -208,24 +212,64 @@ class RequestBindingGenerator(
...
+                        rustTemplate(
+                            "let $derefName = $derefName.as_ref().ok_or_else(|| #{BuildError:W})?;",
+                            *codegenScope,
                         )
                     }
+
+                    when {
+                        target.isStringShape -> {
+                            val asStr = writable {
+                                if (target.hasTrait<EnumTrait>()) {
```
jdisanti, line 236: Seems like an `Option::is_none` check would be sufficient for enums, rather than converting it to a string. Can an enum member have an empty value? That could get interesting...

aws/sdk/integration-tests/s3/tests/required-query-params.rs
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use aws_sdk_s3::operation::AbortMultipartUpload;
+use aws_sdk_s3::Region;
+use aws_smithy_http::operation::error::BuildError;
+
+#[tokio::test]
+async fn test_error_when_required_query_param_is_unset() {
```
jdisanti, line 11: I think we should also have a Kotlin unit test for this query param validation functionality, since it applies more broadly than to just S3.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/RequestBindingGenerator.kt
Velfi: I checked and found that enums always have values in smithy IDL v1 and v2 so I'll implement your suggestion.

ysaito1001 (approved):
Looks great!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/RequestBindingGenerator.kt
```diff
@@ -71,7 +73,7 @@ class RequestBindingGenerator(
     private val httpBindingGenerator =
         HttpBindingGenerator(protocol, codegenContext, codegenContext.symbolProvider, operationShape, ::builderSymbol)
     private val index = HttpBindingIndex.of(model)
-    private val Encoder = CargoDependency.smithyTypes(runtimeConfig).toType().member("primitive::Encoder")
+    private val encoder = CargoDependency.smithyTypes(runtimeConfig).asType().member("primitive::Encoder")
```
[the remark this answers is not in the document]
Velfi: oops, yeah. Thanks for spotting that.

Issue #1926, opened 2022-10-31
Overhaul errors in rust-runtime crates


jdisanti:
- [x] Add `DisplayErrorContext` to all `tracing` error logs
- [x] Hide error source type in `Unhandled` variant in code generated errors
- [x] Remove `Clone` from `ProfileParseError` and any others that have it
- [x] Errors in rust-runtime
  - [x] `aws-smithy-async`
    - [x] `crate::future::TimedOutError`
  - [x] `aws-smithy-checksums`
    - [x] `crate::body::Error`
  - [x] `aws-smithy-client`
    - [x] `crate::hyper_ext::HttpTimeoutError`
    - [x] `crate::timeout::RequestTimeoutError`
  - [x] `aws-smithy-eventstream`
    - [x] `crate::error::Error`
  - [x] `aws-smithy-http`
    - [x] `crate::byte_stream::Error`
    - [x] `crate::endpoint::Error`
    - [x] `crate::endpoint::InvalidEndpoint`
    - [x] `crate::header::ParseError`
    - [x] `crate::operation::BuildError`
    - [x] `crate::operation::SerializationError`
    - [x] `crate::result::ConnectorError`
    - [x] `crate::result::SdkError`
    - [x] `crate::event_stream::receiver::Error`
    - [x] `crate::event_stream::sender::MessageStreamError`
  - [x] `aws-smithy-json`
    - [x] `crate::escape::EscapeError`
    - [x] `crate::deserialize::Error`
  - [x] `aws-smithy-types`
    - [x] `crate::base64::DecodeError`
    - [x] `crate::error::Error`
    - [x] `crate::TryFromNumberError`
    - [x] `crate::primitive::PrimitiveParseError`
    - [x] `crate::retry::RetryModeParseErr`
    - [x] `crate::retry::RetryConfigErr`
    - [x] `crate::timeout::error::ConfigError`
    - [x] `crate::date_time::format::DateTimeParseError`
    - [x] `crate::date_time::format::DateTimeFormatError`
    - [x] `crate::date_time::ConversionError`
  - [x] `aws-smithy-types-convert`
    - [x] `crate::date_time::Error`
  - [x] `aws-smithy-xml`
    - [x] `crate::decode::XmlError`
    - [x] `crate::encode::Error`
  - [x] `inlineable`
    - [x] `crate::endpoint_lib::InvalidArn`
- [x] Errors in aws-runtime
  - [x] `aws-types`
[...]

PR #1974, opened 2022-11-09
Make `SdkError::into_service_error` infallible

jdisanti:
## Motivation and Context
Originally, `into_service_error` returned a `Result<E, Self>`. The intent was that the `ServiceError` could easily be extracted while bubbling up other error types, but as it turns out, this makes it harder to bubble up service errors that don't need to be handled in a match block. Making `into_service_error` makes both use cases easy.

Credit goes to @rcoh for the suggestion.

This PR is part of a series that will fully implement the RFC, tracked in #1926.

Changelog entries added in #1951.

### DIFF
--- a/rust-runtime/aws-smithy-http/src/result.rs
+++ b/rust-runtime/aws-smithy-http/src/result.rs
@@ -122,6 +122,14 @@ impl<E, R> ServiceError<E, R> {
     }
 }
 
+/// Constructs the unhandled variant of a code generated error.
+///
+/// This trait exists so that [`SdkError::into_service_error`] can be infallible.
+pub trait CreateUnhandledError {
+    /// Creates an unhandled error variant with the given `source`.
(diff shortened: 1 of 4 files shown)


rust-runtime/aws-smithy-http/src/result.rs
```diff
@@ -122,6 +122,14 @@ impl<E, R> ServiceError<E, R> {
     }
 }
 
+/// Constructs the unhandled variant of a code generated error.
+///
+/// This trait exists so that [`SdkError::into_service_error`] can be infallible.
+pub trait CreateUnhandledError {
+    /// Creates an unhandled error variant with the given `source`.
+    fn create_unhandled_error(source: Box<dyn Error + Send + Sync + 'static>) -> Self;
+}
```
rcoh, line 131: I wonder if this should be from `SdkError` to enable the future possibility of more precise handling?

Velfi (approved):

PR #1978, opened 2022-11-11
Fix Gradle deprecation warnings

rschmitt:
This change provides compatibility with the latest Gradle 8.0 nightly. Since the top-level code coverage stuff was already broken and apparently no one noticed, I just deleted it. The coverage reports for specific subprojects still work fine.


jdisanti (approved):
LGTM! Thank you.

PR #1981, opened 2022-11-14
Add default for the Smithy test workspace

rcoh:
## Motivation and Context
The test codegeneration has the ability to run all tests in a persistent workspace. This makes multiple subsequent test runs _much_ faster–it can take test runs from 15-20 seconds to just 1 second.

But, prior to this change, this was only accessible in Intellij if you altered the default execution templates to set an environment variable.

## Description
This change sets a default for this variable based on your home directory.

## Testing
Ran tests in the default runtime configuration and verified that the expected directory was used

## Checklist
- internalonly, no changelog


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/Rust.kt
```diff
@@ -55,8 +56,14 @@ private fun tempDir(directory: File? = null): File {
  * This workspace significantly improves test performance by sharing dependencies between different tests.
  */
 object TestWorkspace {
-    private val baseDir =
-        System.getenv("SMITHY_TEST_WORKSPACE")?.let { File(it) } ?: tempDir()
+    private val baseDir by lazy {
+        val homeDir = System.getProperty("user.home")
+        if (homeDir != null) {
+            File(Path.of(homeDir, ".smithy-test-workspace").absolutePathString())
```
david-perez, line 62: Can this be `$XDG_DATA_HOME`? Unsure if something like [`dirs::data_local_dir`](https://docs.rs/dirs/latest/dirs/fn.data_local_dir.html) exists for Kotlin. A directory under `/tmp/` might also be a good candidate for where to place this workspace. The home directory is too invasive.

david-perez (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/Rust.kt
jjant: Note that if we use `XDG_DATA_HOME`, we should default to [`$HOME/.local/share` if it's not set](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
Also, if we use this, should we remove the usage of `SMITHY_TEST_WORKSPACE`?

rcoh: good idea—Updated to match dirs::data_local_dir. I kept the fallback, might be useful for CI

Velfi (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/Rust.kt
```diff
@@ -55,8 +56,18 @@ private fun tempDir(directory: File? = null): File {
  * This workspace significantly improves test performance by sharing dependencies between different tests.
  */
 object TestWorkspace {
-    private val baseDir =
-        System.getenv("SMITHY_TEST_WORKSPACE")?.let { File(it) } ?: tempDir()
+    private val baseDir by lazy {
+        val homeDir = System.getProperty("APPDATA")
+            ?: System.getenv("XDG_DATA_HOME")
+            ?: System.getProperty("user.home")
+                ?.let { Path.of(it, "Library/Application Support").absolutePathString() }
```
jjant, line 63: Shouldn't this be `Path.of(it, ".local/share").absolutePathString()`?

david-perez: Yep, `Library/Application Support` is for Mac I presume.

jjant: AFAIK programs following the XDG spec use `~/.local/share` on Mac too (at least on my machine, things like `neovim`, `nix`, and `ruby` do).

PR #1983, opened 2022-11-14
Respect the sensitive trait on container shapes

### LINKED ISSUE #1745: `std::fmt::Debug` implementation on `enum`s does not honor the `sensitive` trait
This model:

```smithy
@sensitive
@enum([
    { value: "en" },
    { value: "es" },
])
string Enum
```

generates:

```rust
#[derive(
    std::clone::Clone,
    std::cmp::Eq,
    std::cmp::Ord,
    std::cmp::PartialEq,
    std::cmp::PartialOrd,
    std::fmt::Debug,
    std::hash::Hash,
)]
pub struct Enum(String);
impl Enum {
    /// Returns the `&str` value of the enum member.
    pub fn as_str(&self) -> &str {
        &self.0
    }
    /// Returns all the `&str` representations of the enum members.
    pub fn values() -> &'static [&'static str] {
        &["en", "es"]
    }
}
impl<T> std::convert::From<T> for Enum
where
    T: std::convert::AsRef<str>,
{
    fn from(s: T) -> Self {
        Enum(s.as_ref().to_owned())
    }
}
```

Note how we `derive(Debug)`, so the [`sensitive`](https://awslabs.github.io/smithy/2.0/spec/documentation-traits.html#sensitive-trait) trait is not being honored.

We derive `Debug` because of the `SymbolMetadataProvider`:

https://github.com/awslabs/smithy-rs/blob/ffafbec885c1e83a460fed38fc98119bb63d0439/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/SymbolMetadataProvider.kt#L134-L134

which gets rendered here in the case of enums:

https://github.com/awslabs/smithy-rs/blob/ffafbec885c1e83a460fed38fc98119bb63d0439/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/EnumGenerator.kt#L128-L128

---

As an aside, it's not clear to me whether we want `Debug` to honor `sensitive`. The spec for `sensitive` reads:

> Sensitive data MUST NOT be exposed in things like exception messages or log output. Application of this trait SHOULD NOT affect wire logging (i.e., logging of all data transmitted to and from servers or clients).

Although the `Display` implementation is used by default in the logging frameworks I know of, `Debug` can be used in `DEBUG`-level logging e.g. via `{#?}`, so to be on the safe side I would advocate for making `Debug` honor it. However, [according to the standard library](https://doc.rust-lang.org/std/fmt/trait.Debug.html), `Debug` is meant for programmer-facing output in a _debugging_ context: there's an argument to be made that if the Smithy spec calls out that `sensitive` should not affect wire logging (which should generally only be used in a debugging context), then `Debug` should not honor `sensitive`.

In any case, whatever we decide, we're currently being inconsistent, because our implementation of `Debug` for `structure` shapes _does_ honor `sensitive`:

https://github.com/awslabs/smithy-rs/blob/ffafbec885c1e83a460fed38fc98119bb63d0439/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/StructureGenerator.kt#L108-L127

ysaito1001:
## Motivation and Context
Fixes #1745.

## Description
**Updated**
Originally it only addressed the sensitive trait on enums, but per David's [suggestion](https://github.com/awslabs/smithy-rs/pull/1983#discussion_r1022613055), this PR has been expanded to include other container shapes such as `Union` and `Builder`s (#2029 got merged into this).

**Old**
This PR updates how `std::fmt::Debug` trait is implemented on enums. If an enum is marked as as `sensitive`, debug printing it only shows the redacted text. If the enum is not marked as `sensitive`, it uses `#[derive(Debug)]` as before, hence displaying its value. 

## Testing
- [x] Added new unit tests in `EnumGeneratorTest.kt` to verify the said behavior of `std::fmt::Debug` for a named enum and a unnamed enum.
- [x] Passed tests in CI.

## Checklist


jdisanti (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -230,6 +241,19 @@ open class EnumGenerator(
...
+     *
+     * It prints the redacted text regardless of the variant it is asked to print.
+     */
+    private fun renderDebugImplForSensitiveEnum() {
+        writer.rustBlock("impl #T for $enumName", RuntimeType.Debug) {
+            writer.rustBlock("fn fmt(&self, f: &mut #1T::Formatter<'_>) -> #1T::Result", RuntimeType.stdfmt) {
+                rust("""write!(f, "{}", $REDACTION)""")
+            }
+        }
+    }
```
david-perez, line 255: You can do this in a single uninterrupted call. Note that you don't need a format argument in the format string, you can use `$REDACTION` as-is.

```suggestion
        writer.rustTemplate(
            """
            impl #{Debug} for $enumName {
                fn fmt(&self, f: &mut #{StdFmt}::Formatter<'_>) -> #{StdFmt}::Result {
                    write!(f, $REDACTION)
                }
            }
            """,
            "Debug" to RuntimeType.Debug,
            "StdFmt" to RuntimeType.stdfmt,
        )
```

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGeneratorTest.kt
```diff
@@ -426,4 +427,133 @@ class EnumGeneratorTest {
...
+            val generator = EnumGenerator(model, provider, this, shape, trait)
+            generator.render()
+            unitTest(
+                "impl_debug_for_non_sensitive_enum_should_implement_the_derived_debug_trait",
+                """
+                assert_eq!(format!("{:?}", SomeEnum::Foo), "Foo");
+                assert_eq!(format!("{:?}", SomeEnum::Bar), "Bar");
+                assert_eq!(
+                    format!("{:?}", SomeEnum::from("Baz")),
+                    "Unknown(UnknownVariantValue(\"Baz\"))");
```
david-perez, line 456: Try to adhere to rustfmt guidelines:

```suggestion
                assert_eq!(
                    format!("{:?}", SomeEnum::from("Baz")),
                    "Unknown(UnknownVariantValue(\"Baz\"))"
                );
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -230,6 +241,19 @@ open class EnumGenerator(
         }
     }
 
+    /**
+     * Manually implement the Debug trait for the enum if marked as sensitive
```
david-perez, line 245:
```suggestion
     * Manually implement the `Debug` trait for the enum if marked as sensitive.
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -92,7 +94,12 @@ open class EnumGenerator(
 ) {
     protected val symbol: Symbol = symbolProvider.toSymbol(shape)
     protected val enumName: String = symbol.name
-    protected val meta = symbol.expectRustMetadata()
+    private val isSensitive = shape.shouldRedact(model)
+    protected val meta = if (isSensitive) {
+        symbol.expectRustMetadata().withoutDerives(RuntimeType.Debug)
```
david-perez, line 99: I think this should be done centrally in the `SymbolMetadataProvider`, where we can do the change for all applicable shapes: if they have the `sensitive` trait, remove `Debug` from the list of derived traits.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
ysaito1001: I like how you use `Debug` and `StdFmt` for clarify. Thank you for the suggestion!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
ysaito1001: That's a good point. Should we make the change in question incrementally and opportunistically, depending on how many applicable shapes there are? Once we cover enough shapes, letting them handle locally like above, maybe we can promote the logic to a centralized place? I wonder if that'll help keep the scope of this PR focused.

ysaito1001: I've understood the proposal better now. I'm going to create a new branch out of this, make the proposed change to `BaseSymbolMetadataProvider.[structureMeta | unionMeta | enumMeta]`, remove the above code from `EnumGenerator.kt` and `StructureGenerator.kt`, and run the tests in CI to see how many will break.

ysaito1001: Created #2029

PR #1984, opened 2022-11-15
Improve `Endpoint` panic-safety and ergonomics

### LINKED ISSUE #1496: Return errors instead of panicking for Endpoint::set_endpoint
We are happy users of the rust aws_sdk but just recently found out that its underlying libraries use `.expect("...")` in several places and thus will panic under some conditions. One such condition would be e.g. setting a custom endpoint with a slightly invalid URL: https://github.com/awslabs/smithy-rs/blob/e1e9a29d5db1330eb51c1e84e5c6c3acb9b4f65e/rust-runtime/aws-smithy-http/src/endpoint.rs#L84-L105

There are many more places where smithy code could panic. Some of those might be legit as they might never be triggered, but some of those are actually dangerous.

* https://github.com/awslabs/smithy-rs/blob/e1e9a29d5db1330eb51c1e84e5c6c3acb9b4f65e/aws/rust-runtime/aws-inlineable/src/glacier_checksums.rs#L55
* https://github.com/awslabs/smithy-rs/blob/e1e9a29d5db1330eb51c1e84e5c6c3acb9b4f65e/rust-runtime/aws-smithy-xml/src/escape.rs#L23
* https://github.com/awslabs/smithy-rs/blob/e1e9a29d5db1330eb51c1e84e5c6c3acb9b4f65e/rust-runtime/aws-smithy-client/src/dvr/replay.rs#L30
* https://github.com/awslabs/smithy-rs/blob/b7506ec2e56ef1540d9fe35ec12784aa13d2e003/rust-runtime/aws-smithy-client/src/test_connection.rs#L44

In our codebase we take great care to avoid panicking constructs and we would love to have the rust aws_sdk do the same. We are ready to bite the bullet of handling additional `Result<..., ...>` return types in sdk code.

jdisanti:
## Motivation and Context
This PR fixes #1496:
- `Endpoint::set_endpoint` no longer panics when called on an endpoint without a scheme
- `Endpoint::mutable` and `Endpoint::immutable` now both return a result so that constructing an endpoint without a scheme is an error
- `Endpoint::mutable` and `Endpoint::immutable` both now take a string instead of a `Uri` as a convenience
- `Endpoint::mutable_uri` and `Endpoint::immutable_uri` were added to construct an endpoint directly from a `Uri`

## Checklist


CHANGELOG.next.toml
```diff
@@ -339,3 +339,21 @@ Server SDKs now correctly reject operation inputs that don't set values for `req
 references = ["smithy-rs#1714", "smithy-rs#1342", "smithy-rs#1860"]
 meta = { "breaking" = true, "tada" = false, "bug" = true, "target" = "server" }
 author = "david-perez"
+
+[[aws-sdk-rust]]
+message = "Functions on `aws_smithy_http::endpoint::Endpoint` now return a `Result` instead of panicking."
+references = ["smithy-rs#1984", "smithy-rs#1496"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
```
jjant, line 346: Does the missing `target` default to `"all"`?

jdisanti: `target` only applies to `[[smithy-rs]]` since there's no server for the SDKs.

jdisanti: Although, now that I'm taking a second look, this deserves a `[[smithy-rs]]` entry too since its in `aws_smithy_http`.

aws/rust-runtime/aws-config/src/imds/client.rs
```diff
@@ -503,9 +506,8 @@ impl EndpointSource {
                     profile.get(profile_keys::ENDPOINT).map(Cow::Borrowed)
                 };
                 if let Some(uri) = uri_override {
-                    return Ok(
-                        Uri::try_from(uri.as_ref()).map_err(BuildErrorKind::InvalidEndpointUri)?
-                    );
+                    return Ok(Uri::try_from(uri.as_ref())
+                        .map_err(|err| BuildErrorKind::InvalidEndpointUri(err.into()))?);
```
Velfi, line 510: These lines are a bit hard for me to parse.

Velfi (approved):

aws/rust-runtime/aws-config/src/imds/client.rs
jdisanti: Added methods to `BuildError` to construct these, which makes it a lot easier to grok.

PR #1985, opened 2022-11-15
Python: Allow injecting Lambda Context

unexge:
## Motivation and Context

This PR allows users to inject Lambda context to their custom context class by type-hinting `LambdaContext`:

```python
from typing import Optional
from libpokemon_service_server_sdk.aws_lambda import LambdaContext

@dataclass
class Context:
    # Inject Lambda context if service is running on Lambda
    # NOTE: All the values that will be injected by the framework should be wrapped with `Optional`
    lambda_ctx: Optional[LambdaContext] = None

    ...

app.context(Context())

@app.get_pokemon_species
def get_pokemon_species(
    input: GetPokemonSpeciesInput, context: Context
) -> GetPokemonSpeciesOutput:
    if context.lambda_ctx is not None:
        logging.debug(
            "Lambda Context: %s",
            dict(
                request_id=context.lambda_ctx.request_id,
                deadline=context.lambda_ctx.deadline,
                invoked_function_arn=context.lambda_ctx.invoked_function_arn,
                function_name=context.lambda_ctx.env_config.function_name,
                memory=context.lambda_ctx.env_config.memory,
                version=context.lambda_ctx.env_config.version,
            ),
        )
    ...

```

Closes https://github.com/awslabs/smithy-rs/issues/1815

## Description
This PR introduces `PyContext` class to wrap raw Python object provided by the users. At the startup we inspect raw Python object to detect injectable fields, after that `AddPyContextLayer` (a Tower layer), populates `PyContext` from incoming request (i.e. creates `PyLambdaContext` from `LambdaContext`) and then adds `PyContext` to request extensions, which then gets injected by the handlers and then passed to Python handler (thanks to `ToPyObject` implementation of `PyContext`).

## TODOs

- [x] ~~Rebase https://github.com/awslabs/smithy-rs/pull/2035 and make Lambda specific types feature gated.~~ We decided to remove feature flag because we don't have codegen support for feature flags.


codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerModuleGenerator.kt
```diff
@@ -163,6 +164,22 @@ class PythonServerModuleGenerator(
...
+            pyo3::py_run!(
+                py,
+                lambda_,
+                "import sys; sys.modules['$libName.lambda_'] = lambda_"
+            );
+            m.add_submodule(lambda_)?;
+            """,
+            *codegenScope,
+        )
+    }
```
crisidev, line 181: I am not sure I understand why the module is suffixed with underscore..

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
```diff
@@ -78,6 +79,9 @@ def value(self) -> int:
 #   https://docs.python.org/3/library/multiprocessing.html#sharing-state-between-processes
 @dataclass
 class Context:
+    # Inject Lambda context if service is running on Lambda
+    lambda_ctx: LambdaContext = None
```
crisidev, line 83:
```suggestion
    lambda_ctx: Optional[LambdaContext] = None
```

rust-runtime/aws-smithy-http-server-python/src/server.rs
```diff
@@ -424,19 +424,30 @@ event_loop.add_signal_handler(signal.SIGINT,
...
+                .build()
+                .expect("unable to start a new tokio runtime for this process");
+            rt.block_on(async move {
+                let handler = LambdaHandler::new(service);
+                let lambda = lambda_http::run(handler);
+                tracing::debug!("starting lambda handler");
+                if let Err(err) = lambda.await {
+                    tracing::error!(error = %err, "unable to start lambda handler");
+                }
+            });
```
crisidev, line 446: What is the reason for moving from blocking directly on the Lambda await to using a thread and blocking on the Python event loop? Is it to handle signal properly?

unexge: `asyncio` runtime doesn't work property if it does not control the main thread: https://docs.rs/pyo3-asyncio/latest/pyo3_asyncio/#pythons-event-loop-and-the-main-thread. Basically without this it runs into deadlock or something and hangs forever if we `await` on a Python coroutine. I borrowed this logic from `start_hyper_worker` btw, I didn't understand at first why that method is designed that way but now I understand 😄

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
unexge: Our check will fail if we wrap it with `Optional`: https://github.com/awslabs/smithy-rs/pull/1985/files#diff-808ab7f16108679c85b347d9b60605bb96a6fad0f1e6ce86ba679bc922fd105eR100 I'll update PR to support `Optional` types. It'd be better if we enforce users to wrap their types with `Optional` always as it indicates that there is no compile-time guarantee that this value if provided and they should always check at runtime to make sure that value is provided, what do you think?

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerModuleGenerator.kt
crisidev: Right! I propose we call this module `aws_lambda`. WDYT?

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
crisidev: Yep, I agree.

rust-runtime/aws-smithy-http-server-python/src/server.rs
crisidev: Got it, makes sense.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerModuleGenerator.kt
unexge: Yes makes sense, addressed in https://github.com/awslabs/smithy-rs/pull/1985/commits/75bd902168f572e99b6d998ca2f46bc4f94ddda4

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
unexge: Addressed in https://github.com/awslabs/smithy-rs/pull/1985/commits/d82ea4e9b8d8169f81ca31bf1444eb32e808f091

rust-runtime/aws-smithy-http-server-python/src/context.rs
```diff
@@ -0,0 +1,221 @@
...
+    match ty
+        .getattr("__args__")
+        .and_then(|args| args.extract::<(&PyAny, &PyAny)>())
+    {
+        Ok((first_ty, second_ty)) => Ok(first_ty.is(target_ty) && second_ty.is(none_ty)),
+        // Here we can ignore errors because `__args__` is not present on all types
+        // and it is not really an error, it is just a type we don't expect
+        _ => Ok(false),
+    }
+}
```
crisidev, line 137: Nit: this could be moved into the util.rs file.

crisidev (approved):
LGTM

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
```diff
@@ -78,6 +79,10 @@ def value(self) -> int:
 #   https://docs.python.org/3/library/multiprocessing.html#sharing-state-between-processes
 @dataclass
 class Context:
+    # Inject Lambda context if service is running on Lambda
+    # NOTE: All the values that will be injected by the framework should be wrapped with `Optional`
+    lambda_ctx: Optional[LambdaContext] = None
```
crisidev, line 84: Just for my understanding: if the `lambda_ctx` of type `Optional[LambdaContext]` is not defined here, the injection will be just avoided right?

rust-runtime/aws-smithy-http-server-python/src/context/testing.rs
```diff
@@ -0,0 +1,49 @@
...
+            .expect("you should assing your context class to `ctx` variable")
+            .into_py(py);
+        Ok::<_, PyErr>(context)
+    })
+    .unwrap();
+    PyContext::new(inner).unwrap()
+}
+
+pub fn lambda_ctx(req_id: &'static str, deadline_ms: &'static str) -> Context {
+    let mut headers = HeaderMap::new();
```
hlbarber, line 35: NIT: I think this kind of thing is faster when you use `HeadersMap::extend` or `FromIterator` because it's more likely to understand how many items it's going to receive and therefore will only resize the collection once. Probably doesn't make a difference when there's only two inserts though.

rust-runtime/aws-smithy-http-server-python/src/context/layer.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
+        if self.ctx.has_lambda_context_fields() {
+            let py_lambda_ctx = req
+                .extensions()
+                .get::<LambdaContext>()
+                .cloned()
```
hlbarber, line 65: Do we want to `.cloned()` this? What's the difference if we `.remove::<LambdaContext>()`?

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
```diff
@@ -195,6 +200,18 @@ def do_nothing(_: DoNothingInput) -> DoNothingOutput:
 def get_pokemon_species(
     input: GetPokemonSpeciesInput, context: Context
 ) -> GetPokemonSpeciesOutput:
+    if context.lambda_ctx is not None:
+        logging.debug(
+            "Lambda Context: %s",
+            dict(
```
hlbarber, line 206: It's a shame there's no interface on the `logging` package to do structured logging, just serializing this big dict seems wrong.

rust-runtime/aws-smithy-http-server-python/src/server.rs
hlbarber: This confused me too until I opened this conversation. Should we add a brief explanation/the link above as a comment?

hlbarber (approved):
Nice documentation and test coverage

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
crisidev: There is, but it is something the customer should provide through their own logging format (I have used a lot JSON lines in the past).

So to be as compatible as possible, we just log the entire dict.

rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py
```diff
@@ -195,6 +200,18 @@ def do_nothing(_: DoNothingInput) -> DoNothingOutput:
 def get_pokemon_species(
     input: GetPokemonSpeciesInput, context: Context
 ) -> GetPokemonSpeciesOutput:
+    if context.lambda_ctx is not None:
+        logging.debug(
+            "Lambda Context: %s",
```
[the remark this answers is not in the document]
crisidev: AFAIK there is not a consensus. I have always used capitalized logging in Python, but I don't have a reason for it.

rust-runtime/aws-smithy-http-server-python/src/context/layer.rs
unexge: I didn't want to use `remove` because other layers might want to use `LambdaContext`, is it a good practice to remove extensions?

rust-runtime/aws-smithy-http-server-python/src/util.rs
```diff
@@ -42,8 +42,45 @@ fn is_coroutine(py: Python, func: &PyObject) -> PyResult<bool> {
...
+        _ => return Ok(false),
+    };
+
+    let none = py.None();
+    // in typing, `None` is a special case and it is converted to `type(None)`,
+    // so we are getting type of `None` here to match
+    let none_ty = none.as_ref(py).get_type();
+    let target_ty = py.get_type::<T>();
+
+    // `Union` should be tuple of `(T, NoneType)`
```
chadac, line 65: Should the edge case of `None | T` be considered? The convention right now is `T | None` absolutely, but it could produce some confusing errors otherwise.

PR #1988, opened 2022-11-15
Add UNIFORM_NAME and Sealed for OperationShape

82marbag:
## Motivation and Context

* `#` is forbidden in some implementations and `.` is what is used as a replacement.
* `OperationShape` should not be implemented by customers; we can't force them not to, but a hint to avoid it is helpful

## Checklist


LukeMathWalker:
When should one use `UNIFORM_NAME` instead of `NAME`?

Issue #1546, opened 2022-07-12
Add custom behavior for Unit type so that using a union variant doesn't need to provide an empty unit


zclcharles:
Currently when specifying a union structure with a variant that is [Unit type](https://awslabs.github.io/smithy/1.0/spec/core/model.html#unit-type), generated SDK will have a tuple struct variant in an enum. 
For example
```
union TransactionStatus {
  succeed: Unit,
  failed: String,
  in_progress: Unit,
  received: Unit
}
```
will result in 
```
pub enum TransactionStatus {
    #[allow(missing_docs)] // documentation missing in model
    Failed(std::string::String),
    #[allow(missing_docs)] // documentation missing in model
    InProgress(crate::model::Unit),
    #[allow(missing_docs)] // documentation missing in model
    Received(crate::model::Unit),
    #[allow(missing_docs)] // documentation missing in model
    Succeed(crate::model::Unit),
}
```
it will be good if we can just have 
```
pub enum TransactionStatus {
    #[allow(missing_docs)] // documentation missing in model
    Failed(std::string::String),
    #[allow(missing_docs)] // documentation missing in model
    InProgress,
    #[allow(missing_docs)] // documentation missing in model
    Received,
    #[allow(missing_docs)] // documentation missing in model
    Succeed,
}
```

PR #1989, opened 2022-11-15
Render a Union member of type Unit to an enum variant without inner data

ysaito1001:
## Motivation and Context
Addresses #1546. 

## Description
~~**Work in progress. Opened a draft PR to wash out test failures in CI.**~~

Previously, a `Union` member of type `Unit` was rendered as an enum variant whose inner data was `crate::model::Unit`. The way such a variant appears in Rust feels a bit odd as it usually does not carry inner data for `()`.

We now render a `Union` member of type `Unit` to an enum variant without inner data. The change leads us to make further updates in the following places as they now need to respond to the fact that an enum variant generated from a `Union` member of type `Unit` has no inner data.
- JsonParserGenerator
- XmlBindingTraitParserGenerator
- JsonSerializerGenerator
- QuerySerializerGenerator
- XmlBindingTraitSerializerGenerator

## Testing
Added a new unit test in `UnionGeneratorTest.kt`

~~Running tests in CI.~~ Passed tests in CI.

## Checklist


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
```diff
@@ -114,7 +116,44 @@ class UnionGenerator(
...
     }
 }
+
 fun unknownVariantError(union: String) =
     "Cannot serialize `$union::${UnionGenerator.UnknownVariantName}` for the request. " +
         "The `Unknown` variant is intended for responses only. " +
         "It occurs when an outdated client is used after a new enum variant was added on the server side."
+
+private fun RustWriter.renderVariant(symbolProvider: SymbolProvider, member: MemberShape, memberSymbol: Symbol) {
+    if (member.target.name == "Unit") {
```
jdisanti, line 126: The full ID should be checked just in case the model declares its own `Unit` shape. It will be [`smithy.api#Unit`](https://smithy.io/2.0/spec/model.html#unit-type).

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
```diff
@@ -114,7 +116,44 @@ class UnionGenerator(
...
+    funcNamePart: String,
+    unionSymbol: Symbol,
+    memberSymbol: Symbol,
+) {
+    if (member.target.name == "Unit") {
+        rust(
+            "/// Tries to convert the enum instance into [`$variantName`], extracting the inner `()`.",
+        )
+        rust("/// Returns `Err(&Self)` if it can't be converted.")
+        rustBlock("pub fn as_$funcNamePart(&self) -> std::result::Result<(), &Self>") {
```
jdisanti, line 145: Should we make this `is_$funcNamePart` and return a boolean, or is it useful to return `&Self` as the err case for this use-case?

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGeneratorTest.kt
```diff
@@ -124,6 +124,17 @@ class UnionGeneratorTest {
         writer.compileAndTest()
     }
 
+    @Test
+    fun `unit types should not appear in generated enum`() {
+        val writer = generateUnion("union MyUnion { a: Unit, b: String }", unknownVariant = true)
+        writer.compileAndTest(
+            """
+            let a = MyUnion::A;
+            assert!(a.as_a() == Ok(()));
```
jdisanti, line 133:
```suggestion
            assert_eq!(Ok(()), a.as_a());
```

jdisanti (commented):
Nice code clean up!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
ysaito1001: Thanks for point it out! Addressed in 2b148aa, but the way it's done is not ideal
- creating `ShapeId.from("smithy.api#Unit")` every time a comparison is performed
- multiple places need to be aware of special casing

Planning to refactor these in a subsequent commit.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
```diff
@@ -26,6 +28,12 @@ fun CodegenTarget.renderUnknownVariant() = when (this) {
     CodegenTarget.CLIENT -> true
 }
 
+private val unitShapeId = ShapeId.from("smithy.api#Unit")
+
+internal fun MemberShape.ofTypeUnit(): Boolean {
+    return this.target == unitShapeId
+}
```
Velfi, line 35: I lead you astray here. I think this can be replaced with the existing `isUnionShape` method of `MemberShape`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
ysaito1001: The first bullet was taken care of by fd60317.

The second bullet is not so straightforward. I haven't uncovered a reasonable abstraction that covers all the places in question; each place has a slightly different logic in its if-else expression. For instance, some call their file-local functions in the `else` block and those functions cannot be moved over to a refactored function which might be in a different file; callback lambdas are not that effective for abstraction here either because each call site still has to specify `Writable` for the `if` case and and the `else` case, making it aware that there is an if-else bifurcation based on whether a `Union` member is of the `Unit` type.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
ysaito1001: I am not sure if you did. I think you were right about suggesting we create a helper because I suspect `isUnionShape` does not check whether a member is of type `Unit` (and there does not seem to be `isUnitShape`). When I do replace `ofTypeUnit` with `isUnionShape`, the new unit test added in `UnionGeneratorTest.kt` no longer passes.

Velfi: Ok, sounds good then

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
jdisanti: That's a fair point. We could generate both for booleans.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
ysaito1001: For now, we'll keep what we have to keep it simple.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
```diff
@@ -26,6 +28,12 @@ fun CodegenTarget.renderUnknownVariant() = when (this) {
     CodegenTarget.CLIENT -> true
 }
 
+private val unitShapeId = ShapeId.from("smithy.api#Unit")
+
+internal fun MemberShape.ofTypeUnit(): Boolean {
```
jdisanti, line 33: This should get moved into `Smithy.kt` in `codegen-core`, and renamed to `isTargetUnit`.

ysaito1001: `Smithy.kt` looks like the right home for this extension. Thanks. Addressed in de1f5cf.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -447,9 +448,14 @@ class JsonSerializerGenerator(
...
-            inner(objectName)
-            rust("$objectName.finish();")
+            if (context.shape.isTargetUnit()) {
+                rust("let $objectName = ${context.writerExpression}.start_object();")
+                rust("$objectName.finish();")
+            } else {
+                rust("let mut $objectName = ${context.writerExpression}.start_object();")
+                inner(objectName)
+                rust("$objectName.finish();")
+            }
```
jdisanti, line 458: This can be simplified. If clippy complains, you can throw in an "allow unused mut" attribute.
```suggestion
            rust("let mut $objectName = ${context.writerExpression}.start_object();")
            if (context.shape.isTargetUnit()) {
                inner(objectName)
            }
            rust("$objectName.finish();")
```

Can you also add a comment explaining why the `isTargetUnit` check is necessary in the serializer?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -489,8 +495,15 @@ class JsonSerializerGenerator(
                 rustBlock("match input") {
                     for (member in context.shape.members()) {
                         val variantName = symbolProvider.toMemberName(member)
-                        withBlock("#T::$variantName(inner) => {", "},", unionSymbol) {
-                            serializeMember(MemberContext.unionMember(context, "inner", member, jsonName))
+
+                        if (member.isTargetUnit()) {
+                            withBlock("#T::$variantName => {", "},", unionSymbol) {
+                                serializeMember(MemberContext.unionMember(context, "", member, jsonName))
```
jdisanti, line 501: I'm surprised this works with the reference name being empty. Can we instead just short-circuit to serializing an empty object to avoid any future issues?

jdisanti (commented):
Looks mostly good. Just a couple of comments I should have noticed last time 😅

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
ysaito1001: Thanks! Addressed in 4b9a355. I negated the condition at line 455 (in the suggested) for correctness.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
ysaito1001: Thanks for catching. Commit e4f6559 unconditionally adds a reference name `inner`, but in the case of the Unit type, nothing will be rendered due to our change to `RustWriter.jsonObjectWriter`. One issue is that we have also made a similar change around the member reference name to `RustWriter.serializeUnion` in `QuerySerializerGenerator` but we haven't identify a kind of change over there that we made in `RustWriter.jsonObjectWriter`. We'll see what test breaks in CI.

ysaito1001: Updated protocol tests for serializers and parser generators to wash out test failures. Fixes addressed in 45ef80d and 884155e.

jdisanti (approved):

PR #1993, opened 2022-11-16
formatting: run code cleanup on entire project

Velfi:
This is a small formatting PR. I ran IntelliJ's code cleanup on the entire project and it changed a few things.


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkSettings.kt
```diff
@@ -19,10 +19,6 @@ class SdkSettings private constructor(private val awsSdk: ObjectNode?) {
             SdkSettings(coreRustSettings.customizationConfig?.getObjectMember("awsSdk")?.orNull())
     }
 
-    /** Path to the `sdk-default-configuration.json` config file */
```
jdisanti, line 22: Let's keep this for now since we will need to implement this feature soon.

jdisanti (approved):

PR #2000, opened 2022-11-17
Remove `fromCoreProtocol` in favour of `as` downcasting

hlbarber:
## Motivation and Context

Having a `companion object` on `ServerProtocol` limits extensibility.

## Description

- Use `ProtocolGeneratorFactory` and `as` downcasting instead of `fromCoreProtocol`.
- Fix bug where `ServerRestXmlFactory.protocol` returned `RestXml` rather than `ServerRestXmlProtocol`.

### DIFF
--- a/codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerCodegenVisitor.kt
+++ b/codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerCodegenVisitor.kt
@@ -161,7 +161,7 @@ class PythonServerCodegenVisitor(
             rustCrate,
             protocolGenerator,
             protocolGeneratorFactory.support(),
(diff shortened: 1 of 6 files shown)


codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerCodegenVisitor.kt
```diff
@@ -161,7 +161,7 @@ class PythonServerCodegenVisitor(
             rustCrate,
             protocolGenerator,
             protocolGeneratorFactory.support(),
-            ServerProtocol.fromCoreProtocol(protocolGeneratorFactory.protocol(codegenContext)),
+            protocolGeneratorFactory.protocol(codegenContext) as ServerProtocol,
```
hlbarber, line 164: I'm unsure whether we even need this `as ServerProtocol`?

david-perez (approved):

PR #2002, opened 2022-11-17
Python: TLS Support for servers

unexge:
## Motivation and Context
This PR adds support for TLS termination on the process:
```python
from libpokemon_service_server_sdk.tls import TlsConfig

app.run(
    workers=1,
    tls=TlsConfig(
        key_path="./localhost.key",
        cert_path="./localhost.crt",
    )
)
```

Fixes https://github.com/awslabs/smithy-rs/issues/1751

## Testing
It can be tested via enabling TLS support on Pokemon service:
```bash
$ cd rust-runtime/aws-smithy-http-server-python/examples
$ python3 pokemon_service.py --enable-tls \
    --tls-key-path ./pokemon-service-test/tests/testdata/localhost.key \
    --tls-cert-path ./pokemon-service-test/tests/testdata/localhost.crt
```

```bash
$ curl https://localhost:13734/pokemon-species/pikachu -k
```

## TODOs
- [ ] Support for refreshing certificates periodically
- [ ] Feature flag for TLS support?


rust-runtime/aws-smithy-http-server-python/src/tls/listener.rs
```diff
@@ -0,0 +1,66 @@
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Conn, Self::Error>>> {
+        // Replace current acceptor (it also contains TLS config) if there is a new one
+        if let Ok(acceptor) = self.new_acceptor_rx.try_recv() {
+            self.as_mut().project().inner.replace_acceptor_pin(acceptor);
+        }
+
+        loop {
+            match ready!(self.as_mut().project().inner.poll_next(cx)) {
```
hlbarber, line 54: We _could_ flatten this to one match.

rust-runtime/aws-smithy-http-server-python/src/tls/listener.rs
```diff
@@ -0,0 +1,66 @@
...
+                Some(conn) => match conn {
+                    Ok(conn) => return Poll::Ready(Some(Ok(conn))),
+                    Err(err) => {
+                        // Don't propogate errors to Hyper because it causes server to shutdown
+                        tracing::error!(error = ?err, "tls connection error");
```
hlbarber, line 59: This is going to blast us with `tracing::error` if the inner `TlsListener<A, T>` gets into a state where `poll_next` always returns `Some(Err(_))`. Is this possible? Do we want to add a delay or shutdown hyper?

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
```diff
@@ -265,17 +265,18 @@ class PythonApplicationGenerator(
                     Ok(())
                 }
                 /// Main entrypoint: start the server on multiple workers.
-                ##[pyo3(text_signature = "(${'$'}self, address, port, backlog, workers)")]
+                ##[pyo3(text_signature = "(${'$'}self, address, port, backlog, workers, tls)")]
```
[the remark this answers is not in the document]
crisidev: We are not really doing Python changelogs yet..

rust-runtime/aws-smithy-http-server-python/examples/pokemon-service-test/tests/helpers.rs
```diff
@@ -3,9 +3,61 @@
...
+            }
+        }
+
+        let process = Command::new("python3")
+            .args(args)
+            .group_spawn()
+            .expect("failed to spawn the Pokémon Service program");
+
+        // The Python interpreter takes a little to startup.
+        time::sleep(Duration::from_secs(2)).await;
```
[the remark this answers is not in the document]
crisidev: This is not racy since it is in a test and it has been working both in CI and locally for months. Python takes a while to startup, that's it. The effort of using a signal would be useless IMHO since this works perfectly already. We do the same in Rust BTW, just a smaller sleep because its faster to startup.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonApplicationGenerator.kt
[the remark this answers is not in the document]
hlbarber: gotcha

rust-runtime/aws-smithy-http-server-python/src/tls/listener.rs
unexge: You mean removing the loop around the match? If so, I've borrowed this part from `tokio_stream::StreamExt::filter`: https://github.com/tokio-rs/tokio/blob/master/tokio-stream/src/stream_ext/filter.rs#L42-L53 and I think the loop is there to optimise if we skip a value we should poll the next value in same cycle

hlbarber: I mean

```rust
match ... {
    Some(Ok(...)) => ...,
    Some(Err(...) => ...,
    None => ...
}
```

rust-runtime/aws-smithy-http-server-python/examples/pokemon-service-test/tests/helpers.rs
```diff
@@ -3,9 +3,61 @@
...
+
+        process
+    }
+
+    fn base_url(&self) -> String {
```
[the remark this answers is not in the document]
unexge: Done in https://github.com/awslabs/smithy-rs/pull/2002/commits/7a89656b9ff1973cd4adef2305adf4d701d87192

rust-runtime/aws-smithy-http-server-python/src/tls/listener.rs
unexge: Ohh sorry, done in https://github.com/awslabs/smithy-rs/pull/2002/commits/37bc9fff8401534c74983a3d026d300a44482792

rust-runtime/aws-smithy-http-server-python/src/tls/listener.rs
unexge: Good point, inner `TlsListener` returns either an `TlsAcceptor::Error` or a TLS handshake error: https://github.com/tmccombs/tls-listener/blob/main/src/lib.rs#L143-L152

I believe handshake errors shouldn't cause inner listener to stuck, so I've changed to propagate `TlsAcceptor::Error`s to upstream and reduced TLS handshake error logs to `DEBUG` level: https://github.com/awslabs/smithy-rs/pull/2002/commits/1710550576e369b18cd0a16015a04118d20b95ec

hlbarber: I'm unsure of when a `ListenerError` vs a `TlsAcceptError` occurs. If clients can cause a `ListenerError` that's a problem I'm guessing.

unexge: I'm also not sure but I'm looking to `warp` and seems like they also propagate listener errors: https://github.com/seanmonstar/warp/blob/4e9c4fd6ce238197fd1088061bbc07fa2852cb0f/src/tls.rs#L381

hlbarber: Maybe it's worth just doing a little check like scrambling the TLS cert for client/server and see what happens.

unexge: I've adde some tests with https://github.com/awslabs/smithy-rs/pull/2002/commits/ce6a85e9600eacb3802547f0ee7ef6e8e082f70a

crisidev (approved):

rust-runtime/aws-smithy-http-server-python/src/tls.rs
```diff
@@ -0,0 +1,173 @@
...
+
+/// PyTlsConfig represents TLS configuration created from Python.
+#[pyclass(
+    name = "TlsConfig",
+    text_signature = "(*, key_path, cert_path, reload)"
+)]
+#[derive(Clone)]
+pub struct PyTlsConfig {
+    /// Absolute path of the RSA or PKCS private key.
+    key_path: String,
```
chadac, line 29: Question since I'm not too familiar with Rust choices: Why `String` rather than `std::path::Path`?

chadac (approved):

rust-runtime/aws-smithy-http-server-python/src/tls.rs
unexge: Good point! No reason, I've changed it to use `PathBuf` which is owned version of `Path`

PR #2003, opened 2022-11-17
Improve `Plugin` toolkit

hlbarber:
## Motivation and Context

Writing plugins can be made easier if we provide useful tools to do so.

## Description

- Add `OperationNameFn` and `plugin_from_operation_name_fn` which adds the ability to create a `Plugin` from a closure from operation name to a layer.
- Add `HttpLayer` which adds the ability to turn a HTTP layer directly into a `Plugin`.

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
+++ b/rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
@@ -53,6 +53,9 @@ where
 /// // This plugin applies the `PrintService` middleware around every operation.
 /// let plugin = plugin_from_operation_name_fn(f);
 /// ```
-pub fn plugin_from_operation_name_fn<F>(f: F) -> OperationNameFn<F> {
+pub fn plugin_from_operation_name_fn<L, F>(f: F) -> OperationNameFn<F>
+where
+    F: Fn(&'static str) -> L,
+{
     OperationNameFn { f }
 }
--- a/rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
+++ b/rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
@@ -1,0 +1,29 @@
+use tower::layer::util::Stack;
+
+use crate::operation::{Operation, OperationShape};
+
+use super::Plugin;
+
+/// A [`Plugin`] implemented by a closure over the operation name. See [`plugin_from_operation_name_fn`] for more details.
+pub struct OperationNameFn<F> {
+    f: F,
+}
+
+impl<P, Op, S, ExistingLayer, NewLayer, F> Plugin<P, Op, S, ExistingLayer> for OperationNameFn<F>
+where
+    F: Fn(&'static str) -> NewLayer,
+    Op: OperationShape,
+{
+    type Service = S;
+    type Layer = Stack<ExistingLayer, NewLayer>;
+
+    fn map(&self, input: Operation<S, ExistingLayer>) -> Operation<Self::Service, Self::Layer> {
+        input.layer((self.f)(Op::NAME))
+    }
+}
+
+/// Constructs a [`Plugin`] using a closure over the operation name `F: Fn(&'static str) -> L` where `L` is a HTTP
+/// [`Layer`](tower::Layer).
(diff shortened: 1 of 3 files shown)


rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
```diff
@@ -0,0 +1,29 @@
...
+    type Service = S;
+    type Layer = Stack<ExistingLayer, NewLayer>;
+
+    fn map(&self, input: Operation<S, ExistingLayer>) -> Operation<Self::Service, Self::Layer> {
+        input.layer((self.f)(Op::NAME))
+    }
+}
+
+/// Constructs a [`Plugin`] using a closure over the operation name `F: Fn(&'static str) -> L` where `L` is a HTTP
+/// [`Layer`](tower::Layer).
```
jjant, line 31: Could you add an example here?

rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
```diff
@@ -0,0 +1,29 @@
...
+pub fn plugin_from_operation_name_fn<F>(f: F) -> OperationNameFn<F> {
```
jjant, line 32: Would it be useful to add the constraint `F: Fn(&'static str) -> NewLayer` here to force earlier compile errors on wrong usage?

rust-runtime/aws-smithy-http-server/src/plugin/layer.rs
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use tower::layer::util::Stack;
+
+use crate::operation::Operation;
+
+use super::Plugin;
+
+/// A [`Plugin`] which appends a HTTP [`Layer`](tower::Layer) to the existing `L` in [`Operation<S, L>`](Operation).
```
jjant, line 12: What do you think about this?
```suggestion
/// A [`Plugin`] which appends a HTTP [`Layer`](tower::Layer) `L` to the existing `Layer` in [`Operation<S, Layer>`](Operation).
```

rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
```diff
@@ -0,0 +1,34 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use tower::layer::util::Stack;
+
+use crate::operation::{Operation, OperationShape};
+
+use super::Plugin;
+
+/// A [`Plugin`] implemented by a closure over the operation name. See [`plugin_from_operation_name_fn`] for more details.
```
LukeMathWalker, line 12:
```suggestion
/// An adapter to convert a `Fn(&'static str) -> Layer` closure into a [`Plugin`]. See [`plugin_from_operation_name_fn`] for more details.
```

rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
hlbarber: The reason I haven't added it here is because [layer_fn](https://docs.rs/tower/latest/tower/layer/fn.layer_fn.html) doesn't have these bounds and, prior to [`fn-traits`](https://doc.rust-lang.org/beta/unstable-book/library-features/fn-traits.html) I think we'd have to add a second generic on here `<L, F>` to specify the return type of `Fn(&'static str) -> L`.

rust-runtime/aws-smithy-http-server/src/plugin/layer.rs
hlbarber: I think this is probably understandable that this is a `Layer` given that we've doc linked to it.

rust-runtime/aws-smithy-http-server/src/plugin/closure.rs
LukeMathWalker: Yeah, we would have to add an `L` parameter - but I think it's worth it for the more precise error message.

LukeMathWalker (approved):

PR #2005, opened 2022-11-18
@range implementation for integer shapes

LukeMathWalker:
Relevant issue: https://github.com/awslabs/smithy-rs/issues/1968

## Checklist

### DIFF
--- a/codegen-core/common-test-models/malformed-range-extras.smithy
+++ b/codegen-core/common-test-models/malformed-range-extras.smithy
@@ -1,0 +1,662 @@
+$version: "2.0"
+
+namespace aws.protocoltests.extras.restjson.validation
+
+use aws.api#service
+use aws.protocols#restJson1
+use smithy.test#httpMalformedRequestTests
+use smithy.framework#ValidationException
+
+/// A REST JSON service that sends JSON requests and responses with validation applied
+@service(sdkId: "Rest Json Validation Protocol")
+@restJson1
+service RestJsonValidation {
+    version: "2022-11-23",
+    operations: [
+        MalformedRange,
+        MalformedRangeOverride,
+    ]
+}
+
+@suppress(["UnstableTrait"])
+@http(uri: "/MalformedRange", method: "POST")
+operation MalformedRange {
+    input: MalformedRangeInput,
+    errors: [ValidationException]
+}
+
+@suppress(["UnstableTrait"])
+@http(uri: "/MalformedRangeOverride", method: "POST")
+operation MalformedRangeOverride {
+    input: MalformedRangeOverrideInput,
+    errors: [ValidationException]
+}
+
+apply MalformedRange @httpMalformedRequestTests([
+    {
+        id: "RestJsonMalformedRangeShort",
+        documentation: """
+        When a short member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "byte" : $value:L }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value $value:L at '/short' failed to satisfy constraint: Member must be between 2 and 8, inclusive",
+                      "fieldList" : [{"message": "Value $value:L at '/short' failed to satisfy constraint: Member must be between 2 and 8, inclusive", "path": "/short"}]}"""
+                }
+            }
+        },
+        testParameters: {
+            value: ["1", "9"]
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMinShort",
+        documentation: """
+        When a short member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "minShort" : 1 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 1 at '/minShort' failed to satisfy constraint: Member must be greater than or equal to 2",
+                      "fieldList" : [{"message": "Value 1 at '/minShort' failed to satisfy constraint: Member must be greater than or equal to 2", "path": "/minShort"}]}"""
+                }
+            }
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMaxShort",
+        documentation: """
+        When a short member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "maxShort" : 9 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 9 at '/maxShort' failed to satisfy constraint: Member must be less than or equal to 8",
+                      "fieldList" : [{"message": "Value 9 at '/maxShort' failed to satisfy constraint: Member must be less than or equal to 8", "path": "/maxShort"}]}"""
+                }
+            }
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeInteger",
+        documentation: """
+        When a integer member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "byte" : $value:L }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value $value:L at '/integer' failed to satisfy constraint: Member must be between 2 and 8, inclusive",
+                      "fieldList" : [{"message": "Value $value:L at '/integer' failed to satisfy constraint: Member must be between 2 and 8, inclusive", "path": "/integer"}]}"""
+                }
+            }
+        },
+        testParameters: {
+            value: ["1", "9"]
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMinInteger",
+        documentation: """
+        When a integer member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "minInteger" : 1 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 1 at '/minInteger' failed to satisfy constraint: Member must be greater than or equal to 2",
+                      "fieldList" : [{"message": "Value 1 at '/minInteger' failed to satisfy constraint: Member must be greater than or equal to 2", "path": "/minInteger"}]}"""
+                }
+            }
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMaxInteger",
+        documentation: """
+        When a integer member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "maxInteger" : 9 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 9 at '/maxInteger' failed to satisfy constraint: Member must be less than or equal to 8",
+                      "fieldList" : [{"message": "Value 9 at '/maxInteger' failed to satisfy constraint: Member must be less than or equal to 8", "path": "/maxInteger"}]}"""
+                }
+            }
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeLong",
+        documentation: """
+        When a long member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "byte" : $value:L }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value $value:L at '/long' failed to satisfy constraint: Member must be between 2 and 8, inclusive",
+                      "fieldList" : [{"message": "Value $value:L at '/long' failed to satisfy constraint: Member must be between 2 and 8, inclusive", "path": "/long"}]}"""
+                }
+            }
+        },
+        testParameters: {
+            value: ["1", "9"]
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMinLong",
+        documentation: """
+        When a long member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "minLong" : 1 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 1 at '/minLong' failed to satisfy constraint: Member must be greater than or equal to 2",
+                      "fieldList" : [{"message": "Value 1 at '/minLong' failed to satisfy constraint: Member must be greater than or equal to 2", "path": "/minLong"}]}"""
+                }
+            }
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMaxLong",
+        documentation: """
+        When a long member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRange",
+            body: """
+            { "maxLong" : 9 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 9 at '/maxLong' failed to satisfy constraint: Member must be less than or equal to 8",
+                      "fieldList" : [{"message": "Value 9 at '/maxLong' failed to satisfy constraint: Member must be less than or equal to 8", "path": "/maxLong"}]}"""
+                }
+            }
+        }
+    },
+])
+
+// now repeat the above tests, but for the more specific constraints applied to the input member
+apply MalformedRangeOverride @httpMalformedRequestTests([
+    {
+        id: "RestJsonMalformedRangeShortOverride",
+        documentation: """
+        When a short member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRangeOverride",
+            body: """
+            { "short" : $value:L }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value $value:L at '/short' failed to satisfy constraint: Member must be between 4 and 6, inclusive",
+                      "fieldList" : [{"message": "Value $value:L at '/short' failed to satisfy constraint: Member must be between 4 and 6, inclusive", "path": "/short"}]}"""
+                }
+            }
+        },
+        testParameters: {
+            value: ["3", "7"]
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMinShortOverride",
+        documentation: """
+        When a short member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRangeOverride",
+            body: """
+            { "minShort" : 3 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 3 at '/minShort' failed to satisfy constraint: Member must be greater than or equal to 4",
+                      "fieldList" : [{"message": "Value 3 at '/minShort' failed to satisfy constraint: Member must be greater than or equal to 4", "path": "/minShort"}]}"""
+                }
+            }
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMaxShortOverride",
+        documentation: """
+        When a short member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRangeOverride",
+            body: """
+            { "maxShort" : 7 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 7 at '/maxShort' failed to satisfy constraint: Member must be less than or equal to 6",
+                      "fieldList" : [{"message": "Value 7 at '/maxShort' failed to satisfy constraint: Member must be less than or equal to 6", "path": "/maxShort"}]}"""
+                }
+            }
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeIntegerOverride",
+        documentation: """
+        When a integer member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRangeOverride",
+            body: """
+            { "integer" : $value:L }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value $value:L at '/integer' failed to satisfy constraint: Member must be between 4 and 6, inclusive",
+                      "fieldList" : [{"message": "Value $value:L at '/integer' failed to satisfy constraint: Member must be between 4 and 6, inclusive", "path": "/integer"}]}"""
+                }
+            }
+        },
+        testParameters: {
+            value: ["3", "7"]
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMinIntegerOverride",
+        documentation: """
+        When a integer member does not fit within range bounds,
+        the response should be a 400 ValidationException.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedRangeOverride",
+            body: """
+            { "minInteger" : 3 }""",
+            headers: {
+                "content-type": "application/json"
+            }
+        },
+        response: {
+            code: 400,
+            headers: {
+                "x-amzn-errortype": "ValidationException"
+            },
+            body: {
+                mediaType: "application/json",
+                assertion: {
+                    contents: """
+                    { "message" : "1 validation error detected. Value 3 at '/minInteger' failed to satisfy constraint: Member must be greater than or equal to 4",
+                      "fieldList" : [{"message": "Value 3 at '/minInteger' failed to satisfy constraint: Member must be greater than or equal to 4", "path": "/minInteger"}]}"""
+                }
+            }
+        }
+    },
+    {
+        id: "RestJsonMalformedRangeMaxIntegerOverride",
+        documentation: """
+        When a integer member does not fit within range bounds,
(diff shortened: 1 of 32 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -328,15 +334,37 @@ class JsonParserGenerator(
...
+                rustTemplate(
+                    """
+                    #{expect_number_or_null}(tokens.next())?
+                        .map(#{NumberPrimitive}::try_from)
+                        .transpose()?
+                    """,
+                    "NumberPrimitive" to RustType.Integer(precision),
+                    *codegenScope,
+                )
+            } else {
```
david-perez, line 357: `returnSymbolToParse` should help you in coalescing both arms, giving you the unconstrained symbol for both cases.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstrainedShapeSymbolProvider.kt
```diff
@@ -103,6 +104,14 @@ class ConstrainedShapeSymbolProvider(
                     base.toSymbol(shape)
                 }
             }
+            is IntegerShape -> {
```
david-perez, line 107: Can you add a test for this arm in `ConstrainedShapeSymbolProvider.kt`?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstraintViolationSymbolProvider.kt
```diff
@@ -117,6 +118,20 @@ class ConstraintViolationSymbolProvider(
                     .definitionFile(Models.filename)
                     .build()
             }
+            is IntegerShape -> {
```
david-perez, line 121: We should make this DRYer given that the arm is identical to `StringShape`. See how `constraintViolationSymbolForCollectionOrMapOrUnionShape` is used.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RangeTraitValidationErrorMessage.kt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.rust.codegen.server.smithy
+
+import software.amazon.smithy.model.traits.RangeTrait
+
+fun RangeTrait.validationErrorMessage(): String {
```
david-perez, line 10: Should probably make this DRYer with `LengthTraitValidationErrorMessage.kt` if the protocol tests specify the messages coincide.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntegerGenerator.kt
```diff
@@ -0,0 +1,187 @@
...
+                fn from(value: $inner) -> Self {
+                    Self::Unconstrained(value)
+                }
+            }
+
+            impl #{From}<$name> for #{MaybeConstrained} {
+                fn from(value: $name) -> Self {
+                    Self::Constrained(value)
+                }
+            }
```
david-perez, line 113: We don't need this one do we?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntegerGenerator.kt
```diff
@@ -0,0 +1,187 @@
...
+            "AsRef" to RuntimeType.AsRef,
+        )
+
+        val constraintViolationModuleName = constraintViolation.namespace.split(constraintViolation.namespaceDelimiter).last()
+        writer.withModule(RustModule(constraintViolationModuleName, RustMetadata(visibility = constrainedTypeVisibility))) {
+            rust(
+                """
+                ##[derive(Debug, PartialEq)]
+                pub enum ${constraintViolation.name} {
+                    Size($inner),
```
david-perez, line 161: `Range` may be a better name, to match the trait name?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntegerGenerator.kt
```diff
@@ -0,0 +1,187 @@
...
+                fn from(value: $name) -> Self {
+                    value.into_inner()
+                }
+            }
+
+            impl #{AsRef}<$inner> for $name {
+                fn as_ref(&self) -> &$inner {
+                    self.inner()
+                }
+            }
```
david-perez, line 144: I'd remove this one for now. We don't need it for the generated code, and it's unclear what traits we should implement for newtypes.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntGeneratorTest.kt
```diff
@@ -0,0 +1,124 @@
...
+import software.amazon.smithy.rust.codegen.core.smithy.ModelsModule
+import software.amazon.smithy.rust.codegen.core.testutil.TestWorkspace
+import software.amazon.smithy.rust.codegen.core.testutil.asSmithyModel
+import software.amazon.smithy.rust.codegen.core.testutil.compileAndTest
+import software.amazon.smithy.rust.codegen.core.testutil.unitTest
+import software.amazon.smithy.rust.codegen.core.util.lookup
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverTestCodegenContext
+import java.util.stream.Stream
+
+class ConstrainedIntGeneratorTest {
```
david-perez, line 27: Rename the file and class name to match those of of what it's testing, `ConstrainedIntegerGeneratorTest`.

david-perez (changes requested):
Looks promising! We need to expand coverage in `constraints.smithy`, binding integer shapes to the different parts of the HTTP message; I foresee the `!= 0` checks when serializing are going to interfere.

I also think adding support for the other integral shapes (`byte`, `short`, `long`) should be part of this PR, you're not far off.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntGeneratorTest.kt
LukeMathWalker: Done in [f50b7b0](https://github.com/awslabs/smithy-rs/pull/2005/commits/f50b7b0809acb5be4e8601681bdcbb0f08f8fcd1)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntegerGenerator.kt
LukeMathWalker: Done in [0b80d2e](https://github.com/awslabs/smithy-rs/pull/2005/commits/0b80d2e9c2fd13b1a469b07d2b1074aec79469de) (and https://github.com/awslabs/smithy-rs/pull/2005/commits/58c156fd0e1d95caeb4fc7f1b7efd6c98889003f)

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
LukeMathWalker: Done in https://github.com/awslabs/smithy-rs/pull/2005/commits/a959004accbdde4e577482c8b47a347eb1acffc9

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntegerGenerator.kt
LukeMathWalker: No, we don't - removed in https://github.com/awslabs/smithy-rs/pull/2005/commits/d71932886636f7c7a8fecdd62d836d5143f32c55

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntegerGenerator.kt
LukeMathWalker: I agree, fixed in https://github.com/awslabs/smithy-rs/pull/2005/commits/6e54c5e4ad228be25c6cb27d54766cc00c31f832

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ValidateUnsupportedConstraintsAreNotUsedTest.kt
```diff
@@ -195,25 +195,6 @@ internal class ValidateUnsupportedConstraintsAreNotUsedTest {
         validationResult.messages[0].message shouldContain "The string shape `test#PatternString` has the constraint trait `smithy.api#pattern` attached"
     }
 
-    @Test
```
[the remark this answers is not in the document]
LukeMathWalker: Done in https://github.com/awslabs/smithy-rs/pull/2005/commits/19297bca2ac3ee5c7f6767caad046b8d6a373a27

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstrainedShapeSymbolProvider.kt
LukeMathWalker: Done in https://github.com/awslabs/smithy-rs/pull/2005/commits/35dfb824c2bf8fbf8770a852c833063c68979d1b

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstraintViolationSymbolProvider.kt
LukeMathWalker: Done in https://github.com/awslabs/smithy-rs/pull/2005/commits/72cda3a82631a88dfba9caefea450547f3d91fa3

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RangeTraitValidationErrorMessage.kt
LukeMathWalker: The message are currently slightly different, what do you mean by

> if the protocol tests specify the messages coincide.

?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -362,6 +364,21 @@ class JsonSerializerGenerator(
...
+                    } else {
+                        rust("(${value.asValue()}).into()")
+                    }
+                }
+                rustTemplate(
+                    "$writer.number(##[allow(clippy::useless_conversion)]#{Variant}::NegInt(#{Conversion:W}));",
+                    "Variant" to smithyTypes.member("Number"),
+                    "Conversion" to conversion,
+                )
+            }
```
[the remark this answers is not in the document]
LukeMathWalker: I have introduced customizations in a way that keeps the parser generators unaware of constraint traits 👌🏻

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
```diff
@@ -475,18 +474,44 @@ class RustWriter private constructor(
...
+            else -> this.block(outerField)
+        }
+    }
+
+    /**
+     * Generate a wrapping if statement around a primitive field.
+     * The specified block will only be called if the field is not set to its default value - `0` for
+     * numbers, `false` for booleans.
+     */
+    fun ifNotDefault(shape: Shape, outerField: String, block: RustWriter.(field: String) -> Unit) {
```
david-perez, line 486: It seems to me like this or `software.amazon.smithy.rust.codegen.core.smithy.protocols.serialize.SerializerUtil` could be made DRYer now.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstrainedShapeSymbolProviderTest.kt
```diff
@@ -64,24 +73,31 @@ class ConstrainedShapeSymbolProviderTest {
...
-        constrainedMapType shouldBe RustType.Opaque("ConstrainedMap", "crate::model")
+    companion object {
+        @JvmStatic
+        fun getShapes(): Stream<Arguments> {
+            return Stream.of(
+                Arguments.of("ConstrainedInteger", { s: Shape -> s.isIntegerShape }),
+                Arguments.of("ConstrainedString", { s: Shape -> s.isStringShape }),
+                Arguments.of("ConstrainedMap", { s: Shape -> s.isMapShape }),
+            )
+        }
```
david-perez, line 94: If your function body comprises a single expression that is returned, you can use this shorthand way of writing functions.

```suggestion
        fun getShapes(): Stream<Arguments> =
            Stream.of(
                Arguments.of("ConstrainedInteger", { s: Shape -> s is IntegerShape }),
                Arguments.of("ConstrainedString", { s: Shape -> s is StringShape }),
                Arguments.of("ConstrainedMap", { s: Shape -> s is MapShape }),
            )
```

Doesn't really matter in this case, but we prefer the instanceof checks (e.g. `s is StringShape`) over the Smithy library-provided method calls, because the Kotlin compiler can do type narrowing in subsequent code.

Can you also move this function before the first test?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RangeTraitValidationErrorMessage.kt
david-perez: There are protocol tests that assert that we reject malformed requests, and stipulate the format of the error message that we need to return in the response: https://github.com/awslabs/smithy-rs/blob/87e4766f9896663e4e00dc8d33b9cc7c49601ea0/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt#L987-L988

When we implement everything correctly, you should see those tests start passing and hence be forced to remove them from `ExpectFail`.

The format of the error message is [here](https://github.com/awslabs/smithy/blob/a8fd6c537d3d1e0c0258741efceda3ea85a8c75b/smithy-aws-protocol-tests/model/restJson1/validation/malformed-range.smithy#L81-L82). It doesn't begin with "The integer value", but rather with "Value x at...".

My comment above was saying that we could extract the definition of `ending`, since it seems to be the same in both error messages.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ValidateUnsupportedConstraints.kt
```diff
@@ -224,11 +224,12 @@ fun validateUnsupportedConstraints(model: Model, service: ServiceShape, codegenC
         .map { (shape, patternTrait) -> UnsupportedPatternTraitOnStringShape(shape, patternTrait as PatternTrait) }
         .toSet()
 
-    // 6. Range trait on any shape is used. It has not been implemented yet.
+    // 6. Range trait used on a non-integer shape. It has not been implemented yet.
     // TODO(https://github.com/awslabs/smithy-rs/issues/1401)
     val unsupportedRangeTraitOnShapeSet = walker
         .walkShapes(service)
         .asSequence()
+        .filter { !it.isIntegerShape }
```
david-perez, line 232:
```suggestion
        .filterIsInstance<IntegerShape>()
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedIntegerGeneratorTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    fun `type should not be constructible without using a constructor`() {
+        val model = """
+            namespace test
+
+            @range(min: -1, max: 69)
+            integer ConstrainedInteger
+        """.asSmithyModel()
```
david-perez, line 112:
```suggestion
        val model = 
            """
            namespace test

            @range(min: -1, max: 69)
            integer ConstrainedInteger
            """.asSmithyModel()
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RangeTraitValidationErrorMessage.kt
david-perez: I've noticed that Smithy only provides tests for malformed bytes and floats. We should write our own for `short`s, `integer`s, `long`s and `double`s and contribute them upstream. Since you're only generating constrained integers, you won't be able to test that the implementation is correct unless you write tests for `integer`s in this PR too.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -517,72 +531,165 @@ class HttpBindingGenerator(
...
+            val formatted = headerFmtFun(
+                this,
+                shape,
+                timestampFormat,
+                variableName,
+                isMultiValuedHeader = isMultiValuedHeader,
+            )
+            val safeName = safeName("formatted")
+            write("let $safeName = $formatted;")
+            rustBlock("if !$safeName.is_empty()") {
```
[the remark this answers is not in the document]
LukeMathWalker: Fixed in https://github.com/awslabs/smithy-rs/pull/2005/commits/64e3312c7ddc01570645df0f98cce2451d45c6ca

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstrainedShapeSymbolProviderTest.kt
LukeMathWalker: Thanks for the explanation!
Done in https://github.com/awslabs/smithy-rs/pull/2005/commits/0ec9f136a9ff170ed0f3ca4c0a87e0ce77fd90ec

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -331,10 +332,10 @@ class JsonParserGenerator(
             rustTemplate(
                 """
                 #{expect_number_or_null}(tokens.next())?
-                    .map(#{NumberType}::try_from)
+                    .map(#{Type}::try_from)
```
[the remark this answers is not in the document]
LukeMathWalker: Done in [f927eaf](https://github.com/awslabs/smithy-rs/pull/2005/commits/f927eaf44e59f40fb7c8db8dc469edd906baccc9)

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/ValueExpression.kt
```diff
@@ -22,4 +22,11 @@ sealed class ValueExpression {
         is Reference -> name
         is Value -> "&$name"
     }
+
+    override fun toString(): String {
+        return when (this) {
+            is Reference -> name
+            is Value -> name
+        }
+    }
```
jjant, line 31: Do you need the `when` at all? I would've thought no because of line 11.
Why don't we need a `&` for the `Reference` case? 🤔

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
```diff
@@ -67,7 +67,7 @@ fun testRustSettings(
 
 private const val SmithyVersion = "1.0"
 fun String.asSmithyModel(sourceLocation: String? = null, smithyVersion: String = SmithyVersion): Model {
-    val processed = letIf(!this.startsWith("\$version")) { "\$version: ${smithyVersion.dq()}\n$it" }
+    val processed = letIf(!this.trimStart().startsWith("\$version")) { "\$version: ${smithyVersion.dq()}\n$it" }
```
jjant, line 70: Out of curiosity, was this needed because we removed some `trimIndent()` somewhere?

jjant (approved):
Great job! 🌟 
Only left a couple of questions.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
LukeMathWalker: It was needed because I added `$version` to a test model (that was implicitly versioned to 1 before) and since I did 
```
"""
$version 
```
it wasn't picking it up 😅 
So I decided to make it a bit more robust.

codegen-core/common-test-models/malformed-range-extras.smithy
```diff
@@ -0,0 +1,662 @@
+$version: "2.0"
```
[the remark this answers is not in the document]
LukeMathWalker: Yes, these are hand-written. 
I was planning to open an issue on the Smithy repo (linking the file) once this PR is merged.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -997,6 +998,31 @@ class ServerProtocolTestGenerator(
             FailingTest(RestJsonValidation, "RestJsonMalformedRangeMinFloat", TestType.MalformedRequest),
             FailingTest(RestJsonValidation, "RestJsonMalformedPatternSensitiveString", TestType.MalformedRequest),
 
+            // Tests involving using @range on bytes, shorts and longs.
+            // See https://github.com/awslabs/smithy-rs/issues/1968
```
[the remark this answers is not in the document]
LukeMathWalker: I'll update the issue title once this is merged.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/ValueExpression.kt
LukeMathWalker: Good catch! 
The `Reference` case can either be a `&variable_name` or simply `variable_name` with the implicit understanding that the variable points to a reference, so we should avoid adding an additional `&`.

LukeMathWalker: Basically, I want references to be rendered as references and values to be rendered as value if no other method has been explicitly called (i.e. `asValue`/`asRef`).

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/SerializerUtil.kt
```diff
@@ -6,28 +6,17 @@
...
 import software.amazon.smithy.model.shapes.MemberShape
-import software.amazon.smithy.model.shapes.NumberShape
 import software.amazon.smithy.model.shapes.StructureShape
 import software.amazon.smithy.rust.codegen.core.rustlang.RustWriter
 import software.amazon.smithy.rust.codegen.core.rustlang.Writable
 import software.amazon.smithy.rust.codegen.core.rustlang.rustBlock
 
 class SerializerUtil(private val model: Model) {
     fun RustWriter.ignoreZeroValues(shape: MemberShape, value: ValueExpression, inner: Writable) {
-        val expr = when (model.expectShape(shape.target)) {
```
david-perez, line 21: We need to revert this. This affects the client. We said we have to evaluate the impact of serializing zero values in the SDK. This is a place where behavior among clients and servers will be different: clients will implement presence-tracking, servers will always serialize everything.

david-perez (changes requested):
We said we have to evaluate the impact of serializing zero values in the SDK.

If the diff to the generated SDK is functional, @awslabs/rust-sdk-owners need to approve the PR.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/SerializerUtil.kt
LukeMathWalker: I've not changed the behaviour here - this is just a refactoring pending a decision on what we want to do going forward for zero values in the SDK.

codegen-core/common-test-models/malformed-range-extras.smithy
```diff
@@ -0,0 +1,662 @@
...
+@restJson1
+service MalformedRangeValidation {
+    version: "2022-11-23",
+    operations: [
+        MalformedRange,
+        MalformedRangeOverride,
+    ]
+}
+
+@suppress(["UnstableTrait"])
```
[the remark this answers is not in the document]
LukeMathWalker: I've mimicked the annotations on https://github.com/awslabs/smithy/blob/a8fd6c537d3d1e0c0258741efceda3ea85a8c75b/smithy-aws-protocol-tests/model/restJson1/validation/malformed-range.smithy#L16 since we plan to merge this upstream later on. I'm not sure what the purpose of it is in their own model.

david-perez (approved):
My bad, I stopped reading when I saw the deleted hunk in `SerializerUtil`. I now see this has coalesced with `ifNotDefault` as per https://github.com/awslabs/smithy-rs/pull/2005#discussion_r1030239137.

Looking at the diff, it's odd that we're now generating this, can we fix it?

```
- if input.limit != 0 {
+ let var_82 = &input.limit;
+ if *var_82 != 0 {
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
```diff
@@ -465,33 +468,63 @@ class RustWriter private constructor(
...
+     */
+    fun ifSet(
+        shape: Shape,
+        member: Symbol,
+        variable: ValueExpression,
+        block: RustWriter.(field: ValueExpression) -> Unit,
+    ) {
+        ifSome(member, variable) { inner -> ifNotDefault(shape, inner, block) }
+    }
+
```
Velfi, line 527: I think we should start including examples of generated code for these so that people can understand how they work at a glance. Just 2 snippets of code for each would be helpful.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -449,7 +460,9 @@ class HttpBindingGenerator(
      */
     // Rename here technically not required, operations and members cannot be renamed.
```
Velfi, line 461: Looking at this comment, I'm wondering if we still need this function at all. What's the purpose of renaming if we don't need to?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -517,72 +531,165 @@ class HttpBindingGenerator(
...
+            }
+        }
+
+        safeName().also { local ->
+            // This variable assignment ensures that the name of the value expression is always a valid variable name
+            // instead of being, for example, a field access expression (e.g. `my_struct.my_field`).
+            // This allows later customization to leverage variable shadowing effectively (e.g. to unwrap constrained types).
+            rust("""let $local = &input.$memberName;""")
+            val memberSymbol = symbolProvider.toSymbol(memberShape)
+            // If a header is of a primitive type and required (e.g. `bool`), we do not serialize it on the
```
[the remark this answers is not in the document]
Velfi: We're going to stop doing this when we implement support for IDL v2. Once we do that, we'll send defaults unless users have explicitly unset a value that normally has a default.

I've mentioned the practice of not sending default to the smithy team and they've never contradicted the assertion. I don't know why we do it, only that we were expected to in IDL v1. I'm assuming it has to do with the concept of ["boxing" primitives in IDL v1](https://smithy.io/1.0/spec/core/type-refinement-traits.html#box-trait) where fields with no default wouldn't have a set value.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -334,22 +346,35 @@ class JsonSerializerGenerator(
...
-            }
-        } else {
-            with(serializerUtil) {
-                ignoreZeroValues(context.shape, context.valueExpression) {
-                    serializeMemberValue(context, targetShape)
+            } else {
+                // This variable assignment ensures that the name of the value expression is always a valid variable name
+                // instead of being, for example, a field access expression (e.g. `my_struct.my_field`).
+                // This allows customization to leverage variable shadowing effectively (e.g. to unwrap constrained types).
+                rust("""let $local = ${context.valueExpression.asRef()};""")
```
Velfi, line 367:
```suggestion
                rust("let $local = ${context.valueExpression.asRef()};")
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/SerializerUtil.kt
Velfi: Some extra context:
We started [always serializing required query params](https://github.com/awslabs/smithy-rs/pull/1973) after an issue related to S3's multipart upload operations.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -334,22 +346,36 @@ class JsonSerializerGenerator(
 
     private fun RustWriter.serializeMember(context: MemberContext) {
         val targetShape = model.expectShape(context.shape.target)
-        if (symbolProvider.toSymbol(context.shape).isOptional()) {
-            safeName().also { local ->
+        safeName().also { local ->
```
david-perez, line 349: nit: the unconditional call to `safeName()` here is increasing the counter regardless of whether `local` is used or not, hence why sometimes the codegen diff is showing that we "skip" a variable name.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
LukeMathWalker: Does this work? https://github.com/awslabs/smithy-rs/pull/2005/commits/3160a92c17828bceb680b40c7e39c715d512a791

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
LukeMathWalker: [7d4d59f](https://github.com/awslabs/smithy-rs/pull/2005/commits/7d4d59fad23f712f7c37ef6582b09e7330717eaf) should fix it 👍🏻

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
LukeMathWalker: No clue 👀 
This is used on the deserialization side of the shop, so I've avoided touching it in this PR (apart from the automated formatting).

PR #2010, opened 2022-11-18
Re-export service from root of the generated crate

hlbarber:
## Motivation and Context

As the central API for the `smithy-rs` it should be exported from the root of the generated crate.

## Description

- Re-export the service from the root of the generated crate.

## Notes

It's `doc(hidden)` for now.


LukeMathWalker (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
```diff
@@ -40,6 +41,12 @@ open class ServerServiceGenerator(
      * which assigns a symbol location to each shape.
      */
     fun render() {
+        rustCrate.lib {
+            val serviceName = codegenContext.serviceShape.id.name.toString()
```
jjant, line 45: Is this safe to use as a rust name? i.e, if `serviceName` was a rust keyword (`mod`, `async`, etc).

hlbarber: We also do this in `ServerProtocolTestGenerator` and `ServerServiceGeneratorV2`.

IIRC our symbol providers should map shapes in a way which considers reserved keywords, however we've not got a symbol provider which maps `ServiceShape`s at the moment.

https://github.com/awslabs/smithy-rs/issues/1724

PR #2012, opened 2022-11-21
Add CHANGELOG support for multiple reference authors

hlbarber:
## Motivation and Context

Currently only a single author can be provided to a changelog entry. This does not support the cases where a single PR was coauthored by multiple authors and/or multiple PRs were authored by different individuals.

## Description

- Accept a list of strings, in addition to a single string in the `author` field. Each author in the list is added as a contributor to every PR provided in references.
- Accept references of the form `{ "id": <ID>, "authors": <Author> }`, in addition to `ID`. Each author given in `authors` is added as a contributor to that PR. Authors common to all references will be serialized in the same way as using the top-level `author` field.

## Notes

- This should not be a breaking change. Existing `CHANGELOG.next.toml` should be deserialized and then serialized identically.


hlbarber:
Should we remove the top-level `author` field eventually?

Velfi (approved):

tools/smithy-rs-tool-common/src/changelog/authors.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+impl Authors {
+    pub fn iter(&self) -> impl Iterator<Item = &String> {
+        self.0.iter()
+    }
+
+    // Checks whether the number of authors is 0 or any author has a empty name.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty() || self.iter().fold(false, |acc, x| x.is_empty() | acc)
```
jdisanti, line 41: `Iterator` has an `any` method to simplify the `fold` operation: https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.any
```suggestion
        self.0.is_empty() || self.iter().any(String::is_empty)
```

jdisanti (approved):

jdisanti:
> Should we remove the top-level `author` field eventually?

We could check the git history, but this will only work for the smithy-rs changelog and not for aws-sdk-rust. I think it's probably not worth the effort.

tools/smithy-rs-tool-common/src/changelog/authors.rs
hlbarber: Good call

hlbarber:
> We could check the git history, but this will only work for the smithy-rs changelog and not for aws-sdk-rust. I think it's probably not worth the effort.

To clarify, by "remove top-level author" I mean remove the `author` field from the root of the format, but keep it settable new position next to the references. This is distinct from automating the insertion of `author` data.

jdisanti:
> To clarify, by "remove top-level author" I mean remove the `author` field from the root of the format, but keep it settable new position next to the references. This is independent from automating the insertion of `author` data.

Oh, I'm fine with either approach.

Velfi:

rcoh (approved):
LGTM. While your here, you could consider making the `@name` references actually link to that users profile page

jdisanti:
> LGTM. While your here, you could consider making the @name references actually link to that users profile page

They already do when it's rendered in GitHub.

PR #2018, opened 2022-11-21
Normalize URI paths during signing (except for S3)

ysaito1001:
~~**Currently a draft PR to wash out test failures in CI.**~~

## Motivation and Context
Fixes https://github.com/awslabs/smithy-rs/issues/2042.

## Description
This PR adds the functionality for normalizing URI paths as described [here](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html). As documented, however, URI path normalization will be ignored for requests against S3.

Due to a difference in the scope, https://github.com/awslabs/smithy-rs/issues/1308, https://github.com/awslabs/aws-sdk-rust/issues/661 will be addressed in a separate PR.

## Testing
- [x] Added new unit tests for `normalize_uri_path`.
- [x] Will add an integration test for S3, verifying that URI path normalization does NOT happen.
- [x] Passed all tests in CI.

## Checklist

----


david-perez:
I think this should only be done for the services that explicitly require it e.g. S3. Stripping out leading forward slashes _could_ be in violation of the spec. I say _could_ because I raised this with the Smithy team in https://github.com/awslabs/smithy/issues/1024, and no conclusion was made.

Note that if we strip out leading forward slashes we may be stripping out empty path segments that could be bound to string members.

For example, consider the model:


```
operation Operation {
    input: OperationInput
}

@http(uri: "/{label}/suffix", method: "GET")
structure OperationInput {
    @required
    @httpLabel
    label: String,
}
```

If I send the request `GET //suffix`, I'm binding the empty string `""` to `label`. If we strip out leading forward slashes, I have no way of binding the empty string `""` to `label`.

In the server SDK, we're not stripping out empty segments from incoming requests.

CHANGELOG.next.toml
```diff
@@ -485,3 +485,9 @@ message = "Make generated enum `values()` functions callable in const contexts."
 references = ["smithy-rs#2011"]
 meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "all" }
 author = "lsr0"
+
+[[aws-sdk-rust]]
+message = "Normalize URI paths per RFC3986 when constructing canonical requests, except for S3."
+references = ["smithy-rs#2018"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
```
jdisanti, line 492: What is breaking about this? Seems more like a bug fix?

jdisanti (commented):
The overall approach looks good to me, and I think this should work as is. Most of my comments are for implementation details on `normalize_path_segment`.

aws/rust-runtime/aws-sigv4/src/http_request/uri_path_normalization.rs
```diff
@@ -0,0 +1,242 @@
...
+                    c.to_str()
+                        .expect("component in URI path should be valid Unicode")
+                        .to_owned(),
+                );
+            }
+        }
+    }
+
+    let mut result = result.join("/");
+    if uri_path.starts_with('/') {
```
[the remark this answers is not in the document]
ysaito1001: True. I was overly defensive in thinking that "what if `normalize_uri_path` is used in a different context where `uri_path` does not always start with a `/`?", but we can unconditionally prepend a `/`. When this assumption breaks, the unit tests will catch it.

CHANGELOG.next.toml
ysaito1001: Oh my bad. The code prior to the PR wasn't working as intended, so there is nothing to break. It is a bug fix.

aws/rust-runtime/aws-sigv4/src/http_request/uri_path_normalization.rs
```diff
@@ -0,0 +1,242 @@
...
+        result.insert(0, '/');
+    }
+    if ends_with_slash(uri_path) && result != "/" {
```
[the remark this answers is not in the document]
ysaito1001: I should've been clearer. Explained it in the comment added in a02771e.

aws/rust-runtime/aws-sigv4/src/http_request/uri_path_normalization.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    // The canonical URI is the URI-encoded version of the _absolute_ path component of the URI.
+    let result = if uri_path.starts_with('/') {
+        Cow::Borrowed(uri_path)
+    } else {
+        Cow::Owned(format!("/{uri_path}"))
+    };
+
+    if !result.contains('.') && !result.contains("..") {
```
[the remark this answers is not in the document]
ysaito1001: Good point. There are indeed some false positives where the code execution still reaches `normalize_path_segment` when we do not need to normalize a path. This check is just a basic one. Removed the second condition in a02771e.

aws/rust-runtime/aws-sigv4/src/http_request/uri_path_normalization.rs
```diff
@@ -0,0 +1,242 @@
...
+        return result;
+    }
+
+    Cow::Owned(normalize_path_segment(&result))
+}
+
+// Implement 5.2.4. Remove Dot Segments in https://www.rfc-editor.org/rfc/rfc3986
+fn normalize_path_segment(uri_path: &str) -> String {
+    let path = Path::new(uri_path);
```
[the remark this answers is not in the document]
ysaito1001: Agreed. The `Path` part is addressed in a02771e and preallocating the capacity of a vec is addressed in 938437d.

aws/rust-runtime/aws-sigv4/src/http_request/uri_path_normalization.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    let mut result = normalized.join("/");
+
+    // Even though `uri_path` starts with a `/`, that may not be the case for `result`.
+    // An example of this is `uri_path` being "/../foo" where the corresponding `result`
+    // will be "foo".
+    if !result.starts_with('/') {
+        result.insert(0, '/');
+    }
```
jdisanti, line 55: I think `result` will never start with `/` in this code, so the `result.insert` will always run.

ysaito1001: Interestingly, `result` starts with `/` most of the time. For instance, if the input to the function is `/a/b/c`, then the vector `normalized` will contain `["", "a", "b", "c"]`. After `normalized.join`, `result` will then be `/a/b/c`.

aws/rust-runtime/aws-sigv4/src/http_request/settings.rs
```diff
@@ -61,6 +64,21 @@ pub enum PercentEncodingMode {
     Single,
 }
 
+/// Config value to specify whether uri path normalization should be performed when signing.
+/// <https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html>
+///
+/// Uri path normalization is performed based on <https://www.rfc-editor.org/rfc/rfc3986>.
+/// The exception to this is that we do not normalize URI paths for requests to Amazon S3.
```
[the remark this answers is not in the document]
ysaito1001: 👍 Addressed in 182729a

aws/rust-runtime/aws-sigv4/src/http_request/settings.rs
```diff
@@ -61,6 +64,20 @@ pub enum PercentEncodingMode {
...
 }
 
+/// Config value to specify whether URI path normalization should be performed when signing.
+/// <https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html>
+///
+/// URI path normalization is performed based on <https://www.rfc-editor.org/rfc/rfc3986>.
+#[non_exhaustive]
+#[derive(Debug, Eq, PartialEq)]
+pub enum UriPathNormalizationMode {
+    /// We normalize URI paths according to the RFC
```
jdisanti, line 74:
```suggestion
    /// Normalize the URI path according to RFC3986
```

aws/rust-runtime/aws-sigv4/src/http_request/settings.rs
```diff
@@ -61,6 +64,20 @@ pub enum PercentEncodingMode {
...
+    Enabled,
+
+    /// The exception to this is that we do not normalize URI paths for requests to Amazon S3
```
jdisanti, line 77: Rephrased so that S3 is one example instead of _the_ example (there may be others we're unaware of right now):
```suggestion
    /// Don't normalize the URI path (S3, for example, rejects normalized paths in some instances)
```

aws/rust-runtime/aws-sigv4/src/http_request/settings.rs
```diff
@@ -61,6 +64,20 @@ pub enum PercentEncodingMode {
     Single,
 }
 
+/// Config value to specify whether URI path normalization should be performed when signing.
```
jdisanti, line 67:
```suggestion
/// Config value to specify whether the canonical request's URI path should be normalized.
```

aws/rust-runtime/aws-sigv4/src/http_request/uri_path_normalization.rs
jdisanti: Ah, yeah, you're right. I forgot about the leading `/` causing a `""` in the split.

jdisanti (approved):
Looks good! Just some suggestions on doc comment wording.

aws/rust-runtime/aws-sigv4/src/http_request/settings.rs
ysaito1001: Thanks! I like the updated for clarity.

PR #2022, opened 2022-11-22
add: RFC for Supporting `SelectObjectContent`'s unique `ScanRange` behavior

Velfi:
## Motivation and Context

[aws-smithy-sdk#630](https://github.com/awslabs/aws-sdk-rust/issues/630)

---

### DIFF
--- a/design/src/rfcs/rfc0024_supporting_select_object_content_scan_range.md
+++ b/design/src/rfcs/rfc0024_supporting_select_object_content_scan_range.md
@@ -1,0 +1,214 @@
+RFC: Supporting `SelectObjectContent`'s unique `ScanRange` behavior
+=============
+
+> Status: RFC
+>
+> Applies to: AWS client
+
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+The Rust SDK does not currently support all the documented behavior for the `ScanRange` input field of the
+`SelectObjectContent` operation. This RFC defines why and how we should support this behavior.
+
+Terminology
+-----------
+
+- [**Smithy IDL v2**][IDLv2]: The latest version of the Smithy modeling language. AWS services are defined with Smithy,
+  and we use it to generate the SDKs.
+- [**The `@default` trait**][default-trait]: A Smithy trait that allows modelers to define a default value for a struct member.
+- **Service Customizations**: Customizations modify the process of code generation, allowing for the inserting code into
+  pre-defined locations in the `config`, `operation`, or other modules. Service customizations are customization that
+  are specific to a single AWS service.
+- [**SelectObjectContent**][SelectObjectContent]: This action filters the contents of an Amazon S3 object based on a
+  simple structured query language (SQL) statement.
+- [**ScanRange**][ScanRange]: Used as an input to `SelectObjectContent`. Specifies the byte range of the object to get
+  the records from. Its functioning is based on [RFC 2616, Section 14.35.1].
+- **unset / unsetting**: The act of explicitly setting a value, which normally has a default, to be `None`.
+
+Why this RFC should be accepted
+-------------------------------
+
+There is currently one [user-submitted issue] in the `aws-sdk-rust` asking us to implement this behavior.
+
+TODO figure out what other SDKs support this behavior
+
+The user experience if this RFC is implemented
+----------------------------------------------
+
+### In Smithy IDLv1 (the current implementation)
+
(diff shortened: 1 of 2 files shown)


design/src/rfcs/rfc0024_supporting_select_object_content_scan_range.md
```diff
@@ -0,0 +1,214 @@
...
+There is currently one [user-submitted issue] in the `aws-sdk-rust` asking us to implement this behavior.
+
+TODO figure out what other SDKs support this behavior
+
+The user experience if this RFC is implemented
+----------------------------------------------
+
+### In Smithy IDLv1 (the current implementation)
+
+SDKs built from IDLv1 smithy models intentionally omit values from a request if a field has a default and the
```
jdisanti, line 40:
```suggestion
SDKs built from IDLv1 Smithy models intentionally omit values from a request if a field has a default and the
```

design/src/rfcs/rfc0024_supporting_select_object_content_scan_range.md
```diff
@@ -0,0 +1,214 @@
...
+the _"one less than the size of the object being queried"_ behavior that the docs say `end` is meant to default to.
+
+**Once this RFC is implemented,** users will finally be able to use the `ScanRange` input as documented:
+- Default values will be equal to those specified in the `ScanRange` docs.
+    - For `start`, the default value is zero.
+    - For `end`, the default value is one less than the size of the object being queried.
+- If `start` is set and `end` is "unset", the object will be scanned from `start` to then end of the file.
+- If `end` is set and `start` is "unset", the last `N` bytes of the object will be read, where `N == end`.
+- If both fields are set, the object will be scanned from `start` to `end`
+- If both fields are "unset", an error will be returned because at least one must be set.
```
jdisanti, line 122:
```suggestion
- If both fields are "unset", S3 will respond with an error because at least one must be set.
```

design/src/rfcs/rfc0024_supporting_select_object_content_scan_range.md
```diff
@@ -0,0 +1,214 @@
...
+}
+```
+
+To ensure that this handwritten struct doesn't deviate from the S3 model, we'd create a codegen test that would verify
+compatibility of the two models.
+
+Why this RFC should be rejected
+-------------------------------
+
+There are several reasons to reject this RFC:
```
jdisanti, line 187: Another reason: In the unlikely event that S3 modifies `ScanRange`, the handwritten implementation will be out of sync.

design/src/rfcs/rfc0024_supporting_select_object_content_scan_range.md
```diff
@@ -0,0 +1,214 @@
...
+pub enum ScanRange {
+    StartToEnd {
+        start: i64,
+        end: i64,
+    },
+    StartToEndOfObject {
+        start: i64,
+    },
+    LastNBytesOfObject {
+      n_bytes: i64
```
jdisanti, line 146:
```suggestion
        n_bytes: i64
```

design/src/rfcs/rfc0024_supporting_select_object_content_scan_range.md
```diff
@@ -0,0 +1,214 @@
...
+  we write won't necessarily be implemented by other SDKs, causing a fragmented and annoying user experience.
+- The proposed behavior is confusing and difficult to explain.
+- Differentiating between "set", "default", and "unset" struct fields is not currently supported by the SDK. Adding it
+  would take a few weeks of effort. _(We do need to eventually support this anyway for IDL v2.)_
+- [Only one user][user-submitted issue] has asked for this feature at the time of writing this RFC.
+
+### What we should do instead
+
+We have already committed to supporting [IDLv2] and implementing that would allow users to set `start=0` or unset it.
+Although IDLv2 support would NOT support the custom `end` value behavior, we'd be halfway there. Any future S3
```
jdisanti, line 199: This is a little confusing because it seemed like the handwritten `ScanRange` implementation was to be applied to the IDLv2 implementation. But this sentence makes it seem like the `ScanRange` proposal is actually for an IDLv1 implementation?

jdisanti (commented):
Nicely written RFC! Seems to me like we should just implement IDLv2 default support, and then maybe we can consider adding a hook for custom methods/fields and serialization hooks on code generated structs to handle the "last n-bytes" case later.

Velfi:
I'm closing this since we've decided to handle this issue by implementing IDLv2 support instead.

PR #2025, opened 2022-11-23
Python: Use Maturin to build SDKs

unexge:
## Motivation and Context
Currently we are using bare extension module from PyO3 to build our SDKs and as a result we are getting bare shared objects. On the other hand if we use [Maturin](https://github.com/PyO3/maturin) we will get Python Wheels as a result and besides being more idiomatic than bare shared objects it will also allows us to distribute some Python files alongside our shared objects which is required for tasks like: generating better documentation, generating type stubs etc.

## Testing
You can build Pokemon service using Maturin:
```bash
$ cd rust-runtime/aws-smithy-http-server-python/examples
$ make build
$ ls -lh wheels
total 8.1M
-rw-r--r-- 1 _ _ 8.0M Nov 23 13:35 pokemon_service_server_sdk-0.0.1-cp310-cp310-macosx_11_0_arm64.whl
$ python3 -c 'import pokemon_service_server_sdk' # Make sure it is importable
```

----


rust-runtime/aws-smithy-http-server-python/examples/Makefile
```diff
@@ -1,56 +1,65 @@
...
+# Note on `--compatibility linux`: Maturin by default uses `manylinux_x_y` but it is not supported 
+# by our current CI version (3.7.10), we can drop `--compatibility linux` when we switch to higher Python version.
+# For more detail: https://github.com/pypa/manylinux
+build-wheel: ensure-maturin
+	maturin build --manifest-path $(SERVER_SDK_DST)/Cargo.toml --out $(WHEELS) --compatibility linux
 
-py_check: build
-	mypy pokemon_service.py
+build-wheel-release: ensure-maturin
+	maturin build --manifest-path $(SERVER_SDK_DST)/Cargo.toml --out $(WHEELS) --compatibility linux --release
```
crisidev, line 32: Don't we need to check if we are on mac and add the compatibily for it?

rust-runtime/aws-smithy-http-server-python/examples/libpokemon_service_server_sdk.pyi
```diff
@@ -1,20 +0,0 @@
-#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
crisidev, line 1: Out of curiosity, why is this file gone?

rust-runtime/aws-smithy-http-server-python/examples/Makefile
unexge: `compatibility` is only needed for Linux and it is ignored on non-Linux platforms:
```
      --compatibility [<compatibility>...]
          Control the platform tag on linux.

          Options are `manylinux` tags (for example `manylinux2014`/`manylinux_2_24`) or `musllinux` tags (for example `musllinux_1_2`) and `linux` for the native linux tag.

          Note that `manylinux1` and `manylinux2010` is unsupported by the rust compiler. Wheels with the native `linux` tag will be rejected by pypi, unless they are separately validated by `auditwheel`.

          The default is the lowest compatible `manylinux` tag, or plain `linux` if nothing matched

          This option is ignored on all non-linux platforms
```

crisidev: Thanks got it.

crisidev (approved):

rust-runtime/aws-smithy-http-server-python/examples/Makefile
```diff
@@ -1,56 +1,65 @@
...
-	RELEASE_SHARED_LIBRARY_SRC := $(SRC_DIR)/target/release/libpokemon_service_server_sdk.so
-endif
+HAS_MATURIN := $(shell command -v maturin 2> /dev/null)
 
 all: codegen
 
 codegen:
 	$(GRADLE) --project-dir $(SRC_DIR) -P modules='pokemon-service-server-sdk,pokemon-service-client' :codegen-client-test:assemble :codegen-server-test:python:assemble
-	mkdir -p $(SERVER_SDK_DST) $(CLIENT_SDK_DST)
+	mkdir -p $(SERVER_SDK_DST) $(CLIENT_SDK_DST) $(WHEELS)
```
chadac, line 16: Should the wheels dir be created in the `ensure-maturin` task? Or should `build-wheel` and `build-wheel-release` have dependencies on `codegen`? Right now it seems like `make build-wheel` would fail if `make codegen` or `make all` isn't run first.

chadac (changes requested):

rust-runtime/aws-smithy-http-server-python/examples/Makefile
unexge: Good catch! Addressed in https://github.com/awslabs/smithy-rs/pull/2025/commits/dd290a852129cf074664a6c969f87e9bb8a768a6

chadac (approved):

PR #2036, opened 2022-11-29
Implement @range on long and byte shapes

### LINKED ISSUE #1968: `range` on `byte`, `short`, `integer`, `long`, `float`, and `double` shapes.

LukeMathWalker:
Closes #1968 

## Checklist

----


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstrainedShapeSymbolProvider.kt
```diff
@@ -98,10 +100,11 @@ class ConstrainedShapeSymbolProvider(
                 }
             }
 
-            is StringShape, is IntegerShape, is ShortShape -> {
+            is StringShape, is IntegerShape, is ShortShape, is LongShape, is ByteShape -> {
                 if (shape.isDirectlyConstrained(base)) {
                     val rustType = RustType.Opaque(shape.contextName(serviceShape).toPascalCase())
                     symbolBuilder(shape, rustType).locatedIn(ModelsModule).build()
+                    symbolBuilder(shape, rustType).locatedIn(ModelsModule).build()
```
jjant, line 107: Was this duplicated on accident?

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedNumberGeneratorTest.kt
```diff
@@ -116,6 +121,8 @@ class ConstrainedNumberGeneratorTest {
             listOf(
                 Triple("integer", "ConstrainedInteger", "i32"),
                 Triple("short", "ConstrainedShort", "i16"),
+                Triple("long", "ConstrainedLong", "i64"),
+                Triple("byte", "ConstrainedByte", "i8"),
```
jjant, line 125: Nit: could you order these from `i8` to `i64`?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ShapesReachableFromOperationInputTagger.kt
```diff
@@ -52,7 +54,7 @@ object ShapesReachableFromOperationInputTagger {
 
         return ModelTransformer.create().mapShapes(model) { shape ->
             when (shape) {
-                is StructureShape, is UnionShape, is ListShape, is MapShape, is StringShape, is IntegerShape, is ShortShape -> {
+                is StructureShape, is UnionShape, is ListShape, is MapShape, is StringShape, is IntegerShape, is ShortShape, is LongShape, is ByteShape -> {
                     if (shapesReachableFromOperationInputs.contains(shape)) {
                         val builder = when (shape) {
```
jjant, line 59: Do we need this `when` at all? It seems that it covers the same cases as the encompassing `is` branch.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/http/ServerResponseBindingGenerator.kt
```diff
@@ -92,7 +94,7 @@ class ServerResponseBeforeRenderingHeadersHttpBindingCustomization(val codegenCo
                     codegenContext.settings.codegenConfig.publicConstrainedTypes,
                 )
             ) {
-                if (section.context.shape is IntegerShape || section.context.shape is ShortShape) {
+                if (section.context.shape is IntegerShape || section.context.shape is ShortShape || section.context.shape is LongShape || section.context.shape is ByteShape) {
```
jjant, line 97: 💡 Maybe we should move this check to a `NumberShape.isIntegral` function?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ValidateUnsupportedConstraints.kt
```diff
@@ -241,12 +243,12 @@ fun validateUnsupportedConstraints(
...
+    // 5. Range trait used on unsupported shapes.
     // TODO(https://github.com/awslabs/smithy-rs/issues/1401)
     val unsupportedRangeTraitOnShapeSet = walker
         .walkShapes(service)
         .asSequence()
-        .filterNot { it is IntegerShape || it is ShortShape }
+        .filterNot { it is IntegerShape || it is ShortShape || it is LongShape || it is ByteShape }
         .filterMapShapesToTraits(setOf(RangeTrait::class.java))
         .map { (shape, rangeTrait) -> UnsupportedRangeTraitOnShape(shape, rangeTrait as RangeTrait) }
         .toSet()
```
jjant, line 254: Do we still need this? I think `@range` on other shapes is rejected by smithy.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
```diff
@@ -372,6 +376,24 @@ open class ServerCodegenVisitor(
...
     }
 
+    override fun longShape(shape: LongShape) {
+        if (shape.isDirectlyConstrained(codegenContext.symbolProvider)) {
+            logger.info("[rust-server-codegen] Generating a constrained long $shape")
+            rustCrate.withModule(ModelsModule) {
+                ConstrainedLongGenerator(codegenContext, this, shape).render()
+            }
+        }
+    }
```
jjant, line 386: Out of ignorance: Don't we need an `else { super(shape) }` branch? Same for `byteShape`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedNumberGenerator.kt
```diff
@@ -57,6 +59,30 @@ class ConstrainedShortGenerator(
...
+    val codegenContext: ServerCodegenContext,
+    val writer: RustWriter,
+    val shape: ByteShape,
+) {
+    val inner = ConstrainedNumberGenerator(codegenContext, writer, shape, RustType.Integer(8))
+
+    fun render() {
+        inner.render()
+    }
+}
```
jjant, line 84: It feels like these 4 could be DRY'd up a bit more, but all I can see is using inheritance, which I'm not a fan of.

hlbarber (approved):
LGTM

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ValidateUnsupportedConstraints.kt
LukeMathWalker: `@range` is allowed on floats and doubles, according to the spec.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
LukeMathWalker: I don't think so? 
The default handler is a no-op for all types.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ShapesReachableFromOperationInputTagger.kt
LukeMathWalker: I think it's necessary for type refinement, otherwise you have no way to call `toBuilder`.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedNumberGeneratorTest.kt
LukeMathWalker: I had an internal bet going on about receiving a comment on this 😛

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstrainedShapeSymbolProvider.kt
LukeMathWalker: Yes! Good catch ❤️

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/http/ServerResponseBindingGenerator.kt
LukeMathWalker: I'd prefer to add an `IntegralShape` class in the hierarchy if we want to dry some of these up further.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ShapesReachableFromOperationInputTagger.kt
jjant: You're right, I thought `toBuilder()` was defined on `Shape`. My bad!

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedNumberGeneratorTest.kt
jjant: Sorry :grimacing:

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedNumberGenerator.kt
jjant: Lovely

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ValidateUnsupportedConstraints.kt
jjant: You're right (and probably `bigDecimal` and `bigInteger`).
I'd suggest using `filter  { it is FloatShape || it is DoubleShape || .. }` instead of the negative as it makes it easier to see which are missing (but this is just a nit).

jjant (approved):
LGTM!

So many more protocol tests passing 😄

PR #2037, opened 2022-11-29
Switch links to Smithy v2 documentation

hlbarber:
## Motivation and Context

The latest documentation is Smithy v2. In most cases it's identical. As far as I can tell, there are no links to items with breaking changes between v1 and v2.


Velfi:
@hlbarber Do you think this would cause any users to incorrectly assume that we support IDL v2? If you do then I'm thinking we should hold off until we support:
- [ ] The enum stuff (I think we already added this and I just forgot)
- [ ] Nullability
- [ ] Presence tracking

I suppose that the cost of that confusion isn't very high though. They'd just be a little sad.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationRegistryGenerator.kt
```diff
@@ -151,8 +151,8 @@ ${operationImplementationStubs(operations)}
 ///
 /// [`serve`]: https://docs.rs/hyper/0.14.16/hyper/server/struct.Builder.html##method.serve
 /// [`tower::make::MakeService`]: https://docs.rs/tower/latest/tower/make/trait.MakeService.html
-/// [HTTP binding traits]: https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html
-/// [operations]: https://awslabs.github.io/smithy/1.0/spec/core/model.html##operation
+/// [HTTP binding traits]: https://smithy.io/2.0/spec/http-bindings.html
+/// [operations]: https://smithy.io/2.0/spec/service-types.html#operation
```
Velfi, line 155: I think you're missing an extra `#` here:
```suggestion
/// [operations]: https://smithy.io/2.0/spec/service-types.html##operation
```
else the formatter will get confused

hlbarber:
> @hlbarber Do you think this would cause any users to incorrectly assume that we support IDL v2?

Yeah, probably. I'll put this in draft and we can resurrect it when we're ready.

jdisanti (approved):

82marbag (approved):

PR #2038, opened 2022-11-29
Add lambda extractors

hlbarber:
## Motivation and Context

Closes https://github.com/awslabs/smithy-rs/issues/2023

## Description

Add `FromParts` implementation for `ApiGatewayProxyRequestContext`, `ApiGatewayV2httpRequestContext`, and `Context`.


rust-runtime/aws-smithy-http-server/src/request/lambda.rs
```diff
@@ -0,0 +1,49 @@
...
+    aws_lambda_events::apigw::{ApiGatewayProxyRequestContext, ApiGatewayV2httpRequestContext},
+    Context,
+};
+
+use crate::Extension;
+
+use super::{extension::MissingExtension, FromParts};
+
+impl<P> FromParts<P> for Context {
+    type Rejection = MissingExtension;
```
hlbarber, line 17: In each of these cases we're re-using the `MissingExtension` rejection. Do we want rejections tailored specifically to `Context` etc?

rust-runtime/aws-smithy-http-server/src/request/lambda.rs
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! The [`lambda_http`] types included in [`http::Request`]s when [`LambdaHandler`](crate::routing::LambdaHandler) is
+//! used. Each are given a [`FromParts`] implementation for easy use within handlers.
+
+use lambda_http::request::RequestContext;
+pub use lambda_http::{
```
hlbarber, line 10: `#[doc(inline)]` these?

rust-runtime/aws-smithy-http-server/src/request/lambda.rs
LukeMathWalker: I'd suggest a little breaking change to `MissingExtension` - allow it to hold a private `'static str` field which we can use to provide a good nudge. For example, in this case, tell the developer to make sure they are using `LambdaHandler`.

hlbarber: Will whip up a subsequent PR to make this story coherent across all extractors.

LukeMathWalker (approved):
Good baseline, we can refine the errors in a separate PR 👌🏻

david-perez (approved):
Nice. We'll need examples showcasing these.

hlbarber:
> We'll need examples showcasing these.

Yeah we're in a catch 22 at the moment because the current service builder doesn't support `FromParts` so we can't add it to the examples until we migrate. I plan to add it to the lambda example in https://github.com/awslabs/smithy-rs/pull/1886.

PR #2040, opened 2022-11-30
RFC: Better Constraint Violations

david-perez:
[Rendered](https://github.com/awslabs/smithy-rs/blob/de01adac828fb9afec82edfee493f64889ca0f91/design/src/rfcs/rfc0031_better_constraint_violations.md).

----


design/src/rfcs/rfc0024_better_constraint_violations.md
```diff
@@ -0,0 +1,509 @@
...
+  * In general, it's unclear what constraint traits should cause
+    short-circuiting. A probably reasonable rule of thumb is to include
+    traits that can be attached directly to aggregate shapes: as of writing,
+    that would be `@uniqueItems` on list shapes and `@length` on list shapes.
+- Another possiblity is to _do nothing_ and value _complete_ validation
+  exception response messages over trying to mitigate this with special
+  handling. One could argue that these kind of DOS attack vectors should be
+  taken care of with a separate solution e.g. a layer that bounds a request
+  body's size to a reasonable default (see [how Axum added
+  this](https://github.com/tokio-rs/axum/pull/1420)). We will provide a similar
```
david-perez, line 508: A related problem is a stack overflow in deserialization for a sufficiently nested shape. Unless an attacker controls the Smithy model, it's not really an attack vector, but we should probably implement a check at codegen time that aborts if a shape's depth exceeds a reasonable threshold. Said check could probably live in awslabs/smithy.

design/src/rfcs/rfc0024_better_constraint_violations.md
```diff
@@ -0,0 +1,509 @@
...
+
+We will analogously introduce a `ConstraintViolationExceptions` type that
+represents a collection of constraint violations that can occur _within the
+framework's request deserialization code_. This type will be `pub(crate)` and
+will be the one the framework will map to Smithy's `ValidationException` that
+eventually gets serialized into the response.
+
+### Unresolved questions
+
+#### Collecting constraint violations may constitute a DOS attack vector
```
hlbarber, line 472: Some comments to pushing back against https://github.com/awslabs/smithy-rs/pull/1199#discussion_r809300673.

Shouldn't clients be in the business of validating the requests they provide?

- Web forms can do it using a client side script - this is better UX because they get an immediate response.
- Typed languages can do it via fallible request constructors - again, this is a better UX.

Currently, there are three logical phases:

1. Receive bytes into a buffer.
2. Deserialize the buffer into the unconstrained tree.
3. Convert the unconstrained tree into the constrained tree.

Phase (2) requires lots of allocations - copying bytes into `String` for example. Phase (3) also requires allocations, `into_iter()...collect()` dynamic collections and some extra CPU work in the case of `@pattern` etc.

If we don't return early on an error we incur all the allocations related to (2) for invalid requests. Validation doesn't protect the parser if we parse then validate, but if we "parse don't validate" (deserialize directly into the constrained tree) it does to some extent.

If we deserialize the buffer directly into the constrained tree we are also able to avoid the allocations associated with phase 3.

design/src/rfcs/rfc0024_better_constraint_violations.md
```diff
@@ -0,0 +1,509 @@
...
+list ValidationExceptionFieldList {
+    member: ValidationExceptionField
+}
+```
+
+It was mentioned in the [constraint traits
+RFC](https://github.com/awslabs/smithy-rs/pull/1199#discussion_r809300673), and
+implicit in the definition of Smithy's
+[`smithy.framework.ValidationException`](https://github.com/awslabs/smithy/blob/main/smithy-validation-model/model/smithy.framework.validation.smithy)
+shape, that server frameworks should respond with a _complete_ collection of
```
LukeMathWalker, line 343: I'm not sure I agree with this interpretation.
I think it's fair to say that the framework should return an error that is as informative as possible, but it doesn't necessarily have to be complete.

It's bit like compiler errors, to use an analogy - you don't want a compiler to short-circuit as soon as it finds the first error, otherwise you need to run through the fix-check loop several times. At the same time, you value fast feedback and you'd prefer the compiler to return in a reasonable timeframe with a set of relevant errors rather than taking forever to return a complete list.

Going back to Smithy, I think it would be sensible to put in place some upper bounds (e.g. 1000 elements) to prevent degenerate behaviour (or DOS attacks, as mentioned below) and guarantee a timely response.

design/src/rfcs/rfc0024_better_constraint_violations.md
```diff
@@ -0,0 +1,509 @@
...
+deserializers enforce constraint traits in a two-step phase: first, the
+_entirety_ of the unconstrained value is deserialized, _then_ constraint traits
+are enforced by feeding the entire value to the `TryFrom` constructor.
+
+We will introduce a `ConstraintViolations` type (note the plural) that
+represents a collection of constraint violations that can occur _within user
+application code_. Roughly:
+
+```rust
+pub ConstraintViolations<T>(pub(crate) Vec<T>);
```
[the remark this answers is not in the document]
david-perez: `T` is uniform. It's either the user-facing `shape_name::ConstraintViolation`, which only models violations arising from direct constraint traits, or the framework-exclusive `shape_name::ConstraintViolationException`, which has variants pointing to inner constraint violations, if any.

design/src/rfcs/rfc0024_better_constraint_violations.md
```diff
@@ -0,0 +1,509 @@
...
+
+impl<T> IntoIterator<Item = T> for ConstraintViolations<T> { ... }
+
+impl std::convert::TryFrom<std::string::String> for LengthPatternString {
+    type Error = ConstraintViolations<crate::model::length_patterh_string::ConstraintViolation>;
+
+    fn try_from(value: std::string::String) -> Result<Self, Self::Error> {
+        // Check constraints and collect violations.
```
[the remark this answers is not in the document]
david-perez: We collect but we do not flatten. We collect `shape_name::ConstraintViolation`s or `shape_name::ConstraintViolationException`s. The latter ones are the ones that point to inner constraint violations, if any.

The implementation should not over-allocate. `Vec::new()` does not allocate. Only when we `push` to it we will.

design/src/rfcs/rfc0024_better_constraint_violations.md
david-perez: > Shouldn't clients be in the business of validating the requests they provide?

Yes, but client-side validation is best effort and a backend service should not rely on it for data integrity.

---

I'm also unconvinced of the value of collecting constraint violations in the general use cases of smithy-rs server SDKs. But there's certainly particular use cases that need it. Another one that I failed to mention in the document is backwards compatibility. If a service has always returned a custom response with a _list_ of validation error messages, this feature and [custom validation error messages](https://smithy.io/ts-ssdk/validation.html#custom-validation) will enable them to move their service to use a smithy-rs sSDK.

Regarding deserializing directly into the constrained tree, that's a bigger architectural change that was already expressed. Deserializing directly into the constrained tree while collecting constraint violations at the same time is difficult. [My preference](https://github.com/awslabs/smithy-rs/pull/1342#discussion_r1012777985) is to explore it in a completely separate deserialization approach by implementing `serde` visitors.

hlbarber: > Yes, but client-side validation is best effort and a backend service should not rely on it for data integrity.

For sure, we shouldn't forgo server side validation and hope that clients behave. I'm proposing that server side validation returns early and client side validation can collect all errors if it improves UX for them.

I'd argue that this is the norm in most client-server interactions - if a server has a list of criteria that a request needs to meet then the client should expect an error message describing the first criteria which fails, rather than a list of all violations.

> Regarding deserializing directly into the constrained tree, that's a bigger architectural change that was already expressed. Deserializing directly into the constrained tree while collecting constraint violations at the same time is difficult. https://github.com/awslabs/smithy-rs/pull/1342#discussion_r1012777985 is to explore it in a completely separate deserialization approach by implementing serde visitors.

Agree, I'm mentioning it because I think "deserializing directly into the constrained tree" is easiest if we return errors early while "Deserializing directly into the constrained tree while collecting constraint violations" might be impossible within `serde`.

design/src/rfcs/rfc0024_better_constraint_violations.md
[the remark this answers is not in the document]
hlbarber: Supposing we had structure `((A, B), (C, D))` all with a constraints on them. If all constraints failed this would amount to something like `vec![vec![_, _], vec![_, _]]`? If we do flatten we'd get something like `vec![_, _, _, _]`? Both the flattened and unflattened approach three vectors. With the flattened vector requiring some extra work but being more cache friendly.

The alternative I'm thinking of is to only `Vec::new` once for the entire tree and instead push to a `&mut Vec<ValidationExceptionField>`.

design/src/rfcs/rfc0025_better_constraint_violations.md
```diff
@@ -0,0 +1,511 @@
...
+represents a collection of constraint violations that can occur _within user
+application code_. Roughly:
+
+```rust
+pub ConstraintViolations<T>(pub(crate) Vec<T>);
+
+impl<T> IntoIterator<Item = T> for ConstraintViolations<T> { ... }
+
+impl std::convert::TryFrom<std::string::String> for LengthPatternString {
+    type Error = ConstraintViolations<crate::model::length_patterh_string::ConstraintViolation>;
```
hlbarber, line 448: `length_patterh_string`

design/src/rfcs/rfc0025_better_constraint_violations.md
```diff
@@ -0,0 +1,511 @@
...
+        }
+    }
+}
+
+pub mod length_map {
+    pub enum ConstraintViolation {
+        Length(usize),
+        Value(
+            std::string::String,
+            crate::model::length_string::ConstraintViolation,
```
david-perez, line 179:
```suggestion
            crate::model::length_string::ConstraintViolationException,
```

design/src/rfcs/rfc0024_better_constraint_violations.md
```diff
@@ -0,0 +1,509 @@
...
+
+[this comment]: https://github.com/awslabs/smithy-rs/blob/27020be3421fb93e35692803f9a795f92feb1d19/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/MapConstraintViolationGenerator.kt#L66-L69
+
+### Solution proposal
+
+The problem can be mitigated by adding `#[doc(hidden)]` to the internal
+variants and `#[non_exhaustive]` to the enum. We're already doing this in some
+constraint violation types.
+
+However, a "less leaky" solution is achieved by _splitting_ the constraint
```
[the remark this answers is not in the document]
david-perez: Perhaps.

However, if we eagerly convert to `ValidationExceptionFieldList`, we'd be unable to map the `ConstraintViolationExceptions` to custom error messages, a feature some users have requested. I've just published [RFC: Custom Validation Exceptions](https://github.com/awslabs/smithy-rs/pull/2053) describing this feature.

design/src/rfcs/rfc0024_better_constraint_violations.md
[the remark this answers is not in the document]
david-perez: > The alternative I'm thinking of is to only Vec::new once for the entire tree and pass it to each conversion in the form of &mut Vec<ValidationExceptionField>.

Eagerly converting to `ValidationExceptionField`s has the downside that we'd be unable to map the constraint violation exceptions to other error shapes. Let's use [this thread](https://github.com/awslabs/smithy-rs/pull/2040#discussion_r1038169046).

design/src/rfcs/rfc0024_better_constraint_violations.md
```diff
@@ -0,0 +1,509 @@
...
+  having it bound the other validators via effectively permuting the order of
+  the checks and thus short-circuiting.
+  * In general, it's unclear what constraint traits should cause
+    short-circuiting. A probably reasonable rule of thumb is to include
+    traits that can be attached directly to aggregate shapes: as of writing,
+    that would be `@uniqueItems` on list shapes and `@length` on list shapes.
+- Another possiblity is to _do nothing_ and value _complete_ validation
+  exception response messages over trying to mitigate this with special
+  handling. One could argue that these kind of DOS attack vectors should be
+  taken care of with a separate solution e.g. a layer that bounds a request
```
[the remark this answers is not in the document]
crisidev: Good point Harry!

design/src/rfcs/rfc0025_better_constraint_violations.md
```diff
@@ -0,0 +1,511 @@
...
+the `@pattern` trait, and only _after_ is the `@length` trait checked. This
+means that if a client sends a request with `n >>> 3` list members, the
+expensive check runs `n` times, when a constant-time check inspecting the
+length of the input vector would have sufficed to reject the request.
+Additionally, we may want to avoid serializing `n` `ValidationExceptionField`s
+due to performance concerns.
+
+- A possibility to circumvent this is making the `@length` validator special,
+  having it bound the other validators via effectively permuting the order of
+  the checks and thus short-circuiting.
```
crisidev, line 500: I am not sure I understand why this applies only to `@length`. Shouldn't this issue apply to basically any of the constraints?

design/src/rfcs/rfc0025_better_constraint_violations.md
```diff
@@ -0,0 +1,511 @@
...
+    short-circuiting. A probably reasonable rule of thumb is to include
+    traits that can be attached directly to aggregate shapes: as of writing,
+    that would be `@uniqueItems` on list shapes and `@length` on list shapes.
+- Another possiblity is to _do nothing_ and value _complete_ validation
+  exception response messages over trying to mitigate this with special
+  handling. One could argue that these kind of DOS attack vectors should be
+  taken care of with a separate solution e.g. a layer that bounds a request
+  body's size to a reasonable default (see [how Axum added
+  this](https://github.com/tokio-rs/axum/pull/1420)). We will provide a similar
+  request body limiting mechanism regardless.
```
crisidev, line 511: I think the solution Axum is using is too simplistic for our usecase. We may have customers that may want a bigger body size or even a completely unconstrained one. 

We could find a decent default and make it configurable for advanced users..

design/src/rfcs/rfc0025_better_constraint_violations.md
```diff
@@ -0,0 +1,511 @@
...
+   constraint traits, setting its visibility to `pub (crate)`, for exclusive
+   use by the framework.
+2. [`ServerBuilderGeneratorWithoutPublicConstrainedTypes.kt`] renders the
+   builder type the user is exposed to: this builder does not take in
+   constrained types and does not enforce all modeled constraints.
+
+[`ServerBuilderGenerator.kt`]: https://github.com/awslabs/smithy-rs/blob/2226feff8f7fa884204f81a50d7e016386912acc/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
+[`ServerBuilderGeneratorWithoutPublicConstrainedTypes.kt`]: https://github.com/awslabs/smithy-rs/blob/2226feff8f7fa884204f81a50d7e016386912acc/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorWithoutPublicConstrainedTypes.kt
+
+Collecting constraint violations
```
crisidev, line 287: I am not sure this title convey properly what you are explaining below, but I don't have a better name on top of my mind.

design/src/rfcs/rfc0025_better_constraint_violations.md
```diff
@@ -0,0 +1,511 @@
...
+[this comment]: https://github.com/awslabs/smithy-rs/blob/27020be3421fb93e35692803f9a795f92feb1d19/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/MapConstraintViolationGenerator.kt#L66-L69
+
+### Solution proposal
+
+The problem can be mitigated by adding `#[doc(hidden)]` to the internal
+variants and `#[non_exhaustive]` to the enum. We're already doing this in some
+constraint violation types.
+
+However, a "less leaky" solution is achieved by _splitting_ the constraint
+violation type into two types:
```
crisidev, line 251: From this document I cannot really tell if you are proposing the mitigation or the more complex solution of splitting the constraints into two types. I believe the latter, but it's better to be more clear I think.

design/src/rfcs/rfc0025_better_constraint_violations.md
david-perez: I don't see why it should apply to any of the other constraints. As the document mentions, I think that `@uniqueItems` on list shapes and `@length` on list shapes are the only ones that should short-circuit. I think there is no general rule to determine whether a constraint should short-circuit or not; it has to be determined on a case by case basis.

design/src/rfcs/rfc0025_better_constraint_violations.md
david-perez: The document is proposing the "less leaky" solution, splitting the constraint violation type into two types. I'll rephrase.

design/src/rfcs/rfc0025_better_constraint_violations.md
david-perez: The RFC now advocates that we should add both solutions after the rewrite: https://github.com/awslabs/smithy-rs/pull/2040/commits/5033c7af05e7e1aa9fe2308cf52dd27fe0c3deee.

design/src/rfcs/rfc0024_better_constraint_violations.md
david-perez: The RFC now subscribes this opinion after the rewrite: https://github.com/awslabs/smithy-rs/pull/2040/commits/5033c7af05e7e1aa9fe2308cf52dd27fe0c3deee.

hlbarber (approved):

Issue #371, opened 2021-05-13
Add more tracing events


rcoh:
- [x] request
- [x] request body
- [x] response
- [x] response body

Separately, we should also enable tracing inside of sigv4:
- [x] canonical request
- [x] string to sign

Event Stream messages should also have trace logging:
- [x] Event Stream receive `Message`
- [x] Event Stream send `Message`

Make spans consistent and useful:
- [x] Run with `texray` to visually observe spans. Make better span names and make them more useful
- [x] Improve overall debuggability with spans

rcoh:
don't need to do all of them, but adding a couple would be a good tour of the code

PR #2044, opened 2022-12-01
Improve client tracing spans

jdisanti:
## Motivation and Context
This PR improves the tracing spans for clients (for #371).

### `tracing-texray` before:
```
root                                                                     356ms ├───────────────────────────────────────┤
  send_operation{operation: "ListBuckets", service: "s3", status: "ok"}  356ms ├───────────────────────────────────────┤
    provide_credentials{provider: default_chain}                         263μs ┆
      lazy_load_credentials                                              212μs ┆
        load_credentials{provider: Environment}                         1000ns ┆
        load_credentials{provider: Profile}                              166μs ┆
        load_credentials{provider: Profile}                              165μs ┆
          load_config_file{file: Default(Config)}                         42μs ┆
          load_config_file{file: Default(Credentials)}                    31μs ┆
          load_base_credentials                                         1000ns ┆
    load_response                                                        592μs                                          ┆
```

### `tracing-texray` after:
```
root                                                                     353ms ├───────────────────────────────────────┤
  send_operation{operation: "ListBuckets", service: "s3", status: "ok"}  352ms ├───────────────────────────────────────┤
    map_request{name: "resolve_endpoint"}                                 25μs ┆
    map_request{name: "generate_user_agent"}                              14μs ┆
    async_map_request{name: "retrieve_credentials"}                      290μs ┆
      provide_credentials{provider: default_chain}                       254μs ┆
        lazy_load_credentials                                            197μs ┆
          load_credentials{provider: Environment}                       1000ns ┆
          load_credentials{provider: Profile}                            136μs ┆
          load_credentials{provider: Profile}                            134μs ┆
            load_config_file{file: Default(Config)}                       53μs ┆
            load_config_file{file: Default(Credentials)}                  36μs ┆
            load_base_credentials                                       1000ns ┆
    map_request{name: "sigv4_sign_request"}                              104μs ┆
    map_request{name: "recursion_detection"}                               2μs ┆
    dispatch                                                             351ms ├───────────────────────────────────────┤
    load_response                                                        382μs                                          ┆
      parse_unloaded                                                       3μs                                          ┆
      read_body                                                          223μs                                          ┆
      parse_loaded                                                        99μs                                          ┆
```

## Checklist

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
```diff
@@ -160,7 +160,7 @@ class CustomizableOperationGenerator(
                 /// Sends this operation's request
                 pub async fn send<T, E>(self) -> Result<T, SdkError<E>>
                 where
-                    E: std::error::Error + 'static,
+                    E: std::error::Error + Send + Sync + 'static,
```
Velfi, line 163: Does it need both or just `Send`?

rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -226,7 +228,47 @@ where
             .layer(DispatchLayer::new())
             .service(connector);
 
-        check_send_sync(svc).ready().await?.call(input).await
+        // send_operation records the full request-response lifecycle.
+        // NOTE: For operations that stream output, only the setup is captured in this span.
```
Velfi, line 232: Would users want to capture the streaming stuff in this span? If yes, is there anything they can do?

rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -226,7 +228,47 @@ where
...
+            status = field::Empty,
+            message = field::Empty
+        );
+        let (mut req, parts) = op.into_request_response();
+        if let Some(metadata) = &parts.metadata {
+            span.record("operation", &metadata.name());
+            span.record("service", &metadata.service());
+            // This will clone two `Cow::<&'static str>::Borrow`s in the vast majority of cases
+            req.properties_mut().insert(metadata.clone());
+        }
```
Velfi, line 246: Is it worth adding a method to operation that exposes this metadata?

rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -226,7 +228,47 @@ where
...
+        let op = Operation::from_parts(req, parts);
+
+        let result = async move { check_send_sync(svc).ready().await?.call(op).await }
+            .instrument(span.clone())
+            .await;
```
Velfi, line 251: I understand why this is the way it is but I still think it looks a bit odd.

rust-runtime/aws-smithy-http/src/middleware.rs
```diff
@@ -111,12 +129,14 @@ where
     };
 
     let http_response = http::Response::from_parts(parts, Bytes::from(body));
-    trace!(http_response = ?http_response);
-    let parsed = handler.parse_loaded(&http_response);
-    sdk_result(
-        parsed,
-        operation::Response::from_parts(http_response.map(SdkBody::from), properties),
-    )
+    trace!(http_response = ?http_response, "read HTTP response body");
```
Velfi, line 132: I'm thinking we should we `debug!` the request ID header if one is present. Or does hyper already log the headers?

Velfi (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
jdisanti: I don't think I actually need this change anymore. But an earlier iteration of my changes did require it, which raises a compatibility question: Should we make these bounds as complete as possible so that we don't need to change the bounds in future SDK versions?

rust-runtime/aws-smithy-client/src/lib.rs
jdisanti: There's no way to include the streaming body without consuming it, so no.

rust-runtime/aws-smithy-client/src/lib.rs
jdisanti: @rcoh was arguing we should remove metadata from the operation entirely and just have something in the property bag, but that was looking to be a pretty sizable refactor that I didn't want to take on here.

I can take a look and see if it's easy to just borrow metadata from the operation here to eliminate the destructuring/restructuring.

rust-runtime/aws-smithy-http/src/middleware.rs
jdisanti: `?http_response` will output all the headers, but it's at `trace` level. I think request IDs are an AWS concept though? Or at least, this header would be specific to AWS? So this is probably not the right place to add it.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
Velfi: The only way this would blow up in our faces is if some important dependency starts emitting non-`Send`/`Sync` errors. I don't expect that to happen.

rust-runtime/aws-smithy-http/src/middleware.rs
rcoh: I would support adding a default middleware to log this stuff though...or maybe expose it in the property bag?

rust-runtime/aws-smithy-client/src/lib.rs
jdisanti: It seems promising at first, but then ends up not working out since calling `op.metadata()` to borrow the metadata locks the entire operation struct so that a property can't be inserted.

PR #2046, opened 2022-12-01
Fix service builder docs

82marbag:


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -487,129 +487,129 @@ class ServerServiceGeneratorV2(
...
+            //! ## Example
+            //!
+            //! ```rust
+            //! use std::net::SocketAddr;
+            //! use $crateName::$serviceName;
+            //!
+            //! ##[tokio::main]
+            //! pub async fn main() {
+            //!    let app = $serviceName::builder_without_plugins()
+            ${builderFieldNames.values.joinToString("\n") { "//!        .$it($it)" }}
```
jjant, line 588: Jeez this line :laughcry:

hlbarber (approved):

PR #2053, opened 2022-12-02
Add custom validation exceptions RFC

david-perez:
[Rendered](https://github.com/awslabs/smithy-rs/blob/158b09c91e23ea969fc31835e10720ab8da15a5d/design/src/rfcs/rfc0026_custom_validation_exceptions.md).

----

### DIFF
--- a/design/src/rfcs/rfc0026_custom_validation_exceptions.md
+++ b/design/src/rfcs/rfc0026_custom_validation_exceptions.md
@@ -1,0 +1,671 @@
+RFC: Custom validation exceptions
+=================================
+
+> Status: RFC
+>
+> Applies to: server
+
+This RFC describes a feature whereby server SDK application owners will be able
+to return custom error messages when validation of [constraint
+traits][constraint-traits] fails for an operation's input.
+
+[constraint-traits]: https://awslabs.github.io/smithy/2.0/spec/constraint-traits.html
+
+Terminology
+-----------
+
+- **Shape closure**: the set of shapes a shape can "reach", including itself.
+- **Transitively constrained shape**: a shape whose closure includes: 
+    1. a shape with a [constraint trait][constraint-traits] attached,
+    2. a (member) shape with a [`required` trait] attached,
+    3. an [`enum` shape]; or
+    4. an [`intEnum` shape].
+- A **directly constrained shape** is any of these:
+    1. a shape with a [constraint trait][constraint-traits] attached,
+    2. a (member) shape with a [`required` trait] attached,
+    3. an [`enum` shape],
+    4. an [`intEnum` shape]; or
+    5. a [`structure` shape] with at least one `required` member shape.
+- **Constrained type**: the Rust type a constrained shape gets rendered as. For
+  shapes that are not `structure`, `union`, `enum`, or `intEnum` shapes, these
+  are wrapper [newtype]s.
+- **`ConstraintViolation`**: the Rust type a constrained type's [`TryFrom`]
+  constructor exposes to user code when a value violates the direct modeled
+  constraints.
+- **`ConstraintViolationException`**: the Rust type a constrained type's
+  _internal_ framework-exclusive constructor returns when a value violates the
+  modeled constraints.
+- **Constrained operation**: an operation whose input shape is transitively
+  constrained.
+
+In the absence of a qualifier, "constrained shape" should be interpreted as
+"transitively constrained shape".
+
+To learn more about the design of constraint traits in server SDKs and more
+in-depth explanations of the above terminology, the following RFCs are
+recommended reading:
+
+- [RFC: Constraint traits](https://github.com/awslabs/smithy-rs/pull/1199).
+- [RFC: Better Constraint Violations][better-constraint-violations].
+
+[better-constraint-violations]: https://github.com/awslabs/smithy-rs/pull/2040
+[`required` trait]: https://smithy.io/2.0/spec/type-refinement-traits.html#required-trait
+[`enum` shape]: https://smithy.io/2.0/spec/simple-types.html#enum
+[`intEnum` shape]: https://smithy.io/2.0/spec/simple-types.html#intenum
+[`structure` shape]: https://smithy.io/2.0/spec/aggregate-types.html#structure
+[newtype]: https://doc.rust-lang.org/rust-by-example/generics/new_types.html
+[`TryFrom`]: https://doc.rust-lang.org/std/convert/trait.TryFrom.html
+
+### Background
+
+Constrained operations are currently required to have
+`smithy.framework#ValidationException` as a member in their [`errors`
+property](https://smithy.io/2.0/spec/service-types.html#operation). This is the
+shape that is rendered in responses when a request contains data that violates
+the modeled constraints.
+
+The shape is defined in the [`smithy-validation-model` Maven
+package](https://search.maven.org/artifact/software.amazon.smithy/smithy-validation-model),
+[as follows](https://github.com/awslabs/smithy/blob/main/smithy-validation-model/model/smithy.framework.validation.smithy):
+
+```smithy
+$version: "2.0"
+
+namespace smithy.framework
+
+/// A standard error for input validation failures.
+/// This should be thrown by services when a member of the input structure
+/// falls outside of the modeled or documented constraints.
+@error("client")
+structure ValidationException {
+
+    /// A summary of the validation failure.
+    @required
+    message: String,
+
+    /// A list of specific failures encountered while validating the input.
+    /// A member can appear in this list more than once if it failed to satisfy multiple constraints.
+    fieldList: ValidationExceptionFieldList
+}
+
+/// Describes one specific validation failure for an input member.
+structure ValidationExceptionField {
+    /// A JSONPointer expression to the structure member whose value failed to satisfy the modeled constraints.
+    @required
+    path: String,
+
+    /// A detailed description of the validation failure.
+    @required
+    message: String
+}
+
+list ValidationExceptionFieldList {
+    member: ValidationExceptionField
+}
+```
+
+Smithy's protocol tests provide a normative reference for what the error
+message contents' should be. For example, the
+[RestJsonMalformedPatternMapValue](https://github.com/awslabs/smithy/blob/31ddf685d7e3fe287eac51442621975d585972fd/smithy-aws-protocol-tests/model/restJson1/validation/malformed-pattern.smithy#L154-L154)
+test stipulates that upon receiving the JSON value:
+
+```json
+{ "map" : { "abc": "ABC" } }
+```
+
+for the modeled map:
+
+```smithy
+map PatternMap {
+    key: PatternString,
+    value: PatternString
+}
+
+@pattern("^[a-m]+$")
+string PatternString
+```
+
+the JSON response should be:
+
+```json
+{
+    { 
+        "message" : "1 validation error detected. Value abc at '/map/abc' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[a-m]+$$",
+        "fieldList" : [
+            {
+                "message": "Value ABC at '/map/abc' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[a-m]+$$",
+                "path": "/map/abc"
+            }
+        ]
+    }
+}
+```
+
+Currently, the framework is mapping from the operation input shape's
+`ConstraintViolationExceptions` (a [newtype] around a container of
+`ConstraintViolationException`s) to a `ValidationException`. This mapping is
+hardcoded so as to make the protocol tests pass.
+
+[newtype]: https://doc.rust-lang.org/rust-by-example/generics/new_types.html
+
+### Problem
+
+Some users have expressed interest in being able to customize the error
+responses. So far their use cases mainly revolve around:
+
+- using smithy-rs server SDKs in frontend services that are invoked by user
+  interface code: for example, said service might want to internationalize or
+  enhance error messages that are displayed directly to users in a web form.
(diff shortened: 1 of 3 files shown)


design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+}
+```
+
+However, in Rust, they'd need `N` different functions, each with different
+inputs and outputs, even if their core logic is likely to be the same:
+
+```rust
+fn map_operation1_validation_failure_to_error(errors: input::operation1_input::ConstraintViolationExceptions) -> error::Operation1Error { /* ... */ }
+...
+fn map_operation2_validation_failure_to_error(errors: input::operation1_input::ConstraintViolationExceptions) -> error::Operation2Error { /* ... */ }
```
david-perez, line 305:
```suggestion
fn map_operationN_validation_failure_to_error(errors: input::operationN_input::ConstraintViolationExceptions) -> error::OperationNError { /* ... */ }
```

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+        // another `usize` indicating the _modeled_ value in the `@length` trait; by
+        // keeping it inside `length_map` we can hardcode that value in the
+        // implementation of e.g. error messages.
+        pub struct Length(usize);
+
+        pub struct Members {
+            pub(crate) Vec<Member>
+        }
+
+        // This is always an `enum`, even when it only has one variant.
```
david-perez, line 532:
```suggestion
```

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+[smithy-typescript][smithy-typescript-custom-validation] does allow access to
+the incoming request's context.
+
+Since granting access to extractors is a backwards-compatible change, we could
+punt on implementing this until user interest is expressed and we learn about
+use cases.
+
+#### "Tightness" of constraint violations
+
+`ConstraintViolationExceptions` [is not
```
LukeMathWalker, line 440: What type of manipulation do we expect the users of custom validation exceptions to make that would require precise modelling of the constraint violation exception?

LukeMathWalker: In relation to https://github.com/awslabs/smithy-rs/pull/2040 - does this section advocate to use a tight exception model even if the user is not relying on custom validation exceptions?
I.e. is this why you were suggesting that this RFC is tied to https://github.com/awslabs/smithy-rs/pull/2040?

david-perez: Yes to both questions.

It's probably better that I move this tightness problem in #2040, I just thought that the problem was more apparent here.

david-perez: > What type of manipulation do we expect the users of custom validation exceptions to make that would require tight modelling of the constraint violation exception?

I don't know. It's not that they'd _require_ it. For what it's worth, an API that gave them a map from Smithy path to error message string would probably satisfy all of the use cases I know of.

Succinctly phrased, I think this is just a simple problem that raises a tradeoff between type safety/expressiveness and API complexity/understandability. But I'm not sure what option is optimal / strikes the right balance in that spectrum.

david-perez: No advantage in that case. But as the section calls out towards the end, this problem afflicts regular `ConstraintViolation`s too. I think we should only consider two cases: we either make both `ConstraintViolation`s and `ConstraintViolationException`s tight, or we make both not tight.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+
+Since the operation does not have `smithy.framework#ValidationException`
+attached, when registering the handler in the service builder, the user must
+provide the mapping to an `error::MyOperationError` variant when constraint
+trait enforcement fails.
+
+```rust
+let my_service = MyService::builder_without_plugins()
+    .my_operation(operation_handler)
+    .on_my_operation_validation_failure(map_operation_validation_failure_to_error)
```
hlbarber, line 239: Should this API be on the `Operation` struct?

For example,

```rust
let op = MyOperation::from_handler(handler).custom_validation(closure);
```

- This is more inline with how we apply specific layers to operations (`Operation::layer`) - a decision was made to not put more methods on the service builder [here](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0020_service_builder.md#approach-d-add-more-methods-to-the-service-builder).
- More closely resembles the typescript API, [related](https://github.com/awslabs/smithy-rs/pull/2053/files#diff-26f3492c91346ff7493644f733cd3391b418f88fb0d2099c7305b399e695e75bR377).

hlbarber:
Another alternative here would be to broaden the range of permitted inputs.

Currently we require the first argument of the handlers are `Op::Input` - the Smithy input. We could broaden this to allow the first argument to be something like `ParseResult<Op::Input, Exceptions>` where

```rust
enum ParseResult<T, E> {
     Ok(T),
    Err(E)
}
```

the customer can now map `ParseResult::Err` to the Smithy output.

This is probably _too_ general however.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
jjant: Came here to say basically the same thing 😄.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+   violation exceptions to a modeled error shape, as opposed to giving them the
+   freedom of returning _anything_ that can be converted into an HTTP response
+   is _intentional_. Too much freedom would allow a user to return HTTP
+   responses that could break the underlying protocol.
+2. `ConstraintViolationException`s have until now been `pub(crate)`, for
+   exclusive use by the framework. Opting into `disableDefaultValidation`
+   should unhide only those associated to constrained operations where
+   `smithy.framework#ValidationException` is not attached to the operation. The
+   user shouldn't have a need to reach for the other operation's constraint
+   violation exceptions.
```
jjant, line 260: Why do we care about this? It sounds like it wouldn't be a big deal that `ConstraintViolationException`s were public.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+```
+
+#### Synchronicity of the custom validation error mapping
+
+Should the `ConstraintViolationExceptions` mapping be `async`? In other words,
+should we require a [`core::ops::Fn`], or a future type? All use cases so far
+would be solved with a synchronous mapping, but a priori there should be no
+problem in allowing async functions. As a data point, axum's
+[`IntoResponse`](https://docs.rs/axum-core/latest/axum_core/response/trait.IntoResponse.html)
+is synchronous.
```
jjant, line 411: Would supporting async mappings be a backwards-compatible change?

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+modeled data, but a priori there should be no problem in allowing access to
+extractors. As a data point,
+[smithy-typescript][smithy-typescript-custom-validation] does allow access to
+the incoming request's context.
+
+Since granting access to extractors is a backwards-compatible change, we could
+punt on implementing this until user interest is expressed and we learn about
+use cases.
+
+#### "Tightness" of constraint violations
```
jjant, line 438: Not really relevant here, but I'd argue that "tightness" of inputs/outputs is way more important, and we don't do any better there (i.e, exact `@length` lists are not arrays, etc).

hlbarber:
Continuing on [this line of thought](https://github.com/awslabs/smithy-rs/pull/2053#issuecomment-1370808681)...

Suppose a customer wanted to implement their own bespoke "constraints", not described in the Smithy specification, and want to serialize them into an error alongside non-bespoke constraints placed on their model. This would be enabled by an extractor

```rust
fn handler(input: Op::Input, constraint_violations: Option<ConstraintViolations>) -> Result<Op::Output, Op::Error>
```

which returns `Op::Input` whenever is deserializers correctly and `Some(violation)` when it fails the non-bespoke constraints placed on the model.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+
+Terminology
+-----------
+
+- **Shape closure**: the set of shapes a shape can "reach", including itself.
+- **Transitively constrained shape**: a shape whose closure includes: 
+    1. a shape with a [constraint trait][constraint-traits] attached,
+    2. a (member) shape with a [`required` trait] attached,
+    3. an [`enum` shape]; or
+    4. an [`intEnum` shape].
```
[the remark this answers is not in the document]
david-perez: Yes you're right. I haven't realized until now I've been using the terminology incorrectly! I've used it all throughout the codebase too.

Shall we change it to "transitively constrain**ing** shape"? Do you have any suggestions?

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+- **Constrained type**: the Rust type a constrained shape gets rendered as. For
+  shapes that are not `structure`, `union`, `enum`, or `intEnum` shapes, these
+  are wrapper [newtype]s.
+- **`ConstraintViolation`**: the Rust type a constrained type's [`TryFrom`]
+  constructor exposes to user code when a value violates the direct modeled
+  constraints.
+- **`ConstraintViolationException`**: the Rust type a constrained type's
+  _internal_ framework-exclusive constructor returns when a value violates the
+  modeled constraints.
+- **Constrained operation**: an operation whose input shape is transitively
```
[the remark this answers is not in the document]
david-perez: Because the framework doesn't care about constraints on output or error shapes insofar as they are something the user constructs and hence the constrained types take care of honoring the constraints _in their application code_. In this RFC we're just interested in the constraints of shapes that lie in operation inputs, because those are checked _in framework code_ upon request deserialization.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+                "message": "Value ABC at '/map/abc' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[a-m]+$$",
+                "path": "/map/abc"
+            }
+        ]
+    }
+}
+```
+
+Currently, the framework is mapping from the operation input shape's
+`ConstraintViolationExceptions` (a [newtype] around a container of
```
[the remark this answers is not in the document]
david-perez: How? You mean to put an anchor link to the "Terminology" section?

design/src/rfcs/rfc0026_custom_validation_exceptions.md
```diff
@@ -0,0 +1,671 @@
...
+
+    // We can return _any_ error attached to the operation.
+    error::OperationError::ValidationCustomError(
+        error::ValidationCustomError {
+            validation_error_message
+        }
+    )
+}
+
+async fn operation_handler(input: input::MyOperationInput) -> Result<output::MyOperationOutput, error::MyOperationError> {
```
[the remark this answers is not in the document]
david-perez: No. `error::MyOperationError` is the top-level enum (it's a synthetic shape; we derive its name from the operation name), which contains the `error::OperationError1`, ..., `error::OperationErrorN` variants.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
david-perez: As mentioned in the beginning of the document, this feature caters to a small tail of power users. I think making both `ConstraintViolation`s and `ConstraintViolationException`s public would unnecessarily expose API surface (i.e. make users overwhelmed when looking at Rust docs) when most users will just interact with `ConstraintViolation`s.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
david-perez: No, async functions essentially desugar to functions that return `Future`s.

design/src/rfcs/rfc0026_custom_validation_exceptions.md
david-perez: Hmm having exact `@length` list shapes render as Rust arrays would certainly be surprising, since `Vec`s and arrays have entirely different APIs and semantics (storage in stack / heap probably being the most notable).

I think constrained types for input and output shapes are as tight as they can be while still honoring the Smithy spec (i.e. we can't use unsigned integers for positive `@range`s either); do you have any other examples?

hlbarber:
Here's a way we could support easy iteration over a loose error _and_ support for tight errors.

```rust
struct Tight {
    a: Option<A>,
    b: Option<B>,
    c: Option<C>
}

enum Loose<'a> {
    A(&'a A),
    B(&'a B),
    C(&'a C)
}

impl Tight {
    fn iter(&self) -> impl Iterator<Item = Loose<'_>> {
        [
            self.a.as_ref().map(Loose::A),
            self.b.as_ref().map(Loose::B),
            self.c.as_ref().map(Loose::C)
        ]
        .iter()
        .filter_map(std::convert::identity)
    }
}
```

PR #2055, opened 2022-12-02
Fix inability to create service clients when `default-features = false`

Velfi:
This PR still needs a way to run the new tests in CI but I'm publishing it in case we don't want to block release on that.

## Motivation and Context

Users should be able to create service clients when default features are disabled

## Description

add: tests for config/client construction when default features are disabled
update: missing HTTP connector panic messages

## Testing

I wrote 2 tests

## Checklist

----


CHANGELOG.next.toml
```diff
@@ -502,3 +502,23 @@ in non-serverless environments (e.g. via `hyper`).
 references = ["smithy-rs#2035"]
 meta = { "breaking" = true, "tada" = false, "bug" = false, "target" = "server" }
 author = "LukeMathWalker"
+
+[[smithy-rs]]
+message = """
+It is once again possible to create a service client when no default features are enabled. When default features are
```
jdisanti, line 508: It was still possible last release, so really this should just say that there will be a panic if there's no connector provided or defaulted.

Might not even be worth mentioning in the changelog since it should be covered by the previous changelog entries for refactoring HTTP connectors.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -158,13 +158,25 @@ private class AwsFluentClientExtensions(types: Types) {
             rustTemplate(
                 """
                 /// Creates a new client from an [SDK Config](#{aws_types}::sdk_config::SdkConfig).
-                ##[cfg(any(feature = "rustls", feature = "native-tls"))]
+                ///
+                /// ###### Panics
```
jdisanti, line 162: I think you only need two `#`, as we do for `## Examples` in other places
```suggestion
                /// ## Panics
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -158,13 +158,25 @@ private class AwsFluentClientExtensions(types: Types) {
...
+                /// - This method will panic if the `conf` is missing an HTTP connector. If you experience this panic, set the
+                ///     `http_connector` on the Config passed into this function to fix it.
                 pub fn new(sdk_config: &#{aws_types}::sdk_config::SdkConfig) -> Self {
                     Self::from_conf(sdk_config.into())
                 }
 
                 /// Creates a new client from the service [`Config`](crate::Config).
-                ##[cfg(any(feature = "rustls", feature = "native-tls"))]
+                ///
+                /// ###### Panics
```
jdisanti, line 174:
```suggestion
                /// ## Panics
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -158,13 +158,25 @@ private class AwsFluentClientExtensions(types: Types) {
             rustTemplate(
                 """
                 /// Creates a new client from an [SDK Config](#{aws_types}::sdk_config::SdkConfig).
-                ##[cfg(any(feature = "rustls", feature = "native-tls"))]
+                ///
+                /// ###### Panics
+                ///
+                /// - This method will panic if the `conf` is missing an async sleep implementation. If you experience this panic, set
```
jdisanti, line 164:
```suggestion
                /// - This method will panic if the `sdk_config` is missing an async sleep implementation. If you experience this panic, set
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -158,13 +158,25 @@ private class AwsFluentClientExtensions(types: Types) {
...
+                ///     the `sleep_impl` on the Config passed into this function to fix it.
+                /// - This method will panic if the `conf` is missing an HTTP connector. If you experience this panic, set the
```
jdisanti, line 166:
```suggestion
                /// - This method will panic if the `sdk_config` is missing an HTTP connector. If you experience this panic, set the
```

aws/sdk/integration-tests/Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
     "iam",
     "kms",
     "lambda",
+    "no-default-features",
```
jdisanti, line 11: Fine for now, but this will need a follow up to actually run this in CI since CI only runs integration tests named after services.

jdisanti (approved):

CHANGELOG.next.toml
Velfi: I removed them.

aws/sdk/integration-tests/Cargo.toml
Velfi: Issue created: https://github.com/awslabs/smithy-rs/issues/2056

PR #1932, opened 2022-11-01
Migrate to EnumShape

82marbag:
Deprecate EnumTrait and migrate to EnumShape.

Closes #1671

## Checklist

- [x] I have updated `CHANGELOG.next.toml` if I made changes to the smithy-rs codegen or runtime crates
- [x] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2058, opened 2022-12-05
Compile regexes for `@pattern` strings early

jjant:
## Description

Completes https://github.com/awslabs/smithy-rs/issues/2026.

- Initializes regexes during `${service}Builder::build()`
- Code generates tests making sure the regexes compile in rust

----


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -224,9 +233,28 @@ class ServerServiceGeneratorV2(
...
         )
     }
 
+    /**
+     * Renders `PatternString::compile_regex()` function calls for every
+     * `@pattern`-constrained string shape in the service closure.
+     */
+    @Suppress("DEPRECATION")
+    private fun patternInitializations(): Writable =
+        Walker(model).walkShapes(service).mapNotNull { shape ->
```
david-perez, line 246: Can't you do a `filter` and then a `map` instead of a `mapNotNull` here?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -224,9 +233,28 @@ class ServerServiceGeneratorV2(
...
+            if (shape is StringShape && shape.hasTrait<PatternTrait>() && !shape.hasTrait<software.amazon.smithy.model.traits.EnumTrait>()) {
```
david-perez, line 247: nit: import `EnumTrait`.

david-perez (commented):
This is good but we should also code-generate a `#[test]` per pattern compiling each regex to do even earlier detection.

A Kotlin unit test exercising a model whose `pattern` is accepted by Smithy but not by the `regex` crate that asserts that the tests fail and that the service panics at startup would be nice too.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
jjant: The reason I didn't is to be able to use the `@Suppress("DEPRECATION")` annotation in this function, if you import `EnumTrait` you have to suppress deprecations for the whole file :( (unless there's other syntax I'm missing).

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
jjant: I tend to intentionally use `mapNotNull` over `filter + map` for efficiency (iterating the list once). It should be called `filterMap` to make the intent clearer : p.

I don't feel strongly about this, so happy to change it.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
82marbag: You could possibly already use `EnumShape` as `EnumTrait` is deprecated in #1932

LukeMathWalker (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedStringGenerator.kt
```diff
@@ -264,6 +294,8 @@ private data class Pattern(val patternTrait: PatternTrait) : StringTraitInfo() {
                     }
                 }
 
+                /// Attempts to compile the regex for this constrained type's `@pattern`.
+                /// This can fail if the specified regex is not supported by the #{Regex} crate.
```
LukeMathWalker, line 298: Micro-nit:
```suggestion
                /// This can fail if the specified regex is not supported by the `#{Regex}` crate.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/TraitInfo.kt
```diff
@@ -20,7 +22,27 @@ data class TraitInfo(
...
     val validationFunctionDefinition: (constraintViolation: Symbol, unconstrainedTypeName: String) -> Writable,
-)
+    private val testCases: List<Writable> = listOf(),
+) {
+    companion object {
+        fun testCases(constraintsInfo: List<TraitInfo>): List<Writable> {
+            val annotatedTestCases = constraintsInfo.flatMap { it.testCases }
+                .map { testCase ->
+                    writable {
+                        rustTemplate(
```
david-perez, line 32: We could use `RustWriter.unitTest` here, and the list of test writables provided by `TraitInfo` would then just comprise the test function bodies.

david-perez (commented):
Implementation looks good but I'd still like to have guarantees in tests that this is working as expected:

> A Kotlin unit test exercising a model whose pattern is accepted by Smithy but not by the regex crate that asserts that the tests fail and that the service panics at startup would be nice too.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/Util.kt
```diff
@@ -53,16 +52,14 @@ class Types(runtimeConfig: RuntimeConfig) {
     val resolveEndpointError = smithyHttpEndpointModule.resolve("ResolveEndpointError")
 }
 
-private fun String.stringToRustName(): String = RustReservedWords.escapeIfNeeded(this.toSnakeCase())
-
```
rcoh, line 57: this was private for a reason—we should be quite careful about converting strings to rust identifiers in this way. In general, names should be "symbol aware" so that the symbol provider can be the source of truth.

In the case of test modules, this is still appropriate to use but if you make it public, we should give it a name that resists misuse (and a comment explaining why this generally shouldn't be used to name symbols).

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt
```diff
@@ -393,6 +396,27 @@ fun Symbol.Builder.rustType(rustType: RustType): Symbol.Builder = this.putProper
...
+    val module = module()
+
+    return RustModule.new(
+        name = "test_${name.toRustName()}",
+        visibility = Visibility.PRIVATE,
+        inline = true,
+        parent = module.parent,
+        additionalAttributes = listOf(Attribute.Cfg("test")),
+    )
+}
```
rcoh, line 418: slight preference for adding this on `RustModule` instead—I think it's a little odd to put it inside symbol visitor (since it doesn't really have anything to do with the symbol). Callers would do `symbol.module().testModule()`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedStringGenerator.kt
```diff
@@ -239,6 +260,15 @@ private data class Pattern(val patternTrait: PatternTrait) : StringTraitInfo() {
                 )
             },
             this::renderValidationFunction,
+            testCases = listOf {
+                rust(
+                    """
+                    fn regex_compiles() {
+                        ${symbol.name}::compile_regex();
```
rcoh, line 267: I think you probably want `#T::compile_regex()`? Then you also want need the `use super::**`. In general, I try to avoid action-at-a-distance in codegen

rcoh (commented):
overall looks good, just a few small changes requested

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt
jjant: That's what I initially wanted to do, but in this case all `symbol`s (constrained types) are defined in the same `model.rs` module, and, from my understanding, I can't access the symbol _from_ the module, which means that for all these types the `symbol.module().testModule()` would be the same (say, `mod test_model`) which complicates codegen a bunch.

I guess I could do something like: `symbol.module().testModule(symbol.name)` too.

Let me know what you think.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedStringGenerator.kt
jjant: Do you mean this?
```kotlin
rustTemplate("""
  fn regex_compiles() {
    #T::compile_regex();
  }
"""
, "T" to symbol)
```

That will automatically add the `use` directive later? Nice!

rcoh: it uses the fully-qualified path

david-perez: You can also just use `rust` like so:

```kotlin
rust("""
  fn regex_compiles() {
    #T::compile_regex();
  }
""", symbol)

Comment (jjant) [reply codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedStringGenerator.kt]: Done in 339b6080cdb6735a2049a4bbec5615a8fe96db7f.

File: codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt (hunk 6 above)
Comment (jjant) [reply codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt]: Moved this to `SymbolProvider.testModuleForShape` as we discussed, in f03bf200dd14f91035f19764832eee4a289f716d.

File: codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/TraitInfo.kt (hunk 4 above)
Comment (jjant) [reply codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/TraitInfo.kt]: Done in 680857323d283270d8342f2f5b80d3c872fcd137.

File: codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt (hunk 1 above)
Comment (jjant) [reply codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt]: Done in 4213e8840bcb2ff802d1ec30d8c414ac895178aa.

File: codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt [hunk 8]
```diff
@@ -219,6 +224,10 @@ class ServerServiceGeneratorV2(
                         });
                     }
                     let $expectMessageVariableName = "this should never panic since we are supposed to check beforehand that a handler has been registered for this operation; please file a bug report under https://github.com/awslabs/smithy-rs/issues";
+
+                    // Eagerly initialize regexes for `@pattern` strings.
```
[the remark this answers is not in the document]
jjant: Done in 39b6ed828f29f9956b7bbf064731e575c7d4292b.

jjant:
@david-perez Added a Kotlin test in https://github.com/awslabs/smithy-rs/pull/2058/commits/2873529dac9ba79f3ffe2386e43851ddb1c0b9fe.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedStringGeneratorTest.kt
```diff
@@ -184,4 +186,26 @@ class ConstrainedStringGeneratorTest {
...
+        val codegenContext = serverTestCodegenContext(model)
+        val project = TestWorkspace.testProject(codegenContext.symbolProvider)
+
+        project.withModule(ModelsModule) {
+            ConstrainedStringGenerator(codegenContext, this, constrainedStringShape).render()
+        }
+
+        assertThrows<CommandFailed> {
+            project.compileAndTest()
+        }
```
LukeMathWalker, line 209: Can we in any way verify the message it fails with, to make sure it fails for the reason we expect?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -230,9 +238,32 @@ class ServerServiceGeneratorV2(
...
+            .filter { shape -> shape is StringShape && shape.hasTrait<PatternTrait>() && !shape.hasTrait<software.amazon.smithy.model.traits.EnumTrait>() }
+            .map { shape -> codegenContext.constrainedShapeSymbolProvider.toSymbol(shape) }
+            .map { symbol ->
+                writable {
+                    rustTemplate("#{Type}::compile_regex();", "Type" to symbol)
+                }
+            }.toMutableList()
+
+        if (patterns.isNotEmpty()) {
+            patterns += writable { rust("// Eagerly initialize regexes for `@pattern` strings.") }
```
LukeMathWalker, line 261: Extremely minor nit: it'd be nicer to prepend this comment, instead of appending it at the end.

LukeMathWalker (approved):

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedStringGeneratorTest.kt
rcoh: yeah—you can look at the error message. An example: 
```
        val failure = shouldThrow<CommandFailed> { "cargo test".runWithWarnings(testDir) }
        failure.output shouldContain "endpoint::test::test_1"
        failure.output shouldContain "https://failingtest.com"
```

jjant: Lovely, thanks!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
rcoh: can do `patterns.letIf(patterns.isNotEmpty()) { docs + patterns }`

rcoh (approved):
lgtm!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
jjant: Thank you both! This was an oversight, I intended to prepend it. Done in 8475b143a5cc680afb6a2ed6c883c6dc4cb23b90.

LukeMathWalker (approved):

Issue #1304, opened 2022-04-05
Add tests for highly concurrent and single threaded contexts


rcoh:
Important for testing hot locks and other potential concurrency related issues

Velfi:
We may encounter locks in single-threaded contexts due to https://github.com/tokio-rs/tokio/issues/4730

This task will surface issues like that but won't fix them. That's OK.

Velfi:
The Java SDK's tests are located [here](https://github.com/aws/aws-sdk-java-v2/tree/master/test/stability-tests/src/it/java/software/amazon/awssdk/stability/tests).

Velfi:
If we discover any fixable issues re: concurrency, we should respond to this discussion: https://github.com/awslabs/aws-sdk-rust/discussions/604

PR #2061, opened 2022-12-05
Add simple request concurrency tests to S3 integration test suite.

Velfi:
## Motivation and Context

#1304

## Description

This PR adds 3 concurrency tests to our S3 integration test suite.
1. **`test_concurrency_on_multi_thread`**
  This tests that S3 `PutObject` request futures can be run concurrently on a multi-threaded `tokio` executor. The requests are run against an in-memory service that always returns `200 OK`. This runs 10k tasks in total with a concurrency limit of 100.
2. **`test_concurrency_on_single_thread`**
  This tests that S3 `PutObject` request futures can be run concurrently on a single-threaded `tokio` executor. The requests are run against an in-memory service that always returns `200 OK`. This runs 10k tasks in total with a concurrency limit of 100.
3. *`test_concurrency_put_object_against_live` This test is ignored because it runs against S3*.
  When run, this test will upload 10k objects to S3, each with a size of 100kb. Concurrency is limited to 100 in these tests but can be pushed higher depending on the payload size.

Weirdly, when running the tests that hit the dummy server, it appears to be limited to 129 concurrent tasks before `hyper` starts return errors. Running against S3 doesn't have this issue although S3 will throttle requests and we don't implement adaptive retries so properly handling this throttling is not possible without significant effort.

A screenshot showing the histogram output

## Testing

This PR only contains tests.

---


aws/sdk/integration-tests/s3/tests/concurrency.rs
```diff
@@ -0,0 +1,379 @@
...
+
+#[tokio::test(flavor = "current_thread")]
+async fn test_concurrency_on_multi_thread() {
+    test_concurrency_with_dummy_server().await
+}
+
+#[tokio::test(flavor = "current_thread")]
+async fn test_concurrency_on_single_thread() {
+    test_concurrency_with_dummy_server().await
+}
```
rcoh, line 40: maybe we should refactor the test to be parameterized on the `http_connector`? Or you could parameterize it on `fn(SdkConfig::Builder) -> SdkConfig::Builder`?

aws/sdk/integration-tests/s3/tests/concurrency.rs
```diff
@@ -0,0 +1,379 @@
...
+        let mut time_to_respond = false;
+        let mut bytes_left_to_write = response.len();
+
+        loop {
+            match socket.try_read_buf(&mut buf) {
+                // Ok(0) => {
+                //     unreachable!(
+                //         "The connection will be closed before this branch is ever reached"
+                //     );
+                // }
```
ysaito1001, line 130: Can these lines be removed?

aws/sdk/integration-tests/s3/tests/concurrency.rs
```diff
@@ -0,0 +1,379 @@
...
+    );
+    display_histogram(
+        "Request Latency",
+        req_histogram,
+        "s",
+        Duration::from_secs(1).as_nanos() as f64,
+    );
+}
+
+fn display_histogram(name: &str, mut h: SyncHistogram<u64>, unit: &str, scale: f64) {
```
ysaito1001, line 352: Not necessarily need to take an action, but since the histogram code is in place, it'd be cool to add a screenshot of example histograms to the PR if available.

ysaito1001 (commented):
Looks good to me. I'll intentionally keep it unapproved to avoid it getting merged inadvertently before feedback from other reviewers has been addressed.

aws/sdk/integration-tests/s3/tests/concurrency.rs
Velfi: yes, done.

jdisanti (approved):

Velfi:
This is currently failing due to including a dependency that's only used in an ignored test:
```
 unused dependencies:
`aws-sdk-s3 v0.0.0-local (/home/build/workspace/aws-sdk-smoketest/sdk/s3)`
└─── dependencies
     └─── "hdrhistogram"
Note: These dependencies might be used by other targets.
      To find dependencies that are not used by any target, enable `--all-targets`.
Note: They might be false-positive.
      For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
      To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.
```

 I'd say that we should allowlist this case or maybe run `udeps` with the `--all-targets` flag as suggested.

Velfi:
> This is currently failing due to including a dependency that's only used in an ignored test:
> 
> ```
>  unused dependencies:
> `aws-sdk-s3 v0.0.0-local (/home/build/workspace/aws-sdk-smoketest/sdk/s3)`
> └─── dependencies
>      └─── "hdrhistogram"
> Note: These dependencies might be used by other targets.
>       To find dependencies that are not used by any target, enable `--all-targets`.
> Note: They might be false-positive.
>       For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
>       To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.
> ```
> 
> I'd say that we should allowlist this case or maybe run `udeps` with the `--all-targets` flag as suggested.



> This is currently failing due to including a dependency that's only used in an ignored test:
> 
> ```
>  unused dependencies:
> `aws-sdk-s3 v0.0.0-local (/home/build/workspace/aws-sdk-smoketest/sdk/s3)`
> └─── dependencies
>      └─── "hdrhistogram"
> Note: These dependencies might be used by other targets.
>       To find dependencies that are not used by any target, enable `--all-targets`.
> Note: They might be false-positive.
>       For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
>       To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.
> ```
> 
> I'd say that we should allowlist this case or maybe run `udeps` with the `--all-targets` flag as suggested.

Nevermind, I misinterpreted this. I believe I need to set the dependency as a dev dependency in Codegen

PR #2062, opened 2022-12-05
Improve credentials tracing

jdisanti:
## Motivation and Context
This PR adds an `info` event for each time there is a credentials cache miss with the time it took to reload credentials, removes some duplicate provider spans, and adjusts some event/span levels.

## Checklist

----


aws/rust-runtime/aws-config/src/imds/credentials.rs
```diff
@@ -152,7 +152,7 @@ impl ImdsCredentialsProvider {
             Err(ImdsError::ErrorResponse(context))
                 if context.response().status().as_u16() == 404 =>
             {
-                tracing::info!(
+                tracing::warn!(
```
Velfi, line 155: users are gonna love this

aws/rust-runtime/aws-config/src/meta/credentials/lazy_caching.rs
```diff
@@ -77,16 +77,17 @@ impl ProvideCredentials for LazyCachingCredentialsProvider {
...
             } else {
                 // If we didn't get credentials from the cache, then we need to try and load.
                 // There may be other threads also loading simultaneously, but this is OK
                 // since the futures are not eagerly executed, and the cache will only run one
                 // of them.
-                let span = trace_span!("lazy_load_credentials");
+                let span = info_span!("lazy_load_credentials");
                 let future = Timeout::new(loader.provide_credentials(), timeout_future);
-                cache
+                let start_time = Instant::now();
```
Velfi, line 89: I'm wondering if we should use `tokio::time::Instant` in this so that this is more controllable in tests. Maybe it would be worth testing cache miss time? IDK

Velfi (approved):

aws/rust-runtime/aws-config/src/meta/credentials/lazy_caching.rs
jdisanti: Using `tokio::time::Instant` in tests is fine, but I'm not as sold on using it in the actual code (without abstracting it in some way) since it requires the `time` feature, which pulls in Tokio's async sleep. I don't think this specific case really needs a test.

PR #2064, opened 2022-12-05
Establish default max idle connections on default connectors

jdisanti:
## Motivation and Context
This PR establishes a default for max idle connections on the default connectors provided by the SDK for https://github.com/awslabs/aws-sdk-rust/issues/632.

## Checklist

----


ysaito1001 (approved):
Looks good! I suppose it's not straightforward to convert the code in `Reproduction Steps` in aws-sdk-rust#632 into an integration test or something. Even if we manage to write one, it might end up being fragile due to the fact that the test needs to monitor open file handles over some period of time, which might extend the test execution time and we probably don't want that.

Velfi (approved):

PR #2068, opened 2022-12-07
Implement FromParts for Option, Result

82marbag:


rust-runtime/aws-smithy-http-server/src/request/mod.rs
```diff
@@ -36,6 +36,15 @@
 //!
 //! See [Accessing Un-modelled data](https://github.com/awslabs/smithy-rs/blob/main/design/src/server/from_parts.md)
 //! a comprehensive overview.
+//!
+//! Implementations exist for:
```
hlbarber, line 40: The following implementations exist:

- Tuples up to size 8, which extracts each component.
- `Option<T>` attempts to extract `T`, otherwise returns `None`.
- `Result<T, T::Rejection>` extracts `T`, otherwise returns the extraction error.

hlbarber (approved):

LukeMathWalker (approved):

Issue #1784, opened 2022-09-29
Endpoints 2.0


jdisanti:
The approach for endpoint resolution is being redone to better support the edge cases without customizations. This is the tracking issue for that work.
## implementation work
- [x] Endpoints resolver generator: #2030 
- [x] Integrate endpoints resolver: #2074 
- [x] deprecate aws_smithy_http::Endpoint: #2074
- [ ] remove aws_smithy_http::Endpoint
- [x] deprecate & remove AwsEndpoint
- [x] Add FIPS builtIn
- [x] Add DualStack builtIn
- [ ] finalize RFC and ensure it reflects the actual implementation

## testing:
- [x]  presigned URL canary
- [ ]  canary with force path style
- [ ]  canary with virtual addressing
- [x]  s3 object lambda tests

## native smithy
- [ ] documentation of how native smithy services can codegenerate an endpoint resolver
- [ ] add a default middleware for native smithy services that sets the endpoint on the request

PR #2074, opened 2022-12-07
Enable Endpoints 2.0

rcoh:
## Motivation and Context
- #1784 
- enable endpoints 2.0
- create the `endpoint_url` setter which is threaded into the `Sdk::Config `builtIn`
- delete unused code, deprecate the old endpoint struct
- update tests to use `endpoint_url`

## Description

This PR enables Endpoints 2.0 for all services. For Smithy native clients, endpoints 2.0 is enabled but we haven't yet setup a default middleware that would make using that possible.

When endpoint rules are not defined & no resolver is provided, we currently will insert a resolver that always fails. A future PR will change this to panic on config construction but we will delay that until more refactoring and soaking has been done.
## Testing
- [x] endpoint tests
- [x] existing tests pass

Future PRs will follow up with more integration tests 

## Checklist

----


aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -340,6 +344,29 @@ mod loader {
             self
         }
 
+        /// Override the endpoint URL used for **all** AWS services.
+        ///
+        /// This method will override the endpoint URL used for **all** AWS services. This mainly
+        /// exists to set a static endpoint for tools like `LocalStack`. For live traffic, AWS services
+        /// require the service-specific endpoint resolver they provide.
```
Velfi, line 351: I think this could be improved a bit so I took a shot at it.
```suggestion
        /// exists to set a static endpoint for tools like `LocalStack`. This can not be used to send requests to
        /// AWS services as they require the service-specific endpoint resolvers that this option overrides.
```

aws/rust-runtime/aws-types/src/endpoint.rs
```diff
@@ -4,6 +4,7 @@
  */
 
 //! AWS SDK endpoint support.
+#![allow(deprecated)]
```
Velfi, line 7: We expect to eventually remove the resolver fn right? If yes, we should create an issue and add a TODO here

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
```diff
@@ -5,58 +5,70 @@
...
-import software.amazon.smithy.rust.codegen.core.util.dq
-import software.amazon.smithy.rust.codegen.core.util.expectTrait
-import software.amazon.smithy.rust.codegen.core.util.orNull
-import kotlin.io.path.readText
+import software.amazon.smithy.rust.codegen.core.util.letIf
 
 class AwsEndpointDecorator : RustCodegenDecorator<ClientProtocolGenerator, ClientCodegenContext> {
     override val name: String = "AwsEndpoint"
-    override val order: Byte = 0
+    override val order: Byte = -100
```
Velfi, line 40: This is giving me [CSS Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) flashbacks. Perhaps we should introduce a constant that represents "after everything else"?

_(I think I got that order correct)_

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
```diff
@@ -65,143 +77,144 @@ class AwsEndpointDecorator : RustCodegenDecorator<ClientProtocolGenerator, Clien
...
+                /// Temporary shim to allow new and old endpoint resolvers to co-exist
+                ///
+                /// This enables converting from the actual parameters type to the placehold parameters type that
+                /// solely holds a region
+                ##[doc(hidden)]
+                impl From<#{Params}> for #{PlaceholderParams} {
+                    fn from(params: #{Params}) -> Self {
+                        Self::new(params.region().map(|r|#{Region}::new(r.to_string())))
+                    }
+                }
```
Velfi, line 108: Should we add a todo comment here too?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
```diff
@@ -65,143 +77,144 @@ class AwsEndpointDecorator : RustCodegenDecorator<ClientProtocolGenerator, Clien
...
-                    }
-                    """,
-                    *codegenScope,
-                )
+                ##[deprecated(note = "use endpoint_url or set the endpoint resolver directly")]
+                /// Sets the endpoint resolver to use when making requests.
+                ///
+                /// This method is deprecated, use [`Builder::endpoint_url`] or [`Builder::endpoint_resolver`] instead.
+                pub fn set_aws_endpoint_resolver(&mut self, endpoint_resolver: Option<std::sync::Arc<dyn #{ResolveAwsEndpoint}>>) -> &mut Self {
+                    self.endpoint_resolver = endpoint_resolver.map(|res|std::sync::Arc::new(#{EndpointShim}::from_arc(res) ) as _);
```
Velfi, line 182: These don't work without the `as` right? Just checking

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
```diff
@@ -65,143 +77,144 @@ class AwsEndpointDecorator : RustCodegenDecorator<ClientProtocolGenerator, Clien
...
-            else -> emptySection
-        }
-    }
-}
+                /// Sets the endpoint url used to communicate with this service
+                ///
+                /// Note: this is used in combination with other endpoint rules, e.g. an API that applies a host-label prefix
+                /// will be prefixed onto this URL. To fully override the endpoint resolver, use
+                /// [`Builder::endpoint_resolver`].
+                pub fn set_endpoint_url(&mut self, endpoint_url: Option<String>) -> &mut Self {
```
Velfi, line 201: Is the reason for not taking a URL here that we don't want to expose a 3rd party type?

aws/sdk/integration-tests/s3/tests/ignore-invalid-xml-body-root.rs
```diff
@@ -76,5 +77,5 @@ async fn ignore_invalid_xml_body_root() {
         .await
         .unwrap();
 
-    conn.assert_requests_match(&[]);
+    conn.assert_requests_match(&[AUTHORIZATION]);
```
Velfi, line 80: Why must we ignore `AUTHORIZATION` here? Does EP 2.0 mean we can't assume that header is deterministic?

aws/sdk/integration-tests/s3/tests/query-strings-are-correctly-encoded.rs
```diff
@@ -60,7 +60,7 @@ async fn test_s3_signer_query_string_with_all_valid_chars() {
 
     // This is a snapshot test taken from a known working test result
     let snapshot_signature =
-        "Signature=775f88213304a5233ff295f869571554140e3db171a2d4a64f63902c49f79880";
+        "Signature=647aa91c7f91f1f1c498ef376fea370b48d0cd8c80a53c8e2cd64e3fc527a5e0";
```
Velfi, line 63: It must have been a pain to have to find all these tests. Thanks for doing all this.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolTestGeneratorTest.kt
```diff
@@ -235,60 +236,47 @@ class ProtocolTestGeneratorTest {
...
                 clazz.isAssignableFrom(ClientCodegenContext::class.java)
         }
-        val visitor = CodegenVisitor(
-            pluginContext,
-            codegenDecorator,
-        )
-        visitor.execute()
-        println("file:///$testDir/src/operation.rs")
-        return testDir
+        return clientIntegrationTest(model, additionalDecorators = listOf(codegenDecorator)) { _, _ -> }
```
Velfi, line 239: What does the `{ _, _ -> }` at the end mean? Is that a closure that ignores two args and then does nothing? It looks so weird.

rust-runtime/aws-smithy-client/src/test_connection.rs
```diff
@@ -201,6 +201,7 @@ impl<B> TestConnection<B> {
         self.requests.lock().unwrap()
     }
 
+    #[track_caller]
```
Velfi, line 204: TIL about [`#[track_caller]`](https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute)

aws/rust-runtime/aws-types/src/endpoint.rs
Velfi: Actually, now that I think about it, perhaps we just need to create an issue that says "remove all deprecated code" that we pick up before GA

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
rcoh: Just nicer for users since we can take static strings and we're going to convert it back to a string anyway to pass it into ep2

aws/sdk/integration-tests/s3/tests/ignore-invalid-xml-body-root.rs
rcoh: no, we just weren't testing the auth header here and I didn't want to both fixing it

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolTestGeneratorTest.kt
rcoh: yeah, normally it accepts a lambda. Let me put in a default arg for that, good idea.

Velfi (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientDecorator.kt
```diff
@@ -140,7 +140,7 @@ class GenericFluentClient(codegenContext: CodegenContext) : FluentClientCustomiz
                     /// ##     */
                     /// ##     .middleware_fn(|r| r)
                     ///       .build();
-                    /// let config = Config::builder().build();
+                    /// let config = Config::builder().endpoint_resolver("https://www.myurl.com").build();
```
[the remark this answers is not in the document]
rcoh: no, but this is a great question. `endpoint_uri` is an SDKism–it seeds the `SDK::Endpoint` builtIn. Since this depends on rules, and generic services might not have rules, we don't support this directly.

this is using that fact that `EndpointResolver<T>` is impled by `&'static str` for convenience

PR #2078, opened 2022-12-08
Run TODO lint against more files

david-perez:
This change would have prevented me from committing what was reverted in
https://github.com/awslabs/smithy-rs/issues/2076.

----


codegen-core/common-test-models/constraints.smithy
```diff
@@ -10,7 +10,6 @@ use smithy.framework#ValidationException
 @title("ConstraintsService")
 service ConstraintsService {
     operations: [
-        // TODO Rename as {Verb}[{Qualifier}]{Noun}: https://github.com/awslabs/smithy-rs/pull/1342#discussion_r980936650
```
david-perez, line 13: I removed this because I couldn't think of any good names. The academic names of the operations are clear enough.

codegen-core/common-test-models/constraints.smithy
```diff
@@ -876,10 +875,8 @@ map ConBMap {
 
 @error("client")
 structure ErrorWithLengthStringMessage {
-    // TODO Doesn't work yet because constrained string types don't implement
```
david-perez, line 879: I removed this TODO because it was stale.

david-perez:
I'll hijack this to ask about the below. Shouldn't `git diff` always be a subset of `git ls-tree -r HEAD`? We're currently running the lint against changed files twice, which duplicates the output.

https://github.com/awslabs/smithy-rs/blob/66698a8ced83df391769618566b07951b8c039e0/tools/sdk-lints/src/main.rs#L68-L80

jdisanti:
> I'll hijack this to ask about the below. Shouldn't `git diff` always be a subset of `git ls-tree -r HEAD`? We're currently running the lint against changed files twice, which duplicates the output.

Yeah, that is weird. @rcoh, was this intentional or an oversight?

jdisanti (approved):
Thanks for improving this!

rcoh:
Hmm probably an oversight

rcoh:
Maybe new files only in the diff don't show up?

LukeMathWalker (approved):

tools/sdk-lints/src/todos.rs
```diff
@@ -23,6 +23,13 @@ impl Lint for TodosHaveContext {
             extension
                 .map(|ext| {
                     ext.eq_ignore_ascii_case("rs")
+                        || ext.eq_ignore_ascii_case("toml")
+                        || ext.eq_ignore_ascii_case("txt")
+                        // TODO(https://github.com/awslabs/smithy-rs/issues/2077)
+                        // || ext.eq_ignore_ascii_case("md")
+                        || ext.eq_ignore_ascii_case("sh")
+                        || ext.eq_ignore_ascii_case("py")
+                        || ext.eq_ignore_ascii_case("smithy")
```
hlbarber, line 32: `md` is a candidate for this too?

hlbarber (approved):

tools/sdk-lints/src/todos.rs
david-perez: Yes, but we can't enable it because of the [linked issue](https://github.com/awslabs/smithy-rs/issues/2077).

PR #2079, opened 2022-12-08
fix: make alternative async runtime timeout assert more permissive

Velfi:
For Russell 🙂 

----


rcoh (commented):
it's failed up to 50ms on my machine...should we do 100?

rcoh (approved):

jdisanti (approved):
Yay! I won't have to retry this in CI anymore 😄

PR #2080, opened 2022-12-08
Fix clippy lints and run formatter

Velfi:
I noticed while working on the concurrency tests that we had neglected to run clippy in a few places. I also ran `cargo fmt` for good measure. This change doesn't change any behavior.

---

### DIFF
--- a/aws/sdk/integration-tests/kms/tests/sensitive-it.rs
+++ b/aws/sdk/integration-tests/kms/tests/sensitive-it.rs
@@ -55,7 +55,12 @@ async fn client_is_debug() {
 async fn client_is_clone() {
     let conf = kms::Config::builder().build();
     let client = kms::Client::from_conf(conf);
-    let _ = client.clone();
+
+    fn is_clone(it: impl Clone) {
+        drop(it)
+    }
+
+    is_clone(client);
 }
 
 #[test]
--- a/rust-runtime/aws-smithy-http-server/src/protocols.rs
+++ b/rust-runtime/aws-smithy-http-server/src/protocols.rs
@@ -7,6 +7,7 @@
 use crate::rejection::MissingContentTypeReason;
 #[allow(deprecated)]
 use crate::request::RequestParts;
+use http::HeaderMap;
 
 /// When there are no modeled inputs,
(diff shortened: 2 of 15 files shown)


aws/sdk/integration-tests/kms/tests/sensitive-it.rs
```diff
@@ -55,7 +55,12 @@ async fn client_is_debug() {
 async fn client_is_clone() {
     let conf = kms::Config::builder().build();
     let client = kms::Client::from_conf(conf);
-    let _ = client.clone();
+
+    fn is_clone(it: impl Clone) {
+        drop(it)
+    }
+
+    is_clone(client);
```
Velfi, line 63: Clippy was mad that we cloned and didn't use the client it but it's fine if we use a function and drop the client.

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -35,6 +30,16 @@ pub fn content_type_header_empty_body_no_modeled_input<B>(
...
+fn parse_content_type(headers: &HeaderMap) -> Result<mime::Mime, MissingContentTypeReason> {
+    headers
+        .get(http::header::CONTENT_TYPE)
+        .unwrap() // The header is present, `unwrap` will not panic.
+        .to_str()
+        .map_err(MissingContentTypeReason::ToStrError)?
+        .parse::<mime::Mime>()
+        .map_err(MissingContentTypeReason::MimeParseError)
+}
+
```
Velfi, line 42: IntelliJ was complaining about code duplication here so I extracted the common code

rust-runtime/aws-smithy-http-server/src/protocols.rs
```diff
@@ -166,7 +165,7 @@ mod tests {
 
     #[test]
     fn check_invalid_content_type() {
-        let invalid = vec!["application/ajson", "text/xml"];
+        let invalid = vec!["application/jason", "text/xml"];
```
Velfi, line 168: IntelliJ didn't like the spelling of "ajson" and I think "jason" is funnier anyways

jdisanti (approved):

PR #2083, opened 2022-12-09
Add RFC for improving request ID access in SDK clients

jdisanti:
## Motivation and Context
This RFC proposes changes that will make it easier to access request IDs when using the AWS Rust SDK.

[Rendered](https://github.com/awslabs/smithy-rs/blob/jdisanti-rfc-request-id-access/design/src/rfcs/rfc0027_improve_sdk_request_id_access.md)

----


design/src/rfcs/rfc0027_improve_sdk_request_id_access.md
```diff
@@ -0,0 +1,244 @@
...
+
+1. For error cases where the response parsed successfully, the request ID can be retrieved
+   via accessor method on operation error. This also works for unmodeled errors so long as
+   the response parsing succeeds.
+2. For error cases where a response was received but parsing fails, the response headers
+   can be retrieved from the raw response on the error, but customers have to manually extract
+   the request ID from those headers (there's no convenient accessor method).
+3. For all error cases where the request ID header was sent in the response, customers can
+   call `SdkError::into_service_error` to transform the `SdkError` into an operation error,
+   which has a `request_id` accessor on it.
```
Velfi, line 18: does "sent in the response" mean it can be sent as a header OR in the body?

design/src/rfcs/rfc0027_improve_sdk_request_id_access.md
```diff
@@ -0,0 +1,244 @@
...
+4. For success cases, the customer can't retrieve the request ID at all if they use the fluent
+   client. Instead, they must manually make the operation and call the underlying Smithy client
+   so that they have access to `SdkSuccess`, which provides the raw response where the request ID
+   can be manually extracted from headers.
+
+Only one of these mechanisms is convenient and ergonomic. The rest need considerable improvements.
+Additionally, the request ID should be attached to tracing events where possible so that enabling
+debug logging reveals the request IDs without any code changes being necessary.
```
Velfi, line 26:
```suggestion
debug logging reveals the request IDs by default.
```

design/src/rfcs/rfc0027_improve_sdk_request_id_access.md
jdisanti: It says, "request ID _header_".

design/src/rfcs/rfc0027_improve_sdk_request_id_access.md
```diff
@@ -0,0 +1,244 @@
...
+accessor without any confusion. In cases where it is already modeled and is the
+same concept, customers will likely just use it and not even realize they didn't
+import the trait. The only concern is future cases where it is modeled as a
+separate concept, and as long as customers don't import `RequestId` for something
+else in the same file, that confusion can be avoided.
+
+In order to implement `RequestId` for outputs, either the original response needs
+to be stored on the output, or the request ID needs to be extracted earlier and
+stored on the output. The latter will lead to a small amount of header lookup
+code duplication.
```
rcoh, line 123: although we can make an inlineable to do it—this is probably also where we can deal with any service-specific request ID behavior.

design/src/rfcs/rfc0027_improve_sdk_request_id_access.md
```diff
@@ -0,0 +1,244 @@
...
+
+In either case, the `StructureGenerator` needs to be customized in `sdk-codegen`
+(Appendix B outlines an alternative approach to this and why it was dismissed).
+This will be done by adding customization hooks to `StructureGenerator` similar
+to the ones for `ServiceConfigGenerator` so that a `sdk-codegen` decorator can
+conditionally add fields and functions to any generated structs. A hook will
+also be needed to additional trait impl blocks.
```
rcoh, line 130: would be interested in the design here—I'd probably envision something like the following but I'm curious about your thoughts

```rust
struct Meta { ... } // private struct shared between all operations, this could be adapted to hold/expose the entire HTTP response, e.g.

struct Foo { 
  /* regular fields */
  // private field
  _meta: Meta
}
```

rcoh (approved):
LGTM. I'm a little hesitant about this approach because of how it sidesteps composition in favor of duplication but I'm interested to see how it turns out.

design/src/rfcs/rfc0027_improve_sdk_request_id_access.md
```diff
@@ -0,0 +1,244 @@
...
+that takes an error from any operation and logs a request ID, so it's still not ideal.
+
+The `aws-http` crate needs to have a `RequestId` trait on it to facilitate generic
+request ID retrieval:
+
+```rust
+pub trait RequestId {
+    /// Returns the request ID if it's available.
+    fn request_id(&self) -> Option<&str>;
+}
```
Velfi, line 74: I like this UX. From a user's perspective, I think it's what I'd try to call before even reading the docs. I think that we one thing that could possibly be expanded upon is that we may need to educate users on when to expect this to return `Some` vs. `None`. In some cases, returning `None` is expected but users may find code paths that produce an unexpected `None` and we need to direct them to file a bug report for that.

Velfi (approved):

ysaito1001 (approved):

design/src/rfcs/rfc0027_improve_sdk_request_id_access.md
```diff
@@ -0,0 +1,244 @@
+RFC: Improving access to request IDs in SDK clients
+===================================================
+
+> Status: RFC
```
ysaito1001, line 4: We can update the status to something like

Implemented in [smithy-rs#2129](https://github.com/awslabs/smithy-rs/pull/2129)

PR #2097, opened 2022-12-13
Small refactor in `EnumGenerator.kt`

david-perez:
Some Rust code blocks have been trivially made more uninterrupted and
DRYer.

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -204,35 +205,44 @@ open class EnumGenerator(
...
+                impl $UnknownVariantValue {
+                    pub(crate) fn as_str(&self) -> &str {
+                        &self.0
+                    }
                 }
-            }
+                """,
+            )
         }
     }
```
jjant, line 247: Can this be written as this (the syntax might be slightly off)?
```
rustTemplate("""
                /// Opaque struct used as inner data for the `Unknown` variant defined in enums in
                /// the crate.
                /// While this is not intended to be used directly, it is marked as `pub` because it is
                /// part of the enums that are public interface.
                #{Metadata:W}
                struct $UnknownVariantValue(pub(crate) String);
                impl $UnknownVariantValue {
                    pub(crate) fn as_str(&self) -> &str {
                        &self.0
                    }
                }
""",
"Metadata" to meta::render
)
```

jjant (approved):
Nice!
I think we might be able to condense the block below as well.

jdisanti (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -299,7 +309,7 @@ private fun RustWriter.renderForwardCompatibilityNote(
         variant in a current version of SDK, your code should continue to work when you
         upgrade SDK to a future version in which the enum does include a variant for that
         feature.
-        """.trimIndent(),
+        """,
```
ysaito1001, line 312: Interesting that `trimIdent` can be removed 🤔  I had to add it only because a pre-commit hook complained about it.

ysaito1001 (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
jjant: I think they can always be removed, and perhaps the pre-commit hook error message is not clear/just wrong.

david-perez:
Closing because of staleness; these improvements were made in https://github.com/awslabs/smithy-rs/pull/2334.

PR #2099, opened 2022-12-14
Break up `RustCodegenDecorator`

jdisanti:
## Motivation and Context
This change creates `ClientCodegenDecorator` and
`ServerCodegenDecorator` in `codegen-client` and `codegen-server` respectively to replace `RustCodegenDecorator`. Client/server equivalents are created to replace `CombinedCodegenDecorator` as well.

This eliminates the need for the `supportsCodegenContext` method since the decorator interface is no longer generic, so the `ServiceLoader` is now powerful enough to differentiate.

The largest benefit, however, is that now clients and servers can have separate customizations.

## Checklist

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customize/ClientCodegenDecorator.kt
```diff
@@ -0,0 +1,172 @@
...
+            )
+
+            val filteredDecorators = decorators.asSequence()
+                .onEach { logger.info("Discovered Codegen Decorator: ${it.javaClass.name}") }
+                .filter { it.classpathDiscoverable() }
+                .onEach { logger.info("Adding Codegen Decorator: ${it.javaClass.name}") }
+                .toList()
+            return CombinedClientCodegenDecorator(filteredDecorators + extras)
+        }
+    }
```
crisidev, line 171: Wonderful, this is much more readable than what we had before with generics!

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customize/ServerCodegenDecorator.kt
```diff
@@ -0,0 +1,132 @@
...
+import java.util.logging.Logger
+
+typealias ServerProtocolMap = ProtocolMap<ServerProtocolGenerator, ServerCodegenContext>
+
+/**
+ * [ServerCodegenDecorator] allows downstream users to customize code generation.
+ *
+ * For example, AWS-specific code generation generates customizations required to support
+ * AWS services. A different downstream customer may wish to add a different set of derive
+ * attributes to the generated classes.
```
crisidev, line 29: Nit: the AWS example doesn't really fit for the server:

```suggestion
 * For example, a downstream customer may wish to add a different set of derive
 * attributes to the generated classes.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customize/ServerCodegenDecorator.kt
```diff
@@ -0,0 +1,132 @@
...
+ */
+interface ServerCodegenDecorator {
```
crisidev, line 31: I think it's great that now the interface for the Server Codegen is smaller and completely customizable without interfering with the Client..

unexge (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customize/ServerCodegenDecorator.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    fun transformModel(service: ServiceShape, model: Model): Model = model
+}
+
+/**
+ * [CombinedServerCodegenDecorator] merges the results of multiple decorators into a single decorator.
+ *
+ * This makes the actual concrete codegen simpler by not needing to deal with multiple separate decorators.
+ */
+class CombinedServerCodegenDecorator(decorators: List<ServerCodegenDecorator>) : ServerCodegenDecorator {
```
david-perez, line 72: nit: I'd move the common method implementations of `CombinedServerCodegenDecorator` and `CombinedClientCodegenDecorator` somewhere common e.g. a `CombinedCodegenDecorator.kt` file to make it a bit DRYer and ensure we don't deviate.

Same for the calculation of `filteredDecorators` below.

david-perez (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/CodegenIntegrationTest.kt
```diff
@@ -68,15 +66,15 @@ abstract class DecoratableBuildPlugin<T, C : CodegenContext> : SmithyBuildPlugin
 }
 
 // TODO(https://github.com/awslabs/smithy-rs/issues/1864): move to core once CodegenDecorator is in core
-private fun <T, C : CodegenContext> codegenIntegrationTest(
+private fun codegenIntegrationTest(
     model: Model,
-    buildPlugin: DecoratableBuildPlugin<T, C>,
-    additionalDecorators: List<RustCodegenDecorator<T, C>>,
+    buildPlugin: DecoratableBuildPlugin,
+    additionalDecorators: List<ClientCodegenDecorator>,
```
rcoh, line 72: this should be a generic so that this function can also be used by the server if possible

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/CodegenIntegrationTest.kt
```diff
@@ -56,10 +54,10 @@ fun clientIntegrationTest(
  * This exists to allow tests to easily customize the _real_ build without needing to list out customizations
  * or attempt to manually discover them from the path
  */
-abstract class DecoratableBuildPlugin<T, C : CodegenContext> : SmithyBuildPlugin {
+abstract class DecoratableBuildPlugin : SmithyBuildPlugin {
```
rcoh, line 57: this should probably keep a generic

rcoh (approved):
LGTM—the only issue is it prevents the integration test helper from being usable for the serve . It would be nice if we could unit those but probably not mission critical

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/CodegenIntegrationTest.kt
jdisanti: It doesn't look like `codegen-server` uses this at all. I'm more inclined to move it into the `test/` directory and remove the TODO comment about moving the decorator into core.

jdisanti: Looks like the `sdk-codegen` module uses it, so it needs to stay in `src/`.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/CodegenIntegrationTest.kt
jdisanti: It's only used by client tests.

rcoh (approved):

PR #2105, opened 2022-12-15
Break `codegen-server`'s dependency on `codegen-client`

### LINKED ISSUE #1442: `EventStreamUnmarshallerGeneratorTest` is running misconfigured tests for the server
The tests in `EventStreamUnmarshallerGeneratorTest.kt` are being run for the server because in `EventStreamTestTools.kt` we have:

https://github.com/awslabs/smithy-rs/blob/242e6bcd677b5600f27106f8bc45f411c3c429da/codegen/src/test/kotlin/software/amazon/smithy/rust/codegen/smithy/protocols/EventStreamTestTools.kt#L306-L306

That is, each test case is duplicated: one with `CodegenTarget.CLIENT`, one with `CodegenTarget.SERVER`.

The problem is this is incorrect, because both test types are using the `testSymbolProvider` from `TestHelpers.kt`

https://github.com/awslabs/smithy-rs/blob/242e6bcd677b5600f27106f8bc45f411c3c429da/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/testutil/TestHelpers.kt#L66-L71

That is the symbol provider that the _client_ uses, which is very different from the server one (for example, the client generates `Option`al types for `@required` members, and the server doesn't).

It should instead be using `serverTestSymbolProvider`:

https://github.com/awslabs/smithy-rs/blob/242e6bcd677b5600f27106f8bc45f411c3c429da/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/testutil/ServerTestHelpers.kt#L30-L35

However, we can't make it use that symbol provider, because the server symbol provider is in `ServerTestHelpers.kt`, part of the `codegen-server` subproject's test sources. But `EventStreamUnmarshallerGeneratorTest.kt` is in the `codegen` subproject. `codegen` cannot depend on `codegen-server`, it's the other way around.

One way to fix this would be to have `codegen-server`'s test sources depend on `codegen`'s test sources. We would rewrite the test case provider in the server subproject to use the `serverTestSymbolProvider`, and then the test would just delegate to `EventStreamUnmarshallerGeneratorTest.kt` passing it the modified test case. I've tried looking into this and it's not straightforward.

* [How to share test classes across Gradle projects?](https://stackoverflow.com/questions/31814812/how-to-share-test-classes-across-gradle-projects)
* [Multi-project test dependencies with gradle](https://stackoverflow.com/questions/5644011/multi-project-test-dependencies-with-gradle)

The examples above are in Groovy; I don't know how to do it in the Kotlin DSL. Gradle's official documentation [seems to recommend the `testFixtures` approach](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures) for sharing test code across projects, but in our case we would not only want to share the test cases but also the test code from `EventStreamUnmarshallerGeneratorTest.kt`. I don't know if that's abusing the `testFixtures` pattern.

To me it seems like this is yet another indication that we should restructure smithy-rs to have _three_ subprojects: a core library, the client plugin, and the server plugin. Both the client and server plugins would depend on the core library's implementation and test code.

### LINKED ISSUE #1864: Eliminate dependency on `codegen-client` from `codegen-server`
Now that `codegen-core` is a thing, we should work towards eliminating the dependency on `codegen-client` from `codegen-server`. So the goal is to drive this number down to zero:

```
find . -name "*.kt" | xargs grep -n "software.amazon.smithy.rust.codegen.client" | wc -l
```

These are candidates to be moved to `codegen-core` and be made more general if possible, reducing our usage on `CodegenTarget` too: https://github.com/awslabs/smithy-rs/issues/1765.

jdisanti:
## Motivation and Context
This PR and fixes #1864.

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
```diff
@@ -0,0 +1,855 @@
...
+                    .build()
+                ),
+                expect_event(result.unwrap())
+            );
+            """,
+        )
+
+        val (someError, kindSuffix) = when (codegenTarget) {
+            CodegenTarget.CLIENT -> listOf("TestStreamErrorKind::SomeError", ".kind")
+            CodegenTarget.SERVER -> listOf("TestStreamError::SomeError", "")
```
david-perez, line 603:
```suggestion
            CodegenTarget.CLIENT -> "TestStreamErrorKind::SomeError" to ".kind"
            CodegenTarget.SERVER -> "TestStreamError::SomeError" to ""
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/parse/EventStreamUnmarshallerGeneratorTest.kt
```diff
@@ -5,302 +5,45 @@
...
-            test.serviceShape,
-            ShapeId.from(testCase.protocolShapeId),
-            testRustSettings(),
-            target = testCase.target,
+        EventStreamTestTools.runTestCase(
+            testCase,
+            CodegenTarget.SERVER,
+            { model -> serverTestSymbolProvider(model) },
+            { codegenContext, test, protocol ->
+                fun builderSymbol(shape: StructureShape): Symbol = shape.builderSymbol(codegenContext.symbolProvider)
```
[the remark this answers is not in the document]
david-perez: In the server we also need to exercise the two server builders, depending on whether `codegenContext.settings.codegenConfig.publicConstrainedTypes` is `true` or `false`; see my comment above.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/parse/EventStreamUnmarshallerGeneratorTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import software.amazon.smithy.rust.codegen.core.smithy.generators.builderSymbol
+import software.amazon.smithy.rust.codegen.core.smithy.protocols.parse.EventStreamUnmarshallerGenerator
+import software.amazon.smithy.rust.codegen.core.testutil.EventStreamTestModels
+import software.amazon.smithy.rust.codegen.core.testutil.EventStreamTestTools
+import java.util.stream.Stream
+
+class UnmarshallTestCasesProvider : ArgumentsProvider {
+    override fun provideArguments(context: ExtensionContext?): Stream<out Arguments> =
+        EventStreamTestModels.TEST_CASES.map { Arguments.of(it) }.stream()
+}
```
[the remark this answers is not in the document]
david-perez: Perhaps in `EventStreamTestTools`?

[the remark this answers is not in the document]
david-perez: Same goes for `software.amazon.smithy.rust.codegen.server.smithy.protocols.serialize.MarshallTestCasesProvider` and `software.amazon.smithy.rust.codegen.client.smithy.protocols.serialize.MarshallTestCasesProvider`.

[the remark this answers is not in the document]
david-perez: Although in the server we don't support `ec2Query` or `awsQuery`, and we very badly support `restXml`, so maybe we want to filter those out in the server too.

Although then again, if these specific event stream tests pass in the server I guess there's no harm in leaving them in so as to not regress in case we want to support those protocols in the future, so I'd just make the test case providers identical for both subprojects.

david-perez (changes requested):
The builder used in the server tests is still the client one. In the server we need to exercise two builder generators, `ServerBuilderGenerator` and `ServerBuilderGeneratorWithoutPublicConstrainedTypes`, depending on the value of `codegenContext.settings.codegenConfig.publicConstrainedTypes`.

It would also be nice to move `ServerCombinedErrorGenerator` from `codegen-core` to `codegen-server`.

---

I think it should be possible to do the above by passing in more closures in the `test` functions. As an aside, what do you think of us setting up a `codegen-core-test` subproject? It seems odd to have the subproject-specific unit tests depend on the entirety of the `codegen-core` module, when what we really want is to depend on shared test code only (e.g. `EventStreamTestTools`, the test helpers, a central Smithy model bank we can reuse so that we don't duplicate models etc.) I briefly looked into setting up a shared test subproject with Gradle but couldn't figure it out (see the two StackOverflow links at the end of https://github.com/awslabs/smithy-rs/issues/1442#issue-1261778111).

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/parse/EventStreamUnmarshallerGeneratorTest.kt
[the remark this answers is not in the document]
jdisanti: I kept these argument providers in `codegen-client` and `codegen-server` to avoid needing to depend on junit in `codegen-core`. However, if we set up a `codegen-core-test` module, then that won't be a problem anymore.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
```diff
@@ -0,0 +1,855 @@
...
+        val project = TestWorkspace.testProject(symbolProvider)
+        val operationSymbol = symbolProvider.toSymbol(operationShape)
+        project.withModule(ErrorsModule) {
+            val errors = model.shapes()
+                .filter { shape -> shape.isStructureShape && shape.hasTrait<ErrorTrait>() }
+                .map { it.asStructureShape().get() }
+                .toList()
+            when (codegenTarget) {
+                CodegenTarget.CLIENT -> CombinedErrorGenerator(model, symbolProvider, operationSymbol, errors).render(this)
+                CodegenTarget.SERVER -> ServerCombinedErrorGenerator(model, symbolProvider, operationSymbol, errors).render(this)
```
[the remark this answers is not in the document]
jdisanti: I actually did move `ServerCombinedErrorGenerator` into `codegen-server` in https://github.com/awslabs/smithy-rs/pull/2107. I didn't want to make the PR too large.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/parse/EventStreamUnmarshallerGeneratorTest.kt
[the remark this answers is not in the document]
jdisanti: I played around with creating `codegen-core-test`, and it seems like it _might be_ the right path forward long-term. However, it's a large scale refactor that has rippling effects that will require making `codegen-client-test` a Kotlin module so that `sdk-codegen` can depend on test-only code from `codegen-client`.

To do it nicely, we're going to have to move all of `codegen-core`'s unit tests into `codegen-core-test/src/test`, while test code that is intended to be shared will need to reside in `codegen-core-test/src/main`. I was able to create and populate `codegen-core-test` easily enough, but then code in `codegen-client/src/main` no longer compiles due to requiring test-only code in `src/main` so that `sdk-codegen` can reuse that client-specific test-only code.

One consequence of going through with this is that there will be no way to unit test code that is `internal`, and if we decide to reuse the existing `codegen-client-test` and `codegen-server-test` modules, they will now run both unit tests and codegen tests (but maybe that's desirable?).

I think I'm going to punt on this for this PR.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
```diff
@@ -0,0 +1,855 @@
...
+            }
+            for (shape in model.shapes().filter { shape -> shape.isStructureShape && shape.hasTrait<ErrorTrait>() }) {
+                StructureGenerator(model, symbolProvider, this, shape as StructureShape).render(codegenTarget)
+                val builderGen = BuilderGenerator(model, symbolProvider, shape)
```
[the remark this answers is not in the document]
jdisanti: I see. I understand what was broken before now. I was able to restructure everything as you described, but I'm not familiar enough with the server to get the tests passing with `publicConstrainedTypes = false`. I will update this PR's description to say it is not fixing #1442 and leave a couple of TODO comments to finish this work.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
```diff
@@ -30,23 +31,18 @@ private fun hasStreamingOperations(model: Model): Boolean {
...
+private fun structUnionMembersMatchPredicate(model: Model, predicate: (Shape) -> Boolean): Boolean =
+    model.structureShapes.any { structure ->
+        structure.members().any { member -> predicate(model.expectShape(member.target)) }
     } || model.unionShapes.any { union ->
-        union.members().any { member -> model.expectShape(member.target).isBlobShape }
+        union.members().any { member -> predicate(model.expectShape(member.target)) }
     }
-}
+
+/** Returns true if the model has any blob shapes or members */
```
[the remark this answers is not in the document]
jdisanti: Yeah, I think you're right. The implementation is bugged. Filed #2111 to track fixing the logic. Will fix the comment here though.

jdisanti:
@david-perez - Thanks for the thorough feedback! I think I incorporated everything, or left a reply explaining why not. Let me know if I missed something or if you disagree on a point.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/eventstream/EventStreamUnmarshallerGeneratorTest.kt
```diff
@@ -0,0 +1,94 @@
...
+                        ::builderSymbol,
+                    ).render()
+                }
+
+                override fun renderBuilderForShape(
+                    writer: RustWriter,
+                    codegenContext: ServerCodegenContext,
+                    shape: StructureShape,
+                ) {
+                    // TODO(https://github.com/awslabs/smithy-rs/issues/1442): Use the correct builder:
```
david-perez, line 75: If you're punting on this, mention in the TODO that we need to test with two builders: `ServerBuilderGeneratorWithoutPublicConstrainedTypes` and `ServerBuilderGenerator`.

Same goes for the `software.amazon.smithy.rust.codegen.server.smithy.protocols.eventstream.EventStreamMarshallerGeneratorTest` class, which is only testing with `ServerBuilderGenerator`.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/eventstream/EventStreamBaseRequirements.kt
```diff
@@ -0,0 +1,53 @@
...
+import software.amazon.smithy.rust.codegen.client.testutil.clientTestRustSettings
+import software.amazon.smithy.rust.codegen.client.testutil.testSymbolProvider
+import software.amazon.smithy.rust.codegen.core.rustlang.RustWriter
+import software.amazon.smithy.rust.codegen.core.smithy.CodegenTarget
+import software.amazon.smithy.rust.codegen.core.smithy.RustSymbolProvider
+import software.amazon.smithy.rust.codegen.core.smithy.generators.BuilderGenerator
+import software.amazon.smithy.rust.codegen.core.smithy.generators.implBlock
+import software.amazon.smithy.rust.codegen.core.testutil.EventStreamTestRequirements
+
+abstract class EventStreamBaseRequirements : EventStreamTestRequirements<ClientCodegenContext> {
```
david-perez, line 23: Should we agree on a naming strategy once and for all on classes that have equivalents across `codegen-client` and `codegen-server`? _Most_ (but not all) of the classes in `codegen-server` are prefixed with `Server`, while few of the classes in `codegen-client` are prefixed with `Client`. Likewise, some classes from `codegen-core` that we inherit from the other modules are prefixed with `Core`.

My preference would be to not prefix things with `Core`, but have everything prefixed with `Client` and `Server` that comes from `codegen-client` and `codegen-server`, respectively. While redundant (the information is encoded in the classpath), I like that one can tell at a glance the purpose of things. This would only apply to classes that have equivalents across the subprojects (e.g. we have `ServerRestJson` in `codegen-server` and `RestJson` in `codegen-client`, so the latter should be renamed to `ClientRestJson`, but there's no need to prefix `ConstraintViolationSymbolProvider` with `Server`, since constraint traits are ignored in the client).

We could even attempt writing a lint that enforced the naming strategy in CI. What do you think?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
```diff
@@ -0,0 +1,173 @@
...
+        project.withModule(ErrorsModule) {
+            val errors = model.shapes()
+                .filter { shape -> shape.isStructureShape && shape.hasTrait<ErrorTrait>() }
+                .map { it.asStructureShape().get() }
+                .toList()
+            when (codegenTarget) {
+                CodegenTarget.CLIENT -> CombinedErrorGenerator(model, symbolProvider, operationSymbol, errors).render(this)
+                CodegenTarget.SERVER -> ServerCombinedErrorGenerator(model, symbolProvider, operationSymbol, errors).render(this)
+            }
+            for (shape in model.shapes().filter { shape -> shape.isStructureShape && shape.hasTrait<ErrorTrait>() }) {
```
david-perez, line 132:
```suggestion
            for (shape in model.shapes().filter { shape -> shape is StructureShape && shape.hasTrait<ErrorTrait>() }) {
```

nit: I favor `is` checks instead of using the Smithy library's provided methods because Kotlin is able to do type narrowing.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
```diff
@@ -0,0 +1,173 @@
...
+    private fun recursivelyGenerateModels(
+        model: Model,
+        symbolProvider: RustSymbolProvider,
+        shape: Shape,
+        writer: RustWriter,
+        mode: CodegenTarget,
+    ) {
+        for (member in shape.members()) {
+            val target = model.expectShape(member.target)
+            if (target is StructureShape || target is UnionShape) {
```
david-perez, line 163: Is the reason why we only render these shape types that we're not testing with more complex models? If yes, I would crash here upon encountering an unrendered shape type, lest a future dev be puzzled when adding more shapes to the event stream tests.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/eventstream/EventStreamBaseRequirements.kt
```diff
@@ -0,0 +1,52 @@
...
+abstract class EventStreamBaseRequirements : EventStreamTestRequirements<ServerCodegenContext> {
+    override fun createCodegenContext(
+        model: Model,
+        symbolProvider: RustSymbolProvider,
+        serviceShape: ServiceShape,
+        protocolShapeId: ShapeId,
+        codegenTarget: CodegenTarget,
+    ): ServerCodegenContext {
+        val settings = serverTestRustSettings()
+        val serverSymbolProviders = ServerSymbolProviders.from(
```
david-perez, line 30: Use the `software.amazon.smithy.rust.codegen.server.smithy.testutil.serverTestSymbolProviders` function.

david-perez: Or directly `serverTestCodegenContext`, since this test is only testing with `publicConstrainedTypes` enabled.

david-perez (approved):

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/eventstream/EventStreamBaseRequirements.kt
jdisanti: Thanks for the pointers! I'm not so familiar with the server testing helpers.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/eventstream/EventStreamBaseRequirements.kt
jdisanti: I'm fine with your proposed naming strategy. It definitely makes it easier to pull up the correct class in IntelliJ. I'll rename these ones.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/eventstream/EventStreamBaseRequirements.kt
jdisanti: Fixed in be77fa0df3888633117973eed62a7784ed4df258

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
jdisanti: Good call. Fixed in c608a775bf02ce33b73f6827bb12b6a8b94ad6ed.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
jdisanti: Fixed in c608a775bf02ce33b73f6827bb12b6a8b94ad6ed.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/eventstream/EventStreamBaseRequirements.kt
jdisanti: Fixed in 5c73370f087e40863b0960118bfb49e07325d926.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/eventstream/EventStreamUnmarshallerGeneratorTest.kt
jdisanti: Fixed in a47597fccc413b489e6c9d3f93dc3c2149a2036c.

jdisanti: Fixed the marshaller side of things in 0b77d8ffe7380d07a0ff648ae4e526ecc28616d7.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/eventstream/ClientEventStreamBaseRequirements.kt
```diff
@@ -0,0 +1,49 @@
...
+import software.amazon.smithy.rust.codegen.client.smithy.customize.CombinedClientCodegenDecorator
+import software.amazon.smithy.rust.codegen.client.testutil.clientTestRustSettings
+import software.amazon.smithy.rust.codegen.client.testutil.testSymbolProvider
+import software.amazon.smithy.rust.codegen.core.rustlang.RustWriter
+import software.amazon.smithy.rust.codegen.core.smithy.CodegenTarget
+import software.amazon.smithy.rust.codegen.core.smithy.generators.BuilderGenerator
+import software.amazon.smithy.rust.codegen.core.smithy.generators.implBlock
+import software.amazon.smithy.rust.codegen.core.testutil.EventStreamTestRequirements
+
+abstract class ClientEventStreamBaseRequirements : EventStreamTestRequirements<ClientCodegenContext> {
```
Velfi, line 22: I feel like this should be called "ExtendedTestRequirements" or something. It feels weird that the "base" thing isn't actually the first thing in the inheritance chain.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/eventstream/ClientEventStreamMarshallerGeneratorTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import software.amazon.smithy.rust.codegen.core.testutil.EventStreamTestModels
+import software.amazon.smithy.rust.codegen.core.testutil.EventStreamTestTools
+import software.amazon.smithy.rust.codegen.core.testutil.EventStreamTestVariety
+import software.amazon.smithy.rust.codegen.core.testutil.TestEventStreamProject
+import software.amazon.smithy.rust.codegen.core.testutil.TestRuntimeConfig
+import java.util.stream.Stream
+
+class MarshallTestCasesProvider : ArgumentsProvider {
+    override fun provideArguments(context: ExtensionContext?): Stream<out Arguments> =
+        // Don't include awsQuery or ec2Query for now since marshall support for them is unimplemented
```
Velfi, line 27: Do we have an issue related to this? It'd be nice to convert this into a linked TODO

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/CrateVersionCustomization.kt
```diff
@@ -13,7 +13,7 @@ import software.amazon.smithy.rust.codegen.core.smithy.generators.LibRsSection
 /**
  * Add `PGK_VERSION` const in lib.rs to enable knowing the version of the current module
  */
-class CrateVersionGenerator : LibRsCustomization() {
+class CrateVersionCustomization : LibRsCustomization() {
```
Velfi, line 16: Thanks for fixing the names of stuff like this, I really appreciate it!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
jdisanti: Actually, I forgot to remove the `if (target is StructureShape || target is UnionShape)` while testing 🤦‍♂️ 

Correctly fixed in 38598f2e318b201d78a81d8e2c1185587f962c4e.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/eventstream/ClientEventStreamBaseRequirements.kt
jdisanti: Its parent is an interface, so I think "base" actually works fine here since base classes are typically abstract classes.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
```diff
@@ -0,0 +1,174 @@
...
+            val errors = model.shapes()
+                .filter { shape -> shape.isStructureShape && shape.hasTrait<ErrorTrait>() }
+                .map { it.asStructureShape().get() }
+                .toList()
+            when (codegenTarget) {
+                CodegenTarget.CLIENT -> CombinedErrorGenerator(model, symbolProvider, operationSymbol, errors).render(this)
+                CodegenTarget.SERVER -> ServerCombinedErrorGenerator(model, symbolProvider, operationSymbol, errors).render(this)
+            }
+            for (shape in model.shapes().filter { shape -> shape is StructureShape && shape.hasTrait<ErrorTrait>() }) {
+                StructureGenerator(model, symbolProvider, this, shape as StructureShape).render(codegenTarget)
```
Velfi, line 127: It's annoying that we have to say `as StructureShape` even after checking `is StructureShape`. Why is that? Is it b/c we're doing the check inside a closure?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
```diff
@@ -0,0 +1,174 @@
...
+        model: Model,
+        symbolProvider: RustSymbolProvider,
+        shape: Shape,
+        writer: RustWriter,
+        mode: CodegenTarget,
+    ) {
+        for (member in shape.members()) {
+            if (member.target.namespace == "smithy.api") {
+                continue
+            }
```
Velfi, line 158: Why do we need to check the namespace like this. What happens if we don't?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamUnmarshallTestCases.kt
```diff
@@ -0,0 +1,274 @@
...
+                let message = msg("event", "NewUnmodeledMessageType", "application/octet-stream", b"hello, world!");
+                let result = ${format(generator)}().unmarshall(&message);
+                assert!(result.is_ok(), "expected ok, got: {:?}", result);
+                assert_eq!(
+                    TestStream::Unknown,
+                    expect_event(result.unwrap())
+                );
+                """,
+            )
+        }
```
Velfi, line 84: Stripping this out of `core` happens in another PR right?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamUnmarshallTestCases.kt
```diff
@@ -0,0 +1,274 @@
...
+                    TestStreamErrorKind::Unhandled(err) => {
+                        let message = format!("{}", aws_smithy_types::error::display::DisplayErrorContext(&err));
+                        let expected = "message: \"unmodeled error\"";
+                        assert!(message.contains(expected), "Expected '{message}' to contain '{expected}'");
+                    }
+                    kind => panic!("expected generic error, but got {:?}", kind),
+                }
+                """,
+            )
+        }
```
Velfi, line 257: Same with these. Are these addressed in another PR?

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/eventstream/ClientEventStreamMarshallerGeneratorTest.kt
jdisanti: I don't think there's actually anything to do here since those protocols don't support event streams. I will remove the "is unimplemented" piece to make it clearer.

From the [AWS Query docs](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html):
> This protocol does not support any kind of streaming requests or responses, including event streams.

The [EC2 Query docs](https://smithy.io/2.0/aws/protocols/aws-ec2-query-protocol.html) don't mention event streams at all, while all the protocol specs that support event streams mention them.

jdisanti: In fact, I'll just delete the test cases for them entirely.

Velfi (approved):
I left some questions and comments but overall it looks good.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/eventstream/ClientEventStreamMarshallerGeneratorTest.kt
jdisanti: Fixed in 80ed7a9510c43a6cbb030c047694b326de53c15a.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
jdisanti: I think the Kotlin compiler isn't smart enough to recognize the `is` check here since it happened in the filter closure.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
jdisanti: This filters out the Smithy prelude types, such as `smithy.api#Blob` or `smithy.api#String`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamUnmarshallTestCases.kt
jdisanti: There is an issue tracking removing most uses of `CodegenTarget`, but we may not be able to get rid of all of them.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/testutil/ServerTestHelpers.kt
```diff
@@ -39,12 +39,20 @@ private fun testServiceShapeFor(model: Model) =
...
         serviceShape ?: testServiceShapeFor(model),
         ServerTestSymbolVisitorConfig,
-        serverTestRustSettings((serviceShape ?: testServiceShapeFor(model)).id).codegenConfig.publicConstrainedTypes,
+        (
+            settings ?: serverTestRustSettings(
+                (serviceShape ?: testServiceShapeFor(model)).id,
+            )
+            ).codegenConfig.publicConstrainedTypes,
         RustCodegenServerPlugin::baseSymbolProvider,
     )
```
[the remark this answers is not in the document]
jdisanti: Yeah, Ktlint did that 😆

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
Velfi: Ah, ok

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/testutil/ServerTestHelpers.kt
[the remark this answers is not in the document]
Velfi: :smh:

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
david-perez: I simplified this in https://github.com/awslabs/smithy-rs/pull/2130.

jdisanti: Simplified it a little bit more in #2107 when addressing merge conflicts.

PR #2106, opened 2022-12-15
Fix examples check in CI

jdisanti:
## Motivation and Context
The examples check condition added in #2088 seems to be broken since examples are always skipped now. This PR fixes the condition.

----


jjant (approved):
Thanks, I think I'd copied an outdated example 🤦.

PR #2108, opened 2022-12-15
Move types for AWS SDK credentials to a separate crate

ysaito1001:
## Motivation and Context
This PR is purely refactoring and lays the groundwork for [this RFC](https://github.com/awslabs/smithy-rs/pull/2082) and for its [parent RFC](https://github.com/awslabs/smithy-rs/pull/1842). At a high level, we are preparing for making a set of APIs for credential providers and caching more consistent for our customers.

To that end, we will allow `ConfigLoader`, `aws_types::sdk_config::Builder`, and service clients' `config::Builder`s to have new methods `credentials_cache` (which happens in a later PR). This implies that whoever uses one of those builders will need to be able to construct a credential cache (today that's `LazyCachingCredentialsProvider`) and to pass it to the `credentials_cache` method. The issue is that the user of the builder could be in the `aws-types` crate but `LazyCachingCredentialsProvider` is in the `aws-config` crate; the former crate cannot depend on the latter.

## Description
To solve the pain point, we first need to straighten out the dependencies between Rust runtime crates. This PR introduces a new crate called `aws-credential-types` that houses credential-related traits and types. `aws-credential-types` then sits at the top of the dependency graph among Rust runtime crates (as shown in [this figure](https://user-images.githubusercontent.com/15333866/207222687-8dd2430c-2865-4161-85bc-ee1410040d38.png)). `aws-credential-types` will be the new home for `LazyCachingCredentialsProvider`, and the user can create a credential cache even from within the `aws-types` crate.

What we see in `Files changed` is a result of this proposed arrangement and placing the affected items accordingly. You can read more about it in the descriptions for 6058acb, ea0168b, and 67a73a5.

This PR implements only the refactoring portion of the said RFC (the first three items in the [changes checklist](https://github.com/awslabs/smithy-rs/blob/fb28d8c47d3ae7572cc13e430e2724da26bc9041/design/src/rfcs/rfc0027_new_home_for_cred_cache_and_builder_api_consistency.md#changes-checklist)).

## Testing
- [x] Passed all tests in CI but `iam-code-examples` that depends on a feature that has been moved from `aws-types` to `aws-credential-types` (will fix the example once this PR has been merged to main)

## Checklist

----


aws/rust-runtime/aws-credential-types/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "aws-credential-types"
+version = "0.0.0-smithy-rs-head"
+authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "Russell Cohen <rcoh@amazon.com>"]
```
jdisanti, line 4:
```suggestion
authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>"]
```

aws/rust-runtime/aws-credential-types/additional-ci
```diff
@@ -0,0 +1,20 @@
...
+echo "### Checking for duplicate dependency versions in the normal dependency graph with all features enabled"
+cargo tree -d --edges normal --all-features
+
+FEATURES=(
+    "hardcoded-credentials"
+)
+for feature in "${FEATURES[@]}"; do
+    echo "### Checking feature '${feature}'..."
+    cargo test --no-default-features --features "${feature}"
+done
```
jdisanti, line 20: We can use `cargo hack` here:
```suggestion
echo "### Testing every combination of features (excluding --all-features)"
cargo hack test --feature-powerset --exclude-all-features
```

aws/rust-runtime/aws-credential-types/src/lib.rs
```diff
@@ -3,6 +3,19 @@
...
+    rustdoc::missing_crate_level_docs,
+    unreachable_pub
+)]
+
+//! `aws-credential-types` provides the items concerned with AWS SDK credentials including:
+//! * A trait for credential providers
+//! * An opaque struct representing credentials
+//! * Concrete implementations of credential caching
+
 //! AWS SDK Credentials
```
jdisanti, line 19: You may want to take a look at how this looks rendered with `cargo doc`...
```suggestion
//! AWS SDK Credentials
//!
//! `aws-credential-types` provides the items concerned with AWS SDK credentials including:
//! * A trait for credentials providers
//! * An opaque struct representing credentials
//! * Concrete implementations of credentials caching
```

aws/rust-runtime/aws-credential-types/README.md
```diff
@@ -0,0 +1,10 @@
+# aws-credential-types
+
+This crate provides the items concerned with AWS SDK credentials including:
+* A trait for credential providers
```
jdisanti, line 4: In this context, we've been using the plural for "credentials" since the provider provides multiple credentials (access key, secret key, and sometimes a token):
```suggestion
* A trait for credentials providers
```

aws/rust-runtime/aws-credential-types/README.md
```diff
@@ -0,0 +1,10 @@
...
+* An opaque struct representing credentials
+* Concrete implementations of credential caching
```
jdisanti, line 6:
```suggestion
* Concrete implementations of credentials caching
```

aws/rust-runtime/aws-credential-types/src/lib.rs
```diff
@@ -67,13 +81,22 @@ construct credentials from hardcoded values.
...
 pub use credentials_impl::Credentials;
+pub use lazy_caching::LazyCachingCredentialsProvider;
 pub use provider::error;
 pub use provider::error::CredentialsError;
 pub use provider::future;
 pub use provider::ProvideCredentials;
 pub use provider::Result;
 pub use provider::SharedCredentialsProvider;
+pub use time_source::ManualTimeSource;
+pub use time_source::TimeSource;
```
jdisanti, line 102: I think we should consider re-exporting less at the top-level.

aws/rust-runtime/aws-inlineable/Cargo.toml
```diff
@@ -12,15 +12,16 @@ publish = false
 repository = "https://github.com/awslabs/smithy-rs"
 
 [dependencies]
+aws-credential-types = { path = "../aws-credential-types" }
 aws-endpoint = { path = "../aws-endpoint" }
 aws-http = { path = "../aws-http" }
-aws-sig-auth = { path = "../../rust-runtime/aws-sig-auth" }
+aws-sig-auth = { path = "../aws-sig-auth" }
```
jdisanti, line 18: How did this work before?

jdisanti (commented):
Looks good! I mostly have minor comments.

Also, you may want to dig into why there is a server codegen diff here. These changes seem like they shouldn't affect the server at all.

aws/rust-runtime/aws-credential-types/additional-ci
ysaito1001: Good idea! Thanks for the suggestion.

aws/rust-runtime/aws-inlineable/Cargo.toml
ysaito1001: Before and after both point to the same path. It's just that the way it reached `aws-sig-auth` before was a bit roundabout.
```
aws/rust-runtime/aws-inlineable/../../rust-runtime/aws-sig-auth
```
vs.
```
aws/rust-runtime/aws-inlineable/../aws-sig-auth
```

ysaito1001:
Thanks John for your review!

> Also, you may want to dig into why there is a server codegen diff here. These changes seem like they shouldn't affect the server at all.

I believe that was due to this PR not merging the latest `main` when that diff appeared. The latest test run kicked off as part of 3f4f67a no longer shows a server codegen diff.

aws/rust-runtime/aws-credential-types/src/lib.rs
ysaito1001: Absolutely. Reduced the number of re-exports in a1f1f78.

aws/rust-runtime/aws-credential-types/README.md
```diff
@@ -0,0 +1,10 @@
+# aws-credential-types
+
+This crate provides the items concerned with AWS SDK credentials including:
```
Velfi, line 3: Nit: I think "types" is a better word for this
```suggestion
This crate provides types concerned with AWS SDK credentials including:
```

aws/rust-runtime/aws-credential-types/src/time_source.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+// In the future, if needed we can add a time source trait, however, the manual time source
+// should cover most test use cases.
+#[derive(Debug, Clone)]
+enum Inner {
+    Real,
+    Manual(ManualTimeSource),
+}
```
Velfi, line 125: I'm bikeshedding here but what about `Default` and `Testing(TestingTimeSource)` instead? Do we expect people to replace the timesource in a production environment? If so, why?

Per your comment about the trait, I'm thinking the future variant would be named `Dyn(Box<dyn TimeSource>)`.

aws/rust-runtime/aws-credential-types/src/time_source.rs
```diff
@@ -0,0 +1,145 @@
...
+        let real = TimeSource::real();
+        // no panics
+        let _ = real.now();
+
+        let mut manual = ManualTimeSource::new(UNIX_EPOCH);
+        let ts = TimeSource::manual(&manual);
+        assert_eq!(ts.now(), UNIX_EPOCH);
+        manual.advance(Duration::from_secs(10));
+        assert_eq!(ts.now(), UNIX_EPOCH + Duration::from_secs(10));
+    }
```
Velfi, line 144: Nit: imo this should be two tests

Velfi (approved):

aws/rust-runtime/aws-credential-types/src/time_source.rs
ysaito1001: Addressed in c11069f.

aws/rust-runtime/aws-credential-types/src/time_source.rs
ysaito1001: Thanks, I like the new names.
> Do we expect people to replace the timesource in a production environment?

Hmm, I cannot think of such a use case (at least for now).

Renamed the variants to `Default` and `Testing(TestingTimeSource)` in c64684c.

PR #2122, opened 2022-12-21
Ensure SDK credential cache type safety

ysaito1001:
## Motivation and Context
Implements [RFC 28](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0028_sdk_credential_cache_type_safety.md). Built on top of #2108.

## Description
This PR improves APIs around credentials caching and credentials provider. The credentials cache previously named `LazyCachingCredentialsProvider` has been renamed to `LazyCredentialsCache` and no longer treated as a credentials provider. Customers do not interact with `LazyCredentialsCache` directly, and instead use  `CredentialsCache` that understands different flavors of credentials caching, which only supports `Lazy` credentials cache as of today (12/23/2022).

As a result of this change, a new builder method called `credentials_cache` has been introduced to `aws_config::ConfigLoader`, `aws_types::sdk_config::Builder`, and the builder for a service-specific `config::Builder`. This method takes an owned `CredentialsCache` as input, and the `load` method of a service-specific builder is a single-sourced place to create an actual credentials cache `LazyCredentialsCache` out of the `CredentialsCache` and the credentials provider at hand. The primary use cases affected have been documented in `CHANGELOG.next.toml`.

Now that the `load` method of a service-specific builder combines a credentials cache and a credentials provider, ensuring `LazyCredentialsCache` is on by default, no credentials provider needs to store a credentials cache explicitly as previously done by `DefaultCredentialsChain` or `AssumeRoleProvider`; those providers do not store a credentials cache anymore.

## Testing
- [x] Passed tests in CI

## Checklist

- [x] I have included examples in `CHANGELOG.next.toml` showing how to use `credentials_provider` and `credentials_cache` methods on `ConfigLoader`.

----


aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -466,12 +488,20 @@ mod loader {
                 ))
             };
 
+            let credentials_cache = if let Some(cache) = self.credentials_cache {
+                cache
+            } else {
+                let mut builder = CredentialsCache::lazy_builder().time_source(conf.time_source());
+                builder.set_sleep(conf.sleep());
+                builder.into_credentials_cache()
+            };
```
Velfi, line 497: IMO, using `.unwrap_or_else()` is preferred to having two assignments on one line.

aws/rust-runtime/aws-credential-types/src/cache/lazy_caching.rs
```diff
@@ -120,36 +115,25 @@ mod builder {
...
-    /// let provider = LazyCachingCredentialsProvider::builder()
-    ///     .load(provide_credentials_fn(|| async {
-    ///         // An async process to retrieve credentials would go here:
-    ///         Ok(Credentials::new("example", "example", None, None, "my_provider_name"))
-    ///     }))
-    ///     .build();
-    /// ```
-    #[derive(Debug, Default)]
+    /// Callers outside of this crate cannot call `build` directly. They can instead call
+    /// `into_credentials_cache` to obtain a [`CredentialsCache`]. Its `create_cache` then calls
```
Velfi, line 131: Does the `CredentialCache` link here work? I can't remember if doc.rs handles links containing backticks correctly.

aws/rust-runtime/aws-http/src/auth.rs
```diff
@@ -186,14 +183,19 @@ mod tests {
     #[tokio::test]
     async fn async_map_request_apply_populates_credentials() {
         let mut req = operation::Request::new(http::Request::new(SdkBody::from("some body")));
-        set_provider(
+        let credentials_cache = CredentialsCache::lazy_builder()
+            .into_credentials_cache()
+            .create_cache(Arc::new(provide_credentials_fn(|| async {
+                Ok(Credentials::new("test", "test", None, None, "test"))
```
Velfi, line 189: I'm thinking we should consider adding a `test` method to `Credentials` with a `#[cfg(test)]` gate. The function would return these credentials or similar.

aws/rust-runtime/aws-types/src/sdk_config.rs
```diff
@@ -313,23 +316,61 @@ impl Builder {
...
+    /// let config = SdkConfig::builder()
+    ///     .credentials_cache(CredentialsCache::lazy())
+    ///     .build();
+    /// ```
+    pub fn credentials_cache(mut self, cache: CredentialsCache) -> Self {
+        self.set_credentials_cache(Some(cache));
+        self
+    }
+
+    /// Set the credentials cache for the builder
```
Velfi, line 334: We should consider linking here to types/docs that describe in greater depth how configuring credential caching works.

aws/sdk/integration-tests/s3/tests/user-agent-app-name.rs
```diff
@@ -4,15 +4,15 @@
...
 
 #[tokio::test]
 async fn user_agent_app_name() {
     let (conn, rcvr) = capture_request(None);
     let sdk_config = SdkConfig::builder()
-        .credentials_provider(SharedCredentialsProvider::new(Credentials::new(
+        .credentials_provider(Arc::new(Credentials::new(
             "ANOTREAL",
             "notrealrnrELgWzOk3IfjzDKtFBhDby",
             Some("notarealsessiontoken".to_string()),
```
Velfi, line 18: This is another place where we could be calling a `Credentials::test` method to create creds.

aws/rust-runtime/aws-config/src/lib.rs
ysaito1001: Agreed. Addressed in af61b54. Tried to convert other occurrences in the same method but they slightly differ and could not be converted, e.g. some of them have `.await` in their else blocks.

aws/rust-runtime/aws-credential-types/src/cache/lazy_caching.rs
ysaito1001: Yes, confirmed it works.

aws/rust-runtime/aws-types/src/sdk_config.rs
ysaito1001: Addressed in 1368f00.

aws/rust-runtime/aws-http/src/auth.rs
ysaito1001: That's good idea. I'll tackle it in a separate PR as it will probably affect many call sites currently creating a test `Credentials`.

aws/sdk/integration-tests/s3/tests/user-agent-app-name.rs
ysaito1001: Will address it in a separate PR.

aws/rust-runtime/aws-http/src/auth.rs
ysaito1001: [PR](https://github.com/awslabs/smithy-rs/pull/2145) created.

aws/rust-runtime/aws-config/src/sts/assume_role.rs
```diff
@@ -191,9 +191,11 @@ impl AssumeRoleProviderBuilder {
     /// Build a credentials provider for this role authorized by the given `provider`.
     pub fn build(self, provider: impl ProvideCredentials + 'static) -> AssumeRoleProvider {
         let conf = self.conf.unwrap_or_default();
-        let credentials_cache = CredentialsCache::lazy_builder()
-            .configure(conf.sleep(), conf.time_source())
-            .into_credentials_cache();
+
+        let mut builder = CredentialsCache::lazy_builder().time_source(conf.time_source());
```
jdisanti, line 195: This cache should be configurable when creating an `AssumeRoleProvider`.

aws/rust-runtime/aws-http/src/auth.rs
```diff
@@ -3,27 +3,26 @@
...
 use aws_smithy_http::operation::Request;
 use aws_smithy_http::property_bag::PropertyBag;
 use std::future::Future;
 use std::pin::Pin;
 
-/// Sets the credentials provider in the given property bag.
-pub fn set_provider(bag: &mut PropertyBag, provider: SharedCredentialsProvider) {
-    bag.insert(provider);
+/// Sets the credentials cache in the given property bag.
+pub fn set_credentials_cache(bag: &mut PropertyBag, cache: SharedCredentialsCache) {
```
jdisanti, line 15: This should get called out as its own breaking change, as this is called directly by some customers. We should also make sure it's possible to directly create the `SharedCredentialsCache` without using `aws-config` (would be good to include as an example for the changelog).

jdisanti (commented):
Very nice work on this! I especially like the changelog entry with all its examples. Just needs a couple minor changes before approval.

aws/rust-runtime/aws-config/src/sts/assume_role.rs
ysaito1001: 👍  Addressed in b987538.

aws/rust-runtime/aws-http/src/auth.rs
ysaito1001: Thank you for pointing it out. That's certainly another place causing a breaking change. Provided the change log and an example in 9b4c55f.

jdisanti (approved):

PR #2129, opened 2022-12-22
Implement request ID access for SDK clients RFC

jdisanti:
## Motivation and Context
This PR implements the RFC written in #2083 to allow access of request IDs on outputs, and make access on errors consistent.

## Checklist

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExtendedRequestIdDecorator.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package software.amazon.smithy.rustsdk.customize.s3
+
+import software.amazon.smithy.rust.codegen.client.smithy.ClientCodegenContext
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+import software.amazon.smithy.rustsdk.BaseRequestIdDecorator
+import software.amazon.smithy.rustsdk.InlineAwsDependency
+
+class S3ExtendedRequestIdDecorator : BaseRequestIdDecorator() {
```
jdisanti, line 13: Thinking about this more, I wonder if we should just code generate the `RequestId` trait entirely so that for S3, it can have two methods on it instead of one. Then there wouldn't be a need for two separate traits in S3 to access the two request IDs, and then duplicated code could be eliminated.

jdisanti: This would benefit the S3 SDK, but would prevent accessing request IDs for types across SDK crates.

jdisanti:
Remaining work:
- [ ] Verify request IDs work for `Unhandled` service errors
- [ ] Add a `tracing::debug!` or `tracing::debug_span!` somewhere in the AWS middleware to attach the request ID to logs
- [ ] Bug (fix or file): If a service responds with invalid XML/JSON, then the request-id it responded with is lost since the unhandled parsing error is bubbled up without the raw response
- [ ] Update RFC

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExtendedRequestIdDecorator.kt
rcoh: can we have a root trait, then a second trait for S3 that gets re-exported?

aws/rust-runtime/aws-http/src/request_id.rs
```diff
@@ -0,0 +1,181 @@
...
+    Builder as GenericErrorBuilder, Error as GenericError, ErrorMetadata, Unhandled,
+};
+use http::{HeaderMap, HeaderValue};
+
+/// Constant for the [`aws_smithy_types::error::Error`] extra field that contains the request ID
+const AWS_REQUEST_ID: &str = "aws_request_id";
+
+/// Implementers add a function to return an AWS request ID
+pub trait RequestId {
+    /// Returns the request ID if it's available.
```
Velfi, line 19: In what cases is one not available?
```suggestion
    /// Returns the request ID. If the service could not be reached, returns `None`.
```

aws/rust-runtime/aws-http/src/request_id.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+impl<E, R> RequestId for SdkError<E, R>
+where
+    R: HttpHeaders,
+{
+    fn request_id(&self) -> Option<&str> {
+        match self {
+            Self::ResponseError(err) => extract_request_id(err.raw().http_headers()),
+            Self::ServiceError(err) => extract_request_id(err.raw().http_headers()),
```
Velfi, line 30: Is this valid?
```suggestion
            Self::ResponseError(err) | Self::ServiceError(err) => extract_request_id(err.raw().http_headers()),
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt
```diff
@@ -23,41 +26,49 @@ import software.amazon.smithy.rustsdk.endpoints.OperationInputTestDecorator
...
+    ApiGatewayDecorator().onlyApplyTo("com.amazonaws.apigateway#BackplaneControlService"),
+    Ec2Decorator().onlyApplyTo("com.amazonaws.ec2#AmazonEC2"),
+    GlacierDecorator().onlyApplyTo("com.amazonaws.glacier#Glacier"),
+    Route53Decorator().onlyApplyTo("com.amazonaws.route53#AWSDnsV20130401"),
+    "com.amazonaws.s3#AmazonS3".applyDecorators(
+        S3Decorator(),
+        S3ExtendedRequestIdDecorator(),
+    ),
+    S3ControlDecorator().onlyApplyTo("com.amazonaws.s3control#AWSS3ControlServiceV20180820"),
+    STSDecorator().onlyApplyTo("com.amazonaws.sts#AWSSecurityTokenServiceV20110615"),
```
Velfi, line 65: This is a cool idea

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/EndpointTraitBindingsTest.kt
```diff
@@ -67,7 +67,7 @@ internal class EndpointTraitBindingsTest {
                 }
                 """,
             )
-            implBlock(model.lookup("test#GetStatusInput"), sym) {
+            implBlock(sym.toSymbol(model.lookup("test#GetStatusInput"))) {
```
Velfi, line 70: It feels weird to call `toSymbol` on something already called `sym`. Is there a better name?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/CoreCodegenDecorator.kt
```diff
@@ -63,6 +66,31 @@ interface CoreCodegenDecorator<CodegenContext> {
         baseCustomizations: List<LibRsCustomization>,
     ): List<LibRsCustomization> = baseCustomizations
 
+    /**
+     * Hook to customize generated structures.
+     */
```
Velfi, line 71: Are these structures written to a specific module? What separates this from `extras`? It'd be nice to have a "map" of where all this stuff ends up.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt
```diff
@@ -25,43 +22,55 @@ import software.amazon.smithy.rust.codegen.core.rustlang.isDeref
...
-    protected val accessorMembers: List<MemberShape> = when (errorTrait) {
+    private val accessorMembers: List<MemberShape> = when (errorTrait) {
         null -> members
         // Let the ErrorGenerator render the error message accessor if this is an error struct
         else -> members.filter { "message" != symbolProvider.toMemberName(it) }
     }
-    protected val name = symbolProvider.toSymbol(shape).name
+    protected val name: String = symbolProvider.toSymbol(shape).name
 
-    fun render(forWhom: CodegenTarget = CodegenTarget.CLIENT) {
```
Velfi, line 60: ❤️

Velfi (commented):
I only got about halfway through but I'm loving what I see so far

aws/rust-runtime/aws-http/src/request_id.rs
jdisanti: This only works if `err` is the same type between the two, which it is not in this case.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/EndpointTraitBindingsTest.kt
jdisanti: Renamed `sym` to `symbolProvider`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/CoreCodegenDecorator.kt
jdisanti: These are customizations for the `StructureGenerator`. I'll clarify the comment.

rcoh:
general comments from reading the generated code diff:
- can we make this a deprecation instead of a huge break? would be nice for customers if it isn't a huge lift
- rather than having `_meta` I'd make it `meta` add docs and reserve `meta`
- surprised that `.meta()` returns a type called error

CHANGELOG.next.toml
```diff
@@ -132,3 +130,167 @@ Fix request canonicalization for HTTP requests with repeated headers (for exampl
...
+use aws_sdk_s3::types::RequestId;
+println!("Request ID: {:?}", error.request_id());
+```
+"""
+references = ["smithy-rs#76", "smithy-rs#2129"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
+author = "jdisanti"
+
+[[smithy-rs]]
+message = "Generic clients no longer expose a `request_id()` function on errors. To get request ID functionality, use the SDK code generator."
```
david-perez, line 159: This seems too harsh. Whitelabel clients should also be able to easily opt into retrieving request IDs without having to onboard onto the full SDK code generator --- perhaps the decorator could be enabled via a codegen config key?

CHANGELOG.next.toml
```diff
@@ -132,3 +130,167 @@ Fix request canonicalization for HTTP requests with repeated headers (for exampl
...
+references = ["smithy-rs#76", "smithy-rs#2129"]
+meta = { "breaking" = true, "tada" = false, "bug" = false, "target" = "client"}
+author = "jdisanti"
+
+[[aws-sdk-rust]]
```
david-perez, line 164: We should probably teach our changelog format a way so that an entry is meant to be duplicated for the AWS SDK.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/ServiceSpecificDecorator.kt
```diff
@@ -0,0 +1,133 @@
...
+    listOf(ServiceSpecificDecorator(ShapeId.from(serviceId), this))
+
+/** Apply the given decorators only to this service ID */
+fun String.applyDecorators(vararg decorators: ClientCodegenDecorator): List<ClientCodegenDecorator> =
+    decorators.map { it.onlyApplyTo(this) }.flatten()
+
+/**
+ * Delegating decorator that only applies to a configured service ID
+ */
+class ServiceSpecificDecorator(
```
david-perez, line 38: This is very useful! Can we port the generic bits to `codegen-core` so that the server project can use it too in the future?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/DisabledAuthDecorator.kt
```diff
@@ -46,3 +44,5 @@ class DisabledAuthDecorator : ClientCodegenDecorator {
         }
     }
 }
+
+private fun String.shapeId() = ShapeId.from(this)
```
david-perez, line 48: This seems like something that we want to reuse throughout the codebase. Can we put it in e.g. `Smithy.kt`?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
```diff
@@ -376,6 +376,13 @@ private fun Element.changeInto(tagName: String) {
     replaceWith(Element(tagName).also { elem -> elem.appendChildren(childNodesCopy()) })
 }
 
+/** Write an `impl` block for the given symbol */
+fun RustWriter.implBlock(symbol: Symbol, block: Writable) {
```
david-perez, line 380: Yeah there's tons of places in the server where we could use this syntax sugar, thanks for making use of it whenever possible.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/CoreCodegenDecorator.kt
```diff
@@ -62,6 +65,31 @@ interface CoreCodegenDecorator<CodegenContext> {
...
 
+    /**
+     * Hook to customize structures generated by `StructureGenerator`.
+     */
+    fun structureCustomizations(
+        codegenContext: CodegenContext,
+        baseCustomizations: List<StructureCustomization>,
+    ): List<StructureCustomization> = baseCustomizations
+
+    // TODO(https://github.com/awslabs/smithy-rs/issues/1401): Move these customizations into `ClientCodegenDecorator`
```
david-perez, line 76: cc @jjant for #2159.

david-perez: Also, only the first customization needs to be moved right? `errorImplCustomizations` is fed to `ErrorImplGenerator` which the server uses.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationErrorGenerator.kt
```diff
@@ -20,6 +22,21 @@ import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
...
+/**
+ * For a given Operation ([this]), return the symbol referring to the operation error. This can be used
+ * if you, e.g. want to return an operation error from a function:
+ *
+ * ```kotlin
+ * rustWriter.rustBlock("fn get_error() -> #T", operation.errorSymbol(symbolProvider)) {
+ *     write("todo!() // function body")
+ * }
+ * ```
+ */
```
david-perez, line 34: Nice; thanks.

david-perez (approved):
Server changes look good. My only general comment would be that we'll need this request ID extraction logic in generic clients too; we should have raised that when the RFC was proposed, apologies.

ysaito1001 (approved):
Could've broken into smaller PRs for the ease of review, but overall the changes look fantastic!

CHANGELOG.next.toml
jdisanti: If I understand correctly, a whitelabel client would already be using other parts of sdk-codegen for things like credentials, right? So those would be able to also use request ID after these changes. If there is a use case for pure/generic Smithy clients (generated by `codegen-client`) to have request ID, then it really should be in the Smithy spec.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/CoreCodegenDecorator.kt
jdisanti: Rephrased the TODO comment to limit it to just the builder customizations.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/ServiceSpecificDecorator.kt
jdisanti: I think I'm going to punt on this, but you're welcome to take parts of it.

CHANGELOG.next.toml
david-perez: I use the terms whitelabel and generic interchangeably, to refer to clients produced by `codegen-client`, without the SDK decorators in `sdk-codegen`.

There are use cases for generic clients to retrieve request IDs from responses, services return unmodeled request IDs frequently.

jdisanti: I think we should add these kinds of features to `sdk-codegen`, but make them consumable in isolation via the `smithy-build.json` configuration. Essentially we want varying degrees of the AWS SDK.

Issue #1966, opened 2022-11-09
`@length` on non-streaming blob shapes.


crisidev:
Implement the [`@length`](https://smithy.io/2.0/spec/constraint-traits.html#length-trait) trait on non-streaming `blob` shapes.

- `@length` on `@streaming` blob shapes is blocked by https://github.com/awslabs/smithy/issues/1389.

PR #2131, opened 2022-12-22
Implement `@length` on blobs 

82marbag:
## Motivation and Context
Implement `@length` on non streaming blob shapes: #1966

----


codegen-core/common-test-models/constraints.smithy
```diff
@@ -445,6 +445,11 @@ structure ConA {
     maxLengthString: MaxLengthString,
```
LukeMathWalker, line 445: Do we have a test for a blob shape in the closure of an event stream?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedBlobGenerator.kt
```diff
@@ -0,0 +1,190 @@
...
+                this
+            } else {
+                PubCrateConstraintViolationSymbolProvider(this)
+            }
+        }
+    private val constraintsInfo: List<TraitInfo> = listOf(LengthTrait::class.java)
+        .mapNotNull { shape.getTrait(it).orNull() }
+        .map { BlobLength(it).toTraitInfo() }
+
+    fun render() {
```
LukeMathWalker, line 52: Worth reusing the `TraitInfo` abstraction here - it'll make it easier to unify these constrained generators further down the line.

codegen-core/common-test-models/constraints.smithy
LukeMathWalker: Yes we do - https://github.com/awslabs/smithy-rs/pull/2131/files#diff-55153d4b5ba965cb5c4c263983d0b77c0a2ee71ac4a0775e02dc17389e1631faR125

LukeMathWalker (approved):

PR #2133, opened 2022-12-22
update: tune the S3 concurrency tests to finish in about 10s

Velfi:
This will also help users that receive a "Too many open files" error when running these tests.


jdisanti (approved):

aws/sdk/integration-tests/s3/tests/concurrency.rs
```diff
@@ -22,17 +22,21 @@ use tokio::sync::Semaphore;
 use tokio::time::{Duration, Instant};
 use tracing::debug;
 
-const TASK_COUNT: usize = 10_000;
+// WARNING:
+// When testing this on your own computer, be sure to run the tests in several different terminals.
+// Depending on the terminal used, you may run into errors related to "Too many open files".
+
+const TASK_COUNT: usize = 1_000;
```
LukeMathWalker, line 29: Worth suggesting to set `ulimit` to a higher threshold?

Velfi: Eh, I'd rather ensure that the defaults don't require that at all

LukeMathWalker: That's definitely the ideal scenario, but you never know what configuration people might have

PR #2138, opened 2022-12-23
Perform some optimisations on the CI Docker image

LukeMathWalker:
Optimisations:
- Break down the `install_rust` stage into smaller stages that can advance concurrently;
- Invoke `yum updateinfo` once, in bare_base_image;
- Use the unstable sparse registry feature to speed up `cargo`'s registry operations.

----

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -26,6 +26,8 @@ jobs:
         fetch-depth: 0
     - name: Acquire base image
       id: acquire
+      env:
+        DOCKER_BUILDKIT: 1
       run: ./smithy-rs/tools/ci-build/acquire-build-image
     - name: Upload base image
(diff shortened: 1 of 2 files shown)


LukeMathWalker:
This seems to shave a few minutes off the build - see [old](https://github.com/awslabs/smithy-rs/actions/runs/3762222299/jobs/6394720201) (~26 minutes) vs [this branch](https://github.com/awslabs/smithy-rs/actions/runs/3767199212/jobs/6404503368) (~23 minutes).

Due to the higher level of parallelism, the new setup would probably benefit significantly from a larger runner.
We could also introduce caching for the intermediate layers (e.g. why do we rebuild all our non-local CI binaries instead of pulling them down from a cache, assuming the version arg hasn't changed?).

Velfi (approved):

PR #2140, opened 2022-12-27
Extract SdkConfig parts into decorators

rcoh:
## Motivation and Context
The code currently has this awkward comment:
```
// !!NOTE!! As more items are added to aws_types::SdkConfig, use them here to configure the config builder
```
this gets more problematic as things can be conditionally included—we don't want to keep all that logic in SdkConfigDecorator.

## Description

Create `extraSections` as a customization option to allow open-ended customization sections to be defined.

This works in two parts:
1. A `DetachedSection` sets up the "type" of the customization. This is parameterized in terms of the interior section that will be used.
```kotlin
/**
 * Section enabling linkage between `SdkConfig` and <service>::Config
 */
object SdkConfigSection : DetachedSection<SdkConfigSection.CopySdkConfigToClientConfig>("SdkConfig") {
    /**
     * [sdkConfig]: A reference to the SDK config struct
     * [serviceConfigBuilder]: A reference (owned) to the `<service>::config::Builder` struct.
     *
     * Each invocation of this section MUST be a complete statement (ending with a semicolon), e.g:
     * ```
     * rust("${section.serviceConfigBuilder}.set_foo(${section.sdkConfig}.foo());")
     * ```
     */
    data class CopySdkConfigToClientConfig(val sdkConfig: String, val serviceConfigBuilder: String) :
        Section("CopyConfig")
}
```
2. Customizations define `DetachedSection<T> -> T -> Writable` — there is a helper method to set up the signature properly:

```kotlin
    override fun extraSections(codegenContext: ClientCodegenContext): List<Pair<DetachedSection<*>, (Section) -> Writable>> =
        listOf(
            SdkConfigSection.create { section ->
                writable {
                    rust("${section.serviceConfigBuilder}.set_credentials_provider(${section.sdkConfig}.credentials_provider().cloned());")
                }
            },
        )
}
```

## Checklist

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
@@ -31,7 +31,7 @@ import software.amazon.smithy.rust.codegen.core.rustlang.writable
 import software.amazon.smithy.rust.codegen.core.smithy.CodegenContext
 import software.amazon.smithy.rust.codegen.core.smithy.RuntimeConfig
 import software.amazon.smithy.rust.codegen.core.smithy.RustCrate
-import software.amazon.smithy.rust.codegen.core.smithy.customize.DetachedSection
+import software.amazon.smithy.rust.codegen.core.smithy.customize.AdHocSection
 import software.amazon.smithy.rust.codegen.core.smithy.customize.Section
 import software.amazon.smithy.rust.codegen.core.smithy.generators.LibRsCustomization
 import software.amazon.smithy.rust.codegen.core.smithy.generators.LibRsSection
@@ -81,7 +81,7 @@ class AwsEndpointDecorator : ClientCodegenDecorator {
         codegenContext: ClientCodegenContext,
         baseCustomizations: List<ConfigCustomization>,
     ): List<ConfigCustomization> {
-        return baseCustomizations.extendIf(regionalized(codegenContext)) {
+        return baseCustomizations.extendIf(codegenContext.isRegionalized()) {
             AwsEndpointShimCustomization(codegenContext)
         } + SdkEndpointCustomization(
             codegenContext,
@@ -104,7 +104,7 @@ class AwsEndpointDecorator : ClientCodegenDecorator {
             )
         }
         // generate a region converter if params has a region
-        if (!regionalized(codegenContext)) {
+        if (!codegenContext.isRegionalized()) {
             println("not generating a resolver for ${codegenContext.serviceShape}")
             return
         }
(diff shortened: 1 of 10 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
```diff
@@ -127,6 +130,25 @@ class AwsEndpointDecorator : ClientCodegenDecorator {
...
+                        ${section.serviceConfigBuilder}.set_endpoint_url(${section.sdkConfig}.endpoint_url().map(|url|url.to_string()));
+                    """,
+                    )
+
+                }
+            }
+        }
+    }
+
+    private fun regionalized(codegenContext: ClientCodegenContext) =
```
Velfi, line 149:
```suggestion
    private fun isRegionalized(codegenContext: ClientCodegenContext) =
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
```diff
@@ -127,6 +130,25 @@ class AwsEndpointDecorator : ClientCodegenDecorator {
         }
     }
 
+    override fun extraSections(codegenContext: ClientCodegenContext): List<Pair<DetachedSection<*>, (Section) -> Writable>> {
+        return regionalized(codegenContext).thenSingletonListOf {
+            SdkConfigSection.create { section ->
+                {
+                    rust(
+                        """
+                        ${section.serviceConfigBuilder}.set_aws_endpoint_resolver(${section.sdkConfig}.endpoint_resolver().clone());
```
Velfi, line 139: Is setting the `endpoint_resolver` here going to trigger a deprecation warning?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkConfigDecorator.kt
```diff
@@ -12,11 +12,56 @@ import software.amazon.smithy.rust.codegen.client.smithy.generators.config.Confi
...
+
+    override fun extraSections(codegenContext: ClientCodegenContext): List<Pair<DetachedSection<*>, (Section) -> Writable>> =
+        listOf(
+            SdkConfigSection.create { section ->
+                writable {
+                    rust(
+                        """
+                        ${section.serviceConfigBuilder}.set_retry_config(${section.sdkConfig}.retry_config().cloned());
+                        ${section.serviceConfigBuilder}.set_timeout_config(${section.sdkConfig}.timeout_config().cloned());
+                        ${section.serviceConfigBuilder}.set_sleep_impl(${section.sdkConfig}.sleep_impl());
```
Velfi, line 58: @jdisanti refers to this stuff as "Resiliency"
https://github.com/awslabs/smithy-rs/blob/main/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/ResiliencyConfigCustomization.kt

It's worth considering if we should move this code into a similarly-named package in AWS codegen.

Velfi (approved):
This is a cool approach. I'm wondering if it's worth adding some breadcrumbs/doc links that will help future contributors realize that this tool is available. We should at least announce it to the server folks, if only to get them to wonder if they might want to start using a method like this.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsEndpointDecorator.kt
rcoh: this doesn't change the generated code—just changes where it's being generated from

Velfi (approved):

PR #2145, opened 2022-12-29
Add a test helper for creating a dummy Credentials

ysaito1001:
## Motivation and Context
This is a refactoring PR affecting only tests, in response to https://github.com/awslabs/smithy-rs/pull/2122#discussion_r1056499862.

## Description
We have various places in testing that create dummy `Credentials` such as
- `Credentials::new("ANOTREAL", "notrealrnrELgWzOk3IfjzDKtFBhDby", Some("notarealsessiontoken".to_string()), None, "test")`
- `Credentials::new("akid", "secret", None, None, "test")`
- `Credentials::new("example", "example", None, None, "example")`

These essentially mean "it does not matter". To consolidate all of these instances, a test helper `aws_credential_types::Credentials::for_tests` has been introduced whose implementation is taken from the first bullet point above.

We intentionally did not replace certain occurrences of dummy `Credentials::new` with `Credentials::for_tests`:
- Those that appear in code examples in rustdoc (it makes more sense for the customers to see `new` rather than `for_tests`)
- Those that need to configure the expiration time

## Testing
- [x] Passed all tests in CI

----

### DIFF
--- a/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
+++ b/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
@@ -140,6 +140,21 @@ impl Credentials {
         )
     }
 
+    /// Creates a test `Credentials`.
+    ///
+    /// Marked as `#[doc(hidden)]` due to the usage limited to testing.
+    /// TODO(https://github.com/rust-lang/cargo/issues/2911): Consider using an auto-activated feature for testing
+    #[doc(hidden)]
+    pub fn for_tests() -> Self {
+        Self::new(
+            "ANOTREAL",
+            "notrealrnrELgWzOk3IfjzDKtFBhDby",
+            Some("notarealsessiontoken".to_string()),
+            None,
+            "test",
+        )
+    }
+
     /// Returns the access key ID.
     pub fn access_key_id(&self) -> &str {
         &self.0.access_key_id
--- a/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
+++ b/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
@@ -141,10 +141,7 @@ impl Credentials {
     }
 
     /// Creates a test `Credentials`.
-    ///
-    /// Marked as `#[doc(hidden)]` due to the usage limited to testing.
-    /// TODO(https://github.com/rust-lang/cargo/issues/2911): Consider using an auto-activated feature for testing
-    #[doc(hidden)]
+    #[cfg(feature = "test-util")]
     pub fn for_tests() -> Self {
         Self::new(
             "ANOTREAL",
--- a/aws/rust-runtime/aws-sig-auth/src/middleware.rs
+++ b/aws/rust-runtime/aws-sig-auth/src/middleware.rs
@@ -220,7 +220,7 @@ mod test {
                 properties.insert(UNIX_EPOCH + Duration::new(1611160427, 0));
                 properties.insert(SigningService::from_static("kinesis"));
                 properties.insert(OperationSigningConfig::default_config());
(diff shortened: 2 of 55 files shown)


aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
```diff
@@ -140,6 +140,21 @@ impl Credentials {
...
+    #[doc(hidden)]
+    pub fn for_tests() -> Self {
+        Self::new(
+            "ANOTREAL",
+            "notrealrnrELgWzOk3IfjzDKtFBhDby",
+            Some("notarealsessiontoken".to_string()),
+            None,
+            "test",
+        )
+    }
```
rcoh, line 156: I don't think this should be doc-hidden but we should feature gate it behind `test-util`. You can see how we handle this for `aws-smithy-client` here: https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/s3/Cargo.toml#L100-L103

aws/rust-runtime/aws-sig-auth/src/middleware.rs
```diff
@@ -270,6 +270,7 @@ mod test {
                 .apply(req.try_clone().expect("can clone"))
                 .expect_err("no cred provider"),
         );
+        // use `new` instead of `for_tests` because the presence of a session_token will render `auth_header` below `Sensitive`.
```
rcoh, line 273: you can use `unwrap().to_string()` to debug IIRC

rcoh (commented):
Thank you so much for doing this!!

aws/sdk/integration-tests/s3/tests/query-strings-are-correctly-encoded.rs
```diff
@@ -15,13 +15,7 @@ use std::time::{Duration, UNIX_EPOCH};
...
     let (conn, rcvr) = capture_request(None);
     let sdk_config = SdkConfig::builder()
-        .credentials_provider(SharedCredentialsProvider::new(Credentials::new(
-            "ANOTREAL",
-            "notrealrnrELgWzOk3IfjzDKtFBhDby",
-            Some("notarealsessiontoken".to_string()),
-            None,
-            "test",
-        )))
+        .credentials_provider(SharedCredentialsProvider::new(Credentials::for_tests()))
```
Velfi, line 18: What do you think of adding a method to `SharedCredentialsProvider` like so:
```suggestion
        .credentials_provider(SharedCredentialsProvider::for_tests())
```

It's just about as clear, with a little less code.

Velfi (approved):

aws/sdk/integration-tests/s3/tests/query-strings-are-correctly-encoded.rs
ysaito1001: I would normally incorporate the suggestion but knowing `SharedCredentialsProvider` will cease to exist in #2122, we should probably leave it as is.

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
ysaito1001: Considered once the introduction of a `test` feature but did not go for it. Happy to make the change.

ysaito1001: Addressed in dfecd0d.

aws/rust-runtime/aws-sig-auth/src/middleware.rs
ysaito1001: Great, that worked. Thanks for the suggestion. Addressed in 851abbb.

PR #2149, opened 2022-12-30
Python: Type-stub generation for SSDKs

unexge:
### What is a Python type-stub?

Type-stubs allows us to define type information for Python modules. They usually used to define type information for untyped modules, and it is the only way to define type information for the native extensions which is the case for Smithy Python Server SDKs. That’s why we need to provide type-stubs alongside our SSDKs to allow our users to get a good IDE integration, static type checking and better documentation.

Type-stubs uses same syntax with regular Python code except they just use `...` for function or class bodies, and they’ve `.pyi` extensions. An example type-stub:

```python
# foo.pyi

class Foo:
    ...

class Bar:
   baz: int

def qux(foobar: str) -> Bar:
    ...
```

For more details about Python type-stubs: https://typing.readthedocs.io/en/latest/source/stubs.html

### Implementation

There is some ongoing work on PyO3 side to handle this, but it is still in early stages. Since we have most of the information needed in our code generator we thought we can roll our custom solution for that until we have a solution on PyO3 side.

Implementation is mostly based on the idea shared by Tpt: https://github.com/PyO3/pyo3/issues/2454#issuecomment-1273300176, that is:

1. Annotate Rust types with Python type information using [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format) (We don't accurately follow the spec though):
    For code generated types: this PR introduces [PythonType](https://github.com/awslabs/smithy-rs/blob/55a10ec7c83e5dfe9288ff04ba5383a638533c6d/codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonType.kt) to represent and render Python types.
    For runtime types: we manually add type information to the comments of Rust types (as you can see in PR diff for most of the Rust files)
2. Build Python module from Rust crate
3. Inspect Python module, parse doc strings of elements and generate type-stubs based on them. This is done in [stubgen.py](https://github.com/awslabs/smithy-rs/blob/55a10ec7c83e5dfe9288ff04ba5383a638533c6d/rust-runtime/aws-smithy-http-server-python/examples/stubgen.py)
4. Build Python module from Rust crate again but this time Rust crate also includes Python type-stubs

## Testing
You can checkout this branch and run `make build` in `rust-runtime/aws-smithy-http-server-python/examples`, that command will generate and install a Python wheel for Pokémon Service containing Python type-stubs. You should then open `rust-runtime/aws-smithy-http-server-python/examples/pokemon_service.py` in your text editor, and you should be able to use `go to definition`, `autocomplation` etc. in your editor. Alternatively you can also see the generated type-stubs in `rust-runtime/aws-smithy-http-server-python/examples/pokemon-service-server-sdk/python/pokemon_service_server_sdk`.

## TODOs - Questions

- [ ] As you can see in the steps we need to do double build. Is there a way to prevent that? Is it okay to build it twice?
- [x] Use generics for `Context` class, currently we're using `Any` for `Context` (Done in https://github.com/awslabs/smithy-rs/pull/2149/commits/541df2da36ebb63ff4c2363d01f018b09470d938)
- [x] Handle default values, currently we are using `...` in place of default values, but we shouldn't emit `...` if there is no default value for that type (Done in https://github.com/awslabs/smithy-rs/pull/2149/commits/5e56f45879c1e1954aa9786bb8e0946cd522a755)
- [x] Omit `__init__` for classes that doesn't have constructors (Done in https://github.com/awslabs/smithy-rs/pull/2149/commits/fb5bcd79ca9db52cd5d8938fe1b762ef45ba6544)
- [ ] How to handle `getters` and `setters`? Currently, we only annotate `getters` and in type-stubs we generate them as regular fields. Should we generate two methods with `@property` and `@T.setter` with correct type information?
- [x] How to integrate it with CI? Should we generate diffs for generated stubs similar to generated SDKs? Or should we just check out generated type-stubs to Git? Should we run `mypy` to verify we generate correct type-stubs? (Done in https://github.com/awslabs/smithy-rs/pull/2149/commits/cde02764ebc86750e1895de0d76b83dd26e36fef and https://github.com/awslabs/smithy-rs/pull/2149/commits/df4a07920c85026e9f7b6526618a97fa568dab1e)
- [x] Should we preserve comments for types? Is this going to help documentation tools and IDEs? (Done in https://github.com/awslabs/smithy-rs/pull/2149/commits/d05d3a036855d0124ea0abebce41ff7f2d66c537)

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
```diff
@@ -387,6 +387,7 @@ class RustWriter private constructor(
         fun factory(debugMode: Boolean): Factory<RustWriter> = Factory { fileName: String, namespace: String ->
             when {
                 fileName.endsWith(".toml") -> RustWriter(fileName, namespace, "#", debugMode = debugMode)
+                fileName.endsWith(".py") -> RustWriter(fileName, namespace, "#", debugMode = debugMode)
```
unexge, line 390: Can we do this customization in Python namespace?

rust-runtime/aws-smithy-http-server-python/examples/stubgen.py
```diff
@@ -0,0 +1,391 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
crisidev, line 1: Is the `example` folder the right place for this file? We need to be able to vend it to customers somehow, right?

rust-runtime/aws-smithy-http-server-python/examples/stubgen.py
```diff
@@ -0,0 +1,391 @@
...
+    return indent('"""\n' + doc + '\n"""\n', indent_level)
+
+
+def is_fn_like(obj: Any) -> bool:
+    return (
+        inspect.isbuiltin(obj)
+        or inspect.ismethod(obj)
+        or inspect.isfunction(obj)
+        or inspect.ismethoddescriptor(obj)
+    )
```
crisidev, line 224: Don't you need also `inspect.iscoroutinefunction` ?

rust-runtime/aws-smithy-http-server-python/examples/stubgen.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+if __name__ == "__main__":
+    import argparse
+    import importlib
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("module")
+    parser.add_argument("outdir")
+    args = parser.parse_args()
```
crisidev, line 380: This code is quite complex and hard to review (it's a hard problem) :D I would like to see some tests in this file to ensure we are doing the right thing and we do not regress in the future.

I think you have several options:
* Make a test suite using pytest with a standard python module
* Add an option to the main, something like `--run-tests`, which will lazily import `pytest` and run something like `pytest.run()`, which will collect the test cases and run them
* Delegate the testing to the github CI or to something written in kotlin in the code generator.

What do you think?

crisidev: I don't think tests are **a requirement** for this PR, but we should work on them before releasing this feature as stable.

rust-runtime/aws-smithy-http-server-python/examples/stubgen.py
unexge: Yes we need to vend it to customers somehow, but I'm not sure how yet. I think we'll need a build system eventually but currently the closest thing we have is the `Makefile` in the `examples` folder, so that's why I've put this file here.

crisidev:
Looking at the diff, I noticed something weird. I think there is an off-by-one error somewhere since docstrings that should be **above** the attribute are actually **below**. Example:

This attribute taken one of the pyi

```python
body: typing.Awaitable[bytes]
"""
Return the HTTP body of this request.
Note that this is a costly operation because the whole request body is cloned.
"""
```

should be

```python
"""
Return the HTTP body of this request.
Note that this is a costly operation because the whole request body is cloned.
"""
body: typing.Awaitable[bytes]
```

Am I wrong?

The same applies to everything in the pyi files.

crisidev:
> Yes it indeed seems a bit weird but as defined in [PEP-257](http://www.python.org/dev/peps/pep-0257/#what-is-a-docstring)
> 
> > String literals occurring immediately after a simple assignment at the top level of a module, class, or **init** method are called “attribute docstrings”.
> 
> That's the way to define docstring for items. See https://stackoverflow.com/a/6061254 for more detail

Oh wow. TIL..

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
crisidev: I don't think so.. To me it is reasonable to have it here, same goes for `toml` for example, but I would like to hear the SDK team thoughts..

rust-runtime/aws-smithy-http-server-python/examples/stubgen.py
crisidev: I think it is fine for this PR as we are just adding the functionality. We need to start thinking about how to vend this file. We should probably cut an issue about it.

rust-runtime/aws-smithy-http-server-python/examples/stubgen.py
crisidev: This is what inspect gives you:

```
 'isabstract',
 'isasyncgen',
 'isasyncgenfunction',
 'isawaitable',
 'isbuiltin',
 'isclass',
 'iscode',
 'iscoroutine',
 'iscoroutinefunction',
 'isdatadescriptor',
 'isframe',
 'isfunction',
 'isgenerator',
 'isgeneratorfunction',
 'isgetsetdescriptor',
 'ismemberdescriptor',
 'ismethod',
 'ismethoddescriptor',
 'ismodule',
 'isroutine',
 'istraceback',
```

I am pretty sure you'll have to add `iscoroutine` and `iscoroutinefunction`. I am not sure about `isawaitable`, `isgenerator`, `isgeneratorfunction`, `isroutine`, `isasyncgen` and `isasyncgenfunction`.

This can help you understand what is what:

```python
for x in dir(inspect):
    if x.startswith("is"):
        obj = getattr(inspect, x)
        print(f"============= {x} ============")
        print(f"{obj.__doc__}")
        print("==============================================\n")
```

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonType.kt
```diff
@@ -0,0 +1,174 @@
...
+        is RustType.Slice -> PythonType.List(this.member.pythonType())
+        is RustType.HashMap -> PythonType.Dict(this.key.pythonType(), this.member.pythonType())
+        is RustType.HashSet -> PythonType.Set(this.member.pythonType())
+        is RustType.Reference -> this.member.pythonType()
+        is RustType.Option -> PythonType.Optional(this.member.pythonType())
+        is RustType.Box -> this.member.pythonType()
+        is RustType.Dyn -> this.member.pythonType()
+        is RustType.Opaque -> PythonType.Opaque(this.name, this.namespace)
+        // TODO(Constraints): How to handle this?
+        // Revisit as part of https://github.com/awslabs/smithy-rs/issues/2114
```
crisidev, line 130: I don't think you have to.. I was talking with @david-perez and right now Python only handles constraints during deserialization and hands you off the original structure, without newtypes if the validation succeeded or an error otherwise.

rust-runtime/aws-smithy-http-server-python/examples/stubgen_test.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+def create_module(name: str, code: str) -> ModuleType:
+    mod = ModuleType(name)
+    exec(dedent(code), mod.__dict__)
+    if not hasattr(mod, "__all__"):
+        # Manually populate `__all__` with all the members that doesn't start with `__`
+        mod.__all__ = [k for k in mod.__dict__.keys() if not k.startswith("__")]  # type: ignore
+    sys.modules[name] = mod
+    return mod
```
crisidev, line 21: 🤠

crisidev (commented):
This looks good to me. I am waiting to approve it because there are still open question about what to get from `inspect` to understand if an object is function like.

rust-runtime/aws-smithy-http-server-python/examples/stubgen.py
unexge: I've added `iscoroutine` and `iscoroutinefunction`. Since `async fn`s are not permitted by PyO3 and coroutines/generators are currently [experimental](https://doc.rust-lang.org/std/ops/trait.Generator.html) we currently only expose regular functions and so the current check should be sufficient.

crisidev (approved):

PC-LoadLetter (approved):
Thanks for letting me review your PR. I learned something. I made some minor comments but over all it looks good to me.

rust-runtime/aws-smithy-http-server-python/src/lambda.rs
```diff
@@ -86,29 +127,49 @@ struct PyConfig {
...
+
     /// The ARN of the Lambda function being invoked.
+    ///
+    /// :type str:
     #[pyo3(get)]
     invoked_function_arn: String,
+
     /// The X-Ray trace ID for the current invocation.
+    ///
+    /// :type typing.Optional\[str\]:
```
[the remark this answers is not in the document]
unexge: Escape chars are required for Rustdoc because the syntax `[foo]` is for [intra-doc linking](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html) and Rustdoc tries to link it to element `foo` and fails if it can't. We are [unescaping](https://github.com/awslabs/smithy-rs/blob/5f355d1839dc25d8161a3f1a3689cb46f38d027f/rust-runtime/aws-smithy-http-server-python/examples/stubgen.py#L34-L43) those chars in our stubgen script, but I don't know if there is a way to unescape them from `help()` function too.

crisidev:
\0/ Nice!

PR #2152, opened 2022-12-30
Refactor configuration loading to parse profile files only once

### LINKED ISSUE #2142: Profile File Refactoring
- Parse profile file only once instead of everytime info is read from it

rcoh:
## Motivation and Context
- https://github.com/awslabs/aws-sdk-rust/issues/237
- parsing profile files each time produces noisy logs
- parsing profile files multiple times is inefficient
- parsing profile files multiple times makes it difficult to unify profile parsing settings
- fixes #2142 
## Description
1. Refactor to only parse profile files once in a lazy async cell
2. Support setting the profile name and profile file at a SdkConfig level instead of for each provider.

## Testing
- [x] UT

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/lib.rs
+++ b/aws/rust-runtime/aws-config/src/lib.rs
@@ -353,9 +353,8 @@ mod loader {
         /// `~/.aws/config` and `~/.aws/credentials` respectively.
         ///
         /// Any number of config and credential files may be added to the `ProfileFiles` file set, with the
-        /// only requirement being that there is at least one of them. Custom file locations that are added
-        /// will produce errors if they don't exist, while the default config/credentials files paths are
-        /// allowed to not exist even if they're included.
+        /// only requirement being that there is at least one of each. Profile file locations will produce an
+        /// error if they don't exist, but the default config/credentials files paths are exempt from this validation.
         ///
         /// # Example: Using a custom profile file path
         ///
--- a/aws/rust-runtime/aws-config/src/lib.rs
+++ b/aws/rust-runtime/aws-config/src/lib.rs
@@ -161,6 +161,7 @@ mod loader {
     use crate::connector::default_connector;
     use crate::default_provider::{app_name, credentials, region, retry_config, timeout_config};
     use crate::meta::region::ProvideRegion;
+    use crate::profile::profile_file::ProfileFiles;
     use crate::provider_config::ProviderConfig;
 
     /// Load a cross-service [`SdkConfig`](aws_types::SdkConfig) from the environment
@@ -181,6 +182,8 @@ mod loader {
         timeout_config: Option<TimeoutConfig>,
         provider_config: Option<ProviderConfig>,
         http_connector: Option<HttpConnector>,
+        profile_name_override: Option<String>,
+        profile_files_override: Option<ProfileFiles>,
     }
 
     impl ConfigLoader {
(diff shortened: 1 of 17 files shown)


ysaito1001 (approved):
LGTM! I like how parsing profile files is now single-sourced in `ProviderConfig`.

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -344,6 +347,73 @@ mod loader {
...
 
+        /// Provides the ability to programmatically override the profile files that get loaded by the SDK.
+        ///
+        /// The [`Default`] for `ProfileFiles` includes the default SDK config and credential files located in
+        /// `~/.aws/config` and `~/.aws/credentials` respectively.
+        ///
+        /// Any number of config and credential files may be added to the `ProfileFiles` file set, with the
+        /// only requirement being that there is at least one of them. Custom file locations that are added
+        /// will produce errors if they don't exist, while the default config/credentials files paths are
+        /// allowed to not exist even if they're included.
```
Velfi, line 358: I feel like this could be reworded a tiny bit. I tried my hand at it, but feel free to ignore this if you like the current wording.

```suggestion
        /// Any number of config and credential files may be added to the `ProfileFiles` file set, with the
        /// only requirement being that there is at least one of each. Profile file locations will produce an
        /// error if they don't exist, but the default config/credentials files paths are exempt from this validation.
```

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -344,6 +347,73 @@ mod loader {
...
+        /// use aws_config::profile::{ProfileFileCredentialsProvider, ProfileFileRegionProvider};
+        /// use aws_config::profile::profile_file::{ProfileFiles, ProfileFileKind};
+        ///
+        /// # async fn example() {
+        /// let sdk_config = aws_config::from_env()
+        ///     .profile_name("prod")
+        ///     .load()
+        ///     .await;
+        /// # }
+        pub fn profile_name(mut self, profile_name: impl Into<String>) -> Self {
```
Velfi, line 412: It may be worth linking to https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html or similar

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -372,8 +442,8 @@ mod loader {
         ///
         /// Update the `ProviderConfig` used for all nested loaders. This can be used to override
         /// the HTTPs connector used by providers or to stub in an in memory `Env` or `Fs` for testing.
-        /// This **does not set** the HTTP connector used when sending operations. To change that
-        /// connector, use [ConfigLoader::http_connector].
+        ///
+        /// NOTE: Update the provider config with care because it will override any other settings.
```
Velfi, line 446: Can we be more specific about what this will override? I'm wondering if we should rename `ProviderConfig` to better highlight the stuff it provides. I'm thinking of something like this:

- `OsProvider`
- `PlatformProvider`
- `SystemProvider`

aws/rust-runtime/aws-config/src/profile/app_name.rs
```diff
@@ -62,26 +54,14 @@ impl ProfileFileAppNameProvider {
...
-                Err(err) => {
-                    tracing::warn!(err = %err, "`sdk-ua-app-id` property in profile `{}` was invalid", selected_profile_name);
-                    None
-                }
-            })
+        let app_id = self.provider_config.profile().await?.get("sdk-ua-app-id")?;
+        match AppName::new(app_id.to_owned()) {
+            Ok(app_name) => Some(app_name),
+            Err(err) => {
+                tracing::warn!(err = %err, "`sdk-ua-app-id` property `{}` was invalid", app_id);
```
Velfi, line 61: If possible, we should consider the following:

- add a call to action here that tells people how to fix this or stop the warning
- explain to users what value will be used instead
- explain to users how this may affect logging of requests
- explain why you would want to set this
- explain what makes an app ID invalid

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
```diff
@@ -123,15 +122,11 @@ pub(super) fn resolve_chain<'a>(
     // - There is not default profile
```
Velfi, line 122: I think this is a typo. It's not something you touched, so feel free to ignore this
```suggestion
    // - There is no default profile
```

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
```diff
@@ -123,15 +122,11 @@ pub(super) fn resolve_chain<'a>(
...
     // Then:
     // - Treat this situation as if no profiles were defined
-    if profile_override.is_none()
-        && profile_set.selected_profile() == "default"
-        && profile_set.get_profile("default").is_none()
-    {
+    if profile_set.selected_profile() == "default" && profile_set.get_profile("default").is_none() {
         tracing::debug!("No default profile defined");
         return Err(ProfileFileError::NoProfilesDefined);
```
Velfi, line 127: Is it possible for a user to have other profiles but they forgot to set the profile name? If so, then I think this error could be improved. It's not so much that no profiles were defined but that the selected profile (`default`) couldn't be found.

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -195,6 +196,53 @@ impl Property {
...
 }
 
+/// Failed to read or parse the profile file(s)
+#[derive(Debug, Clone)]
+pub enum ProfileFileLoadError {
+    /// The profile could not be parsed
+    #[non_exhaustive]
+    ParseError(ProfileParseError),
+
+    /// The profile file (e.g. `~/.aws/config`) encountered an error reading from the filesystem
```
Velfi, line 206:
```suggestion
    /// Attempt to read the AWS config file (`~/.aws/config` by default) failed with a filesystem error.
```

aws/rust-runtime/aws-config/src/profile/app_name.rs
rcoh: %err will include details so this should be OK

aws/rust-runtime/aws-config/src/provider_config.rs
```diff
@@ -35,6 +42,9 @@ pub struct ProviderConfig {
     connector: HttpConnector,
     sleep: Option<Arc<dyn AsyncSleep>>,
     region: Option<Region>,
+    parsed_profile: Arc<OnceCell<Result<ProfileSet, ProfileFileLoadError>>>,
+    profile_files: ProfileFiles,
+    profile_name_override: Option<Cow<'static, str>>,
```
Velfi, line 47: It'd be nice to have docs for these somewhere. These are private fields so they don't need to be doc comments but something like this would be nice:
```suggestion
    // An AWS profile created from `ProfileFiles` and a `profile_name`
    profile: Arc<OnceCell<Result<ProfileSet, ProfileFileLoadError>>>,
    // A list of [std::path::Path]s to profile files
    profile_files: ProfileFiles,
    // An override to use when constructing a `ProfileSet`
    profile_name_override: Option<Cow<'static, str>>,
```

Otherwise I'm going to get confused about what each part is for.

aws/rust-runtime/aws-config/src/provider_config.rs
```diff
@@ -180,12 +201,57 @@ impl ProviderConfig {
...
+        self.try_profile().await.ok()
+    }
+
     /// Override the region for the configuration
     pub fn with_region(mut self, region: Option<Region>) -> Self {
         self.region = region;
         self
     }
 
+    /// Override profile-file (`~/.aws/config` et al) configuration
```
Velfi, line 234: I think we should avoid conflating the path, the files, and the config info extracted from one or more files. 

```suggestion
    /// Override the profile file paths (`~/.aws/config` by default) and name (`default` by default)
```

PR #2153, opened 2023-01-01
feat(smithy-http-auth): add api key auth types

eduardomourar:
## Motivation and Context
In preparation for supporting the [httpApiKeyAuth trait](https://smithy.io/1.0/spec/core/auth-traits.html#httpapikeyauth-trait) within the rust codegen, the `aws-smithy-http-auth` crate will be created containing the proper types that will be used later. Although out of the scope for this PR, a similar process can be done for [httpBasicAuth](https://smithy.io/1.0/spec/core/auth-traits.html#httpbasicauth-trait), [httpDigestAuth](https://smithy.io/1.0/spec/core/auth-traits.html#httpdigestauth-trait) or [httpBearerAuth](https://smithy.io/1.0/spec/core/auth-traits.html#httpbearerauth-trait) traits.

## Description
There are two major parts for this PR:
1. Create the `AuthApiKey` struct that will be used to hold the API key information (for now only a static string value). Consequently, it needs to be included into the default SDK config builder.
2. Create the `HttpAuthDefinition` struct that will allow the codegen to pass the values from the Smithy model to Rust and also be used while mutating the HTTP request so that we don't apply it a second time.

## Testing
Some basic unit tests

## Checklist
- [] I have updated `CHANGELOG.next.toml` if I made changes to the smithy-rs codegen or runtime crates

----


aws/rust-runtime/aws-types/src/sdk_config.rs
```diff
@@ -153,6 +156,23 @@ impl Builder {
         self
     }
 
+    /// Set the api key to use when making requests.
+    /// # Examples
+    /// ```
+    /// use aws_types::SdkConfig;
+    /// let config = SdkConfig::builder().api_key("some-api-key").build();
+    /// ```
+    pub fn api_key(mut self, api_key: impl Into<AuthApiKey>) -> Self {
```
jdisanti, line 165: We should talk more about your use-case, but I think this is probably the wrong place to add this. Adding it here will impact all AWS SDK users as this configuration will become available to them when its not technically a valid configuration for the AWS SDKs.

Essentially, right now, there are two types of smithy-rs client generation:
1. Generic client generation
2. AWS SDK client generation, which is a bundle of codegen customizations applied to (1) to add AWS SDK specific features. Basically anything in the `aws/` directory is part of this client codegen type.

HTTP auth makes sense to add directly to (1) since it's just handling the built-in http auth traits in Smithy.

rust-runtime/aws-smithy-types/src/auth.rs
```diff
@@ -0,0 +1,270 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+//! Smithy HTTP Auth Types
+
+use std::cmp::PartialEq;
+use std::fmt::Debug;
+
+#[derive(Debug, PartialEq)]
+enum AuthErrorKind {
```
jdisanti, line 11: These types need to go into their own (new) `aws-smithy-http-auth` crate rather than into `aws-smithy-types`. I went ahead and squatted it so we can use the name: https://crates.io/crates/aws-smithy-http-auth

This new crate can be created in `rust-runtime/` side by side with `aws-smithy-types`.

rust-runtime/aws-smithy-http/src/query_writer.rs
```diff
@@ -3,19 +3,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-use crate::http_request::url_escape::percent_encode_query;
+use crate::query::fmt_string as percent_encode_query;
 use http::Uri;
 
 /// Utility for updating the query string in a [`Uri`].
-pub(super) struct QueryWriter {
+pub struct QueryWriter {
```
jdisanti, line 10: Since this is becoming public, I think we should make it `#[doc(hidden)]` to indicate it is not intended to be used directly (it is an internal utility for codegen only, and the flexibility to change it at any time is definitely desired).

rust-runtime/aws-smithy-types/src/auth.rs
```diff
@@ -0,0 +1,270 @@
...
+    InvalidLocation,
+    SchemeNotAllowed,
+}
+
+/// Error for Smithy authentication
+#[derive(Debug)]
+pub struct AuthError {
```
jdisanti, line 18: Good job on the error type!

rust-runtime/aws-smithy-types/src/auth.rs
```diff
@@ -0,0 +1,270 @@
...
+impl From<AuthErrorKind> for AuthError {
+    fn from(kind: AuthErrorKind) -> Self {
+        Self { kind }
+    }
+}
+
+/// Authentication configuration to connect to a Smithy Service
+#[derive(Clone, Debug, PartialEq)]
+pub struct AuthApiKey {
+    api_key: String,
```
jdisanti, line 44: We should use `Zeroizing` on this so that it is wiped from memory when no longer in use (see `Credentials` for example: https://github.com/awslabs/smithy-rs/blob/76ee00eb1737964ce240ae817f573980920f0607/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs#L25).

We also need to implement `Debug` manually so that the API key value can be redacted. Example: https://github.com/awslabs/smithy-rs/blob/76ee00eb1737964ce240ae817f573980920f0607/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs#L47

rust-runtime/aws-smithy-types/src/auth.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// An HTTP-specific authentication scheme that sends an arbitrary
+/// auth value in a header or query string parameter.
+// As described in the Smithy documentation:
+// https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
+#[derive(Clone, Debug, PartialEq)]
+pub struct HttpAuthDefinition {
+    /// Defines the location of where the Auth is serialized. This value
+    /// can be set to `"header"` or `"query"`.
+    location: String,
```
jdisanti, line 89: Seems like this should be an enum with two variants.

rust-runtime/aws-smithy-types/src/auth.rs
```diff
@@ -0,0 +1,270 @@
...
+    name: String,
+
+    /// Defines the security scheme to use on the `Authorization` header value.
+    /// This can only be set if the "in" property is set to `"header"`.
+    scheme: Option<String>,
+}
+
+impl HttpAuthDefinition {
+    /// Constructs a new HTTP auth definition.
+    fn new<S>(location: String, name: String, scheme: S) -> Result<Self, AuthError>
```
jdisanti, line 102: We should use the builder pattern so that more fields can be added in the future without a breaking change.

rust-runtime/aws-smithy-types/src/auth.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+    }
+
+    /// Returns the underlying api key.
+    pub fn api_key(&self) -> &str {
+        &self.api_key
+    }
+
+    /// Sets the value for the api key
+    pub fn set_api_key(mut self, api_key: impl Into<String>) -> Self {
```
jdisanti, line 61: Doesn't seem like this function needs to exist since there's only one field.

jdisanti (commented):
Hey, sorry for getting back to you so late on this! Thank you for making this contribution.

In the future, it will be helpful to file an issue as a starting point to discuss how a feature like this should be implemented. A lot of my comments could have been pointed out in advance and saved you time.

With that, let's get into it and work towards getting this merged! 😄

eduardomourar:
> Hey, sorry for getting back to you so late on this! Thank you for making this contribution.
> 
> 
> 
> In the future, it will be helpful to file an issue as a starting point to discuss how a feature like this should be implemented. A lot of my comments could have been pointed out in advance and saved you time.
> 
> 
> 
> With that, let's get into it and work towards getting this merged! 😄 

I will next time. In this case, we are already using this as is internally (including the other PR with the actual injection through the codegen) so it was simpler to just create a PR.

rust-runtime/Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
     "aws-smithy-checksums",
     "aws-smithy-eventstream",
     "aws-smithy-http",
+    "aws-smithy-http-auth",
```
jdisanti, line 10: Needs to be added to this list also: https://github.com/awslabs/smithy-rs/blob/bb5bbeb5cff2afb4b3664e95f5c63cae3a524379/buildSrc/src/main/kotlin/CrateSet.kt#L18-L31

rust-runtime/aws-smithy-http-auth/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+pub mod auth;
```
jdisanti: Need to add this to the top:
```rust
#![warn(
    missing_debug_implementations,
    missing_docs,
    rustdoc::all
)]
```

rust-runtime/aws-smithy-http-auth/src/auth.rs
```diff
@@ -39,42 +41,49 @@ impl From<AuthErrorKind> for AuthError {
 }
 
 /// Authentication configuration to connect to a Smithy Service
-#[derive(Clone, Debug, PartialEq)]
-pub struct AuthApiKey {
-    api_key: String,
+#[derive(Clone, Eq, PartialEq)]
+pub struct AuthApiKey(Arc<Inner>);
```
jdisanti: Why `Arc` this?

rust-runtime/aws-smithy-http-auth/src/auth.rs
```diff
@@ -187,11 +223,61 @@ impl HttpAuthDefinition {
...
+        }
+
+        /// Sets the the HTTP auth name.
+        pub fn name(mut self, name: impl Into<String>) -> Self {
+            self.name = name.into();
+            self
+        }
+
+        /// Sets the HTTP auth scheme.
+        pub fn scheme(mut self, scheme: impl Into<Option<String>>) -> Self {
```
jdisanti: Our general pattern for this is for the main builder method to take a non-option, and then have a second `set_thing` method that takes option:
```rust
pub fn scheme(mut self, scheme: impl Into<String>) -> Self { ... }
pub fn set_scheme(&mut self, scheme: Option<String>) -> &mut Self { ... }
```

Also note the difference in the `self` types.

We follow this on non-optional fields too so that they can be unset. So the equivalent `set_` methods should be added to `location` and `name`.

rust-runtime/aws-smithy-http-auth/src/auth.rs
```diff
@@ -187,11 +223,61 @@ impl HttpAuthDefinition {
...
+            let scheme: Option<String> = scheme.into();
+            self.scheme = scheme;
+            self
+        }
+
+        /// Constructs a [`HttpAuthDefinition`] from the builder.
+        pub fn build(self) -> Result<HttpAuthDefinition, AuthError> {
+            if self.scheme.is_some() && !self.name.eq_ignore_ascii_case("authorization") {
+                return Err(AuthError::from(AuthErrorKind::SchemeNotAllowed));
```
jdisanti: Seems to me like this should panic instead. This seems like encountering this would be programmer error rather than caused by user input.

jdisanti (commented):
Looking better! After this round of comments, we just need to figure out something for removing the changes from `aws-types`.

rust-runtime/aws-smithy-http-auth/src/auth.rs
jdisanti: That would be fine.

aws/rust-runtime/aws-types/src/sdk_config.rs
eduardomourar: In regards to this, we are actually generating our own SDKs following very closely the pattern from the AWS SDK itself. We do that by using the parts from the Rust client codegen that are relevant for our own services.

Also, we want to use some of the default that are present in the [aws-config crate](https://crates.io/crates/aws-config). I took the approach of recreating our own config version which builds the same `SdkConfig` as defined in the [aws-types crate](https://crates.io/crates/aws-types). The reason being that we can have services with different types of authentication mechanism, one of them is actually passing the AWS provided credentials.

eduardomourar: What would you suggest us do? Create our own version of the `SdkConfig` containing the base options for every service we have?

jdisanti: My initial idea of how this would need to work is:

- The API auth key is configurable at the smithy-rs generic client level (below the SDK level). That way, generic non-SDK clients can use HTTP auth keys like the Smithy spec calls for.
- Your client code generator adds a customization (via `ClientCodegenDecorator`--lots of examples in `sdk-codegen`) to exercise the underlying generic client configuration option.

At a high level, codegen customizations are classes on the codegen classpath that inject additional code into different hook-points. All the differences between the AWS SDK and the generic Smithy clients are done entirely via customizations (and the additional crates that get pulled in by those customizations).

Problem is: `aws-config`/`aws-types` are not customizable since they are not code generated. So any mechanism you customize in would need to be outside of those somehow. It seems hacky, but you may be able to re-use the `Credentials` type for your API key configuration, and customize converting that into the HTTP auth.

Thinking about it more though... the vast majority of the code in `aws-config` is centered around AWS credential resolution, so it's probably better to modify your code generator to drop the need for `aws-config` entirely, and rely on the generic `Client` instantiation. You could even code generate your own more convenient client construction interface if desired. Is there anything in `aws-config` that you absolutely need?

Think about the above for a bit and let me know how you want to proceed, and then I can provide more information on how to accomplish that approach.

eduardomourar: Your idea matches mine to the point that:
- The API auth key has been added in [this PR](https://github.com/awslabs/smithy-rs/pull/2154) (although some modification is required due to changes done here)
- We have our own codegen doing the proper customization as you described.

For now the important parts are API key, retry config, timeout config, our own concept of region, HTTP connector and endpoint resolver. Eventually we will need the AWS Credentials, but we can leave for a second wave of changes we will need to do next month.

eduardomourar: With that said, I will start to port the `SdkConfig` and friends to our own packages in order to eliminate this dependency and I believe I can revert the commits related to the `aws-types` crate.

rust-runtime/aws-smithy-http-auth/src/auth.rs
eduardomourar: For the same reason specified here: https://github.com/awslabs/smithy-rs/blob/76ee00eb1737964ce240ae817f573980920f0607/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs#L18-L19

rust-runtime/aws-smithy-http-auth/src/definition.rs
```diff
@@ -142,17 +155,27 @@ pub mod http_auth_definition {
...
-            HttpAuthDefinition {
-                location: self.location,
-                name: self.name,
+            Ok(HttpAuthDefinition {
+                location: self.location.ok_or_else(|| {
+                    AuthError::from(AuthErrorKind::MissingRequiredField("location"))
+                })?,
+                name: self
+                    .name
+                    .ok_or_else(|| AuthError::from(AuthErrorKind::MissingRequiredField("name")))?,
```
jdisanti, line 176: These also seem like they should panic since they are cases where the builder is being used incorrectly.

jdisanti (commented):
This looks good to me! Just one minor comment on the builder. If everything passes CI after that, I'll be ready to merge.

jdisanti:
We're working on getting a patch release out, so I'm going to hold off on merging this until that goes out (probably early next week).

jdisanti:
I think the CI issues here will be resolved with https://github.com/awslabs/smithy-rs/pull/2359. Will update the PR after the merges.

jdisanti:
Heh, this clippy lint won't stop haunting me:
```
error: you are deriving `PartialEq` and can implement `Eq`
  --> sdk/aws-smithy-http-auth/src/definition.rs:16:33
   |
16 | #[derive(Clone, Debug, Default, PartialEq)]
   |                                 ^^^^^^^^^ help: consider deriving `Eq` as well: `PartialEq, Eq`
   |
   = note: `-D clippy::derive-partial-eq-without-eq` implied by `-D warnings`
```
I'll push a fix up in a bit...

PR #2155, opened 2023-01-02
Redact sensitive structs in debug

82marbag:
## Motivation and Context
Our implementation does not redact sensitive structures from logs and debugs.

Pending https://github.com/awslabs/smithy/issues/1561, the current implementation follows our current approach: sensitive are the values, not names. The structure name is added but no members are.

```
@sensitive
structure MyStruct { ... }
```
will be represented as: `MyStruct` (the name). Similarly for a builder.

## Testing
Tests have been added.

Closes: #2156

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt
@@ -92,14 +92,14 @@ fun ServiceShape.hasEventStreamOperations(model: Model): Boolean = operations.an
 
 fun Shape.shouldRedact(model: Model): Boolean =
     when (this) {
-        is MemberShape -> model.expectShape(this.target).shouldRedact(model)
+        is MemberShape -> model.expectShape(this.target).shouldRedact(model) || model.expectShape(this.container).shouldRedact(model)
         else -> this.hasTrait<SensitiveTrait>()
     }
 
 const val REDACTION = "\"*** Sensitive Data Redacted ***\""
 
-fun Shape.redactIfNecessary(model: Model, safeToPrint: String, sensitiveMember: Boolean = false): String =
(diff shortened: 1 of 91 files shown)


codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGeneratorTest.kt
```diff
@@ -118,4 +119,25 @@ internal class BuilderGeneratorTest {
...
+        builderGenerator.render(writer)
+        writer.implBlock(secretStructure, provider) {
+            builderGenerator.renderConvenienceMethod(this)
+        }
+        writer.compileAndTest(
+            """
+            use super::*;
+            let builder = SecretStructure::builder()
+                .secret_field("secret");
+                 assert_eq!(format!("{:?}", builder), "Builder");
```
jjant, line 139: I think this is one indent too far.

jjant (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt
```diff
@@ -98,8 +98,8 @@ fun Shape.shouldRedact(model: Model): Boolean =
 
 const val REDACTION = "\"*** Sensitive Data Redacted ***\""
 
-fun Shape.redactIfNecessary(model: Model, safeToPrint: String): String =
-    if (this.shouldRedact(model)) {
+fun Shape.redactIfNecessary(model: Model, safeToPrint: String, sensitiveMember: Boolean = false): String =
+    if (sensitiveMember || this.shouldRedact(model)) {
```
jjant, line 102: Shouldn't this `sensitiveMember` check be covered by the `shouldRedact` function? Probably checking `shouldRedact` on its `container`?

jjant (approved):

PR #2157, opened 2023-01-03
Undeprecate `OperationExtension` and add `Plugin`

hlbarber:
## Motivation and Context

There still exist use cases for a central `OperationExtension`.

## Description

- Undeprecate `OperationExtension`
- Add `OperationExtensionPlugin`


jjant (commented):
> There still exist use cases for a central `OperationExtension`.

Can you add an example?

jjant (approved):

PR #2159, opened 2023-01-03
Move `ClientBuilderGenerator` out of `codegen-client`

### LINKED ISSUE #1970: Move `BuilderGenerator.kt` (the entire file), together with its tests, to `codegen-client` (and rename it to `ClientBuilderGenerator`).

jjant:
## Motivation and Context
Closes #1970.

## Description
Move a whole bunch of client-specific code from `codegen-core` to `codegen-client`.

----


codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/InstantiatorTest.kt
```diff
@@ -229,12 +231,12 @@ class InstantiatorTest {
...
-            "k2": { "map": { "k3": {} } },
-            "k3": { }
-        }
-        """,
+            {
+                "k1": { "map": {} },
+                "k2": { "map": { "k3": {} } },
+                "k3": { }
+            }
+            """,
```
jjant, line 239: ktlint complained here and below.

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/InstantiatorTest.kt
```diff
@@ -91,7 +91,9 @@ class InstantiatorTest {
     // This is the exact same behavior of the client.
     private class BuilderKindBehavior(val codegenContext: CodegenContext) : Instantiator.BuilderKindBehavior {
         override fun hasFallibleBuilder(shape: StructureShape) =
-            BuilderGenerator.hasFallibleBuilder(shape, codegenContext.symbolProvider)
+            // ClientBuilderGenerator.hasFallibleBuilder(shape, codegenContext.symbolProvider)
+            // TODO: Not sure how to fix this. Maybe this module should be in codegen-client too?
+            false
```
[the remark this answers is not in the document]
jdisanti: You can take a look at the event stream tests to see how I abstracted out the builder generator so that it can be used by both the client and server tests. Basically, the common test code is in `codegen-core` in `EventStreamTestTools.kt`, and then there are Client/Server test cases (e.g., `ServerEventStreamMarshallerGeneratorTest.kt`) in the `codegen-client`/`codegen-server` modules that run the actual test cases using the correct builders.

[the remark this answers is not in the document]
jjant: I ended up moving these tests directly to `codegen-client`. They literally used a copy-pasted version of `ClientBuilderKindBehavior` so they were pretty much just more `ClientInstantiator` tests.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
```diff
@@ -94,17 +93,24 @@ internal fun testCodegenContext(
     codegenTarget,
 )
 
+interface BuilderGenerator {
+    fun render(writer: RustWriter)
+    fun renderConvenienceMethod(implBlock: RustWriter)
+}
```
jdisanti, line 99: I'm not a fan of this since it makes the builders inherit from unit test code, but it seems like this might go away due to #2342? Perhaps we merge that PR first and then delete this `BuilderGenerator` interface.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -370,3 +376,23 @@ open class Instantiator(
         else -> throw CodegenException("Unrecognized shape `$shape`")
     }
 }
+fun StructureShape.builderSymbol(symbolProvider: RustSymbolProvider): Symbol {
```
jdisanti, line 379: We should add a `symbolForBuilder(shape: Shape)` to `RustSymbolProvider` instead of this approach (there are examples in there for operation error and event stream error symbols). The module location needs to be configurable for my crate reorganization work.

jdisanti (dismissed):
This looks mostly good, just a couple comments.

My current work on crate reorganization is getting blocked on this PR, so I'm open to just merging it as is (once CI passes) and I can address the comments myself.

jdisanti:
Hmmm... Digging into this more, I'm thinking we should actually keep a minimal `BuilderGenerator` in `codegen-core` simply due to the number of unit tests that require builders. The protocol tests should stay in `codegen-core`, or otherwise, `codegen-core` doesn't have adequate testing on its own.

Velfi:

PR #2164, opened 2023-01-04
Enforce downwards walker traversal

hlbarber:
## Motivation and Context

The desired behavior of our walkers is to traverse downwards through the tree of shapes.

## Description

- Filter out `RelationshipDirection.INVERTED` for all usages of `walkShapes`.


82marbag (approved):

david-perez (changes requested):
I'd prefer it if we created a `DirectedWalker`/ `DownwardsWalker` class in `codegen-core` that applies that predicate.

david-perez (approved):

PR #2168, opened 2023-01-04
Enable FIPS and dual-stack from SDK config

### LINKED ISSUE #2143: Add FIPS/Dualstack support for EP 2.0

rcoh:
## Motivation and Context
- #1784 
- fixes #2143 

## Description
- Wire up FIPS & dualstack providers for SdkConfig
- Move tests from `credentials-provider` that belong in `RetryConfig`

## Testing
- local and integration tests

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/lib.rs
+++ b/aws/rust-runtime/aws-config/src/lib.rs
@@ -156,6 +156,7 @@ mod loader {
     use aws_smithy_types::retry::RetryConfig;
     use aws_smithy_types::timeout::TimeoutConfig;
     use aws_types::app_name::AppName;
+    use aws_types::docs_for;
     use aws_types::endpoint::ResolveAwsEndpoint;
     use aws_types::SdkConfig;
 
@@ -445,24 +446,13 @@ mod loader {
             self
         }
 
-        /// When true, send this request to the FIPS-compliant regional endpoint.
-        ///
-        /// If the configured endpoint does not have a FIPS compliant endpoint, dispatching
-        /// the request will return an error.
-        ///
-        /// **Note**: Not all services and regions support FIPS. If a service does not support FIPS,
-        /// this setting will have no effect.
+        #[doc = docs_for!(use_fips)]
         pub fn use_fips(mut self, use_fips: bool) -> Self {
             self.use_fips = Some(use_fips);
             self
         }
 
-        /// When true, send this request to the dual-stack endpoint.
-        ///
-        /// If the configured endpoint does not support dual-stack, the request MAY return an error.
-        ///
-        /// **Note**: Not all services and regions support dual-stack. If a service does not support
-        /// dual-stack, this setting will have no effect.
+        #[doc = docs_for!(use_dual_stack)]
         pub fn use_dual_stack(mut self, use_dual_stack: bool) -> Self {
             self.use_dual_stack = Some(use_dual_stack);
             self
--- a/aws/rust-runtime/aws-config/src/lib.rs
+++ b/aws/rust-runtime/aws-config/src/lib.rs
@@ -160,6 +160,8 @@ mod loader {
     use aws_types::SdkConfig;
 
     use crate::connector::default_connector;
+    use crate::default_provider::use_dual_stack::use_dual_stack_provider;
+    use crate::default_provider::use_fips::use_fips_provider;
     use crate::default_provider::{app_name, credentials, region, retry_config, timeout_config};
     use crate::meta::region::ProvideRegion;
     use crate::profile::profile_file::ProfileFiles;
@@ -185,6 +187,8 @@ mod loader {
         http_connector: Option<HttpConnector>,
         profile_name_override: Option<String>,
         profile_files_override: Option<ProfileFiles>,
+        use_fips: Option<bool>,
+        use_dual_stack: Option<bool>,
     }
 
     impl ConfigLoader {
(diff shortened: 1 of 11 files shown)


aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -441,6 +445,29 @@ mod loader {
             self
         }
 
+        /// When true, send this request to the FIPS-compliant regional endpoint.
+        ///
+        /// If the configured endpoint does not have a FIPS compliant endpoint, dispatching
+        /// the request will return an error.
+        ///
+        /// **Note**: Not all services and regions support FIPS. If a service does not support FIPS,
+        /// this setting will have no effect.
```
Velfi, line 454:
```suggestion
        /// When true, send this request to the FIPS-compliant regional endpoint.
        ///
        /// If the configured endpoint it not a FIPS-compliant endpoint, dispatching
        /// the request will return an error.
        ///
        /// **Note**: Not all services and regions support FIPS. If a service does not support FIPS,
        /// this setting will have no effect.
```

I'm having a bit of trouble understanding which cases will return an error. Is this table correct?

| `use_fips = <bool>` | service supports FIPS | will error |
| -------------------- | ---------------------- | --------- |
| true | yes | no |
| true | no | yes |
| false | yes | no |
| false | no | no|

What is the difference between a configured endpoint being FIPS compliant vs a service supporting FIPS?

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -441,6 +445,29 @@ mod loader {
...
+            self.use_fips = Some(use_fips);
+            self
+        }
+
+        /// When true, send this request to the dual-stack endpoint.
+        ///
+        /// If the configured endpoint does not support dual-stack, the request MAY return an error.
+        ///
+        /// **Note**: Not all services and regions support dual-stack. If a service does not support
+        /// dual-stack, this setting will have no effect.
```
Velfi, line 465: Same here, I'm not clear on the difference between an endpoint supporting dual-stack and a service supporting dual-stack

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -540,6 +567,18 @@ mod loader {
...
+            } else {
+                use_fips_provider(&conf).await
+            };
+
+            let use_dual_stack = if let Some(use_dual_stack) = self.use_dual_stack {
+                Some(use_dual_stack)
+            } else {
+                use_dual_stack_provider(&conf).await
+            };
+
```
Velfi, line 581: I wish there was a better way to handle futures containing options but it looks like `TryFutureExt` only covers `Result`s. I don't like having two `=` on one line but the following alternative doesn't actually work so 🤷‍♀️.

```rust
async fn foo() {
    let config = Config::new();

    let use_dual_stack = future::ready(config.use_dual_stack)
        .or_else(|| use_dual_stack_provider(&config))
        .await;
    
    println!("use_dual_stack == {use_dual_stack:?}");
}
```

aws/rust-runtime/aws-types/src/sdk_config.rs
```diff
@@ -461,6 +465,52 @@ impl Builder {
...
+    /// When true, send this request to the dual-stack endpoint.
+    ///
+    /// If the configured endpoint does not support dual-stack, the request MAY return an error.
+    ///
+    /// **Note**: Not all services and regions support dual-stack. If a service does not support
+    /// dual-stack, this setting will have no effect.
+    pub fn set_use_dual_stack(&mut self, use_dual_stack: Option<bool>) -> &mut Self {
+        self.use_dual_stack = use_dual_stack;
+        self
+    }
```
Velfi, line 512: I wish we could set the `ConfigLoader` docs to pull from these docs so that we only have to update them in one place. If you end up re-wording the `ConfigLoader` docs, don't forget these ones.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AddFIPSDualStackDecorator.kt
```diff
@@ -35,14 +40,11 @@ fun ClientCodegenContext.getBuiltIn(builtIn: Parameter): Parameter? {
...
-    val idx = EndpointRulesetIndex.of(model)
-    val service = model.expectShape(serviceId, ServiceShape::class.java)
-    val rules = idx.endpointRulesForService(service) ?: return model
     // load the builtIn with a matching name from the ruleset allowing for any docs updates
-    val builtIn = rules.getBuiltIn(builtInSrc) ?: return model
+    val builtIn = this.loadBuiltIn(serviceId, builtInSrc) ?: return model
 
     return ModelTransformer.create().mapShapes(model) { shape ->
-        if (shape !is ServiceShape || shape.id != service.id) {
+        if (shape !is ServiceShape || shape.id != serviceId) {
```
Velfi, line 47: Do we even need to check for the shape here? isn't checking for the ID match enough?

aws/rust-runtime/aws-config/src/lib.rs
rcoh: I was considering writing a `or_else_fut` trait but figured it wasn't worth it

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AddFIPSDualStackDecorator.kt
rcoh: checking if it's a service shape means Kotlin will infer it as a service shape in this branch

aws/rust-runtime/aws-config/src/lib.rs
Velfi: Yeah, I agree with you.

aws/rust-runtime/aws-types/src/sdk_config.rs
rcoh: done!

Velfi (approved):

Issue #1785, opened 2022-09-30
Adaptive Retry


jdisanti:
Implement adaptive retry and make it possible to switch between `standard` and `adaptive` for both `aws-config` and non-SDK clients.

PR #2171, opened 2023-01-05
RFC: Implementing Adaptive Retry Behavior

Velfi:
## Motivation and Context

#1785 

----

### DIFF
--- a/design/src/rfcs/rfc0031_implementing_adaptive_retry_behavior.md
+++ b/design/src/rfcs/rfc0031_implementing_adaptive_retry_behavior.md
@@ -1,0 +1,162 @@
+# Implementing Adaptive Retry Behavior
+
+> Status: RFC
+>
+> Applies to: client, server
+
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+This RFC defines how the Rust SDK will implement Adaptive Retry Behavior. This new behavior works on the principle of client-side rate limiting and only takes effect once a client has received a throttling error. The new behavior:
+- Is opt-in. The existing retry behavior will remain the default.
+- Will perform similarly to the existing retry behavior when comparing the rate of successful requests.
+- Will improve the experience of customers that use the SDK to send very large numbers of concurrent requests.
+
+*This RFC does not consider retry behavior related to expired credentials or clock skew errors.*
+
+## Terminology
+
+- **Retry (Behavior / Strategy)**: How the SDK reacts to retryable errors.
+- **Standard Retry Behavior (SRB)**: The default behavior. When a retryable error is received, this behavior calculates an exponential backoff time based on how many retry attemps have already been made and then sleeps for that long before making another attempt.
+- **Adaptive Retry Behavior (ARB)**: When sending requests, clients must acquire a token from a shared bucket. When responses are successful, the bucket capacity and fill rate are increased. When throttling responses are received, the bucket capacity and fill rate are decreased. **Transient Errors** do not affect the rate of token replenishment. When the bucket is empty, requests will sleep until more tokens become available.
+- **Token Bucket**: similar to a [`tokio::sync::Semaphore`][tokio::sync::Semaphore], a token bucket contains a limited amount of "tokens." Before making an HTTP request, we take a token from the bucket. If the request fails, we take another token and try again. If the request succeeds, we release any tokens we're holding. If the token bucket is empty, no requests may be made.
+- [**Retryable Trait**](https://awslabs.github.io/smithy/spec/core.html#retryable-trait): a smithy trait indicating that an error may be retried by the client.
+- **(HTTP / Retry) Request**: An HTTP request. One or more of these will be made during an **Operation Request**.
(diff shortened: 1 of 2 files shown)


design/src/rfcs/rfc0031_implementing_adaptive_retry_behavior.md
```diff
@@ -0,0 +1,162 @@
...
+>
+> Applies to: client, server
+
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+This RFC defines how the Rust SDK will implement Adaptive Retry Behavior. This new behavior works on the principle of client-side rate limiting and only takes effect once a client has received a throttling error. The new behavior:
+- Is opt-in. The existing retry behavior will remain the default.
+- Will perform similarly to the existing retry behavior when comparing the rate of successful requests.
+- Will improve the experience of customers that use the SDK to send very large numbers of concurrent requests.
+
```
drganjoo, line 13: I can see that the RFC is a draft but I had a few comments so I thought I might as well add them now.

1) I couldn't find anything on how does one turn off retries. Is it going to be:

```
    let retry_config = RetryConfig::disabled();
```

Which I am assuming is the same as setting:

```
    let retry_config = RetryConfig::standard().with_max_attempts(1);
```

2) Are we going to allow retry policy to be configured separately for specific operations? If yes, what does the API look for that?

3) The RFC doesn't mention the [RetryClassifier](https://github.com/awslabs/smithy-rs/blob/e2e0d9354548bbf2d76b660571cd0701d016771b/rust-runtime/aws-smithy-client/src/retry.rs#L375) that each operation can have. Can we please add a section on how does that play a role in the policy behavior? and, how does one specify a RetryClassifier on a particular operation or the overall client.

4) What metrics will the client emit by default? and:
a) Can a user disable metrics? 
b) Can a custom metric be emitted on a particular lifecycle event?

design/src/rfcs/rfc0031_implementing_adaptive_retry_behavior.md
```diff
@@ -0,0 +1,162 @@
...
+        - `ThrottlingException`
+        - `ThrottledException`
+        - `RequestThrottledException`
+        - `TooManyRequestsException`
+        - `ProvisionedThroughputExceededException`
+        - `TransactionInProgressException`
+        - `LimitExceededException`
+        - `PriorRequestNotComplete`
+    - `403` errors with the `RequestThrottled` code.
+    - `502` errors with the `EC2ThrottledException` code.
```
drganjoo, line 37: Just to clarify, a 502 that does not have `EC2ThrottledException` is still going to be a transient error?

Also, are the codes case sensitive?

design/src/rfcs/rfc0031_implementing_adaptive_retry_behavior.md
```diff
@@ -0,0 +1,162 @@
...
+- **Token Bucket**: similar to a [`tokio::sync::Semaphore`][tokio::sync::Semaphore], a token bucket contains a limited amount of "tokens." Before making an HTTP request, we take a token from the bucket. If the request fails, we take another token and try again. If the request succeeds, we release any tokens we're holding. If the token bucket is empty, no requests may be made.
+- [**Retryable Trait**](https://awslabs.github.io/smithy/spec/core.html#retryable-trait): a smithy trait indicating that an error may be retried by the client.
+- **(HTTP / Retry) Request**: An HTTP request. One or more of these will be made during an **Operation Request**.
+- **Operation Request**: A request to a smithy service. Each operation request may contain several **HTTP requests**.
+- **Throttling Error**: An error that should trigger the client's rate-limiting implementation. Currently, the Rust SDK has no special behavior for these errors. Once this RFC is implemented, the Rust SDK will treat the following as throttling errors:
```
drganjoo, line 25: I thought the current standard policy retries on Throttling / Transient errors?

design/src/rfcs/rfc0031_implementing_adaptive_retry_behavior.md
```diff
@@ -0,0 +1,162 @@
...
+    .initial_backoff(Duration::from_secs(1))
+    .build();
+
+let sdk_config = aws_config::from_env()
+    .retry_config(retry_config)
+    .load()
+    .await;
+```
+
+Because **ARB** works alongside **SRB** and isn't configurable, no new config fields will be added.
```
drganjoo, line 89: Can the size of token buckets be configured?

design/src/rfcs/rfc0031_implementing_adaptive_retry_behavior.md
```diff
@@ -0,0 +1,162 @@
...
+        } else {
+            Err(RetryModeParseError::new(string))
+        }
+    }
+}
+```
+
+### The Service
+
+Because `tower::Service`s already define an async `ready` method, we'll use that to block sending requests unless they can acquire a token from the token bucket.
```
drganjoo, line 122: I am assuming that in the orchestrator world, we are going to replace this section in the RFC?

PR #2172, opened 2023-01-05
refactor: extract retry quota into a tower service

Velfi:
## Motivation and Context

#1785 

## Description

This PR extracts the retry quota state from the `RetryHandler` and places it in its own `tower::Layer`. This is a first step in the process to introduce adaptive retry behavior which will use its own specialized retry quota layer.

## Testing

Ran existing tests

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-client/src/quota.rs
+++ b/rust-runtime/aws-smithy-client/src/quota.rs
@@ -33,13 +33,13 @@ impl<S, Tb: TokenBucket> QuotaService<S, Tb> {
     }
 }
 
-impl<S, H, R, Tb, To, E, T> Service<Operation<H, R>> for QuotaService<S, Tb>
+impl<S, H, R, Tb, To, T, E> Service<Operation<H, R>> for QuotaService<S, Tb>
 where
     S: Service<Operation<H, R>, Error = SdkError<E>, Response = SdkSuccess<T>>,
     S::Response: 'static,
     S::Future: 'static,
-    E: ProvideErrorKind + 'static,
     T: 'static,
+    E: ProvideErrorKind + 'static,
     To: Token + 'static,
     Tb: TokenBucket<Token = To>,
 {
@@ -70,19 +70,23 @@ where
         };
 
         Box::pin(async move {
-            let res = fut.await;
+            let mut res = fut.await;
             let classifier = DefaultResponseRetryClassifier::new();
             // If this request failed with a reason that's relevant to our token bucket,
             // insert that info into the property bag so we can retrieve the correct number
             // of tokens if the request is retried.
             let response_kind = classifier.classify_retry(res.as_ref());
-            // res.properties_mut().insert(response_kind);
 
-            if res.is_ok() {
-                token.release()
-            } else {
-                token.forget()
-            }
+            match &mut res {
+                Ok(success) => {
+                    token.release();
+                    success.raw.properties_mut().insert(response_kind)
+                }
+                Err(SdkError::ServiceError(failure)) => {
+                    token.forget();
+                    failure.raw().properties_mut().insert(response_kind)
+                }
+            };
 
             res
         })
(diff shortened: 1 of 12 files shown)


rust-runtime/aws-smithy-client/src/retry/adaptive.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// Construct a new standard retry policy from the given policy configuration.
+    pub fn new(config: Config) -> Self {
+        Self {
+            shared_state: CrossRequestRetryState::new(),
+            config,
+        }
+    }
+
+    /// Set the configuration for this retry policy.
+    pub fn with_config(&mut self, config: Config) -> &mut Self {
```
ysaito1001, line 81: Curious to know a use case of this method. A caller who created a `Policy` had the chance to specify `config` when calling `new` but somehow the caller wants to overwrite the initially specified `config` after that?

The same question applies to `standard::Policy::with_config`.

rust-runtime/aws-smithy-client/src/retry/adaptive.rs
```diff
@@ -0,0 +1,332 @@
...
+
+use crate::retry::{NewRequestPolicy, DEFAULT_MAX_ATTEMPTS};
+use aws_smithy_async::rt::sleep::AsyncSleep;
+use aws_smithy_http::operation::Operation;
+use aws_smithy_http::result::{SdkError, SdkSuccess};
+use aws_smithy_http::retry::ClassifyRetry;
+use aws_smithy_types::retry::{ErrorKind, RetryKind};
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::{Arc, Mutex};
```
ysaito1001, line 16: We use `std::sync::Mutex` as opposed to `tokio::sync::Mutex` is because the latter is tied to a specific async runtime construct and we want to avoid that?

rust-runtime/aws-smithy-client/src/retry/standard.rs
```diff
@@ -0,0 +1,597 @@
...
+
+    use super::{calculate_exponential_backoff, Config, Policy, RetryHandler};
+
+    fn test_config() -> Config {
+        Config::default().with_base(|| 1_f64)
+    }
+
+    #[test]
+    fn retry_handler_send_sync() {
+        fn must_be_send_sync<T: Send + Sync>() {}
```
ysaito1001, line 402: This is cool. I know there is a crate out there called [static-assertions](https://github.com/nvzqz/static-assertions-rs), but what we have here will do.

ysaito1001 (commented):
Looks awesome! Excited about these changes 👍

rust-runtime/aws-smithy-client/src/retry/adaptive.rs
Velfi: Ignore this code, I didn't intend to include it in this PR. I've removed it.

rust-runtime/aws-smithy-client/src/retry/adaptive.rs
Velfi: No. `tokio`'s Mutex actually works fine on any async runtime. The real issue is that locking it is async and our current retry code is all sync.

rust-runtime/aws-smithy-client/src/quota.rs
```diff
@@ -0,0 +1,156 @@
...
+    InnerService: tower::Service<Operation<H, R>, Error = SdkError<E>>,
+{
+    type Response = InnerService::Response;
+    type Error = SdkError<E>;
+    type Future = InnerService::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        if self.state.available() == 0 {
+            // If no quota is available, report pending
+            Poll::Pending
```
rcoh, line 121: this won't quite work—we need to schedule a wakeup when state goes back above 0. As written, if state goes to 0, that task will never get woken up

Velfi:
I'm going to focus on implementing the `TokenBucket` as its own testable thing before finishing up this PR.

PR #2183, opened 2023-01-08
Rfc30/feature gate for generated crates

thomas-k-cameron:
## Motivation and Context
part of RFC30

## Description
add `serde::Serialize/Deserialize` to `RuntimeType.kt`
add `CfgUnstable` on `DependencyScope`
add `serde::Serialize/Deserialize` trait to generated data types

## Testing
NA

## Checklist
TODO

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/CargoTomlGenerator.kt
```diff
@@ -54,6 +54,9 @@ class CargoTomlGenerator(
         if (features.isNotEmpty()) {
             cargoFeatures.add("default" to features.filter { it.default }.map { it.name })
         }
+        // add serde related features
+        cargoFeatures.add("serialize" to listOf("aws-smithy-types/serialize"))
+        cargoFeatures.add("deserialize" to listOf("aws-smithy-types/deserialize"))
```
LukeMathWalker, line 59: I'm not super familiar with the feature flag machinery here, but are `serialize`/`deserialize` going to be the names of the feature flags?

thomas-k-cameron: yup!

let me prepare you an example on another crate.

LukeMathWalker: I would suggest going for `serde-deserialize`/`serde-serialize` instead of `serialize`/`deserialize` - we might want to add support for other (de)serialization crates down the line.

thomas-k-cameron: If I understand it correctly, it was agreed that name of the feature gate is going to become `serialize`/`deserialize` on RFC30.

https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0030_serialization_and_deserialization.md

```
Feature Gate for Serialization and De-serialization

Serde traits are implemented behind feature gates. Serialize is implemented behind serialize, while Deserialize is implemented behind deserialize. Users must enable the unstable feature to expose those features.

We considered giving each feature a dedicated feature gate such as unstable-serde-serialize. In this case, we will need to change the name of feature gates entirely once it leaves the unstable status which will cause users to make changes to their code base. We conclude that this brings no benefit to the users.
```

You can find the discussion here.
https://github.com/awslabs/smithy-rs/pull/1944

Furthermore, those feature gates can be enabled only when aws_sdk_unstable is enabled via cfg flag.
like this.
```bash
export RUSTFLAGS="--cfg aws-sdk-unstable"
cargo build
```

thomas-k-cameron: @Velfi 
Am I correct to believe that we are going to implement the way the RFC describes including the name of the feature-gate?

@LukeMathWalker 
@Velfi was heavily involved in the discussion. I'm going to mention her to clarify the situation.

LukeMathWalker: The RFC touches on a few crates that are co-owned by the service framework team (e.g. aws-smithy-http) - I realised that was the case too late (a.k.a. after it was merged), so that's why I'm jumping in on the implementation PR 😁 

The issue I am raising is not about us removing those flags - it is about us wanting to support, for example, the [`deser`](https://github.com/mitsuhiko/deser) crate alongside `serde` as a serialization option. Including the crate name in the feature name allows for this addition without causing confusion.

thomas-k-cameron: Thank you for letting me know!

Yeah, I think you make sense; I prefer to have the RFC updated just to make sure that everyone is on the same pages.

Would you kindly create a PR to update the RFC?

thomas-k-cameron: by the way, here is a repository to show case how feature-gate works.
https://github.com/thomas-k-cameron/feature-gate-examples

thomas-k-cameron: https://github.com/awslabs/smithy-rs/pull/2183#discussion_r1064573442

https://github.com/awslabs/smithy-rs/pull/2186

@LukeMathWalker 

I just created the PR for myself.
can you check it out?

LukeMathWalker: Oh, I now realise the confusion - when I said I wasn't familiar with the feature gate machinery I meant that I'm not familiar with how we handle that in the Kotlin code generator, not in Rust per se 😅 Sorry for having waste your time there 😅

thomas-k-cameron: oh snap

no worries :D

Well, this should work but I'm not too familiar with it too. 

That is why it is in a `draft` state.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/CargoDependency.kt
```diff
@@ -16,6 +16,7 @@ import java.nio.file.Path
 sealed class DependencyScope {
     object Dev : DependencyScope()
     object Compile : DependencyScope()
+    object CfgUnstable : DependencyScope()
```
Velfi, line 19: This is a cool way of doing this!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
```diff
@@ -239,6 +239,10 @@ data class RuntimeType(val path: String, val dependency: RustDependency? = null)
         val ConstrainedTrait = RuntimeType("crate::constrained::Constrained", InlineDependency.constrained())
         val MaybeConstrained = RuntimeType("crate::constrained::MaybeConstrained", InlineDependency.constrained())
 
+        // serde types. They will be behind feature gates
```
Velfi, line 242: Avoid making comments about the future. All we can know is the present.

```suggestion
        // serde types. Gated behind `CfgUnstable`.
```

thomas-k-cameron: OK.

thomas-k-cameron:
@Velfi 

https://github.com/awslabs/smithy-rs/pull/1944#issuecomment-1371160038

Your comment mentions `extras` method.

>Create a SerdeDecorator class that extends ClientCodegenDecorator. This class should be declared in its own file, probably in the smithy-rs/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations folder. 

>When overriding the extras method, you get access to the RustCrate being generated and you can add features to it by calling rustCrate.mergeFeature(Feature("<your feature name>", default = true, listOf(<every CargoDependency needed by this feature>))).

>The decorator can add all features that you want to define.

But there `ClientCodegenDecorator` has no `extras` method.

What am I missing?

Velfi:
> What am I missing?

`ClientCodegenDecorator` extends `CoreCodegenDecorator<ClientCodegenContext>`
`CoreCodegenDecorator` is the class that defines an `extras` method.
You'll want to extend `ClientCodegenDecorator` and add an `override` for `extras`.

If you search for `override fun extras(codegenContext: ClientCodegenContext, rustCrate: RustCrate) {` you can find plenty of decorators to use as examples. I hope that's enough to get you going, but let me know if it isn't.

thomas-k-cameron:
Thanks it seems to be working now. No idea why to be honest because the compiler was complaining to me that there is no `fun extras` to override.

thomas-k-cameron:
There are still things that I haven't figured out so I think I need more time.
https://github.com/awslabs/smithy-rs/discussions/2182

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/UnstableDerive.kt
```diff
@@ -0,0 +1,17 @@
...
+        all(aws_sdk_unstable, feature = "serde-serialize")
+    """
+    val AttrUnstableDeserialize = """
+        all(aws_sdk_unstable, feature = "serde-deserialize")
+    """
+
+    // double `#` is because this data is passed onto writeInLine, which will interpret it as a variable with single `#`
+    val UnstableDerive = """##[cfg_attr(all(aws_sdk_unstable, feature = "serde-serialize"), derive(serde::Serialize))]
+##[cfg_attr(all(aws_sdk_unstable, feature = "serde-deserialize"), derive(serde::Deserialize))]"""
+}
```
Velfi, line 17: We have a new way of constructing attributes and you can add these to the `Attribute` class' companion object. Here's what that would look like:

```kotlin
        fun cfg_attr(vararg attrMacros: Writable): Writable = macroWithArgs("cfg_attr", *attrMacros)
        val UnstableSerdeAny = Attribute(cfg(all(writable("aws_sdk_unstable"), any(pair("feature" to "serde-serialize"), pair("feature" to "serde-deserialize")))))
        val UnstableSerdeAll = all(writable("aws_sdk_unstable"), all(pair("feature" to "serde-serialize"), pair("feature" to "serde-deserialize")))
        val UnstableSerialize = all(writable("aws_sdk_unstable"), pair("feature" to "serde-serialize"))
        val UnstableDeserialize = all(writable("aws_sdk_unstable"), pair("feature" to "serde-deserialize"))
        val UnstableSerdeDerive = writable {
            Attribute(cfg_attr(all(writable("aws_sdk_unstable"), pair("feature" to "serde-serialize"), derive(CargoDependency.Serde.resolve("Serialize"))))).render(this)
            Attribute(cfg_attr(all(writable("aws_sdk_unstable"), pair("feature" to "serde-deserialize"), derive(CargoDependency.Serde.resolve("Deserialize"))))).render(this)
        }
```

Because we actually reference the `CargoDependency` for `serde` in `UnstableDerive`, we'll automatically get it added to the `Cargo.toml` whenever we write `UnstableDerive`.

All of these are `Writable` and you can write them by calling `.render(this)` where `this` is a rust writer:

```kotlin
import software.amazon.smithy.rust.codegen.core.rustlang.Attribute

// ...
private fun someRenderFun(writer: RustWriter) {
    Attribute.UnstableSerdeDerive.render(writer)
}
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -208,7 +209,11 @@ class BuilderGenerator(
     private fun renderBuilder(writer: RustWriter) {
         writer.docs("A builder for #D.", structureSymbol)
         Attribute(derive(builderDerives)).render(writer)
+        writer.writeInline("/// This is the datatype that Builder of this module build itself into.\n")
+        writer.writeInline("pub type OutputShape = #D", structureSymbol)
```
Velfi, line 213: Prefer the Rust-specific writer fns
```suggestion
        writer.rustInline("pub type OutputShape = #D", structureSymbol)
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/UnstableDerive.kt
thomas-k-cameron: Thanks for the help!

thomas-k-cameron:
I think it's ready for reviews!

thomas-k-cameron:
I think it's ready for reviews!

thomas-k-cameron:
I think there are some missing. Let me revert that.

thomas-k-cameron:
Ok, so thanks to NPE the piece of code doesn't work for some reason. No idea why.

I decided to replace it with a string for a time being to move forward with testing other todos ... etc.

thomas-k-cameron:
https://github.com/awslabs/smithy-rs/discussions/2248#discussioncomment-4826519

thomas-k-cameron:
Import errors are handled here https://github.com/awslabs/smithy-rs/pull/2185.

thomas-k-cameron:
I think I messed it up. Let me fix it.

thomas-k-cameron:
@jdisanti 
Can you review this one please?

thomas-k-cameron:
@Velfi 
Would you kindly review this one please?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -206,13 +207,19 @@ class BuilderGenerator(
     }
 
     private fun renderBuilder(writer: RustWriter) {
+        writer.docs("This is the datatype that Builder of this module build itself into.")
```
Velfi, line 210:
```suggestion
        writer.docs("This is the datatype returned when calling `Builder::build()`.")
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustType.kt
```diff
@@ -446,6 +446,16 @@ class Attribute(val inner: Writable) {
...
 
+    // These were supposed to be a part of companion object but we decided to move it out to here to avoid NPE
+    // You can find the discussion here.
+    // https://github.com/awslabs/smithy-rs/discussions/2248
+    public fun SerdeSerialize(): Attribute {
+        return Attribute(cfgAttr(all(writable("aws_sdk_unstable"), feature("serde-serialize")), derive(RuntimeType.SerdeSerialize)))
+    }
+    public fun SerdeDeserialize(): Attribute {
+        return Attribute(cfgAttr(all(writable("aws_sdk_unstable"), feature("serde-deserialize")), derive(RuntimeType.SerdeDeserialize)))
+    }
```
Velfi, line 457: There's no need for these to be functions. Change them into values and put them in the companion object below.
```suggestion
    val SerdeSerialize = Attribute(cfgAttr(all(writable("aws_sdk_unstable"), feature("serde-serialize")), derive(RuntimeType.SerdeSerialize)))
    val SerdeDeserialize = Attribute(cfgAttr(all(writable("aws_sdk_unstable"), feature("serde-deserialize")), derive(RuntimeType.SerdeDeserialize)))
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/RenderSerdeAttribute.kt
```diff
@@ -0,0 +1,25 @@
...
+public object RenderSerdeAttribute {
+    public fun forStructureShape(writer: RustWriter, shape: StructureShape, model: Model) {
+        if (shape.members().none { it.isEventStream(model) }) {
+            writeAttributes(writer)
+        }
+    }
+
+    public fun writeAttributes(writer: RustWriter) {
+        Attribute("").SerdeSerialize().render(writer)
+        Attribute("").SerdeDeserialize().render(writer)
```
Velfi, line 23: After making the change suggested in my earlier comment, you can update this function like so:

```suggestion
        Attribute.SerdeSerialize.render(writer)
        Attribute.SerdeDeserialize.render(writer)
```

Velfi (changes requested):
For the most part, this is good. I only see a few things that need updating and then we can merge it.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustType.kt
thomas-k-cameron: This causes NPE too.

thomas-k-cameron:
@Velfi 

I left some comments.  
Your change request would cause NPE.

I don't want to rush you, but I really appreciate if you could take a look at it.

thomas-k-cameron:
@Velfi 

Hi,

Would you kindly give me some replies to my comment?

Velfi (approved):
@thomas-k-cameron Sorry about the wait on this one. I'll put it in the merge queue

PR #2191, opened 2023-01-10
Add static stability support to IMDS credentials provider 

ysaito1001:
## Motivation and Context
Implements https://github.com/awslabs/smithy-rs/issues/2117

## Description
This PR adds static stability support to the IMDS credentials provider. The basic idea is that `ImdsCredentialsProvider` now holds on to a last retrieved credentials and if it cannot fetch the latest credentials from IMDS possibly due to the service being unavailable, it will serve the last retrieved credentials instead. That way, the Rust SDK will not fail fast in sending a request with expired credentials  and allows the target service to make the ultimate decision as to whether the request sent is valid.

Even though IMDS is available, it may return stale credentials. Static stability support specifies that `ImdsCredentialsProvider` should extend the credentials expiry by the calculated amount. The implementation of that is defined in `ImdsCredentialsProvider::extend_expiration` and references that in [botocore](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L652-L685).

The code changes in the PR that accomplish the above are mostly in `aws-config`. However, we found out that those changes alone are not enough to meet requirements for static stability support. Specifically, the timeout logic in `LazyCredentialsCache` prior to this PR did not work well with the timeout logic required by the updated `ImdsCredentialsProvider` (see cfeab3b for more details). The bottom lines is that the additional changes have been made in `aws-credential-types` and the `ProvideCredentials` trait now has a new method `provide_credentials_with_timeout` to address the said impedance mismatch.

## Testing
- Expired Credential Test Cases
  - [x] SDK can use IMDS credential provider if first IMDS call returns expired credentials (implemented by `test_request_should_be_sent_when_first_call_to_imds_returns_expired_credentials`).
  - [x] SDK can send request if expired credentials are available (there's no explicit test added for this as it is implicitly handled by two tests for `Refresh Failures`).
  - [x] SDK can perform 3 successive requests with expired credentials. IMDS must only be called once (implemented by `test_successive_requests_should_be_sent_with_expired_credentials_and_imds_being_called_only_once`).
- Refresh Failures
  - [x] SDK can send a request after a read timeout during a refresh (implemented by `read_timeout_during_credentials_refresh_should_yield_last_retrieved_credentials`).
  - [x] SDK can send a request after receiving a 500 HTTP response from IMDS during a refresh (implemented by `test_request_should_be_sent_with_expired_credentials_after_imds_returns_500_during_credentials_refresh`).
- Logging
  - [x] SDK must log a message when expired credentials are used (implemented by `log_message_informing_expired_credentials_are_used`).
- [ ] Passed tests in CI

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/imds/credentials.rs
+++ b/aws/rust-runtime/aws-config/src/imds/credentials.rs
@@ -14,11 +14,20 @@ use crate::imds::client::LazyClient;
 use crate::json_credentials::{parse_json_credentials, JsonCredentials, RefreshableCredentials};
 use crate::provider_config::ProviderConfig;
 use aws_credential_types::provider::{self, error::CredentialsError, future, ProvideCredentials};
+use aws_credential_types::time_source::TimeSource;
 use aws_credential_types::Credentials;
+use aws_smithy_async::future::timeout::Timeout;
+use aws_smithy_async::rt::sleep::AsyncSleep;
 use aws_types::os_shim_internal::Env;
+use fastrand;
 use std::borrow::Cow;
 use std::error::Error as StdError;
 use std::fmt;
+use std::sync::Arc;
+use std::time::{Duration, SystemTime};
+use tokio::sync::RwLock;
+
+const CREDENTIAL_EXPIRATION_INTERVAL: Duration = Duration::from_secs(15 * 60);
 
 #[derive(Debug)]
 struct ImdsCommunicationError {
@@ -45,6 +54,8 @@ pub struct ImdsCredentialsProvider {
     client: LazyClient,
     env: Env,
     profile: Option<String>,
+    time_source: TimeSource,
+    last_retrieved_credentials: Arc<RwLock<Option<Credentials>>>,
 }
 
 /// Builder for [`ImdsCredentialsProvider`]
@@ -53,6 +64,7 @@ pub struct Builder {
     provider_config: Option<ProviderConfig>,
     profile_override: Option<String>,
     imds_override: Option<imds::Client>,
+    last_retrieved_credentials: Option<Credentials>,
 }
 
 impl Builder {
@@ -86,6 +98,12 @@ impl Builder {
         self
     }
 
+    #[cfg(test)]
+    fn last_retrieved_credentials(mut self, credentials: Credentials) -> Self {
+        self.last_retrieved_credentials = Some(credentials);
+        self
+    }
+
     /// Create an [`ImdsCredentialsProvider`] from this builder.
     pub fn build(self) -> ImdsCredentialsProvider {
         let provider_config = self.provider_config.unwrap_or_default();
@@ -102,6 +120,8 @@ impl Builder {
             client,
             env,
             profile: self.profile_override,
+            time_source: provider_config.time_source(),
+            last_retrieved_credentials: Arc::new(RwLock::new(self.last_retrieved_credentials)),
         }
     }
 }
@@ -116,6 +136,17 @@ impl ProvideCredentials for ImdsCredentialsProvider {
         Self: 'a,
     {
         future::ProvideCredentials::new(self.credentials())
+    }
+
+    fn provide_credentials_with_timeout<'a>(
+        &'a self,
+        sleeper: Arc<dyn AsyncSleep>,
+        timeout: Duration,
+    ) -> future::ProvideCredentials<'a>
+    where
+        Self: 'a,
+    {
+        future::ProvideCredentials::new(self.credentials_with_timeout(sleeper, timeout))
     }
 }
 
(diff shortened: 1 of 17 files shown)


aws/sdk/integration-tests/Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
     "ec2",
     "glacier",
     "iam",
+    "imds",
```
ysaito1001, line 9: Move all of these to `aws_config`.

aws/rust-runtime/aws-config/src/meta/credentials/chain.rs
```diff
@@ -95,6 +98,35 @@ impl CredentialsProviderChain {
...
         ))
     }
+
+    async fn credentials_with_timeout(
+        &self,
+        sleeper: Arc<dyn AsyncSleep>,
+        timeout: Duration,
+    ) -> provider::Result {
+        for (name, provider) in &self.providers {
+            let timeout_per_provider = timeout.div_f64(self.providers.len() as f64);
```
ysaito1001, line 108: This cannot be a simple division as each provider has a different idea of how long timeout should be.

ysaito1001, line 108: This cannot be a simple division as each provider has a different idea of how long timeout should be. This implies that we should consider what it means for a credentials provider to time out in a broader scope.

ysaito1001: At least preserved today's behavior with 4592ca5, but a set of APIs `provide_credentials` and `provide_credentials_with_timeout` needs to be reconsidered.

aws/sdk/integration-tests/Cargo.toml
ysaito1001: Addressed in 3ca1ab7.

aws/rust-runtime/aws-config/src/imds/credentials.rs
```diff
@@ -222,6 +259,61 @@ impl ImdsCredentialsProvider {
...
+        sleeper: Arc<dyn AsyncSleep>,
+        timeout: Duration,
+    ) -> provider::Result {
+        let sleep_future = sleeper.sleep(timeout);
+        let timeout_future = Timeout::new(self.provide_credentials(), sleep_future);
+        match timeout_future.await {
+            Ok(creds) => creds,
+            _ => match &*self.last_retrieved_credentials.read().await {
+                Some(creds) => Ok(creds.clone()),
+                _ => Err(CredentialsError::provider_timed_out(timeout)),
```
jdisanti, line 284: Seems like this should return the original error rather than a timeout error since this error case can also happen in non-timeout cases (such as the IMDS calls failing).

aws/rust-runtime/aws-config/src/meta/credentials/chain.rs
```diff
@@ -104,4 +136,30 @@ impl ProvideCredentials for CredentialsProviderChain {
...
+    where
+        Self: 'a,
+    {
+        // We need to give `timeout` to the whole chain as well as to each credentials provider in the chain.
+        // One could argue that if the whole chain has `timeout` anyway, there is no point in having it
+        // in individual providers. This is due to the fact that the other method `provide_credentials`
+        // does not respect provider-specific read timeout behavior, e.g. the IMDS credentials provider
+        // wants to provide expired credentials, if any, in the case of read timeout.
+        let sleep_future = sleeper.sleep(timeout);
+        let timeout_future = Timeout::new(
```
jdisanti, line 154: Won't this top-level timeout cause a race condition with the IMDS provider's timeout?

ysaito1001: Yeah, I realized that after pushing the commit. One idea is to update the signature of `provide_credentials_with_timeout` to take a mutable `Sleep`, instead of `sleeper`. `CredentialsProviderChain::provide_credentials_with_timeout` passes it through the chain and each provider only gets remaining timeout from the preceding provider.

Regardless of whether the above idea is implemented as part of the PR, we will work on an RFC because the design around timeout and credentials providers affects not just `ImdsCredentialsProvider`.

ysaito1001: Yeah, I realized that after pushing the commit. One interesting idea is to update the signature of `provide_credentials_with_timeout` to take something like `aws_smithy_async::rt::Sleep` but sharable.
`CredentialsProviderChain::provide_credentials_with_timeout` passes it through the chain like "passing the torch" one provider to the next.

Regardless of whether the above idea is implemented as part of the PR, we will work on an RFC because the design around timeout and credentials providers affects not just `ImdsCredentialsProvider`.

ysaito1001: Yeah, I realized that after pushing the commit. One possible workaround is to update the signature of `provide_credentials_with_timeout` to take something like `aws_smithy_async::rt::Sleep` but sharable.
`CredentialsProviderChain::provide_credentials_with_timeout` passes it through the chain like "passing the torch" one provider to the next.

Regardless of whether the above idea is implemented as part of the PR, we will work on an RFC because the design around timeout and credentials providers affects not just `ImdsCredentialsProvider`.

ysaito1001:
Will close this draft for now. Need to rework on what method to add to the `ProvideCredentials` trait. `provide_credentials_with_timeout` means delegating overall timeout to the individual provider, which can be tricky to get right in the face of races.

PR #2192, opened 2023-01-10
Remove usage of always empty writable in `JsonParserGenerator`

david-perez:
The writable calculates a string that it never writes.

This was added in #2131.

----


hlbarber (approved):

82marbag:
> This was added in https://github.com/awslabs/smithy-rs/pull/2131.

This was the behavior before that commit too. I agree it does not render anything anyway.

jdisanti (approved):

PR #2197, opened 2023-01-11
add rustfmt for cargo fix

thomas-k-cameron:
## Motivation and Context
Stronger consistency in code organization.
`cargo fix` can fix up the import order; Thus, reducing the complexity of code review and improving productivity.

For example, this PR will automatically resolve something like this from even happening.
[example comment](https://github.com/awslabs/smithy-rs/pull/2185#discussion_r1065971687)

## Description
Add cargofmt.toml

## Testing
Ran cargo fix to ensure that it's working.

## Checklist

----


Velfi:
I think we should also run `cargo fix` as part of this PR. Also, we should add it to the pre-commit actions.

Velfi:
@thomas-k-cameron one of my teammates brought up that this is going to result in a lot of code churn so we should start with just the rule for fixing the imports and add the other changes. They also requested that you add this to our CI action `check-style-and-lints`. Since `cargo fix` doesn't have a `--check` mode like `cargo fmt`, you'll have to check a diff to determine if the user forgot to run it.

Do you feel up for that?

thomas-k-cameron:
https://github.com/awslabs/smithy-rs/pull/2197#issuecomment-1379472989

I made a update.
can you check if it works?

Velfi:
> Concerning this one, can't you just cp the source code to another directory and check if everything is the same?

That approach is fine with me. I only care that it's automated.

thomas-k-cameron:
Ok, I seem to have made a mistake;

`cargo fix` doesn't fix the import order; `cargo fmt` is the one that reorder the import order.

`fix` only fixes the lint warning reported by `rustc`.

thomas-k-cameron:
Problem with the `cargo fmt` is that you need nightly to enable fmt's feature for reordering imports.

I don't think the use of nightly channel would cause any trouble if the reordering is the only thing you would do with it.
(You don't want to ship the SDK with nightly right?)

thomas-k-cameron:
I pushed with the changes. Let me know how you think.

thomas-k-cameron:
I think I might want to close this PR because my proposal was based on my misunderstanding.

1. `cargo fix` doesn't fix ordering of module declaration/imports
2. `cargo fmt` can change the ordering of module declaration/imports but not the kind that lead me to open this PR. https://github.com/awslabs/smithy-rs/pull/2185#discussion_r1065971687

I think should've double checked but I think I missed it along the way.  
I'm really sorry for about it.

PR #2204, opened 2023-01-12
Endpoint operation input tests

rcoh:
## Motivation and Context
- #2144 
- #1784 

Endpoints 2.0 exposes integration-style tests–we need to use those (and this found a bug!)

## Description
- enable operationInput tests. I chose to generate tests using the fluent client and capture request—this seemed like the best end-to-end strategy. Because of this, this currently only runs against the fluent client.
- a few other things needed to happen to support this:
  - We needed a way to go from builtIn name to applying it to the service config
  - We needed a way to default config parameters for tests—I added `with_test_defaults()` which we can start using after this PR 

## Testing
- tests pass 

## Checklist

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
```diff
@@ -0,0 +1,179 @@
...
+}
+
+val PromotedBuiltInsDecorators =
+    listOf(
+        decoratorForBuiltIn(Builtins.FIPS),
+        decoratorForBuiltIn(Builtins.DUALSTACK),
+        decoratorForBuiltIn(
+            Builtins.SDK_ENDPOINT,
+            ConfigParam("endpoint_url", RuntimeType.String.toSymbol(), endpointUrlDocs),
+        ),
```
Velfi, line 178: If a customer is using a custom smithy client with SDK codegen disabled, can they easily set endpoints? This looks SDK specific, and I'm wondering how it interacts with the "plain" smithy client codegen.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkConfigDecorator.kt
```diff
@@ -35,6 +35,16 @@ object SdkConfigSection : AdHocSection<SdkConfigSection.CopySdkConfigToClientCon
...
+
+    fun copyField(fieldName: String, map: Writable?) = SdkConfigSection.create { section ->
+        {
+            val mapBlock = map?.let { writable { rust(".map(#W)", it) } } ?: writable { }
+            rustTemplate(
+                "${section.serviceConfigBuilder}.set_$fieldName(${section.sdkConfig}.$fieldName()#{map});",
+                "map" to mapBlock,
+            )
+        }
+    }
```
Velfi, line 47: An example for this `fun` would be appreciated. It looks cool.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
```diff
@@ -79,6 +85,24 @@ class S3Decorator : ClientCodegenDecorator {
         it + S3PubUse()
     }
 
+    override fun endpointCustomizations(codegenContext: ClientCodegenContext): List<EndpointCustomization> {
+        return listOf(object : EndpointCustomization {
+            override fun setBuiltInOnServiceConfig(name: String, value: Node, configBuilderRef: String): Writable? {
+                if (name.startsWith("S3")) {
+                    return null
+                }
```
Velfi, line 93: What kinds of things does this skip? From my naïve point of view, it looks like we're filtering out stuff related to S3 in an S3 decorator, which I find strange.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/AwsEndpointDecorator.kt
```diff
@@ -141,27 +140,13 @@ class AwsEndpointDecorator : ClientCodegenDecorator {
...
-                override fun builtInDefaultValue(parameter: Parameter, configRef: String): Writable? {
-                    return when (parameter.builtIn) {
-                        Builtins.SDK_ENDPOINT.builtIn -> writable { rust("$configRef.endpoint_url().map(|url|url.to_string())") }
-                        else -> null
-                    }
-                }
-            },
-        )
-    }
-
```
Velfi, line 164: This is no longer needed because it's being added by `PromotedBuiltInsDecorators`, right?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+                    generator.generateInput(operationInput)
+                }
+            }
+            tests.join("\n")(this)
+        }
+    }
+}
+
+private val deprecatedBuiltins =
+    setOf(Builtins.S3_USE_GLOBAL_ENDPOINT, Builtins.STS_USE_GLOBAL_ENDPOINT).map { it.builtIn.get() }
```
Velfi, line 60: Is it worth documenting when/why these were deprecated? Is there a source where that info is tracked?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+
+fun usesUnsupportedBuiltIns(testOperationInput: EndpointTestOperationInput): Boolean {
+    return testOperationInput.builtInParams.members.map { it.key.value }.any { deprecatedBuiltins.contains(it) }
+}
```
Velfi, line 64: Other than deprecation, what reasons are there that a built-in would be invalid?
```suggestion
fun usesDeprecatedBuiltIns(testOperationInput: EndpointTestOperationInput): Boolean {
    return testOperationInput.builtInParams.members.map { it.key.value }.any { deprecatedBuiltins.contains(it) }
}
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Eventually, we need to pull this test into generic smithy. However, this relies on generic smithy clients
+ * supporting middleware and being instantiable from config (https://github.com/awslabs/smithy-rs/issues/2194)
+ *
+ * Doing this in AWS codegen allows us to actually integration test generated clients.
+ */
```
Velfi, line 71: I removed the second `*` since I wouldn't classify this as an actual doc comment.
```suggestion
/*
 * TODO(https://github.com/awslabs/smithy-rs/issues/2194) Eventually, we need to pull this test into
 * generic smithy. However, this relies on generic smithy clients supporting middleware and being
 * instantiable from config 
 *
 * Doing this in AWS codegen allows us to actually integration test generated clients.
 */
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+        if (test.expect.endpoint.orNull()?.properties?.get("authSchemes")?.asArrayNode()?.orNull()
+            ?.map { it.expectObjectNode().expectStringMember("name").value }?.contains("sigv4a") == true
+        ) {
+            Attribute.ShouldPanic.render(this)
+        }
+        tokioTest(safeName("operation_input_test_$operationName")) {
+            rustTemplate(
+                """
+                /* builtIns: ${escape(Node.prettyPrintJson(testOperationInput.builtInParams))} */
+                /* clientParams: ${escape(Node.prettyPrintJson(testOperationInput.clientParams))} */
```
Velfi, line 94: Are these lines here to help debug failing tests?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+    private val instantiator = clientInstantiator(ctx)
+
+    private fun EndpointTestOperationInput.operationId() =
+        ShapeId.fromOptionalNamespace(ctx.serviceShape.id.namespace, operationName)
+
+    fun generateInput(testOperationInput: EndpointTestOperationInput) = writable {
+        val operationName = testOperationInput.operationName.toSnakeCase()
+        if (test.expect.endpoint.orNull()?.properties?.get("authSchemes")?.asArrayNode()?.orNull()
+            ?.map { it.expectObjectNode().expectStringMember("name").value }?.contains("sigv4a") == true
+        ) {
```
Velfi, line 87: This is complex enough that we should consider extracting it into its own function and documenting it.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt
```diff
@@ -0,0 +1,172 @@
...
+                            escape("expected error: $error [${test.documentation.orNull() ?: "no docs"}]")
+                        rustTemplate(
+                            """
+                            rcvr.expect_no_request();
+                            let error = _result.expect_err(${expectedError.dq()});
+                            assert!(format!("{:?}", error).contains(${escape(error).dq()}), "expected error to contain `${
+                            escape(
+                                error,
+                            )
+                            }` but it was {}", format!("{:?}", error));
```
Velfi, line 127: Why not just
```suggestion
                            }` but it was {:?}", error));
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/StripEndpointTrait.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.traits.EndpointTrait
+import software.amazon.smithy.model.transform.ModelTransformer
+
+fun stripEndpointTrait(hostPrefix: String): (Model) -> Model {
+    return { model: Model ->
+        ModelTransformer.create()
+            .removeTraitsIf(model) { _, trait ->
+                trait is EndpointTrait && trait.hostPrefix.labels.any {
```
Velfi, line 16: Does this cover both the test trait and the rules trait?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customize/RequiredCustomizations.kt
```diff
@@ -58,6 +60,8 @@ class RequiredCustomizations : ClientCodegenDecorator {
         // Add rt-tokio feature for `ByteStream::from_path`
         rustCrate.mergeFeature(Feature("rt-tokio", true, listOf("aws-smithy-http/rt-tokio")))
 
+        rustCrate.mergeFeature(TestUtilFeature)
```
Velfi, line 63: Checking I understand this:

We have to manually add this feature here so that it gets added to the generated `Cargo.toml`. This is because we don't have magical support for when stuff is feature-gated _(as opposed to how we add deps to a `Cargo.toml`.)_

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
```diff
@@ -43,18 +43,28 @@ interface EndpointCustomization {
...
      */
-    fun customRuntimeFunctions(codegenContext: ClientCodegenContext): List<CustomRuntimeFunction> = listOf()
+    fun setBuiltInOnServiceConfig(name: String, value: Node, configBuilderRef: String): Writable? = null
 
     /**
-     * Set a given builtIn value on the service config builder. If this builtIn is not recognized, return null
+     * Provide a list of additional endpoints standard library functions that rules can use
      */
-    fun setBuiltInOnConfig(name: String, value: Node, configBuilderRef: String): Writable? = null
+    fun customRuntimeFunctions(codegenContext: ClientCodegenContext): List<CustomRuntimeFunction> = listOf()
```
Velfi, line 67: I appreciate the example on `loadBuiltInFromServiceConfig`. Adding examples for these two would be just as appreciated.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointTypesGenerator.kt
```diff
@@ -22,9 +22,9 @@ import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
  * This exposes [RuntimeType]s for the individual components of endpoints 2.0
  */
 class EndpointTypesGenerator(
-    codegenContext: ClientCodegenContext,
+    private val codegenContext: ClientCodegenContext,
     private val rules: EndpointRuleSet?,
-    private val tests: List<EndpointTestCase>,
+    val tests: List<EndpointTestCase>,
```
Velfi, line 27: This is b/c it's used by `OperationInputTestGenerator.kt`, right?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
```diff
@@ -81,12 +89,72 @@ sealed class ServiceConfig(name: String) : Section(name) {
...
+     */
+    data class DefaultForTests(val configBuilderRef: String) : ServiceConfig("DefaultForTests")
+}
+
+data class ConfigParam(val name: String, val type: Symbol, val setterDocs: Writable?, val getterDocs: Writable? = null)
+
+/**
+ * Config customization for a standard config param:
+ * 1. `pub(crate)` field
+ * 2. pub(crate) accessor
```
Velfi, line 104: It feels weird to backtick one but not the other
```suggestion
 * 1. `pub(crate)` field
 * 2. `pub(crate)` accessor
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
```diff
@@ -81,12 +89,72 @@ sealed class ServiceConfig(name: String) : Section(name) {
...
+    /**
+     * The set default value of a field for use in tests, e.g `${configBuilderRef}.set_credentials(Credentials::for_tests())`
+     */
+    data class DefaultForTests(val configBuilderRef: String) : ServiceConfig("DefaultForTests")
+}
+
+data class ConfigParam(val name: String, val type: Symbol, val setterDocs: Writable?, val getterDocs: Writable? = null)
+
+/**
+ * Config customization for a standard config param:
```
Velfi, line 102: What does "standard" mean in this case?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
```diff
@@ -168,6 +239,25 @@ class ServiceConfigGenerator(private val customizations: List<ConfigCustomizatio
             customizations.forEach {
                 it.section(ServiceConfig.BuilderImpl)(this)
             }
+
+            val testUtilOnly =
+                Attribute(Attribute.cfg(Attribute.any(Attribute.feature(TestUtilFeature.name), writable("test"))))
```
Velfi, line 244: This renders `#[cfg(any(feature = "test-util", test))]`, right?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustType.kt
```diff
@@ -466,6 +466,7 @@ class Attribute(val inner: Writable) {
         val DenyMissingDocs = Attribute(deny("missing_docs"))
         val DocHidden = Attribute(doc("hidden"))
         val DocInline = Attribute(doc("inline"))
+        val ShouldPanic = Attribute("should_panic")
```
Velfi, line 469: IMO, we should also support the "message to match against" version and encourage its usage:
```suggestion
        val ShouldPanic = Attribute("should_panic")
        fun shouldPanic(messageToMatchAgainst: String): Writable = macroWithArgs("should_panic", writable(str))
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -92,6 +93,8 @@ open class Instantiator(
         fun doesSetterTakeInOption(memberShape: MemberShape): Boolean
     }
 
+    fun generate(shape: Shape, data: Node, ctx: Ctx = Ctx()) = writable { render(this, shape, data, ctx) }
```
Velfi, line 96: Given that neither this method nor `render` reference their class, I'm wondering why they're even methods. Should this functionality be moved elsewhere?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/LetIf.kt
```diff
@@ -24,3 +24,5 @@ fun <T> Boolean.thenSingletonListOf(f: () -> T): List<T> = if (this) {
 } else {
     listOf()
 }
+
+fun<T> List<T>.nonEmptyOrNull(): List<T>? = this.ifEmpty { null }
```
Velfi, line 28: consider adding an example. Also, more bikeshedding, what about `orNullIfEmpty`?

Velfi (approved):
I only had doc and naming-related comments. Good work on this.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
rcoh: "regular"—a field, a setter, a getter and the same type for all

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
rcoh: yup

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
rcoh: they do eventually (e.g. they need symbolProvider)

rcoh (commented):
thanks for all the excellent comments!

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialProviders.kt
```diff
@@ -42,36 +55,59 @@ class CredentialsProviderDecorator : ClientCodegenDecorator {
         listOf(
             SdkConfigSection.create { section ->
                 writable {
-                    rust("${section.serviceConfigBuilder}.set_credentials_provider(${section.sdkConfig}.credentials_provider().clone());")
+                    rust("${section.serviceConfigBuilder}.set_credentials_provider(${section.sdkConfig}.credentials_provider().cloned());")
```
Velfi: `clone` vs `cloned` always gets me. I wish they had given it a more distinct name.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
```diff
@@ -0,0 +1,179 @@
...
+)
+
+fun Model.loadBuiltIn(serviceId: ShapeId, builtInSrc: Parameter): Parameter? {
+    val model = this
+    val idx = EndpointRulesetIndex.of(model)
+    val service = model.expectShape(serviceId, ServiceShape::class.java)
+    val rules = idx.endpointRulesForService(service) ?: return null
+    // load the builtIn with a matching name from the ruleset allowing for any docs updates
+    return rules.getBuiltIn(builtInSrc)
+}
```
Velfi, line 68: Since `fun EndpointRuleSet.getBuiltIn(builtIn: Parameter) = getBuiltIn(builtIn.builtIn.orNull()!!)` has the non-null assertion, then the return type here can be `Parameter` too.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
```diff
@@ -0,0 +1,179 @@
...
+                        builtIn.builtIn -> writable {
+                            rust("$configRef.$name")
+                            if (parameter.type == ParameterType.STRING) {
+                                rust(".clone()")
+                            }
+                        }
+                        else -> null
+                    }
+
+                override fun setBuiltInOnServiceConfig(name: String, value: Node, configBuilderRef: String): Writable? {
```
Velfi, line 132: We should consider thinking of a word that means "this function constructs and returns a writable" and using it to name things like this. My reasoning is that this looks like something that sets a field in Kotlin when it actually creates rust code that sets something in Rust. I don't think we have too much of a problem with keeping Kotlin vs. Rust stuff straight in codegen, but I think it's worth considering and defining a boundary.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt
```diff
@@ -0,0 +1,218 @@
...
+ *     let error = _result.expect_err("expected error: Invalid configuration: region from ARN `us-east-1` does not match client region `us-west-2` and UseArnRegion is `false` [outposts arn with region mismatch and UseArnRegion=false]");
+ *     assert!(format!("{:?}", error).contains("Invalid configuration: region from ARN `us-east-1` does not match client region `us-west-2` and UseArnRegion is `false`"), "expected error to contain `Invalid configuration: region from ARN `us-east-1` does not match client region `us-west-2` and UseArnRegion is `false`` but it was {}", format!("{:?}", error));
+ * }
+ * ```
+ *
+ * Eventually, we need to pull this test into generic smithy. However, this relies on generic smithy clients
+ * supporting middleware and being instantiable from config (https://github.com/awslabs/smithy-rs/issues/2194)
+ *
+ * Doing this in AWS codegen allows us to actually integration test generated clients.
+ */
```
Velfi, line 114: ❤️

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
rcoh: I don't think so? this can return null because a model may not have a built in

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
```diff
@@ -0,0 +1,179 @@
...
+fun ClientCodegenContext.getBuiltIn(builtIn: Parameter): Parameter? = getBuiltIn(builtIn.builtIn.orNull()!!)
+fun ClientCodegenContext.getBuiltIn(builtIn: String): Parameter? {
+    val idx = EndpointRulesetIndex.of(model)
+    val rules = idx.endpointRulesForService(serviceShape) ?: return null
+    return rules.getBuiltIn(builtIn)
+}
+
+private fun toConfigParam(parameter: Parameter): ConfigParam = ConfigParam(
+    parameter.name.rustName(),
+    when (parameter.type!!) {
```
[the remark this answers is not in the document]
rcoh: this is just the kotlin->java boundary

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
```diff
@@ -0,0 +1,179 @@
...
+import software.amazon.smithy.rust.codegen.core.util.extendIf
+import software.amazon.smithy.rust.codegen.core.util.orNull
+import java.util.Optional
+
+/** load a builtIn parameter from a ruleset by name */
+fun EndpointRuleSet.getBuiltIn(builtIn: String) = parameters.toList().find { it.builtIn == Optional.of(builtIn) }
+
+/** load a builtIn parameter from a ruleset. The returned builtIn is the one defined in the ruleset (including latest docs, etc.) */
+fun EndpointRuleSet.getBuiltIn(builtIn: Parameter) = getBuiltIn(builtIn.builtIn.orNull()!!)
+fun ClientCodegenContext.getBuiltIn(builtIn: Parameter): Parameter? = getBuiltIn(builtIn.builtIn.orNull()!!)
```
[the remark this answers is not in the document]
rcoh: the `!!` is inside the paren

Velfi (approved):

PR #2213, opened 2023-01-13
Support `@sparse` constrained map shapes and list shapes

david-perez:
Turns out we've never supported them, neither directly constrained nor
with constrained members, because of a lack of tests. Yet another data
point to prioritize working on code-generating `constraints.smithy` (see
https://github.com/awslabs/smithy-rs/issues/2101).

The implementation is simple: we just need to call the symbol provider
on the member symbols instead of on the target symbols so we get
`Option<T>` list members / map values if applicable, and handle the
wrapper when converting between unconstrained and constrained types with
help from `match` and `Option<T>::map`.

----

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedMapGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedMapGenerator.kt
@@ -14,7 +14,10 @@ import software.amazon.smithy.rust.codegen.core.rustlang.join
 import software.amazon.smithy.rust.codegen.core.rustlang.rust
 import software.amazon.smithy.rust.codegen.core.rustlang.rustBlock
 import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.core.rustlang.withBlock
 import software.amazon.smithy.rust.codegen.core.rustlang.writable
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.core.smithy.isOptional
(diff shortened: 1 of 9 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedMapGenerator.kt
```diff
@@ -102,31 +105,57 @@ class UnconstrainedMapGenerator(
...
+                                    // DRYing this up with the else branch below would make this less understandable.
+                                    rustTemplate(
+                                        """
+                                        match #{ConstrainedValueSymbol}::try_from(v) {
+                                            Ok(v) => Ok((k, Some(v))),
+                                            Err(inner_constraint_violation) => Err(Self::Error::Value(k, inner_constraint_violation)),
+                                        }
+                                        """,
+                                        "ConstrainedValueSymbol" to constrainedValueSymbol,
+                                        "Epilogue" to epilogueWritable,
```
crisidev, line 142:
```suggestion
```

crisidev: This doesn't look to be used here in this block

crisidev (approved):
Just a small comment on a value that looks to be unused. LGTM.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedMapGenerator.kt
david-perez: Fixed in https://github.com/awslabs/smithy-rs/pull/2213/commits/b4ae60f85991e647748c2bc1a1bce247761ca1f3.

crisidev (approved):

PR #2218, opened 2023-01-17
Add RFC: Providing fallback credentials on timeout

ysaito1001:
## Motivation and Context
This PR adds an RFC to describe how credentials providers can serve fallback credentials in the event of timeout (which the RFC refers to as external timeout). This lays the groundwork for a static stability support [PR](https://github.com/awslabs/smithy-rs/pull/2191); we will reopen it once the RFC has been approved and implemented.

[Rendered](https://github.com/awslabs/smithy-rs/blob/16a14ba0957d5e86ebcabaa733c908caf653954f/design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md)

----


design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+    pub struct OnTimeout<'a>(NowOrLater<Option<Credentials>, BoxFuture<'a, Option<Credentials>>>);
+
+    // impls for OnTimeout similar to those for the ProvideCredentials future newtype
+}
+
+pub trait ProvideCredentials: Send + Sync + std::fmt::Debug {
+    // --snip--
+
+    fn on_timeout<'a>(&'a self) -> future::OnTimeout<'a> {
```
jdisanti, line 91: Is there any reason this shouldn't be synchronous? What's the use-case for asynchronously retrieving the credentials on timeout? Would we have a timeout for the `on_timeout`?

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+The user experience for the code snippet in question will look like this once this proposal is implemented:
+```rust
+let timeout_future = self.sleeper.sleep(self.load_timeout); // by default self.load_timeout is 5 seconds.
+// --snip--
+let future = Timeout::new(provider.provide_credentials(), timeout_future);
+let result = cache
+    .get_or_load(|| {
+        async move {
+           let credentials= match future.await {
```
jdisanti, line 106:
```suggestion
           let credentials = match future.await {
```

jdisanti (commented):
Very nicely written! Just a question on the `on_timeout` return value.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
+RFC: Providing fallback credentials on external timeout
+=======================================================
+
+> Status: RFC
+>
+> Applies to: client
+
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+This RFC proposes a fallback mechanism for credentials providers on external timeout (see the [Terminology](#terminology) section), allowing them to continue serving (possibly stale) credentials for the sake of overall reliability of the intended service; the IMDS credentials provider is an example that must fulfill such a requirement to support static stability.
```
Velfi, line 10: As someone who is an unrepentant semicolon addict, I must note that the first word after a semicolon should be capitalized 😝 .
```suggestion
This RFC proposes a fallback mechanism for credentials providers on external timeout (see the [Terminology](#terminology) section), allowing them to continue serving (possibly stale) credentials for the sake of overall reliability of the intended service; The IMDS credentials provider is an example that must fulfill such a requirement to support static stability.
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+Terminology
+-----------
+
+- External timeout: A timeout within which a call to the `provide_credentials` method must complete. `provide_credentials` returns a future and if it does not complete within this timeout, it gets dropped on the floor and there is no output retrieved from the future.
```
Velfi, line 15: This sentence felt a bit wordy to me so I tried my hand at re-wording it. Let me know if you think it makes anything clearer.

```suggestion
- External timeout: The name of the timeout that occurs when a duration elapses before an async call to `provide_credentials` returns. In this case, `provide_credentials` returns no credentials.
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+- Internal timeout: A timeout imposed on functions/methods internally called by `provide_credentials` that, when failing to meet any of them, renders `provide_credentials` return an `CredentialsError`. For instance, internal timeout includes connection timeout, TLS negotiation timeout, and HTTP request timeout, if `provide_credentials` needs to fetch credentials via HTTP(S). This is a kind of timeout `provide_credentials` can react and handle it accordingly, i.e. returning fallback credentials or an `CredentialsError`.
```
Velfi, line 16: I had a go at this one too
```suggestion
- Internal timeout: The name of the timeout that occurs when a duration elapses before an async call to some function, inside the implementation of `provide_credentials`, returns. Examples include connection timeouts, TLS negotiation timeouts, and HTTP request timeouts. Implementations of `provide_credentials` may handle these failures at their own discretion e.g. by returning _(possibly expired)_ credentials or a `CredentialsError`.
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+Assumption
+----------
+
+This RFC is concerned only with external timeout, as the cost of designing APIs poorly is much higher than that for internal timeout. The former will affect a public trait implemented by all credentials providers whereas the latter can be handled locally by individual credentials providers without affecting one another.
```
Velfi, line 21:
```suggestion
This RFC is concerned only with external timeouts, as the cost of poor API design is much higher than in this case than for internal timeouts. The former will affect a public trait implemented by all credentials providers whereas the latter can be handled locally by individual credentials providers without affecting one another.
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+Problem
+-------
+
+We have mentioned static stability. Supporting it calls for the following functional requirement, among others:
```
Velfi, line 26: "static stability" should be included in the terminology section

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+- REQ 1: Once a credentials provider has served credentials, it should continue serving them in the event of a timeout (whether internal or external) while obtaining refreshed credentials.
+
+Today, we have the following trait method to obtain credentials:
+```rust
+fn provide_credentials<'a>(&'a self) -> future::ProvideCredentials<'a>
+where
+    Self: 'a,
+```
+This method can be used in async contexts, which means that nothing stops it from being raced against a timeout future, as demonstrated by the following code snippet from `LazyCredentialsCache`:
```
Velfi, line 35:
```suggestion
This method returns a future, which can be raced against a timeout future as demonstrated by the following code snippet from `LazyCredentialsCache`:
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+let future = Timeout::new(provider.provide_credentials(), timeout_future);
+let result = cache
+   .get_or_load(|| {
+        async move {
+            let credentials = future.await.map_err(|_err| {
+                CredentialsError::provider_timed_out(load_timeout)
+            })??;
+            // --snip--
+        }
+    }).await;
```
Velfi, line 49: I think this shorter version is also valid
```suggestion
   .get_or_load(|| async move {
        let credentials = future.await.map_err(|_err| {
            CredentialsError::provider_timed_out(load_timeout)
        })??;
        // --snip--
    }).await;
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+**Case 2:** Provider 2 successfully loaded credentials but later was not reached because its preceding provider was still running when an external timeout kicked in.
+
+<p align="center">
+<img width="750" alt="chain-provider-ext-timeout-2" src="https://user-images.githubusercontent.com/15333866/212421712-8c6eab11-a0c1-4229-8ba3-67b0bb6056e7.png">
+</p>
+
+The figure above illustrates an example with the same setting as the previous figure. Again, when `CredentialsProviderChain::provide_credentials` is called the first time, provider 1 does not find credentials but provider 2 does. The next time the method is called, provider 1 is still executing `provide_credentials` and then an external time by `timeout_future` kicked in. Consequently, the execution of `CredentialsProviderChain::provide_credentials` has been terminated. Given the functional requirements, provider 2 should return the previously available credentials but today the code snippet from `LazyCredentialsCache` returns `CredentialsError::ProviderTimedOut` instead.
```
Velfi, line 73:
```suggestion
The figure above illustrates an example with the same setting as the previous figure. Again, when `CredentialsProviderChain::provide_credentials` is called the first time, provider 1 does not find credentials but provider 2 does. The next time the method is called, provider 1 is still executing `provide_credentials` and then an external timeout by `timeout_future` kicked in. Consequently, the execution of `CredentialsProviderChain::provide_credentials` has been terminated. Given the functional requirements, provider 2 should return the previously available credentials but today the code snippet from `LazyCredentialsCache` returns `CredentialsError::ProviderTimedOut` instead.
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+
+Proposal
+--------
+To address the problem in the previous section, we propose to add a new method to the `ProvideCredentials` trait called `on_timeout` that looks something like this:
+```rust
+mod future {
+    // --snip--
+
+    pub struct OnTimeout<'a>(NowOrLater<Option<Credentials>, BoxFuture<'a, Option<Credentials>>>);
```
Velfi, line 83: related, possibly useful: https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.now_or_never

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+// --snip--
+```
+
+
+How to actually implement this RFC
+----------------------------------
+
+Almost all credentials providers do not have to implement their own `on_timeout` except for `CredentialsProviderChain` (`ImdsCredentialsProvider` also needs to implement `on_timeout` when we are adding static stability support to it but that is outside the scope of this RFC).
+
+Considering the two cases we analyzed above, implementing `CredentialsProviderChain::on_timeout` is not so straightforward. Keeping track of whose turn in the chain it is to call `provide_credentials` when an external timeout has ocurred is a challening task. Even if we figured it out, that would still not satisfy `Case 2` above, because it was provider 1 that was actively running when the external timeout kicked in, but the chain should return credentials from provider 2, not from provider 1.
```
Velfi, line 125:
```suggestion
Considering the two cases we analyzed above, implementing `CredentialsProviderChain::on_timeout` is not so straightforward. Keeping track of whose turn in the chain it is to call `provide_credentials` when an external timeout has occurred is a challenging task. Even if we figured it out, that would still not satisfy `Case 2` above, because it was provider 1 that was actively running when the external timeout kicked in, but the chain should return credentials from provider 2, not from provider 1.
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+With all that taken into account, we can consider starting with the following simple approach:
```
Velfi, line 127:
```suggestion
With that in mind, consider instead the following simple approach:
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+                    creds @ Some(_) => return creds,
+                    None => {}
+                }
+            }
+            None
+        })
+    }
+}
+```
+`CredentialsProviderChain::on_timeout` will invoke each provider's `on_timeout` method until credentials are returned by one of them. It ensures that the updated code snippet for `LazyCredentialsCache` can return credentials from provider 2 in both `Case 1` and `Case 2`. Even if `timeout_future` wins the race, the execution subsequently calls `provider.on_timeout().await` to obtain fallback credentials from provider 2, assuming provider 2's `on_timeout` is implemented to return credentials retrieved on the first call to `CredentialsProviderChain::provide_credentials`.
```
Velfi, line 145: > the execution subsequently calls `provider.on_timeout().await` to obtain fallback
this is a possible place to use `now_or_never`

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+
+The downside of this simple approach is that the behavior is not clear if more than one credentials provider in the chain can return credentials from their `on_timeout`. Note, however, that it is the exception rather than the norm for a provider's `on_timeout` to return fallback credentials, at least at the time of writing (01/13/2023). The fact that it returns fallback credentials means that the provider successfully loaded credentials at least once, and it usually continues serving credentials on subsequent calls to `provide_credentials`.
+
+Should we have more than one provider in the chain that can potentially return fallback credentials from `on_timeout`, we could configure the behavior of `CredentialsProviderChain` managing in what order and how each `on_timeout` should be executed; it can be possible to introduce a new API without breaking changes. That being said, we first need to investigate and understand the use case behind it to design the right API.
+
+Alternative
+-----------
+
+In this section, we will describe our alternative approach that we ended up dismissing as unworkable.
```
Velfi, line 154:
```suggestion
In this section, we will describe an alternative approach that we ended up dismissing as unworkable.
```

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
```diff
@@ -0,0 +1,251 @@
...
+```
+There are mainly two problems with this approach. The first problem is that as shown above, there is no sensible way to calculate a timeout for each provider in the chain. The second problem is that exposing a parameter like `timeout` at a public trait's level is giving too much control to users; delegating overall timeout to the individual provider means each provider has to get it right.
+
+Changes checklist
+-----------------
+
+- [ ] Add `on_timeout` method to the `ProvideCredentials` trait with the default implementation
+- [ ] Implement the `OnTimeout` future newtype
+- [ ] Implement `CredentialsProviderChain::on_timeout`
+- [ ] Add unit tests for `Case 1` and `Case 2`
```
Velfi, line 251: When new credentials come through, we should test that all layer's cached credentials are refreshed.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
ysaito1001: No reason it shouldn't. I think each has pros/cons:
- synchronous
**pros**: Easier to reason about. 
**cons**: Will be a breaking change if we want to switch to async in the future.
- asynchronous
**pros**: It's async from the beginning, so less likely to introduce a breaking change.
**cons**: Exactly what you brought up, may need to specify a timeout.

Currently, there are no concrete use cases requiring this to be async.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
ysaito1001: Hmm, it may help obviate the need for a potential timeout on `on_timeout` as [John asked](https://github.com/awslabs/smithy-rs/pull/2218#discussion_r1072657848). We can have `on_method` as `async` and yet sort of behave synchronously; It returns `Some` from `Poll::Ready` but `None` from `Poll::Pending`.

ysaito1001: Interesting. It may help obviate the need for a potential timeout on `on_timeout` as [John asked](https://github.com/awslabs/smithy-rs/pull/2218#discussion_r1072657848). We can have `on_method` as `async` and yet sort of behave synchronously; The future will not wonder off and get polled indefinitely.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
ysaito1001: An async primitive is more suitable as discussed (since we do not have an async critical section). The method is a fallback plan so the execution should not take too long. d413e58 mimics `FutureExt::now_or_never` using our `NowOrLater` with `OnlyReady`, which means no need to have yet another timeout for the method.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
ysaito1001: A good suggestion. Addressed in d413e58.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
ysaito1001: Addressed in 413e58.

ysaito1001: Addressed in d413e58.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
ysaito1001: Thanks for coming up with the use case! As discussed, we can employ a different approach. See the [Appendix](https://github.com/awslabs/smithy-rs/blob/d413e58e57dc0fd7bcd68583414bb7aa250c684c/design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md#appendix) section.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
ysaito1001: I believe we discussed this offline? The execution does need to call `fallback_on_interrupt` to obtain fallback credentials because the future for `provide_credentials` gets dropped and we cannot do anything with it. The concept of `now_or_never` does help in implementing the future returned by `fallback_on_interrupt`.

design/src/rfcs/rfc0031_providing_fallback_credentials_on_timeout.md
ysaito1001: > Would we have a timeout for the on_timeout?

It is possible, although I don't think we specify that as an input argument to `fallback_on_interrupt` for the same reason I described in the Alternative section in the RFC. Whoever uses `fallback_on_timeout` can race it against a timeout future just like `LazyCredentialsCache` does with `provide_credentials` today. It just that the timeout for `fallback_on_timeout` will probably way shorter than that for `provide_credentials`, as the execution of `fallback_on_timeout` shouldn't take too long.

Haven't decided whether the timeout for `fallback_on_timeout` should be introduced today (something like `load_timeout` for `LazyCredentialsCache`). If we are to introduce it later, we need to see if we can do so without causing breaking changes.

ysaito1001: > Would we have a timeout for the on_timeout?

It is possible, although I don't think we specify that as an input argument to `fallback_on_interrupt` for the same reason I described in the Alternative section in the RFC. Whoever uses `fallback_on_timeout` can race it against a timeout future just like `LazyCredentialsCache` does with `provide_credentials` today. It just that the timeout for `fallback_on_timeout` will probably be way shorter than that for `provide_credentials`, as the execution of `fallback_on_timeout` shouldn't take too long.

Haven't decided whether the timeout for `fallback_on_timeout` should be introduced today (something like `load_timeout` for `LazyCredentialsCache`). If we are to introduce it later, we need to see if we can do so without causing breaking changes.

jdisanti (approved):
LGTM!

PR #2223, opened 2023-01-18
Derive `Eq` and `Hash` wherever possible

david-perez:
In server SDKs, these traits can be implemented by any shape _except_ if
the shape's closure contains:

1. A `float`, `double`, or `document` shape: floating point types in
   Rust do not implement `Eq`. Similarly, [`document` shapes] may
   contain arbitrary JSON-like data containing floating point values.
2. A [@streaming] shape: all the streaming data would need to be
   buffered first to compare it.

Additionally, the `Hash` trait cannot be implemented by shapes whose
closure contains:

1. A `map` shape: we render `map` shapes as `std::collections::HashMap`,
   which _do not_ implement `Hash`. See
   https://github.com/awslabs/smithy/issues/1567.

In **client SDKs, these traits cannot be derived on any code-generated
Rust types corresponding to Smithy shapes**, since e.g. adding new
optional members to a structure [is a backwards-compatible change], and
doing so alters the semantics of these traits.

However, this commit does implement these traits for the
`aws_smithy_types::date_time::DateTime` and `aws_smithy_types::Blob`
runtime types.

This change is necessary to efficiently implement the `@uniqueItems`
constraint trait in server SDKs.

[`Eq`]: https://doc.rust-lang.org/std/cmp/trait.Eq.html
[`Hash`]: https://doc.rust-lang.org/std/hash/trait.Hash.html
[`document` shapes]: https://smithy.io/2.0/spec/simple-types.html#document
[@streaming]: https://smithy.io/2.0/spec/streaming.html
[is a backwards-compatible change]: https://smithy.io/2.0/guides/evolving-models.html#updating-structures

## Testing

I added tests.

## Checklist

----


david-perez:
I'll try to pull out the changes to constrained types in a separate PR so that this patch is more easily reviewable by the @awslabs/rust-sdk-owners team.

david-perez:
Actually, I can't break this PR down any further:

* If I remove `Eq`, `Hash` from `aws_smithy_types::date_time::DateTime` or `aws_smithy_types::Blob`, any server shape with these types in its closure cannot implement `Eq` or `Hash`.
* If I pull out the constrained shape symbol metadata provider, I have to derive `Eq` and `Hash` for constrained types, which may have `aws_smithy_types::date_time::DateTime` or `aws_smithy_types::Blob` in its closure.

So this patch has to be merged atomically. I've updated the changelog and the commit message to better reflect all changes.

---

@awslabs/rust-sdk-owners team: you just have to review the changes to these files:
* `rust-runtime/aws-smithy-types/src/date_time/mod.rs`
* `rust-runtime/aws-smithy-types/src/lib.rs`
* `CHANGELOG.next.toml`

@awslabs/smithy-rs-server team: you have to review everything.

rust-runtime/aws-smithy-types/src/date_time/mod.rs
```diff
@@ -47,7 +47,7 @@ const NANOS_PER_SECOND_U32: u32 = 1_000_000_000;
 /// The [`aws-smithy-types-convert`](https://crates.io/crates/aws-smithy-types-convert) crate
 /// can be used for conversions to/from other libraries, such as
 /// [`time`](https://crates.io/crates/time) or [`chrono`](https://crates.io/crates/chrono).
-#[derive(Debug, PartialEq, Clone, Copy)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
```
crisidev, line 50: If I am not mistaken (@unexge can comment on this), we need to derive the same also on the Python wrappers that are found inside `aws-smithy-http-server-python/src/types.rs`.

rust-runtime/aws-smithy-types/src/lib.rs
```diff
@@ -30,7 +30,7 @@ pub use error::Error;
 /// Binary Blob Type
 ///
 /// Blobs represent protocol-agnostic binary content.
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
```
crisidev, line 33: If I am not mistaken (@unexge can comment on this), we need to derive the same also on the Python wrappers that are found inside `aws-smithy-http-server-python/src/types.rs`.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerSymbolProvider.kt
```diff
@@ -106,7 +105,12 @@ class PythonStreamingShapeMetadataProvider(private val base: RustSymbolProvider,
...
-        return base.toSymbol(stringShape).expectRustMetadata()
-    }
+    override fun memberMeta(memberShape: MemberShape) = base.toSymbol(memberShape).expectRustMetadata()
+    override fun enumMeta(stringShape: StringShape) = base.toSymbol(stringShape).expectRustMetadata()
+
+    override fun listMeta(listShape: ListShape) = base.toSymbol(listShape).expectRustMetadata()
+    override fun mapMeta(mapShape: MapShape) = base.toSymbol(mapShape).expectRustMetadata()
+    override fun stringMeta(stringShape: StringShape) = base.toSymbol(stringShape).expectRustMetadata()
+    override fun numberMeta(numberShape: NumberShape) = base.toSymbol(numberShape).expectRustMetadata()
+    override fun blobMeta(blobShape: BlobShape) = base.toSymbol(blobShape).expectRustMetadata()
```
crisidev, line 115: Wouldn't be possible to add `DeriveEqAndHashSymbolMetadataProvider` to the Python symbol visitor to keep the codebase less diverse between the two languages?

crisidev: The same goes for `ConstrainedShapeSymbolMetadataProvider`. Shouldn't we add also that to Python?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedMapGenerator.kt
```diff
@@ -149,6 +154,7 @@ class ConstrainedMapGenerator(
                     type Unconstrained = #{UnconstrainedSymbol};
                 }
                 """,
+                *codegenScope,
```
crisidev, line 157: Maybe is an artefact of the diff, but are you sure you need `codegenScope` here?

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/DeriveEqAndHashSymbolMetadataProviderTest.kt
```diff
@@ -0,0 +1,237 @@
...
+
+            return (
+                shapesWithNeitherEqNorHash.map { Arguments.of(it, emptyList<RuntimeType>()) } +
+                    shapesWithEqAndHash.map { Arguments.of(it, listOf(RuntimeType.Eq, RuntimeType.Hash)) } +
+                    shapesWithOnlyEq.map { Arguments.of(it, listOf(RuntimeType.Eq)) }
+                ).stream()
+        }
+    }
+
+    @ParameterizedTest(name = "(#{index}) Derive `Eq` and `Hash` when possible. Params = shape: {0}, expectedTraits: {1}")
```
crisidev, line 215: TIL how to do parametrized tests in kotlin. Nice.

david-perez:
@crisidev @unexge All right I added `Eq` and `Hash` to `aws_smithy_http_server_python::types::Blob, DateTime}`, and added `DeriveEqAndHashSymbolMetadataProvider` and `ConstrainedShapeSymbolMetadataProvider` to the Python symbol provider in `PythonCodegenServerPlugin`, in https://github.com/awslabs/smithy-rs/pull/2223/commits/f1f0b3ded359b2dbd1a6896ed5e1135486e73186.

rcoh (approved):
lgtm re: eq/hash derives

crisidev (approved):
LGTM. I'd wait for @unexge review as well on this.

PR #2227, opened 2023-01-18
Replace Arc with SharedCredentialsProvider in credentials_provider builder methods

ysaito1001:
## Motivation and Context
#2122 made an API change to expose `std::sync::Arc` to our customers when they called `credentials_provider` builder methods on `ConfigLoader`, `sdk_config::Builder`, or service specific client builders. This PR reverts the change as exposing the implementation detail like `std::sync::Arc` in public APIs makes it difficult for us to evolve them in the future.

The customers can continue passing owned `impl ProvideCredentials` as well as `SharedCredentialsProvider` without `Arc` to those builder methods like prior to the said PR.

## Description
[RFC 28](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0028_sdk_credential_cache_type_safety.md) proposed renaming `SharedCredentialsProvider` to `SharedCredentialsCache`. As stated above, however, we'll resurrect `SharedCredentialsProvider` (which now coexists with `SharedCredentialsCache`) and update call sites, examples, and the release note accordingly.

## Testing
- [x] Passed tests in CI

## Checklist

----


aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -326,22 +326,21 @@ mod loader {
         /// Override the credentials provider but load the default value for region:
         /// ```no_run
         /// # use aws_credential_types::Credentials;
-        /// # use std::sync::Arc;
         /// # fn create_my_credential_provider() -> Credentials {
```
Velfi, line 329: Should this return a `SharedCredentialsProvider`?

Velfi (approved):

rcoh (approved):
lgtm!

PR #2228, opened 2023-01-19
Remove old entries from CHANGELOG.next.toml

LukeMathWalker:
These were already released in 0.53.

----


crisidev (approved):
I skimmed through them and I think you caught them all.

PR #2230, opened 2023-01-19
AwsJson1.0 `@streaming` unique in payload

82marbag:
## Motivation and Context

In RestJson, a streaming member is the unique member in the payload and must be marked with `@httpPayload`. 

Until we have an answer on https://github.com/awslabs/smithy/issues/1584, we should restrict protocols that don't support http traits and have data in payloads to unique streaming member in the payload.

The change is backwards compatible as no release supports `@streaming` in AwsJson1.0 and relaxing this change in the future will not be breaking.

## Checklist

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/AwsJson.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/AwsJson.kt
@@ -5,6 +5,7 @@
 
 package software.amazon.smithy.rust.codegen.core.smithy.protocols
 
+import software.amazon.smithy.codegen.core.CodegenException
 import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.pattern.UriPattern
 import software.amazon.smithy.model.shapes.MemberShape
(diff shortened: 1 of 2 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/AwsJson.kt
```diff
@@ -49,16 +50,21 @@ class AwsJsonHttpBindingResolver(
...
-                if (it.isStreaming(model)) {
-                    HttpBindingDescriptor(it, HttpLocation.PAYLOAD, "document")
-                } else {
-                    HttpBindingDescriptor(it, HttpLocation.DOCUMENT, "document")
-                }
+    private fun bindings(shape: ToShapeId): List<HttpBindingDescriptor> {
+        val members = shape.let { model.expectShape(it.toShapeId()) }.members()
+        if (members.size > 1 && members.any { it.isStreaming(model) }) {
+            throw CodegenException("A streaming members must be the only member in the payload of the request")
+        }
```
crisidev, line 57: I think it's worth to replicate the description of this CR as a comment here.

crisidev (approved):

PR #2232, opened 2023-01-19
Add support for `@uniqueItems`

david-perez:
This commit adds support for the `@uniqueItems` trait on `list` shapes
in server SDKs. Requests with duplicate values for `list` shapes
constrained with `@uniqueItems` will be rejected by servers.

## Testing

Lots of it.

## Checklist

----


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedCollectionGenerator.kt
```diff
@@ -180,14 +179,145 @@ class ConstrainedCollectionGenerator(
...
+                            fn check_unique_items(items: #{Vec}<#{MemberSymbol}>) -> Result<#{Vec}<#{MemberSymbol}>, #{ConstraintViolation}> {
+                                let mut seen = #{HashMap}::new();
+                                let mut duplicate_indices = #{Vec}::new();
+                                for (idx, item) in items.iter().enumerate() {
+                                    if let Some(prev_idx) = seen.insert(item, idx) {
+                                        duplicate_indices.push(prev_idx);
+                                    }
+                                }
+
+                                let mut last_duplicate_indices = #{Vec}::new();
```
david-perez, line 287: Is there a way to convince the compiler that I'll be pushing to disjoint slices of `duplicate_indices` in order to avoid allocating this vector here? 🤔

crisidev: Do you want to modify `duplicate_indices` in place, where you give the assurance that you are going to push after the original `len()` of `duplicate_indices`? You could probably do it with unsafe if we uphold the requirements above. Is there a way in safe rust? 🤔

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedCollectionGeneratorTest.kt
```diff
@@ -46,98 +78,177 @@ class ConstrainedCollectionGeneratorTest {
...
-                list ConstrainedList {
-                    member: String
-                }
+        private fun constraintViolationForDuplicateIndices(duplicateIndices: List<Int>):
+            ((constraintViolation: Symbol, originalValueBindingName: String) -> Writable) {
+            fun ret(constraintViolation: Symbol, originalValueBindingName: String): Writable = writable {
+                // Public documentation for the unique items constraint violation states that callers should not
+                // rely on the order of the elements in `duplicate_indices`. However, the algorithm is deterministic,
+                // so we can internally assert the order. If the algorithm changes, the test cases will need to be
+                // adjusted.
```
crisidev, line 100: 👍🏼

crisidev (approved):
Impeccable PR!

PR #2240, opened 2023-01-20
Dense lists cannot deserialize null

82marbag:


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -389,8 +389,24 @@ class JsonParserGenerator(
                                     withBlock("let value =", ";") {
                                         deserializeMember(shape.member)
                                     }
-                                    rustBlock("if let Some(value) = value") {
-                                        rust("items.push(value);")
+                                    codegenTarget.ifServer {
```
david-perez, line 392: nit: we could also do:

```
rust(
    """
    if let Some(value) = value {
        items.push(value);
    }
    """,
)
codegenTarget.ifServer {
    rust(
        """
        else {
            return Err(#{Error}::custom("dense list cannot contain null values"));
        }
        """,
    )
}
```

and avoid the client branch.

david-perez (approved):

PR #2241, opened 2023-01-20
Must set a member in unions

82marbag:
Deserialization should fail for unions without any explicit member

----


Velfi (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -510,23 +514,30 @@ class JsonParserGenerator(
...
+                                                variant = Some(#{Union}::$variantName)
                                                 """,
                                                 "Union" to returnSymbolToParse.symbol, *codegenScope,
                                             )
                                         } else {
-                                            withBlock("Some(#T::$variantName(", "))", returnSymbolToParse.symbol) {
-                                                deserializeMember(member)
+                                            rust("let (value, is_not_null) =")
+                                            deserializeMember(member)
+                                            rust(".map_or((None, false), |v| (Some(v), true));")
```
jdisanti, line 532: Something seems off here. If we got into this variant match arm, then we have an explicitly set variant, right? So it shouldn't matter if it's null or not?

jdisanti: In fact, I think a variant being explicitly set to `null` is an error condition that we could probably short-circuit on.

82marbag: > then we have an explicitly set variant, right? So it shouldn't matter if it's null or not?

This was because of `unwrap_or_default`

> I think a variant being explicitly set to null is an error condition that we could probably short-circuit on

Nice! Changed to reflect this

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -510,22 +511,26 @@ class JsonParserGenerator(
...
                                                 """,
                                                 "Union" to returnSymbolToParse.symbol, *codegenScope,
                                             )
                                         } else {
-                                            withBlock("Some(#T::$variantName(", "))", returnSymbolToParse.symbol) {
+                                            withBlock("variant = Some(#T::$variantName(", "));", returnSymbolToParse.symbol) {
                                                 deserializeMember(member)
-                                                unwrapOrDefaultOrError(member)
+                                                if (checkValueSet) {
+                                                    rust(""".ok_or_else(|| aws_smithy_json::deserialize::error::DeserializeError::custom("no variant was set in union"))?""")
```
82marbag, line 530: This can be checked in `unwrapOrDefaultOrError`. In `unwrapOrDefaultOrError` the error is returned as "this member cannot be null", but here we're checking that "one (any) member must be set". I can move the check inside `unwrapOrDefaultOrError`

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -510,22 +511,26 @@ class JsonParserGenerator(
                                 """,
                                 *codegenScope,
                             )
-                            withBlock("variant = match key.to_unescaped()?.as_ref() {", "};") {
```
jdisanti, line 513: What's the advantage to moving the `variant =` piece to inside of the match? Seems like it adds more overall code?

82marbag: Changed it back (it was used in the previous commit)

jdisanti (approved):

david-perez (approved):

PR #2253, opened 2023-01-26
Change the release flow to use release branches

LukeMathWalker:
Releases will no longer happen on `main`, they'll be cut from dedicated release branches (one for each release series, e.g. `1.x.y` or `0.5.x`).

Each branch will have a `CHANGELOG.md` file (related to the changes in that release series) and a `CHANGELOG.next.toml`.
`main` will have the `CHANGELOG.md` with historical entries and the `CHANGELOG.next.toml` file with the changes that will go into the next release series.

The `CHANGELOG.next.toml` file on `main` is automatically cleaned up if a new release series is kicked off from the current HEAD of `main`. It will need to be _manually_ cleaned up if a new release series series is kicked off from an older commit. I'm planning to remediate this in a follow-up PR (which will require the introduction of unique ids for each changelog entry, to be able to properly manipulate entries and be resistant to rewording/changes in the user-facing message).

All the required steps (creating the branch using a precise naming scheme, changing the runtime crates version in `gradle.properties`, tagging, publishing) will be handled by the updated release action.

----


tools/publisher/src/subcommand/generate_version_manifest.rs
```diff
@@ -143,14 +143,14 @@ fn find_released_versions(
...
-                    );
-                }
+                // if unrecent_version > recent_version {
+                //     bail!(
+                //         "Version number for `{}` decreased between releases (from `{}` to `{}`)",
+                //         crate_name,
+                //         unrecent_version,
+                //         recent_version
+                //     );
+                // }
```
LukeMathWalker, line 153: I had to comment this out since we are going from `0.54.1` to `0.0.0-smithy-rs-head` on `main` - the version will only be set on release branches in `gradle.properties`.
If my understanding is correct, we can uncomment this (and the related test) after this PR has been merged.
@jdisanti @rcoh @Velfi @ysaito1001

jdisanti: You can just delete it.

jdisanti: Although, I think we might have another problem... The SDK release process always uses the latest `versions.toml` in the `aws-sdk-rust` repo root. If we want to do patch releases of the SDK, I think we'll have to do some modifications to that release process to have it use branches as well.

LukeMathWalker: Yes, we'll need to do some changes there to be able to backport fixes for SDK crates.
I don't think changes will be necessary in order to do patch releases if they are the latest release (in the semver sense)?

.github/workflows/release.yml
```diff
@@ -6,48 +6,44 @@
...
+        type: string
       dry_run:
-        description: Dry runs will only produce release artifacts, but will not cut a release tag in GitHub nor publish to crates.io
+        description: Dry runs will only produce release artifacts, but they will not cut a release tag in GitHub nor publish to crates.io
         required: true
         type: boolean
         default: true
 
 jobs:
-  main-branch-check:
```
jdisanti, line 27: Is there any way we could check that the versions we will be publishing in `gradle.properties` aren't already published? Would be a nice sanity check that the `gradle.properties` was correctly updated.

.github/workflows/release.yml
```diff
@@ -61,17 +57,151 @@ jobs:
...
+        with:
+          ref: ${{ inputs.commit_sha }}
+          token: ${{ secrets.RELEASE_AUTOMATION_BOT_PAT }}
+      - name: Get or create release branch
+        id: branch-push
+        shell: bash
+        env:
+          SEMANTIC_VERSION: ${{ inputs.semantic_version }}
+          DRY_RUN: ${{ inputs.dry_run }}
+        run: |
```
jdisanti, line 86: Consider pulling this out to a separate shell script file so that it can be easily tested locally

.github/workflows/release.yml
```diff
@@ -61,17 +57,151 @@ jobs:
...
+            # The file was actually changed, we need to commit and push the changes
+            git commit gradle.properties -m "Upgrade the smithy-rs runtime crates version to ${SEMANTIC_VERSION}"
+            echo "Pushing upgraded gradle.properties commits..."
+            git push origin
+          fi
+          echo "commit_sha=$(git rev-parse --verify --short HEAD)" > $GITHUB_OUTPUT
+
+  # We'll need to build a base image again if the previous job upgraded the gradle.properties file.
+  # This job will be a no-op if the version in gradle.properties had already been upgraded.
+  acquire-base-image-2:
```
jdisanti, line 171: What in `tools/` uses the `gradle.properties` file? Doesn't seem like this should be necessary. Even if a tool reads `gradle.properties` from the filesystem, it shouldn't need an updated Docker image since that version number isn't being compiled into the tool.

Separately, you will probably want an escape hatch for using the latest tools in `main` for the actual release workflow since you may need an updated `publisher` to actually successfully publish.

.github/workflows/release.yml
```diff
@@ -133,3 +259,31 @@ jobs:
         else
           publisher publish -y --location .
         fi
+
+  trim-changelog-next-on-main:
+    name: Remove released entries from CHANGELOG.next.toml on the main branch
+    if: inputs.dry_run == false && needs.get-or-create-release-branch.outputs.new_release_series == true
```
jdisanti, line 265: SDK changelog entries need to end up in `aws/SDK_CHANGELOG.next.json` upon release so that the SDK changelog can be generated when it is released. It currently uses the `changelogger split` subcommand to accomplish this. Does this new release process preserve that?

.github/workflows/release.yml
LukeMathWalker: It'd be quite cumbersome.
Things can fail at any step, including publishing - leaving us with some unpublished runtime crates. I'm opting for building the workflow in such a way that all steps are idempotent, therefore allowing us to retry any transient failure by re-triggering the workflow from the right commit.

.github/workflows/release.yml
LukeMathWalker: > Separately, you will probably want an escape hatch for using the latest tools in main for the actual release workflow since you may need an updated publisher to actually successfully publish.

My assumption is that you can always merge the upgraded tools into the release branch if that's required, indeed as an escape hatch.

.github/workflows/release.yml
LukeMathWalker: Yes - the `aws/SDK_CHANGELOG.next.json` will be on the release branch and it's generated in the previous step of the workflow.

tools/publisher/src/subcommand/generate_version_manifest.rs
LukeMathWalker: Removed in [f8de002](https://github.com/awslabs/smithy-rs/pull/2253/commits/f8de00227858cef87c63cdb1603b92d755410444)

.github/workflows/release.yml
LukeMathWalker: Done in [dd472e4](https://github.com/awslabs/smithy-rs/pull/2253/commits/dd472e435caa69cc15f378c16bc46c964c9471e2)
Shellcheck found _a lot_ of issues 😅 Thanks for pointing it out.

.github/workflows/release.yml
```diff
@@ -83,56 +83,7 @@ jobs:
...
-                # We can then create create the release branch and set the current commit as its tip
-                git checkout -b ${branch_name} 
-                git push origin ${branch_name}
-              else
-                echo "You must choose a commit from main to create a new release series!"
-                exit 1
-              fi
-            fi
-          fi
+        run: ./tools/get-or-create-release-branch.sh
```
jdisanti, line 86: Prefer this to be in `.github/scripts/`.

tools/get-or-create-release-branch.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
```
jdisanti, line 7: Nit: Would be good to add a high level explanation of what this is and what it's used for, how to test locally, etc.

jdisanti (approved):
LGTM! Couple minor suggestions.

tools/publisher/src/subcommand/upgrade_runtime_crates_version.rs
```diff
@@ -0,0 +1,72 @@
...
+) -> Result<(), anyhow::Error> {
+    let upgraded_version = semver::Version::parse(args.version.as_str())
+        .with_context(|| format!("{} is not a valid semver version", &args.version))?;
+    let fs = Fs::Real;
+    let gradle_properties = read_gradle_properties(fs, &args.gradle_properties_path).await?;
+    let version_regex =
+        Regex::new(r"(?P<field>smithy\.rs\.runtime\.crate\.version=)(?P<version>\d+\.\d+\.\d+-.*)")
+            .unwrap();
+    let current_version = version_regex.captures(&gradle_properties).ok_or_else(|| {
+        anyhow!("Failed to extract the expected runtime crates version from gradle.properties")
```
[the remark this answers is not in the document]
david-perez: We're hardcoding `gradle.properties` but the CLI takes the file as an argument.

tools/publisher/src/subcommand/upgrade_runtime_crates_version.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[derive(Parser, Debug)]
+pub struct UpgradeRuntimeCratesVersionArgs {
+    /// The version of runtime crates you want the code generator to use (e.g. `0.52.0`).
+    #[clap(long)]
+    version: String,
+    /// The path to the `gradle.properties` file. It will default to `gradle.properties` if
+    /// left unspecified.
+    #[clap(long, default_value = "gradle.properties")]
+    gradle_properties_path: PathBuf,
```
david-perez, line 20: Is there a use case for performing this substitution in another file? Unsure this param is needed.

tools/publisher/src/subcommand/upgrade_runtime_crates_version.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::fs::Fs;
+use anyhow::{anyhow, bail, Context};
+use clap::Parser;
+use regex::Regex;
+use std::path::{Path, PathBuf};
+
+#[derive(Parser, Debug)]
+pub struct UpgradeRuntimeCratesVersionArgs {
+    /// The version of runtime crates you want the code generator to use (e.g. `0.52.0`).
+    #[clap(long)]
```
david-perez, line 15: nit: Perhaps this should just be a positional arg, lest it be misconstrued as the version of a part of the CLI (?)

tools/publisher/src/subcommand/upgrade_runtime_crates_version.rs
```diff
@@ -0,0 +1,72 @@
...
+        updated_gradle_properties.as_ref(),
+    )
+    .await?;
+    Ok(())
+}
+
+async fn read_gradle_properties(fs: Fs, path: &Path) -> Result<String, anyhow::Error> {
+    let bytes = fs.read_file(path).await?;
+    let contents =
+        String::from_utf8(bytes).context("The gradle.properties file contained non-UTF8 data")?;
```
david-perez, line 61:
```suggestion
        String::from_utf8(bytes).context("the `gradle.properties` file contained non-UTF8 data")?;
```

Same here, the filename is hardcoded.

david-perez (approved):

.github/scripts/get-or-create-release-branch.sh
```diff
@@ -0,0 +1,82 @@
...
+# The script populates an output file with key-value pairs that are needed in the release CI workflow to carry out
+# the next steps in the release flow: the name of the release branch and a boolean flag that is set to 'true' if this
+# is the beginning of a new release series.
+
+if [ -z "$SEMANTIC_VERSION" ]; then
+  echo "'SEMANTIC_VERSION' must be populated."
+  exit 1
+fi
+
+if [ -z "$1" ]; then
```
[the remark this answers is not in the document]
LukeMathWalker: > Why do we require a filename to be passed in for the output, if then we're cating its contents to $GITHUB_OUTPUT in the invocation? As opposed to just writing to stdout and placing all informative messages in stderr as is typical.

That was my initial approach, but it's quite easy to break it by mistake - invoke a command, forget to redirect what it outputs to stderr.. 💥 
The file is used as a way to be very explicit with respect to what will end up being used to set the step output variables.

The touch is superfluous, I can remove it.

.github/scripts/get-or-create-release-branch.sh
```diff
@@ -0,0 +1,82 @@
...
+  fi
+fi
+
+if [[ "${DRY_RUN}" == "true" ]]; then
+  branch_name="${branch_name}-preview"
+fi
+echo "release_branch=${branch_name}" > "${output_file}"
+
+if [[ "${DRY_RUN}" == "true" ]]; then
+  git push -f origin "HEAD:${branch_name}"
```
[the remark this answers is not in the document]
LukeMathWalker: It's a good default, I agree.

tools/publisher/src/subcommand/upgrade_runtime_crates_version.rs
LukeMathWalker: No, the parameter is mostly there to allow invoking the script from a difference directory (i.e. avoid assuming that the file exists in the current working directory).

.github/scripts/get-or-create-release-branch.sh
```diff
@@ -0,0 +1,82 @@
...
+else
+  commit_sha=$(git rev-parse --verify --short HEAD)
+  if git ls-remote --exit-code --heads origin "${branch_name}"; then
+    # The release branch already exists, we need to make sure that our commit is its current tip
+    branch_head_sha=$(git rev-parse --verify --short refs/heads/patches)
```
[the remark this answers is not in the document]
LukeMathWalker: Leftover from my local testing, good catch.

PR #2256, opened 2023-01-27
Constraint member types are refactored as standalone shapes.

drganjoo:
## Motivation and Context

Code generation for [member constraint traits](https://smithy.io/2.0/spec/model.html#traits)

Traits that target [members](https://smithy.io/2.0/spec/model.html#member) apply only in the context of the member shape and do not affect the shape targeted by the member. Traits applied to a member supersede traits applied to the shape targeted by the member and do not inherently conflict.

Issue: https://github.com/awslabs/smithy-rs/issues/1969

## Description

The model is transformed and all constrained member shapes in structure, union, map and list, are extracted as standalone shapes in the model, with an additional trait `SyntheticStructureFromConstrainedMemberTrait` applied to them.

## Testing

`./gradlew codegen-server-test:test`


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+object RefactorConstrainedMemberType {
+    private data class MemberTransformation(
+        val newShape: Shape,
+        val memberToChange: MemberShape,
+        val traitsToKeep: List<Trait>,
+    )
+
+    /**
+     * Transforms all member shapes that have constraints on them into equivalent
+     * non-constrained member shapes.
```
david-perez, line 29:
```suggestion
     * Transforms all structure member shapes that have constraints on them into equivalent
     * non-constrained structure member shapes targeting synthetic constrained structure shapes with the member's constraints.
```

I'd move this documentation above `object` and include a simple example Smithy model so that the reader can get the gist of this transformation.

E.g.:

```
structure A {
    @length(min: 1, max: 69)
    overriddenConstrainedString: ConstrainedString
}

@length(min: 2, max: 10)
@pattern("^[A-Za-z]+$")
string ConstrainedString
```

to

```
structure A {
    overridenConstrainedString: OverriddenConstrainedString
}

@length(min: 1, max: 69)
@pattern("^[A-Za-z]+$")
OverriddenConstrainedString
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+        val transformations = model.operationShapes
+            .flatMap { operation ->
+                listOfNotNull(operation.input.orNull(), operation.output.orNull())
+            }
+            .map { model.expectShape(it) }
+            .flatMap {
+                val walker = DirectedWalker(model)
+                walker.walkShapes(it)
+            }
+            .filterIsInstance(StructureShape::class.java)
```
david-perez, line 43: All this does is getting reachable structure shapes. We can just use `model.structureShapes` here; the plugin takes care of not generating shapes that are not reachable from an operation.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+        return ModelTransformer.create()
+            .replaceShapes(modelBuilder.build(), memberShapesToChange)
+    }
+
+    /**
+     * Returns a list of members that have constraint traits applied to them
+     */
+    private fun StructureShape.constrainedMembers(): List<MemberShape> =
+        this.allMembers.values.filter {
+            it.hasConstraintTrait()
```
david-perez, line 86: This is transforming member shapes which have `@required` on them, which we don't want to touch (see the definition of `hasConstraintTrait`).

Add a negative test so that we don't transform these.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+
+            val changedMember = it.memberToChange.toBuilder()
+                .target(it.newShape.id)
+                .traits(it.traitsToKeep)
+                .build()
+            memberShapesToChange.add(changedMember)
+        }
+
+        // Change all original constrained member shapes with the new standalone types,
+        // and keep only the non-constraint traits on the member shape.
```
david-perez, line 76:
```suggestion
        // Change all original constrained member shapes with the new non-constrained member shapes,
        // keeping only the non-constraint traits on them.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+    /***
+     * Returns a Model that has all the transformations applied on the original
+     * model.
+     */
+    private fun applyTransformations(
+        model: Model,
+        transformations: List<MemberTransformation>,
+    ): Model {
+        if (transformations.isEmpty())
+            return model
```
david-perez, line 60:
```suggestion
```

The result is the same without this early return, right?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+     * Returns the transformation that would be required to turn the given member shape
+     * into a non-constrained member shape.
+     */
+    private fun MemberShape.makeNonConstrained(
+        model: Model,
+        existingRefactoredNames: MutableSet<ShapeId>,
+    ): MemberTransformation? {
+        val (constrainedTraits, otherTraits) = this.allTraits.values
+            .partition {
+                allConstraintTraits.contains(it.javaClass)
```
david-perez, line 151: This contains `@required`, which we don't want to touch.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Returns the transformation that would be required to turn the given member shape
+     * into a non-constrained member shape.
+     */
+    private fun MemberShape.makeNonConstrained(
+        model: Model,
+        existingRefactoredNames: MutableSet<ShapeId>,
+    ): MemberTransformation? {
+        val (constrainedTraits, otherTraits) = this.allTraits.values
```
david-perez, line 149:
```suggestion
        val (constraintTraits, otherTraits) = this.allTraits.values
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+            .partition {
+                allConstraintTraits.contains(it.javaClass)
+            }
+
+        // No transformation required in case the given MemberShape has no constraints.
+        if (constrainedTraits.isEmpty())
+            return null
+
+        // Create a new standalone shape of the same target as the target of the member e.g.
+        // someX : Centigrade, where integer : Centigrade, then the new shape should be
```
david-perez, line 159: I think a comprehensive example in the `object`'s documentation will suffice in terms of documentation.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+        // integer: newShape
+        val targetShape = model.expectShape(this.target)
+
+        // The shape has to be a buildable type otherwise it cannot be refactored.
+        if (targetShape is ToSmithyBuilder<*>) {
```
david-perez, line 164: Aren't all targets of member shapes buildable shapes? In which case, I'd check for the negative condition and use `software.amazon.smithy.rust.codegen.core.util.UNREACHABLE` to better convey that we're doing this just for Kotlin type narrowing.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+            when (val builder = targetShape.toBuilder()) {
+                is AbstractShapeBuilder<*, *> -> {
+                    // Use the target builder to create a new standalone shape that would
+                    // be added to the model later on. Keep all existing traits on the target
+                    // but replace the ones that are overriden on the member shape
+                    val existingNonOverridenTraits =
+                        builder.allTraits.values.filter { existingTrait ->
+                            !constrainedTraits.any { it.toShapeId() == existingTrait.toShapeId() }
+                        }
```
david-perez, line 173: This is removing constraint traits from the target shape that are also on the member shape? But then we're adding them back in the line below (`existingNonOverridenTraits + constrainedTraits`)?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+        // integer: newShape
+        val targetShape = model.expectShape(this.target)
+
+        // The shape has to be a buildable type otherwise it cannot be refactored.
+        if (targetShape is ToSmithyBuilder<*>) {
+            when (val builder = targetShape.toBuilder()) {
+                is AbstractShapeBuilder<*, *> -> {
+                    // Use the target builder to create a new standalone shape that would
+                    // be added to the model later on. Keep all existing traits on the target
+                    // but replace the ones that are overriden on the member shape
```
david-perez, line 169:
```suggestion
                    // but replace the ones that are overridden on the member shape
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+                    val existingNonOverridenTraits =
+                        builder.allTraits.values.filter { existingTrait ->
+                            !constrainedTraits.any { it.toShapeId() == existingTrait.toShapeId() }
+                        }
+
+                    val newTraits = existingNonOverridenTraits + constrainedTraits + RefactoredStructureTrait(this.id)
+
+                    // Create a new standalone type that would be added to the model later on
```
david-perez, line 177:
```suggestion
                    // Create a new standalone shape that will be added to the model later on
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+
+        // Create a new standalone shape of the same target as the target of the member e.g.
+        // someX : Centigrade, where integer : Centigrade, then the new shape should be
+        // integer: newShape
+        val targetShape = model.expectShape(this.target)
+
+        // The shape has to be a buildable type otherwise it cannot be refactored.
+        if (targetShape is ToSmithyBuilder<*>) {
+            when (val builder = targetShape.toBuilder()) {
+                is AbstractShapeBuilder<*, *> -> {
```
david-perez, line 166: This is needed because `.allTraits` and `.id` are only defined on `AbstractShapeBuilder` right?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+
+                    // Create a new standalone type that would be added to the model later on
+                    val shapeId = refactoredShapeId(model, existingRefactoredNames, this.id)
+                    val standaloneShape = builder.id(shapeId)
+                        .traits(newTraits)
+                        .build()
+
+                    // Since the new shape has not been added to the model as yet, the current
+                    // memberShape's target cannot be changed to the new shape.
+                    // + OriginalShapeIdTrait(this.target)
```
david-perez, line 185:
```suggestion
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+
+                suffix = "_$it"
+            }
+
+            // A unique type could not be found by adding a numeric suffix to it. As a
+            // secondary algorithm, use existing struct names as suffix to generate a unique name.
+            suffix = "_"
+
+            model.listShapes
+                .filterIsInstance<StructureShape>()
```
david-perez, line 124: Isn't this always calculating an empty collection? List shapes cannot be structure shapes.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+     */
+    private fun StructureShape.constrainedMembers(): List<MemberShape> =
+        this.allMembers.values.filter {
+            it.hasConstraintTrait()
+        }
+
+    /**
+     * Returns the unique (within the model) name of the refactored shape
+     */
+    private fun refactoredShapeId(
```
david-perez, line 92: Let's add the new shape id to `existingRefactoredNames` in the caller to avoid mutable state.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+        fun generateUniqueName(): ShapeId {
+            // Ensure the name does not already exist in the model, else make it unique
+            // by appending a new number as the suffix.
+            var suffix = ""
+            (0..100).forEach {
+                val extractedStructName = makeStructName(suffix)
+                if (structNameIsUnique(extractedStructName))
+                    return extractedStructName
+
+                suffix = "_$it"
```
david-perez, line 116: No need to use a `var` for `suffix` here, we can calculate it inside each iteration. I'd drop the underscore too; I've never seen underscores in shape IDs.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
```diff
@@ -0,0 +1,195 @@
...
+            }
+
+            // A unique type could not be found by adding a numeric suffix to it. As a
```
david-perez, line 119: I think I'd abort here; the main algorithm should be good enough and not cause collisions.

david-perez (changes requested):
Looks promising! Haven't looked at the tests yet, but I've published a first round of comments for the implementation code.

david-perez:
This PR has "DRAFT" in its title but is not a draft PR. No need to put "DRAFT" in the title; you can mark a GitHub PR as a draft [like this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft).

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
drganjoo: Oh `listShapes` returns shapes of the type `ListShape.class`. I mistook it for `list all Shapes`. Anyhow, have removed this piece of code as we are aborting early.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
drganjoo: Have added the negative check and UNREACHABLE.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
drganjoo: Well, this is getting all the traits that are defined on the target shape but not on the member shape. There is an outer negation `!` on the `any`. I can see how confusing this was, so I have changed the `any` with the `none` operator now. 

```
                val existingNonOverriddenTraits =
                    builder.allTraits.values.filter { existingTrait ->
                        constraintTraits.none { it.toShapeId() == existingTrait.toShapeId() }
                    }
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/RefactorConstrainedMemberType.kt
drganjoo: Have changed it now to keep `@required` on the member.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
```diff
@@ -160,6 +161,9 @@ open class ServerCodegenVisitor(
             .let { ModelTransformer.create().copyServiceErrorsToOperations(it, settings.getService(it)) }
             // Add `Box<T>` to recursive shapes as necessary
             .let(RecursiveShapeBoxer::transform)
+            // Refactors constrained member shapes into non-constrained shapes of a new standalone type
+            // that has some constraints.
+            .let(RefactorConstrainedMemberType::transform)
```
drganjoo, line 166: Have renamed it to `ConstrainedStructureMemberTransform` and have removed the word `refactor` from wherever it was being used.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ConstrainedMemberTransform.kt
```diff
@@ -0,0 +1,210 @@
...
+
+                // Since the new shape has not been added to the model as yet, the current
+                // memberShape's target cannot be changed to the new shape.
+                return MemberShapeTransformation(standaloneShape, this, otherTraits)
+            }
+
+            else -> throw IllegalStateException("Constraint traits cannot to applied on ${this.id}") // FZ confirm how we are throwing exceptions
+        }
+
+        throw IllegalStateException("Constraint traits can only be applied to buildable types. ${this.id} is not buildable") // FZ confirm how we are throwing exceptions
```
david-perez, line 208: This line is not necessary. We must go through the first arm of the `when`, which `return`s.

You can rewrite this so that it's more apparent as follows, note the `return when`, each arm returns an expression:

```kotlin
return when (val builder = targetShape.toBuilder()) {
    is AbstractShapeBuilder<*, *> -> {
        ...
        MemberShapeTransformation(standaloneShape, this, otherTraits)
    }

    else -> UNREACHABLE("Constraint traits cannot to applied on ${this.id}")
}
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/traits/SyntheticStructureFromConstrainedMemberTrait.kt
```diff
@@ -0,0 +1,14 @@
+package software.amazon.smithy.rust.codegen.server.smithy.traits
+
+import software.amazon.smithy.model.node.Node
+import software.amazon.smithy.model.shapes.ShapeId
+import software.amazon.smithy.model.traits.AnnotationTrait
+
+/**
+ * Trait applied to a refactored shape indicating the structure that contains member of this new structure type
+ */
+class SyntheticStructureFromConstrainedMemberTrait(val sourceMember : ShapeId): AnnotationTrait(SyntheticStructureFromConstrainedMemberTrait.ID, Node.objectNode())  {
```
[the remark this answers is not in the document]
drganjoo: I am not sure calling `model.toBuilder()` and then calling `toBuild()` on it retains the existing JVM shape object. But I will try this out and if the object remains the same, will keep the shape instead of the ID.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ConstrainedMemberTransform.kt
```diff
@@ -0,0 +1,210 @@
...
+        val transformations = model.operationShapes
+            .flatMap { operation ->
+                listOfNotNull(operation.input.orNull(), operation.output.orNull())
+            }
+            .map { model.expectShape(it) }
+            .flatMap {
+                val walker = DirectedWalker(model)
+                walker.walkShapes(it)
+            }
+            .filter { it.isStructureShape || it.isListShape || it.isUnionShape || it.isMapShape }
```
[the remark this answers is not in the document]
drganjoo: Oh okay, changing it.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ConstrainedMemberTransform.kt
```diff
@@ -0,0 +1,210 @@
...
+        model: Model,
+        additionalNames: Set<ShapeId>,
+        memberShape: ShapeId,
+    ): ShapeId {
+        val structName = memberShape.name
+        val memberName = memberShape.member.orElse(null)
+            .replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() }
+
+        fun makeStructName(suffix: String = "") =
+            ShapeId.from("${memberShape.namespace}#Overridden${structName}${memberName}$suffix")
```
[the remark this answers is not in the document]
drganjoo: Oh I had tried doing that but it raises an exception:

```
software.amazon.smithy.model.shapes.ShapeIdSyntaxException: Invalid shape ID: OverriddenMalformedRangeOverrideInputByte
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriter.kt
```diff
@@ -0,0 +1,217 @@
...
+import software.amazon.smithy.rust.codegen.core.rustlang.Writable
+import software.amazon.smithy.rust.codegen.core.rustlang.comment
+import software.amazon.smithy.rust.codegen.core.smithy.RustCrate
+import java.util.concurrent.ConcurrentHashMap
+
+typealias DocWriter = () -> Unit
+
+/**
+ * Creates a sequence of writable, each of which need to write in the same `rust_module::inline_module`. Calling
+ * `render()`, creates an inline module and then invokes all of the writable.
```
david-perez, line 15:
```suggestion
 * Creates a sequence of `Writable`s, each of which need to write in the same `rust_module::inline_module`. Calling
 * `render()`, creates the inline module and then invokes all of the writables.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriter.kt
```diff
@@ -0,0 +1,217 @@
...
+ */
+private class RustCrateInlineModuleComposingWriter {
+    companion object {
+        // A map of RustCrate to the InlineModuleWriter that has been created for that RustCrate
+        private val crateToInlineModule: ConcurrentHashMap<RustCrate, RustCrateInlineModuleComposingWriter> =
+            ConcurrentHashMap()
+
+        /**
+         * If the given shape is an "overridden constrained member" shape, then it registers
+         * the writeable to be called later on at the time of rendering. However, if the shape
```
david-perez, line 25:
```suggestion
         * the writable to be called later on at the time of rendering. However, if the shape
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriter.kt
```diff
@@ -0,0 +1,217 @@
...
+    )
+
+    /**
+     * Keeps a mapping of InlineModule -> Writables that are to be called at render time.
+     */
+    class InlineModules {
+        private val inlineModuleWritables: HashMap<RustModule.LeafModule, InlineCodeAndDocWritable> = HashMap()
+
+        /**
+         * Adds a writeable to the inline module
```
david-perez, line 101:
```suggestion
         * Adds a `Writable` to the inline module.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -151,6 +153,15 @@ class ServerBuilderGenerator(
...
+    fun composeDocAndCodeWritable(rustCrate : RustCrate, writer: RustWriter) {
+        val docWriter : () -> Unit = {
+            writer.docs("See #D.", structureSymbol)
+        }
+        rustCrate.withComposableInlineModule(builderSymbol.module(), codegenContext, docWriter) {
+            renderBuilder(this)
+        }
+    }
+
     fun render(writer: RustWriter) {
```
david-perez, line 165: I'd have guessed that this method would become `private` and tests should use `composeDocAndCodeWritable` too.

`composeDocAndCodeWritable` could simply be named `render`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderSymbol.kt
```diff
@@ -17,13 +17,23 @@ import software.amazon.smithy.rust.codegen.core.smithy.rustType
 import software.amazon.smithy.rust.codegen.core.util.toSnakeCase
 import software.amazon.smithy.rust.codegen.server.smithy.ServerCodegenContext
 
+val alreadyDefined : HashMap<String, RustModule.LeafModule> = HashMap()
```
david-perez, line 20: Perhaps this should be `ConcurrentHashMap` like the one in `RustCrateInlineModuleComposingWriter`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderSymbol.kt
```diff
@@ -17,13 +17,23 @@ import software.amazon.smithy.rust.codegen.core.smithy.rustType
...
+
+
 fun StructureShape.serverBuilderSymbol(codegenContext: ServerCodegenContext): Symbol =
     this.serverBuilderSymbol(
         codegenContext.symbolProvider,
         !codegenContext.settings.codegenConfig.publicConstrainedTypes,
     )
 
-fun StructureShape.serverBuilderSymbol(symbolProvider: SymbolProvider, pubCrate: Boolean): Symbol {
+fun StructureShape.serverBuilderModule(codegenContext: ServerCodegenContext): RustModule.LeafModule {
```
david-perez, line 29: I don't think this is used now.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderSymbol.kt
david-perez: Actually, why does this need to be a map? Can't it be a set of `RustModule.LeafModule`s?

david-perez: Coming back to this again, why do we need to memoize the modules at all? Can't `serverBuilderModule` simply always calculate and return `RustModule.new(builderNamespace, visibility, parent = structureSymbol.module(), inline = true)`?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -794,80 +794,11 @@ class ServerProtocolTestGenerator(
...
-            FailingTest(MalformedRangeValidation, "RestJsonMalformedRangeMinShortOverride", TestType.MalformedRequest),
-            FailingTest(
-                MalformedRangeValidation,
-                "RestJsonMalformedRangeMinIntegerOverride",
-                TestType.MalformedRequest,
-            ),
-            FailingTest(MalformedRangeValidation, "RestJsonMalformedRangeMinLongOverride", TestType.MalformedRequest),
-            FailingTest(RestJsonValidation, "RestJsonMalformedRangeByteOverride_case0", TestType.MalformedRequest),
-            FailingTest(RestJsonValidation, "RestJsonMalformedRangeByteOverride_case1", TestType.MalformedRequest),
+            // Tests involving float range based member constraint traits
```
david-perez, line 797:
```suggestion
            // Tests involving floating point shapes and the `@range` trait; see https://github.com/awslabs/smithy-rs/issues/2007.
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ConstrainedMemberTransform.kt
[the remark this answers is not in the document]
david-perez: That looks like you constructed the shape ID without the `#` character, which delimits the namespace.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriter.kt
drganjoo: I've fixed all `writeable` now :)

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
drganjoo: I am with you on this. But right now, for the inline modules to be written out one has to call `RustCrate.renderComposableInlineModules`. Basically the inline writer needs to know that there are no more writers who would be writing to it, so that it can go ahead and render the registered `Writable`. At the moment, I couldn't figure out a better way of doing this than requiring someone to put an end to it by calling a function.

I will check what needs to be done in the tests for this to work.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderSymbol.kt
drganjoo: Good catch. Actually, this `HashMap` thing was never to be committed. At that time, I thought the reason it keeps creating new inline modules was that the code keeps on calling `RustModule.new` instead of returning the same inline module object.

Have fixed it and now it returns `RustModule.new(...)`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderSymbol.kt
drganjoo: Oh yes! have removed this now.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ConstrainedMemberTransform.kt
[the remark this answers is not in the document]
drganjoo: Oh you mean get rid of the `Overridden` word :o yes you are absolutely right. Getting rid of it.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstraintsMemberShapeTest.kt
```diff
@@ -0,0 +1,1012 @@
...
+
+    @Test
+    fun `extra trait on a constrained member should remain on it`() {
+        val transformedModel = loadModel(outputModelOnly)
+        checkShapeHasTrait(
+            transformedModel,
+            outputModelOnly,
+            "constrainedMemberShape.synthetic#OperationUsingGetOutput\$constrainedPatternString",
+            "constrainedMemberShape#OperationUsingGetOutput\$constrainedPatternString",
+            "aws.api#data",
```
david-perez, line 264: You can pass in `software.amazon.smithy.aws.traits.DataTrait` directly. You can make `checkShapeHasTrait` take in a `software.amazon.smithy.model.traits.Trait`, and use `memberShape.expectTrait<Trait>()` inside, instead of having to always manually construct the trait using `ShapeId.from(traitName)`.

david-perez: I think we should also test that `required` remains.

david-perez (commented):
Tests look solid! All we're missing is for lists/maps/unions to generate their member shapes in their own modules when constrained, like we've done with structure member shapes.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ConstrainedMemberTransform.kt
```diff
@@ -0,0 +1,222 @@
...
+        val memberName = memberShape.member.orElse(null)
+            .replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() }
+
+        fun makeStructName(suffix: String = "") =
+            ShapeId.from("${memberShape.namespace}#${structName}${memberName}$suffix")
+
+        fun structNameIsUnique(newName: ShapeId) =
+            model.getShape(newName).isEmpty && !additionalNames.contains(newName)
+
+        fun generateUniqueName(): ShapeId {
```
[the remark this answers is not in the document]
drganjoo: For the following conformant model:

```
namespace X

structure A {
  @range(min:10)
  b Integer
}

structure Ab {
  .....
}
```

If we do only `${memberShape.namespace}#${structName}${memberName}` then the extracted name `X#Ab` won't be unique.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
drganjoo: Oh yes. Let me fix this.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriter.kt
```diff
@@ -0,0 +1,346 @@
...
+
+/**
+ * For each RustCrate a separate mapping of inline-module to `RustWriter` is maintained.
+ */
+private val crateToInlineModule: ConcurrentHashMap<RustCrate, InnerModule> =
+    ConcurrentHashMap()
+
+class InnerModule(debugMode : Boolean) {
+    private val topLevelModuleWriters: MutableSet<RustWriter> = mutableSetOf()
+    private val inlineModuleWriters: HashMap<RustWriter, MutableList<InlineModuleWithWriter>> = hashMapOf()
```
82marbag, line 136: Doesn't this have to be concurrent too?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/traits/SyntheticStructureFromConstrainedMemberTrait.kt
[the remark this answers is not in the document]
drganjoo: We keep the actual `Shapes` in the `SyntheticTrait` now.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriter.kt
```diff
@@ -0,0 +1,327 @@
...
+        return hierarchy
+    }
+
+
+    /**
+     * Writes out each inline module's code (`toString`) to the respective top level `RustWriter`.
+     */
+    fun render() {
+        fun writeInlineCode(rustWriter: RustWriter, code: String) {
+            val inlineCode = code.drop(emptyLineCount)
```
david-perez, line 247: Haven't seen this before. What is it for?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriter.kt
```diff
@@ -0,0 +1,327 @@
...
+        // with the bottom most inline module that has been passed.
+        if (hierarchy.isNotEmpty()) {
+            val topMost = hierarchy.removeFirst()
+
+            // Create an intermediate writer for all inner modules in the hierarchy
+            rustCrate.withModule(topMost) {
+                var writer = this
+                hierarchy.forEach {
+                    writer = getWriter(writer, it)
+                }
```
david-perez, line 175: What does this ensure? Isn't at the end of this for loop `writer` equal to the last element in `hierarchy` (which is the penultimate module, since we called `removeLast` before).

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerSymbolProviders.kt
```diff
@@ -28,23 +28,25 @@ class ServerSymbolProviders private constructor(
...
+                    publicConstrainedTypes,
+                    true
                 ),
                 unconstrainedShapeSymbolProvider = UnconstrainedShapeSymbolProvider(
                     baseSymbolProviderFactory(
                         model,
                         service,
                         symbolVisitorConfig,
                         false,
+                        false,
```
david-perez, line 49: This is new. Why do we need to exclude the `ConstrainedShapeSymbolProvider` when constructing this one?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedCollectionGenerator.kt
```diff
@@ -121,23 +122,23 @@ class ConstrainedCollectionGenerator(
 
         writer.rustTemplate(
             """
-            impl #{TryFrom}<$inner> for $name {
-                type Error = #{ConstraintViolation};
+        impl #{TryFrom}<$inner> for $name {
```
david-perez, line 125: This looks like it needs to be reverted, the formatting was correct before (the beginning of `impl` should align with the first double quote in the preceding line).

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedCollectionGenerator.kt
```diff
@@ -146,7 +147,8 @@ class ConstrainedCollectionGenerator(
         if (!publicConstrainedTypes &&
             innerShape.canReachConstrainedShape(model, symbolProvider) &&
             innerShape !is StructureShape &&
-            innerShape !is UnionShape
+            innerShape !is UnionShape &&
+            innerShape !is EnumShape
```
david-perez, line 151: Oh, this is interesting. So we had a bug here? A list of enums indeed does not need this converter.

david-perez: Don't we have the same bug in `ConstrainedMapGenerator`, with maps whose values are enums?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedMapGenerator.kt
```diff
@@ -130,14 +132,22 @@ class ConstrainedMapGenerator(
...
             valueShape !is UnionShape
         ) {
+            val keyShape = model.expectShape(shape.key.target, StringShape::class.java)
+            val keyNeedsConversion = keyShape.typeNameContainsNonPublicType(model, symbolProvider, publicConstrainedTypes)
+            val key = if (keyNeedsConversion) {
+                "k.into()"
+            } else {
+                "k"
+            }
+
```
david-perez, line 142: We had a bug here too? Maps whose keys are constrained strings but not enums need this conversion right?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiator.kt
```diff
@@ -47,12 +48,26 @@ class ServerBuilderKindBehavior(val codegenContext: CodegenContext) : Instantiat
...
         // Only operation input builders take in unconstrained types.
         val takesInUnconstrainedTypes = shape.isReachableFromOperationInput()
-        return ServerBuilderGenerator.hasFallibleBuilder(
-            shape,
-            codegenContext.model,
-            codegenContext.symbolProvider,
-            takesInUnconstrainedTypes,
-        )
+
+        val publicConstrainedTypes = if (codegenContext is ServerCodegenContext) {
```
david-perez, line 52: This is correct, although `codegenContext` should always be of type `ServerCodegenContext` here. I don't understand why we're invoking the server instantiator when `publicConstrainedTypes` is `false` though.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedUnionGenerator.kt
```diff
@@ -133,11 +133,15 @@ class UnconstrainedUnionGenerator(
         } else {
             Visibility.PUBCRATE
         }
-        modelsModuleWriter.withInlineModule(
-            constraintViolationSymbol.module(),
+
+        inlineModuleCreator(
+            //constraintViolationSymbol.module(), codegenContext
```
david-perez, line 138:
```suggestion
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/transformers/ConstrainedMemberTransform.kt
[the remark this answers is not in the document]
david-perez: Ah, you're right! Good catch!

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriterTest.kt
```diff
@@ -0,0 +1,280 @@
...
+import software.amazon.smithy.rust.codegen.core.testutil.unitTest
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverTestCodegenContext
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverTestSymbolProvider
+import java.io.File
+
+class RustCrateInlineModuleComposingWriterTest {
+    private val rustCrate: RustCrate
+    private val codegenContext: ServerCodegenContext
+    private val model: Model = """
+            ${'$'}version: "2.0"
```
david-perez, line 27: You can also feed `"2.0"` to `asSmithyModel`

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriterTest.kt
```diff
@@ -0,0 +1,280 @@
...
+            use aws.api#data
+            use aws.protocols#restJson1
+
+            @title("Weather Service")
+            @restJson1
+            service WeatherService {
+                operations: [MalformedPatternOverride]
+            }
+
+            @suppress(["UnstableTrait"])
```
david-perez, line 39:
```suggestion
```

Not sure why this is needed.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriterTest.kt
```diff
@@ -0,0 +1,280 @@
...
+    private val rustCrate: RustCrate
+    private val codegenContext: ServerCodegenContext
+    private val model: Model = """
+            ${'$'}version: "2.0"
+            namespace test
+
+            use aws.api#data
+            use aws.protocols#restJson1
+
+            @title("Weather Service")
```
david-perez, line 33:
```suggestion
```

Try to keep test models as minimal as possible. For example, the `data` trait above is imported but I don't see it used either.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriterTest.kt
```diff
@@ -0,0 +1,280 @@
...
+            }
+
+            structure MalformedPatternOverrideInput {
+                @pattern("^[g-m]+${'$'}")
+                string: PatternString,
+            }
+
+            @pattern("^[a-m]+${'$'}")
+            string PatternString
+        """.trimIndent().asSmithyModel()
```
david-perez, line 53:
```suggestion
        """.asSmithyModel()
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriterTest.kt
```diff
@@ -0,0 +1,280 @@
...
+
+    init {
```
david-perez, line 55: Not sure why this needs to be in an `init` block.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriterTest.kt
```diff
@@ -0,0 +1,280 @@
...
+            writer.rust("""println!("from inside $moduleName");""")
+        }
+    }
+
+//    private fun extraRustFun(writer: RustWriter, moduleName: String) {
+//        writer.rustBlock("pub fn extra()") {
+//            writer.comment("Module $moduleName")
+//            writer.rust("""println!("extra function defined inside $moduleName");""")
+//        }
+//    }
```
david-perez, line 105:
```suggestion
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustCrateInlineModuleComposingWriterTest.kt
```diff
@@ -0,0 +1,280 @@
...
+        testProject.withModule(ServerRustModule.Model) {
+            this.unitTest("test_a") {
+                rust("crate::model::a::hello_world();")
+            }
+        }
+        testProject.compileAndTest()
+    }
+
+    @Test
+    fun `creating nested modules works from different rustWriters`() {
```
david-perez, line 139: Great test.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/eventstream/ServerEventStreamBaseRequirements.kt
```diff
@@ -113,6 +115,8 @@ abstract class ServerEventStreamBaseRequirements : EventStreamTestRequirements<S
             shape.getTrait()!!,
             listOf(),
         ).render(CodegenTarget.SERVER)
-        renderBuilderForShape(writer, codegenContext, shape)
+        // FZ TODO()
+        //TODO()
+        //renderBuilderForShape(writer, codegenContext, shape)
```
david-perez, line 120:
```suggestion
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedMapGeneratorTest.kt
```diff
@@ -53,22 +54,26 @@ class UnconstrainedMapGeneratorTest {
...
         project.withModule(Model) {
-            model.lookup<StructureShape>("test#StructureC").serverRenderWithModelBuilder(model, symbolProvider, this)
+            model.lookup<StructureShape>("test#StructureC").serverRenderWithModelBuilder(project, model, symbolProvider, this)
         }
 
         project.withModule(ServerRustModule.ConstrainedModule) {
             listOf(mapA, mapB).forEach {
-                PubCrateConstrainedMapGenerator(codegenContext, this, it).render()
+                PubCrateConstrainedMapGenerator(codegenContext,
+                    //this,
```
david-perez, line 63:
```suggestion
```

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedCollectionGeneratorTest.kt
```diff
@@ -50,26 +51,29 @@ class UnconstrainedCollectionGeneratorTest {
...
         project.withModule(ServerRustModule.Model) {
-            model.lookup<StructureShape>("test#StructureC").serverRenderWithModelBuilder(model, symbolProvider, this)
+            model.lookup<StructureShape>("test#StructureC").serverRenderWithModelBuilder(project, model, symbolProvider, this)
         }
 
         project.withModule(ServerRustModule.ConstrainedModule) {
             listOf(listA, listB).forEach {
-                PubCrateConstrainedCollectionGenerator(codegenContext, this, it).render()
+                PubCrateConstrainedCollectionGenerator(codegenContext,
+                    //this,
```
david-perez, line 60:
```suggestion
```

david-perez (approved):
There are several places I'm not sure how they work, but they seem to work. Everything looks correct though.

I'm preemptively approving so we can merge this as soon as CI passes.

jdisanti (approved):
Approving the client-side changes.

PR #2258, opened 2023-01-27
Add static stability support to IMDS credentials provider

ysaito1001:
## Motivation and Context
Implements https://github.com/awslabs/smithy-rs/issues/2117

## Description
This PR adds static stability support to the IMDS credentials provider. The goal is to make `ImdsCredentialsProvider` more robust in providing credentials even when IMDS is unreachable. We do this by letting it hold the last retrieved credentials, and these fallback credentials can be served as a backup.

In addition to the failure scenario above, we also need to consider the case where IMDS may return expired credentials. Static stability support specifies that `ImdsCredentialsProvider` should extend the credentials expiry by the calculated amount. The implementation of that is defined in `ImdsCredentialsProvider::maybe_extend_expiration` and references that in [botocore](https://github.com/boto/botocore/blob/develop/botocore/utils.py#L652-L685).

By implementing the above, the Rust SDK will not fail fast in sending a request with expired credentials and allows the target service to make the ultimate decision as to whether the request sent is valid.

## Testing
For the ease of maintaining tests in the future, we'd like to keep them in `aws-config`. Some of the tests require "SDK can send a request...", but we can reinterpret the test specification and bring it closer to the shore of `ImdsCredentialsProvider`. From an `ImdsCredentialsProvider`'s perspective, credentials being returned from `provide_credentials` is equivalent to "SDK can send a request". Furthermore, the Rust SDK, by design, separates credentials providers from credentials caching at the time of writing (01/27/2023); `ImdsCredentialsProvider` itself does not cache credentials the way `LazyCredentialsCache` does. Thus, the test specification like "IMDS must only be called once" can be translated to `ImdsCredentialsProvider` correctly extending expired credentials.

- Expired Credential Test Cases
  - [x] SDK can use IMDS credential provider if first IMDS call returns expired credentials (implemented by `expired_credentials_should_be_extended`).
  - [x] SDK can send request if expired credentials are available (here's no explicit test added for this as it is implicitly handled by two tests for `Refresh Failures`).
  - [x] SDK can perform 3 successive requests with expired credentials. IMDS must only be called once (implicitly realized by `expired_credentials_should_be_extended`).
- Refresh Failures
  - [x] SDK can send a request after a read timeout during a refresh (implemented by `read_timeout_during_credentials_refresh_should_yield_last_retrieved_credentials`).
  - [x] SDK can send a request after receiving a 500 HTTP response from IMDS during a refresh (implemented by `fallback_credentials_should_be_used_when_imds_returns_500_during_credentials_refresh`).
- Logging
  - [x] SDK must log a message when expired credentials are used (implemented by `expired_credentials_should_be_extended`).
- [x] Passed tests in CI

## Checklist

----


CHANGELOG.next.toml
```diff
@@ -56,3 +56,11 @@ let result = cache
 references = ["smithy-rs#2246"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "ysaito1001"
+
+[[aws-sdk-rust]]
+message = """
+Add static stability support to IMDS credentials provider. It does not alter common use cases for the provider, but allows the provider to serve expired credentials in case IMDS is unreachable. This allows requests to be dispatched to a target service with expired credentials. This, in turn, allows the target service to make the ultimate decision as to whether requests sent are valid or not.
+"""
+references = ["smithy-rs#2258"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
```
jdisanti, line 65: Seems like a :tada: to me 😄 
```suggestion
meta = { "breaking" = false, "tada" = true, "bug" = false }
```

jdisanti (approved):
LGTM!

rcoh:
It seems like our logic might be incorrect—I don't think the intention was to always increase the expiry on credentials coming from IMDS (as seems to be the case in the diff). We'll dig into it and get a fix out.

Do you mind verifying if you see ```
        tracing::warn!(
            "Attempting credential expiration extension due to a credential service availability issue. \
            A refresh of these credentials will be attempted again within the next {:.2} minutes.",
            refresh_offset.as_secs_f64() / 60.0,
        );
``` in your logs? just to confirm that this is the culprit?

mustakimali:
Yes I see this log line in the affected pods. In fact this is what lead us to this commit.

ysaito1001:
Thank you for reporting the issue. Opened [a PR](https://github.com/awslabs/smithy-rs/pull/2694) to address it.

rcoh:
this should be fixed in the latest release to aws-sdk-rust

PR #2259, opened 2023-01-28
Make minor fixes to the crate claim workflow

jdisanti:
## Motivation and Context
Noticed some minor things that could be cleaned up in `claim-crate-names.yml` while I was in the area.

----


.github/workflows/claim-crate-names.yml
```diff
@@ -56,6 +56,7 @@ jobs:
   claim:
     name: Claim crate names
     needs:
+    - main-branch-check
```
jdisanti, line 59: Without depending on this, the two jobs race each other. `claim` will _probably_ always lose the race, but better not to chance it.

.github/workflows/claim-crate-names.yml
```diff
@@ -65,8 +66,6 @@ jobs:
         toolchain: ${{ env.rust_version }}
     - name: Checkout smithy-rs
       uses: actions/checkout@v3
-      with:
-        token: ${{ secrets.RELEASE_AUTOMATION_BOT_PAT }}
```
jdisanti, line 69: This isn't needed since we're not pushing to smithy-rs.

.github/workflows/claim-crate-names.yml
```diff
@@ -4,9 +4,9 @@
 # This workflow claims the names of the unpublished crates in this repository
 # on crates.io (by publishing a dummy empty package)
 
-# Allow only one release to run at a time
+# Allow only one crate claim workflow to run at a time
 concurrency:
-  group: release-smithy-rs
+  group: claim-crates-smithy-rs
```
jdisanti, line 9: It wouldn't be so good for this workflow to cancel an ongoing release that happens to be running at 11 UTC 😄

LukeMathWalker (approved):

PR #2262, opened 2023-01-30
fix: time not implemented on wasm

eduardomourar:
## Motivation and Context

Related to https://github.com/awslabs/smithy-rs/issues/2087 and https://github.com/awslabs/aws-sdk-rust/discussions/679#discussioncomment-4357224.

Whenever targeting a platform that the `std::time::SystemTime::now()` or `std::time::Instant::now()` is not supported, it will compile successfully but panic during runtime with the following message:
> time not implemented on this platform

## Description

In order to prevent that from happening, we will use the time library with the [wasm-bindgen feature flag](https://docs.rs/time/latest/time/#feature-flags) enabled. The crate `aws-smithy-types` will have that feature enabled by default and we will consume that function to get the current time for the system.

## Testing
Unit tests already in place
Tested in the browser based on example here: https://github.com/awsdocs/aws-doc-sdk-examples/pull/4325

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/Cargo.toml
+++ b/aws/rust-runtime/aws-config/Cargo.toml
@@ -29,7 +29,7 @@ aws-smithy-json = { path = "../../sdk/build/aws-sdk/sdk/aws-smithy-json" }
 aws-smithy-types = { path = "../../sdk/build/aws-sdk/sdk/aws-smithy-types" }
 aws-types = { path = "../../sdk/build/aws-sdk/sdk/aws-types" }
 hyper = { version = "0.14.12", default-features = false }
-time = { version = "0.3.4", features = ["parsing"] }
+time = { version = "0.3.17", features = ["parsing"] }
 tokio = { version = "1.8.4", features = ["sync"] }
 tracing = { version = "0.1" }
 
--- a/aws/rust-runtime/aws-config/src/sts/util.rs
+++ b/aws/rust-runtime/aws-config/src/sts/util.rs
@@ -6,6 +6,7 @@
 use aws_credential_types::provider::{self, error::CredentialsError};
 use aws_credential_types::Credentials as AwsCredentials;
 use aws_sdk_sts::model::Credentials as StsCredentials;
+use aws_smithy_types::date_time;
 
 use std::convert::TryFrom;
 use std::time::{SystemTime, UNIX_EPOCH};
@@ -46,7 +47,7 @@ pub(crate) fn into_credentials(
 /// provide a name for the session, the provider will choose a name composed of a base + a timestamp,
 /// e.g. `profile-file-provider-123456789`
 pub(crate) fn default_session_name(base: &str) -> String {
-    let now = SystemTime::now()
+    let now = date_time::now()
         .duration_since(UNIX_EPOCH)
         .expect("post epoch");
     format!("{}-{}", base, now.as_millis())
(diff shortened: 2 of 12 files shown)


rcoh:
I think we will probably want to figure out a way to do this without pulling in the wasm-bindgen — `js-sys` is a fairly meaty crate that I don't think we want to pull in for all users in all cases. 

My guess is we'll want to refactor https://github.com/awslabs/smithy-rs/blob/main/aws/rust-runtime/aws-credential-types/src/time_source.rs into a dyn-safe trait like we have to `AsyncSleep`, pull it into `aws-smithy-types`, then make a WASM compatible implementation. 

Probably worth an RFC to get everyone aligned on the design

eduardomourar:
I agree with you with pulling `js-sys` to all consumers is not a good idea (that was my poor implementation because we should be able to set that conditionally based on target). There is a big difference a see between `AsyncSleep` (as you can see it had to be reimplemented for the browser [here](https://github.com/awsdocs/aws-doc-sdk-examples/blob/545d4dc216262157b560378adf5d265781a65bc1/rust_dev_preview/webassembly/src/lib.rs#L79-L87)), because this error can simply break you application during runtime silently and it can happen even when a dependency uses that function from std library.

I don't think we want to every time we need to fetch the current time to always go through the TimeSource (that was passed during instantiation or default), do you? I think this will somewhat become a burden while developing it. In the ideal scenario, we would want the standard library itself to not throw this error. This might happen when there is a standard way of dealing with of the host environment in a WebAssembly context (possibly [WebAssembly component model](https://github.com/WebAssembly/component-model) will solve this).

eduardomourar:
@rcoh , I convert the PR back to draft. Let me know if you still think it is worth going on the route of this PR, otherwise I will park this for now.

jdisanti:
I had a discussion with @rcoh this morning about this. We want to take the following approach:

- Add a `TimeSource` trait to `aws-smithy-types`
- Make `TimeSource` configurable the exact same way as `AsyncSleep`/`sleep_impl` everywhere in smithy-rs
- Refactor all places that consume time to use the `TimeSource` instead
- Have a default `TimeSource` implementation that takes the time from Tokio's `Instant` to make testing easier since we will be able to use Tokio's time travel testing capabilities. This default would be wired up the same way as the default async sleep implementation.

You're welcome to implement this if you want.

On the fallibility of `AsyncSleep`, that browser use case you provided is interesting. Do you know why `Delay` is fallible in `wasm-timer`? From what I could tell from the source code, the timer can disappear, but when would that happen?

eduardomourar:
> I had a discussion with @rcoh this morning about this. We want to take the following approach:
> 
> * Add a `TimeSource` trait to `aws-smithy-types`
> * Make `TimeSource` configurable the exact same way as `AsyncSleep`/`sleep_impl` everywhere in smithy-rs
> * Refactor all places that consume time to use the `TimeSource` instead
> * Have a default `TimeSource` implementation that takes the time from Tokio's `Instant` to make testing easier since we will be able to use Tokio's time travel testing capabilities. This default would be wired up the same way as the default async sleep implementation.
> 
> You're welcome to implement this if you want, and I can provide more details and answer questions if you're interested.

Based on [this comment](https://github.com/tomaka/wasm-timer/issues/21#issuecomment-1380544611) from the wasm-timer maintainer, the approach you guys are proposing seems a good idea. Unfortunately, I will not be able to tackle this anytime soon.

PR #2264, opened 2023-01-30
Prevent test dependencies from leaking into production

rcoh:
## Motivation and Context
- https://github.com/awslabs/aws-sdk-rust/issues/722

## Description
This change fixes the specific issue in https://github.com/awslabs/aws-sdk-rust/issues/722 and goes one step further: It adds `testDependenciesOnly` and `testWriter` mode on `RustWriter`. These two things add mechanisms to prevent this from happening in the future and to automatically convert dependencies created in tests to test scope.

## Testing
- [ ] Specific audit of the codegen diff

## Checklist

----


jdisanti (approved):
Nice!

PR #2276, opened 2023-02-02
Make `OperationExtension` store the absolute shape ID

hlbarber:
## Motivation and Context

Previously, the intention of `OperationExtension` was to store the operation shape ID with `#` replaced with a `.`. This is an opinionated choice to accommodate specific use cases.  Due to a bug the `OperationExtensionExt` was storing the unmodified operation shape ID, but is less opinionated.

## Description

Make `OperationExtension` store the unmodified Operation shape ID.


LukeMathWalker (approved):
This will have to be called out quite loudly in the changelog - users might have alarms matching on those operation names that now won't trigger anymore.

CHANGELOG.next.toml
```diff
@@ -9,4 +9,45 @@
...
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
-# author = "rcoh"
\ No newline at end of file
+# author = "rcoh"
+[[smithy-rs]]
+message = """
+Fix `name` and `absolute` methods on `OperationExtension`.
+
+The older, [now removed](https://github.com/awslabs/smithy-rs/pull/2161), service builder would insert `OperationExtension` into the `http::Response` containing the [absolute shape ID](https://smithy.io/2.0/spec/model.html#grammar-token-smithy-AbsoluteRootShapeId) with the `#` symbol replaced with a `.`. When [reintroduced](https://github.com/awslabs/smithy-rs/pull/2157) into the new service builder machinery the behavior was changed - we now do _not_ perform the replace. This change fixes the documentation and `name`/`abosolute` methods of the `OperationExtension` API to match this new behavior.
```
LukeMathWalker, line 17:
```suggestion
The older, [now removed](https://github.com/awslabs/smithy-rs/pull/2161), service builder would insert `OperationExtension` into the `http::Response` containing the [absolute shape ID](https://smithy.io/2.0/spec/model.html#grammar-token-smithy-AbsoluteRootShapeId) with the `#` symbol replaced with a `.`. When [reintroduced](https://github.com/awslabs/smithy-rs/pull/2157) into the new service builder machinery the behavior was changed - we now do _not_ perform the replace. This change fixes the documentation and `name`/`absolute` methods of the `OperationExtension` API to match this new behavior.
```

CHANGELOG.next.toml
```diff
@@ -9,4 +9,45 @@
...
+
+```rust
+let ext = OperationExtension::new("com.amazonaws.CompleteSnapshot");
+
+// This is expected
+let name = ext.name(); // "CompleteSnapshot"
+let namespace = ext.namespace(); // = "com.amazonaws";
+```
+
+On reintroduction, `OperationExtension` was iniialized, by the `Plugin`, and then used as follows:
```
LukeMathWalker, line 29:
```suggestion
When reintroduced, `OperationExtension` was initialized by the `Plugin` and then used as follows:
```

PR #2295, opened 2023-02-03
Fix CI on main and don't acquire Docker login for forks

jdisanti:
Also fixes a bug where PRs from forks were throwing an exception in `acquire-base-image` because the environment variable existed but was an empty string.


jdisanti:
CI on main test run: https://github.com/awslabs/smithy-rs/actions/runs/4080164490

LukeMathWalker (approved):

PR #2302, opened 2023-02-03
Move converters from constraint violations into `ValidationException` to decorators

david-perez:
This allows "easily" converting to other custom validation exception
shapes by implementing a decorator.

As a simple example, this PR adds a
`CustomValidationExceptionWithReasonDecorator` decorator that is able to
convert into a shape that is very similar to
`smithy.framework#ValidationException`, but that has an additional
`reason` field. The decorator can be enabled via the newly added
`experimentalCustomValidationExceptionWithReasonPleaseDoNotUse` codegen
config flag.

This effectively provides a way for users to use custom validation
exceptions without having to wait for the full implementation of #2053,
provided they're interested enough to write a decorator in a JVM
language. This mechanism is _experimental_ and will be removed once full
support for custom validation exceptions as described in #2053 lands,
hence why the configuration key is strongly worded in this respect.

This commit also ports the mechanism to run codegen integration tests
within Kotlin unit tests for client SDKs to the server. See #1956 for
details. The custom validation exception decorator is tested this way.

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestTools.kt
```diff
@@ -118,11 +119,15 @@ object EventStreamTestTools {
         val symbolProvider = codegenContext.symbolProvider
         val operationShape = model.expectShape(ShapeId.from("test#TestStreamOp")) as OperationShape
         val unionShape = model.expectShape(ShapeId.from("test#TestStream")) as UnionShape
+        val walker = DirectedWalker(model)
 
         val project = TestWorkspace.testProject(symbolProvider)
         val operationSymbol = symbolProvider.toSymbol(operationShape)
         project.withModule(ErrorsModule) {
-            val errors = model.structureShapes.filter { shape -> shape.hasTrait<ErrorTrait>() }
+            val errors = model.serviceShapes
```
david-perez, line 127: @jdisanti What do you think about replacing these tests:

```
ClientEventStreamUnmarshallerGeneratorTest.kt
ServerEventStreamUnmarshallerGeneratorTest.kt
ClientEventStreamMarshallerGeneratorTest.kt
ServerEventStreamMarshallerGeneratorTest.kt
```

by simple `{client,server}IntegrationTests`? (see https://github.com/awslabs/smithy-rs/pull/1956/) They're notoriously hard to get right and a sizable amount of code we could delete.

LukeMathWalker (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/CodegenIntegrationTest.kt
```diff
@@ -0,0 +1,45 @@
...
+    val runtimeConfig: RuntimeConfig? = null,
+    val additionalSettings: ObjectNode = ObjectNode.builder().build(),
+    val overrideTestDir: File? = null,
+    val command: ((Path) -> Unit)? = null,
+)
+
+/**
+ * Run cargo test on a true, end-to-end, codegen product of a given model.
+ */
+fun codegenIntegrationTest(model: Model, params: IntegrationTestParams, invokePlugin: (PluginContext) -> Unit): Path {
```
jdisanti, line 32: Nice!

jdisanti (approved):
Client/core changes LGTM.

PR #2304, opened 2023-02-03
fix: small comment update

DanielBauman88:
The comment talked about the server hanging up while
the response was being read.

The server hanging up while the response is read should not
be a problem. The client will still read whatever the server
has sent before getting to the fin/rst that closes the connection.

This error should only (I think) happen if the server hangs up
before sending a complete response or if the server sends an
invalid response.

## Motivation and Context
Ensuring comments are accurate.

## Description
This is just a comment update, the reason is described above and in the commit message.

## Testing
No testing was done, this is just a comment change.

----


jdisanti (approved):

rcoh (approved):
LGTM!

PR #2306, opened 2023-02-03
Rename Rust server codegen plugins

david-perez:
Rename `RustCodegenServerPlugin` to `RustServerCodegenPlugin`, for
consistency with `RustClientCodegenPlugin`. This is a better name, since
the plugin is named `rust-server-codegen`.

This commit also renames `PythonCodegenServerPlugin` to
`RustServerCodegenPythonPlugin` for the same reasons.

This commit also contains other drive-by improvements made while working
on #2302.

----


hlbarber (approved):

crisidev:
Can we do the same with the Python one as well if it fits?

david-perez:
@crisidev Done in https://github.com/awslabs/smithy-rs/pull/2306/commits/2d5b6bfbc92031a0dda0d3ac12a6b4bfccb5a578; I've amended the commit message and the PR title / description.

crisidev (approved):

PR #2311, opened 2023-02-05
Add unstable flag on tests

thomas-k-cameron:
## Motivation and Context
Add `--cfg aws_sdk_unstable` flag on tests

## Description
It can check whether the implementation of RFC30 is correct and other stuff if we were ever to add feature behind that feature-gate.

## Testing
It was compiling serde when I ran it so I think it's working but I believe we should have a proper tests. Suggestions are much appreciated.

## Checklist

----

### DIFF
--- a/buildSrc/src/main/kotlin/RustBuildTool.kt
+++ b/buildSrc/src/main/kotlin/RustBuildTool.kt
@@ -28,6 +28,7 @@ private fun runCli(
                 }
             }
             .copyTo(action)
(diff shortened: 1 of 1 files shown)


thomas-k-cameron:
@Velfi 
I don't want to rush you or anything, but I really appreciate if you could let me know when you would have the time to review my PR.

Thank you!

Velfi (approved):

PR #2312, opened 2023-02-06
Enforce the same minimum TLS version (1.2) for both TLS backends

LukeMathWalker:
## Motivation and Context
We already require TLS 1.2 or above when using the `rustls` backend. This PR makes sure that the same minimum applies to the `native-tls` backend.

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-client/src/lib.rs
+++ b/rust-runtime/aws-smithy-client/src/lib.rs
@@ -78,7 +78,13 @@ pub mod conns {
 
     #[cfg(feature = "native-tls")]
     pub fn native_tls() -> NativeTls {
-        hyper_tls::HttpsConnector::new()
+        let mut tls = hyper_tls::native_tls::TlsConnector::builder();
+        let tls = tls
+            .min_protocol_version(Some(hyper_tls::native_tls::Protocol::Tlsv12))
+            .build()
+            .unwrap_or_else(|e| panic!("Error while creating TLS connector: {}", e));
+        let mut http = hyper::client::HttpConnector::new();
+        hyper_tls::HttpsConnector::from((http, tls.into()))
     }
 
     #[cfg(feature = "native-tls")]
--- a/rust-runtime/aws-smithy-client/src/lib.rs
+++ b/rust-runtime/aws-smithy-client/src/lib.rs
@@ -72,11 +72,19 @@ pub mod conns {
     }
 
     #[cfg(feature = "rustls")]
+    /// Return a default HTTPS connector backed by the `rustls` crate.
+    ///
+    /// It requires a minimum TLS version of 1.2.
+    /// It allows you to connect to both `http` and `https` URLs.
     pub fn https() -> Https {
         HTTPS_NATIVE_ROOTS.clone()
     }
 
     #[cfg(feature = "native-tls")]
+    /// Return a default HTTPS connector backed by the `hyper_tls` crate.
+    ///
+    /// It requires a minimum TLS version of 1.2.
+    /// It allows you to connect to both `http` and `https` URLs.
     pub fn native_tls() -> NativeTls {
         let mut tls = hyper_tls::native_tls::TlsConnector::builder();
         let tls = tls
--- a/rust-runtime/aws-smithy-client/src/lib.rs
+++ b/rust-runtime/aws-smithy-client/src/lib.rs
@@ -91,7 +91,7 @@ pub mod conns {
             .min_protocol_version(Some(hyper_tls::native_tls::Protocol::Tlsv12))
             .build()
             .unwrap_or_else(|e| panic!("Error while creating TLS connector: {}", e));
-        let mut http = hyper::client::HttpConnector::new();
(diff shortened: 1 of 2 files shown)


rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -78,7 +78,13 @@ pub mod conns {
 
     #[cfg(feature = "native-tls")]
     pub fn native_tls() -> NativeTls {
```
crisidev, line 80: I guess returning an error here is too much of a breaking change right?

crisidev (approved):

rust-runtime/aws-smithy-client/src/lib.rs
LukeMathWalker: It's also unnecessary - that should never fail since we are building the default configuration with known-good settings.

rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -78,7 +78,13 @@ pub mod conns {
...
-        hyper_tls::HttpsConnector::new()
+        let mut tls = hyper_tls::native_tls::TlsConnector::builder();
+        let tls = tls
+            .min_protocol_version(Some(hyper_tls::native_tls::Protocol::Tlsv12))
+            .build()
+            .unwrap_or_else(|e| panic!("Error while creating TLS connector: {}", e));
+        let mut http = hyper::client::HttpConnector::new();
+        hyper_tls::HttpsConnector::from((http, tls.into()))
```
benlcb, line 87: This will still allow `http` connections when supplying an `http` URL. 
Maybe a comment should be added to the `native_tls()` function to make users aware of that?

LukeMathWalker: Added documentation to both `native_tls` and `https` in https://github.com/awslabs/smithy-rs/pull/2312/commits/97b3d33fcd0d8b2041531767f073f38121cf685b - do you think it's clear enough?

Velfi (approved):

rust-runtime/aws-smithy-client/src/lib.rs
benlcb: Yes it's very clear, thank you! :)

rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -72,13 +72,27 @@ pub mod conns {
...
+    /// It allows you to connect to both `http` and `https` URLs.
     pub fn native_tls() -> NativeTls {
-        hyper_tls::HttpsConnector::new()
+        let mut tls = hyper_tls::native_tls::TlsConnector::builder();
+        let tls = tls
+            .min_protocol_version(Some(hyper_tls::native_tls::Protocol::Tlsv12))
+            .build()
+            .unwrap_or_else(|e| panic!("Error while creating TLS connector: {}", e));
+        let http = hyper::client::HttpConnector::new();
+        hyper_tls::HttpsConnector::from((http, tls.into()))
```
david-perez, line 95: Is the `.into()` necessary? Isn't `tls` already of type `TlsConnector`?

rust-runtime/aws-smithy-client/src/lib.rs
```diff
@@ -72,13 +72,27 @@ pub mod conns {
...
+    ///
+    /// It requires a minimum TLS version of 1.2.
+    /// It allows you to connect to both `http` and `https` URLs.
     pub fn native_tls() -> NativeTls {
-        hyper_tls::HttpsConnector::new()
+        let mut tls = hyper_tls::native_tls::TlsConnector::builder();
+        let tls = tls
+            .min_protocol_version(Some(hyper_tls::native_tls::Protocol::Tlsv12))
+            .build()
+            .unwrap_or_else(|e| panic!("Error while creating TLS connector: {}", e));
```
david-perez, line 93:
```suggestion
            .unwrap_or_else(|e| panic!("error creating TLS connector: {}", e));
```

nit: the convention we've been following for panic and `Error` messages is for them to be uncapitalized.

rust-runtime/aws-smithy-client/src/lib.rs
LukeMathWalker: Yes, otherwise you get the following compilation error:
```
error[E0277]: the trait bound `hyper_tls::HttpsConnector<_>: From<(hyper::client::HttpConnector, TlsConnector)>` is not satisfied
  --> aws-smithy-client/src/lib.rs:95:9
   |
95 |         hyper_tls::HttpsConnector::from((http, tls))
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<(hyper::client::HttpConnector, TlsConnector)>` is not implemented for `hyper_tls::HttpsConnector<_>`
   |
   = help: the trait `From<(T, tokio_native_tls::TlsConnector)>` is implemented for `hyper_tls::HttpsConnector<T>`
```

PR #2332, opened 2023-02-08
Update pretty_assertions

Velfi:
## Motivation and Context

https://github.com/awslabs/aws-sdk-rust/pull/734

## Description

`pretty_assertions` v1.0 depends on an unmaintained crate. Upgrading to 1.3 will fix this.

----


jdisanti (approved):

ysaito1001 (approved):
Thank you!

crisidev:
Any reason we don't use `1.3` everywhere? Some entries are `1.3.0`

Velfi:
> Any reason we don't use `1.3` everywhere? Some entries are `1.3.0`

For some reason, I'm under the impression that libraries should use more general version numbers and that tools should use specific versions. I've done some introspection and I'm not sure where the impression comes from.

jjant:
> For some reason, I'm under the impression that libraries should use more general version numbers and that tools should use specific versions. I've done some introspection and I'm not sure where the impression comes from.

I'm not sure what you mean, cargo treats both as equivalent.

Velfi:
> I'm not sure what you mean, cargo treats both as equivalent.

Yeah, If I had any clarity I would have set the tool versions to `=1.3.0`. I'll make them all the same.

crisidev (approved):

Issue #2151, opened 2022-12-30
Investigate the default http connector


rcoh:
Currently the default http connector is `Prebuilt`: https://github.com/awslabs/smithy-rs/blob/bb5bbeb5cff2afb4b3664e95f5c63cae3a524379/aws/rust-runtime/aws-config/src/lib.rs#L463-L466

This means that timeout settings at a service level will be ignored. This should probably be `ConnectorFn` but the implications need to be considered.

ysaito1001:
Fix released in https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2023-03-30.

PR #2333, opened 2023-02-08
test: override connect_timeout and read_timeout in clients

Oliboy50:
## Motivation and Context

A simple PR to reproduce the fact that `connect_timeout` is not properly applied on a DynamoDB client constructed from a SDK config that already has a **real** http_connector set (e.g. when users run `aws_config::load_from_env().await`).

ℹ️ The bug is **also** present for `read_timeout` (but I don't know how to write an integration test for that).
ℹ️ The bug is **not** present for `operation_timeout` and `operation_attempt_timeout` (probably because they are not set on the http_connector).

## Description

Only added an integration test, because I don't know how to fix this bug.

## Checklist

- [ ] waiting for a maintainer to run CI workflows and see that the new test fails

----

### DIFF
--- a/aws/sdk/integration-tests/dynamodb/Cargo.toml
+++ b/aws/sdk/integration-tests/dynamodb/Cargo.toml
@@ -24,6 +24,7 @@ bytes = "1.0.0"
 criterion = { version = "0.4.0" }
 futures-util = "0.3.16"
 http = "0.2.0"
+hyper-rustls = { version = "0.23.2", features = ["webpki-tokio"] }
 serde_json = "1.0.0"
 tokio = { version = "1.8.4", features = ["full", "test-util"] }
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
--- a/aws/sdk/integration-tests/dynamodb/tests/timeouts.rs
+++ b/aws/sdk/integration-tests/dynamodb/tests/timeouts.rs
@@ -10,7 +10,6 @@ use aws_credential_types::provider::SharedCredentialsProvider;
 use aws_credential_types::Credentials;
 use aws_sdk_dynamodb::types::SdkError;
 use aws_smithy_async::rt::sleep::{default_async_sleep, AsyncSleep, Sleep};
-use aws_smithy_client::http_connector::ConnectorSettings;
 use aws_smithy_client::hyper_ext;
 use aws_smithy_client::never::NeverConnector;
 use aws_smithy_types::error::display::DisplayErrorContext;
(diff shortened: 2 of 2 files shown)


rcoh:
yes thanks for adding this test! I had spotted this issue in #2151 . We'll try to get a fix out soon

Velfi:
Hey @Oliboy50, the issue here is not so much that connectors are broken but that this has a bad user-experience.

When any type that implements the `SmithyConnector` trait is passed to a config's `http_connector` setter method, it's converted into an `HttpConnector::Prebuilt` and can't accept any more configuration. Configuring a connector and creating it are the same thing.

However, you can pass an `HttpConnector::ConnectorFn`. This allows the smithy client to create new connectors as needed, and those connectors will respect configuration set at a later "stage" than the connector itself. For example, if you set a prebuilt connector in the SdkConfig, then it wouldn't pick up configuration from the service-specific config.

Here's an example of what creating and passing a `ConnectorFn` looks like:

```rust
let connector_fn =
    |connector_settings: &ConnectorSettings, sleep_impl: Option<Arc<dyn AsyncSleep>>| {
        let mut conn = hyper_ext::Adapter::builder()
            .connector_settings(connector_settings.clone());
        conn.set_sleep_impl(sleep_impl);

        Some(DynConnector::new(
            conn.build(
                hyper_rustls::HttpsConnectorBuilder::new()
                    .with_webpki_roots()
                    .https_or_http()
                    .enable_http1()
                    .build(),
            ),
        ))
    };

let conf = SdkConfig::builder()
    .http_connector(HttpConnector::ConnectorFn(Arc::new(connector_fn)))

    .sleep_impl(default_async_sleep().unwrap())
    .build();

let client = aws_sdk_dynamodb::Client::from_conf(
        aws_sdk_dynamodb::config::Builder::from(&conf)
            .timeout_config(
                TimeoutConfig::builder()
                    .connect_timeout(Duration::from_millis(connect_timeout_value))
                    .build(),
            )
            .build(),
    );
```

Now, the 100ms timeout will work. Let me know if you have any questions about this. I'll try to think of a more friendly API.

Oliboy50:
thanks for the explanation 👍 

actually, as a simple user, I would not even think of building my own `ConnectorFn` as you shown here... nor using any other `http_connector` than the one set by default... 

simply using `aws_config::load_from_env().await` then set different timeout configs at each service level is the way to go IMO

so I'll wait for [the issue linked by @rcoh](https://github.com/awslabs/smithy-rs/issues/2151) to be fixed 👌 

BTW, because of how accurate it is, I really love the `operation_*` timeouts much more than `connect`/`read` timeouts, so I'd love to see [this issue](https://github.com/awslabs/smithy-rs/issues/2086) fixed too ❤️

BTW2, would you want me to close this PR now or keep it open until the issue is fixed?

Velfi:
> would you want me to close this PR now or keep it open until the issue is fixed?

I'll close this PR because we have @rcoh's issue. Thank you for bringing more attention to this issue.

PR #2335, opened 2023-02-08
Add jitter to `LazyCredentialsCache`

ysaito1001:
## Motivation and Context
Addresses https://github.com/awslabs/smithy-rs/issues/2051

## Description
This PR adds jitter to `LazyCredentialsCache`. A jitter provides a mechanism for randomizing the [buffer time](https://github.com/awslabs/smithy-rs/blob/d040f76fd1ae0295b9f4475d994e3dd4f4aa8893/aws/rust-runtime/aws-credential-types/src/cache/lazy_caching.rs#L45) for credentials. This allows credentials with the same expiry to expire at slightly different times, thereby preventing thundering herds. The basic idea is borrowed from [the counterpart](https://github.com/aws/aws-sdk-go-v2/blob/1c05fb6452a1f74985ff6deb7a642b9eb441274a/aws/credential_cache.go#L28-L40) in AWS SDK for Go v2.

Note that adding a jitter does NOT modify the [expires_after](https://github.com/awslabs/smithy-rs/blob/d040f76fd1ae0295b9f4475d994e3dd4f4aa8893/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs#L36) field of credentials. Instead, it "tricks" `ExpiringCache` into thinking that the credentials' buffer time has been jittered. At the end of the day, the [expired](https://github.com/awslabs/smithy-rs/blob/d040f76fd1ae0295b9f4475d994e3dd4f4aa8893/aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs#L101-L103) function behaves as intended.

<p align="center">
Screen Shot 2023-02-08 at 3 47 05 PM
</p>

In case 1, when no jitter is applied, the credentials' `expires_after` and [value's expiration](https://github.com/awslabs/smithy-rs/blob/d040f76fd1ae0295b9f4475d994e3dd4f4aa8893/aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs#L22) in `ExpiringCache` point to the same value. If we subtract the buffer time from that timestamp, that's when the `expired` function starts returning `true`.

In case 2, when a jitter is applied, the credentials' `expires_after` is still the same, and only the value's expiration in `ExpiringCache` is pushed back by the amount of the jitter. If we then subtract the buffer time from that timestamp, we get a new point at which the `expired` function starts returning `true`. The green bar is now shorter, meaning that we have effectively randomized the buffer time with respect to the credentials' `expires_after`.

## Testing
- [x] Unit test `buffer_time_jitter` has been added.
- [x] Passed tests in CI.

## Checklist

----


aws/rust-runtime/aws-credential-types/src/cache/lazy_caching.rs
```diff
@@ -228,6 +240,38 @@ mod builder {
             self
         }
 
+        /// A random percentage by which buffer time is jittered for randomization.
+        ///
+        /// For example, if credentials are expiring in 15 minutes, the buffer time is 10 seconds,
```
jdisanti, line 245: It seems like the Go V2 jitter applies to the whole expiration window rather than to the buffer time. What is your reasoning for applying it to the 10 second buffer time?

ysaito1001: Go V2's term [ExpiryWindow](https://github.com/aws/aws-sdk-go-v2/blob/1c05fb6452a1f74985ff6deb7a642b9eb441274a/aws/credential_cache.go#L16-L23) actually means the buffer time in the Rust SDK.
```
ExpiryWindow will allow the credentials to trigger refreshing prior to the credentials actually expiring.
```
In our case, `ExpiringCache` automatically takes care of buffer time, specifically within the [expired](https://github.com/awslabs/smithy-rs/blob/d040f76fd1ae0295b9f4475d994e3dd4f4aa8893/aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs#L101-L103) function.

jdisanti (approved):

PR #2339, opened 2023-02-09
Simplify `simple.smithy`

david-perez:
`simple.smithy` is currently not that simple. Besides, it serves no
effective purpose: everything that the model exercises is already
exercised in a clearer and more orderly manner in other integration
tests.

This PR makes the model minimal, so that it recovers its purpose as a
simple smoke test. Contributors can also use the model as a scratchpad
for features or bugfixs they're working on.

----


jjant (approved):

crisidev:
This was a simple model we created in the beginning of development to just start out things IIRC. Would not be better to just get rid of it? All it's doing is already covered by other models..

crisidev (commented):
This was a simple model we created in the beginning of development to just start out things IIRC. Would not be better to just get rid of it? All it's doing is already covered by other models..

82marbag:
> This was a simple model we created in the beginning of development to just start out things IIRC. Would not be better to just get rid of it? All it's doing is already covered by other models..

I see two usecases for this to stay:

* it is a way to debug features or wrong behaviors
* any time we need to add a test not covered elsewhere

Sure we can also add it back when needed. It's about the gradle parts that are to be added again.

hlbarber (approved):

crisidev (approved):

PR #2348, opened 2023-02-10
Add crate versions in CHANGELOG

ysaito1001:
## Motivation and Context
Addresses https://github.com/awslabs/aws-sdk-rust/issues/731

## Description

This PR will enable `changelogger` to include a table in `CHANGELOG` showing all crate versions.

The markdown table appears in an expand/collapse section. From a UI standpoint, the section will be rendered last in the release notes. Note that the table contains ~300 rows so when a user expands the section, it will totally occupy the page further down. If the expand/collapse section is placed at the bottom, the table, when expanded downwards, does not get in the way of the current release notes written above the table (I feel it's acceptable to push down the release notes for the previous releases when the table gets expanded).

## Testing
- [x] Added unit tests and an integration test 
- [x] Visually confirmed a rendered `CHANGELOG` to see crate versions are listed in a markdown table

This section
Screen Shot 2023-02-10 at 3 36 44 PM

gets expanded as follows:
Screen Shot 2023-02-10 at 3 37 07 PM

----


Velfi (approved):

jdisanti (approved):
🚀

tools/ci-build/changelogger/src/render.rs
```diff
@@ -592,9 +676,69 @@ author = "rcoh"
...
+        let expected_body = r#"
+**Crate Versions**
+<details>
+<summary>Click to expand to view crate versions...</summary>
+
+Crate|Version|
+-|-|
+aws-config|0.54.1|
+aws-sdk-accessanalyzer|0.24.0|
+aws-smithy-async|0.54.1|
```
LukeMathWalker, line 737: IIRC the recommended formatting is to have a pipe on both ends of the row. I don't think it's strictly necessary though, so feel free to ignore it.

LukeMathWalker (approved):

tools/ci-build/changelogger/src/render.rs
ysaito1001: Thanks for calling it out. I prefer having a pipe on both ends for consistency. Addressed in 7d5524d.

PR #2351, opened 2023-02-13
Ensure services with no operations generate valid Rust

hlbarber:
## Description

- Add test coverage for a service with no operations
- Correct the service builder code to allow for this

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
@@ -10,6 +10,7 @@ import software.amazon.smithy.model.neighbor.Walker
 import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.model.shapes.StringShape
 import software.amazon.smithy.model.traits.PatternTrait
+import software.amazon.smithy.rust.codegen.core.rustlang.Attribute
 import software.amazon.smithy.rust.codegen.core.rustlang.RustReservedWords
 import software.amazon.smithy.rust.codegen.core.rustlang.RustWriter
 import software.amazon.smithy.rust.codegen.core.rustlang.Writable
@@ -207,11 +208,11 @@ class ServerServiceGeneratorV2(
             }
         }
 
-        var missingOperationsUnusedMut = ""
-        var unusedExpected = ""
+        var missingOperationsUnusedMut = writable("")
+        var unusedExpected = writable("")
         if (operations.isEmpty()) {
-            missingOperationsUnusedMut = "##[allow(unused_mut)]"
-            unusedExpected = "##[allow(unused_variables)]"
+            missingOperationsUnusedMut = writable { Attribute.AllowUnusedMut.render(this) }
+            unusedExpected = writable { Attribute.AllowUnusedVariables.render(this) }
         }
 
         rustTemplate(
@@ -225,7 +226,7 @@ class ServerServiceGeneratorV2(
             pub fn build(self) -> Result<$serviceName<#{SmithyHttpServer}::routing::Route<$builderBodyGenericTypeName>>, MissingOperationsError>
             {
                 let router = {
-                    $missingOperationsUnusedMut
+                    #{MissingOperationsUnusedMut:W}
                     let mut $missingOperationsVariableName = std::collections::HashMap::new();
                     #{NullabilityChecks:W}
                     if !$missingOperationsVariableName.is_empty() {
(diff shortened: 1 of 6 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -216,15 +218,14 @@ class ServerServiceGeneratorV2(
             pub fn build(self) -> Result<$serviceName<#{SmithyHttpServer}::routing::Route<$builderBodyGenericTypeName>>, MissingOperationsError>
             {
                 let router = {
-                    use #{SmithyHttpServer}::operation::OperationShape;
+                    ##[allow(unused_mut)]
```
LukeMathWalker, line 221: Can we add this conditionally on the fact that we know there are no operations?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -178,18 +178,20 @@ class ServerServiceGeneratorV2(
 
     private fun buildMethod(): Writable = writable {
         val missingOperationsVariableName = "missing_operation_names"
-        val expectMessageVariableName = "unexpected_error_msg"
+        val expectMessage = "this should never panic since we are supposed to check beforehand that a handler has been registered for this operation; please file a bug report under https://github.com/awslabs/smithy-rs/issues"
```
LukeMathWalker, line 181: We used a variable there to avoid having multiple copies of the same string in the final binary artefact 🤔

hlbarber: I see, I'd be surprised if it didn't deduplicate them. But I'll put a conditional `#[allow(unused_variables)]` instead.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
LukeMathWalker: Loosening lints in all cases can lead to bugs going undetected (because a legitimate lint doesn't trigger). If it's easy enough to be precise, I think we should be.

jjant: > If it's easy enough to be precise, I think we should be.

+1 to this. I don't think we should bend over backwards to have lints be as tight as possible, but if it's easy I'd go for it.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -184,12 +184,14 @@ class ServerServiceGeneratorV2(
             for (operationShape in operations) {
                 val fieldName = builderFieldNames[operationShape]!!
                 val operationZstTypeName = operationStructNames[operationShape]!!
-                rust(
+                rustTemplate(
                     """
                     if self.$fieldName.is_none() {
+                        use #{SmithyHttpServer}::operation::OperationShape;
```
jjant, line 190: Why was this needed? 🤔

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
```diff
@@ -205,6 +207,13 @@ class ServerServiceGeneratorV2(
             }
         }
 
+        var missingOperationsUnusedMut = ""
+        var unusedExpected = ""
+        if (operations.isEmpty()) {
+            missingOperationsUnusedMut = "##[allow(unused_mut)]"
+            unusedExpected = "##[allow(unused_variables)]"
+        }
+
```
jjant, line 216: I think you can use `Attribute.AllowUnusedMut/AllowUnusedVariables.render(writer)` for these.

hlbarber: https://github.com/awslabs/smithy-rs/pull/2351/commits/e0a4d8316c92fa78d12c420ae7f58d89255585e1

LukeMathWalker (approved):

hlbarber:
Following the decision made in https://github.com/awslabs/smithy-rs/pull/1369 we should NOT generate code for this Smithy class of models and instead throw an exception.

Issue #2343, opened 2023-02-10
Nested server structure member shapes targeting simple shapes with `@default` don't work


david-perez:
Minimal reproducer:

```smithy
$version: "1.0"

namespace com.amazonaws.simple

use aws.protocols#restJson1
use smithy.framework#ValidationException

@restJson1
service SimpleService {
    operations: [
        Operation
    ]
}

@http(uri: "/operation", method: "POST")
operation Operation {
    //input: KeyModesOfUse // This works fine.
    input: OperationInput
    errors: [ValidationException]
}

structure OperationInput {
    @required
    KeyModesOfUse: KeyModesOfUse
}

structure KeyModesOfUse {
    Encrypt: PrimitiveBoolean
}
```

The issue is largely within these functions:

https://github.com/awslabs/smithy-rs/blob/bf678fd7d28762e7add375f8485bd7f307b8e317/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderConstraintViolations.kt#L86-L94

https://github.com/awslabs/smithy-rs/blob/bf678fd7d28762e7add375f8485bd7f307b8e317/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/Constraints.kt#L80-L95

Using the example above:

* `OperationInput` is correctly detected as constrained because of `@required` _and_ because its member's target, `KeyModesOfUse`, reaches the `PrimitiveBoolean` constrained shape, since it is _not_ optional ([note that `PrimitiveBoolean` has `@default(false)` in the prelude](https://smithy.io/2.0/spec/model.html#prelude)).

https://github.com/awslabs/smithy-rs/blob/bf678fd7d28762e7add375f8485bd7f307b8e317/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/Constraints.kt#L86-L86

* However, when rendering the constraint violations for the `KeyModesOfUse` structure, _the member is not detected as constrained_: there is no arm in `isDirectlyConstrained` that matches the shape's type (the `BooleanShape` Java class), so we go through the `else` branch.

So this results in the constraint violation for `OperationInput` containing a member pointing to the constraint violation for `KeyModesOfUse`, but the latter is never rendered!

```
error[E0412]: cannot find type `ConstraintViolation` in module `crate::model::key_modes_of_use`
  --> simple/rust-server-codegen/src/input.rs:27:55
   |
[...]

PR #2352, opened 2023-02-13
Nested server structure member shapes targeting simple shapes with default trait

82marbag:
Closes: #2343 

----


CHANGELOG.next.toml
```diff
@@ -10,3 +10,9 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
+
+[[smithy-rs]]
+message = "Mark unconstrained structures with optional _or_ default members."
```
david-perez, line 15: This message is not useful to users at all. It makes no mention of the bug itself.

How about something like "Fix bug whereby nested server structure member shapes targeting simple shapes with `@default` produced Rust code that did not compile"?

CHANGELOG.next.toml
```diff
@@ -10,3 +10,9 @@
...
+references = ["smithy-rs#2352"]
```
david-perez, line 16: Reference the issue too: https://github.com/awslabs/smithy-rs/issues/2343.

david-perez (changes requested):
Fix looks good, but we need tests so that we don't regress on this.

We need _at least_ a regular integration test / Kotlin unit test using `serverIntegrationTest` testing a minimal version of the model in https://github.com/awslabs/smithy-rs/issues/2343.

We should add a test to `ConstraintsTest` to test that indeed `@default` is now not detected as constrained on a simple shape too.

82marbag:
Tests have been added

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ConstraintsTest.kt
```diff
@@ -119,4 +125,9 @@ class ConstraintsTest {
         mapB.canReachConstrainedShape(model, symbolProvider) shouldBe true
         recursiveShape.canReachConstrainedShape(model, symbolProvider) shouldBe true
     }
+
+    @Test
+    fun `it should not consider shapes with the default trait as constrained`() {
+        structWithInnerDefault.canReachConstrainedShape(model, symbolProvider) shouldBe false
```
david-perez, line 131: I would call `isDirectlyConstrained` on `StructWithInnerDefault$PrimitiveBoolean` too, to make the unit tests be more unit-like.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderConstraintViolationsTest.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import org.junit.jupiter.api.Test
+import software.amazon.smithy.rust.codegen.core.testutil.asSmithyModel
+import software.amazon.smithy.rust.codegen.server.smithy.testutil.serverIntegrationTest
+
+class ServerBuilderConstraintViolationsTest {
+
+    @Test
+    fun `it should not generate constraint violations for members with a default value`() {
+        val model = """
```
david-perez, line 16: Nice. I'd leave a comment linking to https://github.com/awslabs/smithy-rs/issues/2343 so the reader can easily find how this test came to be.

In general I like how the Rust compiler adds test for reported issues and links back to the info on GitHub e.g. https://github.com/rust-lang/rust/tree/master/tests/ui/codegen

david-perez (approved):

PR #2364, opened 2023-02-14
Fix version regex.

LukeMathWalker:
It would match `0.0.0-smithy-rs-head` but not `0.54.2`.

----

### DIFF
--- a/tools/ci-build/publisher/src/subcommand/upgrade_runtime_crates_version.rs
+++ b/tools/ci-build/publisher/src/subcommand/upgrade_runtime_crates_version.rs
@@ -28,7 +28,7 @@ pub async fn subcommand_upgrade_runtime_crates_version(
     let fs = Fs::Real;
     let gradle_properties = read_gradle_properties(fs, &args.gradle_properties_path).await?;
     let version_regex =
-        Regex::new(r"(?P<field>smithy\.rs\.runtime\.crate\.version=)(?P<version>\d+\.\d+\.\d+-.*)")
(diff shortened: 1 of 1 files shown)


tools/ci-build/publisher/src/subcommand/upgrade_runtime_crates_version.rs
```diff
@@ -28,7 +28,7 @@ pub async fn subcommand_upgrade_runtime_crates_version(
     let fs = Fs::Real;
     let gradle_properties = read_gradle_properties(fs, &args.gradle_properties_path).await?;
     let version_regex =
-        Regex::new(r"(?P<field>smithy\.rs\.runtime\.crate\.version=)(?P<version>\d+\.\d+\.\d+-.*)")
+        Regex::new(r"(?P<field>smithy\.rs\.runtime\.crate\.version=)(?P<version>\d+\.\d+\.\d+.*)")
```
crisidev, line 31: I am wondering if we should be stricter here and just accept only `-smithy-rs-head` rather than everything after the version..

crisidev (approved):

jjant (approved):

tools/ci-build/publisher/src/subcommand/upgrade_runtime_crates_version.rs
LukeMathWalker: The alternative would be `\d+\.\d+\.\d+(-smithy-rs-head){0,1}\s*$` if my regex-fu holds.

PR #2371, opened 2023-02-14
Fix recursive constraint violations with paths over list and map shapes

david-perez:
There is a widespread assumption throughout the generation of constraint violations that does not hold true all the time, namely, that a recursive constraint violation graph has the same requirements with regards to boxing as the regular shape graph.

Some types corresponding to recursive shapes are boxed to introduce indirection and thus not generate an infinitely recursive type. The algorithm however does not superfluously introduce boxes when the cycle goes through a list shape or a map shape. Why list shapes and map shapes? List shapes and map shapes get rendered in Rust as `Vec<T>` and `HashMap<K, V>`, respectively, they're the only Smithy shapes that "organically" introduce indirection (via a pointer to the heap) in the recursive path. For other recursive paths, we thus have to introduce the indirection artificially ourselves using `Box`. This is done in the `RecursiveShapeBoxer` model transform.

However, the constraint violation graph needs to box types in recursive paths more often. Since we don't collect constraint violations (yet, see #2040), the constraint violation graph never holds `Vec<T>`s or `HashMap<K, V>`s, only simple types. Indeed, the following simple recursive model:

```smithy
union Recursive {
    list: List
}

@length(min: 69)
list List {
    member: Recursive
}
```

has a cycle that goes through a list shape, so no shapes in it need boxing in the regular shape graph. However, the constraint violation graph is infinitely recursive if we don't introduce boxing somewhere:

```rust
pub mod model {
    pub mod list {
        pub enum ConstraintViolation {
            Length(usize),
            Member(
                usize,
                crate::model::recursive::ConstraintViolation,
            ),
        }
    }

    pub mod recursive {
        pub enum ConstraintViolation {
            List(crate::model::list::ConstraintViolation),
        }
    }
}
```

This commit fixes things by making the `RecursiveShapeBoxer` model transform configurable so that the "cycles through lists and maps introduce indirection" assumption can be lifted. This allows a server model transform, `RecursiveConstraintViolationBoxer`, to tag member shapes along recursive paths with a new trait, `ConstraintViolationRustBoxTrait`, that the constraint violation type generation then utilizes to ensure that no infinitely recursive constraint violation types get generated.

For example, for the above model, the generated Rust code would now look like:

```rust
pub mod model {
    pub mod list {
        pub enum ConstraintViolation {
            Length(usize),
            Member(
                usize,
                std::boxed::Box(crate::model::recursive::ConstraintViolation),
            ),
        }
    }

    pub mod recursive {
        pub enum ConstraintViolation {
            List(crate::model::list::ConstraintViolation),
        }
    }
}
```

Likewise, places where constraint violations are handled (like where unconstrained types are converted to constrained types) have been updated to account for the scenario where they now are or need to be boxed.

Parametrized tests have been added to exhaustively test combinations of models exercising recursive paths going through (sparse and non-sparse) list and map shapes, as well as union and structure shapes (`RecursiveConstraintViolationsTest`). These tests even assert that the specific member shapes along the cycles are tagged as expected (`RecursiveConstraintViolationBoxerTest`).

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/RecursiveShapeBoxer.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/RecursiveShapeBoxer.kt
@@ -16,24 +16,41 @@ import software.amazon.smithy.rust.codegen.core.smithy.traits.RustBoxTrait
 import software.amazon.smithy.rust.codegen.core.util.hasTrait
 
 class RecursiveShapeBoxer(
+    /**
+     * A predicate that determines when a cycle in the shape graph contains "indirection". If a cycle contains
+     * indirection, no shape needs to be tagged. What constitutes indirection is up to the caller to decide.
+     */
     private val containsIndirectionPredicate: (Collection<Shape>) -> Boolean = ::containsIndirection,
+    /**
+     * A closure that gets called on one member shape of a cycle that does not contain indirection for "fixing". For
+     * example, the [RustBoxTrait] trait can be used to tag the member shape.
+     */
     private val boxShapeFn: (MemberShape) -> MemberShape = ::addRustBoxTrait,
 ) {
     /**
-     * Transform a model which may contain recursive shapes into a model annotated with [RustBoxTrait].
+     * Transform a model which may contain recursive shapes.
      *
-     * When recursive shapes do NOT go through a `CollectionShape` or a `MapShape` shape, they must be boxed in Rust.
-     * This function will iteratively find loops and add the [RustBoxTrait] trait in a deterministic way until it
-     * reaches a fixed point.
+     * For example, when recursive shapes do NOT go through a `CollectionShape` or a `MapShape` shape, they must be
+     * boxed in Rust. This function will iteratively find cycles and call [boxShapeFn] on a member shape in the
+     * cycle to act on it. This is done in a deterministic way until it reaches a fixed point.
      *
-     * Why `CollectionShape`s and `MapShape`s? Note that `CollectionShape`s get rendered in Rust as `Vec<T>`, and
-     * `MapShape`s as `HashMap<String, T>`; they're the only Smithy shapes that "organically" introduce indirection
-     * (via a pointer to the heap) in the recursive path. For other recursive paths, we thus have to introduce the
-     * indirection artificially ourselves using `Box`.
-     *
-     * This function MUST be deterministic (always choose the same shapes to `Box`). If it is not, that is a bug. Even so
+     * This function MUST be deterministic (always choose the same shapes to fix). If it is not, that is a bug. Even so
      * this function may cause backward compatibility issues in certain pathological cases where a changes to recursive
      * structures cause different members to be boxed. We may need to address these via customizations.
+     *
+     * For example, given the following model,
+     *
+     * ```smithy
+     * namespace com.example
+     *
+     * structure Recursive {
+     *     recursiveStruct: Recursive
+     *     anotherField: Boolean
+     * }
+     * ```
+     *
+     * The `com.example#Recursive$recursiveStruct` member shape is part of a cycle, but the
+     * `com.example#Recursive$anotherField` member shape is not.
      */
     fun transform(model: Model): Model {
         val next = transformInner(model)
@@ -45,8 +62,9 @@ class RecursiveShapeBoxer(
     }
 
     /**
-     * If [model] contains a recursive loop that must be boxed, apply one instance of [RustBoxTrait] return the new model.
-     * If [model] contains no loops, return null.
+     * If [model] contains a recursive loop that must be boxed, return the transformed model resulting form a call to
+     * [boxShapeFn].
+     * If [model] contains no loops, return `null`.
      */
     private fun transformInner(model: Model): Model? {
         // Execute 1 step of the boxing algorithm in the path to reaching a fixed point:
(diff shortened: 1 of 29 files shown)


jjant:
Can you add to the description what the generated `ContraintViolation` structs look like now after the fix?

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RecursiveConstraintViolationsTest.kt
```diff
@@ -0,0 +1,185 @@
+/*
```
LukeMathWalker, line 1: I'm finding this test harness somewhat hard to follow.
I think it'd be easier to decouple the model generation step from the test execution step, persisting (and committing) the generated models with a check to make sure they stay fresh. 

This would require some fiddling with Gradle, which I understand might be unpalatable.

LukeMathWalker (approved):

david-perez:
I updated the commit message and PR description to include the fixed Rust code for the example model.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RecursiveConstraintViolationsTest.kt
david-perez: Not sure how that can be achieved. These test case generators often produce models coupled to values that are used in the assertions (e.g. `shapeIdWithConstraintViolationRustBoxTrait` here).

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/traits/ConstraintViolationRustBoxTrait.kt
```diff
@@ -0,0 +1,24 @@
...
+import software.amazon.smithy.model.traits.Trait
+
+/**
+ * This shape is analogous to [software.amazon.smithy.rust.codegen.core.smithy.traits.RustBoxTrait], but for the
+ * constraint violation graph.
+ *
+ * This is used to handle recursive constraint violations.
+ * See [software.amazon.smithy.rust.codegen.server.smithy.transformers.RecursiveConstraintViolationBoxer].
+ */
+class ConstraintViolationRustBoxTrait : Trait {
```
Velfi, line 19: Why have a separate trait for this?

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RecursiveConstraintViolationsTest.kt
```diff
@@ -0,0 +1,185 @@
...
+
+        private fun recursiveUnionModel(unionPrefix: String = ""): Pair<Model, String> =
+            """
+            $baseModel
+            
+            structure Recursive {
+                attributeValue: ${unionPrefix}AttributeValue
+            }
+
+            // Named `${unionPrefix}AttributeValue` to have reminiscenses of DynamoDB's famous `AttributeValue`.
```
Velfi, line 99:
```suggestion
            // Named `${unionPrefix}AttributeValue` in honor of DynamoDB's famous `AttributeValue`.
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/RecursiveShapeBoxer.kt
```diff
@@ -7,21 +7,29 @@ package software.amazon.smithy.rust.codegen.core.smithy.transformers
...
+) {
     /**
-     * Transform a model which may contain recursive shapes into a model annotated with [RustBoxTrait]
+     * Transform a model which may contain recursive shapes into a model annotated with [RustBoxTrait].
      *
-     * When recursive shapes do NOT go through a List, Map, or Set, they must be boxed in Rust. This function will
-     * iteratively find loops & add the `RustBox` trait in a deterministic way until it reaches a fixed point.
+     * When recursive shapes do NOT go through a `CollectionShape` or a `MapShape` shape, they must be boxed in Rust.
+     * This function will iteratively find loops and add the [RustBoxTrait] trait in a deterministic way until it
+     * reaches a fixed point.
```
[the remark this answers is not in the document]
david-perez: I added an example model.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/transformers/RecursiveShapeBoxer.kt
```diff
@@ -7,21 +7,29 @@ package software.amazon.smithy.rust.codegen.core.smithy.transformers
...
 import software.amazon.smithy.model.shapes.MemberShape
-import software.amazon.smithy.model.shapes.SetShape
 import software.amazon.smithy.model.shapes.Shape
 import software.amazon.smithy.model.transform.ModelTransformer
 import software.amazon.smithy.rust.codegen.core.smithy.traits.RustBoxTrait
 import software.amazon.smithy.rust.codegen.core.util.hasTrait
 
-object RecursiveShapeBoxer {
+class RecursiveShapeBoxer(
+    private val containsIndirectionPredicate: (Collection<Shape>) -> Boolean = ::containsIndirection,
```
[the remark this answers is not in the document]
david-perez: Forgot to add docs on the params. I've now added them.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/traits/ConstraintViolationRustBoxTrait.kt
david-perez: They serve different purposes and we tag different shape sets. I expanded the docs to explicitly call this out.

Velfi (approved):

PR #2372, opened 2023-02-15
Determine modules after running the Rust reserved word symbol provider

jdisanti:
## Motivation and Context
For [RFC-26](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0026_client_crate_organization.md), several types need to be added to modules that are named after operations.

There are actually examples of modules being named after shapes for builder codegen, as can be seen in:
https://github.com/awslabs/smithy-rs/blob/83656ec994581f6e5b9c4701cd0c127d11db31ac/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt#L84-L85

However, for RFC-26, that `symbolProvider.toSymbol` call would actually infinite loop when determining a module for an operation since the operation symbol resolution itself needs a module for the operation symbol, and determining the module requires an operation symbol.

This PR refactors module resolution to the end of the symbol provider chain in a separate `ModuleAttachingSymbolProvider`. This allows names to be resolved in the first half of the chain, and then modules to be resolved based on those names in the second half of the chain.

This PR also eliminates `RustSymbolProvider.toEnumVariantName` by refactoring the code that called that to create fake `MemberShape`s instead, which brings us slightly closer to supporting `EnumShape` rather than `EnumTrait`.

This refactor should be entirely contained to the code generator and should result in no generated code differences.

----


jdisanti:
I'm not liking this approach. Going to close this and work on some additional PRs that should result in something simpler.

PR #2373, opened 2023-02-15
Add caching to the symbol providers

jdisanti:
## Motivation and Context
Smithy's `SymbolProvider` has a `cache` function that wraps a symbol provider chain in a cache layer. I'm curious to see how this impacts codegen performance. This PR applies it to the `RustSymbolProvider`, which required a little extra work due to the type difference.

----


crisidev:
Out of curiosity, are we worried about performances here?

LukeMathWalker:
It's not registering as a significant difference in terms of CI execution time for the generation steps 😢

jdisanti:
> Out of curiosity, are we worried about performances here?

Faster codegen time will result in faster CI, which would be nice, but not worth a ton of investment.

> It's not registering as a significant difference in terms of CI execution time for the generation steps 😢

That's unfortunate. Not worth merging then.

PR #2375, opened 2023-02-15
Disallow `@uniqueItems`-constrained list shapes that reach a map shape

david-perez:
The server SDK codegen generates Rust code that does not compile when a
`@uniqueItems`-constrained list shape reaches a map shape, essentially
because `std::collections::HashMap` does not implement
`std::hash::Hash`.

A ticket with the Smithy team was opened in awslabs/smithy#1567 to
disallow such models.

This commit makes it so that codegen aborts with an informative message
when such models are encountered, instead of going ahead and producing
code that does not compile.

This commit also slightly adjusts the error messages produced when
unsupported constraint traits are found.

----


LukeMathWalker (approved):

mtdowling:
I don't think order matters. Could you use a BTreeMap?

PR #2377, opened 2023-02-15
Remove `toEnumVariantName` from `RustSymbolProvider`

jdisanti:
## Motivation and Context
The `toEnumVariantName` function existed on symbol provider to work around enum definitions not being shapes. In the future when we refactor to use `EnumShape` instead of `EnumTrait`, there will be `MemberShape`s for each enum member. This change incrementally moves us to that future by creating fake `MemberShape`s in the enum generator from the enum definition.

~This change should have no codegen diff.~ The server codegen diff reveals that this improves doc comments emitted by the `UnionGenerator`.

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustReservedWords.kt
```diff
@@ -9,29 +9,30 @@ import software.amazon.smithy.codegen.core.ReservedWordSymbolProvider
...
 import software.amazon.smithy.rust.codegen.core.smithy.RustSymbolProvider
 import software.amazon.smithy.rust.codegen.core.smithy.WrappingSymbolProvider
 import software.amazon.smithy.rust.codegen.core.smithy.generators.UnionGenerator
 import software.amazon.smithy.rust.codegen.core.smithy.renamedFrom
+import software.amazon.smithy.rust.codegen.core.util.hasTrait
 import software.amazon.smithy.rust.codegen.core.util.letIf
-import software.amazon.smithy.rust.codegen.core.util.orNull
-import software.amazon.smithy.rust.codegen.core.util.toPascalCase
 
 class RustReservedWordSymbolProvider(private val base: RustSymbolProvider, private val model: Model) :
```
Velfi, line 25: You may also be able to do this while you're here https://github.com/awslabs/smithy-rs/issues/1766

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
```diff
@@ -71,14 +75,41 @@ abstract class EnumType {
...
+    private val symbolProvider: RustSymbolProvider,
+) {
+    companion object {
+        /**
+         * Return the name of a given `enum` variant. Note that this refers to `enum` in the Smithy context
+         * where enum is a trait that can be applied to [StringShape] and not in the Rust context of an algebraic data type.
+         *
+         * Ordinarily, the symbol provider would determine this name, but the enum trait doesn't allow for this.
+         *
+         * TODO(EnumShape): Remove this function when refactoring to EnumShape.
```
Velfi, line 90:
```suggestion
         * TODO(https://github.com/awslabs/smithy-rs/issues/1700): Remove this function when refactoring to EnumShape.
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
```diff
@@ -198,13 +201,15 @@ private fun RustWriter.renderAsVariant(
...
+        val targetSymbol = symbolProvider.toSymbol(model.expectShape(member.target))
         rust(
             "/// Tries to convert the enum instance into [`$variantName`](#T::$variantName), extracting the inner #D.",
             unionSymbol,
-            memberSymbol,
+            targetSymbol,
         )
         rust("/// Returns `Err(&Self)` if it can't be converted.")
-        rustBlock("pub fn as_$funcNamePart(&self) -> std::result::Result<&#T, &Self>", memberSymbol) {
+        rustBlock("pub fn as_$funcNamePart(&self) -> std::result::Result<&${memberSymbol.rustType().render()}, &Self>") {
```
Velfi, line 212: Why don't we want to render the symbol here? This looks weird.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustReservedWords.kt
jdisanti: There are some bug fixes I plan to make to `RustReservedWordSymbolProvider` today as a separate PR. I can consider it there.

Velfi (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/UnionGenerator.kt
jdisanti: Member shapes are a bit weird. They represent a field on a struct/enum/union, and thus, the name is going to be that field name. Originally, I just changed this to use `targetSymbol` instead of `memberSymbol`, but that resulted in the `Box` type wrapper getting dropped on the floor, because we do boxing on the member shape instead of on the target shape... So this is actually the only correct way to do this particular action.

jdisanti: You'll see a bunch of these if you code search for `rustType().render()`

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/EnumGenerator.kt
jdisanti: Thanks.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustReservedWords.kt
```diff
@@ -55,7 +56,20 @@ class RustReservedWordSymbolProvider(private val base: RustSymbolProvider, priva
...
+            }
+
+            container is EnumShape || container.hasTrait<EnumTrait>() -> when (baseName) {
+                // Self cannot be used as a raw identifier, so we can't use the normal escaping strategy
+                // https://internals.rust-lang.org/t/raw-identifiers-dont-work-for-all-identifiers/9094/4
+                "Self" -> "SelfValue"
+                // Real models won't end in `_` so it's safe to stop here
+                "SelfValue" -> "SelfValue_"
+                // Unknown is used as the name of the variant containing unexpected values
+                "Unknown" -> "UnknownValue"
```
LukeMathWalker, line 69: This is somewhat a spooky action at a distance - perhaps worth putting `Unknown` in a constant and refer to it here and where we generated the unexpected variant?

LukeMathWalker (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustReservedWords.kt
jdisanti: I think this comment applies more to #2382, where this is more heavily refactored. I'll add the comment there.

PR #2389, opened 2023-02-16
Move `AppName` and `PKG_VERSION`

jdisanti:
## Motivation and Context
This PR reorganizes `AppName` and `PKG_VERSION` based on the `enableNewCrateOrganizationScheme` feature flag as part of the new crate organization scheme for [RFC-26](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0026_client_crate_organization.md).

The codegen diff should only reorder things, but otherwise be completely backwards compatible.

----


rcoh (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customize/RequiredCustomizations.kt
```diff
@@ -69,5 +69,13 @@ class RequiredCustomizations : ClientCodegenDecorator {
         rustCrate.withModule(ClientRustModule.Types) {
             pubUseSmithyTypes(codegenContext, codegenContext.model)(this)
         }
+
+        val metaModule = when (codegenContext.settings.codegenConfig.enableNewCrateOrganizationScheme) {
+            true -> ClientRustModule.Meta
+            else -> ClientRustModule.root
+        }
```
rcoh, line 76: could add `CodegenContext.metaModule()` as a extension function since codegenContext _has_ codegen config you can fully contain the logic there

rcoh (approved):
LGTM, but see inline about a way to remove the branching logic from the code

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customize/RequiredCustomizations.kt
jdisanti: Not a bad idea.

LukeMathWalker (approved):

PR #2395, opened 2023-02-20
Add builder symbol/module resolution to symbol providers

jdisanti:
## Motivation and Context
Builder structs, while unmodeled, are integral to inputs, outputs, errors, and modeled types, and this should be reflected in the design in the `codegen-core` module by placing them front and center in the symbol provider interface like we do for operation errors. Individual code generators can replace builder codegen, but referring to a builder should be centralized.

This PR also allows the `ModuleProvider` to provide modules for builders, which is required for implementing [RFC-26](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0026_client_crate_organization.md).

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientRustModule.kt
```diff
@@ -120,6 +126,17 @@ object OldModuleSchemeClientModuleProvider : ModuleProvider {
...
     ): RustModule.LeafModule = ClientRustModule.Error
+
+    override fun moduleForBuilder(context: ModuleProviderContext, shape: Shape, symbol: Symbol): RustModule.LeafModule {
+        val builderNamespace = RustReservedWords.escapeIfNeeded(symbol.name.toSnakeCase())
+        return RustModule.new(
+            builderNamespace,
+            visibility = Visibility.PUBLIC,
+            parent = symbol.module(),
+            inline = true,
+            documentation = "See [${symbol.name}](${symbol.module().fullyQualifiedPath()}::${symbol.name}).",
```
Velfi, line 137:
```suggestion
            documentation = "See [`${symbol.name}`](${symbol.module().fullyQualifiedPath()}::${symbol.name}).",
```

Velfi (approved):

ysaito1001 (approved):

hlbarber (approved):

PR #2415, opened 2023-02-24
Split the `types` and `builders` modules into per-shape private modules

jdisanti:
## Motivation and Context
This PR, behind the `enableNewCrateOrganizationScheme` feature flag, splits up the `types` and `builders` modules so that there is one private module per structure/enum/union shape, with the generated types being re-exported into the correct module. This significantly reduces file sizes which should improve IDE support (see aws-sdk-rust#716).

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientCodegenVisitor.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientCodegenVisitor.kt
@@ -6,6 +6,7 @@
 package software.amazon.smithy.rust.codegen.client.smithy
 
 import software.amazon.smithy.build.PluginContext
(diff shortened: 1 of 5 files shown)


jdisanti:
Examining EC2 with these changes (with the feature enabled), there are only two modules larger than a megabyte:
```
aws-sdk/sdk/ec2$ find . -name '*.rs' -type f -size +1M
./src/client.rs
./src/xml_deser.rs
```

The `client.rs` clocks in at 2.1 MB for EC2, and it is arguably the most important to have under a megabyte since it's the entry point for all use of the SDK, so we'll have to figure out how to split it up as well. The vast majority of that file is documentation currently. I'm not sure if writing the docs to separate files and including them into the file would work for IDEs or make things worse.

`xml_deser.rs` is entirely private, so I'm not so concerned about that one.

I think this PR is a good first step though by splitting up `types` and `builders`.

Velfi (approved):

PR #2416, opened 2023-02-24
Abridge EC2 field docs on the fluent client

jdisanti:
## Motivation and Context
This reduces the size of the `client.rs` for EC2 from 2.1 MB down to 1.7 MB (for https://github.com/awslabs/aws-sdk-rust/issues/716) by keeping only the first sentence in each input field's documentation in the operation summary.

This is still not below the 1 MB cutoff, but is a 20% improvement.

----


ysaito1001 (approved):
Like the test cases!

PR #2418, opened 2023-02-25
Make module docs customizable and update mod docs for crate reorg

jdisanti:
## Motivation and Context
This is needed to correct the module docs behind the `enableNewCrateOrganizationScheme` feature flag, although it can also be used to allow for codegen decorators to change or augment module documentation (e.g., maybe the AWS SDK adds additional examples/links/etc), and for the service model and codegen config to be incorporated into those docs. 

----


Velfi (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustModule.kt
```diff
@@ -25,19 +26,38 @@ import software.amazon.smithy.rust.codegen.core.util.toSnakeCase
...
         software.amazon.smithy.rust.codegen.core.smithy.UnconstrainedModule
     val ConstrainedModule =
         software.amazon.smithy.rust.codegen.core.smithy.ConstrainedModule
 }
 
+class ServerModuleDocProvider : ModuleDocProvider {
+    override fun docs(module: RustModule.LeafModule): String? = when (module) {
+        ServerRustModule.Error -> "All error types that operations can return. Documentation on these types is copied from the model."
+        ServerRustModule.Operation -> "All operations that this crate can perform."
+        // TODO(ServerTeam): Document this module (I don't have context)
```
hlbarber, line 48: This module has a decent amount of documentation. We render it in [ServerOperationShapeGenerator.kt](https://github.com/awslabs/smithy-rs/blob/main/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerOperationShapeGenerator.kt) via a `writer.rustTemplate`. This is similar to our root docs found in [ServerServiceGenerator.kt](https://github.com/awslabs/smithy-rs/blob/main/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt).

For this reason I think perhaps the `docs` method should return a `Writable?` and the dependencies templated in it should be added to the `dev-dependencies` rather than the normal `dependencies`.

jdisanti: Was actually just refactoring it to be that way in https://github.com/awslabs/smithy-rs/pull/2432 in order to fix CI.

hlbarber (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustModule.kt
hlbarber: One additional thing to note during doc generation refactors is that I think it'd be nicer to generate the docs in the actual module rather than on the `mod` statement for the following reasons:

- Huge docs won't crowd their parent module
- They are more discoverable for developers
- Intra-doc links are easier

jdisanti: Yeah, I definitely agree that would be better. Given how our module writing system currently works, though, this will require a significant refactor since it will require buffering and separating writables to output into different logical sections. Definitely something to strive towards in the long term, but will require a considerable amount of work.

PR #2423, opened 2023-03-01
Move using-native-tls-instead-of-rustls to smithy-rs

ysaito1001:
## Motivation and Context
Moves a `native-tls/rustls`-related test, `using-native-tls-instead-of-rustls` from `aws-doc-sdk-examples` to `smithy-rs`.

## Description
The test was previously in `aws-doc-sdk-examples` but it can be more useful to be in `smithy-rs` so it can catch a test failure early prior to cutting a release.

(`using-native-tls-instead-of-rustls` has been deleted from `aws-doc-sdk-examples` as part of [this PR](https://github.com/awsdocs/aws-doc-sdk-examples/pull/4469))

----

### DIFF
--- a/aws/sdk/integration-tests/using-native-tls-instead-of-rustls/Cargo.toml
+++ b/aws/sdk/integration-tests/using-native-tls-instead-of-rustls/Cargo.toml
@@ -17,8 +17,4 @@ aws-config = { path = "../../build/aws-sdk/sdk/aws-config", default-features = f
 aws-sdk-s3 = { path = "../../build/aws-sdk/sdk/s3", default-features = false, features = [
   "native-tls",
 ] }
-# aws-sdk-sts is the same as aws-sdk-s3
-aws-sdk-sts = { path = "../../build/aws-sdk/sdk/sts", default-features = false, features = [
-  "native-tls",
-] }
-tokio = { version = "1.20.1", features = ["full"] }
+tokio = { version = "1.20.1", features = ["rt", "macros"] }
--- a/aws/sdk/integration-tests/using-native-tls-instead-of-rustls/Cargo.toml
+++ b/aws/sdk/integration-tests/using-native-tls-instead-of-rustls/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "using-native-tls-instead-of-rustls"
+version = "0.1.0"
+authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>"]
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dev-dependencies]
+# aws-config pulls in rustls and several other things by default. We have to disable defaults in order to use native-tls
+# and then manually bring the other defaults back
+aws-config = { path = "../../build/aws-sdk/sdk/aws-config", default-features = false, features = [
+  "native-tls",
+  "rt-tokio",
+] }
+# aws-sdk-s3 brings in rustls by default so we disable that in order to use native-tls only
+aws-sdk-s3 = { path = "../../build/aws-sdk/sdk/s3", default-features = false, features = [
+  "native-tls",
+] }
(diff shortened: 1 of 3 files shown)


aws/sdk/integration-tests/using-native-tls-instead-of-rustls/tests/no-rustls-in-dependency.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// The SDK defaults to using RusTLS by default but you can also use [`native_tls`](https://github.com/sfackler/rust-native-tls)
+/// which will choose a TLS implementation appropriate for your platform. This test looks much like
+/// any other. Activating and deactivating `features` in your app's `Cargo.toml` is all that's needed.
+
+async fn list_buckets() -> Result<(), aws_sdk_s3::Error> {
+    let shared_config = aws_config::load_from_env().await;
+
+    let s3_config = aws_sdk_s3::Config::from(&shared_config);
+    let client = aws_sdk_s3::Client::from_conf(s3_config);
```
Velfi, line 14: imo, let's simplify this

```suggestion
    let sdk_config = aws_config::load_from_env().await;
    let client = aws_sdk_s3::Client::new(&sdk_config);
```

aws/sdk/integration-tests/using-native-tls-instead-of-rustls/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+  "rt-tokio",
+] }
+# aws-sdk-s3 brings in rustls by default so we disable that in order to use native-tls only
+aws-sdk-s3 = { path = "../../build/aws-sdk/sdk/s3", default-features = false, features = [
+  "native-tls",
+] }
+# aws-sdk-sts is the same as aws-sdk-s3
+aws-sdk-sts = { path = "../../build/aws-sdk/sdk/sts", default-features = false, features = [
+  "native-tls",
+] }
```
Velfi, line 23: Is bringing this in required? It's not referenced in the test.

aws/sdk/integration-tests/using-native-tls-instead-of-rustls/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+tokio = { version = "1.20.1", features = ["full"] }
```
Velfi, line 24: I think we can change this to just bring in the test utils, although it may also require the `rt` feature

```suggestion
tokio = { version = "1.20.1", features = ["test-util"] }
```

aws/sdk/integration-tests/using-native-tls-instead-of-rustls/tests/no-rustls-in-dependency.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    let resp = client.list_buckets().send().await?;
+
+    for bucket in resp.buckets().unwrap_or_default() {
+        println!("bucket: {:?}", bucket.name().unwrap_or_default())
+    }
```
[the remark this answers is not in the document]
ysaito1001: Good point. We can eliminate a call to `println!` for the test.

aws/sdk/integration-tests/using-native-tls-instead-of-rustls/Cargo.toml
ysaito1001: Thanks for bringing it up. I think we can remove it. Because the test looks at a dependency tree, I originally thought we needed this as part of the test requirement even though `sts` wasn't used at runtime.

aws/sdk/integration-tests/using-native-tls-instead-of-rustls/Cargo.toml
ysaito1001: Yes. Addressed in 86896e5. The minimal feature set required seems to be `rt` and `macros`.

Velfi (approved):

PR #2425, opened 2023-03-02
Split generated code into multiple files

jdisanti:
## Motivation and Context
This PR breaks up the generated fluent client and the operation/http serde functions across multiple files to address https://github.com/awslabs/aws-sdk-rust/issues/716.

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/ProtocolFunctions.kt
```diff
@@ -0,0 +1,127 @@
...
+import software.amazon.smithy.rust.codegen.core.rustlang.RustWriter
+import software.amazon.smithy.rust.codegen.core.smithy.CodegenContext
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.core.smithy.RustSymbolProvider
+import software.amazon.smithy.rust.codegen.core.smithy.contextName
+import software.amazon.smithy.rust.codegen.core.util.toSnakeCase
+
+typealias ProtocolFnWritable = RustWriter.(String) -> Unit
+
+class ProtocolFunctions(
```
Velfi, line 25: consider adding docs and possibly a usage example

Velfi (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -89,139 +93,147 @@ class FluentClientGenerator(
...
+                        pub fn $fnName(&self) -> #{FluentBuilder}${generics.inst} {
+                            #{FluentBuilder}::new(self.handle.clone())
+                        }
+                        """,
+                        "FluentBuilder" to operation.fluentBuilderType(codegenContext, symbolProvider),
+                    )
+                }
             }
         }
     }
```
Velfi, line 239: You didn't touch this in your PR, but on line 249 below, where it says:
```kt
rust(
    """
    /// Fluent builder constructing a request to `${operationSymbol.name}`.
    ///
    """,
)
```

We have an opportunity to use the `docs` writer and turn the `${operationSymbol.name}` into a link to more docs.

LukeMathWalker (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/ProtocolFunctions.kt
```diff
@@ -0,0 +1,162 @@
...
+        else -> shape.contextName(serviceShape).toSnakeCase()
+    }
+    return when (shape) {
+        is MemberShape -> shape.memberName.toSnakeCase()
+        is DocumentShape -> "document"
+        else -> containerName
+    }
+}
+
+fun RustSymbolProvider.lensName(
```
Velfi, line 153: What is a lensName? Why the underscore prefix?

Velfi:
Hey @jdisanti. I looked through the generated docs after setting `"enableNewCrateOrganizationScheme": true` and I have a few comments:

## What are fluent builders?

Screenshot 2023-03-07 at 9 58 38 AM

The docs for the client module itself aren't very helpful. I think it would benefit users if we explained some of the philosophy and language behind the client, like:

- How we use the builder pattern, and what a fluent builder is
- What an operation is
  - How operations and the fluent builders are related
  - Operation has no docs
    Screenshot 2023-03-07 at 10 01 26 AM
- The fact that operations can be customized
- We're generating code with the module declarations and use statements all the way at the bottom. It would be more idiomatic to generate these at the top.
    Screenshot 2023-03-07 at 10 09 52 AM
- I ran `cargo check` on S3 crate and it's complaining about a redundant import of `aws_smithy_types::error::metadata::ProvideErrorMetadata` in the error module.
  - Running `cargo check --all` in the generated S3 crate reveals several compiler errors related to missing modules.
  - Some generated modules (`lens.rs`, `event_stream_serde.rs`, `error_meta.rs`, and more) are missing their copyright headers.
  - We use the terms `ser`, `deser`, and `serde` in different places. I think we should settle on either `ser`+`deser` OR `serde`

I know some of this stuff will be fixed in separate PRs, but hopefully there's something here that's helpful in the context of this PR.

rcoh:
from EC2:
a quick hack shows that `fluent_builders` is 3.8MB on its own so we'll need to split that up as well
```
1.7M    ./src/output.rs
2.0M    ./src/error.rs
2.8M    ./src/xml_deser.rs
5.4M    ./src/client.rs
5.8M    ./src/model.rs
7.0M    ./src/input.rs
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/ProtocolFunctions.kt
```diff
@@ -0,0 +1,162 @@
...
+ * This class should be used for generating all inline functions from a protocol code generator, as it is
+ * responsible for correctly organizing and code splitting those functions into smaller modules.
+ */
+class ProtocolFunctions(
+    private val codegenContext: CodegenContext,
+) {
+    companion object {
+        private val serDeModule = RustModule.pubCrate("protocol_serde")
+
+        fun commonFn(fnName: String, block: ProtocolFnWritable): RuntimeType =
```
rcoh, line 40: `crossOperationFn`? it wasn't immediately obvious to me what "common" meant here

jdisanti:
> from EC2:
> a quick hack shows that fluent_builders is 3.8MB on its own so we'll need to split that up as well

Are you looking at generated code with the crate reorg flag disabled? I already split up fluent builders in another PR.

jdisanti:
> I ran cargo check on S3 crate and it's complaining about a redundant import of aws_smithy_types::error::metadata::ProvideErrorMetadata in the error module. 

All of this stuff is addressed in https://github.com/awslabs/smithy-rs/pull/2433

jdisanti:
> We use the terms ser, deser, and serde in different places. I think we should settle on either ser+deser OR serde

I'm fine with using `ser` and `de`/`deser`, but `serde` will get confused with the `serde` crate.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/ProtocolFunctions.kt
jdisanti: I don't know much about lenses... just preserving functionality here.

rcoh (approved):
couple of naming suggestions, otherwise LGTM

jdisanti:
> We're generating code with the module declarations and use statements all the way at the bottom. It would be more idiomatic to generate these at the top.

I agree it would be more idiomatic, but it would make the code generator much more complex also since we would have to buffer contents from certain writables and emit them later. That seems like it would be a considerable refactor with little added value for customers.

jdisanti:
> Some generated modules (lens.rs, event_stream_serde.rs, error_meta.rs, and more) are missing their copyright headers.

They have the `// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.` comment at the top, which is correct. Copyright headers haven't ever been emitted into the generated code.

jdisanti:
@Velfi - Your feedback on the docs is great. I'm going to apply it to https://github.com/awslabs/smithy-rs/pull/2418 since I think it's more applicable to that PR than this one.

Velfi (approved):

PR #2427, opened 2023-03-02
Implement Python unions

crisidev:
## Motivation and Context
Continue work on #1367.

## Description
Implement support for the `union` shape in the Python server.


unexge:
Wonder if we can leverage [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html) here.

So we generate the following:
```rust
struct Square;

struct Circle;

// :type: typing.Union[Square, Circle]
enum Shape {
    Square(Square),
    Circle(Circle),
}

#[pyo3(name="Square")]
struct PySquare(Square);

#[pyo3(name="Circle")]
struct PyCircle(Circle);

impl ToPyObject for Shape {
    fn to_object(&self, py: Python<'_>) -> PyObject {
        match self {
            Shape::Square(s) => PySquare(s),
            Shape::Circle(c) => PyCircle(c),
        }
    }
}

impl FromPyObject<'_> for Shape {
    fn extract(obj: &PyAny) -> PyResult<Self>{
        if let Ok(s) = obj.extract::<PySquare> {
            Ok(Shape::Square(s.0))
        } else if let Ok(c) = obj.extract::<PyCircle> {
            Ok(Shape::Circle(c.0))
        } else {
            Err("invalid shape")
        }
    }
}
```

and users can write:
```python
Shape = Square | Circle

def stringfy(shape: Shape) -> str:
    if isinstance(shape, Square):
        return "■"
    return "●"
```

I believe with Python 3.10 they can even write: 
```python
def stringfy(shape: Shape) -> str:
    match shape:
        case Square:
            return "■"
        case Circle:
            return "●"
        case _:
            raise NotReachable()
```
🤯

crisidev:
> Wonder if we can leverage [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html) here.

> So we generate the following:

Unfortunately that is not enough as it will not cover nested unions like

```smithy
union Something {
    member_a: AnotherUnion
    member_b: Integer
}

union AnotherUnion {
    member_a: Integer
    member_b: String
}
```

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerCodegenVisitor.kt
```diff
@@ -81,7 +88,7 @@ class PythonServerCodegenVisitor(
             rustSymbolProviderConfig: RustSymbolProviderConfig,
             publicConstrainedTypes: Boolean,
             includeConstraintShapeProvider: Boolean,
-        ) = RustServerCodegenPythonPlugin.baseSymbolProvider(settings, model, serviceShape, rustSymbolProviderConfig, publicConstrainedTypes)
+        ) = RustServerCodegenPythonPlugin.baseSymbolProvider(settings, model, serviceShape, rustSymbolProviderConfig, publicConstrainedTypes, includeConstraintShapeProvider)
```
crisidev, line 91: @unexge

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/RustServerCodegenPythonPlugin.kt
```diff
@@ -74,6 +73,7 @@ class RustServerCodegenPythonPlugin : SmithyBuildPlugin {
             serviceShape: ServiceShape,
             rustSymbolProviderConfig: RustSymbolProviderConfig,
             constrainedTypes: Boolean = true,
+            includeConstrainedShapeProvider: Boolean = true,
```
crisidev, line 76: @unexge

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/RustServerCodegenPythonPlugin.kt
```diff
@@ -82,7 +82,7 @@ class RustServerCodegenPythonPlugin : SmithyBuildPlugin {
                 // In the Python server project, this is only done to generate constrained types for simple shapes (e.g.
                 // a `string` shape with the `length` trait), but these always remain `pub(crate)`.
                 .let {
-                    if (constrainedTypes) ConstrainedShapeSymbolProvider(it, serviceShape, constrainedTypes) else it
+                    if (includeConstrainedShapeProvider) PythonConstrainedShapeSymbolProvider(it, serviceShape, constrainedTypes) else it
```
crisidev, line 85: @unexge

codegen-server-test/python/model/pokemon.smithy
```diff
@@ -41,3 +45,40 @@ structure StreamPokemonRadioOutput {
...
+    //@length(min: 1, max: 100)
+    inner: String
+}
+
+union InnerUnion {
+    a: Integer
+    b: ConstrainedStuff
+}
+
+union MyUnion {
```
unexge, line 78: Should we move those types to an end-to-end test? We will probably get rid of Python specific Pokémon model when we support event streaming tho, so it is also fine to keep them here for now.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerCodegenVisitor.kt
```diff
@@ -82,7 +89,7 @@ class PythonServerCodegenVisitor(
             publicConstrainedTypes: Boolean,
             includeConstraintShapeProvider: Boolean,
             codegenDecorator: ServerCodegenDecorator,
-        ) = RustServerCodegenPythonPlugin.baseSymbolProvider(settings, model, serviceShape, rustSymbolProviderConfig, publicConstrainedTypes, codegenDecorator)
+        ) = RustServerCodegenPythonPlugin.baseSymbolProvider(settings, model, serviceShape, rustSymbolProviderConfig, publicConstrainedTypes, includeConstraintShapeProvider, codegenDecorator)
```
unexge, line 92: Unrelated to this PR, this bug caused because we forgot to pass some variable right? Should we have some linter rules for unused variables to prevent those kinds of issues?

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerSymbolProvider.kt
```diff
@@ -31,10 +31,36 @@ import software.amazon.smithy.rust.codegen.core.smithy.traits.SyntheticOutputTra
...
+private fun toPythonByteStreamSymbolOrOriginal(model: Model, config: RustSymbolProviderConfig, initial: Symbol, shape: Shape): Symbol {
+    if (shape !is MemberShape) {
+        return initial
+    }
+
+    val target = model.expectShape(shape.target)
+    val container = model.expectShape(shape.container)
+
+    // We are only targeting non-synthetic inputs and outputs.
+    if (!container.hasTrait<SyntheticOutputTrait>() && !container.hasTrait<SyntheticInputTrait>()) {
```
unexge, line 48: I think the comment and the code compete here. Shouldn't we use 
```kotlin
if (container.hasTrait<SyntheticOutputTrait>() ||
container.hasTrait<SyntheticInputTrait>()) {
   return initial
}
```
if the comment is correct?

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerUnionGenerator.kt
```diff
@@ -0,0 +1,202 @@
...
+import software.amazon.smithy.rust.codegen.core.smithy.expectRustMetadata
+import software.amazon.smithy.rust.codegen.core.smithy.generators.UnionGenerator
+import software.amazon.smithy.rust.codegen.core.smithy.rustType
+import software.amazon.smithy.rust.codegen.core.util.isTargetUnit
+import software.amazon.smithy.rust.codegen.core.util.toSnakeCase
+import software.amazon.smithy.rust.codegen.server.python.smithy.PythonServerCargoDependency
+import software.amazon.smithy.rust.codegen.server.python.smithy.pythonType
+import software.amazon.smithy.rust.codegen.server.python.smithy.renderAsDocstring
+
+private fun RustType.renderPy() {
```
unexge, line 31: Seems like unused

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerUnionGenerator.kt
```diff
@@ -0,0 +1,202 @@
...
+                rust("/// :rtype bool:")
+                rustBlock("pub fn is_$funcNamePart(&self) -> bool") {
+                    rust("self.0.is_$funcNamePart()")
+                }
+            }
+            if (renderUnknownVariant) {
+                rust("/// Returns true if the union instance is the `Unknown` variant.")
+                rust("/// :rtype bool:")
+                rustBlock("pub fn is_unknown(&self) -> bool") {
+                    rust("matches!(self.0, Self::Unknown)")
```
unexge, line 82: I think this should be `self.0.is_unknown()` because the wrapper type won't have any `Unknown` variant

unexge (approved):
Thank you for doing this 🎉

codegen-server-test/python/model/pokemon.smithy
crisidev: I am really thinking to just get rid of this piece of model until we have an operation with union that makes sense. We have unions covered by `misc` and `ebs` right now. Thoughts?

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerCodegenVisitor.kt
crisidev: This was a change that came with a commit done by a customer and it was just not compiling, so I don't think there is a need for anything here.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/PythonServerSymbolProvider.kt
crisidev: Naaa, the comment is wrong here, this is the right thing to do. I'll fix the comment.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerUnionGenerator.kt
crisidev: Yup, will remove.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerUnionGenerator.kt
crisidev: Good catch.

codegen-server-test/python/model/pokemon.smithy
unexge: Yeah, that makes sense. We'll have unions with event streaming, so I'm fine to remove them from Pokémon service for now

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerUnionGenerator.kt
```diff
@@ -0,0 +1,203 @@
...
+
+    private fun renderPyObjectConverters() {
+        writer.rustBlockTemplate("impl #{pyo3}::IntoPy<#{pyo3}::PyObject> for ${unionSymbol.name}", "pyo3" to pyo3) {
+            rustBlockTemplate("fn into_py(self, py: #{pyo3}::Python<'_>) -> #{pyo3}::PyObject", "pyo3" to pyo3) {
+                rustBlock("match self") {
+                    sortedMembers.forEach { member ->
+                        val variantName = symbolProvider.toMemberName(member)
+                        rust("${unionSymbol.name}::$variantName(variant) => variant.into_py(py),")
+                    }
+                }
```
unexge, line 97: I think this should be just:
```suggestion
                rust("PyUnionMarker${unionSymbol.name}(self).into_py(py)")
```

because we're returning directly variants here, and we don't get the `is_{variant}`, `as_{variant}` methods. We should just return the wrapper type

unexge: Addressed with https://github.com/awslabs/smithy-rs/pull/2427/commits/1fc0a92837c14aebb39f9666004e207d0cac8a6b

Velfi (approved):

PR #2430, opened 2023-03-03
Use a `&'static str` instead of `String` in `AwsJsonRouter`

jjant:
## Description
I noticed they key for the routes map could be a `&'static str` so I made this change to save some allocations.

----


hlbarber (approved):

hlbarber:
The allocation only occurs once at startup but this seems reasonable change. We'll probably need to tweak the code generation slightly too?

PR #2431, opened 2023-03-03
Python: Add tests and fix the issues with `Timestamp` and `ByteStream`

unexge:
## Motivation and Context
This PR adds tests for `Timestamp` and `ByteStream` types and fixes the discovered issues, namely:
- `Timestamp`: It was causing a compilation error previously if you add a `@required Timestamp` to your model
- `ByteStream`: There was a bug in the async next method (`__anext__`), which was causing Rust future to panic

## Testing
I've added some tests for the issues

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
@@ -38,7 +38,6 @@ import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
 import software.amazon.smithy.rust.codegen.core.smithy.RustSymbolProvider
 import software.amazon.smithy.rust.codegen.core.smithy.customize.NamedCustomization
 import software.amazon.smithy.rust.codegen.core.smithy.customize.Section
-import software.amazon.smithy.rust.codegen.core.smithy.generators.TypeConversionGenerator
 import software.amazon.smithy.rust.codegen.core.smithy.generators.UnionGenerator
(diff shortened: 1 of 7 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -407,11 +405,7 @@ class JsonSerializerGenerator(
                 val timestampFormat =
                     httpBindingResolver.timestampFormat(context.shape, HttpLocation.DOCUMENT, EPOCH_SECONDS)
                 val timestampFormatType = RuntimeType.timestampFormat(runtimeConfig, timestampFormat)
-                rustTemplate(
-                    "$writer.date_time(${value.asRef()}#{ConvertInto:W}, #{FormatType})?;",
-                    "FormatType" to timestampFormatType,
-                    "ConvertInto" to typeConversionGenerator.convertViaInto(target),
-                )
+                rust("$writer.date_time(${value.asRef()}, #T)?;", timestampFormatType)
```
unexge, line 408: I've reverted the changes made in https://github.com/awslabs/smithy-rs/commit/5455c4e9cce288f57367176d8e1c64892842167f#diff-41136e956b536306dde14246dbe368a0450d6849773cc95f14f72685f0e3b022 to fix `Timestamp` conversions errors from Python counterpart. We're now using a `Deref` hack to accomplish that because current solution wasn't enough for some cases. `Deref` hack is explained in https://github.com/awslabs/smithy-rs/blob/843d1549d018d21c2e1c8b5fe9f91ff27a09a81d/rust-runtime/aws-smithy-http-server-python/src/types.rs#L8-L16

PR #2436, opened 2023-03-07
Remove uncessary type parameter and constraints from `Upgrade` service

hlbarber:
## Motivation and Context

The type parameter `B` was not only redundant but also removed a degree of freedom from the API.

## Description

- Remove type parameter `B` from `Upgrade` and related code.
- Remove repeated constraint on the upgraded service.

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -239,3 +239,9 @@ message = "`SdkError` variants can now be constructed for easier unit testing."
 references = ["smithy-rs#2428", "smithy-rs#2208"]
 meta = { "breaking" = false, "tada" = true, "bug" = false, "target" = "client" }
 author = "jdisanti"
+
+[[smithy-rs]]
+message = "Remove unnecessary type parameter `B` from `Upgrade` service."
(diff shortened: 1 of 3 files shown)


jjant (commented):
- Why was this type variable there in the first place?
- Will we be removing any of the other phantom type variables in `Upgrade` later?

You might want to update [these docs](https://awslabs.github.io/smithy-rs/design/server/anatomy.html) (they seem to have an outdated version of `Upgrade` with only 3 tyvars).

hlbarber:
> Why was this type variable there in the first place?

Just a mistake - I'm guessing I was mucking around with types and it was required in a previous iteration.

> Will we be removing any of the other phantom type variables in `Upgrade` later?

It's a possibility, this PR was motivated by a specific excess constraint. A more general investigation can be performed and might reveal some extra redundancy.

> You might want to update [these docs](https://awslabs.github.io/smithy-rs/design/server/anatomy.html) (they seem to have an outdated version of Upgrade with only 3 tyvars).

This was intentional to simplify the documentation. There do seems to be some references to `B` here that can be removed though - I'll update it.

PR #2441, opened 2023-03-08
Use a bespoke version of `Either`

hlbarber:
## Motivation and Context

https://docs.rs/tower/latest/tower/util/enum.Either.html

> Both services must be of the same request, response, and error types.

This is not the case, the error types are not required to be the same - as seen in the actual [`Service` implementation](https://docs.rs/tower/latest/tower/util/enum.Either.html#impl-Service%3CRequest%3E). The `tower::util::Either` instead takes either error and converts it to `Box<dyn Error>` which is not the intended behavior and conflicts with the contract of Smithy which wants either `Infallible` in the case of a HTTP middleware and `Op::Error` in the case of a model middleware.

## Description

- Roll bespoke version of `Either` which preserves error.
- Add `Plugin` implementation for `Plugin`.
- Use bespoke `Either` for `FilterByOperationName`.


rust-runtime/aws-smithy-http-server/src/plugin/mod.rs
```diff
@@ -119,6 +119,7 @@
 //!
 
 mod closure;
+pub mod either;
```
hlbarber, line 122: I think I should probably pull `Either` into this module actually, rather than `pub mod`

hlbarber: Yeah that could work except all these `Plugin`s are kinda `util`s though?

- `Stack`
- `FilterByOperationName`
- `OperationNameFn`

PR #2443, opened 2023-03-08
Remove `{Service}::layer` in favour of `PluginPipeline::http_layer`

hlbarber:
## Motivation and Context

The current schematic for middleware for the `smithy-rs` is given in the [Applying Middleware](https://github.com/awslabs/smithy-rs/blob/9d31b6c4a7ba17d38919ddc8f931725f102efb2f/design/src/server/middleware.md#applying-middleware) section of the [Middleware document](https://github.com/awslabs/smithy-rs/blob/main/design/src/server/middleware.md).

Within the document 4 middleware groups are described. Both A and B were present prior to the [service builder refactor](https://github.com/awslabs/smithy-rs/blob/main/design/src/rfcs/rfc0020_service_builder.md) while C and D were added during it. The group B survived the refactor mainly because the lack of infrastructure around C at the time and an attempt to smooth out the breaking changes.

B and C are however nearly identical positions to apply middleware. The API for applying to C is more flexible in the sense that we can use `Plugin`s and `Operation::layer` to apply individualized `Layer`s to each operation - however a blanket application of one `Layer` to every operation is just a special case of this. The only real difference is that B is applied _after_ we `Route::new` and `C` is applied just _before_ we `Route::new`.

**Pros**

- Remove redundant middleware surface
- Simplify types
- The order of `Plugin`s and `Layer`s are now explicit in `PluginPipeline` method chain

**Cons**

- Breaking change and everyone is using `{Service}::layer`.

## Description

- Remove middleware position B by removing `{Service}::layer` and `{Service}::boxed`.
- Add `PluginPipeline::http_layer` which plays the role of `{Service}::layer`.

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGeneratorV2.kt
@@ -52,8 +52,6 @@ class ServerServiceGeneratorV2(
     private val service = codegenContext.serviceShape
     private val serviceName = service.id.name.toPascalCase()
     private val builderName = "${serviceName}Builder"
-    private val builderPluginGenericTypeName = "Plugin"
-    private val builderBodyGenericTypeName = "Body"
 
     /** Calculate all `operationShape`s contained within the `ServiceShape`. */
     private val index = TopDownIndex.of(codegenContext.model)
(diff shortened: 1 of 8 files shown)


Velfi:

david-perez:
Now that we have [`HttpPlugins`](https://docs.rs/aws-smithy-http-server/latest/aws_smithy_http_server/plugin/struct.HttpPlugins.html), I think this patch is no longer needed. `HttpPlugins` applies layers to operations in C position.

david-perez:
Note though that we don't want to remove middleware in B position, but rather have it apply _around_ routing so that it has access to routing errors. So we'll keep `{Service}::layer` around, we just want different semantics. See https://github.com/awslabs/smithy-rs/issues/3074.

PR #2450, opened 2023-03-13
Fix test failure in `no-default-features`

ysaito1001:
## Motivation and Context
This PR fixes a test failure from `test_clients_from_sdk_config`. The test failure has prevented us from releasing a new version of `aws-sdk-rust`.

## Description
It is a negative test expecting a panic, but it failed because it panicked for a different reason, i.e. `no default sleep implementation available` instead of `No HTTP connector was available...`. The fix will add a default async sleep to a service config to ensure that the panic occurs for the expected reason.

----

### DIFF
--- a/aws/sdk/integration-tests/no-default-features/Cargo.toml
+++ b/aws/sdk/integration-tests/no-default-features/Cargo.toml
@@ -16,6 +16,7 @@ publish = false
 [dev-dependencies]
 aws-config = { path = "../../build/aws-sdk/sdk/aws-config", default-features = false }
 aws-sdk-s3 = { path = "../../build/aws-sdk/sdk/s3", default-features = false }
+aws-smithy-async = { path = "../../build/aws-sdk/sdk/aws-smithy-async", features = ["rt-tokio"] }
 futures = "0.3.25"
 tokio = { version = "1.8.4", features = ["full", "test-util"] }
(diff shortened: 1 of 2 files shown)


aws/sdk/integration-tests/no-default-features/Cargo.toml
```diff
@@ -16,6 +16,7 @@ publish = false
 [dev-dependencies]
 aws-config = { path = "../../build/aws-sdk/sdk/aws-config", default-features = false }
 aws-sdk-s3 = { path = "../../build/aws-sdk/sdk/s3", default-features = false }
+aws-smithy-async = { path = "../../build/aws-sdk/sdk/aws-smithy-async", features = ["rt-tokio"] }
```
jdisanti, line 19: Since this is specifically testing no default features, we should just create a no-op sleep impl for the test rather than pull in one of the default features. We don't actually need to sleep for anything in the test.

ysaito1001: Ah, got it. Will add a no-op sleep locally in the test and remove the `rt-tokio` feature. Thanks for pointing it out.

ysaito1001: Addressed in 45b91ab.

jdisanti (approved):

PR #2457, opened 2023-03-14
Remove `PollError` from server type parameters

hlbarber:
## Motivation and Context

https://github.com/awslabs/smithy-rs/issues/2444

## Description

Remove `PollError` from the implementation and documentation.


rust-runtime/aws-smithy-http-server/src/operation/upgrade.rs
```diff
@@ -259,7 +253,16 @@ where
     fn upgrade(self, plugin: &Pl) -> Route<B> {
         let mapped = plugin.map(self);
         let layer = Stack::new(UpgradeLayer::new(), mapped.layer);
-        Route::new(layer.layer(mapped.inner))
+        let svc = layer
+            .layer(mapped.inner)
+            // NOTE: `map_result` has an annoying `Error: From<Self::Error>` bound on it
+            // to ignore this we follow this combinator with a `map_err`
+            .map_result(|res| match res {
```
hlbarber, line 260: We're essentially discarding the `Op::Error` from the `poll_ready` here.

rust-runtime/aws-smithy-http-server/src/operation/upgrade.rs
```diff
@@ -131,15 +132,15 @@ where
...
 
     // Must be able to convert extensions
     Exts: FromParts<P>,
 
     // The signature of the inner service is correct
-    S: Service<(Op::Input, Exts), Response = Op::Output, Error = OperationError<OpError, PollError>>,
+    S: Service<(Op::Input, Exts), Response = Op::Output, Error = Op::Error> + Clone,
 {
-    type Output = Result<http::Response<crate::body::BoxBody>, PollError>;
+    type Output = Result<http::Response<crate::body::BoxBody>, Infallible>;
```
hlbarber, line 143: This is the crux of the PR. The operation service `poll_ready` errors no longer propagate through the `Upgrade` middleware.

hlbarber: Note that we have a constraint above in the stack that asserts that `PollError = Infallible` anyway.

LukeMathWalker: What can we expect the cascading effect to be for users that wrote their own layers?

hlbarber: Good question, for this change to have a knock on effect the customer middleware the customer would have to have the following setup:

1. `Operation::from_service` to specify a `PollError` which is returned by `poll_ready`.
2. Write a HTTP layer which expects `Err = PollError` and apply it in [position C](https://github.com/awslabs/smithy-rs/blob/main/design/src/server/middleware.md#c-operation-specific-http-middleware).
3. Then throw away the `PollError` via a subsequent HTTP layer, again in position C, to adhere to the "constraint above in the stack" - `Error = Infallible` [found here](https://github.com/awslabs/smithy-rs/pull/2457/files#diff-449bb7ee2b877947f8c7df593f64e6dbb8187fa3aec9963bd21ee6674553efd0R247).

I find it unlikely that such a setup would exist.

In the future I think passing data from the handler/"model world" up into the HTTP middleware should be done solely by the use of `IntoResponse`/`IntoParts`(?), dual to how `FromRequest`/`FromParts` works.

jjant, line 143: Can't this just be:
```suggestion
    type Output = http::Response<crate::body::BoxBody>;
```

hlbarber: It can be but then we'd have to `map` in the `Service::call` method which would leak into the `type Future = `, which we don't want.

PR #2465, opened 2023-03-15
Add: runtime plugin placeholder code to Smithy Orchestrator

Velfi:
update: add placeholder impl of runtime plugins
update: separate client and operation interceptors

I thought I had already added this in the other PR but these changes must have become lost somehow.

----


rust-runtime/aws-smithy-runtime-api/src/interceptors.rs
```diff
@@ -565,31 +608,59 @@ impl<ModReq, TxReq, TxRes, ModRes> Interceptors<ModReq, TxReq, TxRes, ModRes> {
...
+        cfg: &mut ConfigBag,
+    ) -> Result<(), InterceptorError> {
+        for interceptor in self.client_interceptors.iter_mut() {
+            interceptor.read_before_execution(context, cfg)?;
+        }
+        Ok(())
+    }
 
-    pub fn read_before_execution(
+    pub fn operation_read_before_execution(
```
jdisanti, line 646: Nit: I think we should make a macro to generate all these functions since they're completely identical except for the function names...

jdisanti (approved):

rust-runtime/aws-smithy-runtime-api/src/runtime_plugin.rs
```diff
@@ -0,0 +1,85 @@
...
+{
+    fn from(t: T) -> Self {
+        Box::new(t) as _
+    }
+}
+
+#[derive(Default)]
+pub struct RuntimePlugins {
+    client_plugins: Vec<Box<dyn RuntimePlugin>>,
+    operation_plugins: Vec<Box<dyn RuntimePlugin>>,
```
ysaito1001, line 26: `Vec` would be a good start. We may later want to change them to `VecDeque` if we need to prepend plugins.

aws/sdk/integration-tests/aws-smithy-runtime-test/src/endpoints.rs
```diff
@@ -0,0 +1,104 @@
...
+#[derive(Debug)]
+pub struct GetObjectEndpointOrc {}
+
+impl GetObjectEndpointOrc {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl RuntimePlugin for GetObjectEndpointOrc {
```
ysaito1001, line 20: Haven't fully decided yet but we may need to define a separate struct `GetObjectEndpointPlugin` and `impl RuntimePlugin` on that.

ysaito1001 (approved):

PR #2469, opened 2023-03-15
Add docs and more color to configbag

rcoh:
## Motivation and Context
Adds more docs & add `Persist` trait to config bag

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime-api/src/config_bag.rs
+++ b/rust-runtime/aws-smithy-runtime-api/src/config_bag.rs
@@ -5,11 +5,10 @@
 
 //! Layered Configuration Bag Structure
 //!
-//! [`ConfigBag`] and [`FrozenConfigBag`] are the two representations of a layered configuration structure
+//! [`config_bag::ConfigBag`] and [`config_bag::FrozenConfigBag`] are the two representations of a layered configuration structure
 //! with the following properties:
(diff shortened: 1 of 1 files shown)


Velfi (approved):

rust-runtime/aws-smithy-runtime/src/config_bag.rs
```diff
@@ -40,13 +61,40 @@ struct Layer {
...
+    /// fn add_more_config(bag: &mut ConfigBag) { /* ... */ }
+    /// use aws_smithy_runtime::ConfigBag;
+    /// let bag = ConfigBag::base().with_fn("first layer", |_| { /* add a property */ });
+    /// let mut bag = bag.add_layer("second layer");
+    /// add_more_config(&mut bag);
+    /// let bag = bag.freeze();
+    /// ```
+    pub fn add_layer(&self, name: &'static str) -> ConfigBag {
+        self.with_fn(name, no_op)
+    }
```
Velfi, line 90: `/// let mut bag = bag.add_layer("second layer");` doesn't compile because `ConfigBag` doesn't have an `add_layer` method. Did you mean to define this on `ConfigBag` instead of `FrozenConfigBag`?

rcoh: ah, should have a `bag.freeze().add_layer`—we can also add `add_layer` to `ConfigBag` as well but moving self

PR #2473, opened 2023-03-16
Add support for omitting session token in canonical requests

martinjlowm:
## Motivation and Context
Some services require the session token to be omitted from the canonical request when calculating the signature. 

## Description
This PR extends the signing settings with `omit_session_token` such that the security token only gets added after a signature has been calculated as is required by some services: https://docs.aws.amazon.com/general/latest/gr/create-signed-request.html#add-signature-to-request

The security token is now added after the signature is calculated, if the new setting states so.

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
+++ b/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
@@ -5,6 +5,7 @@
 
 use crate::date_time::{format_date, format_date_time};
 use crate::http_request::error::CanonicalRequestError;
+use crate::http_request::settings::SessionTokenMode;
 use crate::http_request::settings::UriPathNormalizationMode;
 use crate::http_request::sign::SignableRequest;
 use crate::http_request::uri_path_normalization::normalize_uri_path;
@@ -122,6 +123,8 @@ impl<'a> CanonicalRequest<'a> {
     /// - If `settings.percent_encoding_mode` specifies double encoding, `%` in the URL will be re-encoded as `%25`
     /// - If `settings.payload_checksum_kind` is XAmzSha256, add a x-amz-content-sha256 with the body
     ///   checksum. This is the same checksum used as the "payload_hash" in the canonical request
+    /// - If `settings.session_token_mode` specifies X-Amz-Security-Token to be
+    ///   included before calculating the signature, add it, otherwise omit it.
     /// - `settings.signature_location` determines where the signature will be placed in a request,
     ///   and also alters the kinds of signing values that go along with it in the request.
     pub(super) fn from<'b>(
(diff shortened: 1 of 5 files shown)


aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
```diff
@@ -170,10 +179,19 @@ impl<'a> CanonicalRequest<'a> {
...
             }),
         };
+
+        if params.settings.session_token_mode == SessionTokenMode::Exclude {
+            if let Some(security_token) = params.security_token {
+                let mut sec_header = HeaderValue::from_str(security_token)?;
+                sec_header.set_sensitive(true);
+                canonical_headers.insert(header::X_AMZ_SECURITY_TOKEN, sec_header);
+            }
+        }
```
rcoh, line 193: can we handle this in `Self::headers` instead of mutating the headers later?

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
```diff
@@ -726,6 +747,28 @@ mod tests {
...
+
+        let creq = CanonicalRequest::from(&req, &signing_params).unwrap();
+        assert_eq!(
+            creq.values.signed_headers().as_str(),
+            "host;x-amz-date;x-amz-security-token"
+        );
+
+        signing_params.settings.session_token_mode = SessionTokenMode::Exclude;
+        let creq = CanonicalRequest::from(&req, &signing_params).unwrap();
+        assert_eq!(creq.values.signed_headers().as_str(), "host;x-amz-date");
```
rcoh, line 769: should probably also assert on the headers in the creq

rcoh:
this is great! have you had a chance to make sure that these requests are actually accepted by the upstream services?

rcoh (commented):
couple of small comments, looking good!

martinjlowm:
> this is great! have you had a chance to make sure that these requests are actually accepted by the upstream services?

I sure have :) I tested it against IoT Core (code from the linked issue by Mathias) in order to subscribe to an IoT topic using temporary credentials. That was with query params - not sure about services using just headers, but seeing that they are passed into the function that builds the params, I presume it works for them as well.

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
martinjlowm: I just realized this part is redundant. It's already in the `Self::headers` function. The header is added to the canonical headers in all cases.

I'll remove the mode condition.

rcoh:
LGTM! Looks like there's one clippy lint that needs to be fixed prior to merge

rcoh (approved):

martinjlowm:
@rcoh - just noticed it needed a rustfmt run as well. Feel free to cancel the previous workflow.

rcoh:
sorry one last thing! will you add a changelog entry? That will also ensure you get credit in the release notes :-)

martinjlowm:
> sorry one last thing! will you add a changelog entry? That will also ensure you get credit in the release notes :-)

Done :)

PR #2486, opened 2023-03-22
Respect the `ShapeVisitor` methods - `operation_shape.rs`

hlbarber:
## Motivation and Context

## Description

- Call `ServerOperationGenerator` from the `ShapeVisitor::operationShape` method rather than `ShapeVisitor::serviceShape` method.
- Migrate `operation_shape.rs` to the `ModuleDocProvider` method of documentation.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerRustModule.kt
```diff
@@ -44,23 +49,74 @@ object ServerRustModule {
...
-                ServerRustModule.ConstrainedModule -> "Constrained types for constrained shapes."
-                else -> TODO("Document this module: $module")
-            },
+class ServerModuleDocProvider(private val codegenContext: ServerCodegenContext) : ModuleDocProvider {
+    override fun docsWriter(module: RustModule.LeafModule): Writable? {
+        val strDoc: (String) -> Writable = { str -> writable { docs(escape(str)) } }
+        return when (module) {
+            ServerRustModule.Error -> strDoc("All error types that operations can return. Documentation on these types is copied from the model.")
+            ServerRustModule.Operation -> strDoc("All operations that this crate can perform.")
+            // TODO(ServerTeam): Document this module (I don't have context)
```
david-perez, line 58:
```suggestion
```

PR #2490, opened 2023-03-23
Refactor `ShapeVisitor` methods - `PythonServerOperationErrorGenerator`

hlbarber:
## Motivation and Context

## Description

Move `renderCombinedErrors` logic into `ShapeVisitor::operationShape`.


crisidev:
Thanks Harry!! <3

crisidev (approved):

PR #2493, opened 2023-03-23
Refactor `ShapeVisitor` methods - `ServerServiceGeneratorV2`

hlbarber:
## Motivation and Context

## Description

- Remove `PythonServerServiceGenerator` and move `renderExtras` (`PythonApplicationGenerator`) to `ShapeVisitor::serviceShape`.
- Rename `ServerServiceGenerator` to `ServerRootGenerator`.
- Rename `ServerServiceGeneratorV2` to `ServerServiceGenerator` call it from `ShapeVisitor::serviceShape`.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
hlbarber, line 1: Despite the messy diff, this is just renamed `ServerServiceGeneratorV2`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerRootGenerator.kt
hlbarber, line 1: This is `ServerServiceGenerator` renamed and the call to `ServerServiceGeneratorV2` removed.

david-perez (approved):
Thanks!

PR #2504, opened 2023-03-28
Remove `ProtocolTraitImplGenerator` from `codegen-core`

jdisanti:
## Motivation and Context
The `ProtocolTraitImplGenerator` was originally an implementation detail for client protocol codegen that ended up getting copied into `codegen-core`. With the work to move towards an orchestrator approach for client codegen, this interface needs to change for clients specifically, and thus, it needs to be removed from core in order to not affect the server codegen.

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/HttpBoundProtocolGenerator.kt
```diff
@@ -62,10 +61,10 @@ class HttpBoundProtocolGenerator(
     HttpBoundProtocolTraitImplGenerator(codegenContext, protocol),
 )
 
-class HttpBoundProtocolTraitImplGenerator(
+open class HttpBoundProtocolTraitImplGenerator(
```
82marbag, line 64: I'm assuming it's `open` for the work in progress you mention in the overview?

jdisanti: In `codegen-client`, one of the unit tests inherit and override one of the methods in this class. Previously, the unit tests just implemented the interface.

Velfi (approved):

PR #2513, opened 2023-03-29
Make AppName configurable from within ConfigLoader

ysaito1001:
## Motivation and Context
Fixes https://github.com/awslabs/smithy-rs/issues/2511

## Description
This small PR adds a setter for `app_name` in `ConfigLoader`. Prior to the commit, in a typical config chain of `ConfigLoader` -> `SdkConfig` -> service config, the field was only settable from `SdkConfig` onward.

## Testing
- [x] Added unit test
- [x] Passed tests in CI  

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -52,3 +52,9 @@ message = "The `enableNewCrateOrganizationScheme` codegen flag has been removed.
 references = ["smithy-rs#2507"]
 meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client" }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Make AppName configurable from within `ConfigLoader`."
+references = ["smithy-rs#2513"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
+author = "ysaito1001"
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -54,7 +54,7 @@ meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client" 
 author = "jdisanti"
 
 [[aws-sdk-rust]]
-message = "Make AppName configurable from within `ConfigLoader`."
+message = "`AppName` is now configurable from within `ConfigLoader`."
 references = ["smithy-rs#2513"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "ysaito1001"
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -56,5 +56,5 @@ author = "jdisanti"
 [[aws-sdk-rust]]
 message = "`AppName` is now configurable from within `ConfigLoader`."
 references = ["smithy-rs#2513"]
-meta = { "breaking" = false, "tada" = false, "bug" = false }
+meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "ysaito1001"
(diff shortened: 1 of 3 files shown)


CHANGELOG.next.toml
```diff
@@ -52,3 +52,9 @@ message = "The `enableNewCrateOrganizationScheme` codegen flag has been removed.
 references = ["smithy-rs#2507"]
 meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client" }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Make AppName configurable from within `ConfigLoader`."
```
jdisanti, line 57:
```suggestion
message = "`AppName` is now configurable from within `ConfigLoader`."
```

CHANGELOG.next.toml
```diff
@@ -52,3 +52,9 @@ message = "The `enableNewCrateOrganizationScheme` codegen flag has been removed.
...
+references = ["smithy-rs#2513"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
```
jdisanti, line 59:
```suggestion
meta = { "breaking" = false, "tada" = false, "bug" = true }
```

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -344,6 +344,24 @@ mod loader {
             self
         }
 
+        /// Override the name of the app used to build [`SdkConfig`](aws_types::SdkConfig).
```
jdisanti, line 347:
```suggestion
        /// Override the name of the app used to build [`SdkConfig`](aws_types::SdkConfig).
        ///
```

jdisanti (approved):

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -344,6 +344,25 @@ mod loader {
...
+        ///
+        /// This _optional_ name is used to identify the application in the user agent that
+        /// gets sent along with requests.
```
pose, line 350: Why is it call `AppName` and not `UserAgent`?

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -344,6 +344,25 @@ mod loader {
...
+        /// ```no_run
+        /// # async fn create_config() {
+        /// use aws_config::AppName;
+        /// let config = aws_config::from_env()
+        ///     .app_name(AppName::new("my-app-name").expect("valid app name"))
+        ///     .load().await;
+        /// # }
+        /// ```
+        pub fn app_name(mut self, app_name: AppName) -> Self {
+            self.app_name = Some(app_name);
```
pose, line 362: Do we need to add any sort of validation for this being non-empty string or is that handled by `AppName`?

pose (changes requested):

aws/rust-runtime/aws-config/src/lib.rs
rcoh: it's not the `UserAgent` — it eventually gets composed as part of the user agent. But the user agent itself has far more in it than the app name (platform, version, etc.)

rcoh (approved):

aws/rust-runtime/aws-config/src/lib.rs
ysaito1001: Thanks for pointing it out! It's good to guard against that. Addressed in b296de9.

pose (approved):

PR #2516, opened 2023-03-30
Remove `ignoreUnsupportedConstraintTraits` from integration test models

david-perez:
Now that constraint traits are supported in server SDKs (with some
corner case caveats, see
https://github.com/awslabs/smithy-rs/issues/1401), we can remove
`ignoreUnsupportedConstraintTraits` from the codegen config of the
integration test models.

----


david-perez:
Worth implementing a check in `ValidateUnsupportedConstraints` that aborts codegen if the flag is enabled but all constraint traits in the model are supported. I'll do this in a separate PR, since it's a breaking change that will have to wait to be merged in until the next merge window.

hlbarber (approved):

jjant (approved):

PR #2518, opened 2023-03-30
Add semver-checks CI job

rcoh:
## Motivation and Context
It is important to avoid making semver breaking changes by accident

## Description
This PR invokes cargo semver checks against main and the codegen result of this pull request. If you are intentionally making a breaking change, you can add the `breaking-change` label to your pull request to suppress this check.

## Testing
- [x] tested breaking change label
- [x] tested semver breaking changes 

----


.github/workflows/ci.yml
```diff
@@ -159,6 +159,41 @@ jobs:
       with:
         action: ${{ matrix.test.action }}
 
+  semver-checks:
```
jdisanti, line 162: I think this will break `ci-main.yml` and `ci-merge-queue.yml`, which both invoke `ci.yml`, but won't have `context.issue.number` or `github.event.pull_request.base.sha`. Did you mean for this to be added to `ci-pr.yml` and `ci-pr-forks.yml`?

tools/ci-scripts/codegen-diff/semver-checks.py
```diff
@@ -0,0 +1,64 @@
...
+    for path in os.listdir():
+        eprint(f'checking {path}...', end='')
+        if get_cmd_status(f'git cat-file -e base:{sdk_directory}/{path}/Cargo.toml') == 0:
+            (status, out, err) = get_cmd_output(f'cargo semver-checks check-release '
+                                    f'--baseline-rev {BASE_BRANCH} '
+                                    # in order to get semver-checks to work with publish-false crates, need to specify
+                                    # package and manifest path explicitly
+                                    f'--manifest-path {path}/Cargo.toml '
+                                    f'-p {path} '
+                                    f'--release-type patch', check=False, quiet=True)
```
obi1kenobi, line 46: Are you sure you want `--release-type patch` here? When you publish things that require minor versions (e.g. deprecations), this will still return `failed` _even if_ you correctly bumped the minor version.

rcoh: in this case, the version is always actually the same. We're planning on adding more integrations in the future that actually validate our releases, but in this case the "release" is always `smithy-0.0-alpha` for both sides.

The point of this check is more to flag that a break was made and ensure we handle things accordingly

.github/workflows/ci.yml
rcoh: yes, didn't understand how CI works apparently

pose (approved):

.github/workflows/ci-pr.yml
```diff
@@ -107,3 +107,38 @@ jobs:
...
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+          });
+          const labels = response.data.labels.map(l => l.name);
+          const isBreaking = labels.includes("breaking-change");
+          const data = {
+            labels,
+            isBreaking
+          };
+          console.log("data:", data);
```
Velfi, line 135: Left in for debug purposes I presume?

rcoh: yeah, just to make it easier to see what's going on

PR #2519, opened 2023-03-30
Add stubs to make SRA test fully runnable

Velfi:
This PR adds stubs so that John's SRA test can run to completion. It doesn't add anything real, and the test only passes b/c the stubs are so friendly. No actual operation is getting correctly handled yet.

----


rust-runtime/aws-smithy-runtime-api/src/client/connections/test_connection.rs
```diff
@@ -0,0 +1,220 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Module with client connectors useful for testing.
```
jdisanti, line 6: Seems like this one should go into `aws-smithy-runtime` rather than `-api`.

jdisanti (approved):

PR #2522, opened 2023-03-31
Improve invalid HTTP status code responses failure mode handling

david-perez:
If the model only uses the `@http` trait to define the response code, Smithy already checks that this is a valid status code, so the conversion cannot fail, and we can avoid rejecting with `ResponseRejection::InvalidHttpStatusCode`.

This commit also removes this variant from RPC protocols, where HTTP binding traits are not allowed, and so this failure mode can never happen.

----


hlbarber (approved):
After CI is fixed, this LGTM

david-perez:
I've rebased into a single commit to avoid GitHub's merge queue using an ugly squashed commit message.

PR #2524, opened 2023-03-31
Error log server response rejections

david-perez:
All `ResponseRejection`s are errors; the service owners are to blame. So
we centrally log them here to let them know.

## Testing

I've applied the following patch

```diff
diff --git a/codegen-core/common-test-models/pokemon-common.smithy b/codegen-core/common-test-models/pokemon-common.smithy
index d213a16b1..679ecf2f7 100644
--- a/codegen-core/common-test-models/pokemon-common.smithy
+++ b/codegen-core/common-test-models/pokemon-common.smithy
@@ -38,6 +38,10 @@ structure GetPokemonSpeciesOutput {
     /// A list of flavor text entries for this Pokémon species.
     @required
     flavorTextEntries: FlavorTextEntries
+
+    @required
+    @httpResponseCode
+    responseCode: Integer
 }
 
 /// Retrieve HTTP server statistiscs, such as calls count.
diff --git a/examples/pokemon-service-common/src/lib.rs b/examples/pokemon-service-common/src/lib.rs
index f431270da..ba7636b16 100644
--- a/examples/pokemon-service-common/src/lib.rs
+++ b/examples/pokemon-service-common/src/lib.rs
@@ -190,6 +190,7 @@ pub async fn get_pokemon_species(
             let output = output::GetPokemonSpeciesOutput {
                 name: String::from("pikachu"),
                 flavor_text_entries,
+                response_code: 69,
             };
             Ok(output)
         }
```

And then we get:

```console
➜  pokemon-service git:(davidpz/error-log-server-response-rejections) ✗ RUST_LOG=error cargo run
Finished dev [unoptimized + debuginfo] target(s) in 0.24s
     Running `/home/ANT.AMAZON.COM/davidpz/code/smithy-ws/src/SmithyRsSource/target/debug/pokemon-service`
Hi com.aws.example#GetPokemonSpecies
{"timestamp":"2023-03-31T12:29:14.816095Z","level":"ERROR","fields":{"e":"Http"},"target":"pokemon_service_server_sdk::operation"}
```

Upon doing:

```console
➜  aws-smithy-http-server git:(davidpz/error-log-server-response-rejections) ✗ curl -v http://localhost:13734/pokemon-species/pikachu
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 13734 (#0)
> GET /pokemon-species/pikachu HTTP/1.1
> Host: localhost:13734
> User-Agent: curl/7.58.0
> Accept: */*
>
< HTTP/1.1 400 Bad Request
< content-type: application/json
< x-amzn-errortype: SerializationException
< content-length: 2
< date: Fri, 31 Mar 2023 12:29:14 GMT
<
* Connection #0 to host localhost left intact
{}% 
```

----

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
@@ -135,6 +135,7 @@ class ServerHttpBoundProtocolTraitImplGenerator(
         "ResponseRejection" to protocol.responseRejection(runtimeConfig),
         "PinProjectLite" to ServerCargoDependency.PinProjectLite.toType(),
         "http" to RuntimeType.Http,
+        "Tracing" to RuntimeType.Tracing,
     )
 
     fun generateTraitImpls(operationWriter: RustWriter, operationShape: OperationShape, customizations: List<OperationCustomization>) {
@@ -251,17 +252,22 @@ class ServerHttpBoundProtocolTraitImplGenerator(
         // Implement `into_response` for output types.
         val errorSymbol = symbolProvider.symbolForOperationError(operationShape)
 
+        // All `ResponseRejection`s are errors; the service owners are to blame. So we centrally log them here
+        // to let them know.
         rustTemplate(
             """
             impl #{SmithyHttpServer}::response::IntoResponse<#{Marker}> for #{O} {
                 fn into_response(self) -> #{SmithyHttpServer}::response::Response {
                     match #{serialize_response}(self) {
                         Ok(response) => response,
-                        Err(e) => #{SmithyHttpServer}::response::IntoResponse::<#{Marker}>::into_response(#{RuntimeError}::from(e))
+                        Err(e) => {
+                            #{Tracing}::error!(error = %e, "failed to serialize response");
+                            #{SmithyHttpServer}::response::IntoResponse::<#{Marker}>::into_response(#{RuntimeError}::from(e))
+                        }
                     }
                 }
             }
-            """.trimIndent(),
+            """,
             *codegenScope,
             "O" to outputSymbol,
             "Marker" to protocol.markerStruct(),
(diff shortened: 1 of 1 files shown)


david-perez:
Note we're just logging the enum variant name, which is not helpful. That will be adressed separately once we properly derive `thiserror::Error`, see https://github.com/awslabs/smithy-rs/pull/2517#discussion_r1154202722.

david-perez:
We should also think about having a mechanism to test that the framework is error logging what we expect it to; error logs are very important.

Doing so seems terribly complicated / janky. On one hand the easiest I've found is leveraging `[tracing_test](https://docs.rs/tracing-test/latest/tracing_test/)`; on the other hand, we run integration tests for this binary by spawning a child process via `Command::cargo_bin("pokemon-service")`, so collecting `stdout`/`stderr` might suffice.

However, I'm not too keen on doing any of them: the Pokémon Service does not feel like the right place to perform this kind of framework test; its purpose is for users to inspect and copy paste it.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -253,17 +255,22 @@ class ServerHttpBoundProtocolTraitImplGenerator(
...
+        // to let them know.
         rustTemplate(
             """
             impl #{SmithyHttpServer}::response::IntoResponse<#{Marker}> for #{O} {
                 fn into_response(self) -> #{SmithyHttpServer}::response::Response {
                     match #{serialize_response}(self) {
                         Ok(response) => response,
-                        Err(e) => #{SmithyHttpServer}::response::IntoResponse::<#{Marker}>::into_response(#{RuntimeError}::from(e))
+                        Err(e) => {
+                            #{Tracing}::error!(%e);
```
hlbarber, line 267: Do we want to add a bit of context here?

`error!(error = %e, "failed to serialize response due to internal server error")`

hlbarber (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
hlbarber: We also want to do this for `#{E}` below?

hlbarber:
This partially addresses https://github.com/awslabs/smithy-rs/issues/2521

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
david-perez: I added `failed to serialize response`. "Due to internal server error" will be interpreted as "the framework is to blame".

david-perez:
Oh wow this is the first time we add `tracing` as a dependency of the generated SDKs.

PR #2525, opened 2023-03-31
runtime: add a method `into_segments` to `AggregatedBytes`

parker-timmerman:
## Motivation and Context
This change makes the `AggregatedBytes` type a bit more flexible, by allowing you to convert an `AggregatedBytes` to your own type, without needing to copy any data. This is very useful if you have your own internal non-contiguous data types that you need to convert to. But it doesn't necessarily leak implementation details of `AggregatedBytes` by using an `Iterator` for the return type.

## Description
Add a method called `into_segments(self) -> impl Iterator<Item = Bytes>` to `AggregatedBytes`.

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -74,3 +74,9 @@ message = "Add support for omitting session token in canonical requests for SigV
 references = ["smithy-rs#2473"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
 author = "martinjlowm"
+
+[[aws-sdk-rust]]
+message = "Add `into_segments` method to `AggregatedBytes`, for zero-copy conversions."
+references = ["smithy-rs#2525"]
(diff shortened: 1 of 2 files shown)


parker-timmerman:
Thanks @jdisanti for kicking off CI! There is a failure, but it seems unrelated, something about `tracing-attributes` failing to compile?

jdisanti (approved):
Thank you for the contribution! Looks good to me. The `tracing-attribute` failure is unrelated and non-blocking.

parker-timmerman:
Thanks for merging it @jdisanti! Also, do y'all have a standard release cadence? Just trying to estimate when this change might get released, thanks! :)

jdisanti:
Sorry for the delayed response. I don't have any estimate for when the next release that includes this will be right now. We don't have an established cadence.

parker-timmerman:
No worries at all!

PR #2528, opened 2023-04-03
Use `aws_smithy_types::primitive::Parse` in the server

### LINKED ISSUE #1232: Use `aws_smithy_types::primitive::Parse` to parse URI path labels and query strings as primitive types in the server
The `aws_smithy_types` crate conveniently exposes the [`aws_smithy_types::primitive::Parse`](https://docs.rs/aws-smithy-types/latest/aws_smithy_types/primitive/trait.Parse.html) trait implemented by some types to parse "Smithy primitives".

However, we're not using it in the server implementation to parse URI path labels and query string key-value pairs. We're instead generating this:

https://github.com/awslabs/smithy-rs/blob/f34abdaf2c3a8fb0c160778f36911efb457f9a18/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpProtocolGenerator.kt#L1018-L1036

david-perez:
This further reduces the number of request rejection variants.
https://docs.rs/aws-smithy-types/latest/aws_smithy_types/primitive/trait.Parse.html

Closes #1232.

----


hlbarber (approved):
Great stuff

david-perez:
I've rebased into a single commit to prevent GitHub's merge queue using an ugly squashed commit message.

PR #2531, opened 2023-04-03
Check `Content-Type` header in all server protocols

david-perez:
## Checklist

----


hlbarber (approved):
After lint fixes, LGTM

david-perez:
I can't remember why this is in draft mode and I didn't merge it at the time. Looking at the patch now, I guess a protocol test would be nice so that we don't regress.

david-perez:
It looks like #2330 did this but only for the restJson1 protocol. `ServerHttpBoundProtocolGenerator` cannot be protocol-specific in any case.

I've merged from `main` to freshen up this branch; it's now ready to go in. 

> I guess a protocol test would be nice so that we don't regress.

The [`RestJsonWithBodyExpectsApplicationJsonContentType`](https://github.com/awslabs/smithy/blob/2f98e69c2463c756d002a239dc799d2d30274e81/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-content-type.smithy#L37-L37) protocol test in Smithy suffices as a regression test here.

PR #2540, opened 2023-04-04
Add `AlbHealthCheckLayer`

jjant:
## Motivation and Context
Services often need the ability to report health status via health checks (see [ALB Health Checks](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html)).
This PR adds a simple layer that allows configuring your service to respond to these health check requests.

## Description
Adds `AlbHealthCheckLayer`, and `AlbHealthCheckService`.

## Testing
Added this layer to the `pokemon-service` binary, and added an integration test for it.

----


rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// A handler that returns `200 OK` with an empty body.
+pub async fn default_ping_handler<E>(_req: Request<Body>) -> Result<Response<BoxBody>, E> {
+    let response = Response::builder()
+        .status(StatusCode::OK)
+        .body(body::boxed(Body::empty()))
+        .expect("Couldn't construct response");
+
+    Ok::<_, E>(response)
+}
```
jjant, line 86: I wanted to have a `CheckHealthLayer::with_default_handler` method instead but I couldn't make it work because of not being able to use `impl Future` in 
```rust
CheckHealthLayer<fn(Request<Body>) -> impl Future<Output = Result<Response<BoxBody>, E>>>
```

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,86 @@
...
+pub struct CheckHealthService<H, S> {
+    inner: S,
+    layer: CheckHealthLayer<H>,
+}
+
+impl<H, HandlerFuture, S> Service<Request<Body>> for CheckHealthService<H, S>
+where
+    S: Service<Request<Body>, Response = Response<BoxBody>> + Clone,
+    S::Future: std::marker::Send + 'static,
+    HandlerFuture: Future<Output = Result<Response<BoxBody>, S::Error>>,
```
jjant, line 49: Should this be a `Future<Output = Response<BoxBody>` (no error)? 🤔

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
hlbarber: You can make the function sync and return a [Ready](https://doc.rust-lang.org/std/future/struct.Ready.html).

jjant: Good idea.

jjant: Done in 7fb5c4800.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,96 @@
...
+    H: Fn(Request<Body>) -> HandlerFuture,
+{
+    type Response = S::Response;
+
+    type Error = S::Error;
+
+    type Future = Either<HandlerFuture, S::Future>;
```
hlbarber, line 66: Can we make this opaque using the `opaque_future` macro?

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    fn poll_ready(&mut self, cx: &mut std::task::Context<'_>) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let clone = self.inner.clone();
```
hlbarber, line 73: I don't think we need to `clone` here? Can't we just `self.service.call`?

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,96 @@
...
+    H: Fn(Request<Body>) -> HandlerFuture,
+{
+    type Response = S::Response;
+
+    type Error = S::Error;
+
+    type Future = Either<HandlerFuture, S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut std::task::Context<'_>) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
```
hlbarber, line 69: This `Service` is kinda a "router" style branch. The `poll_ready` is actually too restrictive for the case where the `http::Request` directs us to the ping handler.

_Maybe_ what we want here is `Poll::Ready(Ok())` and then in the handler we use `ServiceExt::oneshot` on the inner service.

hlbarber: @LucioFranco What's the guidance here? If there's a service which contains multiple services, each of which are called depending on control flow controlled by the `Request`, then what should we do:

a) Check that all inner services are `Poll::Ready`?
b) Always return `Poll::Ready` and use `oneshot` in the `call`?

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
hlbarber: This is not the case if we `oneshot` [(b) in this comment](https://github.com/awslabs/smithy-rs/pull/2540/files#r1158283055)

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
LucioFranco: So I think you can look at balance/steer in the tower repo to see how they handle this stuff. Though in this case I feel like since this is server side and generally on server side you don't really use back pressure since the only real mechanism for back pressure is load shedding (and for h2 not reading from the body which stops the window from moving). So that being said I feel like here you can do what @hlbarber said and return read and then oneshot. I would maybe even review the middleware that you expect to be below this in the stack and see if they do anything with poll_ready it might be the case that they don't do anything and its safe to just call directly into call.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+impl<R, S: Service<R>, F: Future<Output = S::Response>> Future for MappedHandlerFuture<R, S, F> {
+    type Output = Result<S::Response, S::Error>;
+
+    fn poll(self: std::pin::Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.project().inner.poll(cx).map(Ok)
+    }
+}
```
jjant, line 79: Is there a simpler way to do this? I basically just want to `map` `HandlerFuture` from a `Future<Output = Response<BoxBody>>` to a `Future<Output = Result<Response<BoxBody>, S::Error>>`.

I tried using [`futures_util::future::Map`](https://docs.rs/futures-util/latest/futures_util/future/struct.Map.html) but I ran into a compiler error when trying to write the type in line 92 as (expected function pointer but got function item):

```rust
type Future = Either<
  Map<HandlerFuture, fn(Response<BoxBody) -> Result<Response<BoxBody>, S::Error>>,
  Oneshot<S, Request<Body>>
>;
```

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
jjant: Done in 80309496b2ce01917cb6db6fe4daf503042ef0e5.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
jjant: Used `oneshot` in bdbed1b012aa191884021398850deefb5a7e63f0.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
jjant: I went for `(b)`, pushed in bdbed1b012aa191884021398850deefb5a7e63f0.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,115 @@
...
+    S: Service<Request<Body>, Response = Response<BoxBody>> + Clone,
+    S::Future: std::marker::Send + 'static,
+    HandlerFuture: Future<Output = Response<BoxBody>>,
+    H: Fn(Request<Body>) -> HandlerFuture,
+{
+    type Response = S::Response;
+
+    type Error = S::Error;
+
+    type Future = Either<MappedHandlerFuture<Request<Body>, S, HandlerFuture>, Oneshot<S, Request<Body>>>;
```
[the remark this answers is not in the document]
jjant: Done in 6735a32f114dd2d04c228a1fa4331fde9d2d05c0.

examples/pokemon-service/src/main.rs
```diff
@@ -46,7 +51,17 @@ pub async fn main() {
         // `Response::extensions`, or infer routing failure when it's missing.
         .insert_operation_extension()
         // Adds `tracing` spans and events to the request lifecycle.
-        .instrument();
+        .instrument()
+        // Handle `/ping` health check requests.
+        .http_layer(CheckHealthLayer::new("/ping", |_req| {
```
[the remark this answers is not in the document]
jjant: Done in 6772f5a799085db3c09f9d8b9af973f4271a5ed. I didn't realise you could use async blocks with closures 👍.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,148 @@
...
+/// A middleware [`Service`] responsible for handling health check requests.
+#[derive(Clone, Debug)]
+pub struct CheckHealthService<H, S> {
+    inner: S,
+    layer: CheckHealthLayer<H>,
+}
+
+pin_project! {
+    /// A future that converts `F` into a compatible [`Service::Future`].
+    pub struct MappedHandlerFuture<R, S, F> {
```
hlbarber, line 59: I don't think we want this to be `pub`? Should we just inline it into `CheckHealthFuture`?

jjant: Inlined in 9fd208889ccaf8c9393640bcb02c4a755e6938f8.

rust-runtime/aws-smithy-http-server/src/plugin/mod.rs
```diff
@@ -118,6 +118,7 @@
 //! ```
 //!
 
+pub mod check_health;
```
hlbarber, line 121: I'm hesitant to put it here but I have no better suggestions. Maybe we should think about making a separate crate dedicated to middleware at some point soon perhaps.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Middleware for handling health check requests.
```
hlbarber, line 6: Little example in the docs here perhaps?

hlbarber (approved):

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
jjant: Done in https://github.com/awslabs/smithy-rs/pull/2540/commits/5bed399c97fb322d9e779b0576cb660d8deef0f8.

rust-runtime/aws-smithy-http-server/src/plugin/mod.rs
jjant: Yes, I think we'll end up moving these to a separate crate eventually.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,143 @@
...
+    inner: S,
+    layer: CheckHealthLayer<H>,
+}
+
+impl<H, HandlerFuture, S> Service<Request<Body>> for CheckHealthService<H, S>
+where
+    S: Service<Request<Body>, Response = Response<BoxBody>> + Clone,
+    S::Future: std::marker::Send + 'static,
+    HandlerFuture: Future<Output = Response<BoxBody>>,
+    H: Fn(Request<Body>) -> HandlerFuture,
```
david-perez, line 78: Why restrict the ping handler to an async function? Why not an arbitrary full-blown `Service`?

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,143 @@
...
+use tower::{util::Oneshot, Layer, Service, ServiceExt};
+
+use crate::body::BoxBody;
+
+use super::either::EitherProj;
+use super::Either;
+
+/// A [`tower::Layer`] used to apply [`CheckHealthService`].
+#[derive(Clone, Debug)]
+pub struct CheckHealthLayer<PingHandler> {
```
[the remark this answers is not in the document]
david-perez: The naming here should be specific to ALB. We'll have integrations with several systems running health checks on our systems, and each integration will have its own quirks, with some commonalities.

david-perez (changes requested):

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
hlbarber: Two constructors would be nice, using `service_fn` in the case where we pass a handler.

examples/pokemon-service/Cargo.toml
```diff
@@ -24,6 +24,9 @@ async-stream = "0.3"
 rand = "0.8.5"
 serial_test = "1.0.0"
 
+# We use hyper client on tests
```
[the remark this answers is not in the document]
jjant: I'll never get these right 🤦‍♂️. Done in 32007cd52ec8aa27c1fafe1c37ae90540b3ab0dd.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,143 @@
...
+    health_check_uri: &'static str,
```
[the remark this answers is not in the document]
jjant: That's true, I've made this a non-static reference in 7fad221f12c07ad5fd1506efc0ceadf358f99f8a.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,143 @@
...
+    ping_handler: PingHandler,
```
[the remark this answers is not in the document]
jjant: Done in 3d15ad851cda0e9d57b1b02e1d09fa334ff78334.

rust-runtime/aws-smithy-http-server/src/plugin/pipeline.rs
```diff
@@ -168,6 +170,11 @@ impl<P> PluginPipeline<P> {
     pub fn push<NewPlugin>(self, new_plugin: NewPlugin) -> PluginPipeline<PluginStack<NewPlugin, P>> {
         PluginPipeline(PluginStack::new(new_plugin, self.0))
     }
+
+    /// Applies a single [`tower::Layer`] to all operations _before_ they are deserialized.
+    pub fn http_layer<L>(self, layer: L) -> PluginPipeline<PluginStack<HttpLayer<L>, P>> {
```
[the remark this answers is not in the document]
jjant: Added a changelog entry for this in b822c5388.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Middleware for handling health check requests.
+//!
+//! # Example
+//!
+//! ```rust
```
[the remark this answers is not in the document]
jjant: Done in 887fbdc8659becb82992d84c4b87a5e59b630fd0.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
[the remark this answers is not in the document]
jjant: Done in 05af5e784ae3eccdec3b62647765a6efa58d8aa4.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
jjant: Given that this is going to be `ALB`-specific, I'm not sure we need to allow a full blown service in here, even this closure is too general: AFAIK, ALB only cares about the status code of the response you send, so maybe we would just need a `Fn() -> StatusCode` here?

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// A [`tower::Layer`] used to apply [`CheckHealthService`].
+#[derive(Clone, Debug)]
+pub struct CheckHealthLayer<PingHandler> {
+    health_check_uri: &'static str,
+    ping_handler: PingHandler,
+}
+
+impl CheckHealthLayer<()> {
+    pub fn new<HandlerFuture: Future<Output = Response<BoxBody>>, H: Fn(Request<Body>) -> HandlerFuture>(
```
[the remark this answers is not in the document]
jjant: Documented in a2719d4a3ba9489bddd6fe3efa0425d37c437621.

rust-runtime/aws-smithy-http-server/src/plugin/check_health.rs
david-perez: We could make the check just be a determination of a status code if that's all that ALB cares about; the interface would model a tighter reality and more apparent to the user as to what really matters. I like it.

As to what we should take in here, my thoughts are: healthchecks should be fast and simple, e.g. verifying some in-memory state or ensuring the filesystem is ok. An `async` context is thus probably preferred, since, despite it allowing the user to possibly do some dumb costly stuff, filesystem operations are async. It seems a bit arbitrary to restrict it to be an async function here though, since our composition unit in the framework has been the Tower service. I expect almost all healthchecks to be implemented as simple async functions, so I think having a constructor for `AlbHealthCheckLayer` that takes in an async function is good. But it'd also be neat to have a second constructor take in a service; we can use `service_fn` to lift the function in the former constructor and continue working in the library code with just Tower services like Harry suggests. The Tower service can have `type Response = http::StatusCode` too, it doesn't have to return HTTP responses.

jjant: Done in https://github.com/awslabs/smithy-rs/pull/2540/commits/f22e0129e9ae88e0447b525132a8105804d1343d.
I'm not happy with how complex the type of `new_fn` is, but anything simpler than that caused the pokemon service example to fail to compile :(.

rust-runtime/aws-smithy-http-server/src/plugin/alb_health_check.rs
```diff
@@ -43,7 +45,24 @@ pub struct AlbHealthCheckLayer<'a, HealthCheckHandler> {
 
 impl<'a> AlbHealthCheckLayer<'a, ()> {
     /// Handle health check requests at `health_check_uri` with the specified handler.
-    pub fn new<HandlerFuture: Future<Output = Response<BoxBody>>, H: Fn(Request<Body>) -> HandlerFuture>(
+    pub fn new_fn<HandlerFuture: Future<Output = StatusCode>, H: Fn(Request<Body>) -> HandlerFuture + Clone>(
```
hlbarber, line 48: `from_handler` and `from_service` like [Operation](https://docs.rs/aws-smithy-http-server/0.55.1/aws_smithy_http_server/operation/struct.Operation.html)?

rust-runtime/aws-smithy-http-server/src/plugin/alb_health_check.rs
```diff
@@ -0,0 +1,178 @@
...
+
+use crate::body::BoxBody;
+
+use super::either::EitherProj;
+use super::Either;
+
+/// A [`tower::Layer`] used to apply [`AlbHealthCheckService`].
+#[derive(Clone, Debug)]
+pub struct AlbHealthCheckLayer<'a, HealthCheckHandler> {
+    health_check_uri: &'a str,
```
hlbarber, line 42: I don't think this will work in any case other than `'static`, this probably needs to be a `Cow<'static, str>`.

jjant: I think you're right, but this can be a `Cow<'a, str>`, right?

rust-runtime/aws-smithy-http-server/src/plugin/alb_health_check.rs
jjant: Done in a961ecb40.

rust-runtime/aws-smithy-http-server/src/plugin/alb_health_check.rs
hlbarber: If `&'a str` doesn't work I'd expect `Cow::<'a, str>::Borrow` to not work either. Lemme try it in the `pokemon-service` quickly.

hlbarber: Yeah
```rust
    let ping = "/ping".to_string();

/* ... */

        .http_layer(AlbHealthCheckLayer::from_handler(&ping, |_req| async {
            StatusCode::OK
        }));
```

yields

```
error[E0597]: `ping` does not live long enough
  --> pokemon-service/src/main.rs:57:55
   |
57 |           .http_layer(AlbHealthCheckLayer::from_handler(&ping, |_req| async {
   |                                                         ^^^^^ borrowed value does not live long enough
...
61 |       let app = PokemonService::builder_with_plugins(plugins)
   |  _______________-
62 | |         // Build a registry containing implementations to all the operations in the service. These
63 | |         // are async functions or async closures that take as input the operation's input and
64 | |         // return the operation's output.
65 | |         .get_pokemon_species(get_pokemon_species)
   | |_________________________________________________- argument requires that `ping` is borrowed for `'static`
...
94 |   }
   |   - `ping` dropped here while still borrowed
```

jjant: Yeah, you're totally right. Done in d56d5cd46.

david-perez (approved):
Very nice.

PR #2541, opened 2023-04-04
[Python] Support more testing model

crisidev:
## Motivation and Context
Python should be tested and compliant with all the testing models used in Rust. To achieve this, there are several changes required in terms of customizations, constraint traits and general bug fixing for edge cases.

In languages that are not Rust, we swap non primitive symbols like `DateTime`, `Blob`, `Document`, etc with wrappers that are compatible with the target language.

This was done using a hack that was looking at the original symbol. If that differed from the new one, we called `from()` or `into()` to make rustc happy.

## Description

This PR implements several changes:

### Symbol swapping

In languages that are not Rust, we swap non primitive symbols like `DateTime`, `Blob`, `Document`, etc with wrappers that are compatible with the target language.

During deserialization the runtime parser requires the Rust symbols, while we are handed back the Python wrappers. Customizations are added call `.map($type::from)` for `Option<T>`, `.into()` for `T`, etc. The customizations are added to the `JsonParserGenerator`, in the `ServerHttpBoundProtocolGenerator` and in `HttpBindingGenerator`.

This should close https://github.com/awslabs/smithy-rs/issues/2479

### Recursive shapes

Implements `FromPyObject` and `IntoPy` traits for `Box<T>` to support recursive shapes. This should close https://github.com/awslabs/smithy-rs/issues/2476

### Naming obstacle

Ensure we use proper casing and escapes for reserved symbol in all Python code (see also https://github.com/awslabs/smithy-rs/issues/2552)

### Constrained Blob

Rust Blobs can be constrained so we need to implement the `From` trait to allow conversion with the Python Blob type.

This should close https://github.com/awslabs/smithy-rs/issues/2551

### Generate diff for all models

We now also generate diffs for all tested models. This should close https://github.com/awslabs/smithy-rs/issues/2483

### Bug fixing

This PR also fixes a bug in the `PythonServerUnionGenerator` where we are missing a closing `)` in a case of union Unit.

----


crisidev:
I need to fix a bunch of unit tests, but the code is almost ready.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -352,7 +355,9 @@ class HttpBindingGenerator(
             rustType to targetShape
         }
         val parsedValue = safeName()
-        if (coreType == dateTime) {
+        // We only check for the type name to allow other implementations that are swapping symbols like
+        // Python to execute the right branch and not fall back to the string deserialization.
+        if (coreType.name == dateTime.name) {
```
crisidev, line 360: I am not sure around relaxing this check. The main issue here is that since we are swapping out the standard `DateTime` symbol with the Python one, we don't end up in the right branch and we go in the else.

An option would be to check if the `AfterDeserializingIntoADateTimeOfHttpHeaders` customization is present to make the check stronger, something like in pseudocode:

```
if (AfterDeserializingIntoADateTimeOfHttpHeaders in customizations and coreType.name == dateTime.name) or (coreType == dateType) {
    ...
}
```

I really need some advices here.

unexge (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerRestJson.kt
```diff
@@ -21,11 +23,23 @@ import software.amazon.smithy.rust.codegen.server.smithy.generators.protocol.Ser
  * RestJson1 server-side protocol factory. This factory creates the [ServerHttpProtocolGenerator]
  * with RestJson1 specific configurations.
  */
-class ServerRestJsonFactory : ProtocolGeneratorFactory<ServerHttpBoundProtocolGenerator, ServerCodegenContext> {
-    override fun protocol(codegenContext: ServerCodegenContext): Protocol = ServerRestJsonProtocol(codegenContext)
+class ServerRestJsonFactory(
+    private val additionalParserCustomizations: List<JsonParserCustomization> = listOf(),
```
hlbarber, line 27: NIT: Maybe we should have a `ServerCustomizations` class to house all these arguments?

hlbarber (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
rcoh: Matteo and I discussed offline—we'll change the check to look at the Smithy shape instead of the Rust type

tools/ci-scripts/codegen-diff/diff_lib.py
```diff
@@ -61,12 +61,8 @@ def generate_and_commit_generated_code(revision_sha, targets=None):
...
             get_cmd_output(f"mv {target}/build/smithyprojections/{target} {OUTPUT_PATH}/")
             if target == target_codegen_server:
-                get_cmd_output(
-                    f"mv {PYTHON_EXAMPLES_PATH}/pokemon-service-server-sdk/ {OUTPUT_PATH}/codegen-server-test-python/",
-                    check=False)
-                get_cmd_output(
-                    f"mv codegen-server-test/typescript/build/smithyprojections/codegen-server-test-typescript {OUTPUT_PATH}/",
-                    check=False)
+                get_cmd_output(f"mv {target}/python/build/smithyprojections/{target}-python {OUTPUT_PATH}/")
+                get_cmd_output(f"mv {target}/typescript/build/smithyprojections/{target}-typescript {OUTPUT_PATH}/")
```
rcoh, line 65: should this be `target == target_codegen_typescript/python` instead?

rcoh (approved):
codegen-core changes LGTM, small non-blocking comment on diff_lib

tools/ci-scripts/codegen-diff/diff_lib.py
crisidev: At the end it does the same.. I have some other rearchitectures of the diff lib in the next PR. I'll refactor this part in it as well if that's ok?

PR #2542, opened 2023-04-04
Tighten server rejection types

david-perez:
Rejection types used to be loose, but now that they are
protocol-specific (see #2517), they can be tightened up to be more
useful.

Where possible, variants now no longer take in the dynamic
`crate::Error`, instead taking in concrete error types arising from
particular fallible invocations in the generated SDKs.This makes it
easier to see how errors arise, and allows for more specific and helpful
error messages that can be logged.

We `#[derive(thiserror::Error)]` on rejection types to cut down on
boilerplate code.

This commit removes the dependency on `strum_macros`.

----


crisidev:
Is this a breaking change?

david-perez:
> Is this a breaking change?

No, see:

https://github.com/awslabs/smithy-rs/blob/27b923db944ac99000110ad3e61b10d5cef455d8/rust-runtime/aws-smithy-http-server/src/proto/rest_json_1/rejection.rs#L25-L27

hlbarber (approved):

crisidev (commented):
LGTM

PR #2543, opened 2023-04-04
add: orchestrator design doc

Velfi:
This document is intended to capture the philosophy and ideas underpinning the current state of the orchestrator and should grow alongside the orchestrator.

Abhinav requested something like this and I think it'll be helpful for future people.

----


design/src/client/orchestrator.md
```diff
@@ -0,0 +1,205 @@
...
+	D-->|Unretryable Failure|F("Err(SdkError)")
+	D-->|Retryable Failure|C
+```
+
+This process is also referred to as the "request/response lifecycle." In this example, the types of "transmit request" and "transmit response" are protocol-dependent. Typical operations use [HTTP], but we plan to support other protocols like [MQTT] in the future.
+
+In addition to the above steps, the orchestrator must also handle:
+
+- **Endpoint resolution:** figuring out which URL to send a request to.
+- **Authentication, identity resolution, and request signing:** Figuring out is sending the request, their credentials, and how should we insert the credentials into a request.
```
jdisanti, line 20:
```suggestion
- **Authentication, identity resolution, and request signing:** Figuring out who is sending the request, their credentials, and how we should insert the credentials into a request.
```

design/src/client/orchestrator.md
```diff
@@ -0,0 +1,205 @@
...
+- **Interceptors**: Running lifecycle hooks at each point in the request/response lifecycle.
+- **Runtime Plugins:** Resolving configuration from config builders.
+- **Retries:** Categorizing response from services and deciding whether to retry and how long to wait before doing so.
```
jdisanti, line 23:
```suggestion
- **Retries:** Categorizing responses from services and deciding whether to retry and how long to wait before doing so.
```

design/src/client/orchestrator.md
```diff
@@ -0,0 +1,205 @@
...
+
+While the structure of an orchestrator is fixed, the actions it takes during its lifecycle are highly configurable. Users have two ways to configure this process:
+
+- **Runtime Plugins**:
+    - **When can these be set?** Any time before calling `orchestrate`.
+    - **When are they called by the orchestrator?** In two batches, at the very beginning of `orchestrate`.
+    - **What can they do?**
+        - They can set configuration to be used by the orchestrator or in interceptors.
+        - They can set interceptors.
+    - **Are they user-definable?** No. At present, only smithy maintainers may define these.
```
jdisanti, line 36:
```suggestion
    - **Are they user-definable?** No. At present, only smithy-rs maintainers may define these.
```

design/src/client/orchestrator.md
```diff
@@ -0,0 +1,205 @@
...
+    - *This phase is infallible.*
+    - An interceptor context is created. This will hold request and response objects, making them available to interceptors.
+    - All runtime plugins set at the client-level are run. These plugins can set config and mount interceptors. Any _"read before execution"_ interceptors that have been set get run.
+    - All runtime plugins set at the operation-level are run. These plugins can also set config and mount interceptors. Any new _"read before execution"_ interceptors that have been set get run.
+1. **Request Construction**
+    - *This phase is fallible.*
+    - The _"read before serialization"_ and _"modify before serialization"_ interceptors are called.
+    - The input is serialized into a transmit request.
+    -  The _"read after serialization"_ and _"modify before retry_ loop" interceptors are called.
+    - Before making an attempt, the retry handler is called to check if an attempt should be made. The retry handler makes this decision for an initial attempt as well as for the retry attempts. If an initial attempt should be made, then the orchestrator enters the Dispatch phase. Otherwise, a throttling error is returned.
```
jdisanti, line 64: Seems like there can be wasted effort if we decide this after serialization.

design/src/client/orchestrator.md
```diff
@@ -0,0 +1,205 @@
...
+#[derive(Debug)]
+pub struct TypeErasedBox {
+    inner: Box<dyn Any + Send + Sync>,
+}
+```
+
+The orchestrator itself doesn't know about any concrete types. Instead, it passes boxed data between the various components of the request/response lifecycle. Individual components access data in two ways:
+
+- **From the `ConfigBag`:** `let retry_strategy: Box<dyn RetryStrategy> = cfg.get()?;`
+- **From the `InterceptorContext`:** `let put_object_input: PutObjectInput = ctx.input().downcast()?;`
```
jdisanti, line 171: It would really be:
```rust
let put_object_input: PutObjectInput = ctx.take_input().unwrap().downcast().unwrap()?;
```
Of if you just want a reference:
```rust
let put_object_input = ctx.input().unwrap().downcast_ref::<PutObjectInput>().unwrap();
```

design/src/client/orchestrator.md
```diff
@@ -0,0 +1,205 @@
...
+/// A new-type around `Box<dyn Any + Send + Sync>`
+#[derive(Debug)]
+pub struct TypeErasedBox {
+    inner: Box<dyn Any + Send + Sync>,
+}
+```
+
+The orchestrator itself doesn't know about any concrete types. Instead, it passes boxed data between the various components of the request/response lifecycle. Individual components access data in two ways:
+
+- **From the `ConfigBag`:** `let retry_strategy: Box<dyn RetryStrategy> = cfg.get()?;`
```
jdisanti, line 170: The `ConfigBagAccessors` centralize this to avoid traps around inserting boxed implementations correctly.

jdisanti (commented):
NIce doc! Just some minor comments

design/src/client/orchestrator.md
Velfi: The internal spec defines it this way. I'll follow up and see if that's deliberate or just an oversight.

jdisanti (approved):

PR #2564, opened 2023-04-11
Implement `StdError::source()` for Error enum

abusch:
## Motivation and Context
This is an attempt at fixing https://github.com/awslabs/aws-sdk-rust/issues/784.

The service-level `Error` enum implements `std::error::Error` but does not implement its `source()` method. This means that an error library like `anyhow` or `eyre` won't be able to display the root cause of an error, which is especially problematic for the `Unhandled` variant.

## Description
I modified `ServiceErrorGenerator` in the `codegen-client` crate and replaced the line that output `impl std::error::Error for Error {}` with an impl block that implements the `source()` method by delegating to the inner error structure.

## Testing
I've added a simple unit test to `ServiceErrorGeneratorTest`.

## Checklist

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/ServiceErrorGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/ServiceErrorGenerator.kt
@@ -80,7 +80,16 @@ class ServiceErrorGenerator(
                         errors.map { it.id },
                     )
                 }
-            rust("impl #T for Error {}", RuntimeType.StdError)
+            rustBlock("impl #T for Error", RuntimeType.StdError) {
+                rustBlock("fn source(&self) -> std::option::Option<&(dyn #T + 'static)>", RuntimeType.StdError) {
+                    rustBlock("match self") {
+                        allErrors.forEach {
+                            rust("Error::${symbolProvider.toSymbol(it).name}(inner) => inner.source(),")
+                        }
+                        rust("Error::Unhandled(inner) => inner.source()")
+                    }
+                }
+            }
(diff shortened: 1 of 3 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/ServiceErrorGenerator.kt
```diff
@@ -80,7 +80,16 @@ class ServiceErrorGenerator(
                         errors.map { it.id },
                     )
                 }
-            rust("impl #T for Error {}", RuntimeType.StdError)
+            rustBlock("impl #T for Error", RuntimeType.StdError) {
+                rustBlock("fn source(&self) -> std::option::Option<&(dyn #T + 'static)>", RuntimeType.StdError) {
+                    rustBlock("match self") {
+                        allErrors.forEach {
+                            rust("Error::${symbolProvider.toSymbol(it).name}(inner) => inner.source(),")
```
rcoh, line 87: hmm...should this be `inner.source()` or `Some(&inner as _)` — I think maybe the latter? `inner.source()` might be an extra level

rcoh (commented):
this looks great overall—definitely exactly what's needed. Just one question about whether we want to double-delegate through the inner source or mark the inner as source directly.

Curious for your thoughts.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/ServiceErrorGenerator.kt
Velfi:
```
Unhandled error
: some error
```

looks more correct to me

Velfi (approved):

crisidev (approved):

PR #2567, opened 2023-04-11
add map_immutable to allow signing to work with a wrapped SdkBody

rcoh:
## Motivation and Context

aws-sdk-rust#749

## Description

When wrapping an `SdkBody`, the resulting body is no longer signable. Some services support this and others don't. This change adds a new method for mapping an `SdkBody` called `map_readonly_body`. If called on a body that is signable, it will produce a signable body. However, this function should never when modifying the inner body is necessary, as that will result in a signing error.

## Testing

Zelda wrote a test

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-http/src/body.rs
+++ b/rust-runtime/aws-smithy-http/src/body.rs
@@ -35,6 +35,7 @@ pin_project! {
         // In the event of retry, this function will be called to generate a new body. See
         // [`try_clone()`](SdkBody::try_clone)
         rebuild: Option<Arc<dyn (Fn() -> Inner) + Send + Sync>>,
+        bytes_contents: Option<Bytes>
     }
 }
 
@@ -94,6 +95,7 @@ impl SdkBody {
         Self {
             inner: Inner::Dyn { inner: body },
             rebuild: None,
+            bytes_contents: None,
         }
     }
 
@@ -110,6 +112,7 @@ impl SdkBody {
         SdkBody {
             inner: initial.inner,
             rebuild: Some(Arc::new(move || f().inner)),
+            bytes_contents: initial.bytes_contents,
         }
     }
 
@@ -119,6 +122,7 @@ impl SdkBody {
         Self {
             inner: Inner::Taken,
             rebuild: None,
+            bytes_contents: None,
         }
     }
 
@@ -127,6 +131,7 @@ impl SdkBody {
         Self {
             inner: Inner::Once { inner: None },
             rebuild: Some(Arc::new(|| Inner::Once { inner: None })),
+            bytes_contents: None,
         }
     }
 
@@ -157,10 +162,9 @@ impl SdkBody {
     /// If this SdkBody is NOT streaming, this will return the byte slab
     /// If this SdkBody is streaming, this will return `None`
     pub fn bytes(&self) -> Option<&[u8]> {
-        match &self.inner {
-            Inner::Once { inner: Some(b) } => Some(b),
-            Inner::Once { inner: None } => Some(&[]),
-            _ => None,
+        match &self.bytes_contents {
+            Some(b) => Some(b),
+            None => None,
         }
     }
 
@@ -172,6 +176,7 @@ impl SdkBody {
             Self {
                 inner: next,
                 rebuild: self.rebuild.clone(),
+                bytes_contents: self.bytes_contents.clone(),
             }
         })
     }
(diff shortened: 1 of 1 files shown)


Velfi (approved):

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -183,14 +188,34 @@ impl SdkBody {
...
+    pub fn map_readonly_body(
+        mut self,
+        f: impl Fn(SdkBody) -> SdkBody + Sync + Send + 'static,
+    ) -> SdkBody {
+        let contents = self.bytes_contents.take();
+        let mut out = if self.rebuild.is_some() {
+            SdkBody::retryable(move || f(self.try_clone().unwrap()))
+        } else {
+            f(self)
+        };
```
rcoh, line 215: hmmm I'm not positive this is actually correct. I think this means that in `f` `.bytes()` will return `None`

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -183,14 +188,32 @@ impl SdkBody {
     }
 
     /// Given a function to modify an `SdkBody`, run that function against this `SdkBody` before
-    /// returning the result.
+    /// returning the result. An `SdkBody` returned from this function is not signable. If you
+    /// want to wrap a body and still have it be signable, see [`SdkBody::map_immutable`].
```
rcoh, line 192: I'd be more specific about `bytes()` rather than referencing signing directly

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -183,14 +188,32 @@ impl SdkBody {
...
             SdkBody::retryable(move || f(self.try_clone().unwrap()))
         } else {
             f(self)
         }
     }
+
+    /// Given a function to modify an `SdkBody`, run that function against this `SdkBody` before
+    /// returning the result. **This function MUST NOT alter the contents of the body.**
+    /// An `SdkBody` returned from this function will be signable so long as the inner
+    /// `SdkBody` is signable. If altering the contents of the inner body is required,
```
rcoh, line 204: same here

CHANGELOG.next.toml
```diff
@@ -92,3 +92,14 @@ message = "Update the `std::fmt::Debug` implementation for `aws-sigv4::SigningPa
 references = ["smithy-rs#2562"]
 meta = { "breaking" = false, "tada" = true, "bug" = true }
 author = "Velfi"
+
+[[aws-sdk-rust]]
+message = """
+When wrapping an `SdkBody`, the resulting body is no longer signable. Some services support this and others don't.
+This change adds a new method for mapping an `SdkBody` called `map_immutable`. If called on a body that is signable,
+it will produce a signable body. However, this function should never when modifying the inner body is necessary,
```
ysaito1001, line 100: should never *be used* when ...?

ysaito1001, line 100: `s/should never when/should never be used when/` ?

ysaito1001, line 100:
```suggestion
it will produce a signable body. However, this function should never be used when modifying the inner body is necessary,
```

CHANGELOG.next.toml
```diff
@@ -92,3 +92,14 @@ message = "Update the `std::fmt::Debug` implementation for `aws-sigv4::SigningPa
...
+as that will result in a signing error.
+"""
+references = ["aws-sdk-rust#749"]
+meta = { "breaking" = false, "tada" = true, "bug" = false }
```
ysaito1001, line 104:
```suggestion
meta = { "breaking" = false, "tada" = true, "bug" = true }
```

ysaito1001 (approved):

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -191,6 +196,22 @@ impl SdkBody {
             f(self)
         }
     }
+
+    /// Given a function to modify an `SdkBody`, run that function against this `SdkBody` before
+    /// returning the result. **This function MUST NOT alter the contents of the body.**
+    pub fn map_preserve_contents(
```
jdisanti, line 202: I think this could use more documentation. I'm still not sure what this is for or why I should use it from reading this doc comment.

Also, is there any way we can assert the content didn't change in debug mode? Probably not?

jszwedko:
Thanks for resurrecting this PR! We'd love to see it merged in so we can stop maintaining a fork for https://github.com/vectordotdev/vector 😄

jszwedko:
Thank you @rcoh ! 🎉 I look forward to dropping our fork.

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -191,6 +196,25 @@ impl SdkBody {
             f(self)
         }
     }
+
+    /// Update this `SdkBody` with `map`. **This function MUST NOT alert the data of the body.**
```
DavidSouther, line 200: Should this be "MUST NOT alter"?

jdisanti: Yeah. I believe it was corrected in a later PR.

Issue #2489, opened 2023-03-23
Move `rust-runtime/aws-smithy-http-server-python/examples` to `/examples`


hlbarber:
We have moved `rust-runtime/aws-smithy-http-server/examples` to `/examples` in [this PR](https://github.com/awslabs/smithy-rs/pull/2481).

We should do the same for `rust-runtime/aws-smithy-http-server-python/examples` and attempt to merge the `make` files and `README.md` perhaps.

If this doesn't make sense then maybe we should have `examples-python`?

PR #2576, opened 2023-04-13
[Python] Automatically generate stubs

crisidev:
## Motivation and Context
We want to automatically generate stubs in the codegen diff to ensure they can be reviewed and have a simple way to generate and include the stubs inside the Maturin wheel.

## Description
The Python example has been moved to the `examples` folder and refactored. The refactoring 
ensures the script `stubgen.py` is included in the codegeneration of the SDK crate. The script is later used to generate stubs automatically during testing and can be used by customers to add their own stubs before the Maturin build

----

----


crisidev:
Closes: #2489

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/customizations/PythonServerCodegenDecorator.kt
```diff
@@ -194,6 +195,31 @@ class PyTypedMarkerDecorator : ServerCodegenDecorator {
...
+        val stubgen_python_location = "$runtime_crates_path/aws-smithy-http-server-python/stubgen.py"
+        val stubgen_python_content = File(stubgen_python_location).readText(Charsets.UTF_8)
+        rustCrate.withFile("stubgen.py") {
+            writeWithNoFormatting("$stubgen_python_content")
+        }
+        val stubgen_shell_location = "$runtime_crates_path/aws-smithy-http-server-python/stubgen.sh"
+        val stubgen_shell_content = File(stubgen_shell_location).readText(Charsets.UTF_8)
+        rustCrate.withFile("stubgen.sh") {
+            writeWithNoFormatting("$stubgen_shell_content")
+        }
```
unexge, line 219: Variables should be camelCase

examples/python/README.md
```diff
@@ -0,0 +1,94 @@
+# Smithy Rust/Python Server SDK example
+
+This folder contains an example service called Pokémon Service used to showcase
+the service framework Python bindings capabilities and to run benchmarks.
+
+The Python implementation of the service can be found inside
+[pokemon_service.py](/rust-runtime/aws-smithy-http-python-server/examples/pokemon_service.py).
```
unexge, line 7:
```suggestion
[pokemon_service.py](./pokemon_service.py).
```

rust-runtime/aws-smithy-http-server-python/src/tls.rs
```diff
@@ -146,16 +146,17 @@ mod tests {
...
     const TEST_CERT: &str = concat!(
         env!("CARGO_MANIFEST_DIR"),
-        "/examples/pokemon-service-test/tests/testdata/localhost.crt"
+        "/../../examples/python/pokemon-service-test/tests/testdata/localhost.crt"
     );
 
     #[test]
     fn creating_tls_config_in_python() -> PyResult<()> {
         pyo3::prepare_freethreaded_python();
+        println!("{}", env!("CARGO_MANIFEST_DIR"));
```
unexge, line 159: Do we want to keep this?

rust-runtime/aws-smithy-http-server-python/src/tls/listener.rs
```diff
@@ -188,10 +188,11 @@ mod tests {
             // We have a certificate with invalid date, so request should fail
             let config = client_config_with_cert(&invalid_cert);
             let response = make_req(config, &addr).await;
+            println!("{}", response.as_ref().unwrap_err().to_string());
```
unexge, line 191: Maybe we can pass `response.as_ref().unwrap_err().to_string()` as second argument to below `assert!` statement

tools/ci-scripts/codegen-diff/diff_lib.py
```diff
@@ -39,18 +39,19 @@ def checkout_commit_and_generate(revision_sha, branch_name, targets=None):
...
+        target_codegen_server_python,
+        target_codegen_server_typescript
+    ]
     # Clean the build artifacts before continuing
     assemble_tasks = ' '.join([f'{t}:assemble' for t in targets])
     clean_tasks = ' '.join([f'{t}:clean' for t in targets])
     get_cmd_output("rm -rf aws/sdk/build")
     get_cmd_output(f"./gradlew --rerun-tasks {clean_tasks}")
     get_cmd_output(f"./gradlew --rerun-tasks {assemble_tasks}")
+    get_cmd_output(f"./gradlew --rerun-tasks {target_codegen_server_python}:stubs")
```
unexge, line 54: I think we should put this into an if like `if target_codegen_server_python in targets:`

unexge (approved):

rust-runtime/aws-smithy-http-server-python/stubgen.sh
```diff
@@ -0,0 +1,40 @@
...
+        rm -rf "$CARGO_TARGET_DIR"
+    }
+    # register the cleanup function to be called on the EXIT signal
+    trap cleanup EXIT
+fi
+export CARGO_TARGET_DIR
+
+# generate the Python stubs
+cargo build --manifest-path "$manifest"
+ln -sf "$CARGO_TARGET_DIR/debug/lib$package.so" "$CARGO_TARGET_DIR/debug/$package.so"
```
unexge, line 37:
```suggestion
if [ "$(uname)" == "Darwin" ]; then
    export SHARED_OBJECT_EXT="so"
else
	export SHARED_OBJECT_EXT="dylib"
fi
export CARGO_TARGET_X86_64_APPLE_DARWIN_RUSTFLAGS="-C link-arg=-undefined -C link-arg=dynamic_lookup"
export CARGO_TARGET_AARCH64_APPLE_DARWIN_RUSTFLAGS="-C link-arg=-undefined -C link-arg=dynamic_lookup"
cargo build --manifest-path "$manifest"
ln -sf "$CARGO_TARGET_DIR/debug/lib$package.$SHARED_OBJECT_EXT" "$CARGO_TARGET_DIR/debug/$package.so"
```

I think we need to make some changes for macOS here, 
- Export some flags to fix linking issues
- Use `.dylib` instead of `.so`, but note that target file still needs to have `.so` extension on macOS too because Python expects `.so` extensions even on macOS

rust-runtime/aws-smithy-http-server-python/stubgen.py
```diff
@@ -401,6 +386,19 @@ def walk_module(writer: Writer, mod: Any):
             print(f"Unknown type: {member}")
```
unexge, line 386: With the new `CODEGEN_VERSION` constant this script will fail here with: `Unknown type: 0.0.0-smithy-rs-head`.

I think we can add support for scalars here:

```
elif is_scalar(member):
    writer.define(f"{name}: {type(member).__name__} = ...")
else:
    print(f"Unknown type: {member}")

...

def is_scalar(obj: Any) -> bool:
    return isinstance(obj, (str, float, int, bool))
```

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/customizations/PythonServerCodegenDecorator.kt
crisidev: Lol.. I am writing Python Rust and Kotlin at the same time 😬 

I'll fix this.

rust-runtime/aws-smithy-http-server-python/src/tls/listener.rs
crisidev: This is a leftover when I was trying to fix the tests for the new version of `rustls`. I'll remove this line.

tools/ci-scripts/codegen-diff/diff_lib.py
crisidev: Yes you are right.

rust-runtime/aws-smithy-http-server-python/stubgen.sh
crisidev: Good point!

pose (approved):

PR #2577, opened 2023-04-13
Add DefaultEndpointResolver to the orchestrator

ysaito1001:
## Motivation and Context
This PR adds `DefaultEndpointResolver`, a default implementer of the [EndpointResolver](https://github.com/awslabs/smithy-rs/blob/1e27efe05fe7b991c9f9bbf3d63a297b2dded334/rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs#L182-L184) trait, to the orchestrator.

## Description
Roughly speaking, the endpoint-related work is currently done by `make_operation` and `SmithyEndpointStage`, where `make_operation` constructs endpoint params and applies an endpoint resolver and the `SmithyEndpointStage` later updates the request header based on the resolved endpoint. In the orchestrator world, that work is handled by `DefaultEndpointResolver::resolve_and_apply_endpoint`.

The way endpoint parameters and an endpoint prefix are made available to `DefaultEndpointResolver::resolve_and_apply_endpoint` is done by interceptors because they _may_ require pieces of information only available in an operation input struct. The place that has access to both an operation input struct and a config bag happens to be an interceptor.

There are two interceptors `<Operation>EndpointParamsInterceptor` and  `<Operation>EndpointParamsFinalizerInterceptor` per operation. We pass around endpoint params _builder_ through interceptors to allow it to be configured with more information at a later point; An end point parameters builder is first initialized within the `ServiceRuntimePlugin` with the field values obtained from the service config, and is stored in a config bag. The finalizer interceptor "seals" the builder and creates the actual endpoint params before it is passed to `DefaultEndpointResolver::resolve_and_apply_endpoint`. These interceptors implement `read_before_execution` because they need to access an operation input before it gets serialized (otherwise, `context.input()` would return a `None`).

## Testing
Replaced `StaticUriEndpointResolver` with `DefaultEndpointResolver` in `sra_test` and verified the test continued passing.

UPDATE: The test currently fails in the `main` branch (it returns a `None` when extracting `SigV4OperationSigningConfig` from the config bag in `OverrideSigningTimeInterceptor`, hence `unwrap` fails), and by merging the `main` branch this PR no longer passes the test, but it does not add new failures either.

----


Velfi (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+        val operationInput = symbolProvider.toSymbol(operationShape.inputShape(model))
+        rustTemplate(
+            """
+            let input = context.input()?;
+            let _input = input
+                .downcast_ref::<${operationInput.name}>()
+                .ok_or_else(|| #{InterceptorError}::invalid_input_access())?;
+            let params_builder = cfg
+                .get::<#{ParamsBuilder}>()
+                .ok_or(#{InterceptorError}::read_before_execution("missing endpoint params builder"))?
```
jdisanti, line 89: Should be `.ok_or_else(|| "missing endpoint params builder".to_string())?`

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+            val endpointTraitBindings = EndpointTraitBindings(
+                codegenContext.model,
+                symbolProvider,
+                codegenContext.runtimeConfig,
+                operationShape,
+                epTrait,
+            )
+            withBlockTemplate(
+                "let endpoint_prefix = ",
+                ".map_err(#{InterceptorError}::read_before_execution)?;",
```
jdisanti, line 114: Noting this error as well.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    private val implInterceptorBodyForEndpointParamsFinalizer: Writable = writable {
+        rustTemplate(
+            """
+            let _ = context;
+            let params_builder = cfg
+                .get::<#{ParamsBuilder}>()
+                .ok_or(#{InterceptorError}::read_before_execution("missing endpoint params builder"))?
```
jdisanti, line 133: And these two errors.

rust-runtime/aws-smithy-runtime-api/src/client/endpoints.rs
```diff
@@ -28,8 +34,77 @@ impl StaticUriEndpointResolver {
...
+        _endpoint_prefix: Option<&EndpointPrefix>,
+        request: &mut HttpRequest,
+    ) -> Result<(), BoxError> {
         apply_endpoint(request.uri_mut(), &self.endpoint, None)?;
         Ok(())
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct DefaultEndpointResolver<Params> {
```
jdisanti, line 49: This should be in the `aws-smithy-runtime` crate rather than `-api` since it is a trait implementation rather than a trait.

rust-runtime/aws-smithy-runtime-api/src/client/endpoints.rs
```diff
@@ -28,8 +34,77 @@ impl StaticUriEndpointResolver {
...
+            for value in header_values {
+                request.headers_mut().insert(
+                    HeaderName::from_str(header_name).map_err(|err| {
+                        ResolveEndpointError::message("invalid header name")
+                            .with_source(Some(err.into()))
+                    })?,
+                    HeaderValue::from_str(value).map_err(|err| {
+                        ResolveEndpointError::message("invalid header value")
+                            .with_source(Some(err.into()))
+                    })?,
```
jdisanti, line 103: Should these errors even be possible if the endpoint config is valid?

jdisanti (commented):
Missed this PR since I was out last week. I have some comments around errors.

rust-runtime/aws-smithy-runtime-api/src/client/endpoints.rs
ysaito1001: I am not entirely sure. I've decided to preserve the existing error handling logic from the [original PR](https://github.com/awslabs/smithy-rs/pull/2063) for the sake of porting. I don't see how an endpoint resolver comes into being after a series of config steps in the orchestrator world is all that different from the old world, so maybe it makes sense to keep the existing error handling for now (we can later turn them into non-recoverable errors, if we want).

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private fun implInterceptorBodyForEndpointParams(operationShape: OperationShape): Writable = writable {
+        val operationInput = symbolProvider.toSymbol(operationShape.inputShape(model))
+        rustTemplate(
+            """
+            let input = context.input()?;
+            let _input = input
+                .downcast_ref::<${operationInput.name}>()
+                .ok_or_else(|| #{InterceptorError}::invalid_input_access())?;
```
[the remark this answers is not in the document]
ysaito1001: Thanks for sharing an important practice! Addressed in 10a1967.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
ysaito1001: Addressed in 10a1967.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
ysaito1001: Addressed in 10a1967.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
ysaito1001: Addressed in 10a1967.

rust-runtime/aws-smithy-runtime-api/src/client/endpoints.rs
ysaito1001: Makes sense. Addressed in 8219785.

PR #2592, opened 2023-04-18
Follow up on DefaultEndpointResolver in the orchestrator

ysaito1001:
## Motivation and Context
This PR incorporates post-merge feedback left in #2577.

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -83,10 +83,10 @@ class EndpointParamsInterceptorGenerator(
             let input = context.input()?;
             let _input = input
                 .downcast_ref::<${operationInput.name}>()
-                .ok_or_else(|| #{InterceptorError}::invalid_input_access())?;
+                .ok_or_else(|| "failed to downcast to ${operationInput.name}")?;
             let params_builder = cfg
                 .get::<#{ParamsBuilder}>()
-                .ok_or(#{InterceptorError}::read_before_execution("missing endpoint params builder"))?
+                .ok_or_else(|| "missing endpoint params builder".to_owned())?
```
jdisanti, line 89: Is the `to_owned()` needed? Looks like it wasn't on the one above.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -111,7 +111,7 @@ class EndpointParamsInterceptorGenerator(
             )
             withBlockTemplate(
                 "let endpoint_prefix = ",
-                ".map_err(#{InterceptorError}::read_before_execution)?;",
+                """.map_err(|err| format!("endpoint prefix could not be built: {err:?}"))?;""",
```
jdisanti, line 114: I'm starting to think we need some kind of `ErrorMessage` struct (perhaps there's a better name for it) that is just a string message and a `Box<dyn Error>` source. We shouldn't be string formatting errors.

This could be something like:
```rust
.map_err(|err| ErrorMessage::new("endpoint prefix could not be built", err))?
```

Essentially we just need a convenient way to attach additional context to an error. Or maybe better, could do something inverse with traits:

```rust
.map_err(|err| "endpoint prefix could not be built".source(err))?
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
ysaito1001: Good catch. Fixed in 51026e9.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
ysaito1001: I like that idea. Added a convenience type `ContextAttachedError` in 51026e9.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/error.rs
```diff
@@ -169,3 +169,31 @@ impl std::error::Error for InterceptorError {
...
+
+/// A convenience error that allows for adding additional `context` to `source`
+#[derive(Debug)]
+pub struct ContextAttachedError {
+    context: String,
+    source: Option<BoxError>,
+}
+
+impl ContextAttachedError {
+    pub fn new(context: impl Into<String>, source: BoxError) -> Self {
```
jdisanti, line 181: You can make source `impl Into<BoxError>` to simplify calling `new`.

ysaito1001: Why not 👍  Addressed in 93d044d.

jdisanti (approved):

Velfi (approved):

PR #2593, opened 2023-04-19
Add a middleware vs. orchestrator performance comparison benchmark

jdisanti:
## Motivation and Context

This PR adds a benchmark suite to the `aws-sdk-s3` test in `aws/sra-test` so that the performance of client middleware can be compared against the new orchestrator approach.

The comparison is not yet apples to apples since the orchestrator isn't complete enough to be comparable, but this is a benchmark that we can check regularly as they become more comparable.

Initially, the performance was really bad (43x slower than middleware), but that turned out to be a trivial mistake in orchestrator implementation that was causing some configuration JSON to get reparsed for every request. Fixing that makes the orchestrator only about 1.5x slower, and there is another trivial optimization that can be made to bring that down more.

The benchmark can be run with the following:
```bash
smithy-rs$ ./gradlew aws:sra-test:assemble
smithy-rs$ cd aws/sra-test/integration-tests/aws-sdk-s3
smithy-rs$ cargo bench
```

----

### DIFF
--- a/aws/sra-test/integration-tests/aws-sdk-s3/benches/middleware_vs_orchestrator.rs
+++ b/aws/sra-test/integration-tests/aws-sdk-s3/benches/middleware_vs_orchestrator.rs
@@ -1,0 +1,285 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#[macro_use]
+extern crate criterion;
+use aws_sdk_s3 as s3;
+use aws_smithy_client::erase::DynConnector;
+use aws_smithy_client::test_connection::infallible_connection_fn;
+use aws_smithy_runtime_api::type_erasure::TypedBox;
+use criterion::Criterion;
+use s3::operation::list_objects_v2::{ListObjectsV2Error, ListObjectsV2Input, ListObjectsV2Output};
+
+async fn middleware(client: &s3::Client) {
+    client
+        .list_objects_v2()
+        .bucket("test-bucket")
+        .prefix("prefix~")
+        .send()
+        .await
+        .expect("successful execution");
+}
+
+async fn orchestrator(connector: &DynConnector) {
+    // TODO(enableNewSmithyRuntime): benchmark with `send_v2` directly once it works
+    let runtime_plugins = aws_smithy_runtime_api::client::runtime_plugin::RuntimePlugins::new()
+        .with_client_plugin(orchestrator::ManualServiceRuntimePlugin(connector.clone()))
+        .with_operation_plugin(aws_sdk_s3::operation::list_objects_v2::ListObjectsV2::new())
+        .with_operation_plugin(orchestrator::ManualOperationRuntimePlugin);
+    let input = ListObjectsV2Input::builder()
+        .bucket("test-bucket")
+        .prefix("prefix~")
+        .build()
+        .unwrap();
+    let input = TypedBox::new(input).erase();
+    let output = aws_smithy_runtime::client::orchestrator::invoke(input, &runtime_plugins)
+        .await
+        .map_err(|err| {
+            err.map_service_error(|err| {
+                TypedBox::<ListObjectsV2Error>::assume_from(err)
+                    .expect("correct error type")
+                    .unwrap()
+            })
+        })
+        .unwrap();
+    TypedBox::<ListObjectsV2Output>::assume_from(output)
+        .expect("correct output type")
+        .unwrap();
+}
+
+fn test_connection() -> DynConnector {
+    infallible_connection_fn(|req| {
+        assert_eq!(
+            "https://test-bucket.s3.us-east-1.amazonaws.com/?list-type=2&prefix=prefix~",
+            req.uri().to_string()
+        );
+        assert!(req.headers().contains_key("authorization"));
+        http::Response::builder()
+            .status(200)
+            .body(
+                r#"<?xml version="1.0" encoding="UTF-8"?>
+<ListBucketResult>
+    <Name>test-bucket</Name>
+    <Prefix>prefix~</Prefix>
+    <KeyCount>1</KeyCount>
+    <MaxKeys>1000</MaxKeys>
+    <IsTruncated>false</IsTruncated>
+    <Contents>
+        <Key>some-file.file</Key>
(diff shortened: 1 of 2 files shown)


aws/sra-test/integration-tests/aws-sdk-s3/benches/middleware_vs_orchestrator.rs
```diff
@@ -0,0 +1,285 @@
...
+            );
+
+            //cfg.set_endpoint_resolver(DefaultEndpointResolver::new(
+            //    aws_smithy_http::endpoint::SharedEndpointResolver::new(
+            //        aws_sdk_s3::endpoint::DefaultResolver::new(),
+            //    ),
+            //));
+            cfg.set_endpoint_resolver(StaticUriEndpointResolver::uri(Uri::from_static(
+                "https://test-bucket.s3.us-east-1.amazonaws.com/",
+            )));
```
jdisanti, line 173: This was where the endpoint JSON was getting reparsed for each request, which was causing the massive performance difference.

aws/sra-test/integration-tests/aws-sdk-s3/benches/middleware_vs_orchestrator.rs
```diff
@@ -0,0 +1,285 @@
...
+    impl RuntimePlugin for ManualServiceRuntimePlugin {
+        fn configure(&self, cfg: &mut ConfigBag) -> Result<(), BoxError> {
+            let identity_resolvers =
+                aws_smithy_runtime_api::client::orchestrator::IdentityResolvers::builder()
+                    .identity_resolver(
+                        aws_runtime::auth::sigv4::SCHEME_ID,
+                        aws_runtime::identity::credentials::CredentialsIdentityResolver::new(
+                            SharedCredentialsCache::new(CredentialsCache::lazy().create_cache(
+                                SharedCredentialsProvider::new(Credentials::for_tests()),
+                            )),
```
jdisanti, line 141: Another optimization can be made here: the credential cache should not be getting recreated for every request.

aws/sra-test/integration-tests/aws-sdk-s3/benches/middleware_vs_orchestrator.rs
ysaito1001: Related to [the discussion](https://github.com/awslabs/smithy-rs/pull/2593/files#r1171662473), once we get the configuration working, we will no longer need to `aws_sdk_s3::endpoint::DefaultResolver::new()` within the `configure` method; `DefaultResolver` has already been created as part of the service config construction and been placed into the config bag. So the commented out code above will eventually look like the following:
```
cfg.set_endpoint_resolver(DefaultEndpointResolver::new(
    cfg.get::<SharedEndpointResolver>().expect("...").clone(),
));
```

ysaito1001 (approved):

hlbarber:
A large portion of the call here is probably going to be serialization/deserialization, even in the case of small inputs/outputs? Which means that the larger the inputs/outputs the smaller, percentage wise, the difference between the two approaches will be.

Do we want to bypass serialization/deserialization in both approaches so we are comparing just the operation of the middleware in each? Perhaps we could pick an operation with `smithy.api#Unit` as the input and output?

I think we do want to quantify how much time is spent in middleware as a percentage of the total call time (more generally some sort of inter-event timing over the lifecycle with something like [tracing-timing](https://github.com/jonhoo/tracing-timing)). It'd help us decide whether were our focus should be. Perhaps I think this should be done as a separate performance analysis however?

jdisanti:
> Do we want to bypass serialization/deserialization in both approaches so we are comparing just the operation of the middleware in each?

For what I'm trying to do here, bypassing serialization/deserialization isn't going to matter much. I just want to make sure we don't regress significantly on performance by switching to the orchestrator, and I can do that by just comparing the overall times of the two (calling the same operation with the same input/response with the two different architectures).

In the longer term, we should definitely do further benchmarking and profiling to improve overall performance, but that's separate from the intent of this PR.

aws/sra-test/integration-tests/aws-sdk-s3/benches/middleware_vs_orchestrator.rs
```diff
@@ -0,0 +1,285 @@
...
+        .unwrap();
+}
+
+fn test_connection() -> DynConnector {
+    infallible_connection_fn(|req| {
+        assert_eq!(
+            "https://test-bucket.s3.us-east-1.amazonaws.com/?list-type=2&prefix=prefix~",
+            req.uri().to_string()
+        );
+        assert!(req.headers().contains_key("authorization"));
```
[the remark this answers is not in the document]
jdisanti: I think for the purposes of this specific benchmark, I want to keep these since I want to make sure we're actually comparing the same thing and that we're not comparing a broken orchestrator to a functioning middleware stack. Since this benchmark isn't trying to determine the actual overhead, but rather the difference in the overheads, I'm not worried about it.

hlbarber:
@jdisanti Gotcha. Out-of-curiosity I tried it and these are the results on my machine:

`Unit -> Unit`
```
middleware              time:   [37.380 µs 37.896 µs 38.608 µs]
                        change: [-31.112% -30.135% -29.258%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe

orchestrator            time:   [72.979 µs 73.321 µs 73.772 µs]
                        change: [-19.921% -19.247% -18.417%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe
```

`ListObjectsV2Request -> ListObjectsV2Output`

```
middleware              time:   [54.186 µs 54.959 µs 56.099 µs]
                        change: [+42.781% +44.289% +45.807%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) high mild
  5 (5.00%) high severe

orchestrator            time:   [90.730 µs 91.068 µs 91.474 µs]
                        change: [+23.407% +24.513% +25.558%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
```

Issue #2521, opened 2023-03-31
Improve the instrumentation generated server code


hlbarber:
At the moment `smithy-rs` server does not have great instrumentation coverage.

We should identify key points in the request/response lifecycle to emit events/open spans.

Care MUST be taken to ensure that we do not log anything that _could_ be sensitive.

david-perez:
Something to keep in mind, from https://github.com/awslabs/smithy-rs/pull/2524#issuecomment-1491955247

> We should also think about having a mechanism to test that the framework is error logging what we expect it to; error logs are very important.

> Doing so seems terribly complicated / janky. On one hand the easiest I've found is leveraging [`tracing_test`](https://docs.rs/tracing-test/latest/tracing_test/); on the other hand, we run integration tests for this binary by spawning a child process via `Command::cargo_bin("pokemon-service")`, so collecting `stdout`/`stderr` might suffice.

This would be an easy thing to test if we had a plugin to generate application binaries using generated SDKs.

PR #2597, opened 2023-04-19
`DEBUG`-log server request rejections

david-perez:
This commit logs server request rejections at the `DEBUG` level in an
operation's `FromRequest` implementation.

This commit is analogous to the one in PR #2524 for response rejections.
However, request rejections are _not_ errors, so they shouldn't be
logged at the `ERROR` level. Indeed, they happen every time the server
rejects a malformed request.

Prior to this commit, the `RuntimeError::NotAcceptable` variant was the
only `RuntimeError` variant that was manually constructed. This commit
makes it so that it now results from a conversion from a new
`RequestRejection::NotAcceptable` variant.

We now leverage `futures_util::future::TryFutureExt::map` to map a future
that uses `RequestRejection` as its error into a future that uses
`RuntimeError`, and centrally log the rejection there. `futures_util` is
already a transitive dependency of server SDKs (via e.g. `hyper` and
`tower`), so adding it is a direct dependency is not worse.

This helps with #2521.

----

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
@@ -137,6 +137,7 @@ class ServerHttpBoundProtocolTraitImplGenerator(
         "Cow" to RuntimeType.Cow,
         "DateTime" to RuntimeType.dateTime(runtimeConfig),
         "FormUrlEncoded" to ServerCargoDependency.FormUrlEncoded.toType(),
+        "FuturesUtil" to ServerCargoDependency.FuturesUtil.toType(),
         "HttpBody" to RuntimeType.HttpBody,
         "header_util" to RuntimeType.smithyHttp(runtimeConfig).resolve("header"),
         "Hyper" to RuntimeType.Hyper,
@@ -182,7 +183,7 @@ class ServerHttpBoundProtocolTraitImplGenerator(
                 rustTemplate(
                     """
                     if !#{SmithyHttpServer}::protocols::accept_header_classifier(request.headers(), ${contentType.dq()}) {
-                        return Err(#{RuntimeError}::NotAcceptable)
+                        return Err(#{RequestRejection}::NotAcceptable);
                     }
                     """,
                     *codegenScope,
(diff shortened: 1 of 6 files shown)


rust-runtime/aws-smithy-http-server/src/proto/aws_json/rejection.rs
```diff
@@ -18,6 +18,8 @@ pub enum ResponseRejection {
 pub enum RequestRejection {
     #[error("error converting non-streaming body to bytes: {0}")]
     BufferHttpBodyBytes(crate::Error),
+    #[error("request contains invalid value for `Accept` header")]
+    NotAcceptable,
```
david-perez, line 22: I'll make these take in the invalid `HeaderValue` in a follow-up PR.

crisidev (approved):
Looks good.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -252,13 +251,19 @@ class ServerHttpBoundProtocolTraitImplGenerator(
                             .await
                             .map_err(Into::into)
                     };
+                    use #{FuturesUtil}::future::FutureExt;
```
hlbarber, line 254: Can we use `TryFutureExt::map_err`?

hlbarber (approved):
LGTM

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
david-perez: Nice, TIL about `TryFutureExt::map_err`. I switched over to that in https://github.com/awslabs/smithy-rs/pull/2597/commits/e61ed468c90299e485a4c62495b9337922de048d.

david-perez:
Oops, I realized I copy-pasted the `tracing::error!` from response rejection logging, contrary to what the commit message says 😅. I fixed that in https://github.com/awslabs/smithy-rs/pull/2597/commits/62fe0f13e9de4c707c17ecb0cddb2db5c4a0ec1d.

crisidev:
> Oops, I realized I copy-pasted the `tracing::error!` from response rejection logging, contrary to what the commit message says sweat_smile. I fixed that in [62fe0f1](https://github.com/awslabs/smithy-rs/commit/62fe0f13e9de4c707c17ecb0cddb2db5c4a0ec1d).

I was wondering about that :)

PR #2599, opened 2023-04-19
Fix `rewrite_base_url`: don't discard query params

jjant:
## Description
The previous version of this function discarded query params, which meant that if our model had a `@httpQuery` bound input, those would not be sent to the server and the requests would fail.

## Testing
Ran existing tests

----

### DIFF
--- a/examples/pokemon-service-common/src/lib.rs
+++ b/examples/pokemon-service-common/src/lib.rs
@@ -46,7 +46,7 @@ pub fn rewrite_base_url(
         let mut uri_parts = http_req.uri().clone().into_parts();
         uri_parts.authority = Some(authority.clone());
         uri_parts.scheme = Some(scheme.clone());
-        *http_req.uri_mut() = Uri::from_parts(uri_parts).expect("valid uri");
+        *http_req.uri_mut() = Uri::from_parts(uri_parts).expect("failed to create uri from parts");
         req
     }
 }
--- a/examples/pokemon-service-common/src/lib.rs
+++ b/examples/pokemon-service-common/src/lib.rs
@@ -17,6 +17,10 @@ use std::{
 use async_stream::stream;
 use aws_smithy_http::operation::Request;
 use aws_smithy_http_server::Extension;
+use http::{
+    uri::{Authority, Scheme},
+    Uri,
+};
 use pokemon_service_server_sdk::{
     error, input, model, model::CapturingPayload, output, types::Blob,
 };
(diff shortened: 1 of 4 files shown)


examples/pokemon-service-common/tests/plugins_execution_order.rs
```diff
@@ -64,10 +64,7 @@ struct SentinelPlugin {
 
 impl SentinelPlugin {
     pub fn new(name: &'static str, output: Arc<Mutex<Vec<&'static str>>>) -> Self {
-        Self {
-            name,
-            output: output,
-        }
+        Self { name, output }
```
jjant, line 67: Unrelated clippy lint fix.

david-perez: Isn't clippy run in CI already? https://github.com/awslabs/smithy-rs/blob/c3691087d434a886a259399cea4a9d15b0780a06/examples/Makefile#L23-L24

examples/pokemon-service-common/src/lib.rs
```diff
@@ -33,11 +37,16 @@ const PIKACHU_JAPANESE_FLAVOR_TEXT: &str =
...
+    authority: Authority,
+) -> impl Fn(Request) -> Request + Clone {
     move |mut req| {
         let http_req = req.http_mut();
-        let uri = format!("{base_url}{}", http_req.uri().path());
-        *http_req.uri_mut() = uri.parse().unwrap();
+        let mut uri_parts = http_req.uri().clone().into_parts();
+        uri_parts.authority = Some(authority.clone());
+        uri_parts.scheme = Some(scheme.clone());
+        *http_req.uri_mut() = Uri::from_parts(uri_parts).expect("valid uri");
```
david-perez, line 49: You're being misguided here and in the below usages by `.expect()`'s unfortunate naming into thinking that the message here is expressing what you expect, whereas this is supposed to be an error message :) See https://github.com/chris-morgan/rfcs/blob/unwrap_or_panic/text/0000-rename-expect.md#rationale-and-alternatives:

> [...] the name expect isn’t great on a couple of axes:
> It guides you in the direction of writing the message back to front, expressing what you expected to be the case, whereas it is supposed to be an error message when the expectation failed, meaning what went wrong, which is often expressed as the exact opposite.

david-perez (approved):

examples/pokemon-service-common/tests/plugins_execution_order.rs
jjant: Yeah, I guess I have a newer version locally.

PR #2602, opened 2023-04-19
Prevent build stalling with overlapping projections

hlbarber:
## Motivation and Context

Smithy CLI runs projections concurrently, for some currently unknown reason this causes the build to stall.

## Description

- Switch `defaultRustMetadata` in `BaseSymbolMetadataProvider` from a `val` to a `fun`.
- Remove unused `RustType::TestModule` companion object.

## Notes

A deadlock occurs when the client/server plugins both attempt to grab file locks. While this fixes this deadlock it is unknown whether there still exist conditions where it can happen.

## Testing

1. Checkout https://github.com/crisidev/smithy-rs-pokemon-service/commit/d276bb95bbd7bf5c5f029953b8f16cecfa8af24f, run `./gradlew assemble`, and observe the build halting.
2. Run the following commands to switch to this branch:
```bash
git submodule update --init --recursive --remote
cd smithy-rs
git checkout harryb/remove-build-deadlock
cd ..
./gradlew assemble
```
3. Observe the build succeeding.


crisidev (approved):
While this confuses the heck out of me, LGTM.

jdisanti (approved):

Velfi (approved):

PR #2603, opened 2023-04-19
Add credentials exposure test & fix STS + SSO

rcoh:
## Motivation and Context
- credentials providers may leak credentials in the HTTP body at the debug level

## Description
This adds a test to aws-config that looks for leaked credentials in all of our provider integration tests—since these test use AWS APIs under the hood, this also serves to test AWS services in general.

To support this, `sensitive` was added to the ParseHttpResponse trait and code was generated to take action based on this change.

- [x] Add environment variable to force logging of the body
- [x] consider if we want to suppress request body logging as well

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/lib.rs
+++ b/aws/rust-runtime/aws-config/src/lib.rs
@@ -374,7 +374,7 @@ mod loader {
         ///
         /// # Example: Using a custom profile file path
         ///
-        /// ```
+        /// ```no_run
         /// use aws_config::profile::{ProfileFileCredentialsProvider, ProfileFileRegionProvider};
         /// use aws_config::profile::profile_file::{ProfileFiles, ProfileFileKind};
         ///
@@ -417,7 +417,7 @@ mod loader {
         ///
         /// # Example: Using a custom profile name
         ///
-        /// ```
+        /// ```no_run
         /// use aws_config::profile::{ProfileFileCredentialsProvider, ProfileFileRegionProvider};
         /// use aws_config::profile::profile_file::{ProfileFiles, ProfileFileKind};
         ///
--- a/aws/rust-runtime/aws-config/src/test_case.rs
+++ b/aws/rust-runtime/aws-config/src/test_case.rs
@@ -16,11 +16,17 @@ use serde::Deserialize;
 use crate::connector::default_connector;
 use aws_smithy_types::error::display::DisplayErrorContext;
 use std::collections::HashMap;
+use std::env;
 use std::error::Error;
 use std::fmt::Debug;
 use std::future::Future;
+use std::io::Write;
 use std::path::{Path, PathBuf};
+use std::sync::{Arc, Mutex};
 use std::time::{Duration, UNIX_EPOCH};
+use tracing::dispatcher::DefaultGuard;
+use tracing::Level;
+use tracing_subscriber::fmt::TestWriter;
 
 /// Test case credentials
 ///
@@ -84,7 +90,7 @@ impl AsyncSleep for InstantSleep {
     }
 }
 
-#[derive(Deserialize)]
+#[derive(Deserialize, Debug)]
 pub(crate) enum GenericTestResult<T> {
     Ok(T),
     ErrorContains(String),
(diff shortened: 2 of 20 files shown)


aws/rust-runtime/aws-config/test-data/profile-provider/credential_process/fs/home/.aws/config
```diff
@@ -1,6 +1,7 @@
 [default]
 source_profile = base
-credential_process = echo '{ "Version": 1, "AccessKeyId": "ASIARTESTID", "SecretAccessKey": "TESTSECRETKEY", "SessionToken": "TESTSESSIONTOKEN", "Expiration": "2022-05-02T18:36:00+00:00" }'
+# base64 encoded to prevent triggering the secret scanner
```
jdisanti, line 3: Reworded to make this sound less scary:
```suggestion
# these fake credentials are base64 encoded to prevent triggering the unit test that scans logs for secrets
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/sso/SSODecorator.kt
```diff
@@ -0,0 +1,32 @@
...
+import java.util.logging.Logger
+
+class SSODecorator : ClientCodegenDecorator {
+    override val name: String = "SSO"
+    override val order: Byte = 0
+    private val logger: Logger = Logger.getLogger(javaClass.name)
+
+    private fun isAwsCredentials(shape: Shape): Boolean = shape.id == ShapeId.from("com.amazonaws.sso#RoleCredentials")
+
+    override fun transformModel(service: ServiceShape, model: Model): Model =
```
jdisanti, line 26: Consider short circuiting for services that aren't SSO so that this thing isn't wasting time scanning over every shape in every other service model.

It would also be reassuring to have a sanity check that it found at least one `RoleCredentials` when it is the SSO service model.

jdisanti (commented):
Looks good!

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/sso/SSODecorator.kt
rcoh: this is used with the service selector when it's injected in the main list

ysaito1001 (approved):
Great PR, handling the business of logging sensitive info holistically both in models and AWS runtime crates.

aws/rust-runtime/aws-config/src/test_case.rs
```diff
@@ -129,6 +135,79 @@ pub(crate) struct Metadata {
...
+    quiet: bool,
+    inner: W,
+}
+
+/// Capture logs from this test.
+///
+/// The logs will be captured until the `DefaultGuard` is dropped.
+///
+/// *Why use this instead of traced_test?*
+/// This captures _all_ logs, not just logs produced by the current crate.
```
Velfi, line 149: ty for adding this comment! It's exactly what I would have asked.

aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
```diff
@@ -256,7 +256,7 @@ fn calculate_signing_headers<'a>(
     // Step 4: https://docs.aws.amazon.com/en_pv/general/latest/gr/sigv4-add-signature-to-request.html
     let values = creq.values.as_headers().expect("signing with headers");
     let mut headers = HeaderMap::new();
-    add_header(&mut headers, header::X_AMZ_DATE, &values.date_time);
+    add_header(&mut headers, header::X_AMZ_DATE, &values.date_time, false);
```
Velfi, line 259: It may be worth creating an enum for `is_sensitive` in order to make this easier to understand.

Velfi (approved):

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/sso/SSODecorator.kt
jdisanti: Oh, you're right. I completely missed that.

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -417,7 +417,7 @@ mod loader {
         ///
         /// # Example: Using a custom profile name
         ///
-        /// ```
+        /// ```no_run
```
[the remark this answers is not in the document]
rcoh: we shouldn't be running doc tests ever, these were misses. They should only be compiled—running them will try to load up an SSL chain etc. and makes them very slow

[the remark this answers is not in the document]
Velfi: We should probably do a pass (as a separate PR) and check that we aren't running them elsewhere.

PR #2606, opened 2023-04-20
[Python] Improve Python stubs generation

crisidev:
## Motivation and Context
This PR improves the Python stubs generation.

## Description
The main change is about avoiding to setup a placeholder for the Python module and use the real module name, which allows to generate correct docstrings during codegeneration.

We also change the stubs layout on disk, with the main stub entrypoint called `__init__.pyi` instead of `$module_name.pyi`.

The README from the Rust runtime crate has been moved completely to the example folder and I run autoformatting and style checks on the Python example code.

----


examples/python/README.md
```diff
@@ -26,22 +26,28 @@ build.
 
 Ensure these dependencies are installed.
 
-```
+```bash
 pip install maturin uvloop aiohttp mypy
 ```
 
+The server can depend on [uvloop](https://pypi.org/project/uvloop/) for a faster
+event loop implementationand it will be automatically used instead of the standard
```
unexge, line 34:
```suggestion
event loop implementation and it will be automatically used instead of the standard
```

unexge (approved):

PR #2621, opened 2023-04-21
feature: orchestrator retry classifiers

Velfi:
## Motivation and Context

To retry a response, we must first classify it as retryable.

## Description

feature: add AWS error code classifier
feature: add x-amz-retry-after header classifier
feature: add smithy modeled retry classifier
feature: add error type classifier
feature: add HTTP status code classifier
add: tests for classifiers
remove: redundant `http` dep from `aws-http`
move: `NeverRetryStrategy` to smithy-runtime crate
add: RuntimePluginImpls codegen section for operation-specific runtime plugin definitions
update: orchestrator retries to work with `ShouldAttempt`
add: retry classifier config bag accessor
add: raw response getter to SdkError
update: RetryStrategy trait signatures to use `ShouldAttempt`
add: `RetryClassifiers` struct for holding and calling retry classifiers
update: `RetryClassifierDecorator` to define orchestrator classifiers
add: `default_retry_classifiers` fn to codegen
update: `ServiceGenerator` to add feature flag for aws-smithy-runtime/test-util
update: SRA integration test to insert retry classifier plugin

## Testing

this change includes tests

----


aws/rust-runtime/aws-http/Cargo.toml
```diff
@@ -28,7 +28,6 @@ aws-smithy-checksums = { path = "../../../rust-runtime/aws-smithy-checksums" }
 aws-smithy-protocol-test = { path = "../../../rust-runtime/aws-smithy-protocol-test" }
 bytes-utils = "0.1.2"
 env_logger = "0.9"
-http = "0.2.3"
```
Velfi, line 31: I noticed that this was in both the regular and dev dependencies so I removed it from dev dependencies.

aws/sra-test/integration-tests/aws-sdk-s3/tests/sra_test.rs
```diff
@@ -261,6 +263,19 @@ async fn sra_manual_test() {
 "#).unwrap(),
             )]);
 
+    struct ListObjectsV2RetryClassifiers;
+
+    // TODO(orchestrator) perform this step in Codegen
```
jdisanti, line 268: I've been tagging these with the feature flag
```suggestion
    // TODO(enableNewSmithyRuntime): perform this step in Codegen
```

Is this difficult to code generate? Seems like it should be pretty easy to add to `OperationRuntimePluginGenerator`

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationRuntimePluginGenerator.kt
```diff
@@ -83,6 +88,8 @@ class OperationRuntimePluginGenerator(
                     Ok(())
                 }
             }
+
+            #{runtime_plugin_impls}
```
jdisanti, line 92: `_impls` doesn't seem like the right suffix... `_supporting_types` maybe?

rust-runtime/aws-smithy-runtime-api/src/client/retries.rs
```diff
@@ -3,32 +3,79 @@
...
+
+        for retry_classifier in self.inner.iter() {
+            retry_reason = retry_classifier.classify_retry(error);
+
+            if retry_reason.is_some() {
+                break;
+            }
+        }
+
+        retry_reason
```
jdisanti, line 79: Minor simplification:
```suggestion
        for retry_classifier in &self.inner {
            if let Some(retry_reason) = retry_classifier.classify_retry(error) {
                return retry_reason;
            }
        }

        None
```

rust-runtime/aws-smithy-runtime/src/client/retries/classifier.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+const TRANSIENT_ERROR_STATUS_CODES: &[u16] = &[500, 502, 503, 504];
+
+/// A retry classifier that will treat HTTP response with those status codes as retryable.
+/// The `Default` version will retry 500, 502, 503, and 504 errors.
+#[derive(Debug)]
+pub struct HttpStatusCodeClassifier {
+    retryable_status_codes: Vec<u16>,
```
jdisanti, line 53: This could be more memory efficient by making it a `Cow<'static, [u16]>` since the majority of the time it will be a constant.

jdisanti (commented):
I think I had a little confusion about what's in scope and not in scope for this PR while reviewing... feel free to push back on comments if they're intended to be done in a follow up.

rust-runtime/aws-smithy-runtime-api/src/client/retries.rs
ysaito1001: Or possibly:
```rust
fn classify_retry(&self, error: &Error) -> Option<RetryReason> {
    self.inner
        .iter()
        .find_map(|classifier| classifier.classify_retry(error))
}
```
?

ysaito1001: Or possibly:
```rust
fn classify_retry(&self, error: &Error) -> Option<RetryReason> {
    self.inner
        .iter()
        .find_map(|classifier| classifier.classify_retry(error))
}
```
?

rust-runtime/aws-smithy-runtime-api/src/client/retries.rs
```diff
@@ -3,32 +3,79 @@
...
+pub struct RetryClassifiers {
+    inner: Vec<Box<dyn ClassifyRetry>>,
+}
 
-impl NeverRetryStrategy {
+impl RetryClassifiers {
     pub fn new() -> Self {
-        Self {}
+        Self {
+            inner: Vec::with_capacity(1),
```
ysaito1001, line 53: Is it worth adding a comment for why the capacity is 1 (unless it's obvious)?

ysaito1001 (commented):
Looks great! I'll take a second pass, leaving minor comments.

rust-runtime/aws-smithy-runtime-api/src/client/retries.rs
Velfi: It's always expected that at least one classifier will be defined, so we eagerly allocate for it.

I'll add a comment for this.

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -61,9 +62,16 @@ pub async fn invoke(
         let retry_strategy = cfg.retry_strategy();
         match retry_strategy.should_attempt_initial_request(cfg) {
             // Yes, let's make a request
-            Ok(_) => {}
+            Ok(ShouldAttempt::Yes) => {}
+            // No, this request shouldn't be retried
+            Ok(ShouldAttempt::No) => {
+                todo!("Create a new dispatch error for this case.")
```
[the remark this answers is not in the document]
Velfi: For this one, I'm not sure if we'll ever hit this case. I'm expecting that when we don't want to send an initial request, it'll always have some "reasoning" message. If we just get a "No" back, then we can't say why we shouldn't attempt the request. I'll add a provisional error message for this, but I think I may mark it as unreachable in a later PR and add a message about how retry strategies should always either return:
- yes
- yes, after a delay
- no, because <ERROR>

and that the "No, for no reason" case is a panic. I'm open to other ideas though.

aws/sra-test/integration-tests/aws-sdk-s3/tests/sra_test.rs
Velfi: You're right. IDK why I didn't do this initially. I have now added codegen to set this.

rust-runtime/aws-smithy-runtime-api/src/client/retries.rs
Velfi: TIL about `find_map`! That's exactly what I wanted; Thanks @ysaito1001.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -84,10 +84,10 @@ class EndpointParamsInterceptorGenerator(
             let input = context.input()?;
             let _input = input
                 .downcast_ref::<${operationInput.name}>()
-                .ok_or_else(|| "failed to downcast to ${operationInput.name}")?;
+                .ok_or("failed to downcast to ${operationInput.name}")?;
```
ysaito1001, line 87: Thanks for updating these. For some reason, I misremembered and thought `.ok_or` would be yelled at by `clippy`.

ysaito1001 (approved):

PR #2643, opened 2023-04-27
Recalculate signing time for every retry attempt

jdisanti:
## Motivation and Context
When SigV4 signing was ported over to the orchestrator, the request time got calculated once before the retry loop, which was incorrect. This PR moves that request time calculation into the request signer so that it happens on every attempt (unless there is a configured request time override).

This PR also refactors auth to use the `ConfigBag` instead of a separate signing properties `PropertyBag`.

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/auth.rs
+++ b/aws/rust-runtime/aws-runtime/src/auth.rs
@@ -11,11 +11,10 @@ pub mod sigv4 {
         SignableRequest, SignatureLocation, SigningParams, SigningSettings,
         UriPathNormalizationMode,
     };
-    use aws_smithy_http::property_bag::PropertyBag;
+    use aws_smithy_runtime_api::client::auth::{AuthSchemeId, HttpAuthScheme, HttpRequestSigner};
     use aws_smithy_runtime_api::client::identity::{Identity, IdentityResolver, IdentityResolvers};
-    use aws_smithy_runtime_api::client::orchestrator::{
-        BoxError, HttpAuthScheme, HttpRequest, HttpRequestSigner,
-    };
+    use aws_smithy_runtime_api::client::orchestrator::{BoxError, ConfigBagAccessors, HttpRequest};
+    use aws_smithy_runtime_api::config_bag::ConfigBag;
     use aws_types::region::SigningRegion;
     use aws_types::SigningService;
     use std::time::{Duration, SystemTime};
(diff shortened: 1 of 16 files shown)


aws/rust-runtime/aws-runtime/src/auth.rs
```diff
@@ -168,11 +164,11 @@ pub mod sigv4 {
...
             credentials: &'a Credentials,
             operation_config: &'a SigV4OperationSigningConfig,
+            request_timestamp: SystemTime,
         ) -> SigningParams<'a> {
             if let Some(expires_in) = settings.expires_in {
                 if let Some(creds_expires_time) = credentials.expiry() {
-                    let presigned_expires_time =
-                        operation_config.signing_options.request_timestamp + expires_in;
+                    let presigned_expires_time = request_timestamp + expires_in;
                     if presigned_expires_time > creds_expires_time {
```
Velfi, line 172: Possible alternate word: [expiry](https://www.merriam-webster.com/dictionary/expiry)
```suggestion
                    let presigned_expiry = request_timestamp + expires_in;
                    if presigned_expiry > creds_expiry {
```

Velfi (approved):

rust-runtime/aws-smithy-runtime-api/src/client/auth/option_resolver.rs
```diff
@@ -3,22 +3,21 @@
...
+use crate::client::orchestrator::BoxError;
 use std::borrow::Cow;
 
 /// New-type around a `Vec<HttpAuthOption>` that implements `AuthOptionResolver`.
 ///
 /// This is useful for clients that don't require `AuthOptionResolverParams` to resolve auth options.
 #[derive(Debug)]
 pub struct AuthOptionListResolver {
-    auth_options: Vec<HttpAuthOption>,
+    auth_options: Vec<AuthSchemeId>,
```
rcoh, line 15: - might consider naming this something like `StaticAuthOptionResolver` or something
- should this field be called `scheme_ids` now?

rcoh (approved):

rust-runtime/aws-smithy-runtime-api/src/client/auth/option_resolver.rs
jdisanti: > might consider naming this something like StaticAuthOptionResolver or something
That's a good idea. Will do.

> should this field be called scheme_ids now?

They are still the auth options, even if they're represented by scheme IDs.

PR #2653, opened 2023-04-27
implement `Ord` and `PartialOrd` for `DateTime`

### LINKED ISSUE #2406: Implement `Ord` on `aws_smithy_types::date_time::DateTime`
https://docs.rs/aws-smithy-types/latest/aws_smithy_types/date_time/struct.DateTime.html

henriiik:
## Motivation and Context

This change will allow easy sorting or comparing anything that contains a DateTime. My example is wanting to sort a list of S3 objects by last modified.

This PR fixes #2406

## Description
It's a pretty small PR, it implements the `Ord` trait for `DateTime` by comparing the `seconds` property of `self` and `other`, if they are equal it compares the `subsec_nanos` properties instead.

The `PartialOrd` trait is implemented by calling the `Ord` trait.

## Testing
I added a unit test that compares a number of `DateTime` values with different combinations of positive/zero/negative `seconds`/`subsec_nanos`.

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-types/src/date_time/mod.rs
+++ b/rust-runtime/aws-smithy-types/src/date_time/mod.rs
@@ -9,6 +9,7 @@ use crate::date_time::format::rfc3339::AllowOffsets;
 use crate::date_time::format::DateTimeParseErrorKind;
 use num_integer::div_mod_floor;
 use num_integer::Integer;
+use std::cmp::Ordering;
 use std::convert::TryFrom;
 use std::error::Error as StdError;
 use std::fmt;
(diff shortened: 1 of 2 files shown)


Velfi (approved):
Thanks for submitting this.

rust-runtime/aws-smithy-types/src/date_time/mod.rs
```diff
@@ -301,6 +302,21 @@ impl From<SystemTime> for DateTime {
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for DateTime {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self.seconds.cmp(&other.seconds) {
+            Ordering::Equal => self.subsecond_nanos.cmp(&other.subsecond_nanos),
+            ordering => ordering,
```
rcoh, line 315: can you delegate to `as_nanos` instead? I'm 99% sure this is correct but the interaction of subsecond nanos negative dates is non trivial and if we use `as_nanos` it will definitely be correct

rcoh (changes requested):
looks great! couple of small changes

henriiik:
@rcoh Thank you for the feedback! I have created a new PR that addresses it: #2656

Whilst implementing the proptest i noticed that `Format::DateTime` does not support formatting nanoseconds which lead to false positives. So the protest uses miliseconds as input instead.

PR #2654, opened 2023-04-27
Adds an experimental store/load add to config bag

rcoh:
## Motivation and Context
- store multiple objects of the same type in the bag
- obtain mutable access to items in the bag

## Description
This adds `store_append` to enable storing multiple items of the same type in the bag and `get_mut`, `get_mut_or_else` and `get_mut_or_default` to reduce clones when possible

## Testing
UTs

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime-api/src/config_bag.rs
+++ b/rust-runtime/aws-smithy-runtime-api/src/config_bag.rs
@@ -9,9 +9,18 @@
 //! with the following properties:
 //! 1. A new layer of configuration may be applied onto an existing configuration structure without modifying it or taking ownership.
 //! 2. No lifetime shenanigans to deal with
-use aws_smithy_http::property_bag::PropertyBag;
+mod typeid_map;
+
+use crate::config_bag::typeid_map::TypeIdMap;
+
+use std::any::{type_name, Any, TypeId};
+use std::collections::HashSet;
+
 use std::fmt::{Debug, Formatter};
+use std::iter::Rev;
+use std::marker::PhantomData;
 use std::ops::Deref;
+use std::slice;
 use std::sync::Arc;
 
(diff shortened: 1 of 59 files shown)


rust-runtime/aws-smithy-runtime-api/src/config_bag.rs
```diff
@@ -59,42 +61,196 @@ impl Deref for FrozenConfigBag {
     }
 }
 
-pub trait Persist {
-    fn layer_name(&self) -> &'static str;
-    fn persist(&self, layer: &mut ConfigBag);
+pub(crate) mod value {
```
jdisanti, line 64: What is the purpose of the module?

rust-runtime/aws-smithy-runtime-api/src/config_bag.rs
```diff
@@ -310,55 +544,45 @@ mod test {
...
+        assert_eq!(open_bag.layers().count(), 4);
     }
 
     #[test]
-    fn persist_trait() {
-        #[derive(Debug, Eq, PartialEq, Clone)]
-        struct MyConfig {
-            a: bool,
-            b: String,
+    fn store_append() {
```
jdisanti, line 560: Should add a test for iteration if there isn't one already

jdisanti (approved):
Looks good! Couple of minor things

rust-runtime/aws-smithy-runtime-api/src/config_bag.rs
rcoh: keep value private while allowing it to be exposed in the trait

jdisanti: Can you add a comment?

jdisanti (approved):
LGTM!

PR #2656, opened 2023-04-27
Use `as_nanos` in `Ord` impl for `DateTime`

henriiik:
## Motivation and Context
This PR adresses on [feedback from my previous PR](https://github.com/awslabs/smithy-rs/pull/2653#pullrequestreview-1404515899).

## Description
- use `as_nanos` in `Ord` impl for `DateTime` instead of manual impl.

## Testing
- add proptest that checks that `Ord` impl matches RFC 3339 comparison.

## Checklist

----


jdisanti (approved):
This looks great. Thanks for fixing @rcoh's comment on #2653!

PR #2657, opened 2023-04-28
Change the orchestrator feature gate into a quad state

jdisanti:
## Motivation and Context
So far we have been testing with the orchestrator on or off, with "on" meaning that both the middleware and orchestrator implementations exist, and you have to opt into the orchestrator via a `send_v2` function call (instead of `send`). This approach makes it difficult to test against the orchestrator exclusively, and also test the existing integration tests with the orchestrator implementation.

This PR changes `enableNewSmithyRuntime` into a quad-state with:
- middleware
- both middleware and orchestrator, defaulting to middleware
- both middleware and orchestrator, defaulting to orchestrator
- orchestrator

This allows for generating a client with the orchestrator exclusively, or generating both and defaulting to the orchestrator, to reduce testing friction.

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpResponseChecksumDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpResponseChecksumDecorator.kt
@@ -35,7 +35,7 @@ class HttpResponseChecksumDecorator : ClientCodegenDecorator {
 
     // TODO(enableNewSmithyRuntime): Implement checksumming via interceptor and delete this decorator
     private fun applies(codegenContext: ClientCodegenContext, operationShape: OperationShape): Boolean =
-        !codegenContext.settings.codegenConfig.enableNewSmithyRuntime && operationShape.outputShape != ShapeId.from("com.amazonaws.s3#GetObjectOutput")
+        codegenContext.smithyRuntimeMode.generateMiddleware && operationShape.outputShape != ShapeId.from("com.amazonaws.s3#GetObjectOutput")
 
     override fun operationCustomizations(
         codegenContext: ClientCodegenContext,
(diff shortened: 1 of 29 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpResponseChecksumDecorator.kt
```diff
@@ -35,7 +35,7 @@ class HttpResponseChecksumDecorator : ClientCodegenDecorator {
 
     // TODO(enableNewSmithyRuntime): Implement checksumming via interceptor and delete this decorator
     private fun applies(codegenContext: ClientCodegenContext, operationShape: OperationShape): Boolean =
-        !codegenContext.settings.codegenConfig.enableNewSmithyRuntime && operationShape.outputShape != ShapeId.from("com.amazonaws.s3#GetObjectOutput")
+        codegenContext.smithyRuntimeMode.generateMiddleware && operationShape.outputShape != ShapeId.from("com.amazonaws.s3#GetObjectOutput")
```
ysaito1001, line 38: Is `generateMiddleware` intended as opposed to `exclusivelyGenerateMiddleware`?

ysaito1001 (approved):

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpResponseChecksumDecorator.kt
jdisanti: I needed to make the request one exclusive since it relies on the middleware property bag, which doesn't exist when generating the serializer for the orchestrator (its customization ends up customizing both serializer implementations). So the exclusivity was to prevent a compiler error when both implementations were enabled.

PR #2658, opened 2023-04-28
Generate endpoint params in the orchestrator codegen

### LINKED ISSUE #2644: Code-gen setters on endpoint ParamsBuilder within runtime plugins/interceptors
This issue keeps track of `TODO(EndpointResolver)`s left in #2577

jdisanti:
## Motivation and Context
This PR adds the codegen logic to generate endpoint parameters in the endpoint params interceptor.

Fixes #2644.

----

### DIFF
--- a/aws/sdk-adhoc-test/build.gradle.kts
+++ b/aws/sdk-adhoc-test/build.gradle.kts
@@ -46,7 +46,6 @@ val allCodegenTests = listOf(
             ,
             "codegen": {
                 "includeFluentClient": false,
-                "enableNewCrateOrganizationScheme": true
             },
             "customizationConfig": {
                 "awsSdk": {
@@ -63,7 +62,6 @@ val allCodegenTests = listOf(
             ,
             "codegen": {
                 "includeFluentClient": false,
-                "enableNewCrateOrganizationScheme": true
             },
             "customizationConfig": {
                 "awsSdk": {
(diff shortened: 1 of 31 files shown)


ysaito1001 (commented):
Thanks for implementing this! You can add a link to [this issue](https://github.com/awslabs/smithy-rs/issues/2644) and remove `TODO(EndpointResolver)` as part of this PR (I think there are two places).

Originally, the way I thought about implementing this was to do call setters on endpoint `ParamsBuilder` in two places: in `ServiceRuntimePlugin` and in `<Operation>EndpointParamsInterceptor`. In `ServiceRuntimePlugin`, we would only call setters whose passed-in values can be extracted from `handle.conf` (in other words, call setters for which we can get values from service config). Then in `<Operation>EndpointParamsInterceptor`, we pull out the half-baked `ParamsBuilder` from the config bag and call setters on it whose passed-in values can be extracted from the operation input.

In the PR, looks like we call all setters on `ParamsBuilder` in one place, i.e. `<Operation>EndpointParamsInterceptor` and do nothing in `ServiceRuntimePlugin`. I think that could simplify things, but at the same time, we need to put the handle into the bag as denoted as `HACK` because `<Operation>EndpointParamsInterceptor` needs to access service config related information.

jdisanti:
I'm thinking we should just do it all in one place. It will be possible to get rid of the hack once config is backed by the `ConfigBag`.

ysaito1001 (approved):

PR #2659, opened 2023-04-28
add: customization + interceptor for HttpChecksumRequired

Velfi:
This translates the HttpChecksumRequired code to an interceptor and mounts that interceptor for non-streaming operations with that trait. 


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpChecksumRequiredGenerator.kt
```diff
@@ -8,14 +8,18 @@ package software.amazon.smithy.rust.codegen.client.smithy.customizations
...
 import software.amazon.smithy.rust.codegen.core.rustlang.Writable
+import software.amazon.smithy.rust.codegen.core.rustlang.rust
 import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
 import software.amazon.smithy.rust.codegen.core.rustlang.writable
 import software.amazon.smithy.rust.codegen.core.smithy.CodegenContext
 import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
 import software.amazon.smithy.rust.codegen.core.smithy.customize.OperationCustomization
 import software.amazon.smithy.rust.codegen.core.smithy.customize.OperationSection
 import software.amazon.smithy.rust.codegen.core.smithy.generators.operationBuildError
+import software.amazon.smithy.rust.codegen.core.util.PANIC
```
jdisanti, line 22:
```suggestion
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpChecksumRequiredGenerator.kt
```diff
@@ -54,6 +58,37 @@ class HttpChecksumRequiredGenerator(
                     "BuildError" to codegenContext.runtimeConfig.operationBuildError(),
                 )
             }
+
+            else -> emptySection
+        }
+    }
+}
+
+class HttpChecksumRequiredInterceptorGenerator(
```
jdisanti, line 67: This should be called `HttpChecksumRequiredInterceptorCustomization`

rust-runtime/aws-smithy-runtime/src/client/interceptors/http_checksum_required.rs
```diff
@@ -0,0 +1,47 @@
...
+        ctx: &mut InterceptorContext<HttpRequest, HttpResponse>,
+        _cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        let req = ctx.request_mut()?;
+        let data = req
+            .body()
+            .bytes()
+            .expect("checksum can only be computed for non-streaming operations");
+
+        let checksum = Md5::digest(data);
```
jdisanti, line 37: Oh, interesting. This is the old checksum implementation before flexible checksums became a thing. That's why the S3 model doesn't have it anymore.

I think we still need to preserve it since it's part of the Smithy spec, but its test should move into the generic clients instead of S3.

Velfi: I still don't understand why the test isn't failing in man since S3 removed the attribute from their model.

jdisanti: The flexible checksums implementation preserves the functionality, so the outdated test doesn't fail.

PR #2660, opened 2023-04-28
Allow canary-runner to specify expected speech text by Transcribe

ysaito1001:
## Motivation and Context
During a release for `aws-sdk-rust`, we've found that the canary test received an improved speech text from Transcribe. This broke the canary test because the test hard-coded the string that should be expected by Transcribe. This PR updates the runner so that the expected speech text can be passed-in from outside of the program.

## Testing
Added unit tests, mainly for argument parsing.

## Checklist
If the next release is created off of the main branch (because it's a non-breaking release), we only need to merge this PR the main branch. If we need to release from a release branch at some point in the future, this PR will need to be duplicated and merged to the release branch as well.

----


tools/ci-cdk/canary-runner/src/run.rs
```diff
@@ -83,6 +83,11 @@ pub struct RunArgs {
     #[clap(long)]
     musl: bool,
 
+    /// Expected speech text generated by Transcribe. This needs to be passed-in
+    /// because it can change as the accuracy of generated text improves over time.
+    #[clap(long)]
+    expected_speech_text_by_transcribe: String,
```
jdisanti, line 89: I think we should still make this optional just so that it's still easy to run in local dev.

ysaito1001: That is useful for local development. Addressed in 4073c70. If `expected_speech_text_by_transcribe` is unspecified, the updated code does not set the environment variable `CANARY_EXPECTED_TRANSCRIBE_RESULT`. If it needs to be set, the argument must be passed-in.

jdisanti (approved):

PR #2663, opened 2023-05-01
Only include in versions.toml crates under the sdk directory

ysaito1001:
## Motivation and Context
We have found that `aws-sdk-rust` release-2023-05-01 brought in the `aws-wasm` crate to `versions.toml`. The crate comes from an example in the `aws-doc-sdk-examples` repository. This PR fixes gradle build in `sdk` so that the `publisher` only considers crates under the `sdk` directory when generating `versions.toml`.

## Testing
Manually ran
```
./gradlew :aws:sdk:assemble
```
and confirmed that `versions.toml` got created under `smithy-rs/aws/sdk/build/aws-sdk` ~~`smithy-rs/aws/sdk/build/aws-sdk/sdk`~~ and `README` was  successfully hydrated.

----

### DIFF
--- a/aws/sdk/build.gradle.kts
+++ b/aws/sdk/build.gradle.kts
@@ -360,7 +360,7 @@ tasks.register<ExecRustBuildTool>("hydrateReadme") {
     binaryName = "publisher"
     arguments = listOf(
         "hydrate-readme",
-        "--versions-manifest", sdkOutputDir.resolve("versions.toml").toString(),
+        "--versions-manifest", outputDir.resolve("versions.toml").toString(),
         "--msrv", getRustMSRV(),
         "--input", rootProject.projectDir.resolve("aws/SDK_README.md.hb").toString(),
         "--output", outputDir.resolve("README.md").absolutePath,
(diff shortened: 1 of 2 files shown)


aws/sdk/build.gradle.kts
```diff
@@ -360,7 +360,7 @@ tasks.register<ExecRustBuildTool>("hydrateReadme") {
     binaryName = "publisher"
     arguments = listOf(
         "hydrate-readme",
-        "--versions-manifest", outputDir.resolve("versions.toml").toString(),
+        "--versions-manifest", sdkOutputDir.resolve("versions.toml").toString(),
```
jdisanti, line 363: This path seems wrong. The `versions.toml` resides in the aws-sdk-rust repo root rather than in `sdk/`.

ysaito1001: Ah, thanks for catching. `publisher`'s subcommand `generate-version-manifest` needs to be able to specify different directories for input and output. Addressed in 7af15c5 and 7d54876.

PR #2671, opened 2023-05-04
Simplify event stream message signer configuration

jdisanti:
## Motivation and Context

This PR creates a `DeferredSigner` implementation that allows for the event stream message signer to be wired up by the signing implementation later in the request lifecycle rather than by adding an event stream signer method to the config.

Refactoring this brings the middleware client implementation closer to how the orchestrator implementation will work, which unblocks the work required to make event streams work in the orchestrator.

----


jdisanti:
This is a breaking change, but I don't think we should defer its release to the next breaking release because of how unlikely it is to impact anyone.

jdisanti:
I made the changes semver compliant. They're still behavioral breaking for anyone implementing a custom event stream signer, but I think that should be OK since the fix is easy, and the chances anyone is doing that are low.

CHANGELOG.next.toml
```diff
@@ -23,3 +23,19 @@ message = "Implement `Ord` and `PartialOrd` for `DateTime`."
 author = "henriiik"
 references = ["smithy-rs#2653"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
+
+[[smithy-rs]]
+message = """
+<details>
+<summary>Breaking change in how event stream signing works (click to expand more details)</summary>
+
+This change will only impact you if you are wiring up their own event stream signing/authentication scheme. If you're using `aws-sig-auth` to use AWS SigV4 event stream signing, then this change will **not** impact you.
```
Velfi, line 32: We should say in the summary how people can know if this affects them e.g. "This impacts you if <reasoning>".

Velfi (approved):

aws/rust-runtime/aws-sig-auth/src/event_stream.rs
```diff
@@ -15,6 +18,115 @@ use std::time::SystemTime;
 
 /// Event Stream SigV4 signing implementation.
 #[derive(Debug)]
+pub struct SigV4MessageSigner {
+    last_signature: String,
+    credentials: Credentials,
+    signing_region: SigningRegion,
+    signing_service: SigningService,
+    time: Option<SystemTime>,
+}
```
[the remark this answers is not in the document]
jdisanti: I made both implementations use it, so it has to remain in `aws-sig-auth` for now.

CHANGELOG.next.toml
jdisanti: It seems like a lot to include in the summary line.

82marbag (approved):

Issue #2676, opened 2023-05-05
UnknownOperationException from API Gateway + Lambda using lambda_http 0.8.0 and lambda_runtime 0.8


anthonydandrea:
I've created an API Gateway connected to a Rust Lambda following the Pokemon service guide. I've wrapped my Smithy service using the crates `lambda_http = 0.8.0` and `lambda_runtime = 0.8` like so:

```
  #[tokio::main]
  pub async fn main() {
     setup_tracing();

      let app = BrokerService::builder_without_plugins()                                                                                                                                                                         
          .do_nothing(do_nothing)
          .check_health(check_health)
          .build()
          .expect("failed to build an instance of BrokerService");

      let handler = LambdaHandler::new(app);
      let lambda = lambda_http::run(handler);

      if let Err(err) = lambda.await {
          eprintln!("lambda error: {}", err);
      }
  }
```

However, when I downgrade those lambda crates to `0.7.0` and `0.7` respectively, I receive a `200` response from both my routes.

david-perez:
Can reproduce. Adding two `tower_http::trace::TraceLayer::new_for_http()` layers, one before and one after wrapping the `Service` in `LambdaHandler` yields:

```
{"timestamp":"2023-05-05T17:03:53.336874Z","level":"DEBUG","fields":{"message":"started processing request"},"target":"tower_http::trace::on_request","span":{"method":"GET","uri":"https://redacted.execute-api.us-east-1.amazonaws.com/prod/do-nothing","version":"HTTP/1.1","name":"request"},"spans":[{"requestId":"bc321be8-58b7-4dfb-affe-255427fbd2ff","xrayTraceId":"Root=1-645536f9-2a73cf0f49c91ea55799e5d1;Parent=704b2a5c2460b6c7;Sampled=0","name":"Lambda runtime invoke"},{"method":"GET","uri":"https://redacted.execute-api.us-east-1.amazonaws.com/prod/do-nothing","version":"HTTP/1.1","name":"request"}]}
[...]

PR #2684, opened 2023-05-09
Instrument `LambdaHandler`'s' `convert_event` function

david-perez:
This would have come in handy when debugging #2676.

----

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/routing/lambda_handler.rs
+++ b/rust-runtime/aws-smithy-http-server/src/routing/lambda_handler.rs
@@ -47,10 +47,14 @@ where
 }
 
 /// Converts a `lambda_http::Request` into a `http::Request<hyper::Body>`
-/// Issue: <https://github.com/awslabs/smithy-rs/issues/1125>
+/// The body conversion would not be required if we were to tackle this issue:
+/// <https://github.com/awslabs/smithy-rs/issues/1125>
 ///
 /// While converting the event the [API Gateway Stage] portion of the URI
-/// is removed from the uri that gets returned as a new `http::Request`.
+/// is removed from the URI that gets returned as the new `http::Request`.
+///
+/// So, for example, if the `https://redacted.execute-api.us-east-1.amazonaws.com/prod/operation`
+/// was hit, then the raw HTTP path is just `/operation`, since `prod` is the API Gateway stage.
 ///
 /// [API Gateway Stage]: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-stages.html
 fn convert_event(request: Request) -> HyperRequest {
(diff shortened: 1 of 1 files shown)


rust-runtime/aws-smithy-http-server/src/routing/lambda_handler.rs
```diff
@@ -59,24 +63,28 @@ fn convert_event(request: Request) -> HyperRequest {
     let mut path = String::from(parts.uri.path());
 
     if !raw_path.is_empty() && raw_path != path {
+        tracing::trace!(raw_path, path, "rewriting request URI to use raw HTTP path");
```
hlbarber, line 66: Do we want to take this out of the `if` block?

hlbarber (approved):

rust-runtime/aws-smithy-http-server/src/routing/lambda_handler.rs
david-perez: Why? In this branch we're rewriting the URI, but in the `else` branch we're not. In both cases we log the values, but the message is different.

hlbarber: I didn't see the `tracing::trace!(raw_path, path, "not rewriting request URI");` in `else`. Ignore me.

hlbarber (commented):
On second thought, the URI here could be sensitive. So perhaps we shouldn't log it?

david-perez:
Let's change it so that the URI is logged only when the feature `unredacted-logging` is enabled.

hlbarber:
> Let's change it so that the URI is logged only when the feature unredacted-logging is enabled.

An alternative here is that we create a wrapper which is the union of all redactions that can be applied prior to routing. e.g. if operation A had path segment 0 sensitive and operation B had path segment 1 sensitive then prior to routing we redact both 0 and 1.

david-perez:
That would be tricky to implement correctly; it wouldn't be as simple as maintaining a list of indices on sensitive path segments, since [greedy labels](https://smithy.io/2.0/spec/http-bindings.html#greedy-labels) can capture a different number of path segments each time.

hlbarber:
> since [greedy labels](https://smithy.io/2.0/spec/http-bindings.html#greedy-labels) can capture a different number of path segments each time

Isn't the join of two [greedy labels](https://docs.rs/aws-smithy-http-server/latest/aws_smithy_http_server/instrumentation/sensitivity/uri/struct.GreedyLabel.html) the component wise `min` over the arguments `segment_index`, `end_offset`?

david-perez:
Yes that would work. E.g. with the two patterns:

1. `/prefix/{greedy1+}/suffix`
2. `/{greedy2+}/foo-bar-baz`

where the `greedy1` and `greedy2` labels are bound to `@sensitive` shapes, we get a component wise minimum of:

```
GreedyLabel {
    segment_index: 0,
    end_of_offset: 7,
}
```

So upon receiving a request URI like `/bound/to/greedy2/foo-bar-baz`, we'd only be able to print something like `/<redacted+>/<redacted>bar-baz`, which is a bit weird.

drganjoo (approved):

PR #2685, opened 2023-05-09
Bump `lambda_http` dependency of `aws-smithy-http-server` to 0.8.0

david-perez:
This patch also removes the unneeded dependency on `lambda_runtime` by
`aws-smithy-http-server-python`.

This patch also refactors `LambdaHandler`'s `convert_event` to avoid
cloning the URI path when not needed.

This is a breaking change. See #2676 why.

This patch also bumps `aws-smithy-http-server` dependency on `mime` to 0.3.4.

`cargo +nightly-2022-11-16 minimal-versions check --all-features` otherwise
fails when using 0.3.0, because we require `impl fmt::Display for
mime::FromStrError`, which was first introduced in 0.3.4.

As to why `minimal-versions` is now picking `mime` 0.3.0 with this patch, it's
because in `lambda_http` 0.7.3, they had [`mime =
"0.3.16"`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/99dba6447253ac87cf3cefeb2ba130b50514f9df/lambda-http/Cargo.toml#L35-L35),
and in `lambda_http` 0.8.0, they've now relaxed that to [`mime =
"0.3"`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/393d6447bea0502e1f939d197f4facc228e6e007/lambda-http/Cargo.toml#L36).

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -23,3 +23,9 @@ message = "Implement `Ord` and `PartialOrd` for `DateTime`."
 author = "henriiik"
 references = ["smithy-rs#2653"]
 meta = { "breaking" = false, "tada" = false, "bug" = false }
+
+[[smithy-rs]]
+message = "Bump dependency on `lambda_http` by `aws-smithy-http-server` to 0.8.0. This version of `aws-smithy-http-server` is only guaranteed to be compatible with 0.8.0, or semver-compatible versions of 0.8.0 of the `lambda_http` crate. It will not work with versions prior to 0.8.0 _at runtime_, making requests to your smithy-rs service unroutable, so please make sure you're running your service in a compatible configuration"
+author = "david-perez"
+references = ["smithy-rs#2676", "smithy-rs#2685"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
--- a/examples/pokemon-service-lambda/Cargo.toml
+++ b/examples/pokemon-service-lambda/Cargo.toml
@@ -13,7 +13,7 @@ hyper = {version = "0.14.25", features = ["server"] }
 tokio = "1.26.0"
 tracing = "0.1"
 
-lambda_http = "0.7.3"
+lambda_http = "0.8.0"
 
 # Local paths
 aws-smithy-http-server = { path = "../../rust-runtime/aws-smithy-http-server", features = ["aws-lambda"] }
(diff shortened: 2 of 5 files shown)


jjant (approved):

david-perez:
In CI, `cargo +nightly-2022-11-16 minimal-versions check --all-features` on `aws-smithy-http-server` [fails with](https://github.com/awslabs/smithy-rs/actions/runs/4927708586/jobs/8805103726?pr=2685#step:3:2375):

```
Checking mime v0.3.0
...
error[E0599]: the method `as_display` exists for reference `&FromStrError`, but its trait bounds were not satisfied
  --> aws-smithy-http-server/src/rejection.rs:18:13
   |
18 |     #[error("invalid `Content-Type` header value mime type: {0}")]
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method cannot be called on `&FromStrError` due to unsatisfied trait bounds
   |
  ::: /opt/cargo/registry/src/github.com-1ecc6299db9ec823/mime-0.3.0/src/lib.rs:72:1
   |
72 | pub struct FromStrError {
   | ----------------------- doesn't satisfy `FromStrError: std::fmt::Display`
   |
   = note: the following trait bounds were not satisfied:
           `FromStrError: std::fmt::Display`
           which is required by `&FromStrError: DisplayAsDisplay`
```

Indeed, we require `mime::FromStrError: std::fmt::Display`, but said trait was first implemented in [this commit](https://github.com/hyperium/mime/commit/1031854), which first made its way out in `mime` 0.3.4. As to why `minimal-versions` is now picking `mime` 0.3.0 with this patch, it's because in `lambda_http` 0.7.3, they had [`mime = "0.3.16"`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/99dba6447253ac87cf3cefeb2ba130b50514f9df/lambda-http/Cargo.toml#L35-L35), and in `lambda_http` 0.8.0, they've now relaxed that to [`mime = "0.3"`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/393d6447bea0502e1f939d197f4facc228e6e007/lambda-http/Cargo.toml#L36).

I'll go ahead and bump the dependency of `aws-smithy-http-server` on `mime` to 0.3.4 and update the PR description.

PR #2696, opened 2023-05-12
Fix client operation name collisions with the standard library prelude

jdisanti:
## Motivation and Context
Operations named `Send` or `Sync` (and probably others) were colliding with the types in the standard library prelude and causing compiler errors. This PR adds tests that include all the type names from the Rust prelude, and fixes the compiler errors they cause.

In the future, the `no_implicit_prelude` attribute can be added to certain code generated modules to better enforce that there can't be name collisions, but for now, the `tokio::test` macro doesn't compile with that attribute enabled (and likely other macros from other libraries).

## Checklist

----


codegen-client-test/build.gradle.kts
```diff
@@ -92,6 +92,22 @@ val allCodegenTests = "../codegen-core/common-test-models".let { commonModels ->
...
+            imports = listOf("$commonModels/naming-obstacle-course-prelude-ops.smithy"),
+        ),
+        CodegenTest(
+            "crate#Config",
+            "naming_test_prelude_structs",
+            """
+            , "codegen": { "renameErrors": false }
+            """.trimIndent(),
+            imports = listOf("$commonModels/naming-obstacle-course-prelude-structs.smithy"),
+        ),
```
ysaito1001, line 110: Great tests to add!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
```diff
@@ -193,14 +193,57 @@ data class RuntimeType(val path: String, val dependency: RustDependency? = null)
      * The companion object contains commonly used RuntimeTypes
      */
     companion object {
+        /** Scope that contains all Rust prelude types, but not macros or functions */
```
ysaito1001, line 196: Nice! Maybe it's useful to include a link to prelude contents like [this one?](https://doc.rust-lang.org/std/prelude/index.html#prelude-contents)

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -223,6 +223,7 @@ private fun renderCustomizableOperationSendMethod(
     val combinedGenerics = operationGenerics + handleGenerics
 
     val codegenScope = arrayOf(
+        *RuntimeType.preludeScope,
```
david-perez, line 226: I wonder if we should just merge this scope directly in `rustTemplate` and friends.

ysaito1001 (approved):
Thanks for this effort! To contribute, if we find more instances of unqualified prelude in the codebase, we can fix them on-the-fly as we go in respective PRs?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -112,12 +113,12 @@ class FluentClientGenerator(
                 }
 
                 #{client_docs:W}
-                ##[derive(std::fmt::Debug)]
+                ##[derive(::std::fmt::Debug)]
```
david-perez, line 116: Yeah I think prepending all types with `::` is a very good idea and something we should somehow enforce everywhere.

codegen-core/common-test-models/naming-obstacle-course-prelude-ops.smithy
```diff
@@ -0,0 +1,92 @@
...
+       ExactSizeIterator,
+       Option,
+       Some,
+       None,
+       Result,
+       Ok,
+       Err,
+       String,
+       ToString,
+       Vec,
```
david-perez, line 51: `TryFrom`, `TryInto`, and `FromIterator` are missing? https://doc.rust-lang.org/std/prelude/index.html

Would be nice to put the prelude link in these files.

We generate Rust 2021 edition crates.

codegen-core/common-test-models/naming-obstacle-course-prelude-structs.smithy
```diff
@@ -0,0 +1,125 @@
...
+use smithy.test#httpRequestTests
+use smithy.test#httpResponseTests
+use aws.protocols#awsJson1_1
+use aws.api#service
+use smithy.framework#ValidationException
+
+/// Confounds model generation machinery by using structs named after every item in the Rust prelude
+@awsJson1_1
+@service(sdkId: "Config")
+service Config {
```
david-perez, line 13: It feels like there's not a need for having these be two separate integration tests.

codegen-core/common-test-models/naming-obstacle-course-prelude-ops.smithy
```diff
@@ -0,0 +1,92 @@
...
+use smithy.test#httpRequestTests
+use smithy.test#httpResponseTests
+use aws.protocols#awsJson1_1
+use aws.api#service
+use smithy.framework#ValidationException
+
+/// Confounds model generation machinery by using operations named after every item in the Rust prelude
+@awsJson1_1
+@service(sdkId: "Config")
+service Config {
```
david-perez, line 13: These are the kinds of tests that I think are better done in `{client,server}IntegrationTests`. In Kotlin we can easily code generate Smithy models that more exhaustively test this e.g. union / enum shapes with names of types from the prelude are not covered.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
```diff
@@ -142,6 +143,21 @@ fun <T : AbstractCodeWriter<T>> T.conditionalBlock(
     return this
 }
 
+fun RustWriter.conditionalBlockTemplate(
```
david-perez, line 146: Thanks for adding this.

david-perez (approved):

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
jdisanti: I definitely thought about it. That and deleting `rust()` and renaming `rustTemplate()` to `rust()`. I'd like to get rid of the old style `#T` formatters entirely to tidy things up. All of this is quite a bit of work, and I wanted to get this initial fix out first.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
jdisanti: Fixed in 49d10aecb5d768cece726b2aaab87a1136852040

codegen-core/common-test-models/naming-obstacle-course-prelude-ops.smithy
jdisanti: Great catch! Fixed in 49d10aecb5d768cece726b2aaab87a1136852040

codegen-core/common-test-models/naming-obstacle-course-prelude-ops.smithy
jdisanti: That's a great idea. I refactored them into Kotlin unit tests in 49d10aecb5d768cece726b2aaab87a1136852040 and also added them to the server codegen test suite.

ysaito1001 (approved):
Re-approved after new commits. Looks fantastic! I really like how test models are now automatically derived from `preludeScope` and used by both client & server tests as per David's comment.

PR #2697, opened 2023-05-12
 Allow for configuring interceptors on generic client

ysaito1001:
## Description
This PR allows users to pass-in interceptors to a generic client. Client-level configured interceptors are eventually added to `client_interceptors` and operation-level interceptors to `operation_interceptors`, both of which are fields in the `aws-smithy-runtime-api::client::interceptors::Interceptors`. The relevant code is generated only in the orchestrator mode.

The SDK registers a default set of (client-level & operation-level) interceptors, and the passed-in interceptors by a user will run _after_ those default interceptors.

## Testing
- Added integration tests `operation_interceptor_test` and `interceptor_priority` in `sra_test`.

----


rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -216,7 +216,7 @@ async fn make_an_attempt(
     Ok(checkpoint)
 }
 
-#[cfg(all(test, feature = "test-util"))]
+#[cfg(all(test, feature = "test-util", feature = "anonymous-auth"))]
```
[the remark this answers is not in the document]
jdisanti: I don't think it's just for testing, but I'm not 100% on that yet. The question will be: is the absence of an auth scheme/option automatically anonymous, or should we keep this thing around? Probably the former so that a model can explicitly say that an operation doesn't require auth. However, I don't think Smithy has any way to communicate that currently.

aws/sra-test/integration-tests/aws-sdk-s3/tests/interceptors.rs
```diff
@@ -0,0 +1,120 @@
...
+                RequestTimeAdvanceInterceptor(Duration::from_secs(1624036048))
+            ))
+            .bucket("test-bucket")
+            .prefix("prefix~")
+            .send_orchestrator_with_plugin(Some(fixup))
+            .await
+    );
+    let resp = resp.expect("valid e2e test");
+    assert_eq!(resp.name(), Some("test-bucket"));
+    conn.full_validate(MediaType::Xml).await.expect("failed")
```
jdisanti, line 119: `expect("failed")`? I think this is flipped. Should be `expect("success")`.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/InterceptorConfigCustomization.kt
```diff
@@ -0,0 +1,193 @@
...
+        "SharedInterceptor" to interceptors.resolve("SharedInterceptor"),
+    )
+
+    override fun section(section: ServiceConfig) =
+        writable {
+            when (section) {
+                ServiceConfig.ConfigStruct -> rustTemplate(
+                    """
+                    // TODO(enableNewSmithyRuntime): Unused until we completely switch to the orchestrator
+                    ##[allow(dead_code)]
```
jdisanti, line 31: This customization shouldn't run unless we're generating the orchestrator. Otherwise, this will pull in the runtime crates.

It looks like this is done correctly where the customization is registered, so this `allow` shouldn't be necessary?

jdisanti (approved):
This looks great! Just some minor comments.

aws/sra-test/integration-tests/aws-sdk-s3/tests/interceptors.rs
ysaito1001: Ah, copied & pasted without realizing it. I've seen two camps for what to say in `.expect`: one camp for saying what should've happened and the other for saying what actually happened. I've been arbitrarily choosing one over the other whenever I use `.expect`, but saying what should've happened makes more sense to me now. Addressed in 44af0db.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/InterceptorConfigCustomization.kt
ysaito1001: Thanks for catching! Addressed in 73c3515.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationRuntimePluginGenerator.kt
```diff
@@ -30,12 +30,10 @@ sealed class OperationRuntimePluginSection(name: String) : Section(name) {
             val smithyRuntimeApi = RuntimeType.smithyRuntimeApi(runtimeConfig)
             writer.rustTemplate(
                 """
-                $interceptorName.register_operation_interceptor(std::sync::Arc::new(#{interceptor}) as _);
+                $interceptorName.register(#{SharedInterceptor}::new(#{interceptor}) as _);
```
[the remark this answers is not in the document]
ysaito1001: Makes sense. Addressed in 24e176d.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/ServiceRuntimePluginGenerator.kt
```diff
@@ -132,6 +130,12 @@ class ServiceRuntimePluginGenerator(
                     cfg.set_connection(connection);
 
                     #{additional_config}
+
+                    // Client-level Interceptors are registered after default Interceptors.
+                    self.handle.conf.interceptors.iter().for_each(|interceptor| {
+                        _interceptors.register(interceptor.clone());
+                    });
```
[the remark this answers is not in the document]
ysaito1001: I like how the updated code looks. Added the `extend` method in 9205f49.

aws/sra-test/integration-tests/aws-sdk-s3/tests/interceptors.rs
jdisanti: I take the stance of reading it as plain English: "do this action and _expect_ success"

PR #2699, opened 2023-05-12
Fix: orchestrator flow

Velfi:
## Motivation and Context

Necessary before we can implement retries.

## Description

I noticed that we weren't handling the flow quite right when errors occurred. This PR fixes that and adds interceptor-based tests to ensure things are working right. I still think we could use more tests but the PR is already quite large.

## Testing

This PR contains tests

----


Velfi:
I still have some rough edges to sand off, but the interceptor tests are passing.

Velfi:
Resolving the conflict with Yuki's changes seems a bit tough. I'll save that for next week.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context/wrappers.rs
```diff
@@ -0,0 +1,207 @@
...
+impl<'a, I, O, E> From<&'a mut InterceptorContext<I, O, E>>
+    for AfterDeserializationInterceptorContextMut<'a, I, O, E>
+{
+    fn from(inner: &'a mut InterceptorContext<I, O, E>) -> Self {
+        Self { inner }
+    }
+}
+
+impl<'a, I, O, E> AfterDeserializationInterceptorContextMut<'a, I, O, E> {
+    pub fn input_mut(&mut self) -> &mut I {
```
jdisanti, line 191: We should make a macro for each accessor method so that we don't have to repeat the documentation everywhere and worry about consistency. The `-Mut` impls should also have the non-mut functions, and a macro will probably make that easier.

rust-runtime/aws-smithy-runtime-api/src/type_erasure.rs
```diff
@@ -224,6 +224,25 @@ impl TypeErasedError {
     }
 }
 
+impl From<Box<dyn StdError + Send + Sync>> for TypeErasedError {
```
jdisanti, line 227: Needs tests.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors.rs
```diff
@@ -679,48 +697,124 @@ impl Interceptors {
...
+                }
+                result = Err(new_error);
+            }
+        }
+        result.map_err(InterceptorError::modify_before_completion)
+    }
+
+    pub fn read_after_execution(
+        &self,
+        ctx: &InterceptorContext<Input, Output, Error>,
```
jdisanti, line 804: Seems like the final hooks should also get a context wrapper to hide the internals, and that will make it possible to use the macro to make these.

jdisanti (commented):
Very nice work on this! It's a huge improvement.

rust-runtime/aws-smithy-runtime-api/src/type_erasure.rs
Velfi: It's good you suggested this b/c I discovered you can't downcast back to a `BoxError`.

I wrote this test but it doesn't compile. I'll see if I can think of another way to handle this.
```rust
    #[test]
    fn test_typed_erased_errors_can_be_downcast() {
        let test_err: BoxError = TestErr::new("something failed!").into();
        let type_erased_test_err: TypeErasedError = test_err.into();
        let actual = type_erased_test_err.downcast::<Box<dyn std::error::Error + Send + Sync + 'static>>().expect("type erased error can be downcast into original type");
        let expected: BoxError = TestErr::new("something failed!").into();
        assert_eq!(test_err.to_string(), actual.to_string());
    }
```

Velfi: I thought of a better way. Thanks for your advice John.

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator/error.rs
```diff
@@ -0,0 +1,117 @@
...
+use crate::client::interceptors::context::phase::Phase;
+use crate::client::interceptors::InterceptorError;
+use crate::client::orchestrator::HttpResponse;
+use crate::type_erasure::TypeErasedError;
+use aws_smithy_http::result::{ConnectorError, SdkError};
+use std::fmt::Debug;
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum OrchestratorError<E: Debug> {
```
jdisanti, line 16: @rcoh had a good idea: We can use `SdkError<E, ()>` inside of the orchestrator (could even do something like `type OrchestratorError = SdkError<Error, ()>`), and then map the response on that to a full `SdkError<E, HttpResponse>` when returning from `invoke`.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
```diff
@@ -575,75 +293,54 @@ impl AttemptCheckpoint {
...
-            .expect("at least one phase must be in progress");
-        let fresh_request =
-            Self::try_clone(self.checkpointed_request.as_ref().expect("checked above"))
-                .expect("cloneable request");
-        self.before_transmit = Some(InterceptorContext::<phase::BeforeTransmit>::new(
-            input,
-            fresh_request,
-        ));
+        self.request = try_clone(self.request_checkpoint.as_ref().expect("checked above"));
+        self.phase = Phase::BeforeTransmit;
```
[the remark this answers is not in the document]
Velfi: Good catch

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
```diff
@@ -575,75 +293,54 @@ impl AttemptCheckpoint {
...
-            .expect("builder has not been modified, headers must be valid") =
-            request.headers().clone();
-        Some(
-            cloned_request
-                .body(cloned_body)
-                .expect("a clone of a valid request should be a valid request"),
-        )
+    #[doc(hidden)]
+    pub fn is_failed(&self) -> bool {
+        self.is_failed
```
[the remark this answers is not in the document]
Velfi: There used to be, but at this point there isn't. I'll update this.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/error.rs
```diff
@@ -170,6 +176,12 @@ impl std::error::Error for InterceptorError {
     }
 }
 
+impl From<InterceptorError> for TypeErasedError {
```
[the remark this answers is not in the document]
Velfi: Yeah, I think this was a holdout. Thanks for spotting it.

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -70,134 +50,225 @@ pub async fn invoke(
...
-    handle_err!([checkpoint] => interceptors.modify_before_signing(checkpoint.before_transmit(), cfg));
-    handle_err!([checkpoint] => interceptors.read_before_signing(checkpoint.before_transmit(), cfg));
-
-    checkpoint = orchestrate_auth(checkpoint, cfg).await?;
-
-    handle_err!([checkpoint] => interceptors.read_after_signing(checkpoint.before_transmit(), cfg));
-    handle_err!([checkpoint] => interceptors.modify_before_transmit(checkpoint.before_transmit(), cfg));
-    handle_err!([checkpoint] => interceptors.read_before_transmit(checkpoint.before_transmit(), cfg));
+) {
+    halt_on_err!(
```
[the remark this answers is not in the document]
Velfi: TIL. I'll switch these over to that style in order to avoid the extra lines.

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator/error.rs
Velfi: I think that this change would be OK, but I want to hold off on making it until I get to the retry PRs. I think I'm getting the response from the context in those cases, but I can't remember rn.

jdisanti (approved):

PR #2704, opened 2023-05-15
Port Glacier customizations to the orchestrator

jdisanti:
## Motivation and Context
This PR gets the Glacier integration tests to pass against the orchestrator implementation by porting the Glacier customizations from `make_operation` changes to interceptors.

----

### DIFF
--- a/aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
+++ b/aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
@@ -7,7 +7,7 @@
 #![allow(dead_code)]
 
 use aws_runtime::auth::sigv4::SigV4OperationSigningConfig;
-use aws_sig_auth::signer::SignableBody;
+use aws_sigv4::http_request::SignableBody;
 use aws_smithy_http::body::SdkBody;
 use aws_smithy_http::byte_stream;
 use aws_smithy_runtime_api::client::interceptors::{
--- a/aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
+++ b/aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
@@ -1,0 +1,388 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// This code is referenced in generated code, so the compiler doesn't realize it is used.
+#![allow(dead_code)]
+
+use aws_runtime::auth::sigv4::SigV4OperationSigningConfig;
+use aws_sig_auth::signer::SignableBody;
+use aws_smithy_http::body::SdkBody;
+use aws_smithy_http::byte_stream;
+use aws_smithy_runtime_api::client::interceptors::{
+    BeforeSerializationInterceptorContextMut, BeforeTransmitInterceptorContextMut, BoxError,
+    Interceptor,
+};
+use aws_smithy_runtime_api::client::orchestrator::LoadedRequestBody;
+use aws_smithy_runtime_api::config_bag::ConfigBag;
+use bytes::Bytes;
+use http::header::{HeaderName, HeaderValue};
+use http::Request;
+use ring::digest::{Context, Digest, SHA256};
+use std::fmt;
+use std::marker::PhantomData;
+
+/// The default account ID when none is set on an input
+const DEFAULT_ACCOUNT_ID: &str = "-";
+
+const TREE_HASH_HEADER: &str = "x-amz-sha256-tree-hash";
+const X_AMZ_CONTENT_SHA256: &str = "x-amz-content-sha256";
+const API_VERSION_HEADER: &str = "x-amz-glacier-version";
+
+/// Adds an account ID autofill method to generated input structs
+///
+/// Some Glacier operations have an account ID field that needs to get defaulted to `-` if not set.
+/// This trait is implemented via codegen customization for those operation inputs so that
+/// the [`GlacierAccountIdAutofillInterceptor`] can do this defaulting.
+pub(crate) trait GlacierAccountId: fmt::Debug {
+    /// Returns a mutable reference to the account ID field
+    fn account_id_mut(&mut self) -> &mut Option<String>;
+
+    /// Autofills the account ID with the default if not set
+    fn autofill_account_id(&mut self) {
+        let account_id = self.account_id_mut();
+        if account_id.as_deref().unwrap_or_default().is_empty() {
+            *account_id = Some(DEFAULT_ACCOUNT_ID.into());
+        }
+    }
+}
+
+/// Autofills account ID input fields with a default if no value is set
+#[derive(Debug)]
+pub(crate) struct GlacierAccountIdAutofillInterceptor<I> {
+    _phantom: PhantomData<I>,
+}
+
+impl<I> GlacierAccountIdAutofillInterceptor<I> {
+    /// Constructs a new [`GlacierAccountIdAutofillInterceptor`]
+    pub(crate) fn new() -> Self {
+        Self {
+            _phantom: Default::default(),
+        }
+    }
+}
+
+impl<I: GlacierAccountId + Send + Sync + 'static> Interceptor
+    for GlacierAccountIdAutofillInterceptor<I>
+{
+    fn modify_before_serialization(
+        &self,
+        context: &mut BeforeSerializationInterceptorContextMut<'_>,
+        _cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        let erased_input = context.input_mut();
+        let input: &mut I = erased_input
+            .downcast_mut()
+            .expect("typechecked at registration");
+        input.autofill_account_id();
+        Ok(())
+    }
+}
+
+/// Attaches the `x-amz-glacier-version` header to the request
+#[derive(Debug)]
+pub(crate) struct GlacierApiVersionInterceptor {
+    api_version: &'static str,
+}
+
+impl GlacierApiVersionInterceptor {
+    /// Constructs a new [`GlacierApiVersionInterceptor`] with the given API version.
+    pub(crate) fn new(api_version: &'static str) -> Self {
+        Self { api_version }
+    }
+}
+
+impl Interceptor for GlacierApiVersionInterceptor {
+    fn modify_before_signing(
+        &self,
+        context: &mut BeforeTransmitInterceptorContextMut<'_>,
+        _cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        context.request_mut().headers_mut().insert(
+            API_VERSION_HEADER,
+            HeaderValue::from_static(self.api_version),
+        );
+        Ok(())
+    }
+}
+
+/// Adds a glacier tree hash checksum to the HTTP Request
+#[derive(Debug, Default)]
+pub(crate) struct GlacierTreeHashHeaderInterceptor;
+
+impl Interceptor for GlacierTreeHashHeaderInterceptor {
+    fn modify_before_serialization(
+        &self,
+        _context: &mut BeforeSerializationInterceptorContextMut<'_>,
+        cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        // Request the request body to be loaded into memory immediately after serialization
+        // so that it can be checksummed before signing and transmit
+        cfg.put(LoadedRequestBody::Requested);
+        Ok(())
+    }
+
+    fn modify_before_retry_loop(
+        &self,
+        context: &mut BeforeTransmitInterceptorContextMut<'_>,
+        cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        let maybe_loaded_body = cfg.get::<LoadedRequestBody>();
+        if let Some(LoadedRequestBody::Loaded(body)) = maybe_loaded_body {
+            let content_sha256 = add_checksum_treehash(context.request_mut(), body)?;
+
+            // Override the signing payload with this precomputed hash
+            let mut signing_config = cfg
+                .get::<SigV4OperationSigningConfig>()
+                .ok_or("SigV4OperationSigningConfig not found")?
+                .clone();
+            signing_config.signing_options.payload_override =
+                Some(SignableBody::Precomputed(content_sha256));
+            cfg.put(signing_config);
+        } else {
+            return Err(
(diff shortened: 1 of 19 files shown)


aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
```diff
@@ -0,0 +1,454 @@
...
+        // Get the content SHA-256 calculated during serialization (via the `RequestSerializer` Glacier customization)
+        let content_sha256 = context
+            .request()
+            .headers()
+            .get(X_AMZ_CONTENT_SHA256)
+            .ok_or_else(|| {
+                format!("{X_AMZ_CONTENT_SHA256} should have been set during serialization")
+            })?
+            .to_str()?
+            .to_owned();
```
Velfi, line 130: My gut tells me we should be handling the `or_else` after converting the possible OK value to a string to avoid allocating two strings when the header is missing. Not a big deal, though.

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
```diff
@@ -0,0 +1,454 @@
...
+use aws_smithy_runtime_api::client::interceptors::context::phase::{
+    BeforeSerialization, BeforeTransmit,
+};
+use aws_smithy_runtime_api::client::interceptors::{BoxError, Interceptor, InterceptorContext};
+use aws_smithy_runtime_api::config_bag::ConfigBag;
+use bytes::Buf;
+use bytes_utils::SegmentedBuf;
+use http::header::{HeaderName, HeaderValue};
+use http::Request;
+use ring::digest::{Context, Digest, SHA256};
```
Velfi, line 22: Isn't there some AWS-owned crypto thing we can use for this? I remember people mentioning it when people were discussing FIPS compliance a few weeks ago.

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
```diff
@@ -0,0 +1,454 @@
...
+        }
+        impl GlacierAccountId for SomeInput {
+            fn account_id_mut(&mut self) -> &mut Option<String> {
+                &mut self.account_id
+            }
+        }
+
+        let mut cfg = ConfigBag::base();
+        let mut context =
+            InterceptorContext::<()>::new(TypedBox::new(SomeInput { account_id: None }).erase());
```
Velfi, line 264: We create structs for this so much in testing it'd be cool to have a derive macro for `From<T> for TypedBox`. I'm guessing those require us to use `syn` tho.

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
```diff
@@ -0,0 +1,454 @@
...
+                .signing_options
+                .payload_override
+                .as_ref()
+                .unwrap()
+        );
+    }
+}
+
+#[cfg(test)]
+mod treehash_checksum_tests {
```
Velfi, line 359: Is this the sort of thing where generating test cases with prop testing is useful?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/glacier/GlacierDecorator.kt
```diff
@@ -6,21 +6,206 @@
...
+            }
+        }
+    }
+}
+
+/**
+ * Adds two interceptors:
+ * 1. `GlacierAccountIdAutofillInterceptor`: Uses the `GlacierAccountId` trait to correctly autofill the account ID field
+ * 2. `GlacierSetPrecomputedSignableBodyInterceptor`: Reuses the tree hash computation during signing rather than allowing
+ *    the `aws-sigv4` module to recalculate the payload hash.
```
Velfi, line 137: Another chance for the orc to be faster than the old client :D

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
```diff
@@ -0,0 +1,454 @@
...
+) -> Result<(Digest, Vec<Digest>), byte_stream::error::Error> {
+    let mut hashes = vec![];
+    let mut remaining_in_chunk = chunk_size;
+    let mut body = ByteStream::new(body);
+    let mut local = Context::new(&SHA256);
+    let mut full_body = Context::new(&SHA256);
+    let mut segmented = SegmentedBuf::new();
+    while let Some(data) = body.try_next().await? {
+        segmented.push(data);
+        while segmented.has_remaining() {
```
ysaito1001, line 199: Just to check, once `segmented` has NO remaining, what will happen when `segmented.push(data)` is called in the next iteration of `while let Some(data) = ...`, assuming the full `segmented` will continue being used in the next iteration?

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs
```diff
@@ -27,7 +27,7 @@ pub type BoxFuture<T> = Pin<Box<dyn StdFuture<Output = Result<T, BoxError>> + Se
 pub type Future<T> = NowOrLater<Result<T, BoxError>, BoxFuture<T>>;
 
 pub trait RequestSerializer: Send + Sync + fmt::Debug {
-    fn serialize_input(&self, input: Input) -> Result<HttpRequest, BoxError>;
+    fn serialize_input(&self, input: Input) -> Future<HttpRequest>;
```
ysaito1001, line 30: It's now async because we need to compute hashes in a async manner?

ysaito1001 (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/RequestSerializerGenerator.kt
```diff
@@ -103,6 +116,59 @@ class RequestSerializerGenerator(
...
+                )
+            } else {
+                writer.rustTemplate(
+                    """
+                    async fn serialize(input: #{ConcreteInput}) -> #{Result}<#{HttpRequest}, #{BoxError}> {
+                        #{serialize_inner}
+                        #{async_after_serialization}
+                        Ok(request)
+                    }
+                    let input = #{TypedBox}::<#{ConcreteInput}>::assume_from(input).expect("correct type").unwrap();
```
rcoh, line 164: can we allow `#{async_after_serialization}` to be empty instead of having two different paths? Doesn't seem like having serialize be async given that the resulting generated state machine will only contain a ready variant is a problem

rcoh (commented):
IMO we should consider this as a need for async interceptors as opposed to making serialization itself async but I'm open to suggestions.

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
jdisanti: AWS LC? Definitely out of scope for this PR.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/glacier/GlacierDecorator.kt
jdisanti: This is all adapted from the middleware implementation. It already had this.

jdisanti:
Going to hold off on merging this until we have a chance to revisit the interceptor design.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/glacier/GlacierDecorator.kt
```diff
@@ -6,21 +6,206 @@
...
+import software.amazon.smithy.rust.codegen.core.util.dq
+import software.amazon.smithy.rust.codegen.core.util.hasTrait
+import software.amazon.smithy.rust.codegen.core.util.letIf
+import software.amazon.smithy.rustsdk.AwsCargoDependency
+import software.amazon.smithy.rustsdk.InlineAwsDependency
 
+/**
+ * Glacier has three required customizations:
+ *
+ * 1. Add the `x-amz-glacier-version` header to all requests
```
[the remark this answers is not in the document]
jdisanti: I think you're assuming this was part of some migration, but I don't know enough history to say one way or another.

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
jdisanti: I don't think so in this case.

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
jdisanti: Implementing a generic `From<T> for Whatever` seems to always conflict with the `From<T> for T` that's built into the language, so that's not really feasible.

rcoh: you don't need a derive macro, just `erase!(...)` which is shorthand for

```
macro_rules! erase {
 ($input: expr) => { TypedBox::new($input).erase() }
```

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
jdisanti: There's no allocation occurring until the `to_owned` call. It's all references. The error case will allocate for the error message though, but then that's immediately bubbled up.

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
jdisanti: I like that macro idea.

rcoh:
After thinking on this some more, I think my preference is for a sentinel like:

```
enum LoadBodyIntoMemory {
  NotNeeded,
  Requested,
  Loaded(Bytes)
}
```

The orchestrator would look for this sentinel during the pre-signing phase. If it's set to requested, the orchestrator will asynchronously read the body and replace it with `Loaded(Bytes)`.

The glacier plugin would be one interceptor to set `LoadBodyIntoMemory::Requested`, then a second interceptor that runs later the uses `Loaded(...)` to add the header.

We can use this pattern generally if there are any services where we need to always sign the body, e.g.

aws/rust-runtime/aws-inlineable/src/glacier_interceptors.rs
jdisanti: I simplified this by moving the request body reading into the orchestrator.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/RequestSerializerGenerator.kt
jdisanti: Removed this entirely.

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs
```diff
@@ -100,6 +101,23 @@ impl RequestTime {
     }
 }
 
+/// Informs the orchestrator on whether or not the request body needs to be loaded into memory before transmit.
+///
+/// This enum gets placed into the `ConfigBag` to change the orchestrator behavior.
+/// Immediately after serialization (before the `read_after_serialization` interceptor hook),
+/// if it was set to `Requested` in the config bag, it will be replaced back into the config bag as
+/// `Loaded` with the request body contents for use in later interceptors.
```
rcoh, line 109: can you add docs about how this interacts with retries?

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -80,45 +89,53 @@ fn apply_configuration(
...
         ctx.set_request(request);
     }
 
+    // Load the request body into memory if configured to do so
+    if let LoadedRequestBody::Requested = cfg.loaded_request_body() {
+        let mut body = SdkBody::taken();
+        mem::swap(&mut body, ctx.request_mut().body_mut());
+        let loaded_body = halt_on_err!([ctx] => ByteStream::new(body).collect().await).into_bytes();
+        *ctx.request_mut().body_mut() = SdkBody::from(loaded_body.clone());
+        cfg.set_loaded_request_body(LoadedRequestBody::Loaded(loaded_body));
```
rcoh, line 120: could make a comment/todo: we can do slightly better here by keeping an `Box<dyn Buf>` or `SegmentedBuf` instead and avoiding the second copy of the parts into a contiguous buffer.

rcoh (approved):
nice!

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -156,30 +173,30 @@ async fn try_attempt(
     cfg: &mut ConfigBag,
     interceptors: &Interceptors,
 ) {
-    halt_on_err!([ctx] => interceptors.read_before_attempt(ctx, cfg).map_err(Into::into));
-    halt_on_err!([ctx] => orchestrate_endpoint(ctx, cfg).map_err(Into::into));
-    halt_on_err!([ctx] => interceptors.modify_before_signing(ctx, cfg).map_err(Into::into));
-    halt_on_err!([ctx] => interceptors.read_before_signing(ctx, cfg).map_err(Into::into));
+    halt_on_err!([ctx] => interceptors.read_before_attempt(ctx, cfg));
```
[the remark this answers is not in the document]
jdisanti: It's just how the macro was written. They don't serve any purpose now. Previously when we had context _and_ `AttemptCheckpoint`, the presence of the `[]` determined which of those the macro was using. However, now that `AttemptCheckpoint` has been removed, they're just vestigial.

We are intentionally using `=>` instead of `,` though so that rustfmt doesn't bloat each of these lines to 4-5 lines of code.

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
jdisanti: I think if we end up optimizing this, it will be to call a function for each chunk that gets read as talked about in https://github.com/awslabs/smithy-rs/issues/2729.

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs
jdisanti: Added in 7ed3d3fdb54c2d0b690b9e2f1d4ec2ab9dab5bef

PR #2706, opened 2023-05-17
Port customizable operation to orchestrator

ysaito1001:
## Motivation and Context
Port [Customizable Operation](https://github.com/awslabs/smithy-rs/pull/1647) to orchestrator

## Description
This PR implements `CustomizableOperation` in the orchestrator. Just like the counterpart in the middleware, it is created when the `customize` method (in the orchestrator mode) on a fluent builder is called. The `customize` method in the orchestrator could technically be made a synchronous method because there is no need to create an operation, which requires `async`, therefore making the `customize` method in the middleware `async`. However, during the transition from the middleware to the orchestrator, the integration tests ([example](https://github.com/awslabs/smithy-rs/blob/31c152d9af53afb9a5e6edf9df3def57931b9c1e/aws/sdk/integration-tests/s3/tests/signing-it.rs#L36)) need to be able to run in both modes. For this reason, the `customize` method in the orchestrator is temporarily marked as `async`.

Regarding methods defined on the new `CustomizableOperation`, they include `mutate_request` and `map_request` from the counterpart in the middleware. However, it did not port `map_operation` because there is no operation to map on. Most use cases for `map_operation` is put things in a property bag. The new `CustomizableOperation` provides an `interceptor` method to accomplish the same, i.e putting things in a config bag.

Finally, for integration tests to run in both modes, the code gen emits the implementation of the `customize` method differently depending on the active Smithy runtime mode, similar to what the implementation of `send` method does.

## Testing
Added one `sra-test` for mutating a request.

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
```diff
@@ -126,6 +128,145 @@ class CustomizableOperationGenerator(
...
+                /// The fields in the builder that are `Some` override those applied to the service
+                /// configuration level. For instance,
+                ///
+                /// Config A     overridden by    Config B          ==        Config C
+                /// field_1: None,                field_1: Some(v2),          field_1: Some(v2),
+                /// field_2: Some(v1),            field_2: Some(v2),          field_2: Some(v2),
+                /// field_3: Some(v1),            field_3: None,              field_3: Some(v1),
+                pub fn config_override(
+                    mut self,
+                    config_override: impl Into<crate::config::Builder>,
```
jdisanti, line 223:
```suggestion
                    config_override: impl #{Into}<crate::config::Builder>,
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -349,6 +353,15 @@ class FluentClientGenerator(
                     #{send_bounds:W} {
                         self.send_middleware().await
                     }
+
+                    /// Consumes this builder, creating a customizable operation that can be modified before being
+                    /// sent. The operation's inner [http::Request] can be modified as well.
+                    pub async fn customize(self) -> std::result::Result<
```
jdisanti, line 359:
```suggestion
                    pub async fn customize(self) -> #{Result}<
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -402,6 +417,12 @@ class FluentClientGenerator(
                             })?;
                         #{Ok}(#{TypedBox}::<#{OperationOutput}>::assume_from(output).expect("correct output type").unwrap())
                     }
+
+                    ##[doc(hidden)]
+                    // TODO(enableNewSmithyRuntime): Remove `async` once we switch to orchestrator
+                    pub async fn customize_orchestrator(self) -> #{CustomizableOperation} {
+                        #{CustomizableOperation} { fluent_builder: self, config_override: None, interceptors: vec![] }
```
jdisanti, line 424: Will the initial config override be lost if I do something like the following?
```rust
client.some_operation()
    .config_override(some_config)
    .customize()
    .await
    .add_interceptor(some_interceptor)
    .send()
```

rust-runtime/aws-smithy-runtime/src/client/interceptor.rs
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+use aws_smithy_http::body::SdkBody;
+use aws_smithy_runtime_api::client::interceptors::context::phase::BeforeTransmit;
+use aws_smithy_runtime_api::client::interceptors::{BoxError, Interceptor, InterceptorContext};
+use aws_smithy_runtime_api::config_bag::ConfigBag;
+use std::fmt;
+use std::marker::PhantomData;
+
+pub struct RequestMapInterceptor<F, E> {
```
jdisanti, line 13: `MapRequestInterceptor` may be a nicer name since it has the verb in front.

rust-runtime/aws-smithy-runtime/src/client/interceptor.rs
```diff
@@ -0,0 +1,80 @@
...
+        _cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        let request = context.request_mut();
+        (self.f)(request)?;
+
+        Ok(())
+    }
+}
+
+pub struct RequestMutationInterceptor<F> {
```
jdisanti, line 50: Similarly, `MutateRequestInterceptor` is shorter and has the verb in front.

jdisanti (commented):
This looks good! Just some minor things.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
ysaito1001: Good question. The given snippet no longer complies after this PR, since there is no `config_override` on a fluent builder anymore. `config_override` is only accessible after you call `.customize`. Besides regular parameters for an operation input, I wanted to shove everything that's related to configuring a single operation invocation under `CustomizableOperation`. That being said, I'm curious to know what you think.

rust-runtime/aws-smithy-runtime/src/client/interceptor.rs
ysaito1001: Renamed in be5a126.

rust-runtime/aws-smithy-runtime/src/client/interceptor.rs
ysaito1001: Renamed in be5a126.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
jdisanti: Sounds good.

jdisanti (approved):

PR #2707, opened 2023-05-17
Add support for TimeStreamWrite and TimeStreamQuery

rcoh:
TODO:
- [x] docs
- [x] integration test (canary even?)
- [x] customize README for timestream
## Motivation and Context
- #613 
- https://github.com/awslabs/aws-sdk-rust/issues/114

## Description
This adds support for TSW and TSQ by adding endpoint discovery as a customization. This is made much simpler by the fact that endpoint discovery for these services **has no parameters** which means that there is no complexity from caching the returned endpoint.

Customers call `.enable_endpoint_discovery()` on the client to create a version of the client with endpoint discovery enabled. This returns a new client and a Reloader from which customers must spawn the reload task if they want endpoint discovery to rerun.

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs
+++ b/aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs
@@ -59,14 +59,6 @@ where
             .map(|(creds, _expiry)| creds)
     }
 
-    /// Attempts to load the cached value if it has been set
-    ///
-    /// # Panics
-    /// This function panics if it is called from an asynchronous context
-    pub fn try_blocking_get(&self) -> Option<T> {
-        self.value.blocking_read().get().map(|(v, _exp)| v.clone())
-    }
-
     /// Attempts to refresh the cached value with the given future.
     /// If multiple threads attempt to refresh at the same time, one of them will win,
     /// and the others will await that thread's result rather than multiple refreshes occurring.
--- a/aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs
+++ b/aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs
@@ -59,6 +59,14 @@ where
             .map(|(creds, _expiry)| creds)
     }
 
+    /// Attempts to load the cached value if it has been set
+    ///
+    /// # Panics
+    /// This function panics if it is called from an asynchronous context
+    pub fn try_blocking_get(&self) -> Option<T> {
+        self.value.blocking_read().get().map(|(v, _exp)| v.clone())
+    }
+
     /// Attempts to refresh the cached value with the given future.
     /// If multiple threads attempt to refresh at the same time, one of them will win,
     /// and the others will await that thread's result rather than multiple refreshes occurring.
--- a/aws/rust-runtime/aws-inlineable/src/endpoint_discovery.rs
+++ b/aws/rust-runtime/aws-inlineable/src/endpoint_discovery.rs
@@ -154,12 +154,10 @@ impl EndpointCache {
 
 #[cfg(test)]
 mod test {
-    use crate::endpoint_discovery::{create_cache, EndpointCache};
-    use aws_credential_types::time_source::TimeSource;
+    use crate::endpoint_discovery::create_cache;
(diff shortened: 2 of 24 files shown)


aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs
```diff
@@ -59,6 +59,14 @@ where
             .map(|(creds, _expiry)| creds)
     }
 
+    /// Attempts to load the cached value if it has been set
+    ///
+    /// # Panics
+    /// This function panics if it is called from an asynchronous context
+    pub fn try_blocking_get(&self) -> Option<T> {
```
jdisanti, line 66: Doesn't look like this is used?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/timestream/TimestreamDecorator.kt
```diff
@@ -0,0 +1,75 @@
...
+                            },
+                            sleep,
+                        )
+                        .await?;
+                        new_conf.endpoint_resolver = std::sync::Arc::new(resolver);
+                        Ok((Self::from_conf(new_conf), reloader))
+                    }
+                }
+
+                """,
```
jdisanti, line 67: `preludeScope` will need to be imported for this to work:
```suggestion
                """,
                *preludeScope,
```

jdisanti (commented):
This looks good, but seems like it is missing tests.

aws/rust-runtime/aws-credential-types/src/cache/expiring_cache.rs
rcoh: yeah deleted, an artifact from a previous impl

aws/rust-runtime/aws-inlineable/src/endpoint_discovery.rs
```diff
@@ -0,0 +1,287 @@
...
+    Ok((cache, reloader))
+}
+
+impl EndpointCache {
+    fn resolve_endpoint(&self) -> aws_smithy_http::endpoint::Result {
+        self.endpoint
+            .lock()
+            .unwrap()
+            .as_ref()
+            .map(|e| e.endpoint.clone())
```
Velfi, line 144: Do we need a `map`? Can't we just call `clone` directly?
```suggestion
            .clone()
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/timestream/TimestreamDecorator.kt
```diff
@@ -0,0 +1,106 @@
...
+import software.amazon.smithy.rustsdk.AwsCargoDependency
+import software.amazon.smithy.rustsdk.DocSection
+import software.amazon.smithy.rustsdk.InlineAwsDependency
+
+/**
+ * This decorator does two things:
+ * 1. Adds the `endpoint_discovery` inlineable
+ * 2. Adds a `enable_endpoint_discovery` method on client that returns a wrapped client with endpoint discovery enabled
+ */
+class TimestreamDecorator : ClientCodegenDecorator {
```
Velfi, line 30: Did you actually use the crate docs decorator anywhere? I would have assumed there'd be something in here about how timestream support is experimental.

aws/rust-runtime/aws-inlineable/src/endpoint_discovery.rs
rcoh: this is plucking out the endpoint in the map (not just cloning)

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/timestream/TimestreamDecorator.kt
rcoh: I generated docs that add `enable_endpoint_discovery`—I figure our standard boilerplate is probably good for starters

PR #2708, opened 2023-05-18
Make the SDK ad hoc tests pass against the orchestrator

jdisanti:
## Motivation and Context
This PR refactors the client protocol test generator machinery to use a client instead of calling `make_operation` directly, and then fixes the ad hoc tests for the orchestrator.

The ad hoc tests revealed that overriding the signing region/service via endpoint config was lost when porting SigV4 signing to the orchestrator, so this PR updates the SigV4 `HttpRequestSigner` implementation to restore this functionality. It is doing this in the signer directly rather than via an interceptor since it should only run this logic when SigV4 is the selected auth scheme.

Other notable changes:
- Adds `--no-fail-fast` arg to `cargoTest` targets so that all Rust tests run in CI rather than stopping on the first failure
- Changes `EndpointResolver::resolve_and_apply_endpoint` to just `resolve_endpoint` so that the orchestrator can place the endpoint config into the request state, which is required for the signer to make use of it
- Adds a `set_region` method to SDK service configs
- Deletes the API Gateway model and integration test from the SDK smoke test since it is covered by the ad hoc tests
- Adds a comment explaining where the API Gateway model comes from in the ad hoc tests
- Adds a `smithy.runtime.mode` Gradle property to `aws:sdk` and `aws:sdk-adhoc-test` to trivially switch between middleware and orchestrator when testing/generating locally

----


aws/sdk-adhoc-test/build.gradle.kts
```diff
@@ -46,6 +48,7 @@ val allCodegenTests = listOf(
             ,
             "codegen": {
                 "includeFluentClient": false,
+                "enableNewSmithyRuntime": "${getSmithyRuntimeMode()}"
```
rcoh, line 51: are you intentionally passing the string middleware here? I'd expect this to be true/false

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -96,6 +98,27 @@ class AwsFluentClientDecorator : ClientCodegenDecorator {
...
+    ): ProtocolTestGenerator = DefaultProtocolTestGenerator(
+        codegenContext,
+        baseGenerator.protocolSupport,
+        baseGenerator.operationShape,
+        renderClientCreation = { params ->
+            rustTemplate(
+                """
+                let config = ${params.configBuilderName}
+                    .http_connector(${params.connectorName})
+                    .region(crate::config::Region::new("us-east-1"))
```
rcoh, line 114: another way to do this is to use vendor params and customize the tests themselves...I wonder if that's better actually.

https://github.com/awslabs/smithy-rs/blob/6beec943ab767a791a414bbaef1acb4b363f7042/aws/sdk/aws-models/glacier-tests.smithy#L23-L29

I think the existing tests almost all already have a region set?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolTestGenerator.kt
```diff
@@ -166,33 +200,56 @@ class ProtocolTestGenerator(
             writable {
                 val customizations = codegenContext.rootDecorator.endpointCustomizations(codegenContext)
                 params.getObjectMember("builtInParams").orNull()?.members?.forEach { (name, value) ->
-                    customizations.firstNotNullOf { it.setBuiltInOnServiceConfig(name.value, value, "builder") }(this)
+                    customizations.firstNotNullOf {
+                        it.setBuiltInOnServiceConfig(name.value, value, "config_builder")
+                    }(this)
                 }
```
rcoh, line 206: this code should be handling the region already

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolTestGenerator.kt
```diff
@@ -166,33 +200,56 @@ class ProtocolTestGenerator(
...
+                            bindings.render(this, "input", codegenContext.smithyRuntimeMode, generateValidation = false)
+                        }
+                    }
+                }
                 rustTemplate(
                     """
                     let mut http_request = http_request;
                     let ep = #{SmithyHttp}::endpoint::Endpoint::mutable(${withScheme.dq()}).expect("valid endpoint");
-                    ep.set_endpoint(http_request.uri_mut(), parts.acquire().get()).expect("valid endpoint");
+                    ep.set_endpoint(http_request.uri_mut(), endpoint_prefix.as_ref()).expect("valid endpoint");
```
rcoh, line 252: do we still need to do it this way now that we're capturing the entire request?

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs
```diff
@@ -63,12 +63,7 @@ impl EndpointResolverParams {
 }
 
 pub trait EndpointResolver: Send + Sync + fmt::Debug {
-    fn resolve_and_apply_endpoint(
-        &self,
-        params: &EndpointResolverParams,
-        endpoint_prefix: Option<&EndpointPrefix>,
-        request: &mut HttpRequest,
-    ) -> Result<(), BoxError>;
+    fn resolve_endpoint(&self, params: &EndpointResolverParams) -> Result<Endpoint, BoxError>;
```
rcoh, line 66: I like this API better, 👍🏻

tools/ci-scripts/check-aws-sdk-adhoc-tests
```diff
@@ -4,6 +4,17 @@
...
-./gradlew aws:sdk-adhoc-test:test
+
+# TODO(enableNewSmithyRuntime): Remove the middleware test run when cleaning up middleware
+echo -e "## ${C_YELLOW}Running SDK adhoc tests against the middleware implementation...${C_RESET}"
+./gradlew aws:sdk-adhoc-test:clean
+./gradlew aws:sdk-adhoc-test:check -Psmithy.runtime.mode=middleware
+
+echo -e "## ${C_YELLOW}Running SDK adhoc tests against the orchestrator implementation...${C_RESET}"
+./gradlew aws:sdk-adhoc-test:clean
+./gradlew aws:sdk-adhoc-test:check -Psmithy.runtime.mode=orchestrator
```
rcoh, line 20: does this work cleanly in GitHub actions re: the coloring?

rcoh (approved):
Only major comment is using vendor params instead of hardcoding the region to us-east-1

rcoh (approved):
Only major comment is using vendor params instead of hardcoding the region to us-east-1

aws/sdk-adhoc-test/build.gradle.kts
jdisanti: It's a quad state: middleware, both defaulting to middleware, both defaulting to orchestrator, orchestrator

aws/rust-runtime/aws-runtime/src/auth.rs
```diff
@@ -178,12 +277,112 @@ pub mod sigv4 {
...
+                        operation_config.service = service;
+                    }
+                    Ok(Cow::Owned(operation_config))
+                }
+            }
+        }
+
+        fn extract_endpoint_auth_scheme_config(
+            value: &SmithyEndpoint,
+        ) -> Result<EndpointAuthSchemeConfig, EndpointAuthSchemeConfigError> {
```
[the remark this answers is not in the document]
jdisanti: Yeah, not a bad idea. I split the endpoint document parsing so that the orchestrator resolves the selected auth scheme config and passes it directly to the signer.

aws/rust-runtime/aws-runtime/src/auth.rs
```diff
@@ -270,8 +469,8 @@ pub mod sigv4 {
                 "test",
             );
             let operation_config = SigV4OperationSigningConfig {
-                region: SigningRegion::from_static("test"),
-                service: SigningService::from_static("test"),
+                region: Some(SigningRegion::from_static("test")),
+                service: Some(SigningService::from_static("test")),
```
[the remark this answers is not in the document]
jdisanti: I needed them to be optional so that the endpoint auth scheme config could set their value when they weren't otherwise set (where previously, it would panic because they didn't have values).

tools/ci-scripts/check-aws-sdk-adhoc-tests
jdisanti: It has worked for the other scripts that do it.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
jdisanti: Oh, I didn't know this was available. Easy fix.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolTestGenerator.kt
jdisanti: Ahhh, I see why it wasn't working now. The `ApiGatewayAccept` test doesn't have endpoint parameters, so this short-circuits and doesn't set a region for that test specifically. I guess I just need to add a default. I don't think there's an easy way to customize the trait to only add endpoint parameters while preserving the rest.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolTestGenerator.kt
jdisanti: It's needed for cases like this where the test supplies a custom host with a path in it:
https://github.com/awslabs/smithy/blob/be68f3bbdfe5bf50a104b387094d40c8069f16b1/smithy-aws-protocol-tests/model/restJson1/endpoint-paths.smithy#L19

aws/rust-runtime/aws-runtime/src/auth.rs
```diff
@@ -25,6 +31,53 @@ pub mod sigv4 {
...
+        MissingSigningRegion,
+        MissingSigningService,
+        WrongIdentityType(Identity),
+        BadTypeInEndpointAuthSchemeConfig(&'static str),
+    }
+
+    impl fmt::Display for SigV4SigningError {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            use SigV4SigningError::*;
+            let mut w = |s| f.write_str(s);
```
Velfi, line 51: avoiding line breaks?

aws/rust-runtime/aws-runtime/src/auth.rs
```diff
@@ -25,6 +31,53 @@ pub mod sigv4 {
...
+
+    impl StdError for SigV4SigningError {
+        fn source(&self) -> Option<&(dyn StdError + 'static)> {
+            match self {
+                Self::MissingOperationSigningConfig => None,
+                Self::MissingSigningRegion => None,
+                Self::MissingSigningService => None,
+                Self::WrongIdentityType(_) => None,
+                Self::BadTypeInEndpointAuthSchemeConfig(_) => None,
+            }
```
Velfi, line 77: imo, just do:
```suggestion
            // SigV4 errors never have a source.
            None
```

aws/rust-runtime/aws-runtime/src/auth.rs
```diff
@@ -178,12 +231,75 @@ pub mod sigv4 {
...
+        fn extract_endpoint_auth_scheme_config(
+            endpoint_config: AuthSchemeEndpointConfig<'_>,
+        ) -> Result<EndpointAuthSchemeConfig, SigV4SigningError> {
+            let (mut signing_region_override, mut signing_service_override) = (None, None);
+            if let Some(config) = endpoint_config.config().and_then(Document::as_object) {
+                use SigV4SigningError::BadTypeInEndpointAuthSchemeConfig as UnexpectedType;
+                signing_region_override = match config.get("signingRegion") {
+                    Some(Document::String(s)) => Some(SigningRegion::from(Region::new(s.clone()))),
+                    None => None,
+                    _ => return Err(UnexpectedType("signingRegion")),
```
Velfi, line 291: It's weird to see a `_` case for this. Is `config.get` not returning an `Option`?

Velfi (approved):
Thanks for patching up the broken classifiers in codegen

aws/rust-runtime/aws-runtime/src/auth.rs
jdisanti: Exactly.

aws/rust-runtime/aws-runtime/src/auth.rs
jdisanti: The errors RFC specifically calls out not doing this 😄

aws/rust-runtime/aws-runtime/src/auth.rs
jdisanti: It's because I'm matching on the inner part of `Some()`. So this catches cases such as `Some(Document::Number(_))`.

PR #2712, opened 2023-05-18
Cherry-pick commits from 'main' onto 'smithy-rs-release-0.55.x'

david-perez:
This tree is the result of:

```console
git clone https://github.com/awslabs/smithy-rs.git
cd smithy-rs
git checkout -b davidpz/smithy-rs-release-0.55.x-cherry-pick origin/smithy-rs-release-0.55.x
git cherry-pick 2ebbfad49..origin/main
```

----


david-perez:
Ran `git cherry-pick a514793fa..upstream/main` to tack on recent changes.

david-perez:
The problem with why the semver checks CI action is failing is that it's trying to generate a super old version of smithy-rs, and then a super new version, and the `publisher` tool in the build image is only compatible with the new version: indeed, https://github.com/awslabs/smithy-rs/pull/2663 contained a breaking change that removed the `--location` parameter from the `publisher` tool.
It's the PR bot that is generating these two versions: it's generating the first commit and then the last commit, and diffing them, but it uses the `publisher` tool that is built into the build image by the last commit.
This will also break SDK release too during `sdk-sync`. We need to update the `publisher` tool to be backwards compatible (still use `--location`) and include that update in the patch release.
It's easy to forget that all changes to `tools/` need to be backwards compatible _across releases_.

david-perez:
Now that https://github.com/awslabs/smithy-rs/pull/2716 has landed, I'm closing this in favor of https://github.com/awslabs/smithy-rs/pull/2718.

PR #2713, opened 2023-05-18
fix type-erasure bug for errors

Velfi:
## Description

We were unintentionally double-boxing type-erased errors. This fix correctly de-references the errors, solving the problem.

## Testing

I wrote a test.

----


jdisanti (approved):

ysaito1001 (approved):
Thanks for the fix!

PR #2715, opened 2023-05-19
Test the orchestrator against the SDK in CI

jdisanti:
## Motivation and Context
This PR adds the SDK smoke test with the orchestrator implementation to CI, and excludes services that don't yet pass tests.

----


aws/sdk/README.md
```diff
@@ -12,7 +12,7 @@ Generate an SDK:
 `./gradlew :aws:sdk:assemble`
 
 Generate, compile, and test an SDK:
```
Velfi, line 14:
```suggestion
Generate, compile, and `cargo check` an SDK:
```

Velfi (approved):

aws/sdk/README.md
jdisanti: No, it actually tests it too.

Velfi: ic

PR #2726, opened 2023-05-24
Render `CustomizableOperation` per crate in orchestrator

ysaito1001:
## Motivation and Context
Render `CustomizableOperation` once in the `customizable` module instead of rendering it in every operation's `builders` module.

## Description
This PR is a follow-up on #2706. The previous PR ported `CustomizableOperation` in the middleware to the orchestrator but it had a flaw in that the type was rendered per every operation. This was clearly a codegen regression because the code for `CustomizableOperation` is repeated many times and is rendered even if no customization is requested for an operation.

This PR attempts to fix that problem. Just like `CustomizableOperation` in the middleware, we render the new `CustomizableOperation` once in the `customize` module per crate. To accomplish that, the new `CustomizableOperation` uses a closure, called `customizable_send`, to encapsulate a call to a fluent builder's `send` method and invokes it when its own `send` method is called.

~~As a side note, this PR will not take care of the following. We have [a separate PR](https://github.com/awslabs/smithy-rs/pull/2723) to fix them and once we've merged it to this PR, they should be addressed automatically (there will be merge conflicts, though):~~
- ~~Removing `send_orchestrator_with_plugin` (which is why a type parameter `R` is present in the code changes, but it'll soon go away)~~
- ~~Incorrect signature of a closure in orchestrator's `CustomizableOperaton::map_request`~~

## Testing
No new tests added as it's purely refactoring. Confirmed `sra-test` passed (except for the one that's known to fail).

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCustomizableOperationDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCustomizableOperationDecorator.kt
@@ -5,6 +5,7 @@
 
 package software.amazon.smithy.rustsdk
 
+import software.amazon.smithy.rust.codegen.client.smithy.SmithyRuntimeMode
 import software.amazon.smithy.rust.codegen.client.smithy.generators.client.CustomizableOperationCustomization
 import software.amazon.smithy.rust.codegen.client.smithy.generators.client.CustomizableOperationSection
 import software.amazon.smithy.rust.codegen.core.rustlang.CargoDependency
@@ -37,12 +38,8 @@ class CustomizableOperationTestHelpers(runtimeConfig: RuntimeConfig) :
     override fun section(section: CustomizableOperationSection): Writable =
         writable {
             if (section is CustomizableOperationSection.CustomizableOperationImpl) {
-                if (section.operationShape == null) {
+                if (section.runtimeMode == SmithyRuntimeMode.Middleware) {
                     // TODO(enableNewSmithyRuntime): Delete this branch when middleware is no longer used
-                    // This branch customizes CustomizableOperation in the middleware. section.operationShape being
-                    // null means that this customization is rendered in a place where we don't need to figure out
-                    // the module for an operation (which is the case for CustomizableOperation in the middleware
-                    // that is rendered in the customize module).
                     rustTemplate(
                         """
                         ##[doc(hidden)]
@@ -62,7 +59,7 @@ class CustomizableOperationTestHelpers(runtimeConfig: RuntimeConfig) :
                         *codegenScope,
                     )
                 } else {
-                    // The else branch is for rendering customization for the orchestrator.
+                    assert(section.runtimeMode == SmithyRuntimeMode.Orchestrator)
                     rustTemplate(
                         """
                         ##[doc(hidden)]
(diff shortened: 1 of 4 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -493,10 +515,18 @@ class FluentClientGenerator(
...
                         // TODO(enableNewSmithyRuntime): Remove `async` and `Result` once we switch to orchestrator
-                        pub async fn customize(self) -> #{Result}<
-                            #{CustomizableOperation},
-                            #{SdkError}<#{OperationError}>
-                        > {
+                        pub async fn customize(
+                            self,
+                        ) -> #{Result}<
+                            #{CustomizableOperation}<
+                                #{Box}<dyn #{CustomizableSend}<
```
jdisanti, line 522: Is there a way we can hide the `Box<dyn CustomizableSend` piece from the public API to maximize flexibility in refactoring this in the future?

jdisanti (commented):
This looks good. Just a couple comments.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCustomizableOperationDecorator.kt
```diff
@@ -37,12 +38,8 @@ class CustomizableOperationTestHelpers(runtimeConfig: RuntimeConfig) :
     override fun section(section: CustomizableOperationSection): Writable =
         writable {
             if (section is CustomizableOperationSection.CustomizableOperationImpl) {
-                if (section.operationShape == null) {
+                if (section.runtimeMode == SmithyRuntimeMode.Middleware) {
```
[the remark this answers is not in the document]
ysaito1001: Only re-purposed two states out of the quad states, hence I totally see your point. They are being used in the wrong context, hence look confusing. Since it's true or false, a boolean works.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
ysaito1001: Maybe add yet another alias `BoxedCustomizableSend` ?

jdisanti: An alias won't necessarily let us change the type later. If we're using a `Box<dyn>` as the generic, then it seems like maybe it doesn't need to be a generic in the first place. Perhaps `CustomizableOperation` only takes the output and error types as generic args.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCustomizableOperationDecorator.kt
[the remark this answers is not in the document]
ysaito1001: Fixed in c9b1a85.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
ysaito1001: Ah, I see what you're saying; Just expose the output type & the error type and hide the function part of the type parameter, `S`,  from `CustomizableOperation`. Let me try making that change.

Issue #2087, opened 2022-12-09
WASM support for the SDK


jdisanti:
Related SDK issue: https://github.com/awslabs/aws-sdk-rust/issues/59

Currently the SDK doesn't work so well when compiled for WASM. These are the issues we know about:
- [x] The currently released version of `ring` (0.16.20) pulls in `wasm-bindgen`, which is undesirable for cases where the WASM is targeting servers instead of browsers
- [x] `aws-config` pulls in all of `ring` just for a SHA-1 hash, which was deemed OK at the time since the SDK uses `rustls` by default, and `rustls` pulls in `ring`. However, WASM applications are unlikely to use `rustls`.
- [ ] For signing, the time source needs to be configurable so that it can come from the WASM runtime to avoid a `time not implemented on this platform` panic. (#2262)

There may be additional issues that appear once the above are solved.

DavidSouther:
For reference - https://github.com/awsdocs/aws-doc-sdk-examples/issues/4637

aajtodd:
Closing as we have some WASM support and tests implemented at this point. Anything missing should be a new issue and/or favor the existing SDK [issue](https://github.com/awslabs/aws-sdk-rust/issues/59)

PR #2728, opened 2023-05-25
Integrate TimeSource into the orchestrator and middleware (breaking change warning is spurious)

rcoh:
## Motivation and Context
- #2262 
- #2087 
- #2707 

This adds `TimeSource` in SDK and service configs so we can effectively control time when executing requests with the SDK at the client level.

_note:_ the breaking change is in a trait implementation of a struct we're going to delete, not a real breaking change

## Description
- Add `SharedTimeSource` and use it in SdkConfig / `aws-config` / <service>::Config
- Wire up the signer and other uses of `SystemTime::now()` that I could find
- track down broken tests

## Testing
- [x] various unit tests that all still pass

## Checklist

----


aws/rust-runtime/aws-credential-types/src/time_source.rs
```diff
@@ -3,31 +3,22 @@
...
-/// #     fn with_timesource(ts: TimeSource) -> Self {
-/// #         Client { }
-/// #     }
-/// # }
-/// use aws_credential_types::time_source::TimeSource;
-/// let time = TimeSource::default();
-/// let client = Client::with_timesource(time);
-/// ```
 #[derive(Debug, Clone)]
+// TODO(breakingChangeWindow): Delete this struct
```
jdisanti, line 21: Should we also add `#[deprecated]`? You can also probably just remove it now...

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs
```diff
@@ -288,12 +265,12 @@ impl ConfigBagAccessors for ConfigBag {
...
-    fn request_time(&self) -> Option<RequestTime> {
-        self.get::<RequestTime>().cloned()
+    fn request_time(&self) -> Option<SharedTimeSource> {
+        self.get::<SharedTimeSource>().cloned()
     }
 
-    fn set_request_time(&mut self, request_time: RequestTime) {
-        self.put::<RequestTime>(request_time);
+    fn set_request_time(&mut self, request_time: impl TimeSource + 'static) {
+        self.put::<SharedTimeSource>(SharedTimeSource::new(request_time));
```
jdisanti, line 273: Does this end up being a `Box<Arc<...>>` under the hood due to the `ConfigBag` implementation?

jdisanti (approved):
Looks good.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCustomizableOperationDecorator.kt
```diff
@@ -70,7 +72,7 @@ class CustomizableOperationTestHelpers(runtimeConfig: RuntimeConfig) :
                         pub fn request_time_for_tests(mut self, request_time: ::std::time::SystemTime) -> Self {
                             use #{ConfigBagAccessors};
                             let interceptor = #{TestParamsSetterInterceptor}::new(move |_: &mut #{BeforeTransmitInterceptorContextMut}<'_>, cfg: &mut #{ConfigBag}| {
-                                cfg.set_request_time(#{RequestTime}::new(request_time));
+                                cfg.set_request_time(#{StaticTimeSource}::new(request_time));
```
ysaito1001, line 75: Is `StaticTimeSource` available in `codegenScope`?

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs
rcoh: yeah (but so does every thing we put in the ConfigBag, unfortunately)

there is a little trickery we could do to avoid that double-allocation with the new load/store traits because they can offer an enum that exposes the internals of the object so the config bag can use the smart pointer directly. not sure if it's worth it

PR #2730, opened 2023-05-25
Make SigningInstructions public

cholcombe973:
## Motivation and Context

I'm building an s3 server in rust and I need signature verification to work from the server side when receiving requests. To do that the server needs to generate a signature and then compare it to the one that the client has already sent along. I believe that the sign module in aws-sigv4 http_request contains the functions required to do that. 

## Description

Exported the `SigningInstructions` struct in the sign module of http_request for aws-sigv4.

## Testing


jdisanti:
Why can't you just use the `aws_sigv4::http_request::sign` method? Example here: https://docs.rs/aws-sigv4/0.55.3/aws_sigv4/http_request/index.html

The signing instructions give you access to just the signature string if that's all you need: https://docs.rs/aws-sigv4/0.55.3/aws_sigv4/struct.SigningOutput.html#method.signature

cholcombe973:
Well I wanted to return the signing instructions from the function and I can't complete the function signature because SigningInstructions is private.

jdisanti:
I see. I think the correct fix would be to add `SigningInstructions` to the export here: https://github.com/awslabs/smithy-rs/blob/b023426d1cfd05e1fd9eef2f92a21cad58b93b86/aws/rust-runtime/aws-sigv4/src/http_request/mod.rs#L59

We probably just missed that during initial implementation since I guess it works if you only need to call a function on that returned value rather than name it.

cholcombe973:
Thanks! I'll do that in the morning

cholcombe973:
@jdisanti I pushed that update.

jdisanti:
Looks good. Could you also update the PR description/title and add an entry to `CHANGELOG.next.toml` for `[[aws-sdk-rust]]`?

CHANGELOG.next.toml
```diff
@@ -11,6 +11,18 @@
...
+author = "rcoh"
+references = ["smithy-rs#2612"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
+
+[[smithy-rs]]
+message = "Implement `Ord` and `PartialOrd` for `DateTime`."
+author = "henriiik"
+references = ["smithy-rs#2653", "smithy-rs#2656"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
+
```
jdisanti, line 25: These were deleted in main:
```suggestion
```

CHANGELOG.next.toml
```diff
@@ -34,3 +46,9 @@ message = "`SsoCredentialsProvider`, `AssumeRoleProvider`, and `WebIdentityToken
 references = ["smithy-rs#2720"]
 meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "ysaito1001"
+
+[[aws-sdk-rust]]
+message = "The SigningInstructions in the aws-sigv4 module are now public. This allows them to be named in a function signature."
+references = ["smithy-rs#2730"]
+author = "cholcombe973"
+meta = { "breaking" = false, "tada" = false, "bug" = false }
```
jdisanti, line 54:
```suggestion
message = "The `SigningInstructions` in the `aws-sigv4` module are now public. This allows them to be named in a function signature."
references = ["smithy-rs#2730"]
author = "cholcombe973"
meta = { "breaking" = false, "tada" = false, "bug" = true }
```

jdisanti (approved):
Looks good. I'll apply a couple minor changes and get this merged. Thank you!

CHANGELOG.next.toml
cholcombe973: Yeah those got pulled in by accident with a rebase. whoops

cholcombe973:
@jdisanti did you want me to update the docs for the SigningInstructions as well? It looks like that's what most of these test breakages are about.

jdisanti:
That would be great!

jdisanti (approved):

PR #2736, opened 2023-05-31
Fix various small issues with the orchestrator

Velfi:
## Motivation and Context

This change fixes many of the smaller issues I ran into during the implementation of standard retries for the orchestrator. Merges from `main` were getting difficult with my [other PR](https://github.com/awslabs/smithy-rs/pull/2725) so I'm breaking things up.

## Description

- when orchestrator attempt timeout occurs, error is now set in context
- update test connection to allow defining connection events with optional latency simulation update orchestrator attempt loop to track iteration count
- set request attempts from the attempt loop
- add comment explaining "rewind" step of making request attempts add `doesnt_matter` method to `TypeErasedBox`,  useful when testing update tests to use the new `TypeErasedBox::doesnt_matter` method
- add more doc comments
- add `set_subsec_nanos` method to `DateTime`.
  - I added this to make it easier to string-format a datetime that didn't include the nanos.
- fix Invocation ID interceptor not inserting the expected header update input type for `OperationError::other` to be more user-friendly
- add `test-util` feature to `aws-smithy-runtime-api`
- add `test-util` feature to `aws-runtime`
- fix presigining inlineable to pull in tower dep during codegen

## Testing

tests have been updated where necessary
----


rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
```diff
@@ -285,23 +283,34 @@ where
...
+        self.request_checkpoint = try_clone(self.request());
+    }
+
     // Returns false if rewinding isn't possible
-    pub fn rewind(&mut self, _cfg: &mut ConfigBag) -> bool {
-        // If before transmit was never touched, then we don't need to rewind
-        if !self.tainted {
-            return true;
-        }
+    pub fn rewind(&mut self, _cfg: &mut ConfigBag) -> RewindResult {
```
jdisanti, line 293: This should probably also be `#[doc(hidden)]`

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -142,12 +142,19 @@ async fn try_op(ctx: &mut InterceptorContext, cfg: &mut ConfigBag, interceptors:
...
         // Break from the loop if we can't rewind the request's state. This will always succeed the
         // first time, but will fail on subsequent iterations if the request body wasn't retryable.
-        if !ctx.rewind(cfg) {
-            break;
+        match ctx.rewind(cfg) {
+            RewindResult::Impossible => {
+                break;
+            }
+            r @ RewindResult::Occurred => debug!("{r}"),
+            r @ RewindResult::Unnecessary => debug!("{r}"),
```
jdisanti, line 156: Are these debug lines useful?

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
```diff
@@ -327,6 +336,32 @@ where
     }
 }
 
+/// The result of attempting to rewind a request.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum RewindResult {
```
jdisanti, line 341: Seems like this should be `non_exhaustive`, or private somehow. Not blocking this PR, but maybe we should move `InterceptorContext` into `aws-smithy-runtime` and have separate customer facing interceptor contexts in the `-api` crate so that all the orchestrator details are well hidden.

jdisanti (approved):

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
Velfi: I can delete them or make them `trace`s if you like. I think they're useful only insofar as they expose an important part of the request/response lifecycle managed by the orchestrator. However, since the user would never be calling `rewind`, it's not very helpful to them.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
Velfi: I'll make it `#[doc(hidden)]` for now. I'm guessing we'll do a reorganization pass before GA.

PR #2737, opened 2023-05-31
Update retry classifiers to act on InterceptorContext instead of OrchestratorError

Velfi:
## Description

- update retry classifiers to act directly on the `InterceptorContext`
- add codegen method for inserting retry classifiers into the config bag
- update `InterceptorContext` to return options instead of panicking

## Testing

Tests were updated as needed.

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/retries/classifier.rs
+++ b/aws/rust-runtime/aws-runtime/src/retries/classifier.rs
@@ -34,29 +34,14 @@ pub const THROTTLING_ERRORS: &[&str] = &[
 pub const TRANSIENT_ERRORS: &[&str] = &["RequestTimeout", "RequestTimeoutException"];
 
 /// A retry classifier for determining if the response sent by an AWS service requires a retry.
-#[derive(Debug)]
-pub struct AwsErrorCodeClassifier<E>
-where
-    E: StdError + ProvideErrorMetadata + Send + Sync + 'static,
-{
+#[derive(Debug, Default)]
+pub struct AwsErrorCodeClassifier<E> {
     _inner: PhantomData<E>,
 }
 
-impl<E> AwsErrorCodeClassifier<E>
-where
-    E: StdError + ProvideErrorMetadata + Send + Sync + 'static,
-{
+impl<E> AwsErrorCodeClassifier<E> {
     /// Create a new AwsErrorCodeClassifier
     pub fn new() -> Self {
-        Self::default()
-    }
-}
-
-impl<E> Default for AwsErrorCodeClassifier<E>
-where
-    E: StdError + ProvideErrorMetadata + Send + Sync + 'static,
-{
-    fn default() -> Self {
         Self {
             _inner: PhantomData,
         }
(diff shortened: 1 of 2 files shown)


aws/rust-runtime/aws-runtime/src/retries/classifier.rs
```diff
@@ -32,15 +35,45 @@ pub const TRANSIENT_ERRORS: &[&str] = &["RequestTimeout", "RequestTimeoutExcepti
...
-impl AwsErrorCodeClassifier {
-    /// Classify an error code to check if represents a retryable error. The codes of retryable
-    /// errors are defined [here](THROTTLING_ERRORS) and [here](TRANSIENT_ERRORS).
-    pub fn classify_error<E: ProvideErrorMetadata, R>(
-        &self,
-        error: &SdkError<E, R>,
-    ) -> Option<RetryReason> {
+pub struct AwsErrorCodeClassifier<E>
+where
+    E: StdError + ProvideErrorMetadata + Send + Sync + 'static,
```
jdisanti, line 40: What do you think about just passing in the error metadata rather than making it generic over `E`?

aws/rust-runtime/aws-runtime/src/retries/classifier.rs
```diff
@@ -32,15 +35,45 @@ pub const TRANSIENT_ERRORS: &[&str] = &["RequestTimeout", "RequestTimeoutExcepti
...
+{
+    _inner: PhantomData<E>,
+}
+
+impl<E> AwsErrorCodeClassifier<E>
+where
+    E: StdError + ProvideErrorMetadata + Send + Sync + 'static,
```
ysaito1001, line 47: Carrying over [this comment](https://github.com/awslabs/smithy-rs/pull/2725#discussion_r1207358260) from the previous PR, my understanding is that in Rust, you delay adding bounds until necessary. Yes, we need these bounds eventually when calling `classify_retry` so I see the claim as to why not also add them here if we need them eventually? The counterargument is that if someone just wants to hold on to `AwsErrorCodeClassifier<E>` as a struct field but has no intention to call `classify_retry` on it, these bounds will have to bubble up to the definition of that struct.

aws/rust-runtime/aws-runtime/src/retries/classifier.rs
Velfi: The retry strategy is responsible for calling the classifiers and it does so by calling a single wrapper struct that works similar to how the `Interceptors` wrapper thing worked. The classifiers are all stored within this struct, which itself is stored in the config bag. So, either we can use an API where everything has the same args, or we can codegen the "access the right type for a classifier and then pass it to the classifier" stuff.

I wanted to avoid generating more code since this would affect all operations. Hence, I prefer the current approach.

aws/rust-runtime/aws-runtime/src/retries/classifier.rs
Velfi: We need the `E` because the `classify_retry` method calls `downcast_ref::<E>()`. Putting the generic on the struct was how I avoided putting it on the `classify_retry` trait method. I'm open to suggestions for another way to accomplish this.

Velfi: Thanks for explaining it to me Yuki. I made the changes.

ysaito1001 (approved):

PR #2741, opened 2023-06-01
No really, add the standard retry strategy to the orchestrator

Velfi:
## Description

This PR adds support for the Standard retry strategy. The standard strategy will be inserted into a service's config bag if a retry config was set. Otherwise, unless another retry strategy was already set, a `NeverRetryStrategy` will be set. This seemed like a reasonable default, but I'm open to suggestions.

## Testing

tests are included

---


ysaito1001 (approved):
Looks good! Do we need to port [existing retry-related unit tests in retry.rs](https://github.com/awslabs/smithy-rs/blob/main/rust-runtime/aws-smithy-client/src/retry.rs#L407-L614) to the orchestrator?

PR #2744, opened 2023-06-02
Implement `SdkConfig` using `FrozenLayer`

ysaito1001:
## Motivation and Context
The `SdkConfig` is now implemented using `FrozenLayer` internally, instead of having individual fields for those configured.

## Description
This PR updates the `SdkConfig` to use `FrozenLayer` internally. What that means is that when a config is asked to query a configured value, it'll retrieve the value from the frozen layer.

We're planning to implement the `SdkConfig` _builder_ using a mutable `Layer` and delete individual fields. However, `Layer` is currently not cloneable, whereas the builder requires its members to implement the `Clone` trait. Until `Layer` is cloneable, the builder will contain individual fields just like today.

This PR is currently a draft until a series of RuntimePlugin and ConfigBag refactorings is done.

## Testing
- [x] Add unit tests in `SdkConfig` to verify the implementation using `ConfigBag` works as expected
- [x] Passed tests in CI

----


rust-runtime/aws-smithy-async/src/test_util.rs
```diff
@@ -24,7 +24,7 @@ pub struct ManualTimeSource {
 
 impl TimeSource for ManualTimeSource {
     fn now(&self) -> SystemTime {
-        self.start_time + self.log.lock().unwrap().iter().sum()
+        self.start_time + self.log.lock().unwrap().iter().sum::<Duration>()
```
ysaito1001, line 27: Without the change, it caused
```
27 |         self.start_time + self.log.lock().unwrap().iter().sum()
   |                         -                                 ^^^ cannot infer type of the type parameter `S` declared on the associated function `sum`
```

aws/rust-runtime/aws-types/src/endpoint.rs
```diff
@@ -0,0 +1,31 @@
...
+
+//! Newtypes for endpoint-related parameters
+//!
+//! Parameters require newtypes so they have distinct types when stored in layers in config bag.
+
+use aws_smithy_types::config_bag::{Storable, StoreReplace};
+
+/// Newtype for `use_fips`
+#[derive(Clone, Debug)]
+pub struct UseFips(pub bool);
```
jdisanti, line 14: Would it be nicer to implement `Deref` for these rather than making the inner boolean public so that it could be accessed using `*use_fips`?

PR #2747, opened 2023-06-06
Write an S3 throughput benchmark

jdisanti:
## Motivation and Context
This PR adds a S3 throughput benchmark for upload, download, multipart upload across multiple threads, and multipart download across multiple threads.

----


aws/sdk/s3-benchmark/benchmark/src/multipart_get.rs
```diff
@@ -0,0 +1,92 @@
...
+                (i + 1) * args.part_size_bytes as i64 - 1,
+                args.part_size_bytes as i64,
+            )
+        }
+    });
+
+    let semaphore = Arc::new(Semaphore::new(args.concurrency));
+    let mut tasks = Vec::new();
+    for range in ranges {
+        tasks.push(tokio::spawn(download_part(
```
jdisanti, line 42: There's a slight problem here in that the tasks will run in a non-deterministic order, which means later parts of the file can start downloading first. Since the tasks are iterated in order later, and since the body is read upon awaiting the tasks, this could potentially lead to issues if it takes a long time to get to one of the later parts that started sooner than everything else.

Not sure if this is really a problem in practice, or just hypothetical. Also not really sure if there's an easy way to fix it.

rcoh: I'd fix it by:
1. Pushing the work items onto a channel (in order)
2. Have the tasks 1) Acquire semaphore, 2) pull item from channel

That keeps the work items weighted to the front of the file which means the early tasks finish first. 

This is likely to be a little faster in practice, but more importantly, it will result in lower (and bounded!) memory usage.

jdisanti: Ah, good idea. Thanks for the suggestion!

Velfi (approved):

PR #2754, opened 2023-06-07
Update runtime plugin trait

rcoh:
## Motivation and Context
Update the RuntimePlugin trait based on discussion:
1. Methods are infallible
2. Split out `Config` and `Interceptors`
3. `ConfigBag` now has an explicit field `interceptor_state`
4. Refactor `ConfigBagAccessors` so that we can build the core around the trait and keep the trait together with a `where Self` trick 

## Description
- Update the `RuntimePlugin` trait
- Deal with resulting implications
## Testing
- [x] CI

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationCustomization.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationCustomization.kt
@@ -101,10 +101,15 @@ sealed class OperationSection(name: String) : Section(name) {
      */
     data class AdditionalRuntimePluginConfig(
         override val customizations: List<OperationCustomization>,
-        val configBagName: String,
+        val newLayerName: String,
+        val operationShape: OperationShape,
+    ) : OperationSection("AdditionalConfig")
+
+    data class AdditionalInterceptors(
+        override val customizations: List<OperationCustomization>,
         val interceptorRegistrarName: String,
         val operationShape: OperationShape,
-    ) : OperationSection("AdditionalConfig") {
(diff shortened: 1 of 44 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationCustomization.kt
```diff
@@ -101,10 +101,15 @@ sealed class OperationSection(name: String) : Section(name) {
      */
     data class AdditionalRuntimePluginConfig(
         override val customizations: List<OperationCustomization>,
-        val configBagName: String,
+        val newLayerName: String,
+        val operationShape: OperationShape,
+    ) : OperationSection("AdditionalConfig")
```
jdisanti, line 106:
```suggestion
    ) : OperationSection("AdditionalRuntimePluginConfig")
```

rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs
```diff
@@ -98,196 +99,252 @@ pub enum LoadedRequestBody {
...
+    fn config_bag(&self) -> &ConfigBag {
+        self
+    }
+}
 
-impl ConfigBagAccessors for ConfigBag {
-    fn auth_option_resolver_params(&self) -> &AuthOptionResolverParams {
-        self.get::<AuthOptionResolverParams>()
+pub trait ConfigBagAccessors {
+    fn auth_option_resolver_params(&self) -> &AuthOptionResolverParams
```
jdisanti, line 129: Seems like we may want to macro this

jdisanti (approved):
This is awesome! ❤️

PR #2757, opened 2023-06-08
Add the ability to disable interceptors via the config bag

rcoh:
This replaces the existing custom interceptor disable logic with shared logic to allow generally disabling any public interceptors from Runtime plugins. The previous behavior was very brittle because it relied heavily on runtime plugin execution order.

## Motivation and Context
- simplify presigning behavior
- generalize logic to disable interceptors

## Description
Create `disable_interceptor` struct, which, when inserted into the configuration bag can disable an interceptor via the `Interceptors` execution interface.

## Testing
- ` (cd aws/sdk/build/aws-sdk/sdk/s3 && cargo test --test presigning)`

## Checklist

----


ysaito1001 (approved):
Really like the change 👍

PR #2762, opened 2023-06-12
Make service config just contain a `FrozenLayer` in the orchestrator mode

ysaito1001:
## Motivation and Context
Service config structs now only contain a `aws_smithy_types::config_bag::FrozenLayer` in the orchestrator mode (no changes for the middleware mode).

## Description
This PR reduces the individual fields of service configs to contain just `FrozenLayer`. This makes service configs work more seamlessly with runtime plugins in the orchestrator mode.

Note that service config _builder_ s still contain individual fields. We're planning to make them just contain `aws_smithy_types::config_bag::Layer`. To do that, though, we need to make `Layer` cloneable and that will be handled in a separate PR. For builders, the only change you will in the PR is that their `build` method will put fields into a `Layer`, freeze it, and pass it to service configs.

This PR is marked as a breaking change because it's based on [another PR](https://github.com/awslabs/smithy-rs/pull/2728) that's also breaking change. 

## Testing
- [x] Passed tests in CI

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointConfigCustomization.kt
```diff
@@ -38,21 +39,38 @@ internal class EndpointConfigCustomization(
...
+                            """
+                            /// Returns the endpoint resolver.
+                            pub fn endpoint_resolver(&self) -> $sharedEndpointResolver {
+                                self.endpoint_resolver.clone()
+                            }
+                            """,
+                            *codegenScope,
+                        )
+                    }
+                }
```
Velfi, line 73: Looking at these customizations, they all seem to have the same outer block and different innards. Therefore you could reduce duplication with something like this:

```suggestion
                is ServiceConfig.ConfigImpl -> {
                    docs("Returns the endpoint resolver.")
                    rustBlockTemplate(
                        "pub fn endpoint_resolver(&self) -> $sharedEndpointResolver",
                        *codegenScope,
                    ) {
                        if (runtimeMode.defaultToOrchestrator) {
                            """self.inner.load::<$sharedEndpointResolver>().expect("endpoint resolver should be set").clone()"""
                        } else {
                            "self.endpoint_resolver.clone()"
                        }
                    }
                }
```

It's a bit shorter, although it may make things a little less clear. What do you think?

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGeneratorTest.kt
```diff
@@ -76,45 +85,100 @@ internal class ServiceConfigGeneratorTest {
...
+
                     ServiceConfig.BuilderStruct -> writable { rust("config_field: Option<u64>") }
                     ServiceConfig.BuilderImpl -> emptySection
                     ServiceConfig.BuilderBuild -> writable {
-                        rust("config_field: self.config_field.unwrap_or_default(),")
+                        if (runtimeMode.defaultToOrchestrator) {
+                            rustTemplate(
+                                """
+                                layer.store_or_unset(self.config_field.map(#{T}));
+                                """,
```
Velfi, line 137: If it's just a single line and doesn't contain any double quotes, you don't need triple quotes:
```suggestion
                                "layer.store_or_unset(self.config_field.map(#{T}));",
```
Saves a few lines.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGeneratorTest.kt
```diff
@@ -76,45 +85,100 @@ internal class ServiceConfigGeneratorTest {
...
+                unitTest(
+                    "set_config_fields",
+                    """
+                    let mut builder = Config::builder();
+                    builder.config_field = Some(99);
+                    let config = builder.build();
+                    assert_eq!(config.config_field, 99);
+                    """,
+                )
+            }
```
Velfi, line 181: Are these branches both the same or am I missing something?

rust-runtime/aws-smithy-async/src/rt/sleep.rs
```diff
@@ -68,6 +69,10 @@ impl AsyncSleep for SharedAsyncSleep {
     }
 }
 
+impl Storable for SharedAsyncSleep {
+    type Storer = StoreReplace<SharedAsyncSleep>;
+}
```
Velfi, line 74: Thanks for updating all of these!

rcoh (commented):
Overall I really like the direction here.

Some things we should think about:

1. In a RuntimePlugin world, I don't think we need to set defaults in the service config do we? We might want a separate plugin that sets defaults and keep the fields on service config optional...

2. Some more thought about how we handle identity resolvers / interceptors is required I think
3. In a RuntimePlugin world, I guess service config **only** needs to handle fields we aren't taking care of elsewhere, maybe eventually, it just holds onto the SdkConfig in it's runtime plugin state?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointConfigCustomization.kt
ysaito1001: I originally took the exact approach you suggested. Then I realized the clarity suffers a bit, and more importantly, the ease of deleting the middleware code when we switch to the orchestrator. Throughout the PR, we'll see duplicated snippets for the orchestrator and the middleware, but they are "optimized" for future removal.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGeneratorTest.kt
ysaito1001: Admittedly, the downside of code duplication mentioned above is this: can be hard to spot where the difference is. The difference in this case is line 168 vs. line 178. One needs `config_field()` to access the field and the other can access it directly.

ysaito1001:
> Overall I really like the direction here.
> 
> Some things we should think about:
> 
>     1. In a RuntimePlugin world, I don't think we need to set defaults in the service config do we? We might want a separate plugin that sets defaults and keep the fields on service config optional...
> 
>     2. Some more thought about how we handle identity resolvers / interceptors is required I think
> 
>     3. In a RuntimePlugin world, I guess service config **only** needs to handle fields we aren't taking care of elsewhere, maybe eventually, it just holds onto the SdkConfig in it's runtime plugin state?

Shared with Russell off-line that this PR is meant to be as dumb and mechanical as possible, simply translating what we currently have into the new way. We'll address provided feedback in a series of upcoming PRs (especially 1 and 3 above).

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpConnectorConfigDecorator.kt
```diff
@@ -147,7 +161,11 @@ private class HttpConnectorConfigCustomization(
             }
 
             is ServiceConfig.BuilderBuild -> writable {
-                rust("http_connector: self.http_connector,")
+                if (runtimeMode.defaultToOrchestrator) {
+                    rust("layer.store_or_unset(self.http_connector);")
```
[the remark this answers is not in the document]
ysaito1001: Agree that we should treat defaults in the service config in a different way in the orchestrator. This code change was originally to preserve the old behavior; Fixed in 52a0d6e.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/InterceptorConfigCustomization.kt
```diff
@@ -46,7 +39,7 @@ class InterceptorConfigCustomization(codegenContext: CodegenContext) : ConfigCus
                     ##[doc(hidden)]
                     /// Returns interceptors currently registered by the user.
                     pub fn interceptors(&self) -> impl Iterator<Item = &#{SharedInterceptor}> + '_ {
-                        self.interceptors.iter()
+                        self.inner.load::<#{SharedInterceptor}>()
```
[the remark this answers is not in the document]
ysaito1001: As discussed, b988ad9 now leaves `interceptors` and `identity_resolvers` as fields in a service config.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/ResiliencyConfigCustomization.kt
```diff
@@ -216,21 +244,30 @@ class ResiliencyConfigCustomization(codegenContext: CodegenContext) : ConfigCust
...
-                    """,
-                    *codegenScope,
-                )
+                ServiceConfig.BuilderBuild -> {
+                    if (runtimeMode.defaultToOrchestrator) {
+                        rustTemplate(
+                            """
+                            layer.store_or_unset(self.retry_config);
+                            layer.store_or_unset(self.sleep_impl.clone());
+                            layer.store_or_unset(self.timeout_config);
```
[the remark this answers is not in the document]
ysaito1001: Fixed in 52a0d6e.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/timestream/TimestreamDecorator.kt
```diff
@@ -92,9 +95,15 @@ class TimestreamDecorator : ClientCodegenDecorator {
...
                 }
-
                 """,
                 "endpoint_discovery" to endpointDiscovery.toType(),
+                "time_source" to writable {
+                    if (runtimeMode.defaultToOrchestrator) {
+                        rust("time_source()")
+                    } else {
+                        rust("time_source")
+                    }
```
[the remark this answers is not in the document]
ysaito1001: Good catch. The accessor was already defined in both runtime modes, so did not need to bifurcate the codegen in the first place. Fixed in 7333e70.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpAuthDecorator.kt
```diff
@@ -324,23 +325,41 @@ private class HttpAuthConfigCustomization(
             }
 
             is ServiceConfig.BuilderBuild -> {
-                rust("identity_resolvers: self.identity_resolvers,")
```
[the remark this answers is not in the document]
ysaito1001: Yes we do. It's just that what we need to apply `StoreAppend` to is `(AuthSchemeId, Arc<dyn IdentityResolver>)` so we need to define a type out of it. Will tackle it in a separate PR.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointConfigCustomization.kt
Velfi: That's smart! I didn't even consider that.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/IdempotencyTokenProviderCustomization.kt
```diff
@@ -15,25 +16,42 @@ import software.amazon.smithy.rust.codegen.core.smithy.customize.NamedCustomizat
...
-                    """,
-                    RuntimeType.IdempotencyToken,
-                )
+                if (runtimeMode.defaultToOrchestrator) {
+                    rustTemplate(
+                        """
+                        /// Returns a copy of the idempotency token provider.
+                        /// If a random token provider was configured,
+                        /// a newly-randomized token provider will be returned.
+                        pub fn make_token(&self) -> #{IdempotencyTokenProvider} {
```
jdisanti, line 36: I guess it was named this before, but it's kind of weird that this is called `make_token`.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/IdempotencyTokenProviderCustomizationTest.kt
```diff
@@ -5,12 +5,51 @@
...
+        // to use the `Storable` trait (there is no smithy runtime mode when handwriting Rust source files since that's
+        // a codegen concept). When that inlinable is brought in during testing, the orchestrator mode is ok because a
+        // service config already depends on the `aws_smithy_types` crate, causing it to appear in `Cargo.toml` of the
+        // synthesized test crate. However, a service config in the middleware mode does not depend on the
+        // `aws_smithy_types` crate, so `Cargo.toml` of the synthesized test crate will not include it, causing the
+        // inlinable fail to compile. To work around it, the code below will force the `aws_smithy_types` to appear in
+        // `Cargo.toml`.
+        project.withModule(ClientRustModule.Config) {
+            if (smithyRuntimeMode.defaultToMiddleware) {
+                rustTemplate(
```
jdisanti, line 41: For future reference, you can just call `addDependency`:
```kotlin
addDependency(CargoDependency.smithyTypes(runtimeConfig))
```

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/IdempotencyTokenProviderCustomizationTest.kt
```diff
@@ -5,12 +5,51 @@
...
+        val project = TestWorkspace.testProject(symbolProvider)
+        // The following is a workaround for the middleware mode, and without this, the synthesized test crate would
+        // fail to compile. `IdempotencyTokenProvider` is a smithy inlineable and depends on `aws_smithy_types` crate
+        // to use the `Storable` trait (there is no smithy runtime mode when handwriting Rust source files since that's
+        // a codegen concept). When that inlinable is brought in during testing, the orchestrator mode is ok because a
+        // service config already depends on the `aws_smithy_types` crate, causing it to appear in `Cargo.toml` of the
+        // synthesized test crate. However, a service config in the middleware mode does not depend on the
+        // `aws_smithy_types` crate, so `Cargo.toml` of the synthesized test crate will not include it, causing the
+        // inlinable fail to compile. To work around it, the code below will force the `aws_smithy_types` to appear in
+        // `Cargo.toml`.
```
jdisanti, line 38: Inline modules have a list of dependencies, and that list should be corrected rather than adding a workaround.

jdisanti (commented):
Would it be simpler to use a `ConfigBag` in both middleware and orchestrator mode, and just change middleware code to call accessor methods instead?

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/IdempotencyTokenProviderCustomizationTest.kt
ysaito1001: Thanks for pointing me to the right direction! Fixed in 0e77721

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/IdempotencyTokenProviderCustomizationTest.kt
ysaito1001: Fixed in 0e77721.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/IdempotencyTokenProviderCustomization.kt
ysaito1001: Yeah, it's good to be renamed. Made it `token_provider` in 89185fa.

ysaito1001:
> Would it be simpler to use a `ConfigBag` in both middleware and orchestrator mode, and just change middleware code to call accessor methods instead?

It's true that using `ConfigBag` in the middleware mode as well makes code changes simpler (since there's no need to bifurcate between two runtime modes), we should leave the middleware mode untouched in case there is a potential bug in `ConfigBag` we're not aware of.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/IdempotencyTokenProviderCustomization.kt
jdisanti: Actually, we should make it something like `idempotency_token_provider` since `token_provider` will be used for token based auth later in the SDK.

jdisanti:
We should add a changelog entry for the `make_token` rename

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/IdempotencyTokenProviderCustomization.kt
ysaito1001: Will take care of it in its own PR since it needs be accompanied by a changelog entry and we want that entry to refer to a small, dedicated PR, but not this large PR 😅

ysaito1001:
> We should add a changelog entry for the `make_token` rename

Will do that as part of the next PR. Thanks for the suggestion.

jdisanti (approved):

PR #2768, opened 2023-06-13
Fix protocol tests against the orchestrator

jdisanti:
This PR fixes the protocol tests in orchestrator mode, and adds `--all-targets` to the orchestrator CI checks.

----


Velfi (approved):
Thanks for fixing this John

PR #2771, opened 2023-06-14
Make RuntimePlugin & RuntimePlugins Send/Sync

rcoh:
## Motivation and Context
We will eventually want to store additional runtime plugins on fluent builders, clients, etc. To make this work, the RuntimePlugin trait needs to assert Send/Sync

## Description
Small tweaks + test on the RuntimePlugin trait

## Testing
CI

----


ysaito1001 (approved):

jdisanti (approved):

jdisanti:
Breaking change? I don't think it is since nothing in these runtime crates are actually used by anything yet.

PR #2772, opened 2023-06-14
Parameterize `Plugin` by service rather than protocol

hlbarber:
## Motivation and Context

Closes https://github.com/awslabs/smithy-rs/issues/1839

Currently, `Plugin` is parameterized by protocol and operation. To improve symmetry, extensibility and uniformity we switch this to be parameterized by service instead. The protocol can still be recovered via a associated type on `ServiceShape`.

## Description

- Add `ServiceShape` trait, encoding the properties of a Smithy service.
- Change `Plugin<Protocol, Operation, S>` to `Plugin<Service, Operation, S>`.

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
@@ -137,22 +137,22 @@ class ServerServiceGenerator(
                     HandlerType: #{SmithyHttpServer}::operation::Handler<crate::operation_shape::$structName, HandlerExtractors>,
 
                     ModelPlugin: #{SmithyHttpServer}::plugin::Plugin<
-                        #{Protocol},
+                        $serviceName,
                         crate::operation_shape::$structName,
                         #{SmithyHttpServer}::operation::IntoService<crate::operation_shape::$structName, HandlerType>
                     >,
                     #{SmithyHttpServer}::operation::UpgradePlugin::<UpgradeExtractors>: #{SmithyHttpServer}::plugin::Plugin<
-                        #{Protocol},
+                        $serviceName,
                         crate::operation_shape::$structName,
                         ModelPlugin::Service
                     >,
                     HttpPlugin: #{SmithyHttpServer}::plugin::Plugin<
-                        #{Protocol},
+                        $serviceName,
                         crate::operation_shape::$structName,
                         <
                             #{SmithyHttpServer}::operation::UpgradePlugin::<UpgradeExtractors>
                             as #{SmithyHttpServer}::plugin::Plugin<
-                                #{Protocol},
+                                $serviceName,
                                 crate::operation_shape::$structName,
                                 ModelPlugin::Service
                             >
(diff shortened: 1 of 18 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerServiceGenerator.kt
```diff
@@ -590,6 +590,51 @@ class ServerServiceGenerator(
...
+            }
+
+            """,
+            *codegenScope,
+        )
+
+        for ((_, value) in operationStructNames) {
+            rustTemplate(
+                """
+                impl #{SmithyHttpServer}::service::ServiceShapeMember<crate::operation_shape::$value> for $serviceName {
```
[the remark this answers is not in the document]
david-perez: Hmm I understand now how it's used to get the enum value corresponding to the operation being hit inside a plugin:

```rust
impl<Ser, Op, S, Inner, F> Plugin<Ser, Op, S> for FilterByOperation<Inner, F>
where
    Ser: ServiceShapeMember<Op>,
...
        let either_plugin = if (self.predicate)(<Ser as ServiceShapeMember<Op>>::ENUM_VALUE) {
```

Can you add some Rust docs in `rust-runtime/aws-smithy-http-server/src/service.rs`?

I'm uneasy about the trait name, it feels like the word "operation" should appear in it.

rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
```diff
@@ -68,3 +65,61 @@ where
...
+/// Filters the application of an inner [`Plugin`] using a predicate over the
+/// [`ServiceShape::Operations`](crate::service::ServiceShape::Operations).
+///
+/// See [`filter_by_operation`] for more details.
+pub struct FilterByOperation<Inner, F> {
+    inner: Inner,
+    predicate: F,
+}
+
+impl<Ser, Op, S, Inner, F> Plugin<Ser, Op, S> for FilterByOperation<Inner, F>
```
david-perez, line 78: Hmm not a compelling use case, since it does the exact same thing as `FilterByOperationId`. I think we should only have one of them.

What useful things can we do with the operations enum?

david-perez: Right! 🤦 We should call that out in the docs.

Is there a use case for keeping `FilterByOperationId` then?

What other useful things does this operations enum enable? I can only come up with conditional application of a plugin based on the operation being invoked.

hlbarber: > Is there a use case for keeping `FilterByOperationId` then?

Someone might want to filter on the namespace or perform some action on `&'static str`. This can be subsumed by generating a `fn id(&self) -> ShapeId` on the generated enum though.

design/src/server/middleware.md
```diff
@@ -268,14 +268,17 @@ Suppose we want to apply a different `Layer` to every operation. In this case, p
...
+use aws_smithy_http_server::shape_id::ShapeId;
 use std::task::{Context, Poll};
 use tower::Service;
 
 /// A [`Service`] that adds a print log.
 pub struct PrintService<S> {
     inner: S,
-    name: &'static str,
+    op_id: ShapeId,
+    ser_id: ShapeId
```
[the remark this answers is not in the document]
hlbarber: https://github.com/awslabs/smithy-rs/pull/2772/commits/0485432836a6db6c585b50be1c7aceedd3c8cb09

rust-runtime/aws-smithy-http-server/src/plugin/filter.rs
david-perez: Nobody is going to filter on namespace and there's not really a need to filter specifically on `&'static str`, an enum variant serves the same filtering purpose.

What I'm trying to get at is that "filtering by operation" is, in essence, all this enables. Which is fine and definitely needed, but then there isn't really a need to additionally have `FilterByOperationId`.

PR #2774, opened 2023-06-14
Categorize orchestrator TODO comments into `Launch` and `Cleanup`

jdisanti:
This PR categorizes orchestrator TODO comments into `Launch` and `Cleanup`:

- `enableNewSmithyRuntimeLaunch`: Comment needs to be addressed before orchestrator launch
- `enableNewSmithyRuntimeCleanup`: Comment needs to be addressed after launch when removing middleware

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
@@ -102,7 +102,6 @@ class EndpointsDecorator : ClientCodegenDecorator {
     override val name: String = "Endpoints"
     override val order: Byte = 0
 
-    // TODO(enableNewSmithyRuntime): Remove `operationCustomizations` and `InjectEndpointInMakeOperation`
     override fun operationCustomizations(
         codegenContext: ClientCodegenContext,
         operation: OperationShape,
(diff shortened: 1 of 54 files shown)


rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
```diff
@@ -294,7 +294,6 @@ impl<I, O, E: Debug> InterceptorContext<I, O, E> {
         }
 
         // Otherwise, rewind to the saved request checkpoint
-        // TODO(enableNewSmithyRuntime): Also rewind the ConfigBag
```
ysaito1001, line 297: Has this already been taken care of?

ysaito1001 (approved):
Thanks for making them more explicit!

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
```diff
@@ -102,7 +102,6 @@ class EndpointsDecorator : ClientCodegenDecorator {
     override val name: String = "Endpoints"
     override val order: Byte = 0
 
-    // TODO(enableNewSmithyRuntime): Remove `operationCustomizations` and `InjectEndpointInMakeOperation`
```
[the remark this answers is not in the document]
jdisanti: Things have changed since that comment was added. Previously, I thought we would remove `operationCustomizations`, but then I ended up adding orchestrator-specific operation customizations. So now, we'll just remove the middleware-specific customizations from it.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
jdisanti: From talking to Zelda, it sounds like it doesn't need to be done.

Velfi (approved):
Thanks for keeping all this stuff groomed and relevant

PR #2782, opened 2023-06-15
Update OperationInputTestDecorator to find operation shape by name 

mcmasn-amzn:
## Motivation and Context
Addresses https://github.com/awslabs/smithy-rs/issues/2767

This fixes an issue for users who write endpoint tests that rely on an operation and service shape from different namespaces.

## Description
Instead of assuming operation namespace matches service namespace, 

## Testing
`./gradlew :aws:sdk-codegen:test`

----


rcoh:
LGTM! did a quick audit to make sure this doesn't cause a performance regression—generation times for the SDK seem to be about the same.

rcoh:
If you don't mind adding a quick entry with the bugfix to `CHANGELOG.next.toml` under the `[smithy-rs]` tag we'll get this merged

rcoh (commented):
LGTM—pending changelog.

mcmasn-amzn:
Thanks @rcoh. Just pushed a commit to add the changelog entry

rcoh (approved):

PR #2776, opened 2023-06-14
Make `Layer` `Clone`able

ysaito1001:
## Checklist

- [ ] I have updated `CHANGELOG.next.toml` if I made changes to the smithy-rs codegen or runtime crates
- [ ] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #2784, opened 2023-06-16
Add `CloneableLayer` that can be `Clone`d

ysaito1001:
## Motivation and Context
Alternative to #2776, making `Layer` cloneable so it can be stored in a type that implements `Clone`.

## Description
In an attempt to reduce struct fields of a service config builder so it only contains `Layer` (plus some exceptions such as `interceptors`), we need to make `Layer` clone since the service config builder `#[derive(Clone)]`.

There are two options. One, adopted by this PR, is to have a separate `CloneableLayer` that primarily caters for the need for service config builders. The other option is #2776, where we require that items in `Layer` be `Clone`. Since the latter is rather a sledge hammer, changing the requirement for all of the items to be stored, we will go for the former, which is a more targeted approach and a two-way door if we want to switch to the latter in the future.

## Testing
Added unit tests in `type_erase.rs` and `config_bag.rs`.

----


rust-runtime/aws-smithy-types/src/config_bag.rs
```diff
@@ -79,6 +79,115 @@ impl<T: Default> Default for Value<T> {
...
+    fn clone(&self) -> Self {
+        Self(
+            self.try_clone()
+                .expect("only cloneable types can be inserted"),
+        )
+    }
+}
+
+// We need to "override" the mutable methods to encode the information that an item being stored
+// implements `Clone`. For the immutable methods, they can just be deligated via the `Deref` trait.
```
Velfi, line 109:
```suggestion
// implements `Clone`. For the immutable methods, they can just be delegated via the `Deref` trait.
```

rust-runtime/aws-smithy-types/src/type_erasure.rs
```diff
@@ -101,9 +102,11 @@ impl<T: fmt::Debug + Send + Sync + 'static> DerefMut for TypedBox<T> {
 pub struct TypeErasedBox {
     field: Box<dyn Any + Send + Sync>,
     #[allow(clippy::type_complexity)]
-    debug: Box<
+    debug: Arc<
```
jdisanti, line 105: Is it possible to to create a struct with a PhantomData to do the casting instead of using an Arc? Something like:
```rust
#[derive(Debug, Clone)]
struct DebugAndClone<T> {
    _: PhantomData<T>,
}
impl<T> DebugAndClone<T> {
    fn new() -> Self { Self { _: Default::default() } }

    fn debug(&self, value: &Box<dyn Any + Send + Sync>, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        fmt::Debug::fmt(value.downcast_ref::<T>().expect("type-checked"), f)
    }

    // ...
}
```
If it is, then this should compile down to zero overhead.

ysaito1001: I've considered ways to make it happen since I like the idea, but it seems that an obstacle here is a generic parameter `T`, e.g. if `DebugAndClone` is to be stored in `TypeErasedBox` the generic parameter is unfortunately propagated up to `TypeErasedBox`:
```rust
struct TypeErasedBox<T> {
    field: Box<dyn Any + Send + Sync>,
    debug_and_clone: DebugAndClone<T>,
}
```
This, in turn, makes `T` also appear in `Layer` because it contains `TypeErasedBox<T>`.

I assume we'd want to avoid all of this?

ysaito1001: https://github.com/awslabs/smithy-rs/issues/2794 created in case we come up with a different approach.

jdisanti (approved):

PR #2785, opened 2023-06-17
Fix `@httpChecksum` in the orchestrator implementation

jdisanti:
This PR fixes codegen for the Smithy `@httpChecksum` trait when generating in orchestrator mode, and also fixes an issue in the unit tests where the wrong body was being tested to be retryable. The request body should be retryable rather than the response body.

----


aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
```diff
@@ -0,0 +1,303 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#![allow(dead_code)]
```
Velfi, line 6: Is this still needed? I'm guessing it's for the `pub(crate)` stuff but I wanted to be sure.

Velfi (approved):
Thanks for picking this up from me John.

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
jdisanti: It's mostly to keep CI on the `rust-runtime` crates happy.

Velfi (approved):

PR #2792, opened 2023-06-20
Add get_foo() -> &Option<Foo> accessors for builders and fluent builders.

### LINKED ISSUE #2791: Add Accessors to Builders
Currently, there is no way to get the information in a Builder while creating a request. It would be nice to have some way to get that information. This would be helpful in testing, to be able to verify the contents of a Builder without mocking the request. It may also be useful for implementing a multi stage request, where one builder is passed through several functions that make changes to the request based on prior information.

For an example of a test that would be improved by this pattern, [see this commit](https://github.com/DavidSouther/aws-doc-sdk-examples/commit/017e8a2e407eba6a4c259bca0334f1356ab822d9).

DavidSouther:
This allows testing, as well as making decisions while creating builders. The references are not mutable, so these props remain read only keeping the Builder invariants.

## Motivation and Context

Some operations are built up using various logic (for instance, choosing an S3 bucket based on a logged in user). It is a lot of overhead to test at the mock level, and the Debug formatting is typically not stable.

Closes #2791 

## Description

1. Add `get_foo(&self) -> &Option<Foo>` to Builders
2. Add `as_input(&self) -> &OperationInputBuilder` to Fluent Builders
3. Add `get_foo(&self) -> &Option<Foo>` to FluentBuilder which delegates to Builders

## Testing

Included unit tests, as well as [proof of concept tests](https://github.com/DavidSouther/aws-doc-sdk-examples/commit/017e8a2e407eba6a4c259bca0334f1356ab822d9)

## Checklist

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
@@ -48,6 +48,7 @@ import software.amazon.smithy.rust.codegen.core.smithy.RustSymbolProvider
 import software.amazon.smithy.rust.codegen.core.smithy.customize.writeCustomizations
 import software.amazon.smithy.rust.codegen.core.smithy.expectRustMetadata
 import software.amazon.smithy.rust.codegen.core.smithy.generators.setterName
+import software.amazon.smithy.rust.codegen.core.smithy.generators.getterName
 import software.amazon.smithy.rust.codegen.core.smithy.rustType
 import software.amazon.smithy.rust.codegen.core.util.inputShape
 import software.amazon.smithy.rust.codegen.core.util.orNull
@@ -409,6 +410,14 @@ class FluentClientGenerator(
                 }
             }
 
+            rust("/// Access the inner ${operationSymbol.name} builder as a reference.\n")
+            withBlockTemplate(
+                "pub fn inner(&self) -> &#{Inner} {", "}",
+                "Inner" to symbolProvider.symbolForBuilder(input),
+            ) {
+                write("&self.inner")
+            }
+
             if (smithyRuntimeMode.generateMiddleware) {
                 val middlewareScope = arrayOf(
                     *preludeScope,
@@ -630,6 +639,9 @@ class FluentClientGenerator(
                 val setterName = member.setterName()
                 val optionalInputType = outerType.asOptional()
                 with(core) { renderInputHelper(member, setterName, optionalInputType) }
+
+                val getterName = member.getterName()
+                with(core) { renderGetterHelper(member, getterName, optionalInputType) }
             }
         }
     }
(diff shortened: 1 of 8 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -409,6 +410,14 @@ class FluentClientGenerator(
                 }
             }
 
+            rust("/// Access the inner ${operationSymbol.name} builder as a reference.\n")
+            withBlockTemplate(
+                "pub fn inner(&self) -> &#{Inner} {", "}",
```
jdisanti, line 415: What do you think about calling this `as_input()`? That should be less likely to name collide, and then people won't have to look at what the inner of the fluent builder is to understand what it's for.

Should we also add `to_input()` and `into_input()` for the owned return value side of things? The `to_input()` one may not be possible though unless the input is cloneable.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -92,6 +92,8 @@ class OperationBuildError(private val runtimeConfig: RuntimeConfig) {
 
 // Setter names will never hit a reserved word and therefore never need escaping.
 fun MemberShape.setterName() = "set_${this.memberName.toSnakeCase()}"
+// Setter names will never hit a reserved word and therefore never need escaping.
```
jdisanti, line 95:
```suggestion
// Getter names will never hit a reserved word and therefore never need escaping.
```

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGeneratorTest.kt
```diff
@@ -76,4 +88,50 @@ class FluentClientGeneratorTest {
...
+                            .endpoint_resolver("http://localhost:1234")
+                            #{set_http_connector}
+                            .build();
+                        let smithy_client = aws_smithy_client::Builder::new()
+                            .connector(connector.clone())
+                            .middleware_fn(|r| r)
+                            .build_dyn();
+                        let client = $moduleName::Client::with_config(smithy_client, config); 
+                    
+                        let say_hello_input_builder = client.say_hello().byte_value(4).foo("hello!");
```
jdisanti, line 112: Nit: should probably be `fluent_builder` instead of `input_builder`

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
DavidSouther: Done.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGeneratorTest.kt
DavidSouther: Done.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
DavidSouther: Love it. Done.

jdisanti (approved):
Thank you!

PR #2795, opened 2023-06-20
Make service config builder contain `CloneableLayer`

ysaito1001:
## Motivation and Context
This PR will reduce fields in service config builders to contain only `CloneableLayer` (except for `interceptors` and `identity_resolvers`).

## Description
This is the third PR in a series of config refactoring in the orchestrator mode. Just like #2762, this PR reduces fields in service config builders to contain `CloneableLayer`.

The code changes follow a straightforward pattern where builders' setters are implemented via a config layer.

## Testing
Relies on the existing tests in CI

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -103,13 +103,24 @@ class AwsFluentClientDecorator : ClientCodegenDecorator {
         baseGenerator.protocolSupport,
         baseGenerator.operationShape,
         renderClientCreation = { params ->
+            rust("let mut ${params.configBuilderName} = ${params.configBuilderName};")
+            if (codegenContext.smithyRuntimeMode.defaultToOrchestrator) {
+                // TODO(enableNewSmithyRuntimeLaunch): A builder field cannot be accessed directly in the orchestrator mode.
```
jdisanti, line 108: https://github.com/awslabs/smithy-rs/pull/2792 adds getters to the builders, so that should unblock this comment

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpConnectorConfigCustomization.kt
```diff
@@ -150,18 +154,33 @@ class HttpConnectorConfigCustomization(
...
                     *codegenScope,
                 )
-            }
-            is ServiceConfig.BuilderBuild -> writable {
                 if (runtimeMode.defaultToOrchestrator) {
-                    rust("layer.store_or_unset(self.http_connector);")
+                    rustTemplate(
+                        """
+                        pub fn set_http_connector(&mut self, http_connector: #{Option}<impl #{Into}<#{HttpConnector}>>) -> &mut Self {
+                            http_connector.map(|c| self.inner.store_put(c.into()));
```
jdisanti, line 164: Should we be using `ConfigBagAccessors` to set these directly to ensure the types match on both ends?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/ResiliencyConfigCustomization.kt
```diff
@@ -74,17 +76,17 @@ class ResiliencyConfigCustomization(codegenContext: ClientCodegenContext) : Conf
                         rustTemplate(
                             """
                             /// Return a reference to the retry configuration contained in this config, if any.
-                            pub fn retry_config(&self) -> Option<&#{RetryConfig}> {
+                            pub fn retry_config(&self) -> #{Option}<&#{RetryConfig}> {
```
jdisanti, line 79: Thanks for fixing these ❤️

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpConnectorConfigCustomization.kt
ysaito1001: Agreed, that's what we should. We actually use `ConfigBagAccessors` already in [ServiceRuntimePluginGenerator](https://github.com/awslabs/smithy-rs/blob/main/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/ServiceRuntimePluginGenerator.kt#LL165-L165) for setting up a connection, and that line of code will be moved to builder's build method in the next PR (by which time we are using `ConfigBagAccessors` in the `build` method to set up a `connection` when a service config is created). Within `set_http_connector`, I believe we can continue using `store_put` because "the other end" is right under its nose, e.g. the `build` method.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
ysaito1001: Thanks for letting me know. Updated the comment c41a167 accordingly.

jdisanti (approved):

PR #2802, opened 2023-06-21
Configure orchestrator's key components in builder's build method

ysaito1001:
## Motivation and Context
Moves setting orchestrator components out of `ServiceRuntimePlugin` and puts it in service config builders' build method.

## Description
This PR is the forth in a series of config refactoring. Here, we move pieces of code out of `ServiceRuntimePlugin::config` method so that key orchestrator components meant for the service-level config should only be constructed once when a service config is created, e.g. during builder's `build` method. Previously, those components were newly created every time an operation is invoked.

Wherever `self.handle.conf...` is used, the PR has moved it from `ServiceRuntimePlugin::config` to the builders' `build` method.

Note that there will be a separate PR to better handle auth resolver & identity resolver in the context of the ongoing config refactoring.

## Testing
- [x] Passed tests in CI

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpConnectorConfigDecorator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpConnectorConfigDecorator.kt
@@ -38,7 +38,12 @@ private class HttpConnectorConfigCustomization(
     private val moduleUseName = codegenContext.moduleUseName()
     private val codegenScope = arrayOf(
         *preludeScope,
+        "Connection" to RuntimeType.smithyRuntimeApi(runtimeConfig).resolve("client::orchestrator::Connection"),
+        "ConnectorSettings" to RuntimeType.smithyClient(runtimeConfig).resolve("http_connector::ConnectorSettings"),
+        "DynConnectorAdapter" to RuntimeType.smithyRuntime(runtimeConfig).resolve("client::connections::adapter::DynConnectorAdapter"),
         "HttpConnector" to RuntimeType.smithyClient(runtimeConfig).resolve("http_connector::HttpConnector"),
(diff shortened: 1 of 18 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpConnectorConfigDecorator.kt
```diff
@@ -183,7 +188,29 @@ private class HttpConnectorConfigCustomization(
             }
 
             is ServiceConfig.BuilderBuild -> writable {
-                if (runtimeMode.defaultToMiddleware) {
+                if (runtimeMode.defaultToOrchestrator) {
+                    rustTemplate(
+                        """
+                        let sleep_impl = layer.load::<#{SharedAsyncSleep}>().cloned();
+                        let timeout_config = layer.load::<#{TimeoutConfig}>().cloned().unwrap_or_else(|| #{TimeoutConfig}::disabled());
```
jdisanti, line 195: So if we set a `TimeoutConfig` on the top-level config, and then change the `http_connector` in a `config_override`, will it use the top-level `TimeoutConfig` or `TimeoutConfig::disabled()`?

ysaito1001: The implementation of `config_override` has been [under construction](https://github.com/awslabs/smithy-rs/blob/main/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt#L459) (and still is as of this PR), so I'm speculating from my mental model. I think it uses the top-level `TimeoutConfig`.

Since whatever is set through `config_override` is applied to the operation-level configuration, as opposed to the top-level service configuration, during the execution of the orchestrator, an operation runtime plugin will have access to a config bag that has accumulated layers so far (among which is a frozen layer from the top-level service config). So it _should_ use the top-level `TimeoutConfig` rather than `TimeoutConfig::disabled()`.

jdisanti (approved):

PR #2807, opened 2023-06-23
Remove `get`/`put` methods from `ConfigBag` and `Layer`

jdisanti:
This PR removes the non-storable `get`/`put` methods from `ConfigBag` and `Layer`.

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
@@ -244,6 +244,9 @@ class FluentClientGenerator(
                     }
                     """,
                     *clientScope,
+                    // TODO(enableNewSmithyRuntimeLaunch): Remove the handle config hack
+                    "Storable" to RuntimeType.smithyTypes(runtimeConfig).resolve("config_bag::Storable"),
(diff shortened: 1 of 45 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -244,6 +244,9 @@ class FluentClientGenerator(
                     }
                     """,
                     *clientScope,
+                    // TODO(enableNewSmithyRuntimeLaunch): Remove the handle config hack
+                    "Storable" to RuntimeType.smithyTypes(runtimeConfig).resolve("config_bag::Storable"),
+                    "StoreReplace" to RuntimeType.smithyTypes(runtimeConfig).resolve("config_bag::StoreReplace"),
```
ysaito1001, line 249: I cannot find occurrences of these in the above rust code snippet in `rustTemplate`, but assume something is still referring to them?

jdisanti: Oh, these can be removed.

jdisanti, line 249:
```suggestion
```

PR #2808, opened 2023-06-24
Fix optional auth in the orchestrator

jdisanti:
This PR:
- Renames "anonymous auth" to "no auth"
- Removes fallback to other auth schemes when an identity fails to resolve (this was not complying to the reference architecture)
- Adds the ability to opt out of credentials in `ConfigLoader`, and removes defaulting of the shared credentials cache if no credentials provider is given
- Makes `ConfigBagAccessors` work on `FrozenLayer` and `CloneableLayer`
- Fixes STS and aws-config tests

----


aws/rust-runtime/aws-config/src/default_provider/credentials.rs
```diff
@@ -161,9 +161,9 @@ impl Builder {
     }
 
     /// Override the configuration used for this provider
-    pub fn configure(mut self, config: ProviderConfig) -> Self {
-        self.region_chain = self.region_chain.configure(&config);
-        self.conf = Some(config);
+    pub fn configure(mut self, config: &ProviderConfig) -> Self {
+        self.region_chain = self.region_chain.configure(config);
+        self.conf = Some(config.clone());
```
ysaito1001, line 166: I know _in general_, we have [this API guideline](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control) where if the implementation ends up using an owned value, the function/method should take as input an owned value from the beginning, rather than taking it by reference and internally clone it. What do we think?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationRuntimePluginGenerator.kt
```diff
@@ -106,4 +120,45 @@ class OperationRuntimePluginGenerator(
...
+                        ?: throw IllegalStateException("missing 'no auth' implementation")
+                    authOption.constructor(this)
+                    atLeastOneScheme = true
+                }
+                if (!atLeastOneScheme) {
+                    throw IllegalStateException(
+                        "this client won't have any auth schemes " +
+                            "(not even optional/no-auth auth), which means the generated client " +
+                            "won't work at all for the ${operationShape.id} operation",
+                    )
```
Velfi, line 159: We should consider linking to smithy docs which explain auth schemes. I'm guessing this will come up for some people generating clients internally.

rust-runtime/aws-smithy-runtime-api/src/client/config_bag_accessors.rs
```diff
@@ -0,0 +1,443 @@
...
+use crate::client::identity::{
+    ConfiguredIdentityResolver, IdentityResolvers, SharedIdentityResolver,
+};
+use crate::client::orchestrator::{
+    Connection, DynConnection, DynEndpointResolver, DynResponseDeserializer, EndpointResolver,
+    EndpointResolverParams, LoadedRequestBody, ResponseDeserializer, SharedRequestSerializer,
+    NOT_NEEDED,
+};
+use crate::client::retries::RetryStrategy;
+use crate::client::retries::{DynRetryStrategy, RetryClassifiers};
```
Velfi, line 19:
```suggestion
use crate::client::retries::{DynRetryStrategy, RetryClassifiers, RetryStrategy};
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationRuntimePluginGenerator.kt
jdisanti: Good call. Fixed in [040d3fe](https://github.com/awslabs/smithy-rs/pull/2808/commits/040d3fe6221420994aa93c776b32d37d87a3cb1d)

Velfi (approved):
Great work!

Issue #2810, opened 2023-06-26
Remove special handling logic for the old SDK examples directory layout


jdisanti:
Address the `TODO(https://github.com/awslabs/smithy-rs/issues/2810)` comments after confirming that there is no longer a `Cargo.toml` file in `aws-sdk-rust/examples/` in the aws-sdk-rust main branch.

PR #2811, opened 2023-06-27
Migrate to a new SDK example workspace structure

jdisanti:
## Motivation and Context
When the WebAssembly SDK example was added some months ago, we changed the build process to make each SDK example its own Cargo workspace. This allowed the `.cargo/config.toml` that changed the compiler target to work correctly. However, this has led to other issues: dependency compilation is no longer shared between examples which greatly increases the time it takes for that CI step to run, and now it is even running out of disk space on the GitHub Actions runners.

This PR adds support for a new workspace layout where the [`aws-doc-sdk-examples`](https://github.com/awsdocs/aws-doc-sdk-examples) repo gets to decide how the workspaces are logically grouped. If a `Cargo.toml` file exists at the example root, then the build system assumes that the _old_ "one example, one workspace" layout should be used. If there is no root `Cargo.toml`, then it assumes the new layout should be used.

The `sdk-versioner` tool had to be adapted to support more complex relative path resolution to make this work, and the `publisher fix-manifests` subcommand had to be fixed to ignore workspace-only `Cargo.toml` files.

The build system in this PR needs to work for both the old and new examples layout so that the `sdk-sync` process will succeed. #2810 has been filed to track removing the old example layout at a later date.

[aws-doc-sdk-examples#4997](https://github.com/awsdocs/aws-doc-sdk-examples/pull/4997) changes the workspace structure of the actual examples to the new one.

## Testing
- [x] Generated a full SDK with the old example layout, manually examined the output, and spot checked that some examples compile
- [x] Generated a full SDK with the new example layout, manually examined the output, and spot checked that some examples compile
- [x] Examples pass in CI with the old example layout
- [x] Examples pass in CI with the new example layout

----


buildSrc/src/main/kotlin/aws/sdk/ServiceLoader.kt
```diff
@@ -19,6 +19,38 @@ data class RootTest(
...
+     * since the dependencies would get recompiled for every example.
+     */
+    Flat,
+
+    /**
+     * Current directory layout where there are a small number of workspaces
+     * rooted in `rust_dev_preview/`.
+     */
+    Workspaces,
+    ;
```
Velfi, line 43: it's so weird to see this little guy hanging out on his own.

tools/ci-build/publisher/src/package.rs
```diff
@@ -185,14 +184,19 @@ pub async fn discover_package_manifests(path: PathBuf) -> Result<Vec<PathBuf>> {
...
-                manifests.extend(discover_package_manifests(package_path).await?.into_iter());
             }
+            manifests.extend(discover_manifests(package_path).await?.into_iter());
         }
     }
     Ok(manifests)
 }
 
+/// Discovers and parses all Cargo.toml files that are packages (as opposed to being exclusively workspaces)
+pub async fn discover_packages(fs: Fs, path: PathBuf) -> Result<Vec<Package>> {
```
Velfi, line 195: "crate" would be more accurate than "package", no?

Velfi (approved):

tools/ci-build/publisher/src/package.rs
jdisanti: Not going to rename this since it's referred to as "package" throughout.

tools/ci-build/publisher/src/subcommand/claim_crate_names.rs
```diff
@@ -79,20 +77,11 @@ async fn discover_publishable_crate_names(repository_root: &Path) -> anyhow::Res
...
-            })?;
-            let package = manifest
-                .package
-                .ok_or(Error::InvalidManifest(manifest_path))
-                .context("crate manifest doesn't have a `[package]` section")?;
-            let name = package.name;
-            if let cargo_toml::Publish::Flag(true) = package.publish {
-                publishable_package_names.insert(name);
+        for package in packages {
+            if let Publish::Allowed = package.publish {
```
ysaito1001, line 83: Mostly for my understanding. Is it ok to change `Publish::Flag(true)` (previous code) to `Publish::Allowed`?

ysaito1001 (approved):

tools/ci-build/publisher/src/subcommand/claim_crate_names.rs
jdisanti: They're different types. `Publish::Flag` is part of the `cargo_toml` library, whereas `Publish::Allowed` is from `crate::package`. This change made it use our existing package code rather than directly parsing the manifest and interacting with that parse result.

PR #2814, opened 2023-06-28
Implement the remaining part of supporting operation level configuration

ysaito1001:
## Motivation and Context
Implements the actual meat of `config_override` introduced [as a skeleton in the past](https://github.com/awslabs/smithy-rs/pull/2589).

## Description
This PR enables operation-level config via `config_override` on a customizable operation (see [config-override.rs](https://github.com/awslabs/smithy-rs/blob/8105dd46b43854e7909aed82c85223414bc85df5/aws/sdk/integration-tests/s3/tests/config-override.rs) integration tests for example code snippets).

The way it's implemented is through `ConfigOverrideRuntimePlugin`. The plugin holds onto two things: a `Builder` passed to `config_override` and a `FrozenLayer` derived from a service config (the latter is primarily for retrieving  default values understood by a service config). The plugin then implements the `RuntimePlugin` trait to generate its own `FrozenLayer` that contains operation-level orchestrator components. That `FrozenLayer` will then be added to a config bag later in the orchestrator execution in a way that it takes precedence over the client-level configuration (see [register_default_runtime_plugins](https://github.com/awslabs/smithy-rs/blob/8105dd46b43854e7909aed82c85223414bc85df5/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationGenerator.kt#L65-L71)).

A couple of things to note:
- The size of `ConfigOverrideRuntimePlugin::config` code-generated is getting large. Maybe each customization defines a helper function instead of inlining logic directly in `config` and we let the `config` method call those generated helpers.
- The PR does not handle a case where `retry_partition` within `config_override` since it's currently `#[doc(hidden)]`, e.g.
```
client
    .some_operation()
    .customize()
    .await
    .unwrap()
    .config_override(Config::builder().retry_partition(/* ... */))
    ...
```

## Testing
- Added tests in Kotlin [ConfigOverrideRuntimePluginGeneratorTest.kt](https://github.com/awslabs/smithy-rs/pull/2814/files#diff-04a76a43c2adb3a2ee37443157c68ec6dad77fab2484722b370a7ba14cf02086) and [CredentialCacheConfigTest.kt](https://github.com/awslabs/smithy-rs/pull/2814/files#diff-32246072688cd11391fa10cd9cb38a80ed88b587e95037225dbe9f1a482ebc5d). ~~These tests are minimal in that they only check if the appropriate orchestrator components are inserted into a config override layer when a user calls a certain builder method as part of `config_override`.~~
- Added integration tests [config-override.rs](https://github.com/awslabs/smithy-rs/pull/2814/files#diff-6fd7a1e70b1c3fa3e9c747925f3fc7a6ce0f7b801bd6bc46c54eec44150f5158)

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialCaches.kt
```diff
@@ -191,6 +191,48 @@ class CredentialCacheConfig(codegenContext: ClientCodegenContext) : ConfigCustom
                 }
             }
 
+            ServiceConfig.RuntimePluginConfig("layer") -> {
```
jdisanti, line 194: Could all this logic be computed once during ServiceRuntimePlugin construction rather than per request?

I'm currently working on a PR that makes the ServiceRuntimePlugin constructed once per client rather than per request.

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/CredentialCacheConfigTest.kt
```diff
@@ -0,0 +1,150 @@
...
+                        *codegenScope,
+                    )
+                }
+
+                unitTest("test_overriding_credentials_cache_leads_to_shared_credentials_cache_in_layer") {
+                    rustTemplate(
+                        """
+                        use #{RuntimePlugin};
+
+                        // TODO(enableNewSmithyRuntimeCleanup): Should not need to provide a default once smithy-rs##2770
```
jdisanti, line 86: What default is this referring to?

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/CredentialCacheConfigTest.kt
```diff
@@ -0,0 +1,150 @@
...
+                        let sut_layer = sut.config().unwrap();
+                        assert!(sut_layer
+                            .load::<#{SharedCredentialsCache}>()
+                            .is_some());
+                        """,
+                        *codegenScope,
+                    )
+                }
+
+                unitTest("test_not_overriding_cache_and_provider_leads_to_no_shared_credentials_cache_in_layer") {
```
jdisanti, line 128: Should there also be a test for:
```rust
                        let client_config = Config::builder()
                            .credentials_cache(#{CredentialsCache}::lazy())
                            .credentials_provider(#{Credentials}::for_tests())
                            .build();
                        let config_override = Config::builder().build();
```
?

What about:
```rust
                        let client_config = Config::builder()
                            .credentials_cache(#{CredentialsCache}::lazy())
                            .credentials_provider(#{Credentials}::for_tests())
                            .build();
                        let config_override = Config::builder()
                            .credentials_cache(#{CredentialsCache}::lazy())
                            .credentials_provider(<<<something else here>>>)
                            .build();
```
?

It should be possible to make a custom credentials provider and resolve credentials to verify it used the right one.

jdisanti (commented):
This is looking good. Some comments on making ServiceRuntimePlugin more efficient, and on testing.

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/CredentialCacheConfigTest.kt
ysaito1001: Admittedly, the wording could be better. What default refers to is the client-level config in this case. Since `config_override` layer is overriding the client-level config, it sometimes needs to grab a value from the client-level config rather than a "pure" default value when newly constructing an orchestrator component within the `config_override` layer.

For instance, if a user just `config_override`s `credentials_cache`, we probably should respect whatever `credentials_provider` specified in the client-level config, rather than an arbitrarily chosen default (like `Credentials::for_tests`, although that's a bad example but hopefully you get the idea).

jdisanti: Safest option is probably to require both of them to be set if one of them is set (for now).

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialCaches.kt
jdisanti: Sorry, I think I misunderstood where the `ServiceConfig.RuntimePluginConfig` hook was emitting code into. I was thinking it was going into the `ServiceRuntimeConfig::config` method, but reading it closer, it looks like it goes somewhere into the service config. So I think it's fine the way it is.

Note that matching on `ServiceConfig.RuntimePluginConfig("layer")` is probably not what you want though:
```suggestion
            is ServiceConfig.RuntimePluginConfig -> {
```

ysaito1001: No worries, thanks for asking this question. Updated the section name (95318ea) and fixed the matching (9cc52cf).

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/CredentialCacheConfigTest.kt
ysaito1001: Updated tests more thoroughly in 0e94e37.

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/CredentialCacheConfigTest.kt
ysaito1001: That works, updated in a78489d.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
```diff
@@ -322,7 +322,16 @@ class AwsPresignedFluentBuilderMethod(
 
             let runtime_plugins = #{Operation}::operation_runtime_plugins(
                 self.handle.runtime_plugins.clone(),
-                self.config_override
+                self.config_override.map(|config_override| {
+                    crate::config::ConfigOverrideRuntimePlugin {
```
jdisanti, line 326: Should `Operation::operation_runtime_plugins` do this mapping? It looks like its identical in both places it is done.

jdisanti (approved):
This is awesome! Great work. Just one suggestion.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
ysaito1001: 165a013 now avoids a map operation at call sites and push that responsibility to the function body of `operation_runtime_plugins` which already has a check `if let Some(config_override) = config_override`.

PR #2815, opened 2023-06-28
Expand skipped headers for sigv4 canonical request signing to include x-amzn-trace-id and authorization headers.

relevantsam:
## Motivation and Context
When customers add x-ray headers to requests, the SigV4 signer should exclude them, or the generated canonical signature will not match the remote service's, since many services being called are written with non-rust SDKs that automatically exclude these common headers.

The Rust SDK should exclude a similar set of headers to the other popular AWS SDKs. While this is not uniform across the SDKs, a minimal set should be excluded and others should be considered to be excluded in future PRs.

## Description

* Expands the set of headers excluded from canonical request calculation to include "x-amzn-trace-id" and "authorization" (since authorization will be added as a part of this process).

## Testing

* Added headers to exclusion test & validated with `cargo test`
* `./gradlew :aws:sdk:test`  

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -12,9 +12,9 @@
 # author = "rcoh"
 
  [[aws-sdk-rust]]
- message = "Automatically exclude x-ray trace id headers and authorization headers from Sigv4 canonical request calculations."
- references = ["smithy-rs#2813"]
- meta = { "breaking" = false, "tada" = false, "bug" = false }
+ message = "Automatically exclude X-Ray trace ID headers and authorization headers from SigV4 canonical request calculations."
+ references = ["smithy-rs#2815"]
+ meta = { "breaking" = false, "tada" = false, "bug" = true }
  author = "relevantsam"
 
  [[aws-sdk-rust]]
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -12,6 +12,12 @@
 # author = "rcoh"
 
  [[aws-sdk-rust]]
+ message = "Automatically exclude x-ray trace id headers and authorization headers from Sigv4 canonical request calculations."
+ references = ["smithy-rs#2813"]
+ meta = { "breaking" = false, "tada" = false, "bug" = false }
+ author = "relevantsam"
+
+ [[aws-sdk-rust]]
  message = "Add accessors to Builders"
  references = ["smithy-rs#2791"]
  meta = { "breaking" = false, "tada" = false, "bug" = false }
--- a/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
+++ b/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
@@ -774,14 +774,16 @@ mod tests {
         assert_eq!(creq.values.signed_headers().as_str(), "host;x-amz-date");
     }
 
-    // It should exclude user-agent and x-amz-user-agent headers from presigning
+    // It should exclude authorization, user-agent, x-amz-user-agent, x-amzn-trace-id headers from presigning
     #[test]
     fn presigning_header_exclusion() {
         let request = http::Request::builder()
             .uri("https://some-endpoint.some-region.amazonaws.com")
+            .header("authorization", "test-authorization")
             .header("content-type", "application/xml")
             .header("content-length", "0")
             .header("user-agent", "test-user-agent")
+            .header("x-amzn-trace-id", "test-trace-id")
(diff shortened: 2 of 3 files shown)


CHANGELOG.next.toml
```diff
@@ -11,6 +11,12 @@
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
 
+ [[aws-sdk-rust]]
+ message = "Automatically exclude x-ray trace id headers and authorization headers from Sigv4 canonical request calculations."
+ references = ["smithy-rs#2813"]
```
relevantsam, line 16: I'll fix this with any other changes that are requested + squashing the rebase pull that happened here

aws/rust-runtime/aws-sigv4/src/http_request/settings.rs
```diff
@@ -3,12 +3,14 @@
...
  */
 
-use http::header::{HeaderName, USER_AGENT};
+use http::header::{HeaderName, AUTHORIZATION, USER_AGENT};
 use std::time::Duration;
 
 /// HTTP signing parameters
 pub type SigningParams<'a> = crate::SigningParams<'a, SigningSettings>;
 
+const X_RAY_TRACE_HEADER: &str = "x-amzn-trace-id";
```
jdisanti, line 12: [`HeaderName::from_static`](https://docs.rs/http/0.2.9/http/header/struct.HeaderName.html#method.from_static) is `const`, so you can actually put the entire `HeaderName` in this constant rather than creating it every time signing settings is created.

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
```diff
@@ -774,14 +774,16 @@ mod tests {
         assert_eq!(creq.values.signed_headers().as_str(), "host;x-amz-date");
     }
 
-    // It should exclude user-agent and x-amz-user-agent headers from presigning
+    // It should exclude authorization, user-agent, x-amz-user-agent, x-amzn-trace-id headers from presigning
```
jdisanti, line 777: Huh, I'm surprised this is only tested for presigning. Would you be willing to copy/paste this test and modify it for non-presigning? The `SignatureLocation` would get changed to `Headers`, `expires_in` would not be set, and `canonical.values.into_query_params()` would change to `canonical.values.as_headers()`. Otherwise, it should be identical.

CHANGELOG.next.toml
```diff
@@ -11,6 +11,12 @@
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
 
+ [[aws-sdk-rust]]
+ message = "Automatically exclude x-ray trace id headers and authorization headers from Sigv4 canonical request calculations."
```
jdisanti, line 15:
```suggestion
 message = "Automatically exclude X-Ray trace ID headers and authorization headers from SigV4 canonical request calculations."
```

CHANGELOG.next.toml
```diff
@@ -11,6 +11,12 @@
...
+ references = ["smithy-rs#2813"]
+ meta = { "breaking" = false, "tada" = false, "bug" = false }
```
jdisanti, line 17:
```suggestion
 meta = { "breaking" = false, "tada" = false, "bug" = true }
```

jdisanti (commented):
Thank you for fixing this!

aws/rust-runtime/aws-sigv4/src/http_request/settings.rs
relevantsam: 👍 I got caught up trying to make the array static

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
relevantsam: There are a couple other changes here. 

1. x-amz-user-agent actually only gets filtered for presigning: https://github.com/awslabs/smithy-rs/blob/main/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs#L253 so `x-amz-user-agent` is actually still present
2. x-amz-date is added on non-presigning requests

relevantsam:
```
12 | const X_RAY_TRACE_HEADER: HeaderName = HeaderName::from_static("x-amzn-trace-id");
```
We looked at suppressing the clippy warning and adding a const array. The Vec input for SigningSettings reduces the value for this, so we went with the original approach.

jdisanti (approved):

PR #2818, opened 2023-06-28
Profile mutation and serialization

jonathanmorley:
## Motivation and Context

Supports mutating `ProfileSet`s, so that applications that fetch STS credentials can update a `ProfileSet` with the new credentials.

Supports serializing `ProfileSet`s, so that those credentials can be written back to disk.

## Testing

New test added to `src/profile/parser.rs`

## Checklist
----

### DIFF
--- a/aws/rust-runtime/aws-config/src/profile/parser.rs
+++ b/aws/rust-runtime/aws-config/src/profile/parser.rs
@@ -5,10 +5,10 @@
 
 use crate::profile::parser::parse::parse_profile_file;
 use crate::profile::parser::source::Source;
-use crate::profile::profile_file::ProfileFiles;
+use crate::profile::profile_file::{ProfileFileKind, ProfileFiles};
 use aws_types::os_shim_internal::{Env, Fs};
 use std::borrow::Cow;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::error::Error;
 use std::fmt::{Display, Formatter};
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -114,6 +114,11 @@ impl ProfileSet {
         self.profiles.get(profile_name)
     }
 
+    /// Set a named profile in the profile set
```
rcoh, line 117: let's add docs that clarify that this will replace another profile with the same name. Might as well change the API to return the removed profile if it existed as well

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -129,6 +134,35 @@ impl ProfileSet {
         self.profiles.keys().map(String::as_ref)
     }
 
+    /// Returns a string representation of the profiles, with syntax based on `kind`.
+    pub fn to_string(&self, kind: ProfileFileKind) -> String {
```
rcoh, line 138: I think we'll need a few caveats in this method:
1. It is going to delete any comments the user may have had
2. We need to test how it handles multi-line values, e.g. for S3:
```
[profile development]
s3 =
  max_concurrent_requests = 20
  max_queue_size = 10000
  multipart_threshold = 64MB
  multipart_chunksize = 16MB
  max_bandwidth = 50MB/s
  use_accelerate_endpoint = true
  addressing_style = path
```

rcoh (commented):
I think ultimately, this isn't going to be the right approach. A full round trip of the file is probably not going to be what users want—I think we will almost certainly want to append to the file instead. That way we can preserve comments, keep the users profile in order, and not worry about unsupported features. 

For that reason, I think the thing I'd add serializability for is a profile itself. That also removes the need to make `ProfileSet` mutable.

When parsing profiles, we are able to keep track of line number information. If we include that, you could even update your profile by tracking the offsets.

I don't think we'll be able to merge this PR as is unfortunately for the reasons above.

rcoh:
closing this for now—let us know (or submit a PR) if you'd like to make some more public methods

PR #2820, opened 2023-06-29
Python: Make sure to log errors from async handlers

unexge:
## Motivation and Context
We were skipping logging of the errors occurred during execution of asynchronous Python handlers due to the question mark at the end:
```
result.await.map(|r| #{pyo3}::Python::with_gil(|py| r.extract::<$output>(py)))? // <-
```

Any exception thrown in asynchronous Python handlers will be logged properly with this change.

## Checklist

----


crisidev (approved):
Uh, good catch

PR #2824, opened 2023-06-29
add connection poisoning support to the orchestrator

Velfi:
This PR updates the orchestrator to support connection poisoning for the `hyper` connector. It also renames many usages of "connection" to "connector" in order to make talking about these concepts less confusing. In short:

```
/// A "connector" manages one or more "connections", handles connection timeouts, re-establishes
/// connections, etc.
///
/// "Connections" refers to the actual transport layer implementation of the connector.
/// By default, the orchestrator uses a connector provided by `hyper`.
```

One possibly controversial decision I made is that `reconnect_mode` is now a standalone configurable field, rather that a sub-field of `RetryConfig`. I think we should get together as a team and discuss what kinds of things we want to allow users to do when configuring connections.

My thoughts on this are that we should either:

- **A.** Make connection-related settings their own types instead of including them within other config types
- **B.** Make a single config struct for all connector-related stuff/use the preëxisting `ConnectorSettings` struct.

Personally, I'm partial to A.

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/ResiliencyConfigCustomization.kt
```diff
@@ -430,12 +430,12 @@ class ResiliencyReExportCustomization(private val runtimeConfig: RuntimeConfig)
         }
         rustCrate.withModule(ClientRustModule.Config.retry) {
             rustTemplate(
-                "pub use #{types_retry}::{RetryConfig, RetryConfigBuilder, RetryMode};",
+                """
+                pub use #{types_retry}::{RetryConfig, RetryConfigBuilder, RetryMode, ReconnectMode};
+                pub use #{runtime_types_retry}::RetryPartition;
```
jdisanti, line 435: This should only be re-exported in orchestrator mode.

Also, it looks like aws-smithy-runtime snuck into middleware mode... so if you want to track that down, that would be fantastic 😆

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context/wrappers.rs
```diff
@@ -158,6 +158,15 @@ declare_wrapper!(
...
 );
 
+impl<'a, I, O, E: Debug> BeforeDeserializationInterceptorContextMut<'a, I, O, E> {
+    /// Downgrade this helper struct, returning the underlying InterceptorContext. There's no good
+    /// reason to use this unless you're writing tests or you have to interact with an API that
+    /// doesn't support the helper structs.
+    pub fn into_inner(&mut self) -> &'_ mut InterceptorContext<I, O, E> {
+        self.inner
+    }
+}
```
jdisanti, line 168: Perhaps we should modify the RetryClassifier trait to just take what it needs from the context rather than the whole context.

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        if error_is_transient && connection_poisoning_is_enabled {
+            debug!("received a transient error, poisoning the connection...");
+
+            if let Some(captured_connection) = captured_connection.and_then(|conn| conn.get()) {
+                captured_connection.poison();
+                debug!("the connection was poisoned")
+            } else {
+                error!(
+                    "no connection found! The underlying HTTP connector never set a connection."
```
jdisanti, line 85:
```suggestion
                    "unable to poison the connection because no connection was found! The underlying HTTP connector never set a connection."
```

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
```diff
@@ -0,0 +1,132 @@
...
+            .load::<RetryConfig>()
+            .map(RetryConfig::reconnect_mode)
+            .unwrap_or(ReconnectMode::ReconnectOnTransientError);
+        let captured_connection = cfg.load::<CaptureSmithyConnectionWrapper>().cloned();
+        let retry_classifiers = cfg.retry_classifiers();
+
+        let error_is_transient = retry_classifiers
+            .classify_retry(context.into_inner())
+            .map(|reason| reason == RetryReason::Error(ErrorKind::TransientError))
+            .unwrap_or_default();
```
jdisanti, line 73: Maybe the orchestrator should add the retry classification to the ConfigBag?

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
```diff
@@ -0,0 +1,132 @@
...
+        Ok(())
+    }
+}
+
+// TODO(enableNewSmithyRuntimeLaunch) We won't need this once we absorb aws_smithy_http into the
+//    new runtime crate.
+#[derive(Clone, Default)]
+pub struct CaptureSmithyConnectionWrapper {
+    inner: CaptureSmithyConnection,
+}
```
jdisanti, line 99: You can just copy/paste CaptureSmithyConnection into aws-smithy-runtime, and add a cleanup TODO to remove it from aws_smithy_http.

rust-runtime/aws-smithy-types/src/retry.rs
```diff
@@ -277,6 +277,7 @@ pub struct RetryConfig {
     max_attempts: u32,
     initial_backoff: Duration,
     max_backoff: Duration,
+    // TODO(enableSmithyRuntimeLaunch) remove this field. It's now its own setting.
```
jdisanti, line 280: Out of date?

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
Velfi: I can't b/c then I'd also have to move the hyper adapter that references it. I can do it as a separate PR after this one if you like, though.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustType.kt
```diff
@@ -6,6 +6,7 @@
 package software.amazon.smithy.rust.codegen.core.rustlang
 
 import software.amazon.smithy.rust.codegen.core.rustlang.Attribute.Companion.derive
+import software.amazon.smithy.rust.codegen.core.rustlang.Attribute.Companion.serde
```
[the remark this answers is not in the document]
Velfi: Yeah

rust-runtime/aws-smithy-types/src/retry.rs
Velfi: good catch, I forgot this guy

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/ResiliencyConfigCustomization.kt
Velfi: Good catch. I found a few of them.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context/wrappers.rs
Velfi: Maybe, although I like passing the whole thing because it feels more future-proof. However, will we ever need to check anything other than the response pre- and post-serialization? I can't think of why we would.

I'm OK making this change but I'd rather do it as a separate PR.

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
Velfi: Yeah, I can see how that could be useful to users. However, the retry strategy doesn't have write access to the bag, and it's the thing that's calling the error classifier (not the orchestrator.)

I'm going to leave this as-is for now.

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
jdisanti: Ah, OK. Nevermind then :)

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context/wrappers.rs
jdisanti: Maybe make this `#[doc(hidden)]` in the mean time.

PR #2825, opened 2023-06-30
Fix the S3 alternate runtime retries test in orchestrator mode

jdisanti:
This PR fixes the S3 `alternative-async-runtime` retry tests. The retry backoff time was being included in the operation attempt timeout, which I think is undesirable as it makes retry config much harder to get right since the backoff times are not predictable (they have randomness incorporated into them). The overall operation timeout is the only one that needs to incorporate backoff times.

In addition, this PR also:
- Updates READMEs for the `aws-smithy-runtime-api` and `aws-smithy-runtime` crates
- Adds top-level crate docs to describe features to `aws-smithy-runtime`
- Copies `capture_test_logs` into `aws-smithy-runtime` so that it can be used (just about) everywhere instead of just in `aws-config`
- Adds service/operation name to the tracing `invoke` span so it's possible to tell which operation the events are for
- Makes the `Debug` impl for `Identity` useful
- Adds a ton of trace/debug spans and events to the orchestrator
- Fixes an issue in `aws-smithy-runtime` where a huge amount of the orchestrator tests weren't being compiled due to a removed feature flag

----


rust-runtime/aws-smithy-runtime-api/README.md
```diff
@@ -1,8 +1,8 @@
-# aws-smithy-retries
+# aws-smithy-runtime-api
```
Velfi, line 1: lol I'm surprised it took us so long to notice this one.

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -138,12 +159,15 @@ async fn try_op(
     stop_point: StopPoint,
 ) {
     // Before serialization
+    trace!("running read_before_serialization interceptors");
     halt_on_err!([ctx] => interceptors.read_before_serialization(ctx, cfg));
+    trace!("running modify_before_serialization interceptors");
     halt_on_err!([ctx] => interceptors.modify_before_serialization(ctx, cfg));
```
Velfi, line 165: Consider calling `trace` from the `Interceptors` thing. That way, there's less lines in the invoke methods.

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -196,23 +223,25 @@ async fn try_op(
...
+        }
         let attempt_timeout_config = cfg.maybe_timeout_config(TimeoutKind::OperationAttempt);
+        trace!(attempt_timeout_config = ?attempt_timeout_config);
         let maybe_timeout = async {
-            // We must await this here or else timeouts won't work as expected
-            if let Some(delay) = retry_delay.take() {
-                delay.await;
-            }
-
+            debug!("beginning attempt #{i}");
```
Velfi, line 244: Thanks for fixing this. I really wasn't sure that I'd gotten it right.

Velfi (approved):

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
jdisanti: Good idea...

Velfi (approved):

PR #2827, opened 2023-07-03
Better distinguish model and HTTP plugins

david-perez:
So far, servers have tacitly worked with the notion that plugins come in
two flavors: "HTTP plugins" and "model plugins":

- A HTTP plugin acts on the HTTP request before it is deserialized, and
  acts on the HTTP response after it is serialized.
- A model plugin acts on the modeled operation input after it is
  deserialized, and acts on the modeled operation output or the modeled
  operation error before it is serialized.

However, this notion was never reified in the type system. Thus, users
who pass in a model plugin where a HTTP plugin is expected or
viceversa in several APIs:

```rust
let pipeline = PluginPipeline::new().push(http_plugin).push(model_plugin);

let app = SimpleService::builder_with_plugins(http_plugins, IdentityPlugin)
    .post_operation(handler)
    /* ... */
    .build()
    .unwrap();
```

would get the typical Tower service compilation errors, which can get
very confusing:

```
error[E0631]: type mismatch in function arguments
  --> simple/rust-server-codegen/src/main.rs:47:6
   |
15 | fn new_svc<S, Ext>(inner: S) -> model_plugin::PostOperationService<S, Ext> {
   | -------------------------------------------------------------------------- found signature defined here
...
47 |     .post_operation(post_operation)
   |      ^^^^^^^^^^^^^^ expected due to this
   |
   = note: expected function signature `fn(Upgrade<RestJson1, (PostOperationInput, _), PostOperationService<aws_smithy_http_server::operation::IntoService<simple::operation_shape::PostOperation, _>, _>>) -> _`
              found function signature `fn(aws_smithy_http_server::operation::IntoService<simple::operation_shape::PostOperation, _>) -> _`
   = note: required for `LayerFn<fn(aws_smithy_http_server::operation::IntoService<..., ...>) -> ... {new_svc::<..., ...>}>` to implement `Layer<Upgrade<RestJson1, (PostOperationInput, _), PostOperationService<aws_smithy_http_server::operation::IntoService<simple::operation_shape::PostOperation, _>, _>>>`
   = note: the full type name has been written to '/local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/target/debug/deps/simple-6577f9f79749ceb9.long-type-4938700695428041031.txt'
```

This commit introduces the `HttpPlugin` and `ModelPlugin` marker traits,
allowing plugins to be marked as an HTTP plugin, a model plugin, or
both. It also removes the primary way of concatenating plugins,
`PluginPipeline`, in favor of `HttpPlugins` and `ModelPlugins`, which
eagerly check that whenever a plugin is `push`ed, it is of the expected
type. The generated service type in the server SDKs'
`builder_with_plugins` constructor now takes an `HttpPlugin` as its
first parameter, and a `ModelPlugin` as its second parameter.

I think this change perhaps goes counter to the generally accepted
wisdom that trait bounds in Rust should be enforced "at the latest
possible moment", that is, only when the behavior encoded in the trait
implementation is relied upon in the code (citation needed).
However, the result is that exposing the concepts of HTTP plugin and
model plugin in the type system makes for code that is easier to reason
about, and produces more helpful compiler error messages.

Documentation about the plugin system has been expanded, particularly on
how to implement model plugins.

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -210,7 +210,7 @@ message = """The middleware system has been reworked as we push for a unified, s
 
 - A `ServiceShape` trait has been added.
 - The `Plugin` trait has been simplified.
-- The `HttpPlugin` and `ModelPlugin` marker traits have been added to better distinguish when plugins run and what they have access to.
+- The `HttpMarker` and `ModelPlugin` marker traits have been added to better distinguish when plugins run and what they have access to.
 - The `Operation` structure has been removed.
 - A `Scoped` `Plugin` has been added.
 
@@ -373,7 +373,7 @@ where
     }
 }
 
-impl HttpPlugin for PrintPlugin { }
+impl HttpMarker for PrintPlugin { }
 ```
 
 Alternatively, using the new `ServiceShape`, implemented on `Ser`:
@@ -400,7 +400,7 @@ let app = PokemonService::builder_with_plugins(/* HTTP plugins */, /* model plug
     .unwrap();
 ```
 
-To better distinguish when a plugin runs and what it has access to, `Plugin`s now have to additionally implement the `HttpPlugin` marker trait, the `ModelPlugin` marker trait, or both:
+To better distinguish when a plugin runs and what it has access to, `Plugin`s now have to additionally implement the `HttpMarker` marker trait, the `ModelPlugin` marker trait, or both:
 
 - A HTTP plugin acts on the HTTP request before it is deserialized, and acts on the HTTP response after it is serialized.
 - A model plugin acts on the modeled operation input after it is deserialized, and acts on the modeled operation output or the modeled operation error before it is serialized.
@@ -423,7 +423,7 @@ let layer = LayerPlugin::new::<SomeProtocol, SomeOperation>(plugin);
 
 ## Removal of `PluginPipeline`
 
-Since plugins now come in two flavors (`HttpPlugin` and `ModelPlugin`) that shouldn't be mixed in a collection of plugins, the primary way of concatenating plugins, `PluginPipeline` has been removed in favor of the `HttpPlugins` and `ModelPlugins` types, which eagerly check that whenever a plugin is pushed, it is of the expected type.
+Since plugins now come in two flavors (those marked with `HttpMarker` and those marked with `ModelPlugin`) that shouldn't be mixed in a collection of plugins, the primary way of concatenating plugins, `PluginPipeline` has been removed in favor of the `HttpPlugins` and `ModelPlugins` types, which eagerly check that whenever a plugin is pushed, it is of the expected type.
 
 This worked before, but you wouldn't be able to do apply this collection of plugins anywhere; if you tried to, the compilation error messages would not be very helpful:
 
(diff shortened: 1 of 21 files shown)


rust-runtime/aws-smithy-http-server/src/plugin/layer.rs
```diff
@@ -23,6 +23,12 @@ where
     }
 }
 
+// Without more information about what the layer `L` does, we can't know whether it's appropriate
```
david-perez, line 26: Not too happy about this. With `PluginLayer` and LayerPlugin`, a user can effectively circumvent the whole point of this PR by converting an e.g. model plugin to a layer and using `LayerPlugin` to get a HTTP plugin.

rust-runtime/aws-smithy-http-server/src/plugin/mod.rs
```diff
@@ -188,3 +244,222 @@ where
         <Pl as Plugin<Ser, Op, T>>::apply(self, inner)
     }
 }
+
+/// A HTTP plugin is a plugin that acts on the HTTP request before it is deserialized, and acts on
```
hlbarber, line 248: > A HTTP plugin is a plugin

I think this is misleading and why perhaps we should call it `HttpMarker` or something. This description would make sense if it was a super trait of `Plugin`.

hlbarber (approved):

CHANGELOG.next.toml
```diff
@@ -586,3 +586,9 @@ message = "The naming `make_token` for fields and the API of `IdempotencyTokenPr
 references = ["smithy-rs#2783"]
 meta = { "breaking" = true, "tada" = false, "bug" = false }
 author = "ysaito1001"
+
+[[smithy-rs]]
+message = "Server `Plugin`s now need to be marked as a `HttpPlugin`, a `ModelPlugin`, or both. Check out the `plugin` module documentation for more information."
```
hlbarber, line 591: I've got a huge changelog entry about the type refactors above - perhaps it'd be nice to squash this one into that and uphold its narrative?

rust-runtime/aws-smithy-http-server/src/plugin/layer.rs
david-perez: It seems a bit heavy-handed to force them to use a collection of plugins to make use of a _single_ plugin that they construct via a layer.

I think it's ok we go ahead with this. I can't think of valid reasons for why someone might want to do a plugin -> layer -> plugin conversion.

rust-runtime/aws-smithy-http-server/src/plugin/mod.rs
david-perez: Why is it misleading?

CHANGELOG.next.toml
david-perez: Addressed in https://github.com/awslabs/smithy-rs/pull/2827/commits/a58ec4897e4a06fc46f28ea3594dd59ecb8d428a.

rust-runtime/aws-smithy-http-server/src/plugin/mod.rs
david-perez: I see. Ok, will change.

PR #2830, opened 2023-07-06
Add `Metadata` to a config layer through operation runtime plugin

ysaito1001:
## Motivation and Context
Adds `Metadata` to a config layer through `OperationRuntimePlugin`.

## Description
We have had a customer's use case where a service name and a operation name are obtained from [Metadata](https://github.com/awslabs/smithy-rs/blob/ddba46086a9754c01f3b11d7521c49d4489de84b/rust-runtime/aws-smithy-http/src/operation.rs#L17-L22). The end goal was to use their names as part of metrics collection. Previously, it was done using `map_operation` on a `CustomizableOperation`, e.g.
```
client
    .some_operation()
    .customize()
    .await?
    .map_operation(|operation| {
        operation.try_clone().map(|operation| {
            let (_, parts) = operation.into_request_response();
            parts.metadata.map(|metadata| {
                let service_name = metadata.service().to_string().to_uppercase();
                let operation_name = metadata.name().to_string();
                /*
                 * do something with `service_name` and `operation_name`
                 */
            })
        });

        Ok(operation)
    })?
    .send()
    .await;
```
The orchestrator no longer supports `map_operation` on `CustomizableOperation`. We therefore add `Metadata` to a config layer through `OperationRuntimePlugin`. See the added integration test for how `Metadata` is retrieved from within an interceptor.

## Testing
Added an integration-test to verify `Metadata` is properly set.

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/MetadataCustomization.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    override fun section(section: OperationSection): Writable = writable {
+        when (section) {
+            is OperationSection.AdditionalRuntimePluginConfig -> {
+                rustTemplate(
+                    """
+                    ${section.newLayerName}.store_put(#{Metadata}::new(
+                        ${operationName.dq()}.to_owned(),
+                        ${codegenContext.serviceShape.sdkId().dq()}.to_owned(),
```
jdisanti, line 38: The `.to_owned()` shouldn't be necessary since the underlying type is `Cow<'static, str>`.

aws/sdk/integration-tests/s3/tests/customizable-operation.rs
```diff
@@ -57,3 +57,59 @@ async fn test_s3_ops_are_customizable() {
         snapshot_signature
     );
 }
+
+#[cfg(aws_sdk_orchestrator_mode)]
+#[tokio::test]
+async fn test_extract_metadata_via_customizable_operation() {
```
jdisanti, line 63: This should be a Kotlin unit test for `MetadataCustomization`.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/MetadataCustomization.kt
ysaito1001: Thanks for catching. Addressed in d4c7bbc.

aws/sdk/integration-tests/s3/tests/customizable-operation.rs
ysaito1001: Thanks for bringing this up! I was actually on the fence about this. I originally wrote this test as a Kotlin test but then thought that writing a test fully in a Rust source file could better serve as documentation for the customers who needs to upgrade their code, instead of Rust snippets embedded in a Kotlin file. But I suppose they have no trouble following the Rust code even in a Kotlin unit test and more importantly, from an engineering process point of view, having a Kotlin unit test is better than having it in `aws/sdk/integration-tests`.

ysaito1001: Addressed in 9472734.

jdisanti (approved):

Velfi (approved):

PR #2842, opened 2023-07-13
Fix panic from repeat clones of `TypeErasedBox`

jdisanti:
This PR fixes a panic that would result from cloning a cloneable `TypeErasedBox` twice.

----


ysaito1001 (approved):
Thanks for the fix!

PR #2843, opened 2023-07-13
Don't allow `test-util` feature in compile time dependencies

jdisanti:
During the orchestrator implementation, a bug was introduced that always enabled the `test-util` feature on `aws-smithy-runtime` in the generated crates. This led to several pieces of non-test code relying on test-only code, specifically around `SystemTime` implementing `TimeSource`.

This PR fixes that issue, and also fixes another issue where the `RuntimeComponentsBuilder` was being initialized without a name for the service config.

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
```diff
@@ -374,9 +375,10 @@ class ServiceConfigGenerator(
         }
 
         writer.docs("Builder for creating a `Config`.")
-        writer.raw("#[derive(Clone, Default)]")
-        if (runtimeMode.defaultToOrchestrator) {
-            Attribute(Attribute.derive(RuntimeType.Debug)).render(writer)
+        if (runtimeMode.defaultToMiddleware) {
+            writer.raw("#[derive(Clone, Default)]")
+        } else {
+            writer.raw("#[derive(Clone, Debug)]")
```
Velfi, line 381: Should we be fully qualifying these traits?

Velfi (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
jdisanti: That's probably the right call, but I'm going to leave it as is for now.

Velfi (approved):

Issue #2413, opened 2023-02-23
Remove 3rd-party types from public APIs


ysaito1001:
This PR contains a sub list from https://github.com/awslabs/smithy-rs/issues/1193. Types in this list will probably not go away as a result of future code reorganization/refactoring, hence working on this task won't be a wasted effort.

- [x] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/aws/rust-runtime/aws-http/external-types.toml#L9
- [x] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/aws/sdk/sdk-external-types.toml#L20
- [x] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/rust-runtime/aws-smithy-async/external-types.toml#L5
- [ ] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/rust-runtime/aws-smithy-client/external-types.toml#L13
- [ ] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/rust-runtime/aws-smithy-http/external-types.toml#L23
- [ ] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/rust-runtime/aws-smithy-http/external-types.toml#L26
- [x] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/rust-runtime/aws-smithy-http/external-types.toml#L32
- [x] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/rust-runtime/aws-smithy-http/external-types.toml#L38
- [x] https://github.com/awslabs/smithy-rs/blob/0da92a4319ceef8eae87af09a37e168f6f4adc93/rust-runtime/aws-smithy-xml/external-types.toml#L2

PR #2845, opened 2023-07-13
Remove third party types from public APIs

ysaito1001:
## Motivation and Context
Addresses item 1, 3, and 9 in #2413 

## Description
This PR removes the third party types as follows:
- removes `InvalidHeaderValue` from public API in `aws-http`
- removes `SendError` from public API in `aws-smithy-async`
- removes `xmlparser` from public API in `aws-smithy-xml`

Those types were exposed to public APIs primarily due to the implementation of traits, e.g. `From` and `Iterator`. Those implementations are used internally (such as XML deserialization and converting an error type) so we should be able to hide them within crates.

## Testing
- [x] Passed tests in CI

## Checklist

----


Velfi (approved):

rust-runtime/aws-smithy-async/src/future/rendezvous.rs
```diff
@@ -38,6 +37,35 @@ pub fn channel<T>() -> (Sender<T>, Receiver<T>) {
...
 }
 
+/// Errors for rendezvous channel
+pub mod error {
+    use std::fmt;
+
+    /// Error when [crate::future::rendezvous::Sender] fails to send a value to the associated `Receiver`
+    #[derive(Debug)]
+    pub struct SendError<T> {
+        source: tokio::sync::mpsc::error::SendError<T>,
```
jdisanti, line 47: Nit: import things when not in codegen =)

rust-runtime/aws-smithy-xml/src/decode.rs
```diff
@@ -256,6 +254,13 @@ impl<'inp> Document<'inp> {
     }
 }
 
+/// A new-type wrapper around `Token` to prevent the wrapped third party type from showing up in
+/// public API
+#[derive(Debug)]
+pub struct XmlToken<'inp> {
+    token: Token<'inp>,
+}
```
jdisanti, line 262: Nit: might be slightly easier to read as `XmlToken<'inp>(Token<'inp>)`

jdisanti (approved):
This is great! Thanks.

rust-runtime/aws-smithy-async/src/future/rendezvous.rs
ysaito1001: Used an alias to bind the import to `TokioSendError` in 59ffe31.

rust-runtime/aws-smithy-xml/src/decode.rs
ysaito1001: Agreed. Addressed in 3105f8f.

PR #2847, opened 2023-07-14
Change the default runtime mode to orchestrator

jdisanti:
This PR changes the default runtime mode to orchestrator for generated clients and the AWS SDK.

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientRustSettings.kt
```diff
@@ -125,6 +125,8 @@ data class ClientCodegenConfig(
     val eventStreamAllowList: Set<String> = defaultEventStreamAllowList,
     // TODO(SmithyRuntime): Remove this once we commit to switch to aws-smithy-runtime and aws-smithy-runtime-api
     val enableNewSmithyRuntime: SmithyRuntimeMode = defaultEnableNewSmithyRuntime,
+    /** If true, adds `endpoint_url`/`set_endpoint_url` methods to the service config */
```
ysaito1001, line 128: Just curious, is there a case where we want to disable those methods?

ysaito1001 (approved):
🚀

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientRustSettings.kt
jdisanti: Yeah, the SDK replaces those methods with alternate implementations that are slightly more sophisticated.

82marbag (approved):

PR #2848, opened 2023-07-14
Remove third party types from public APIs Part 2

ysaito1001:
## Motivation and Context
Addresses item 8 in #2413 

## Description
This PR removes the third party types as follows:
- removes `SegmentedBuf` from public API in `aws-smithy-http` (used when the feature `event-stream` was enabled)

## Testing
- [x] Passed tests in CI

## Checklist

----


jdisanti (approved):
🚀

PR #2851, opened 2023-07-16
Add RUSTFLAG="--cfg aws_sdk_unstable" to ci-scripts

thomas-k-cameron:
## Motivation and Context
This is a sub-PR of https://github.com/awslabs/smithy-rs/pull/2615.
It adds RUSTFLAG="--cfg aws_sdk_unstable" to some ci-scripts to so that it tests with unstable features.
----

### DIFF
--- a/tools/ci-scripts/check-aws-sdk-adhoc-tests
+++ b/tools/ci-scripts/check-aws-sdk-adhoc-tests
@@ -7,6 +7,8 @@
 C_YELLOW='\033[1;33m'
 C_RESET='\033[0m'
 
+export RUSTFLAG="--cfg aws_sdk_unstable"
+
 set -eu
 cd smithy-rs
 
--- a/tools/ci-scripts/check-aws-sdk-orchestrator-impl
+++ b/tools/ci-scripts/check-aws-sdk-orchestrator-impl
@@ -9,6 +9,8 @@
 C_YELLOW='\033[1;33m'
 C_RESET='\033[0m'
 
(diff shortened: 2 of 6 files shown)


rcoh:
I think something like this is necessary. We probably want a single parallel task to sanity check everything with the flag enabled but do most testing with it disabled. Need to think about this. The main issue is the coverage that we're losing with the flag disabled (which is, of course, where most customers are going to be)

thomas-k-cameron:
@rcoh 

I added `test-with-serde` to `ci.yml`. 
https://github.com/smithy-lang/smithy-rs/pull/2851/files

I think it's working.  
https://github.com/smithy-lang/smithy-rs/actions/runs/7279552536?pr=2851

thomas-k-cameron:
I got 2 errors.

1. `:jar: No valid plugin descriptors were found in META-INF/gradle-plugins` from `generate`

and 

2. `Verify client TLS configuration`
```
output: Connecting to https://localhost:40835 with root CA bundle from /tmp/tmpybhd04f_/2: Adding root CA: O=Fake Certificate Authority (65:84:52:d7:3e:03:25:d5)DispatchFailure { source: ConnectorError { kind: Io, source: hyper::Error(Connect, Custom { kind: Other, error: Custom { kind: InvalidData, error: InvalidCertificate(NotValidForName) } }), connection: Unknown } }
```

Since I made no changes to both jobs, I have no idea where it's coming from :(
I don't think my changes are causing them.

thomas-k-cameron:
I think this one's ready for review.

PR #2872, opened 2023-07-26
New-type the interceptor context `Input`, `Output`, and `Error`

jdisanti:
This PR creates new-types for `Input`, `Output`, and `Error` so that the type system can enforce that an output isn't given to an input and vice versa. It also removes `TypedBox` since that isn't really needed anymore.

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
+++ b/rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
@@ -36,13 +36,65 @@ use std::fmt::Debug;
 use std::{fmt, mem};
 use tracing::{debug, error, trace};
 
-// TODO(enableNewSmithyRuntimeLaunch): New-type `Input`/`Output`/`Error`
-/// Type-erased operation input.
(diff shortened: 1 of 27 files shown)


rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs
```diff
@@ -36,13 +36,65 @@ use std::fmt::Debug;
...
 use tracing::{debug, error, trace};
 
-// TODO(enableNewSmithyRuntimeLaunch): New-type `Input`/`Output`/`Error`
-/// Type-erased operation input.
-pub type Input = TypeErasedBox;
-/// Type-erased operation output.
-pub type Output = TypeErasedBox;
-/// Type-erased operation error.
-pub type Error = TypeErasedError;
+macro_rules! new_type_box {
```
ysaito1001, line 39: Love the macro.

ysaito1001 (approved):
Good call on removing `TypedBox`, easier to read the code without it.

PR #2892, opened 2023-08-01
Upgrade Smithy to 1.35

jdisanti:

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolTestGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolTestGenerator.kt
@@ -614,6 +614,20 @@ class DefaultProtocolTestGenerator(
 
         // These tests are not even attempted to be generated, either because they will not compile
         // or because they are flaky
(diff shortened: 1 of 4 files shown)


david-perez:
Server-side restXml support is broken beyond repair, so excluding the newly added `S3EscapeObjectKeyInUriLabel` and `S3EscapePathObjectKeyInUriLabel` tests is fine. I'll remove the `com.amazonaws.s3#AmazonS3` test suite from the server integration tests; in fact, I'm leaning towards removing the server restXml code all together to not give the impression to users that we actually support it.

As for the `RestJsonQueryStringEscaping` test, the test actually passes as-is. I had put in a hotfix while I waited for https://github.com/smithy-lang/smithy/pull/1562 to land, but the fix was upstreamed and we're now consuming it. The test was further tweaked in https://github.com/smithy-lang/smithy/pull/1759 to prepend a space to the query string, which is why my patch no longer worked. I've pushed a commit https://github.com/awslabs/smithy-rs/pull/2892/commits/3dddedfbb73a5adac0c091443dd30995578fdf56 to remove my patch and just use Smithy's definition of the test.

82marbag (approved):

david-perez (approved):

ysaito1001 (approved):

PR #2898, opened 2023-08-02
Add TLS docs page

82marbag:
## Motivation and Context
The TLS connector page is missing

## Description
Render the TLS connector documentation page

## Checklist

----


david-perez (commented):
The book is now checked in CI, you'll need to ensure the examples in https://github.com/awslabs/smithy-rs/blob/main/design/src/transport/connector.md compile. You can prepend lines with `#` in the code so they won't be rendered to the user but you can still rely on their effects in the rest of the example code:

```
# let foo = 69;
dbg!(&foo);
```

david-perez (changes requested):
Adding `ignore` of course works but negates the purpose of having the code checked by CI. Is there any reason in particular why we cannot check this code?

david-perez:
The example code did not compile as-is and it wasn't representative of what a user would write (since they'd deal with the code-generated client and not with the raw `aws_smithy_client`. We also can't depend on `hyper_tls` in the mdBook because there's no crate in the repo depending on the crate (we build the mdBook with `mdbook test design -L target/debug/deps`). So I moved the TLS connector example to the `pokemon-service-tls` example crate in https://github.com/awslabs/smithy-rs/pull/2898/commits/c4e69c91c2738c54a0f0d117d5b3919322d043a3, adding a test to ensure we can call the server with a custom TLS connector based on `native_tls`.

david-perez (approved):

rcoh (approved):

PR #2899, opened 2023-08-02
Add an RFC for limiting the HTTP public dependency

rcoh:
## Motivation and Context
[Rendered](https://github.com/awslabs/smithy-rs/blob/d3eaf5eea561af90146b7aa80da0307d49f63506/design/src/rfcs/rfc0036_http_dep_elimination.md)

## Description
An RFC describing the need and proposed path for limiting the scope of our HTTP dependency.

----

### DIFF
--- a/design/src/rfcs/rfc0036_http_dep_elimination.md
+++ b/design/src/rfcs/rfc0036_http_dep_elimination.md
@@ -1,0 +1,107 @@
+<!-- Give your RFC a descriptive name saying what it would accomplish or what feature it defines -->
+RFC: Eliminating Public `http` dependencies
+=============
+
+> Status: RFC
+>
+> Applies to: client
+
+<!-- A great RFC will include a list of changes at the bottom so that the implementor can be sure they haven't missed anything -->
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+<!-- Insert a short paragraph explaining, at a high level, what this RFC is for -->
+This RFC defines how we plan to refactor the SDK to allow the SDK to consume a `1.0` version of `hyper`, `http-body`,
+and `http` at a later date. Currently, `hyper` is `0.14.x` and a `1.0` release candidate series is in progress. However,
+there are [open questions](https://github.com/orgs/hyperium/projects/1/views/4) that may significantly delay the launch
+of
+these three crates. We do not want to tie the `1.0` of the Rust SDK to these crates.
+
+<!-- The "Terminology" section is optional but is really useful for defining the technical terms you're using in the RFC -->
+Terminology
+-----------
+
+- **http-body**: A crate (and trait) defining how HTTP bodies work. Notably, the change from `0.*` to `1.0` changes `http-body`
+  to operate on frames instead of having separate methods.
+- `http` (crate): a low level crate of `http` primitives (no logic, just requests and responses)
+- ossified dependency: An ossified dependency describes a dependency that, when a new version is released, cannot be utilized without breaking changes. For example, if the `mutate_request` function on every operation operates on `&mut http::Request` where `http = 0.2`, that dependency is "ossified." Compare this to a function that offers the ability to convert something into an `http = 0.2` request—since http=1 and http=0.2 are largely equivalent, the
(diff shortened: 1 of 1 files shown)


design/src/rfcs/rfc0036_http_dep_elimination.md
```diff
@@ -0,0 +1,107 @@
...
+    // this uses the http = 0.2 request. In the future, we can make an internal enum to allow storing an http = 1
+    http_0: http::Request<B>
+}
+```
+
+**Conversion to/from `http::Request`**
+One key property here is that although converting to/from an `http::Request` **can** be expensive, this is *not* ossification of the API. This is because the API can support converting from/to both `http = 0` and `http = 1` in the future—because it offers mutation of the request via a unified interface, the request would only need to be converted once for dispatch if there was a mismatch (instead of repeatedly). At some point in the future, the `http = 0` representation could be deprecated and removed or feature gated.
+
+**Challenges**
+1. Creating an HTTP API which is forwards compatible, idiomatic and "truthful" without relying on existing types from Hyper—e.g. when adding a header, we need to account for the possibility that a header is invalid.
```
jdisanti, line 85: Doesn't the current API just panic if its invalid (at least for `HeaderName`)? This could potentially be an area for improvement by not allowing panicking construction methods to be used. Although, that would probably make us lose out on `const fn` constructors.

design/src/rfcs/rfc0036_http_dep_elimination.md
```diff
@@ -0,0 +1,107 @@
...
+2. Allow for future forwards-compatible evolution in the API—A lot of thought went into the `http` crate API w.r.t method parameters, types, and generics. Although we can aim for a simpler solution in some cases (e.g. accepting `&str` instead of `HeaderName`), we need to be careful that we do so while allowing to API evolution.
```
jdisanti, line 86:
```suggestion
2. Allow for future forwards-compatible evolution in the API—A lot of thought went into the `http` crate API w.r.t method parameters, types, and generics. Although we can aim for a simpler solution in some cases (e.g. accepting `&str` instead of `HeaderName`), we need to be careful that we do so while allowing API evolution.
```

jdisanti (commented):
Nice work! This is a very well thought out RFC.

design/src/rfcs/rfc0036_http_dep_elimination.md
rcoh: `HeaderName::from_static` panics but the other APIs are fallible

design/src/rfcs/rfc0036_http_dep_elimination.md
```diff
@@ -0,0 +1,107 @@
...
+The SigV4 crate signs a number of `HTTP` types directly. We should change it to accept strings, and when appropriate, iterators of strings for headers.
+
+### Removing the HTTP dependency from generated clients
+Generated clients currently include a public HTTP dependency in `customize`. This should be changed to accept our `HTTP` wrapper type instead or be restricted to a subset of operations (e.g. `add_header`) while forcing users to add an interceptor if they need full control.
+
+
+In order to implement this feature, we need to add X and update Y...
+
+<!-- Include a checklist of all the things that need to happen for this RFC's implementation to be considered complete -->
+Changes checklist
```
jdisanti, line 98: Need an item for `~` dependencies?

design/src/rfcs/rfc0036_http_dep_elimination.md
```diff
@@ -0,0 +1,107 @@
...
+-----------------
```
ysaito1001, line 99: We also want an item to remove `http` types from `external-types.xml`s?

ysaito1001 (approved):
Thank you for this excellent write-up!

design/src/rfcs/rfc0036_http_dep_elimination.md
rcoh: we actually can't do that—we _may_ want the ability to have something like `#[allow_external]` somehow for the external types checker

jdisanti: To clarify, the only allowed instances are the ones that are used exclusively from generated code?

rcoh: no, the wrapper type itself will have conversion functions so that e.g. we can use it with Hyper

rcoh: but it should be removed from almost all locations

PR #2900, opened 2023-08-03
Default collections RFC

rcoh:
[Rendered](https://github.com/awslabs/smithy-rs/blob/flatten-collections-rfc/design/src/rfcs/rfc0035_collection_defaults.md)

Could use a bit more color but I think it gets the point across.


jdisanti (approved):
I like it!

PR #2905, opened 2023-08-07
Update valid `Content-Type`s for example Python server

david-perez:
The service has a modeled event stream operation. It is exercised in the
`event_stream_test`.

----

### DIFF
--- a/examples/python/pokemon_service.py
+++ b/examples/python/pokemon_service.py
@@ -176,8 +176,8 @@ Next = Callable[[Request], Awaitable[Response]]
 @app.middleware
 async def check_content_type_header(request: Request, next: Next) -> Response:
     content_type = request.headers.get("content-type")
-    if content_type == "application/json":
-        logging.debug("found valid `application/json` content type")
+    if content_type in ["application/json", "application/vnd.amazon.eventstream"]:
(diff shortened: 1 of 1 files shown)


examples/python/pokemon_service.py
```diff
@@ -176,8 +176,8 @@ def get_random_radio_stream(self) -> str:
 @app.middleware
 async def check_content_type_header(request: Request, next: Next) -> Response:
     content_type = request.headers.get("content-type")
-    if content_type == "application/json":
-        logging.debug("found valid `application/json` content type")
+    if content_type in ["application/json", "application/vnd.amazon.eventstream"]:
+        logging.debug("found valid `%s` content type", content_type)
     else:
         logging.warning(
```
drganjoo, line 182: nit: Should we add a comment here indicating that sensitive information may be logged by dumping all headers and that this should not be done in production?

drganjoo (approved):

examples/python/pokemon_service.py
david-perez: Sorry, I had "add to merge queue after approval" enabled and the PR got merged before I saw this comment.

I've addressed it in https://github.com/awslabs/smithy-rs/pull/3083.

PR #2909, opened 2023-08-08
Make it possible to nest runtime components

jdisanti:
Runtime plugins need to be able to wrap components configured in other runtime components. For example, one runtime plugin should be able to wrap the HTTP connector configured in another runtime plugin.

This PR makes this possible by:
- Introducing the ability to order runtime plugins within the service/operation plugin "levels".
- Adding an argument to `RuntimePlugin::runtime_components` so that implementations can reference components configured by previous plugins.

The `order` function has three separate order values: `Defaults`, `Overrides`, and `NestedComponents`. The `Defaults` order is currently unused, but will be used later when we refactor how defaults in config work. Everything defaults to `Overrides` since most runtime plugins will want to be in this slot. The `NestedComponents` order is specifically for runtime plugins that want to create nested components, and runs at the very end.

----


ysaito1001 (approved):
LGTM!

PR #2910, opened 2023-08-09
Remove `futures_core::stream::Stream` from public API

ysaito1001:
## Motivation and Context
Addresses boxes 2 and 5 in #2413 

## Description
This PR removes the `futures_core::stream::Stream` trait from public API, as enforced by our tool [check-external-types](https://github.com/awslabs/cargo-check-external-types). In our codebase, we previously relied on the `Stream`  trait for the following use cases:
- calling `.next().await` to retrieve items from a stream
- calling `.collect().await` to dump elements into a collection
- passing types to functions/methods accepting `impl Stream`

Removing the `Stream` trait from the use cases above affected the following types:
- `aws_smithy_async::fn_steram::{FnStream, TryFlatMap}`
- `aws_smithy_http::byte_stream::ByteStream`
- `aws_smithy_http::event_stream::sender::{EventStreamSender, MessageStreamAdapter}`

The primary approach taken in this PR involves
- defining `async fn next(&mut self) -> XXX` for types that used to implement the `impl Stream`
- implementing a `.collect` stream extension method for `FnStream` since it was previously used in its example
- using `FnStream` at its core when types needed to hold/pass/return `Stream` trait objects
- using `FnStream` to replace usages of a `stream!` macro from `async_stream`
- introducing new-types `HyperBodyWrapEventStream` and `HyperBodyWrapByteStream` for those that were previously passed to `hyper::body::Body::wrap_stream`, which requires `impl Stream`, but no longer implement the trait.

## Testing
Ported existing tests (many of which used a `stream!` macro from the `async-stream` crate) to use `FnStream`.

## Checklist

----


rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -403,6 +427,29 @@ impl ByteStream {
     }
 }
 
+pin_project! {
+    // A new-type wrapper around `ByteStream` so we can pass it to `tokio_util::io::StreamReader`.
+    struct StreamWrapper {
+        #[pin]
+        byte_stream: ByteStream,
+    }
```
ysaito1001, line 435: This is pretty much the same as `HyperBodyWrapEventStream` in `hyper_body_wrap_stream.rs`. It'd be nice to remove duplication 🤔

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
```diff
@@ -239,12 +309,22 @@ class HttpBoundProtocolPayloadGenerator(
         val ref = if (payloadMetadata.takesOwnership) "" else "&"
         val serializer = protocolFunctions.serializeFn(member, fnNameSuffix = "http_payload") { fnName ->
             val outputT = if (member.isStreaming(model)) {
-                symbolProvider.toSymbol(member)
+                if (fromPythonServerRuntime(member)) {
```
jdisanti, line 312: Rather than having `codegen-core` be aware of the Python server, we should make wrapping the streams a configurable feature of this class. See `renderEventStreamBody` as an example.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -838,15 +856,24 @@ class ServerHttpBoundProtocolTraitImplGenerator(
...
-                httpTrait.uri.segments.slice((greedyLabelIndex + 1) until httpTrait.uri.segments.size).joinToString(prefix = "/", separator = "/")
+                httpTrait.uri.segments.slice((greedyLabelIndex + 1) until httpTrait.uri.segments.size)
+                    .joinToString(prefix = "/", separator = "/")
             } else {
                 ""
             }
         val labeledNames = segments
             .mapIndexed { index, segment ->
-                if (segment.isLabel) { "m$index" } else { "_" }
+                if (segment.isLabel) {
```
jdisanti, line 866: For next time, try to only do code formatting in the areas you're changing, or post a separate PR that only formats files. It makes code review a lot easier.

rust-runtime/aws-smithy-async/src/future/fn_stream.rs
```diff
@@ -39,52 +40,86 @@ pin_project! {
...
             Poll::Pending => {
-                if let Some(generator) = me.generator.as_mut().as_pin_mut() {
-                    if generator.poll(cx).is_ready() {
+                if let Some(generator) = me.generator {
+                    if generator.as_mut().poll(cx).is_ready() {
                         // if the generator returned ready we MUST NOT poll it again—doing so
                         // will cause a panic.
-                        me.generator.set(None);
+                        *me.generator = None;
                     }
```
jdisanti, line 99: I'm a little confused by this. Wouldn't this mean that the stream could only ever yield one item?

rust-runtime/aws-smithy-async/src/future/fn_stream/collect.rs
```diff
@@ -0,0 +1,83 @@
...
+
+/// A trait that signifies that elements can be collected into `T`.
+///
+/// Currently the trait may not be implemented by clients so we can make changes in the future
+/// without breaking code depending on it.
+pub trait Collectable<T>: sealed::CollectablePrivate<T> {}
+
+pub(crate) mod sealed {
+    #[doc(hidden)]
+    pub trait CollectablePrivate<T> {
```
jdisanti, line 20: Why have two traits instead of putting Collectable inside of sealed?

rust-runtime/aws-smithy-async/src/future/fn_stream/collect.rs
```diff
@@ -0,0 +1,83 @@
...
+        type Collection;
+
+        fn initialize() -> Self::Collection;
+
+        fn extend(collection: &mut Self::Collection, item: T) -> bool;
+
+        fn finalize(collection: &mut Self::Collection) -> Self;
```
jdisanti, line 27: Why not just take `self` directly instead of `&mut Self`? It doesn't look like this trait needs object safety.

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -567,6 +608,11 @@ where
     }
 }
 
+const SIZE_HINT_32_BIT_PANIC_MESSAGE: &str = r#"
```
jdisanti, line 611: Can we get rid of this now that we don't have a trait mandating that we use `usize`? Is it possible to use `u64` now?

rust-runtime/aws-smithy-http/src/event_stream/sender.rs
```diff
@@ -4,20 +4,25 @@
...
 use std::task::{Context, Poll};
 use tracing::trace;
 
 /// Input type for Event Streams.
 pub struct EventStreamSender<T, E> {
-    input_stream: Pin<Box<dyn Stream<Item = Result<T, E>> + Send + Sync>>,
+    // `FnStream` does not have a `Sync` bound but this struct needs to be `Sync`
+    // as demonstrated by a unit test `event_stream_sender_send_sync`.
+    // Wrapping `input_stream` with a `Mutex` will make `EventStreamSender` `Sync`.
+    input_stream: Mutex<FnStream<Result<T, E>>>,
```
jdisanti, line 25: Was there a reason that the boxed future inside FnStream doesn't have a Sync bound? Is it just to make implementing the stream easier?

tools/ci-cdk/canary-runner/src/build_bundle.rs
```diff
@@ -56,16 +56,19 @@ tracing-texray = "0.1.1"
 reqwest = { version = "0.11.14", features = ["rustls-tls"], default-features = false }
 "#;
 
+const REQUIRED_SMITHY_CRATES: &[&str] = &["aws-smithy-async"];
```
jdisanti, line 59: If `FnStream` is required to implement event streams now, then we need to re-export it from the generated crates (when generating for event streams).

jdisanti (commented):
Great work on getting this working!

rust-runtime/aws-smithy-async/src/future/fn_stream.rs
ysaito1001: It comes from this bug fix https://github.com/awslabs/smithy-rs/pull/1903, and the stream yields more than one item (otherwise, the unit tests would fail).

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
ysaito1001: Thanks for pointing this out. Avoiding `codegen-core` being aware of the Python server is crucial from a design point of view. Addressed in bedfb04.

The key ideas in the above commit are
- `HttpBoundProtocolPayloadGenerator` now takes additional parameter `StreamPayloadSerializer` containing two callbacks. This parameter allows client/server to pass in how they want to specify the type of serialized payload stream and how they want to wrap (or not wrap) in a new-type wrapper enabling `futures_core::stream::Stream` trait.
- For `StreamPayloadSerializer`, the server side needs to further distinguish between regular server-codegen and python server-codegen. We use `ServerHttpBoundProtocolCustomization` to make that happen; `ServerProtocolLoader.kt` and `PythonServerProtocolLoader.kt` each defines the appropriate customization.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
ysaito1001: Enabled the option for Reformat only changes uncommitted in my IDE. Sorry about that.

rust-runtime/aws-smithy-async/src/future/fn_stream/collect.rs
ysaito1001: As discussed offline `extend` can keep `&mut Self::Collection` for the convenience of the call sites. We can let `finalize` take an owned `Self::Collection` so we can get rid of `std::mem::replace` in the implementation of `finalize` in a consuming type.

Addressed in 5698ee3.

tools/ci-cdk/canary-runner/src/build_bundle.rs
ysaito1001: Good point. Canary shouldn't need to directly depend on `aws-smithy-async` that way. Addressed in b281aee.

rust-runtime/aws-smithy-async/src/future/fn_stream.rs
ysaito1001: Updated comment with more details in a778886.

rust-runtime/aws-smithy-http/src/byte_stream.rs
ysaito1001: I think we can. For our own `ByteStream::size_hint` we can just directly return what's returned by the underlying `http_body::Body::size_hint`. Addressed in 0f5f6be.

rust-runtime/aws-smithy-http/src/event_stream/sender.rs
ysaito1001: Yes, the underlying `generator` is a future that is only `Send`, which is usually the case for `Future` ([discussion](https://users.rust-lang.org/t/future-is-not-sync-when-calling-async-fn-of-dyn-trait-obj-in-another-async-fn/83285/2)). Added a comment in d2e4411.

rust-runtime/aws-smithy-async/src/future/fn_stream.rs
jdisanti: Thanks. That makes a lot more sense now.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
```diff
@@ -53,11 +54,24 @@ data class EventStreamBodyParams(
...
+    val symbolProvider: SymbolProvider,
+    val runtimeConfig: RuntimeConfig,
+    val member: MemberShape,
+    val payloadName: String?,
+)
+
+data class StreamPayloadSerializer(
+    val outputT: (RustWriter, StreamPayloadSerializerParams) -> Unit,
+    val renderPayload: (RustWriter, StreamPayloadSerializerParams) -> Unit,
+)
```
jdisanti, line 67: Seems like this should be an interface rather than a data class.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -87,6 +89,12 @@ import java.util.logging.Logger
 sealed class ServerHttpBoundProtocolSection(name: String) : Section(name) {
     data class AfterTimestampDeserializedMember(val shape: MemberShape) :
         ServerHttpBoundProtocolSection("AfterTimestampDeserializedMember")
+
+    data class TypeOfSerializedPayloadStream(val params: StreamPayloadSerializerParams) :
+        ServerHttpBoundProtocolSection("TypeOfSerializedPayloadStream")
+
+    data class WrapStreamAfterPayloadGenerated(val params: StreamPayloadSerializerParams) :
+        ServerHttpBoundProtocolSection("WrapStreamAfterPayloadGenerated")
```
jdisanti, line 97: Add comments explaining these and how to implement them.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
```diff
@@ -26,6 +26,14 @@ private fun hasStreamingOperations(model: Model): Boolean {
     }
 }
 
+/** Returns true if the model has event streaming operations */
+private fun hasEventStreamOperations(model: Model): Boolean =
```
jdisanti, line 30: Can you use https://github.com/awslabs/smithy-rs/blob/847ed0b66afe6eca3c64a764a4e7c7c16b0851df/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt#L93? The codegen context has the service shape in it.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/HttpBoundProtocolPayloadGenerator.kt
ysaito1001: Now that I'm looking at it, yes totally. Addressed in ab57dce.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
ysaito1001: Addressed in 3db0525.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
ysaito1001: Addressed in 91163de.

jdisanti (approved):
LGTM

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerEventStreamWrapperGenerator.kt
```diff
@@ -166,6 +169,14 @@ class PythonServerEventStreamWrapperGenerator(
                         })
                     });
 
+                    let stream = #{FnStream}::new(|tx| {
+                        Box::pin(async move {
+                            while let #{Some}(item) = stream.next().await {
+                                tx.send(item).await.expect("send should succeed");
```
unexge, line 175: Is it safe to panic here?

rust-runtime/inlineable/src/hyper_body_wrap_stream.rs
```diff
@@ -0,0 +1,133 @@
...
+use aws_smithy_http::byte_stream::ByteStream;
+use aws_smithy_http::event_stream::MessageStreamAdapter;
+use aws_smithy_http::result::SdkError;
+use bytes::Bytes;
+use futures_core::stream::Stream;
+use std::error::Error as StdError;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+pub(crate) struct HyperBodyWrapEventStream<T, E>(MessageStreamAdapter<T, E>);
```
david-perez, line 17: nit: I'd prefer it if this were not an inlineable and instead lived in a runtime crate, `#[doc(hidden)]`. We can add the docs we've sprinkled in the code generator as proper Rust docs, contributors have a better development experience, and each user won't have to pay the cost of running the below tests when compiling their generated SDK.

rust-runtime/aws-smithy-async/src/future/fn_stream.rs
```diff
@@ -6,12 +6,14 @@
 //! Utility to drive a stream with an async function and a channel.
 
 use crate::future::rendezvous;
-use futures_util::StreamExt;
```
david-perez, line 9: nit: I guess we could mention in the changelog that the type now doesn't implement this or `TryStream`, `TryStreamExt` either, for completion.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/protocols/PythonServerProtocolLoader.kt
```diff
@@ -61,6 +61,29 @@ class PythonServerAfterDeserializedMemberServerHttpBoundCustomization() :
...
+
+/**
+ * Customization class used to determine the type of serialized stream payload and how it should be wrapped in a
+ * new-type wrapper to enable `futures_core::stream::Stream` trait.
+ */
+class PythonServerStreamPayloadSerializerCustomization() : ServerHttpBoundProtocolCustomization() {
+    override fun section(section: ServerHttpBoundProtocolSection): Writable = when (section) {
+        is ServerHttpBoundProtocolSection.TypeOfSerializedStreamPayload -> writable {
+            // `aws_smithy_http_server_python::types::ByteStream` already implements
+            // `futures::stream::Stream`, so no need to wrap it in a futures' stream-compatible
```
david-perez, line 77: Why the asymmetry though? I'd prefer it if we made `aws_smithy_http_server_python::types::ByteStream` more similar to `aws_smithy_http::byte_stream::ByteStream` and have both not implement `Stream`, and wrap it there too in a `HyperBodyWrapByteStream`.

david-perez: The generator code would also become simpler; we wouldn't need to draw distinctions based on what we're generating.

rust-runtime/aws-smithy-http/src/event_stream/sender.rs
david-perez: Why does `EventStreamSender` need to be `Sync`? i.e. why does that test exist?

examples/pokemon-service-common/src/lib.rs
```diff
@@ -242,59 +242,63 @@ pub async fn capture_pokemon(
...
+                                    .unwrap_or_else(|| CapturingPayload::builder().build());
+                                let pokeball = payload.pokeball().unwrap_or("");
+                                if !matches!(pokeball, "Master Ball" | "Great Ball" | "Fast Ball") {
+                                    tx.send(Err(
                                     crate::error::CapturePokemonEventsError::InvalidPokeballError(
                                         crate::error::InvalidPokeballError {
                                             pokeball: pokeball.to_owned()
                                         }
                                     )
-                                );
```
david-perez, line 262: Related to Burak's comment, I'm having trouble understanding when the receiver can disconnect and this can panic. Can we document that somewhere?

* If the receiver can indeed drop, we shouldn't have examples where we panic, and instead teach users in the example how best to handle the disconnection.
* If the receiver can't really drop, we shouldn't expose a fallible interface.

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -48,7 +48,8 @@
 //!
 //! ### Stream a ByteStream into a file
 //! The previous example is recommended in cases where loading the entire file into memory first is desirable. For extremely large
-//! files, you may wish to stream the data directly to the file system, chunk by chunk. This is posible using the `futures::Stream` implementation.
+//! files, you may wish to stream the data directly to the file system, chunk by chunk.
+//! This is possible using the [`.next()`](crate::byte_stream::ByteStream::next).
```
david-perez, line 52:
```suggestion
//! This is possible using the [`.next()`](crate::byte_stream::ByteStream::next) method.
```

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/protocols/PythonServerProtocolLoader.kt
ysaito1001: One of the goals for the task is to keep things untouched, if possible, which is the primary reason `aws_smithy_http_server_python::types::ByteStream` still implements `Stream`.

That being said, if the server team thinks we can get rid of `impl Stream` for `aws_smithy_http_server_python::types::ByteStream`, I'm for it since the code gen will be simpler that way, i.e. no need to have special customization like `PythonServerStreamPayloadSerializerCustomization`.

ysaito1001: I tried keeping the server side code untouched as much as possible, thus leaving python's `ByteStream` still implementing `Stream`. That said, if the server team feels we can get rid of `impl Stream` for `aws_smithy_http_server_python::types::ByteStream`, I'm for it. That'll obviate the need for customizations like `PythonServerStreamPayloadSerializerCustomization`.

rust-runtime/inlineable/src/hyper_body_wrap_stream.rs
ysaito1001: Having `HyperBodyWrapEventStream` in a runtime crate might help solve https://github.com/awslabs/smithy-rs/pull/2910#discussion_r1306191209 as well. I think it's worth incorporating this as part of the PR.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/protocols/PythonServerProtocolLoader.kt
ysaito1001: The asymmetry comes from the fact that `HyperBodyWrapByteStream` is only capable of wrapping `aws_smithy_http::byte_stream::ByteStream`; even if we dropped `impl Stream for aws_smithy_http_server_python::types::ByteStream`, it could not be wrapped in `HyperBodyWrapByteStream`.

For this reason, if something already implements the `Stream` trait, it's better to leave it as-is since we don't need to wrap it in a new-type to enable the trait.

aws/sdk/integration-tests/transcribestreaming/tests/test.rs
```diff
@@ -16,20 +15,25 @@ use aws_sdk_transcribestreaming::{Client, Config};
...
 
 #[tokio::test]
 async fn test_success() {
-    let input_stream = stream! {
-        let pcm = pcm_data();
-        for chunk in pcm.chunks(CHUNK_SIZE) {
-            yield Ok(AudioStream::AudioEvent(AudioEvent::builder().audio_chunk(Blob::new(chunk)).build()));
-        }
-    };
+    let input_stream = FnStream::new(|tx| {
```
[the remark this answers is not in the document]
rcoh: FnStream is an implementation of generators–the network isn't involved in all use cases. By bounding it, you all the caller to avoid using more memory than is necessary

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/PaginatorGenerator.kt
```diff
@@ -141,8 +141,9 @@ class PaginatorGenerator private constructor(
 
                 /// Create the pagination stream
                 ///
-                /// _Note:_ No requests will be dispatched until the stream is used (eg. with [`.next().await`](tokio_stream::StreamExt::next)).
-                pub fn send(self) -> impl #{Stream}<Item = #{item_type}> + #{Unpin} {
+                /// _Note:_ No requests will be dispatched until the stream is used
+                /// (e.g. with [`.next().await`](aws_smithy_async::future::fn_stream::FnStream::next)).
+                pub fn send(self) -> #{fn_stream}::FnStream<#{item_type}> {
```
rcoh, line 146: hmmm...not sure I like this. It's pretty constraining and also exposes a weird name to customers. I think we should wrap this in a newtype so we aren't publicly exposing the `FnStream` internals

rcoh (changes requested):
need more discussion here about how we will do this with paginators. I think we should probably split out some smaller PRs—I don't think `FnStream` is the interface we want to be generally exposing to people—it's not really use friendly and has a confusing name.

When stream is used as an input type (e.g. transcribe streaming) I think we may want to attempt to retain some amount of compatibility with the `Stream` trait, possibly via the smithy conversions / compatibility crate (or our own trait?)

I don't think that making it a concrete struct is really the right option here.

Can we make this PR much smaller to start piecing off individual chunks?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/PaginatorGenerator.kt
ysaito1001: > Can we make this PR much smaller to start piecing off individual chunks?

We can create a smaller PR just to remove `impl Stream` from `aws-smithy-async` and but keep it for event streaming (that'll take care of Burak & David's questions around `FnStream` since the code around event stream will be unchanged).

The above `send` methods needs to return a new-type hiding `FnStream`.

rust-runtime/aws-smithy-http/src/byte_stream.rs
ysaito1001: Moving `hyper_body_wrap_stream.rs` to a doc-hidden module in `aws-smithy-http` helped remove this one-off struct. See f5c8a96.

rust-runtime/inlineable/src/hyper_body_wrap_stream.rs
ysaito1001: In addition to what you pointed out, moving this to a runtime crate also helped address https://github.com/awslabs/smithy-rs/pull/2910#discussion_r1306191209. Thanks for the suggestion. Addressed in f5c8a96.

rust-runtime/inlineable/src/hyper_body_wrap_stream.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl<T, E> HyperBodyWrapEventStream<T, E> {
+    #[allow(dead_code)]
+    pub(crate) fn new(adapter: MessageStreamAdapter<T, E>) -> Self {
+        Self(adapter)
+    }
+}
+
+impl<T, E: StdError + Send + Sync + 'static> Unpin for HyperBodyWrapEventStream<T, E> {}
```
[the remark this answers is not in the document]
ysaito1001: It actually wasn't and removed in f5c8a96.

ysaito1001:
Closing this PR for the reason https://github.com/awslabs/smithy-rs/pull/2910#discussion_r1319214680.

PR #2912, opened 2023-08-09
HTTP Wrapper Type RFC & Implementation

rcoh:
## Motivation and Context
Implementation of the HTTP request wrapper type from RFC

## Description
RFC + implementation of HttpRequest

## Testing
- IT/UT
- Separate PR which uses this type in implementation of the SDK

----


rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -0,0 +1,237 @@
...
+use crate::box_error::BoxError;
+use crate::client::interceptors::context::Error;
+use aws_smithy_http::body::SdkBody;
+use http as http0;
+use std::borrow::Cow;
+use std::fmt::{Debug, Formatter};
+use std::str::FromStr;
+
+/// An HTTP Request Type
+pub struct Request<B = SdkBody> {
```
jdisanti, line 15: Do we want the body to be generic, even?

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -0,0 +1,237 @@
...
+        write!(f, "{:?}", self.inner)
+    }
+}
+
+impl<B> Request<B> {
+    /// Converts this request into an http 0.x request.
+    ///
+    /// Depending on the internal storage type, this operation may be free or it may have an internal
+    /// cost.
+    pub fn into_http0x(self) -> http::Request<B> {
```
jdisanti, line 30: Hypothetically, if there was ever a http 0.3 release before 1.0, this would need an additional method that would break the naming convention, right?

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -0,0 +1,237 @@
...
+        Headers::from(&self.inner)
+    }
+
+    /// Returns a mutable reference to the header map
+    pub fn headers_mut(&mut self) -> HeadersMut<'_> {
+        HeadersMut::from(&mut self.inner)
+    }
+
+    /// Returns the body associated with the request
+    pub fn body(&self) -> &B {
```
jdisanti, line 45: Would need `body_mut` too.

Also, it would be super nice to have a conveniences like `map_body`/`take_body` since we frequently have to juggle `mem::swap`s in the orchestrator code to accomplish that.

jdisanti (commented):
This looks great so far!

ysaito1001 (commented):
Looks good so far! I originally had a question whether
```
impl TryFrom<http0::HeaderValue> for HeaderValue
```
is considered exposing a 3rd party type in public API, but this should be ok as it's done in a way that is not ossified, as called out in the RFC.

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
rcoh: yeah, we should either make this a From impl or name it more precisely

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
rcoh: yeah there are a bunch of methods that need to be added to support everything

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
rcoh: I think so—really no reason _not_ to have it be generic.

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
jdisanti: Should it be fallible in case it's not possible to convert in the future (due to losing information or something)?

design/src/rfcs/rfc0037_http_wrapper.md
```diff
@@ -0,0 +1,96 @@
...
+3. We will add SDK specific helper functionality when appropriate, e.g. first-level support for applying an endpoint to a request.
+
+<!-- Explain the implementation of this new feature -->
+How to actually implement this RFC
+----------------------------------
+
+We will need to add two types, `HttpRequest` and `HttpResponse`.
+
+#### To string or not to String
+The `http` library is very precise in its representation—specifically it allows for `HeaderValue`s that are both a super and subset of `String`—a superset because headers support arbitrary binary data but a subset because headers can contain any sequence of non-zero codepoints including sequences that are not valid UTF-8.
```
jdisanti, line 40: I don't think this section actually answers the question of String or not String

design/src/rfcs/rfc0037_http_wrapper.md
```diff
@@ -0,0 +1,96 @@
...
+These types will be used by all orchestrator functionality, so they will be housed in `aws-smithy-runtime-api`
+
+#### What's in and what's out?
+At the onset, these types focus on supporting the most ossified usages: `&mut` modification of HTTP types. They **do not**
+support construction of HTTP types, other than `impl From<http::Request>` and `From<http::Response>`. We will also make it
+possible to use `http::HeaderName` / `http::HeaderValue` in a zero-cost way.
+
+#### The `AsHeaderComponent` trait
+All header insertion methods accept `impl AsHeaderComponent`. This allows us to provide a nice user experience while taking
+advantage of zero-cost usage of `'static str`. We will seal this trait to prevent external usage.
```
jdisanti, line 52: Can you explain this more? How does the trait let us take advantage of `&'static str`?

rust-runtime/aws-smithy-runtime-api/src/client/http/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    fn invalid_header_value(err: impl Into<BoxError>) -> Self {
+        // TODO: better error internals
+        Self(err.into())
+    }
+}
+
+impl Display for HttpError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "An error occurred creating an HTTP Request")
```
jdisanti, line 39:
```suggestion
        write!(f, "an error occurred creating an HTTP Request")
```

rust-runtime/aws-smithy-runtime-api/src/client/http/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+impl Deref for HttpError {
+    type Target = dyn StdError + Send + Sync + 'static;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_ref()
+    }
+}
```
jdisanti, line 49: Why `Deref` instead of `std::error::Error` with a `source()` impl?

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -0,0 +1,476 @@
...
+use http as http0;
+use http0::header::Iter;
+use http0::{Extensions, HeaderMap, Method};
+use std::borrow::Cow;
+use std::fmt::Debug;
+use std::str::FromStr;
+
+#[derive(Debug)]
+/// An HTTP Request Type
+pub struct Request<B = SdkBody> {
```
jdisanti, line 17: Have you tried using the Request/Response with Surf? Does it play nicely with that API?

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -0,0 +1,476 @@
...
+    body: B,
+    uri: Uri,
+    method: Method,
+    extensions: Extensions,
+    headers: Headers,
+}
+
+/// A Request URI
+#[derive(Debug, Clone)]
+pub struct Uri {
```
jdisanti, line 27: Would love to have nice URI manipulations on this in the future 😄

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -0,0 +1,476 @@
...
+    where
+        U: TryInto<Uri>,
+    {
+        let uri = uri.try_into().map_err(|e| e.into())?;
+        self.uri = uri;
+        Ok(())
+    }
+
+    /// Adds an extension to the request extensions
+    pub fn add_extension<T: Send + Sync + Clone + 'static>(&mut self, extension: T) {
```
jdisanti, line 136: Should this be fallible in case the underlying HTTP library doesn't support extensions?

design/src/rfcs/rfc0037_http_wrapper.md
```diff
@@ -0,0 +1,96 @@
...
+2. Support for `&mut self.apply_endpoint(...)`
+
+#### Handling failure
+There is no stdlib type that cleanly defines what may be placed into headers—String is too broad (even if we restrict to ASCII). This RFC proposes moving fallibility to the APIs:
+```rust,ignore
+impl HeadersMut<'_> {
+    pub fn try_insert(
+        &mut self,
+        key: impl AsHeaderName,
+        value: impl AsHeaderName,
```
jdisanti, line 67:
```suggestion
        key: impl AsHeaderComponent,
        value: impl AsHeaderComponent,
```

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -0,0 +1,476 @@
...
+}
+
+use sealed::AsHeaderComponent;
+
+mod sealed {
+    use super::*;
+    /// Trait defining things that may be converted into a header component (name or value)
+    pub trait AsHeaderComponent {
+        fn into_maybe_static(self) -> Result<MaybeStatic, HttpError>;
+        fn already_header_name(self) -> Result<http0::HeaderName, Self>
```
jdisanti, line 309:
```suggestion
        fn already_header_name_http_03x(self) -> Result<http0::HeaderName, Self>
```
?

Or, since we're always dealing with concrete objects instead of dyn traits, I think we could even make this generic over the return type so the caller can request whatever concrete type they want. Although, implementing that may be quite gross...

I guess this isn't too much of a concerns since this is a private trait though.

rcoh: changed the name to `try_as_http03_header_name`

design/src/rfcs/rfc0037_http_wrapper.md
```diff
@@ -0,0 +1,108 @@
...
+<details>
+<summary>Proposed Implementation of `request`</summary>
+
+```rust,ignore
+{{#include ../../../rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs}}
+```
+</details>
+
+### Future Work
+Currently, the only way to construct `Request` is from a compatyible type (e.g. `http03x::Request`)
```
jdisanti, line 100:
```suggestion
Currently, the only way to construct `Request` is from a compatible type (e.g. `http03x::Request`)
```

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
rcoh: so _currently_ we have fallibility in the conversion to the final request type—you actually don't know what the final request type is going to be until the request is dispatched.

We can probably refactor at some point to remove the need to set request extensions and instead handle it during the conversion to `http03x::Request`...not a bad idea actually, hmmm

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
rcoh: the API works fine _except_ for extensions. We'd need to maintain a copy of each extension map we aim to target or the maps would need to support construction from TypeId directly

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -0,0 +1,613 @@
...
+            .body(SdkBody::from("hello"))
+            .unwrap();
+        let mut req = super::Request::try_from(req).unwrap();
+        assert_eq!(req.uri(), "http://foo.com/");
+        req.set_uri("http://bar.com").unwrap();
+        assert_eq!(req.uri(), "http://bar.com");
+        let http0 = req.into_http03x().unwrap();
+        assert_eq!(http0.uri(), "http://bar.com");
+    }
+
```
jdisanti, line 578: Just note somewhere that there are more tests to add. For example, the `Uri::set_endpoint` method.

jdisanti (approved):

PR #2913, opened 2023-08-09
Use `Identity` instead of `Credentials` in signing code

Velfi:
This PR replaces the access_key, secret_key, and session token fields of signing params with the Orchestrator's `Identity` type.

## Checklist

----


aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
```diff
@@ -736,7 +747,8 @@ mod tests {
             .body("")
             .unwrap();
         let req = SignableRequest::from(&req);
-        let signing_params = signing_params(SigningSettings::default());
+        let identity = Identity::new(Credentials::for_tests(), None);
```
rcoh, line 750: lets add `impl Into<Identity>` for credentials. That way you don't have to have this constructor everywhere + it will handle setting the expiry properly.

rcoh: signing params can then accept `impl Into<Identity>`

aws/rust-runtime/aws-sigv4/src/lib.rs
```diff
@@ -59,14 +56,17 @@ impl<'a, S> SigningParams<'a, S> {
     pub fn service_name(&self) -> &str {
         self.service_name
     }
+
+    /// If the identity in params contains AWS credentials, return them. Otherwise, return `None`.
+    pub fn credentials(&self) -> Option<&Credentials> {
+        self.identity.data::<Credentials>()
+    }
```
rcoh, line 63: do we need this (or any!) public functions on signing params?

aws/rust-runtime/aws-sigv4/src/lib.rs
```diff
@@ -59,14 +56,17 @@ impl<'a, S> SigningParams<'a, S> {
...
 }
 
 impl<'a, S: fmt::Debug> fmt::Debug for SigningParams<'a, S> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("SigningParams")
-            .field("access_key", &"** redacted **")
-            .field("secret_key", &"** redacted **")
-            .field("security_token", &"** redacted **")
+            .field("identity", &"** redacted **")
```
rcoh, line 69: might as well delegate to `Identity` and trust its redaction? I guess this is safer but it might be helpful to, e.g. see the expiration time. And if we can't trust the debug impl on identity we have bigger issues

rcoh: hmm and then we don't need to impl debug anymore at all actually

rcoh (commented):
overall I think this is the right change. Personally, though, I'd keep `Credentials` on all the private locations and validate very early on that it is indeed credentials inside. We can also have a `.credentials` method on the signing params builder that accepts credentials—no reason we can't have both credentials and identity.

aws/rust-runtime/aws-sigv4/src/lib.rs
Velfi: I'd say no, but I'll just make the new fn `pub(crate)` for now

aws/rust-runtime/aws-sigv4/src/lib.rs
jdisanti: The `Identity` debug impl can't really redact anything itself, but it delegates to the inner `Debug`, which for `Credentials`, is already redacted. It should be safe to use its `Debug` implementation directly.

aws/rust-runtime/aws-sigv4/src/http_request/error.rs
```diff
@@ -52,6 +65,7 @@ enum CanonicalRequestErrorKind {
     InvalidHeaderValue { source: InvalidHeaderValue },
     InvalidUtf8InHeaderValue { source: Utf8Error },
     InvalidUri { source: InvalidUri },
+    UnsupportedCredentialType,
```
jdisanti, line 68: Should be called `UnsupportedIdentityType`

aws/rust-runtime/aws-sigv4/src/lib.rs
```diff
@@ -29,14 +30,11 @@ pub mod event_stream;
 pub mod http_request;
 
 /// Parameters to use when signing.
+// #[derive(Debug)] is safe because `Identity` handles redaction.
```
jdisanti, line 33: `Identity` doesn't handle redaction, but the data inside of it is _supposed to_.

jdisanti (approved):
Looks good. Just two minor comments.

aws/rust-runtime/aws-sigv4/src/lib.rs
Velfi: I'll clarify that.

PR #2916, opened 2023-08-10
add support for nullable struct members when generating AWS SDKs

Velfi:
## Motivation and Context

smithy-rs#1767 aws-sdk-rust#536

## Description

This PR adds support for nullability i.e. much less unwraps will be required when using the AWS SDK. For generic clients, this new behavior can be enabled in codegen by setting `nullabilityCheckMode: "Client"` in their codegen config:
```
      "plugins": {
        "rust-client-codegen": {
          "codegen": {
            "includeFluentClient": false,
            "nullabilityCheckMode": "CLIENT_CAREFUL"
          },
     }
```

## Testing

Ran existing tests

## Checklist

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolParserGenerator.kt
```diff
@@ -166,14 +166,18 @@ class ProtocolParserGenerator(
...
+                                if (errorMessageMember != null) {
+                                    if (errorMessageMember.isOptional) {
+                                        rust(
+                                            """
+                                            if tmp.message.is_none() {
+                                                tmp.message = _error_message;
+                                            }
+                                            """,
+                                        )
+                                    }
```
rcoh, line 180: is this dropping it if it's not optional?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -509,10 +509,20 @@ class JsonParserGenerator(
...
+                    // TODO the need for this check should be eliminated or the server code should be updated.
+                    if (codegenContext.target == CodegenTarget.SERVER) {
+                        if (returnSymbolToParse.isUnconstrained) {
+                            rust("Ok(Some(builder))")
+                        } else {
+                            rust("Ok(Some(builder.build()))")
+                        }
                     } else {
-                        rust("Ok(Some(builder.build()))")
+                        if (hasFallibleBuilder(shape, symbolProvider)) {
```
rcoh, line 520: I think this can be if `(hasFaillibleBuilder || !returnSymbolToParse.isUnconstrained)`?

Alternatively, can we push `returnSymbolToParse` as an argument to `hasFallibleBuilder`? That might help us make sure we caught all of these locations.

aws/rust-runtime/aws-inlineable/src/route53_resource_id_preprocessor.rs
```diff
@@ -27,28 +27,24 @@ fn trim_resource_id(resource_id: &mut Option<String>) {
...
+            *resource_id = trimmed_id.to_string();
             return;
         }
     }
 }
 
 pub(crate) struct Route53ResourceIdInterceptor<G, T>
 where
-    G: for<'a> Fn(&'a mut T) -> &'a mut Option<String>,
+    G: for<'a> Fn(&'a mut T) -> Option<&'a mut String>,
```
rcoh, line 39: side note, might be nice to replace these existential types with a trait for clarity

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/route53/Route53Decorator.kt
```diff
@@ -96,10 +96,11 @@ class TrimResourceIdCustomization(
                         RuntimeType.forInlineDependency(
                             InlineAwsDependency.forRustFile("route53_resource_id_preprocessor"),
                         ).resolve("Route53ResourceIdInterceptor")
+
                     rustTemplate(
                         """
                         #{Route53ResourceIdInterceptor}::new(|input: &mut #{Input}| {
-                            &mut input.$fieldName
+                            Some(&mut input.$fieldName)
```
rcoh, line 103: can remove this option from both places right?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientCodegenVisitor.kt
```diff
@@ -78,7 +78,11 @@ class ClientCodegenVisitor(
         val rustSymbolProviderConfig = RustSymbolProviderConfig(
             runtimeConfig = settings.runtimeConfig,
             renameExceptions = settings.codegenConfig.renameExceptions,
-            nullabilityCheckMode = NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1,
+            nullabilityCheckMode = if (settings.codegenConfig.generateOptionsForRequiredShapes) {
+                NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1
+            } else {
+                NullableIndex.CheckMode.CLIENT_CAREFUL
+            },
```
rcoh, line 85: I think we want to allow generic clients to use `Client` instead of client careful

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/XmlBindingTraitParserGenerator.kt
```diff
@@ -474,16 +474,17 @@ class XmlBindingTraitParserGenerator(
...
-                        rustTemplate(""".map_err(|_|#{XmlDecodeError}::custom("missing field"))?""", *codegenScope)
-                    }
+
+                if (BuilderGenerator.hasFallibleBuilder(shape, symbolProvider)) {
+                    // NOTE:(rcoh) This branch is unreachable given the current nullability rules.
+                    // Only synthetic inputs can have fallible builders, but synthetic inputs can never be parsed
+                    // (because they're inputs, only outputs will be parsed!)
+
+                    // I'm leaving this branch here so that the binding trait parser generator would work for a server
+                    // side implementation in the future.
```
rcoh, line 484: I don't think this comment is true anymore

rust-runtime/aws-smithy-http/src/operation/error.rs
```diff
@@ -185,3 +186,9 @@ impl Error for BuildError {
         }
     }
 }
+
+impl From<BuildError> for DeserializeError {
+    fn from(build_error: BuildError) -> Self {
+        DeserializeError::custom_source("deserialization failed due to a build error", build_error)
```
rcoh, line 192: should we make a nicer error here? e.g `because a required field was missing`, or something?

rcoh: `builderror` is just a weird name to expose to a customer

rcoh:
I think we should also note that certain fields are required in the generated docs

rcoh (changes requested):
I'm so excited about this!

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolParserGenerator.kt
Velfi: AFAICT, this code is meant to set an error message. When the message is optional and wasn't provided, we set a generic message. This code isn't relevant when the message is required, as we can count on one being included in the response.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreRustSettings.kt
```diff
@@ -41,21 +41,25 @@ const val CODEGEN_SETTINGS = "codegen"
 open class CoreCodegenConfig(
     open val formatTimeoutSeconds: Int = defaultFormatTimeoutSeconds,
     open val debugMode: Boolean = defaultDebugMode,
+    open val generateOptionsForRequiredShapes: Boolean = defaultGenerateOptionsForRequiredShapes,
 ) {
     companion object {
         const val defaultFormatTimeoutSeconds = 20
         const val defaultDebugMode = false
+        const val defaultGenerateOptionsForRequiredShapes = true
```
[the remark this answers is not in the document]
Velfi: note to self: use CLIENT for generic clients

aws/sdk/integration-tests/dynamodb/tests/movies.rs
```diff
@@ -28,25 +28,29 @@ async fn create_table(client: &Client, table_name: &str) {
...
+                .build()
+                .unwrap(),
         )
         .attribute_definitions(
             AttributeDefinition::builder()
                 .attribute_name("title")
                 .attribute_type(ScalarAttributeType::S)
-                .build(),
+                .build()
+                .unwrap(),
```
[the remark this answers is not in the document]
Velfi: put another way, what if users could pass a builder directly?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientCodegenVisitor.kt
```diff
@@ -78,10 +77,12 @@ class ClientCodegenVisitor(
         val rustSymbolProviderConfig = RustSymbolProviderConfig(
             runtimeConfig = settings.runtimeConfig,
             renameExceptions = settings.codegenConfig.renameExceptions,
-            nullabilityCheckMode = NullableIndex.CheckMode.CLIENT_ZERO_VALUE_V1,
+            nullabilityCheckMode = settings.codegenConfig.nullabilityCheckMode,
             moduleProvider = ClientModuleProvider,
             nameBuilderFor = { symbol -> "${symbol.name}Builder" },
         )
+
+        println("rustSymbolProviderConfig:\n$rustSymbolProviderConfig")
```
jdisanti, line 85:
```suggestion
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/QuerySerializerGenerator.kt
```diff
@@ -171,11 +174,20 @@ abstract class QuerySerializerGenerator(codegenContext: CodegenContext) : Struct
...
-                serializeStructureInner(context)
+                context.copy(writerExpression = "writer", valueExpression = ValueExpression.Reference("input"))
+                    .also { inner ->
+                        for (member in inner.shape.members()) {
+                            val memberContext = MemberContext.structMember(inner, member, symbolProvider)
+                            structWriter(memberContext) { writerExpression ->
+                                serializeMember(memberContext.copy(writerExpression = writerExpression))
+                            }
+                        }
+                    }
```
jdisanti, line 185: This looks identical to `serializeStructureInner`?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
```diff
@@ -100,10 +100,10 @@ private object CodegenCoreTestModules {
     }
 }
 
-val TestRustSymbolProviderConfig = RustSymbolProviderConfig(
+fun testRustSymbolProviderConfig(nullabilityCheckMode: NullableIndex.CheckMode = NullableIndex.CheckMode.CLIENT) = RustSymbolProviderConfig(
```
jdisanti, line 103: This shouldn't have a default of `CLIENT` since it is in `codegen-core`.

rust-runtime/aws-smithy-http/src/operation/error.rs
```diff
@@ -185,3 +186,21 @@ impl Error for BuildError {
...
+            build_error,
+        )
+    }
+}
+
+#[cfg(feature = "event-stream")]
+impl From<BuildError> for aws_smithy_eventstream::error::Error {
+    fn from(build_error: BuildError) -> Self {
+        aws_smithy_eventstream::error::Error::unmarshalling(format!(
+            "error unmarshalling failed because a required field was missing: {build_error}"
```
jdisanti, line 203: How does this error know the underlying cause was a required field?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolParserGenerator.kt
```diff
@@ -166,14 +166,18 @@ class ProtocolParserGenerator(
...
-                                if (errorShape.errorMessageMember() != null) {
-                                    rust(
-                                        """
-                                        if tmp.message.is_none() {
-                                            tmp.message = _error_message;
-                                        }
-                                        """,
-                                    )
+                                val errorMessageMember = errorShape.errorMessageMember()
+                                // If the message member is optional and wasn't set, we set a generic error message.
```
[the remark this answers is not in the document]
Velfi: We should, but those structs won't be generated with an option so we'd have to handle it differently b/c this message is getting set after `.build()` has been called. Want me to create an issue to investigate possible improvements? I'd rather do that than prolong the review of this PR.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/QuerySerializerGenerator.kt
Velfi: It is. I think I broke it out b/c I wanted to change something but then it didn't end up being needed and I just forgot to switch back to using `serializeStructureInner`. I'll update this.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/Rust.kt
```diff
@@ -343,7 +343,7 @@ fun TestWriterDelegator.compileAndTest(
     val env = mapOf("RUSTFLAGS" to "-A dead_code")
     val testOutput = "cargo test".runCommand(baseDir, env)
     if (runClippy) {
-        "cargo clippy".runCommand(baseDir, env)
+        "cargo clippy --all-features --tests".runCommand(baseDir, env)
```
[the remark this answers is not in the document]
Velfi: Probably not, I'll leave `--all-features` but remove `--tests`

rust-runtime/aws-smithy-http/src/operation/error.rs
Velfi: Currently, that's the only reason that this occurs. I changed this from a more generic message b/c of [Russell's comment here](https://github.com/awslabs/smithy-rs/pull/2916#discussion_r1294734978). I'm open to suggestions. Another option is that I code-generate a `map_err` function for each place we need to convert this, but I wanted to avoid adding so many new lines to the SDKs.

Velfi:
Currently blocked on https://github.com/aws/aws-models/pull/819

CHANGELOG.next.toml
```diff
@@ -34,3 +34,15 @@ message = "Fix requests to S3 with `no_credentials` set."
 references = ["smithy-rs#2907", "aws-sdk-rust#864"]
 meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Struct members modeled as required are no longer wrapped in `Option`s."
+references = ["smithy-rs#2916", "aws-sdk-rust#536"]
```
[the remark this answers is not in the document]
Velfi: done https://github.com/awslabs/smithy-rs/discussions/2929

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolParserGenerator.kt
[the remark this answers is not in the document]
jdisanti: Oh, I see. Yeah, that seems challenging.

rust-runtime/aws-smithy-http/src/operation/error.rs
```diff
@@ -185,3 +186,21 @@ impl Error for BuildError {
         }
     }
 }
+
+impl From<BuildError> for DeserializeError {
+    fn from(build_error: BuildError) -> Self {
+        DeserializeError::custom_source(
```
david-perez, line 192: This conversion should probably live in the generated crate.

Otherwise with this  _all_ clients (that have fallible builders and hence use `BuildError` from `aws-smithy-http`) will depend on `aws-smithy-json` for this conversion, _regardless of their protocol_. So e.g. restXml clients will start depending on the `aws-smithy-json` crate, compiling a lot of unused code and causing confusion. See https://github.com/awslabs/smithy-rs/issues/1947, which is related.

david-perez (changes requested):

rcoh:
@Velfi the blocked changes have been merged

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/Rust.kt
```diff
@@ -343,7 +343,7 @@ fun TestWriterDelegator.compileAndTest(
     val env = mapOf("RUSTFLAGS" to "-A dead_code")
     val testOutput = "cargo test".runCommand(baseDir, env)
     if (runClippy) {
-        "cargo clippy".runCommand(baseDir, env)
+        "cargo clippy --all-features".runCommand(baseDir, env)
```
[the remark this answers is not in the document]
Velfi: I've broken out this change into a separate PR: https://github.com/awslabs/smithy-rs/pull/2963

rcoh (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -92,6 +97,37 @@ interface ServerProtocol : Protocol {
             .toType().resolve("protocol::$protocolModulePath::runtime_error::RuntimeError")
 }
 
+class ServerBuilderInstantiator(private val symbolProvider: RustSymbolProvider, private val symbolParseFn: (Shape) -> ReturnSymbolToParse) :
```
david-perez, line 100: This class should not be in this file. There's nothing about builders that is protocol-aware.

Also, we shouldn't rely on `ClientBuilderInstantiator` here. `BuilderGenerator.kt` (which hosts `ClientBuilderInstantiator`) is currently in `codegen-core`, but it shouldn't (see https://github.com/awslabs/smithy-rs/issues/1970), it should be in `codegen-client`. I'm ok merging this PR before #1970 gets tackled, but we shouldn't add _more_ tech debt.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -127,6 +129,34 @@ fun MemberShape.setterName() = "set_${this.memberName.toSnakeCase()}"
 // Getter names will never hit a reserved word and therefore never need escaping.
 fun MemberShape.getterName() = "get_${this.memberName.toSnakeCase()}"
 
+interface BuilderInstantiator {
+    fun setField(builder: String, value: Writable, e: MemberShape): Writable
```
david-perez, line 133:
```suggestion
    fun setField(builder: String, value: Writable, member: MemberShape): Writable
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/AwsJson.kt
```diff
@@ -148,6 +149,7 @@ open class AwsJson(
             codegenContext,
             httpBindingResolver,
             ::awsJsonFieldName,
+            builderInstantiator = ClientBuilderInstantiator(codegenContext.symbolProvider),
```
david-perez, line 152: nit: I'd prefer it if we didn't default to the client implementation. Historically `codegen-server` depended on `codegen-client` and when I made changes, I defaulted to the client implementation in the core classes to not have to change the client code, since it came first.

But now that we've untangled things and the server and the client sit on equal footing, I think it's best that we don't default to any implementation in the `codegen-core` classes. There's still some old code that defaults to the client implementation, but new code should not follow that example.

david-perez (changes requested):

david-perez (approved):
I think the PR description needs to be updated. The option is `nullabilityCheckMode`.

jdisanti (approved):

PR #2919, opened 2023-08-14
ban mio dependency so we don't accidentally break WASM users

Velfi:
When we bring in `mio`, it breaks WASM users. We don't currently have a need for `mio` so I've added a ban so that we don't accidentally bring it in.

This change also
- synchronizes the versions of `async-trait` we're using so that we don't bring in two versions of `syn`.
- fixes up the dep versions for the publisher so they can be less outdated.

----

### DIFF
--- a/aws/rust-runtime/aws-inlineable/Cargo.toml
+++ b/aws/rust-runtime/aws-inlineable/Cargo.toml
@@ -11,6 +11,9 @@ license = "Apache-2.0"
 publish = false
 repository = "https://github.com/awslabs/smithy-rs"
 
+[features]
+rt-tokio = ["aws-smithy-async/rt-tokio", "tokio/rt"]
+
 [dependencies]
 aws-credential-types = { path = "../aws-credential-types" }
 aws-endpoint = { path = "../aws-endpoint" }
@@ -34,7 +37,7 @@ http = "0.2.9"
 http-body = "0.4.5"
 md-5 = "0.10.1"
 ring = "0.16"
-tokio = { version = "1.23.1", features = ["full"] }
+tokio = { version = "1.23.1" }
 tokio-stream = "0.1.5"
 tower = { version = "0.4", default-features = false }
 tracing = "0.1"
@@ -47,6 +50,7 @@ aws-smithy-runtime-api = { path = "../../../rust-runtime/aws-smithy-runtime-api"
 tempfile = "3.6.0"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 aws-smithy-async = { path = "../../../rust-runtime/aws-smithy-async", features = ["test-util"] }
+tokio = { version = "1.23.1", features = ["rt", "macros"] }
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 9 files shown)


Velfi:
Putting this on hold since I couldn't figure it out in a short time.

aws/rust-runtime/aws-inlineable/Cargo.toml
```diff
@@ -34,7 +37,7 @@ http = "0.2.9"
 http-body = "0.4.5"
 md-5 = "0.10.1"
 ring = "0.16"
-tokio = { version = "1.23.1", features = ["full"] }
+tokio = { version = "1.23.1" }
```
simbleau, line 40: I think you need `default-features = false` here. I'm not sure that the default-features are wasm-compat, and I wouldn't count on it.

aws/rust-runtime/aws-inlineable/Cargo.toml
```diff
@@ -47,6 +50,7 @@ aws-smithy-runtime-api = { path = "../../../rust-runtime/aws-smithy-runtime-api"
 tempfile = "3.6.0"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 aws-smithy-async = { path = "../../../rust-runtime/aws-smithy-async", features = ["test-util"] }
+tokio = { version = "1.23.1", features = ["rt", "macros"] }
```
simbleau, line 53: Same here, needs `default-features = false`

aws/sdk/integration-tests/no-default-features/Cargo.toml
```diff
@@ -19,5 +19,5 @@ aws-sdk-s3 = { path = "../../build/aws-sdk/sdk/s3", default-features = false }
 aws-smithy-async = { path = "../../build/aws-sdk/sdk/aws-smithy-async" }
 aws-credential-types = { path = "../../build/aws-sdk/sdk/aws-credential-types", features = ["test-util"] }
 futures = "0.3.25"
-tokio = { version = "1.23.1", features = ["full", "test-util"] }
+tokio = { version = "1.23.1", features = ["macros", "test-util"] }
```
simbleau, line 22: needs `default-features = false`

rust-runtime/aws-smithy-client/Cargo.toml
```diff
@@ -47,7 +47,7 @@ aws-smithy-async = { path = "../aws-smithy-async", features = ["rt-tokio"] }
 hyper-tls = { version = "0.5.0" }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-tokio = { version = "1.23.1", features = ["full", "test-util"] }
+tokio = { version = "1.23.1", features = ["macros", "test-util"] }
```
simbleau, line 50: needs `default-features = false`

simbleau (changes requested):
The only features that are wasm-compat for wasm32-unknown-unknown are:
 
    sync
    macros
    io-util
    rt
    time

https://docs.rs/tokio/latest/tokio/#wasm-support

The default features are seemingly not wasm-compatible.

aws/rust-runtime/aws-inlineable/Cargo.toml
rcoh: this cargo.toml isn't actually used in anything

PR #2931, opened 2023-08-18
Fix re-export of `SdkError`

jdisanti:
`SdkError` was re-exported in generated crates with the wrong generic argument for the response type, which causes the SDK examples to fail to compile. The example breaks weren't caught due to a bug in the CI script that checks them, addressed in smithy-rs#2930.

----


jdisanti:
It looks like this fixes the examples, but a change to aws-doc-sdk-examples needs to get merged before this can: https://github.com/awsdocs/aws-doc-sdk-examples/pull/5286

PR #2935, opened 2023-08-21
Fix `SDK::Endpoint` built-in

jdisanti:
Setting `endpoint_url` on a SDK Config currently has no effect due to a bug in deciding if a parameter is a built-in or not in EndpointBuiltinsDecorator. The generated code is placing the endpoint URL into the ConfigBag with the correct `aws_types::endpoint_config::EndpointUrl` type, but then the operation runtime plugin is trying to pull it out of the ConfigBag with the `crate::config::EndpointUrl` type, which always yields nothing.

Or, to illustrate with the generated code itself:
```
    pub fn set_endpoint_url(&mut self, endpoint_url: Option<::std::string::String>) -> &mut Self {
        self.config.store_or_unset(endpoint_url.map(::aws_types::endpoint_config::EndpointUrl));
        self
    }
```
vs.
```
let params = crate::config::endpoint::Params::builder()
            .set_endpoint(cfg.load::<crate::config::EndpointUrl>().map(|ty| ty.0.clone()))
```

This PR adds a unit test that reproduces the issue, and then fixes it.

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
@@ -57,7 +57,9 @@ fun ClientCodegenContext.getBuiltIn(builtIn: String): Parameter? {
 }
 
 private fun promotedBuiltins(parameter: Parameter) =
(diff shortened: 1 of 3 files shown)


ysaito1001 (approved):

rcoh:
side note: we may want to add a diff target that captures changes like this

PR #2960, opened 2023-08-29
Generate a region setter in config when the model uses SigV4

jdisanti:
SDK ad-hoc clients sometimes have endpoint rulesets without a region, but use SigV4. In these cases, the region setter on config is still necessary for signing to succeed, so they should retain the setters for region.

----


ysaito1001 (approved):
Unit tests helped understand from concrete use cases!

Issue #1668, opened 2022-08-25
Verify XML APIs are returning XML with root elements that match the models 


Velfi:
@jdisanti did some extra [investigative](https://github.com/awslabs/smithy-rs/pull/1665#discussion_r954371345) [work](https://github.com/awslabs/smithy-rs/pull/1665#discussion_r954373006) while reviewing my PR [Add "invalid xml body root" check exemption for S3's `GetObjectAttributes`](https://github.com/awslabs/smithy-rs/pull/1665) and found several other API that may require the same customization:

- `s3#GetBucketLocation`
- `s3#ListObjects`
- `s3#ListObjectVersions`
- `ec2#GetConsoleOutput`

Test these APIs to see if they return XML errors. If they do, apply the same customization from my PR.

**Alternatively,**

Implement [John's suggestion](https://github.com/awslabs/smithy-rs/pull/1665#discussion_r954373006) to use `xmlName` instead of `allowInvalidXmlRoot` and update any buggy operations (don't forget `s3#GetObjectAttributes`)

jdisanti:
This gets especially muddy since some of the S3 operations overlap. For example, making a call to `ListObjectsV2` without a bucket set results in the XML tag name being `ListAllMyBucketsResult` and a list of buckets in the response.

jdisanti:
Errors are also complicated: https://github.com/awslabs/aws-sdk-rust/issues/873

rcoh:
For `ListObjectsV2` (and other S3 operations with bucket) I _think_ the right answer is actually making `bucket` required—`Bucket` is bound as a httpLabel effectively so it it's missing we're really not sending the right request

jdisanti:
I don't see value in keeping this issue open anymore. I think most of the XML root issues have been discovered and fixed at this point.

PR #2964, opened 2023-08-30
Make required context parameters required

rcoh:
## Motivation and Context

When a `@contextParam` is marked as required, we will enforce it on inputs. Since these fields may influence endpoint, omitting them can result in a different target being hit.

- #1668 
- aws-sdk-rust#873

## Testing
- [x] S3 Integration test

## Checklist

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -76,10 +80,59 @@ abstract class BuilderCustomization : NamedCustomization<BuilderSection>()
...
+private val checkSetString = RuntimeType.forInlineFun("non_empty_str", RustModule.private("serde_util")) {
+    rustTemplate(
+        """
+        pub (crate) fn non_empty_str<T: AsRef<str>>(field: Option<T>) -> Option<T> {
+            if let Some(field) = field {
+                if field.as_ref() != "" {
+                    return Some(field)
+                }
+            }
+            None
```
jdisanti, line 123: I think this can be one line:
```suggestion
field.map(AsRef::as_ref).filter(|s| s != "")
```

jdisanti (approved):

PR #2972, opened 2023-09-05
Port redacting sensitive body to orchestrator

ysaito1001:
## Motivation and Context
Fixes https://github.com/awslabs/smithy-rs/issues/2926

## Description
This PR ports logic implemented in https://github.com/awslabs/smithy-rs/pull/2603. Thankfully, even though we did not port this at the time of the orchestrator launch, the orchestrator has not logged sensitive bodies because we have never logged response bodies in the orchestrator code.

The code changes in this PR
- now logs response bodies in `try_attempt`
- ports the logic from the previous PR in question to the orchestrator, via an interceptor

Now, when credentials providers in `aws_config` need to say "I want to redact a response body" ([example](https://github.com/awslabs/smithy-rs/blob/2c27834f90b0585dba60606f9da6246b341227d6/aws/rust-runtime/aws-config/src/http_credential_provider.rs#L48)) when middleware is gone, they can pass an interceptor `SensitiveOutputInterceptor` to `Config` of whatever clients they are using.

## Testing
Depends on the existing tests.

Without the logic ported over the orchestrator and by logging response bodies unconditionally in `try_attempt`, we got the following failures. After we've ported the logic, they now pass.
```
    default_provider::credentials::test::ecs_assume_role
    default_provider::credentials::test::imds_assume_role
    default_provider::credentials::test::sso_assume_role
    default_provider::credentials::test::web_identity_token_env
    default_provider::credentials::test::web_identity_token_profile
    default_provider::credentials::test::web_identity_token_source_profile
    profile::credentials::test::e2e_assume_role
    profile::credentials::test::region_override
    profile::credentials::test::retry_on_error
```

## Checklist

----


CHANGELOG.next.toml
```diff
@@ -102,3 +102,15 @@ message = "Required members with @contextParam are now treated as client-side re
...
 author = "rcoh"
+
+[[aws-sdk-rust]]
+message = "Port redacting sensitive HTTP response bodies to the orchestrator"
+references = ["smithy-rs#2926", "smithy-rs#2972"]
+meta = { "breaking" = false, "tada" = false, "bug" = true }
+author = "ysaito1001"
+
+[[smithy-rs]]
+message = "Port redacting sensitive HTTP response bodies to the orchestrator"
```
jdisanti, line 113: Consider calling this, "Fix regression with redacting sensitive HTTP response bodies"

rust-runtime/aws-smithy-runtime/src/client/orchestrator/http.rs
```diff
@@ -33,3 +41,41 @@ pub(crate) async fn read_body(response: &mut HttpResponse) -> Result<(), <SdkBod
...
+        "SensitiveOutputInterceptor"
+    }
+
+    fn modify_before_deserialization(
+        &self,
+        _context: &mut BeforeDeserializationInterceptorContextMut<'_>,
+        _runtime_components: &RuntimeComponents,
+        cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        cfg.interceptor_state().store_put(SensitiveOutput);
```
jdisanti, line 66: Should the operation runtime plugin add this to config rather than it being a separate interceptor?

rust-runtime/aws-smithy-runtime/src/client/orchestrator/http.rs
```diff
@@ -33,3 +41,41 @@ pub(crate) async fn read_body(response: &mut HttpResponse) -> Result<(), <SdkBod
...
+
+pub(crate) fn log_response_body(response: &HttpResponse, cfg: &ConfigBag) {
+    if cfg.load::<SensitiveOutput>().is_none()
+        || std::env::var(LOG_SENSITIVE_BODIES)
+            .map(|v| v.eq_ignore_ascii_case("true"))
+            .unwrap_or_default()
+    {
+        trace!(response = ?response, "read HTTP response body");
+    } else {
+        trace!(response = "** REDACTED **. To print, set LOG_SENSITIVE_BODIES=true")
```
jdisanti, line 79:
```suggestion
        trace!(response = ?response, "read HTTP response body");
    } else {
        trace!(response = "** REDACTED **. To print, set LOG_SENSITIVE_BODIES=true", "read HTTP response body")
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/SensitiveOutputDecorator.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    override fun operationCustomizations(
+        codegenContext: ClientCodegenContext,
+        operation: OperationShape,
+        baseCustomizations: List<OperationCustomization>,
+    ): List<OperationCustomization> =
+        baseCustomizations + listOf(SensitiveOutputCustomization(codegenContext, operation))
+}
+
+private class SensitiveOutputCustomization(
```
jdisanti, line 31: Should we have a Kotlin clientIntegrationTest that examines this end to end?

rust-runtime/aws-smithy-runtime/src/client/orchestrator/http.rs
ysaito1001: If the operation runtime plugin added it, then the question would be how can existing call sites of a generic client in `aws_config` ([example](https://github.com/awslabs/smithy-rs/blob/2c27834f90b0585dba60606f9da6246b341227d6/aws/rust-runtime/aws-config/src/http_credential_provider.rs#L48)) express they want to redact response bodies, once the middleware has been removed?

Previously with the middleware, they were able to express that through the implementation of the [sensitive](https://github.com/awslabs/smithy-rs/blob/75461b9db9a68131efcfaee05ae8440548fed945/aws/rust-runtime/aws-config/src/imds/client.rs#L284-L286) method of the `ParseStringResponse` trait. But that trait is gone now in the orchestrator.

With the orchestrator, I believe those call sites can only specify redacting response bodies through an interceptor.

CHANGELOG.next.toml
ysaito1001: +1. Addressed in 590e28f.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/http.rs
jdisanti: A runtime plugin can also add config, so I don't think it's limited to just interceptors. The `SensitiveOutput` config value could still be available for aws-config to add.

ysaito1001: Addressed in e6c60da

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/SensitiveOutputDecorator.kt
ysaito1001: Test added in 71c39a9

rust-runtime/aws-smithy-runtime/src/client/orchestrator/http.rs
```diff
@@ -33,3 +37,25 @@ pub(crate) async fn read_body(response: &mut HttpResponse) -> Result<(), <SdkBod
 
     Ok(())
 }
+
+/// Marker value stored in the config bag to indicate that a response body should be redacted.
+#[derive(Debug)]
+pub struct SensitiveOutput;
```
jdisanti, line 43: One last thing, I think this should be in aws-smithy-runtime-api alongside `LoadedRequestBody`.

ysaito1001: Thanks for suggesting a better place. Addressed in b4a68c4.

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -36,7 +36,8 @@ use tracing::{debug, debug_span, instrument, trace, Instrument};
 mod auth;
 /// Defines types that implement a trait for endpoint resolution
 pub mod endpoints;
-mod http;
+/// Defines types that work with HTTP types
+pub mod http;
```
jdisanti, line 40: Looks like this doesn't need to be pub anymore.

jdisanti (commented):
`cargo-semver-checks` is saying this is a breaking change, but I don't see it...

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
jdisanti, line 40:
```suggestion
mod http;
```

jdisanti (approved):

Issue #2417, opened 2023-02-24
[Feature Request] Allow configuration of RetryClassifier on the client


DanielBauman88:
This lack of functionality is a significant regression over clients in other languages.
The default retry classifier in smithy is the AwsResponseRetryClassifier.

Smithy allows for the retry config to be modified (max attempts and backoff params) but it doesn't allow for any way to inject customer retry classification. It's possible (but not super ergonomic) to inject one's own retry policy but then one loses out on being able to use the supplied configuration for max attempts and backoff and the good default classification which one might want to just add a few cases to.

The generated code does create a customize method on the client but it is parametrized to the specific AwsResponseRetryClassifier so even when using customize on the request level a custom classifier cannot be supplied. Even if it could, doing this at the request level might be needed in some cases but the client should allow a client level configuration of a retry classifier.

The locked down cuztomize method:
```
        pub async fn customize(
            self,
        ) -> std::result::Result<
            crate::operation::customize::CustomizableOperation<
                crate::operation::SOME-OPERATION,
                aws_http::retry::AwsResponseRetryClassifier,
            >,
            aws_smithy_http::result::SdkError<crate::error::SOME-ERROR>,

```

TLDR, support a function on the client builder to set a retry classifier. This will allow consumers of smithy clients to use the AwsResponseRetryClassifier and based on its result to modify the classification however they choose. This can be done with java and go, and is functionality that a client to be used in production by customers should support.

This more specific issue was created after some digging - resolved the old [issue](https://github.com/awslabs/smithy-rs/issues/2405).

jdisanti:
What is your use-case for needing to do this?

DanielBauman88:
I may want to retry a wider set of exceptions that the default retrclassifier.

For example - I'm interacting with DDB and my transactions are written in such a way that I want to retry any  TransactionCanceledException that happens because of a TransactionConflict reason.

Or if I'm using a service that throws ConflictExceptions and for my application I always want to retry these.

Or if I'm using an eventually consistent service and I always want to retry the not-found errors that might come after fresh creates.

DanielBauman88:
One more thing to tag onto this issue - the smithy rs client is in general extremely restrictive.

In other languages its easy to keep the default behaviour but plug into the logic to write some logs or post a metric when retries happen.

The rust sdk makes that impossible. Either an entire retry config has to be written which cannot benefit from the default logic, or no additional behaviour can be customized.

[...]

PR #2977, opened 2023-09-08
implement user-configurable retry classifiers

Velfi:
[Read the RFC here](https://github.com/awslabs/smithy-rs/pull/3018)

## Motivation and Context

#2417

## Description

Exactly what it says on the tin. I have a related RFC to publish that goes into more depth.

## Testing

I wrote an integration test that ensures a custom retry classifier can be set and is called.

## Checklist

----


rust-runtime/aws-smithy-runtime-api/src/client/retries/classifiers.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn classify_retry(
+        &self,
+        ctx: &InterceptorContext,
+        result_of_preceding_classifier: Option<RetryClassifierResult>,
+    ) -> Option<RetryClassifierResult>;
+
+    /// The name of this retry classifier.
+    ///
+    /// Used for debugging purposes
+    fn name(&self) -> &'static str;
```
rcoh, line 104: do we want this only for debugging purposes? Or would we also want to be able to use this (or something like it) to remove classifiers from the list

rust-runtime/aws-smithy-runtime-api/src/client/retries/classifiers.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub fn new(retry_classifier: impl ClassifyRetry + 'static) -> Self {
+        Self(Arc::new(retry_classifier))
+    }
+}
+
+impl ClassifyRetry for SharedRetryClassifier {
+    fn classify_retry(
+        &self,
+        ctx: &InterceptorContext,
+        result_of_preceding_classifier: Option<RetryClassifierResult>,
```
rcoh, line 129: in the RFC, I want to see the customer use cases this enables

rust-runtime/aws-smithy-runtime-api/src/client/retries/classifiers.rs
```diff
@@ -0,0 +1,170 @@
...
+use crate::client::interceptors::context::InterceptorContext;
+use aws_smithy_types::retry::ErrorKind;
+use std::fmt;
+use std::sync::Arc;
+use std::time::Duration;
+
+/// The result of running a [`ClassifyRetry`] on a [`InterceptorContext`].
+#[non_exhaustive]
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum RetryClassifierResult {
```
rcoh, line 17: result implies error to me—maybe `RetryAction`?

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
```diff
@@ -72,14 +73,10 @@ impl Interceptor for ConnectionPoisoningInterceptor {
...
-        let retry_classifiers = runtime_components
-            .retry_classifiers()
-            .ok_or("retry classifiers are required for connection poisoning to work")?;
-
-        let error_is_transient = retry_classifiers
-            .classify_retry(context.inner_mut())
-            .map(|reason| reason == RetryReason::Error(ErrorKind::TransientError))
-            .unwrap_or_default();
+        let retry_classifier_result =
+            run_classifiers_on_ctx(runtime_components.retry_classifiers(), context.inner_mut());
```
rcoh, line 77: seems odd to be passing a mutable reference to context here

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
```diff
@@ -119,41 +167,78 @@ impl HttpStatusCodeClassifier {
...
+        result = classifier.classify_retry(ctx, result);
+        match result.as_ref() {
+            Some(result) => tracing::debug!(
+                "running retry classifier '{}' returned result '{result}'",
+                classifier.name()
+            ),
+            None => tracing::debug!(
+                "running retry classifier '{}' returned no result",
+                classifier.name()
+            ),
```
rcoh, line 224: this is _probably_ too noisy at the debug level because we're going to get multiple log lines for every request. Maybe only log if a retry is actually indicated?

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
```diff
@@ -119,41 +167,78 @@ impl HttpStatusCodeClassifier {
...
+/// context. Each classifier is passed the classification result from the previous classifier (the
+/// 'root' classifier is passed `None`.)
+pub fn run_classifiers_on_ctx(
+    classifiers: impl Iterator<Item = SharedRetryClassifier>,
+    ctx: &InterceptorContext,
+) -> RetryClassifierResult {
+    let mut result = None;
+
+    for classifier in classifiers {
+        result = classifier.classify_retry(ctx, result);
```
rcoh, line 215: the one thing that might be helpful to log here is if one classifier said we should retry but then we set it to `None` afterwards because that's a classifier explicitly saying "disable"

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
```diff
@@ -63,13 +89,31 @@ impl<E> SmithyErrorClassifier<E> {
...
 
 impl<E> ClassifyRetry for SmithyErrorClassifier<E>
 where
     E: StdError + Send + Sync + 'static,
 {
-    fn classify_retry(&self, ctx: &InterceptorContext) -> Option<RetryReason> {
+    fn classify_retry(
+        &self,
+        ctx: &InterceptorContext,
+        _: Option<RetryClassifierResult>,
```
rcoh, line 115: is this a bug? I think this will ignore previous classification results?

rcoh (changes requested):
If I'm reading this PR correctly, all (?) of our built-in classifiers are actually incorrect since they ignore the previous result causing retries to be blown away.

I think this is probably a hint that this interface as written is a tripping hazard / face rake and we should try to revise the implementor experience to make it easier to do the right thing—possibly by just altering the behavior of the main loop to not replace `Some` with `None`?

Can dig in more in the text based RFC

rust-runtime/aws-smithy-runtime-api/src/client/retries/classifiers.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// "A retryable error was received. This is what kind of error it was,
+    /// in case that's important."
+    Error(ErrorKind),
+    /// "The server told us to wait this long before retrying the request."
+    Explicit(Duration),
+    /// "This response should not be retried."
```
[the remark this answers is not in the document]
Velfi: It's meant to communicate the cases where you'd commonly see these results. I'll make the comments more declarative and less conversational.

rust-runtime/aws-smithy-runtime-api/src/client/retries/classifiers.rs
Velfi: We can change the comment if we decide to implement that feature in the future.

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
Velfi: It's b/c I need this to be the actual context type, not `BeforeDeserializationInterceptorContextMut` and because that's a `mut` context type, it doesn't have an `inner` method, only an `inner_mut` method. Want me to add an `inner` method?

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
```diff
@@ -63,13 +89,31 @@ impl<E> SmithyErrorClassifier<E> {
...
+    }
 }
 
 impl<E> ClassifyRetry for SmithyErrorClassifier<E>
 where
     E: StdError + Send + Sync + 'static,
 {
-    fn classify_retry(&self, ctx: &InterceptorContext) -> Option<RetryReason> {
+    fn classify_retry(
+        &self,
```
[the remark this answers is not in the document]
Velfi: This behavior was encompassed in a single check in previous versions and that's why timeouts are grouped with IO. I named this `Smithy` because it's the most general classifier, the default which all clients use (alongside the 'modeled error' and 'HTTP status code' classifiers). I think splitting it makes sense and would make naming easy so that's what I'll do.

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
Velfi: OK, I'll only log disagreements

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/RetryClassifierConfigCustomization.kt
```diff
@@ -0,0 +1,120 @@
...
+    )
+
+    override fun section(section: ServiceConfig) =
+        writable {
+            when (section) {
+                ServiceConfig.ConfigImpl -> rustTemplate(
+                    """
+                    /// Returns retry classifiers currently registered by the user.
+                    pub fn retry_classifiers(&self) -> impl Iterator<Item = #{SharedRetryClassifier}> + '_ {
+                        self.runtime_components.retry_classifiers()
```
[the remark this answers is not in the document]
Velfi: I chose to return an iterator here because that's what we're doing for other 'list collected' things like interceptors.

rust-runtime/aws-smithy-runtime-api/src/client/retries/classifiers.rs
```diff
@@ -0,0 +1,170 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RetryClassifierPriority {
+    /// The default priority for the [`HttpStatusCodeClassifier`].
+    HttpStatusCodeClassifier,
+    /// The default priority for the [`ModeledAsRetryableClassifier`].
+    ModeledAsRetryableClassifier,
+    /// The default priority for the [`SmithyErrorClassifier`].
+    SmithyErrorClassifier,
+    /// The priority of some other classifier.
+    Other(i8),
```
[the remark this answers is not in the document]
Velfi: You're right. I need to add a wrapper struct.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/operation.rs
```diff
@@ -204,15 +205,18 @@ impl<I, O, E> OperationBuilder<I, O, E> {
         self
     }
 
-    pub fn no_retry(mut self) -> Self {
+    pub fn retry_classifier(
+        mut self,
+        retry_classifier: impl IntoShared<SharedRetryClassifier>,
```
jdisanti, line 210: `impl RetryClassifier + 'static`

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/fixed_delay.rs
```diff
@@ -72,31 +73,29 @@ impl RetryStrategy for FixedDelayRetryStrategy {
...
-        let backoff = match retry_reason {
-            Some(RetryReason::Explicit(_)) => self.fixed_delay,
-            Some(RetryReason::Error(_)) => self.fixed_delay,
-            Some(_) => {
-                unreachable!("RetryReason is non-exhaustive. Therefore, we need to cover this unreachable case.")
-            }
-            None => {
-                tracing::trace!(
+        let backoff = match classifier_result {
+            RetryAction::RetryAfter(_) => self.fixed_delay,
```
jdisanti, line 80: Shouldn't this use the delay inside of `RetryAfter`? Is there a test for explicit retry delay?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/RetryClassifierConfigCustomization.kt
```diff
@@ -0,0 +1,156 @@
...
+                        /// ## Examples
+                        /// ```no_run
+                        /// ## ##[cfg(test)]
+                        /// ## mod tests {
+                        /// ## ##[test]
+                        /// ## fn example() {
+                        /// ## }
+                        /// ## }
+                        /// ```
+                        pub fn push_retry_classifier(&mut self, retry_classifier: #{SharedRetryClassifier}) -> &mut Self {
```
Velfi, line 81: TODO add examples

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/fixed_delay.rs
Velfi: It's up for debate imo. If you set a retry strategy that's meant to always use the same delay, then I think no. I think we could just remove this strategy now. I believe I originally created it before we had Russell's `InstantSleep`. I'll remove it in a separate PR.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/RetryClassifierConfigCustomization.kt
```diff
@@ -0,0 +1,156 @@
...
+                            self.runtime_components.push_retry_classifier(retry_classifier);
```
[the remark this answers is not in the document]
Velfi: I was trying to mirror how the Interceptor setters API.

rust-runtime/aws-smithy-runtime/src/client/connectors/connection_poisoning.rs
Velfi: never mind, John changed the interceptor context object used here so the point is moot.

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
Velfi: Yeah, it's fixed now.

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/fixed_delay.rs
Velfi: nvm, I just removed it in this one since nothing was using it.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -471,7 +473,7 @@ class FluentClientGenerator(
     }
 }
 
-private fun baseClientRuntimePluginsFn(runtimeConfig: RuntimeConfig): RuntimeType =
+private fun baseClientRuntimePluginsFn(runtimeConfig: RuntimeConfig, customizations: List<FluentClientCustomization>): RuntimeType =
```
rcoh, line 476: extra arg is unused?

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
```diff
@@ -23,76 +25,100 @@ impl<E> ModeledAsRetryableClassifier<E> {
             _inner: PhantomData,
         }
     }
+
+    /// Return the priority of this retry classifier.
+    pub fn priority() -> RetryClassifierPriority {
+        RetryClassifierPriority::modeled_as_retryable_classifier()
```
rcoh, line 31: consider renaming this something like `RetryTraitClassifier` (non blocking)

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
```diff
@@ -119,45 +145,86 @@ impl HttpStatusCodeClassifier {
...
+        tracing::trace!(
+            "Classifier '{}' set the result of classification to '{}'",
+            classifier.name(),
+            new_result
+        );
+        result = new_result;
+
+        // If the result is `RetryForbidden`, stop running classifiers.
+        if result == RetryAction::RetryForbidden {
+            break;
```
rcoh, line 208: I think we should probably log something here

rust-runtime/aws-smithy-runtime-api/src/client/retries/classifiers.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Run this classifier on the [`InterceptorContext`] to determine if the previous request
+    /// should be retried. Returns a [`RetryAction`].
+    fn classify_retry(&self, ctx: &InterceptorContext) -> RetryAction;
+
+    /// The name of this retry classifier.
+    ///
+    /// Used for debugging purposes.
+    fn name(&self) -> &'static str;
+
+    /// The priority of this retry classifier. Classifiers with a higher priority will run before
```
rcoh, line 144: rather than noting the order it should probably say that higher priority overrides lower priority

rcoh (approved):
Overall looks great. My only top-level thought to consider is that we may want to split `DontCare` out from the list of RetryActions since that's something only a classifier cares about (and the orchestrator doesn't really).


Thinking something like:
```
enum RetryAction {
  ActionIndicated(RetryReason),
  NoActionIndicated
}
```
or something. This is definitely not critical though.

aws/sdk/integration-tests/s3/tests/data/request-information-headers/three-retries_and-then-success.json
```diff
@@ -84,7 +84,7 @@
       "action": {
         "Data": {
           "data": {
-            "Utf8": "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n    \"<Error>\n    <Type>Server</Type>\n    <Code>InternalError</Code>\n    <Message>We encountered an internal error. Please try again.</Message>\n    <RequestId>foo-id</RequestId>\n</Error>\""
```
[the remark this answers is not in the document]
Velfi: Yeah, IDK why. I probably copy-pasted something wrong when I created these tests.

aws/rust-runtime/aws-config/src/profile/credentials/exec.rs
```diff
@@ -35,11 +33,12 @@ impl AssumeRoleProvider {
         input_credentials: Credentials,
         sdk_config: &SdkConfig,
     ) -> provider::Result {
-        let config = sdk_config
+        let config = aws_sdk_sts::Config::from(sdk_config)
             .to_builder()
-            .credentials_provider(SharedCredentialsProvider::new(input_credentials))
+            .credentials_provider(input_credentials)
             .build();
-        let client = StsClient::new(&config);
```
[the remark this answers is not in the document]
Velfi: Yeah, this had a purpose at one point but then I changed things. I'll just removed these changes.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
Velfi: I think this was another leftover thing. I've remove the change.

rust-runtime/aws-smithy-runtime/src/client/retries/classifiers.rs
Velfi: I'll add a trace log

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/RetryClassifierDecorator.kt
```diff
@@ -24,43 +25,49 @@ class RetryClassifierDecorator : ClientCodegenDecorator {
         baseCustomizations: List<OperationCustomization>,
     ): List<OperationCustomization> = baseCustomizations +
         OperationRetryClassifiersFeature(codegenContext, operation)
+
+    override fun serviceRuntimePluginCustomizations(
```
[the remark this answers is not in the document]
Velfi: It used to mount all 5 classifiers. Now it only mounts the two that need knowledge of a specific error type to work.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customize/RequiredCustomizations.kt
```diff
@@ -49,15 +52,17 @@ class RequiredCustomizations : ClientCodegenDecorator {
         baseCustomizations +
             MetadataCustomization(codegenContext, operation) +
             IdempotencyTokenGenerator(codegenContext, operation) +
-            HttpChecksumRequiredGenerator(codegenContext, operation)
+            HttpChecksumRequiredGenerator(codegenContext, operation) +
+            RetryClassifierOperationCustomization(codegenContext, operation)
```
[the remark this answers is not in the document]
Velfi: Previously, all retry classifiers were set per-operation in the `RetryClassifierDecorator`. Now, they can be set at the config level AND the operation level.

aws/sdk/integration-tests/s3/tests/retry-classifier-customization.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        // Interceptors may call this classifier before a response is received. If a response was received,
+        // ensure that it has the expected status code.
+        if let Some(res) = ctx.response() {
+            assert_eq!(
+                res.status(),
+                500,
+                "expected a 500 response from test connection"
+            );
+        }
```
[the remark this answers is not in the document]
Velfi: _(I'm worried I don't understand what specifically you're questioning here, lmk if I got it wrong)_
Interceptors can get all classifiers in an iterator from the runtime components. The retry information headers thing therefore will call this classifier when running the below test.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/RetryClassifierDecorator.kt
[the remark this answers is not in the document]
Velfi: Scratch that, I combined the AWS classifiers because of the new "retry after" handling. This is now covered by the AWS Error code classifier

aws/rust-runtime/aws-config/src/imds/client.rs
```diff
@@ -548,32 +546,26 @@ impl EndpointResolver for ImdsEndpointResolver {
...
+                _ if status.as_u16() == 401 => RetryAction::Retry(ErrorKind::ServerError),
+                // This catch-all includes successful responses that fail to parse. These should not be retried.
+                _ => RetryAction::DontCare,
+            }
         } else {
             // Don't retry timeouts for IMDS, or else it will take ~30 seconds for the default
             // credentials provider chain to fail to provide credentials.
             // Also don't retry non-responses.
-            None
+            RetryAction::DontCare
```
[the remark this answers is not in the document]
Velfi: I've updated the structure of RetryAction/RetryReason.

aws/rust-runtime/aws-config/src/imds/client.rs
```diff
@@ -548,32 +546,26 @@ impl EndpointResolver for ImdsEndpointResolver {
...
+    fn classify_retry(&self, ctx: &InterceptorContext) -> RetryAction {
         if let Some(response) = ctx.response() {
-            Self::classify(response)
+            let status = response.status();
+            match status {
+                _ if status.is_server_error() => RetryAction::Retry(ErrorKind::ServerError),
+                // 401 indicates that the token has expired, this is retryable
+                _ if status.as_u16() == 401 => RetryAction::Retry(ErrorKind::ServerError),
+                // This catch-all includes successful responses that fail to parse. These should not be retried.
+                _ => RetryAction::DontCare,
```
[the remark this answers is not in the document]
Velfi: I've updated the structure of RetryAction/RetryReason.

ysaito1001 (approved):

PR #2978, opened 2023-09-09
Remove `futures_core::stream::Stream` from `aws-smithy-async`

ysaito1001:
## Motivation and Context
Removes `futures_core::stream::Stream` from the `aws-smithy-async` crate.

## Description
This PR is part of our ongoing effort, https://github.com/awslabs/smithy-rs/issues/2413. We remove the `futures_core::stream::Stream` trait from the public API in the `aws-smithy-async` crate. While doing so, we compensate
- `FnStream` by providing the explicit `.next` and `.collect` methods to let the previously working code continue working.
- `TryFlatMap` by making it return a new-type wrapper `PaginationStream` to hide `FnStream` from those who use paginators.

With this change, the latest canary no longer uses `tokio_stream::StreamExt`, since the paginator does not work in terms of the `Stream` trait. Furthermore, `aws-sdk-rust` has been more than [3 releases since release-2023-01-26](https://github.com/awslabs/aws-sdk-rust/releases), so the module `release-2023-01-26` has been removed from `canary-lambda`.

## Testing
No new tests added, but made sure the existing tests keep working with the change.

## Checklist

----


rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
```diff
@@ -183,10 +255,16 @@ mod test {
...
-        assert!(test_stream.poll_next().is_pending());
+        let _ = test_stream.enter(|ctx, pin| {
+            let polled = pin.poll_next(ctx);
+            assert!(polled.is_pending());
+        });
+        let _ = test_stream.enter(|ctx, pin| {
+            let polled = pin.poll_next(ctx);
+            assert!(polled.is_pending());
+        });
     }
```
rcoh, line 268: can you explain what this change does?

rcoh (commented):
nice! like the direction here.

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
ysaito1001: Yes, added a comment explaining why in 2b1fcb1.

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
```diff
@@ -3,16 +3,51 @@
...
+impl<Item> PaginationStream<Item> {
+    /// Creates a `PaginationStream` from the given [`FnStream`].
+    pub fn new(stream: FnStream<Item>) -> Self {
+        Self(stream)
+    }
+
+    /// Consumes and returns the next `Item` from this stream.
+    pub async fn next(&mut self) -> Option<Item>
+    where
+        Self: Unpin,
```
rcoh, line 34: was this bound here before?

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
```diff
@@ -3,16 +3,51 @@
...
+    {
+        self.0.next().await
+    }
+
+    /// Consumes this stream and gathers elements into a collection.
+    pub async fn collect<T: collect::sealed::Collectable<Item>>(self) -> T {
+        self.0.collect().await
+    }
+}
```
rcoh, line 43: I guess you can't map streams anymore? We could add a conversion to `aws-smithy-types-convert` I guess to regain access to `StreamExt`?

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
```diff
@@ -3,16 +3,51 @@
...
+    /// Consumes and returns the next `Item` from this stream.
+    pub async fn next(&mut self) -> Option<Item>
+    where
+        Self: Unpin,
+    {
+        self.0.next().await
+    }
+
+    /// Consumes this stream and gathers elements into a collection.
+    pub async fn collect<T: collect::sealed::Collectable<Item>>(self) -> T {
```
rcoh, line 40: what do you think about `try_collect` to do the `Result` flattening? I think that's more intuitive to use since it doesn't need a crazy turbofish. It will also remove the need for the Collectible trait

rcoh: (collect would return `Vec<Result`, try collect returns `Result<Vec`)

rcoh: Or we could add this method in addition and keep the generic 🤷🏻 

I just don't like typing `.collect::<Result<Vec<_>, _>()`

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
ysaito1001: Thanks for catching. It must've been reminiscent of trial and error. Fixed in 8147d81.

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
ysaito1001: As we discussed, can be done in an additive manner outside of the PR. https://github.com/awslabs/smithy-rs/issues/2993 keeps track of it.

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
ysaito1001: Thanks for the suggestion! Addressed in  `43318e7`.

ysaito1001: Thanks for the suggestion! Addressed in  43318e7.

CHANGELOG.next.toml
```diff
@@ -156,3 +156,15 @@ message = "Source defaults from the default trait instead of implicitly based on
...
+references = ["smithy-rs#2978"]
+meta = { "breaking" = true, "tada" = false, "bug" = false }
+author = "ysaito1001"
+
+[[smithy-rs]]
+message = "The `futures_core::stream::Stream` trait has been removed from public API. [`FnStream`](https://docs.rs/aws-smithy-async/latest/aws_smithy_async/future/pagination_stream/struct.FnStream.html) only supports `next`, `try_next`, and `collect` methods. [`TryFlatMap::flat_map`](https://docs.rs/aws-smithy-async/latest/aws_smithy_async/future/pagination_stream/struct.TryFlatMap.html#method.flat_map) returns [`PaginationStream`](https://docs.rs/aws-smithy-async/latest/aws_smithy_async/future/pagination_stream/struct.PaginationStream.html), which should be preferred to `FnStream` at an interface level. For stream operations previously made available through the trait or its extension traits, we should be able to add them as needed in a backward compatible manner."
+references = ["smithy-rs#2978"]
+meta = { "breaking" = true, "tada" = false, "bug" = false, "target" = "client" }
+author = "ysaito1001"
```
rcoh, line 170: fn_stream was also moved to a new module

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
```diff
@@ -0,0 +1,385 @@
...
+use std::task::{Context, Poll};
+
+pub mod collect;
+
+/// A wrapper around [`FnStream`].
+///
+/// This type provides the same set of methods as [`FnStream`], but that is meant to be used
+/// internally and not by external users.
+#[derive(Debug)]
+pub struct PaginationStream<Item>(FnStream<Item>);
```
rcoh, line 23: can you update these docs to be better focused on the end user? It's probably more relevant what methods they can call, not that it's a wrapper type around another type ;-)

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl<T, E> PaginationStream<Result<T, E>> {
+    /// Yields the next item in the stream or returns an error if an error is encountered.
+    pub async fn try_next(&mut self) -> Result<Option<T>, E> {
+        self.next().await.transpose()
+    }
+
+    /// Convenience method for `.collect::<Result<Vec<_>, _>()`.
+    pub async fn try_collect(self) -> Result<Vec<T>, E> {
```
rcoh, line 49: we _may_ want to make `Vec` a generic here so that you collect other types in the future. I don't feel storngly though.

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
```diff
@@ -0,0 +1,385 @@
...
+
+use crate::future::rendezvous;
+use pin_project_lite::pin_project;
+use std::fmt;
+use std::future::poll_fn;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+pub mod collect;
```
rcoh, line 16: a meta nit, can you put `fn_stream` back in it's own module? that would make this much easier to review

rcoh (changes requested):
Looking good! Just a couple of small tweaks. Nice work on this.

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
ysaito1001: A bit torn on that, but I'd leave `try_collect` generic free for now. If someone wants a different collection type while using `try_collect`, they can use `collect` with a turbo fish instead.

rust-runtime/aws-smithy-async/Cargo.toml
```diff
@@ -14,13 +14,18 @@ test-util = []
...
 
 [dev-dependencies]
+pin-utils = "0.1"
 tokio = { version = "1.23.1", features = ["rt", "macros", "test-util"] }
 tokio-test = "0.4.2"
 
+# futures-util is used by `now_or_later`, for instance, but the tooling
+# reports a false positive, saying it is unused.
+[package.metadata.cargo-udeps.ignore]
+normal = ["futures-util"]
```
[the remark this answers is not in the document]
ysaito1001: I wonder if we're hitting https://github.com/est31/cargo-udeps/issues/65#issuecomment-603999547.

CHANGELOG.next.toml
ysaito1001: Good catch! Addressed in 7571c8c.

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
ysaito1001: Absolutely. Updated docs with an example in b8f37ea.

rust-runtime/aws-smithy-async/src/future/pagination_stream.rs
ysaito1001: Yep, updated in 1add37f.

rcoh (approved):

tools/ci-cdk/canary-lambda/src/release_2023_08_23/s3_canary.rs
```diff
@@ -8,8 +8,8 @@ use crate::{mk_canary, CanaryEnv};
 use anyhow::Context;
 use aws_config::SdkConfig;
 use aws_sdk_s3 as s3;
-use aws_sdk_s3::presigning::config::PresigningConfig;
-use s3::types::ByteStream;
+use s3::presigning::PresigningConfig;
+use s3::primitives::ByteStream;
```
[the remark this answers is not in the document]
ysaito1001: As discussed, this can be done later. Tracked by https://github.com/awslabs/smithy-rs/issues/3008

Veetaha:
I understand the impulse here, but removing `futures::Stream` implementation is almost akin to removing `Iterator` implementation.

What we have now is that paginators are a stripped down `futures::Stream` which has no combinators that it previously had. It means we have to write our own wrappers around every single AWS SDK paginator that maps it into a `futures::Stream`. That's bad user experience, and I think you overdid this

Issue #1767, opened 2022-09-23
Support for IDLv2 nullability semantics


sugmanue:
In order to be able to take full advantage of the new nullability semantics in IDLv2 the clients needs to have a mechanism to keep track whether or not a member with a default value has been updated, quoting from the [Defaults and Model Evolution](https://github.com/awslabs/smithy/blob/main/designs/defaults-and-model-evolution.md#default-value-serialization) document

> To allow servers to change default values if necessary, clients SHOULD NOT serialize default values unless the member is explicitly set to the default value. This implies that clients SHOULD implement a kind of "presence tracking" of defaulted members.

jdisanti:
> I believe this is being tracked in another issue, and it is certainly on the roadmap to match v2 semantics in their entirety before a 1.0 release.

Yeah, there is a SDK issue for it: https://github.com/awslabs/aws-sdk-rust/issues/536

I'm going to leave this open to track implementation in smithy-rs.

jdisanti:
I think we need to make some code generator changes to `StructureGenerator` and `BuilderGenerator` to fully support the new nullability semantics. I generated a new SDK with the latest IDLv2 SDK models and manually inspected the codegen diff looking for changes in optionality, and there was none.

I wrote a quick test in `StructureGeneratorTest` which fails:
```kotlin
    @Test
    fun `IDLv2 client @required fields are not options`() {
        val model = """
            namespace com.test
            structure MyStruct {
               @required
               foo: String,
               @required
               baz: Integer,
            }
        """.asSmithyModel(smithyVersion = "2.0")
        val struct = model.lookup<StructureShape>("com.test#MyStruct")

        val provider = testSymbolProvider(model)
        val project = TestWorkspace.testProject(provider)

        project.useShapeWriter(inner) { writer ->
            writer.withModule("model") {
                StructureGenerator(model, provider, writer, struct).render()
            }

            writer.rustBlock("fn check_optionality(s: &MyStruct)") {
                rust(
                    """
                    let _: &String = &s.foo;
                    let _: i64 = &s.baz;
                    """,
                )
            }
        }
        project.compileAndTest()
    }
```
The `SymbolVisitor` is coded up to use Smithy's `NullabilityIndex` to correctly determine if a shape should be optional or not, but it seems like something is getting lost in translation between that decision and the actual code generation for it.

Velfi:
We're still waiting on some internal developments before we can begin work on this issue.

jnicholls:
[...]

PR #2982, opened 2023-09-12
Use the default(..) trait to source default information

rcoh:
## Motivation and Context
Because of Smithy IDL v1 semantics, we were incorrectly providing generalized default values. This isn't correct in Smithy IDLv2 where the `default` trait controls default behavior.

## Description
- Source default information from the default trait instead of automatically providing it for all primitive shapes.
- Add "error_correcting_builder" when required which effectively adds back the old lax-defaults but only for deserialization

## Testing
- unit tests

## Checklist

----


ysaito1001 (commented):
The change looks good.

> Because of Smithy IDL v1 semantics, we were incorrectly providing generalized default values

Is there any test that is broken because of the reason above and is now fixed with this PR?

rcoh:
> The change looks good.
> 
> > Because of Smithy IDL v1 semantics, we were incorrectly providing generalized default values
> 
> Is there any test that is broken because of the reason above and is now fixed with this PR?

No, I wouldn't expect there to be—this change is only load-bearing in combination with #1767

PR #2983, opened 2023-09-12
Remove `futures_core::stream::Stream` from `aws_smithy_http::byte_stream::ByteStream`

ysaito1001:
## Motivation and Context
Removes `futures_core::stream::Stream` from `aws_smithy_http::byte_stream::ByteStream`

## Description
This PR is part of our ongoing effort, https://github.com/awslabs/smithy-rs/issues/2413. We remove the `futures_core::stream::Stream` trait from `aws_smithy_http::byte_stream::ByteStream`. To continue support existing places that relied upon `ByteStream` implementing the `Stream` trait, we provide explicit `.next`, `.try_next`, and `.size_hints` methods on that type. As part of it, a doc-hidden type `FuturesStreamCompatByteStream`, which implements `Stream`, has been added to `aws_smithy_http` to cater to those who need a type implementing `Stream` (see [discussion](https://github.com/awslabs/smithy-rs/pull/2910#discussion_r1317112233) why doc-hidden).

Another place we need to update is codegen responsible for rendering stream payload serializer, and this affects the server. The regular server and the python server have slightly different rendering requirements, since the former uses `aws_smithy_http::byte_stream::ByteStream` (that does _not_ implement the `Stream` trait) and latter uses its own [ByteStream](https://github.com/awslabs/smithy-rs/blob/cb79a68e3c38d1e62d3980d5e7baedc1144bacc7/rust-runtime/aws-smithy-http-server-python/src/types.rs#L343) (that does implement `Stream`). We use `ServerHttpBoundProtocolCustomization` to handle the difference: `StreamPayloadSerializerCustomization` and `PythonServerStreamPayloadSerializerCustomization`.

## Testing
No new behavior has been added, relied on the existing tests in CI.

## Checklist

----


rcoh (commented):
seems pretty reasonable. I might consider putting the `impl Stream` bridge in the types-convert crate instead so customers can easily use it if they want

ysaito1001:
> seems pretty reasonable. I might consider putting the `impl Stream` bridge in the types-convert crate instead so customers can easily use it if they want

The `impl Stream` bridge is currently done for `FuturesStreamCompatByteStream` and it's meant to be used directly by codegen to pass `ByteStream` (in the case of python-server, it's own `aws_smithy_http_server_python::types::ByteStream`) to `hyper::body::Body::wrap_stream` for stream payload serializers.

For external customers' use cases for `aws_smithy_http::ByteStream`, we may not need the `impl Stream` bridge, i.e. `next`, `try_next`, `collect`, and `into_async_read` should do?

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/protocols/PythonServerProtocolLoader.kt
```diff
@@ -61,6 +61,29 @@ class PythonServerAfterDeserializedMemberServerHttpBoundCustomization() :
...
+    override fun section(section: ServerHttpBoundProtocolSection): Writable = when (section) {
+        is ServerHttpBoundProtocolSection.TypeOfSerializedStreamPayload -> writable {
+            // `aws_smithy_http_server_python::types::ByteStream` already implements
+            // `futures::stream::Stream`, so no need to wrap it in a futures' stream-compatible
+            // wrapper.
+            rust("#T", section.params.symbolProvider.toSymbol(section.params.member))
+        }
+
+        is ServerHttpBoundProtocolSection.WrapStreamPayload -> writable {
+            // payloadName is always non-null within WrapStreamAfterPayloadGenerated
```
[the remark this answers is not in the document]
david-perez: If it's always non-null, why is `payloadName` nullable in `StreamPayloadSerializerParams`?

david-perez (approved):

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/protocols/PythonServerProtocolLoader.kt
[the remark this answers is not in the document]
ysaito1001: I refactored in e076f64 so that `payloadName` is not used by `StreamPayloadSerializerParams`. The refactoring is about wrapping stream payload with a `Stream`-trait-enabling new type now takes place at the call sites of a serializer rather than within the serializer itself.

That effectively removed the need for the `StreamPayloadSerializerRenderer` interface, since rendering customization can be hard-coded on the server side (and no customization is needed on the client side).

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -393,7 +417,9 @@ impl ByteStream {
     /// # }
     /// ```
     pub fn into_async_read(self) -> impl tokio::io::AsyncRead {
-        tokio_util::io::StreamReader::new(self)
+        tokio_util::io::StreamReader::new(
+            crate::futures_stream_adapter::FuturesStreamCompatByteStream::new(self),
```
rcoh, line 421: ah I see—this is why we need the bridge. Makes sense. Although this _could_ be pub crate for this use case.

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -524,6 +538,25 @@ impl<B> Inner<B> {
         Self { body }
     }
 
+    async fn next(&mut self) -> Option<Result<Bytes, B::Error>>
+    where
+        Self: Unpin,
```
rcoh, line 543: this `Unpin` bound isn't new right? I guess this is on a internal type so it's fine.

rcoh (approved):
Awesome work on this!! I love how narrow this PR ended up being

rust-runtime/aws-smithy-http/src/byte_stream.rs
ysaito1001: Right, and this needs to be `pub` as opposed to `pub(crate)` because codegen server is also [using it](https://github.com/awslabs/smithy-rs/pull/2983/files#diff-3cb0946710c8ff96daf68b3347899e049ffc8dd77bb307c3d4fa43c691a4c49bR32-R36).

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
```diff
@@ -407,9 +407,10 @@ data class RuntimeType(val path: String, val dependency: RustDependency? = null)
         fun retryErrorKind(runtimeConfig: RuntimeConfig) = smithyTypes(runtimeConfig).resolve("retry::ErrorKind")
         fun eventStreamReceiver(runtimeConfig: RuntimeConfig): RuntimeType =
             smithyHttp(runtimeConfig).resolve("event_stream::Receiver")
-
         fun eventStreamSender(runtimeConfig: RuntimeConfig): RuntimeType =
             smithyHttp(runtimeConfig).resolve("event_stream::EventStreamSender")
+        fun futuresStreamCompatByteStream(runtimeConfig: RuntimeConfig): RuntimeType =
```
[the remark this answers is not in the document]
ysaito1001: Yes, we do as you noticed it [here](https://github.com/awslabs/smithy-rs/pull/2983#discussion_r1337298337).

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -276,14 +275,39 @@ impl ByteStream {
...
+    }
+
+    /// Attempt to pull out the next value of this stream, returning `None` if the stream is
+    /// exhausted.
+    pub fn poll_next(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Bytes, Error>>> {
+        self.project().inner.poll_next(cx).map_err(Error::streaming)
+    }
```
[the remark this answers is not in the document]
ysaito1001: Thanks for catching. No it doesn't anymore. At one point during development, `FuturesStreamCompatByteStream` was in `rust-runtime/inlineable` and needed to access the above `poll_next` (hence was `pub`) but now that the compat struct is in the same crate as `ByteStream`, the above method can be `pub(crate)`.

Addressed in a802799.

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -276,14 +275,39 @@ impl ByteStream {
...
+    /// Consume the `ByteStream`, returning the wrapped SdkBody.
     // Backwards compatibility note: Because SdkBody has a dyn variant,
     // we will always be able to implement this method, even if we stop using
     // SdkBody as the internal representation
     pub fn into_inner(self) -> SdkBody {
         self.inner.body
     }
 
+    /// Return the next item in the `ByteStream`.
+    pub async fn next(&mut self) -> Option<Result<Bytes, Error>> {
```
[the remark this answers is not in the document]
ysaito1001: Linked in 5d07b16.

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -276,14 +275,39 @@ impl ByteStream {
...
+        Some(self.inner.next().await?.map_err(Error::streaming))
+    }
+
+    /// Attempt to pull out the next value of this stream, returning `None` if the stream is
```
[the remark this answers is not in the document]
ysaito1001: Linked in 6fc9a96.

rust-runtime/aws-smithy-http/src/lib.rs
```diff
@@ -27,6 +27,8 @@
 
 pub mod body;
 pub mod endpoint;
+#[doc(hidden)]
```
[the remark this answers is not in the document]
ysaito1001: Added a comment in b01b1d4.

rust-runtime/aws-smithy-http/src/byte_stream.rs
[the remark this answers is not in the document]
ysaito1001: Turns out it's used by the python server's own `aws_smithy_http_server_python::types::ByteStream` that wraps `aws_smithy_http::byte_stream::ByteStream` ([relevant code](https://github.com/awslabs/smithy-rs/blob/bc418284516975461dc0bee44ea999361f5c95be/rust-runtime/aws-smithy-http-server-python/src/types.rs#L352)). Added back `pub` in 5bcf5ca.

rust-runtime/aws-smithy-http/src/byte_stream.rs
```diff
@@ -393,7 +423,9 @@ impl ByteStream {
     /// # }
     /// ```
     pub fn into_async_read(self) -> impl tokio::io::AsyncRead {
```
utkarshgupta137, line 425: Earlier we could do this wrapping ourselves & since `tokio_util::io::StreamReader` implements `AsyncBufRead`, we would get a buffered reader for free. Would it be acceptable to change this to:
```suggestion
    pub fn into_async_read(self) -> impl tokio::io::AsyncBufRead {
```

utkarshgupta137: Or perhaps you would want a separate `into_async_buf_read`?

jdisanti: We don't have any buffering in our implementation. You should be able to wrap the ByteStream with a new-type and implement Stream for it to get the same functionality you used to have. The Stream impl should be a simple call through to the poll_next and size_hint methods.

utkarshgupta137: Yes, I know. What I'm saying is that in this function, you're wrapping bytestream in `StreamReader` from tokio, which implements `AsyncBufRead`. So you could also return `impl AsyncBufRead` from this function. 
Otherwise, one has to wrap this in a `BufReader` for no reason, since it is not possible to wrap `ByteStream` in `StreamReader` easily (without wrapping & implementing Stream manually).

jdisanti: Oh, I see what you're saying. Yeah, I don't see why we couldn't have two methods. You're welcome to submit a PR to add that.

utkarshgupta137: I would say it makes more sense to change the signature of the existing function as I suggested above. It won't be a breaking change at all since AsyncBufRead is a super trait of AsyncRead.

jdisanti: Sounds good.

PR #2985, opened 2023-09-14
Source defaults from the model instead of implicitly

rcoh:
## Motivation and Context
We weren't correctly computing defaults which lead to incorrect behavior when coupled with nullability.

## Description
Minimal changeset to source defaults from the model. Other changes:
- Unify enum parsing across client and server to always use `from_str` in protocol tests
- Extract `PrimitiveInstantiator` from `Instantiator` so it can be used to instantiate defaults

## Testing
- regular codegen tests

## Checklist

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt
@@ -116,7 +116,7 @@ fun Shape.contextName(serviceShape: ServiceShape?): String {
  */
 open class SymbolVisitor(
     settings: CoreRustSettings,
-    override val model: Model,
+    final override val model: Model,
     private val serviceShape: ServiceShape?,
     override val config: RustSymbolProviderConfig,
 ) : RustSymbolProvider, ShapeVisitor<Symbol> {
@@ -272,8 +272,7 @@ open class SymbolVisitor(
             when (val value = trait.toNode()) {
                 Node.from(""), Node.from(0), Node.from(false), Node.arrayNode(), Node.objectNode() -> Default.RustDefault
                 Node.nullNode() -> Default.NoDefault
-                else -> { Default.NonZeroDefault(value)
-                }
+                else -> Default.NonZeroDefault(value)
             }
         } ?: Default.NoDefault
         // Handle boxing first, so we end up with Option<Box<_>>, not Box<Option<_>>.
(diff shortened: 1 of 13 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -131,64 +128,7 @@ open class Instantiator(
...
-                    """
-                    let json_bytes = br##"${Node.prettyPrintJson(data)}"##;
-                    let mut tokens = #{json_token_iter}(json_bytes).peekable();
-                    #{expect_document}(&mut tokens).expect("well formed json")
-                    """,
-                    "expect_document" to smithyJson.resolve("deserialize::token::expect_document"),
-                    "json_token_iter" to smithyJson.resolve("deserialize::json_token_iter"),
-                )
-            }
+            is SimpleShape -> PrimitiveInstantiator(runtimeConfig, symbolProvider).instantiate(shape, data)(writer)
```
ysaito1001, line 131: Great `SimpleShape` reduces the multiple branches!

ysaito1001 (approved):
Thanks for adding a test to show what's supported now.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -416,3 +357,77 @@ open class Instantiator(
...
+                )
+            }
+        }
+    }
+
+    private fun renderString(shape: StringShape, arg: StringNode): Writable = {
+        val data = escape(arg.value).dq()
+        if (shape.hasTrait<EnumTrait>() || shape is EnumShape) {
+            val enumSymbol = symbolProvider.toSymbol(shape)
+            rust("$data.parse::<#T>().unwrap()", enumSymbol)
```
david-perez, line 428:
```suggestion
            rust("$data.parse::<#T>().expect("this is only used in tests")", enumSymbol)
```

david-perez (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/SymbolVisitor.kt
```diff
@@ -263,13 +268,21 @@ open class SymbolVisitor(
 
     override fun memberShape(shape: MemberShape): Symbol {
         val target = model.expectShape(shape.target)
+        val defaultValue = shape.getMemberTrait(model, DefaultTrait::class.java).orNull()?.let { trait ->
+            when (val value = trait.toNode()) {
```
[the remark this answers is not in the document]
rcoh: - triggers clippy lint
- causes a huge codegen diff
- non-zero defaults are only allowed on simple shapes so it enables some simplification of generating defaults

PR #2986, opened 2023-09-14
Filter test inputs for presigning header exclusion proptest

ysaito1001:
## Motivation and Context
During our internal build, the `aws-sigv4` crate failed on a proptest as follows:
```
Diff < left / right > :
<content-length;content-type;host
>content-length;content-type

.
minimal failing input: excluded_headers = [
    "host",
]
        successes: 74
        local rejects: 0
        global rejects: 0
', aws-sigv4/src/http_request/canonical_request.rs:838:5

failures:
    http_request::canonical_request::tests::presigning_header_exclusion_with_explicit_exclusion_list_specified
```
This says that `excluded_headers` should not contain `host`.

To address it, this PR will filter out test inputs from proptest that contain `content-length`, `content-type`, or `host` (all of which appear in the expected value for the test at line 886 in the new revision).

## Testing
No new tests have been added, relied on the existing tests in CI

----

### DIFF
--- a/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
+++ b/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
@@ -835,11 +835,24 @@ mod tests {
         );
     }
 
+    fn valid_input(input: &Vec<String>) -> bool {
+        [
+            "content-length".to_owned(),
+            "content-type".to_owned(),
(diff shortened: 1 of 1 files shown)


rcoh (approved):

rcoh:
merged—we need to merge 0.56 back into main

PR #1725, opened 2022-09-09
Add support for IDL 2.0 nullability semantics

sugmanue:
## Motivation and Context
Changed the code to properly handle the nullability semantics for both, 1.0 and 2.0 models.

## Description
* Removed the use of the deprecated method `isNullable` and instead used the `isMemberNullable` with `CheckMode.CLIENT_ZERO_VALUE_V1_NO_INPUT` which will make the semantics equivalent to looking for a `Box` trait which will work for both 1.0 and 2.0 models.
* Changed a unit test which now we need to specify is a 1.0 model in order for the loader to properly work 
* Changed the EC2 customization to add `ClientOptionalTrait` instead of adding a `Box` trait using a transformer, this will make will have the same semantic implications that the previous code.
* Finally added a bunch of references to `mavenLocal` which was needed to test the changes.

## Testing

* Generated all the models using `aws-sdk-rust` and validated that there were no changes in the generated code.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2988, opened 2023-09-15
Extract builderInstantiator interface to prepare for nullability changes

rcoh:
## Motivation and Context
#1725 exposed the need for easily configuring builder behavior between client & server

## Description
- extract builderGenerator interface and attach it to `codegenContext` to avoid loads of threading it up and down

## Testing
- codegen unit tests
- [x] codegen diff audit

## Checklist

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderInstantiator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderInstantiator.kt
@@ -11,7 +11,7 @@ import software.amazon.smithy.rust.codegen.core.rustlang.Writable
 import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
 import software.amazon.smithy.rust.codegen.core.rustlang.writable
 
-/** Abstraction for instantiating a builders.
+/** Abstraction for instantiating builders.
  *
  * Builder abstractions vary—clients MAY use `build_with_error_correction`, e.g., and builders can vary in fallibility.
  * */
@@ -19,11 +19,14 @@ interface BuilderInstantiator {
     /** Set a field on a builder. */
     fun setField(builder: String, value: Writable, field: MemberShape): Writable
 
-    /** Finalize a builder, turning into a built object (or in the case of builders-of-builders, return the builder directly).*/
+    /** Finalize a builder, turning it into a built object
+     *  - In the case of builders-of-builders, the value should be returned directly
+     *  - If an error is returned, you MUST use `mapErr` to convert the error type
+     */
     fun finalizeBuilder(builder: String, shape: StructureShape, mapErr: Writable? = null): Writable
 
     /** Set a field on a builder using the `$setterName` method. $value will be passed directly. */
-    fun setFieldBase(builder: String, value: Writable, field: MemberShape) = writable {
+    fun setFieldWithSetter(builder: String, value: Writable, field: MemberShape) = writable {
         rustTemplate("$builder = $builder.${field.setterName()}(#{value})", "value" to value)
     }
 }
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderInstantiator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderInstantiator.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.rust.codegen.core.smithy.generators
+
+import software.amazon.smithy.model.shapes.MemberShape
+import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.rust.codegen.core.rustlang.Writable
+import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.core.rustlang.writable
+
+/** Abstraction for instantiating a builders.
+ *
+ * Builder abstractions vary—clients MAY use `build_with_error_correction`, e.g., and builders can vary in fallibility.
+ * */
+interface BuilderInstantiator {
+    /** Set a field on a builder. */
+    fun setField(builder: String, value: Writable, field: MemberShape): Writable
+
+    /** Finalize a builder, turning into a built object (or in the case of builders-of-builders, return the builder directly).*/
+    fun finalizeBuilder(builder: String, shape: StructureShape, mapErr: Writable? = null): Writable
+
+    /** Set a field on a builder using the `$setterName` method. $value will be passed directly. */
+    fun setFieldBase(builder: String, value: Writable, field: MemberShape) = writable {
+        rustTemplate("$builder = $builder.${field.setterName()}(#{value})", "value" to value)
+    }
+}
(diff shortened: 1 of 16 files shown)


rcoh:
audited codegen, no diff

rcoh:
no changelog, internal only change

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderInstantiator.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package software.amazon.smithy.rust.codegen.core.smithy.generators
+
+import software.amazon.smithy.model.shapes.MemberShape
+import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.rust.codegen.core.rustlang.Writable
+import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.core.rustlang.writable
+
+/** Abstraction for instantiating a builders.
```
david-perez, line 14:
```suggestion
/** 
 * Abstraction for instantiating builders.
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt
```diff
@@ -339,6 +340,7 @@ class EventStreamUnmarshallerGenerator(
                     // TODO(EventStream): Errors on the operation can be disjoint with errors in the union,
                     //  so we need to generate a new top-level Error type for each event stream union.
                     when (codegenTarget) {
+                        // TODO(https://github.com/awslabs/smithy-rs/issues/1970) It should be possible to unify these branches now
```
david-perez, line 343: Why should it be possible?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt
```diff
@@ -352,9 +354,19 @@ class EventStreamUnmarshallerGenerator(
                                         })?;
                                     builder.set_meta(Some(generic));
```
david-perez, line 355: Aside: not sure why we bind the error metadata to a variable named `generic`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderInstantiator.kt
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Builder abstractions vary—clients MAY use `build_with_error_correction`, e.g., and builders can vary in fallibility.
+ * */
+interface BuilderInstantiator {
+    /** Set a field on a builder. */
+    fun setField(builder: String, value: Writable, field: MemberShape): Writable
+
+    /** Finalize a builder, turning into a built object (or in the case of builders-of-builders, return the builder directly).*/
```
david-perez, line 22:
```suggestion
    /** Finalize a builder, turning it into a built object (or in the case of builders-of-builders, return the builder directly).*/
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderInstantiator.kt
```diff
@@ -0,0 +1,29 @@
...
+    fun finalizeBuilder(builder: String, shape: StructureShape, mapErr: Writable? = null): Writable
```
david-perez, line 23: Worth documenting `mapErr`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
```diff
@@ -508,12 +513,14 @@ class JsonParserGenerator(
...
-                    // Only call `build()` if the builder is not fallible. Otherwise, return the builder.
-                    if (returnSymbolToParse.isUnconstrained) {
-                        rust("Ok(Some(builder))")
-                    } else {
-                        rust("Ok(Some(builder.build()))")
+                    val builder = builderInstantiator.finalizeBuilder(
+                        "builder", shape,
+                    ) {
+                        rustTemplate(
+                            """|err|#{Error}::custom_source("Response was invalid", err)""", *codegenScope,
```
david-perez, line 520: This doesn't look correct to me. This only applies to clients, not servers. It only works because `ServerBuilderInstantiator` does not rely on the passed in `mapErr` writable.

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/TestBuilderInstantiator.kt
```diff
@@ -0,0 +1,6 @@
+/*
```
david-perez, line 1: I don't understand what this file's purpose is.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiator.kt
```diff
@@ -86,3 +91,38 @@ fun serverInstantiator(codegenContext: CodegenContext) =
         defaultsForRequiredFields = true,
         customizations = listOf(ServerAfterInstantiatingValueConstrainItIfNecessary(codegenContext)),
     )
+
+class ServerBuilderInstantiator(
+    private val symbolProvider: RustSymbolProvider,
+    private val symbolParseFn: (Shape) -> ReturnSymbolToParse,
+) :
+    BuilderInstantiator {
```
david-perez, line 99:
```suggestion
) : BuilderInstantiator {
```

The current formatting looks odd.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderInstantiator.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /** Set a field on a builder using the `$setterName` method. $value will be passed directly. */
+    fun setFieldBase(builder: String, value: Writable, field: MemberShape) = writable {
```
david-perez, line 26: nit:

```suggestion
    fun setFieldWithSetter(builder: String, value: Writable, field: MemberShape) = writable {
```

Base denotes we're doing inheritance. Which we kind of are doing manually in the implementations of this interface.

david-perez (approved):
The only blocking comment is the one in `JsonParserGenerator.kt`, which should be artifact agnostic, but is currently passing in a value to `map_err` that only applies to client generation.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt
rcoh: mapErr defines how to convert the error back into the caller error—since the server never returns an error, it doesn't need to use it.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt
rcoh: because we have the builderInstantiator, I think. But it's possible this is more divergence that can't be handled like that.

PR #2989, opened 2023-09-18
Omit fractional seconds from `http-date` format

### LINKED ISSUE #2831: Update `aws_smithy_types::date_time::Format::HttpDate` to omit fractional seconds
Per https://github.com/awslabs/smithy/pull/1847, we need to update [`aws_smithy_types::date_time::Format::HttpDate`](https://docs.rs/aws-smithy-types/latest/aws_smithy_types/date_time/enum.Format.html#variant.HttpDate) to omit fractional seconds.

rschmitt:
## Motivation and Context
Closes #2831.

## Description
Fractional seconds will still be accepted during parsing, but not emitted during serialization.

## Testing
```
./gradlew codegen-core:check codegen-client:check codegen-server:check codegen-client-test:check codegen-server-test:check
```

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -156,3 +156,9 @@ message = "Omit fractional seconds from `date-time` format."
 references = ["smithy-rs#2831", "aws-sdk-rust#818"]
 meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "all" }
 author = "rschmitt"
+
+[[smithy-rs]]
+message = "Source defaults from the default trait instead of implicitly based on type. This has minimal changes in the generated code."
+references = ["smithy-rs#2985"]
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
+author = "rcoh"
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -150,3 +150,9 @@ message = "Fix regression with redacting sensitive HTTP response bodies."
 references = ["smithy-rs#2926", "smithy-rs#2972"]
 meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
 author = "ysaito1001"
+
+[[smithy-rs]]
+message = "Omit fractional seconds from `date-time` format."
+references = ["smithy-rs#2831"]
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "all" }
+author = "rschmitt"
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -35,6 +35,21 @@ meta = { "breaking" = true, "tada" = false, "bug" = false }
 author = "Velfi"
 
 [[aws-sdk-rust]]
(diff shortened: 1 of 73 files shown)


rcoh:
doesn't need to be in this PR, but we should also add an [S3-specific protocol test](
https://github.com/awslabs/smithy-rs/blob/main/aws/sdk/aws-models/s3-tests.smithy) for https://github.com/awslabs/aws-sdk-rust/issues/818

Can you also link to aws-sdk-rust#818 from the changelog?

It would be good to also verify that that API actually works after this change.

We should also contribute a protocol test upstream that mandates this behavior—currently there are non that I'm aware of.

rcoh (approved):

jdisanti (commented):
I think you have the wrong date format. This change should be applied to `http-date` rather than `date-time`. We definitely want fractional seconds on date-time.

CHANGELOG.next.toml
```diff
@@ -150,3 +150,9 @@ message = "Fix regression with redacting sensitive HTTP response bodies."
 references = ["smithy-rs#2926", "smithy-rs#2972"]
 meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
 author = "ysaito1001"
+
+[[smithy-rs]]
+message = "Omit fractional seconds from `date-time` format."
```
rcoh, line 155:
```suggestion
message = "Omit fractional seconds from `http-date` format."
```

rust-runtime/aws-smithy-json/src/serialize.rs
```diff
@@ -309,7 +309,7 @@ mod tests {
         object.finish();
 
         assert_eq!(
-            r#"{"epoch_seconds":5.2,"date_time":"2021-05-24T15:34:50.123Z","http_date":"Wed, 21 Oct 2015 07:28:00 GMT"}"#,
+            r#"{"epoch_seconds":5.2,"date_time":"2021-05-24T15:34:50Z","http_date":"Wed, 21 Oct 2015 07:28:00 GMT"}"#,
```
rcoh, line 312: We should change this test to put a fraction in http_date—and I think we need to back out the changes to date_time, only http_date omits the subsecond. My bad on that—good catch from @jdisanti

rust-runtime/aws-smithy-types/src/date_time/format.rs
```diff
@@ -479,26 +466,9 @@ pub(crate) mod rfc3339 {
...
-            into.push('.');
-            let (mut remaining, mut place) = (micros, 100_000);
-            while remaining > 0 {
-                let digit = (remaining / place) % 10;
-                into.push(char::from(b'0' + (digit as u8)));
-                remaining -= digit * place;
-                place /= 10;
-            }
-        }
-    }
```
rcoh, line 501: We need to bring this back—sorry about that

rcoh (changes requested):
Wow, good catch all on the HTTP Date, glad we actually tested the end-bug. I think we need to back out the changes to the RFC3339 serializer than we're g2g

PR #2992, opened 2023-09-19
Move endpoint resolution into the same phase as signing

jdisanti:
The endpoint resolution and signing steps need to be combined in the near future. This PR moves endpoint resolution so that combining it with signing in the future won't break interceptors.

----


ysaito1001 (approved):
Looks good, as it says on the tin.

jdisanti:
This seems like it is the wrong approach since it will make it impossible to see the endpoint prior to signing. What really needs to happen is `orchestrate_auth` needs to be broken into two separate steps so that the resolution of the auth scheme happens at the same time as endpoint resolution, but signing happens later. Thus, the current code is actually correct rather than this update. Closing this.

PR #2995, opened 2023-09-21
Produce `&[T]` instead of `Option<&[T]>` for list fields

rcoh:
## Motivation and Context
When the returned field is a list, there is _almost_ never a reason to know if the returned value was null or `[]` — this addresses the 99% case.

Before:
```rust
fn blah(&self) -> Option<&[T]> { self.blah.as_deref() }
```

After:
```rust
fn blah(&self) -> &[T] { self.blah.as_deref().unwrap_or_default() }
```

**note**: no impact on servers by default, see codegen diff.

## Description
Update accessors for lists.

## Testing
- [x] codegen diff audit (no server changes)
- [x] unit tests

## Checklist

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreRustSettings.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreRustSettings.kt
@@ -41,16 +41,19 @@ const val CODEGEN_SETTINGS = "codegen"
 open class CoreCodegenConfig(
     open val formatTimeoutSeconds: Int = defaultFormatTimeoutSeconds,
     open val debugMode: Boolean = defaultDebugMode,
+    open val flattenCollectionAccessors: Boolean = defaultFlattenMode,
 ) {
     companion object {
         const val defaultFormatTimeoutSeconds = 20
         const val defaultDebugMode = false
+        const val defaultFlattenMode = false
 
         fun fromNode(node: Optional<ObjectNode>): CoreCodegenConfig =
             if (node.isPresent) {
                 CoreCodegenConfig(
                     formatTimeoutSeconds = node.get().getNumberMemberOrDefault("formatTimeoutSeconds", defaultFormatTimeoutSeconds).toInt(),
                     debugMode = node.get().getBooleanMemberOrDefault("debugMode", defaultDebugMode),
+                    flattenCollectionAccessors = node.get().getBooleanMemberOrDefault("flattenCollectionAccessors", defaultFlattenMode),
                 )
             } else {
                 CoreCodegenConfig(
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
@@ -37,6 +37,7 @@ import software.amazon.smithy.rust.codegen.core.smithy.RustSymbolProviderConfig
 import software.amazon.smithy.rust.codegen.core.smithy.SymbolVisitor
 import software.amazon.smithy.rust.codegen.core.smithy.generators.BuilderGenerator
 import software.amazon.smithy.rust.codegen.core.smithy.generators.BuilderInstantiator
+import software.amazon.smithy.rust.codegen.core.smithy.generators.StructSettings
 import software.amazon.smithy.rust.codegen.core.smithy.generators.StructureGenerator
 import software.amazon.smithy.rust.codegen.core.smithy.module
 import software.amazon.smithy.rust.codegen.core.smithy.traits.SyntheticInputTrait
@@ -194,7 +195,7 @@ fun StructureShape.renderWithModelBuilder(
 ) {
     val struct = this
     rustCrate.withModule(symbolProvider.moduleForShape(struct)) {
-        StructureGenerator(model, symbolProvider, this, struct, emptyList()).render()
+        StructureGenerator(model, symbolProvider, this, struct, emptyList(), StructSettings(true)).render()
         implBlock(symbolProvider.toSymbol(struct)) {
             BuilderGenerator.renderConvenienceMethod(this, symbolProvider, struct)
         }
--- a/codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGeneratorTest.kt
+++ b/codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGeneratorTest.kt
@@ -7,14 +7,17 @@ package software.amazon.smithy.rust.codegen.core.smithy.generators
 
 import org.junit.jupiter.api.Test
 import software.amazon.smithy.codegen.core.Symbol
+import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.knowledge.NullableIndex
 import software.amazon.smithy.model.shapes.Shape
 import software.amazon.smithy.model.shapes.StructureShape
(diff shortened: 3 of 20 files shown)


Velfi (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreRustSettings.kt
```diff
@@ -41,16 +41,19 @@ const val CODEGEN_SETTINGS = "codegen"
...
         const val defaultFormatTimeoutSeconds = 20
         const val defaultDebugMode = false
+        const val defaultFlattenMode = false
 
         fun fromNode(node: Optional<ObjectNode>): CoreCodegenConfig =
             if (node.isPresent) {
                 CoreCodegenConfig(
                     formatTimeoutSeconds = node.get().getNumberMemberOrDefault("formatTimeoutSeconds", defaultFormatTimeoutSeconds).toInt(),
                     debugMode = node.get().getBooleanMemberOrDefault("debugMode", defaultDebugMode),
+                    flattenCollectionAccessors = node.get().getBooleanMemberOrDefault("flattenCollectionAccessors", defaultFlattenMode),
```
drganjoo, line 56: Will this change allow someone on the server side to override how we generate accessor methods for collection types by setting "flattenCollectionAccessor" in smithy-build.json? 

If the user cannot set this in the settings, then disregard the following:

Currently, on the server side that behavior is controlled by whether the field is marked as `@required` in the model. 

This approach is fine for optional server inputs, where not sending a value for a collection (resulting in None) and sending an empty collection are treated the same by the business logic. However, there are scenarios where the server's output could make a meaningful distinction between an empty list and a None list. Additionally, using `$memberName.is_none()` on the client side to check whether the server sent an output will be unreliable if the server sends an empty list instead of None.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
```diff
@@ -294,6 +294,7 @@ open class ServerCodegenVisitor(
                 this,
                 shape,
                 codegenDecorator.structureCustomizations(codegenContext, emptyList()),
+                structSettings = codegenContext.structSettings(),
```
drganjoo, line 297: If there is no chance for the user to set this flag on the server side then this is okay. Otherwise, I think let's set this to:  `StructSettings(flattenVecAccessors = false)` for now.

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerStructureGenerator.kt
```diff
@@ -40,7 +40,7 @@ class PythonServerStructureGenerator(
     private val codegenContext: ServerCodegenContext,
     private val writer: RustWriter,
     private val shape: StructureShape,
-) : StructureGenerator(model, codegenContext.symbolProvider, writer, shape, emptyList()) {
+) : StructureGenerator(model, codegenContext.symbolProvider, writer, shape, emptyList(), codegenContext.structSettings()) {
```
drganjoo, line 43: If there is no chance for the user to set this flag on the server side then this is okay. Otherwise, I think let's set this to:  `StructSettings(flattenVecAccessors = false)` for now.

rcoh: a customer can set this on the server side with the smithy-build.json setting

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreRustSettings.kt
rcoh: this is only the accessor method—it has absolutely no impact on serialization, it's just for convenience when reading the field.

> Additionally, using $memberName.is_none() on the client side to check whether the server sent an output will be unreliable if the server sends an empty list instead of None.

The only reason to do that is explicitly if you are trying to tell the difference.

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGeneratorTest.kt
```diff
@@ -94,12 +97,14 @@ internal class BuilderGeneratorTest {
         project.compileAndTest()
     }
 
+    private fun generator(model: Model, provider: RustSymbolProvider, writer: RustWriter, shape: StructureShape) = StructureGenerator(model, provider, writer, shape, emptyList(), StructSettings(flattenVecAccessors = true))
```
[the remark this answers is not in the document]
rcoh: there are existing tests in StructureGeneratorTest testing the old behavior

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreRustSettings.kt
```diff
@@ -41,16 +41,19 @@ const val CODEGEN_SETTINGS = "codegen"
 open class CoreCodegenConfig(
     open val formatTimeoutSeconds: Int = defaultFormatTimeoutSeconds,
     open val debugMode: Boolean = defaultDebugMode,
+    open val flattenCollectionAccessors: Boolean = defaultFlattenMode,
```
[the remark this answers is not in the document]
rcoh: there is a relationship but I wanted default true for clients and default false for servers

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
```diff
@@ -194,7 +195,7 @@ fun StructureShape.renderWithModelBuilder(
 ) {
     val struct = this
     rustCrate.withModule(symbolProvider.moduleForShape(struct)) {
-        StructureGenerator(model, symbolProvider, this, struct, emptyList()).render()
+        StructureGenerator(model, symbolProvider, this, struct, emptyList(), StructSettings(true)).render()
```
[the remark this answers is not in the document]
rcoh: actually no! let me add one, good catch. (Although _all_ existing server tests are testing this)

codegen-server/python/src/main/kotlin/software/amazon/smithy/rust/codegen/server/python/smithy/generators/PythonServerStructureGenerator.kt
drganjoo: Keep it like it is. I wrongly thought that the underlying field is also going to get changed. Only the accessor gets changed, so this is good.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CoreRustSettings.kt
drganjoo: Oh yes! only the accessor gets changed. The underlying serialization format and the field remains the same.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
drganjoo: Disregard. Only the accessor is getting changed so let's keep it the way you have it right now.

drganjoo (approved):

PR #3001, opened 2023-09-26
Add injection hook for structures in no service

82marbag:
## Motivation and Context
This change will allow to render structures in no service closures with a decorator.

## Checklist

----

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
@@ -621,6 +621,7 @@ open class ServerCodegenVisitor(
      *  - Operations ser/de
      *  - Errors via `ServerOperationErrorGenerator`
      *  - OperationShapes via `ServerOperationGenerator`
+     *  - Additional structure shapes via `postprocessGenerateAdditionalStructures`
      */
     override fun operationShape(shape: OperationShape) {
         // Generate errors.
@@ -637,6 +638,9 @@ open class ServerCodegenVisitor(
         rustCrate.withModule(ServerRustModule.Operation) {
             protocolGenerator.renderOperation(this, shape)
         }
+
+        codegenDecorator.postprocessGenerateAdditionalStructures(shape)
+            .forEach { structureShape -> this.structureShape(structureShape) }
     }
 
     override fun blobShape(shape: BlobShape) {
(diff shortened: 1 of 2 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customize/ServerCodegenDecorator.kt
```diff
@@ -31,14 +33,20 @@ interface ServerCodegenDecorator : CoreCodegenDecorator<ServerCodegenContext, Se
      * constrained but the `ValidationException` shape is not attached to the operation's errors.
      */
     fun postprocessValidationExceptionNotAttachedErrorMessage(validationResult: ValidationResult) = validationResult
+
+    /**
+     * Injection point to allow a decorator to point to structures not in the service closure
+     * for which to generate Rust.
+     */
+    fun postprocessGenerateAdditionalStructures(operationShape: OperationShape): List<StructureShape> = listOf()
```
david-perez, line 41:
```suggestion
    fun postprocessGenerateAdditionalStructures(operationShape: OperationShape): List<StructureShape> = emptyList()
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt
```diff
@@ -637,6 +638,10 @@ open class ServerCodegenVisitor(
         rustCrate.withModule(ServerRustModule.Operation) {
             protocolGenerator.renderOperation(this, shape)
         }
+
+        // Postprocessing hook: generate structures not in the service closure.
```
david-perez, line 642:
```suggestion
```

The hook allows for arbitrary structure shapes to be generated, they don't have to necessarily be outside of the service's closure. I've added a documentation suggestion to the decorator's method explaining why that would be bad.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customize/ServerCodegenDecorator.kt
```diff
@@ -31,14 +33,20 @@ interface ServerCodegenDecorator : CoreCodegenDecorator<ServerCodegenContext, Se
      * constrained but the `ValidationException` shape is not attached to the operation's errors.
      */
     fun postprocessValidationExceptionNotAttachedErrorMessage(validationResult: ValidationResult) = validationResult
+
+    /**
+     * Injection point to allow a decorator to point to structures not in the service closure
```
david-perez, line 38:
```suggestion
     * For each operation in the service's closure, this hook allows decorators to return a collection of structure shapes that will additionally be generated.
     * If a structure shape is in the service's closure, note that returning it here will cause for it to be generated more than once, making the resulting crate not compile (since it will contain more than one struct with the same name). Therefore, ensure that all the structure shapes returned by this method are not in the service's closure.
```

david-perez (approved):

PR #3007, opened 2023-09-27
fix use_fips in provider config

Velfi:
I'm not 100% that I fixed this in the right way. Feel free to set me straight if that's the case.

## Motivation and Context

aws-sdk-rust#882

## Description

This change causes the`ProviderConfig` to respect both `use_fips` and `use_dual_stack` when those settings are configured in a user's environment or profile.

## Testing

I wrote two tests

## Checklist

----


CHANGELOG.next.toml
```diff
@@ -183,3 +183,9 @@ message = "Produce better docs when items are marked @required"
 references = ["smithy-rs#2996"]
 meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client" }
 author = "rcoh"
+
+[[smithy-rs]]
```
jdisanti, line 187:
```suggestion
[[aws-sdk-rust]]
```

aws/rust-runtime/aws-config/src/provider_config.rs
```diff
@@ -285,6 +320,25 @@ impl ProviderConfig {
...
+
+    /// Use the [default use_dual_stack provider](crate::default_provider::use_dual_stack::use_dual_stack_provider) to set the
+    /// `use_dual_stack` setting for this configuration
+    ///
+    /// Note: the `env` and `fs` already set on this provider will be used when loading the default region.
+    pub async fn load_default_use_dual_stack(self) -> Self {
+        let use_dual_stack =
+            crate::default_provider::use_dual_stack::use_dual_stack_provider(&self).await;
+        self.with_use_dual_stack(use_dual_stack)
+    }
```
rcoh, line 340: I think doing it this way increases the chances for bugs a little bit. In an ideal world we would:
- make the use_fips/use_dualstack accessors async and have them fallback to the default providers
- call provider_config.use_dual_stacl() methods from the STS constructor. We also need to audit

But that's not going to work well in practice because current callers of `sts_client_config()` don't have the ability to make async calls. So instead, here's what I'm thinking:
Tenets:
1. Customers should never need to call these methods, it should always happen automatically. I think this is what people would expect.
2. It should be hard/impossible to write code that fails to load these values.

Add something like:
```rust
struct ProviderConfig { initialized: bool, ... }

async fn initialize_dependent_fields(&mut self) {
  if self.region == None {
    self.region = region_provider.await;
  }   
  if self.use_dual_stack == None {
    self.use_dual_stack = crate::default_provider::use_dual_stack::use_dual_stack_provider(&self).await;
  }
  initialized = true;
  // etc.
}

// debug asserts so that in tests, we can ensure that all code paths call the initializer.
pub fn use_dual_stack(&self) { if self.use_dual_stack.is_none() { debug_assert!(self.initialized, "must call initialize default fields"); } }
```

Then we call this method in the following places:
- `async fn load()`
- Anywhere else that initializes an AWS client from provider config directly. (e.g. SSO?)

One extremely bulletproof way to do this would be to add a generic to provider config, `Initialized` and only accept an initialized one in these places? Food for though.

rcoh:
main comment is here: https://github.com/awslabs/smithy-rs/pull/3007/files#r1340126880

rcoh:
in either case, I think we also need to tests the case where STS is configured in the other way, via the profile.

aws/rust-runtime/aws-config/src/provider_config.rs
```diff
@@ -285,6 +320,25 @@ impl ProviderConfig {
...
     }
 
+    /// Use the [default use_fips provider](crate::default_provider::use_fips::use_fips_provider) to set the
+    /// `use_fips` setting for this configuration
+    ///
+    /// Note: the `env` and `fs` already set on this provider will be used when loading the default region.
+    pub async fn load_default_use_fips(self) -> Self {
+        let use_fips = crate::default_provider::use_fips::use_fips_provider(&self).await;
+        self.with_use_fips(use_fips)
+    }
```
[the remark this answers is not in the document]
Velfi: no

CHANGELOG.next.toml
```diff
@@ -261,3 +261,9 @@ For more information, see the [Change Log Discussion](https://github.com/awslabs
 meta = { "breaking" = true, "tada" = false, "bug" = false }
 references = ["smithy-rs#3014"]
 author = "rcoh"
+
+[[aws-sdk-rust]]
+message = "`ProviderConfig` will now respect both `use_fips` and `use_dual_stack` when those settings are configured in a user's environment or profile."
+references = ["aws-sdk-rust#882", "smithy-rs#3007"]
+meta = { "breaking" = false, "tada" = true, "bug" = true }
+author = "Velfi"
```
rcoh, line 269: message should be updated to be customer facing (referencing SSO/STS credential providers)

aws/rust-runtime/aws-config/src/profile/credentials/exec.rs
```diff
@@ -30,13 +34,18 @@ impl AssumeRoleProvider {
...
-            .clone()
-            .credentials_provider(input_credentials)
+        let config = sdk_config
+            .to_builder()
+            .credentials_provider(SharedCredentialsProvider::new(provide_credentials_fn(
+                move || {
+                    let input_credentials = input_credentials.clone();
+                    async { Ok(input_credentials) }
+                },
+            )))
```
rcoh, line 46: I thought credentials impl provider? can you pass those directly?

aws/rust-runtime/aws-config/src/sts/assume_role.rs
```diff
@@ -362,7 +363,7 @@ mod test {
     #[tokio::test]
     async fn configures_session_length() {
         let (server, request) = capture_request(None);
-        let provider_conf = SdkConfig::builder()
+        let sdk_config = SdkConfig::builder()
```
rcoh, line 366: 👍🏻

aws/rust-runtime/aws-config/src/sts/assume_role.rs
```diff
@@ -459,6 +458,56 @@ mod test {
...
+        let provider = AssumeRoleProvider::builder("myrole")
+            .configure(&sdk_config)
+            .build_from_provider(provide_credentials_fn(|| async {
+                Ok(Credentials::for_tests())
+            }))
+            .await;
+        let _ = provider.provide_credentials().await;
+        let req = request.expect_request();
+        assert!(req.uri().host().unwrap().contains("sts-fips"));
+    }
```
rcoh, line 480: these tests are basically equivalent to the tests that already existed—we need tests of the profile file provider / default chain itself

rcoh (commented):
changes LGTM–need other tests

aws/rust-runtime/aws-config/src/sts/assume_role.rs
Velfi: fixed.

rcoh (approved):

PR #3009, opened 2023-09-28
Specify tilde version requirements for runtime crates in manifest files

ysaito1001:
## Motivation and Context
This PR converts version requirements for runtime crates (those under `aws/rust-runtime` and those under `rust-runtime`) in manifest files into tilde requirements, allowing customers to safely incorporate only patch versions (from [The Cargo Book](https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html)).

## Description
Prior to the PR, if one opens the manifest file for an SDK create, a version requirement for a runtime crate looks like [this](https://github.com/awslabs/aws-sdk-rust/blob/655d490682e97ef799da85703520b5501fe91574/sdk/s3/Cargo.toml#L24-L27). Post general availability (GA), those runtime crates will have versions like `1.X.X` and if specified as such in a manifest file, it will be interpreted as `>=1.X.X, <2.0.0`. To protect customers from minor version bumps from runtime crates, we want to change their version requirements to [tilde requirements](https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#tilde-requirements) `~1.X`, which then supports a version range `>=1.X.0, <1.X+1.0`.

Note that the change being discussed here is only concerned with version requirements of runtime crates. There is a runtime crate that depends on SDK crates (e.g. `aws-config` depending on `aws-sdk-sts`). In that case, we do _not_ want to turn the version requirement of the SDK crate into a tilde version because customers may depend on that SDK crate by themselves, causing multiple versions of it to be brought in to their crate graph.

To support the above functionality, `publisher` has been updated. Specifically, when the `fix-manifests` subcommand runs, for runtime crates it will render the tilde version requirements into manifest files. Regarding a implementation detail, since the `semver::Version` used in `publisher` requires three components, `major.minor.patch`, to appear when version requirements are turned into in-memory data structure, it does not quite work well with non-semver version requirements. To address it, our own `crate::package::Version` has been introduced to handle both semver and versions that do not adhere to the semver specification. It then boils down to the `parse_version` function being updated to be able to parse an input string into `crate::package::Version`, as specified at call sites through a generic parameter `P: ParseIntoVersion`.

## Testing
- Updated an existing unit test `test_validate_packages` in the `package` module.
- Added a new unit test `sdk_crate_version_should_not_be_turned_into_tilde_requirement` to the `fix-manifests` subcommand.
- Ran our internal release pipeline and confirmed the generated SDK crates have tilde dependencies for runtime crates in their manifest files.

## Checklist

----


tools/ci-build/publisher/src/package.rs
```diff
@@ -255,38 +340,60 @@ fn read_package(path: &Path, manifest_bytes: &[u8]) -> Result<Option<Package>> {
...
 /// across all of their local dependencies.
 fn validate_packages(packages: &[Package]) -> Result<()> {
-    let mut versions: BTreeMap<String, Version> = BTreeMap::new();
-    let track_version = &mut |handle: &PackageHandle| -> Result<(), Error> {
-        if let Some(version) = versions.get(&handle.name) {
+    let mut versions: BTreeMap<String, LenientVersion> = BTreeMap::new();
+    let mut version_requirements: BTreeMap<String, LenientVersion> = BTreeMap::new();
+    let track_version = &mut |handle: &PackageHandle,
+                              map: &mut BTreeMap<String, LenientVersion>|
+     -> Option<(String, LenientVersion, LenientVersion)> {
```
jdisanti, line 347: Consider returning a result instead of an option since you're mapping it into an error later anyway.

tools/ci-build/publisher/src/package.rs
```diff
@@ -255,38 +340,60 @@ fn read_package(path: &Path, manifest_bytes: &[u8]) -> Result<Option<Package>> {
 /// Validates that all of the publishable crates use consistent version numbers
 /// across all of their local dependencies.
 fn validate_packages(packages: &[Package]) -> Result<()> {
-    let mut versions: BTreeMap<String, Version> = BTreeMap::new();
-    let track_version = &mut |handle: &PackageHandle| -> Result<(), Error> {
-        if let Some(version) = versions.get(&handle.name) {
+    let mut versions: BTreeMap<String, LenientVersion> = BTreeMap::new();
+    let mut version_requirements: BTreeMap<String, LenientVersion> = BTreeMap::new();
+    let track_version = &mut |handle: &PackageHandle,
+                              map: &mut BTreeMap<String, LenientVersion>|
```
jdisanti, line 346: Doesn't look like this refers to anything outside of the closure, so you can just declare it as a normal nested function.

tools/ci-build/publisher/src/subcommand/fix_manifests.rs
```diff
@@ -25,6 +25,8 @@ use tracing::{debug, info};
 
 mod validate;
 
+const AWS_SDK_CRATE_NAME_PREFIX: &str = "aws-sdk-";
```
jdisanti, line 28: There is already code somewhere that categorizes package names. Shouldn't need to do this manually.

tools/ci-build/publisher/src/subcommand/fix_manifests.rs
```diff
@@ -210,6 +233,21 @@ fn update_dep(table: &mut Table, dep_name: &str, versions: &VersionView) -> Resu
     }
 }
 
+// Convert `package_version` into a tilde version requirement
+//
+// For instance, given `package_version` like 0.12.3, the function returns `~0.12`.
+fn convert_to_tilde_requirement(package_version: &str) -> Result<String> {
+    // `package_version` is from the `semver` crate which requires versions to have 3 components,
```
jdisanti, line 240: There's no type-level enforcement of this assumption. Should the SemVer or Version be passed in directly here rather than a str?

tools/ci-build/publisher/src/subcommand/publish.rs
```diff
@@ -249,10 +249,16 @@ mod test {
...
     #[tokio::test]
     async fn crate_published_works() {
-        let handle = PackageHandle::new("aws-smithy-http", "0.27.0-alpha.1".parse().unwrap());
+        let handle = PackageHandle::new(
+            "aws-smithy-http",
+            "0.27.0-alpha.1".parse::<String>().unwrap(),
+        );
         assert!(is_published(&handle).await.expect("failed"));
         // we will never publish this version
-        let handle = PackageHandle::new("aws-smithy-http", "0.21.0-alpha.1".parse().unwrap());
```
jdisanti, line 255: `0.21.0-alpha.1` is a good example that breaks the tilde conversion logic, although it's unlikely we'll ever use a version number like this.

tools/ci-build/publisher/src/package.rs
```diff
@@ -15,22 +15,90 @@ use smithy_rs_tool_common::package::PackageCategory;
...
+    type ParsedT = Version;
+    type Err = semver::Error;
+}
+
+/// A type that indicates the `Lenient` variant of `LenientVersion` will be created as a result of
+/// parsing
+pub struct Lenient;
+impl ParseIntoLenientVersion for Lenient {
+    type ParsedT = String;
+    type Err = std::convert::Infallible;
```
rcoh, line 43: is this trait really required? I'm trying to understand how it's actually used...

tools/ci-build/publisher/src/package.rs
```diff
@@ -192,18 +262,28 @@ pub async fn discover_manifests(path: PathBuf) -> Result<Vec<PathBuf>> {
...
-pub fn parse_version(manifest_path: &Path, version: &str) -> Result<Version, Error> {
-    Version::parse(version)
+/// Parses `version` into `LenientVersion` and adds additional error context when parsing fails.
+pub fn parse_version<P: ParseIntoLenientVersion>(
+    manifest_path: &Path,
+    version: &str,
+) -> Result<LenientVersion, Error> {
+    P::ParsedT::from_str(version)
+        .map(|ver| ver.into())
         .map_err(|err| Error::InvalidCrateVersion(manifest_path.into(), version.into(), err.into()))
```
rcoh, line 280: it seems like this function should return the type you ask for? (Like `parse::<T>`)?

rcoh (commented):
overall this seems fine—I'm not sure we need that ParseLenient trait though...

tools/ci-build/publisher/src/package.rs
ysaito1001: It is more of a convenience trait, and without it, call site like below
```
read_package<SemVer>(...)
```
would need to specify
```
read_package<semver::Version, semver::Error>(...)
```
and functions being called down the road need to have two generic parameters instead of one.

I'll add a comment as such.

tools/ci-build/publisher/src/package.rs
ysaito1001: I think I'm looking for a nicer way myself here, but if this returns a result, the implementation needs to know who calls `track_version` because we're mapping a returned tuple to different error enum variants; if it's called with a map `versions` then it ends up `Error::MultipleVersions` but if it's called with a map `version_requirements`, then it ends up with `Error::MultipleVersionRequirements`. Being inside `track_version`, we don't know which error variant should be returned.

ysaito1001: Maybe I can pass in a closure that returns `Error` to `track_version`, and each call site gets to specify what error variant it wants the function to return.

ysaito1001: Maybe I can pass in to `track_version` a closure that takes a tuple and returns `Error`, and each call site gets to specify what error variant it wants the function to return.

CHANGELOG.next.toml
```diff
@@ -200,4 +200,10 @@ author = "ysaito1001"
 message = "Python middleware can set URI. This can be used to route a request to a different handler."
 references = ["smithy-rs#3005"]
 meta = { "breaking" = false, "tada" = true, "bug" = false, "target" = "server" }
-author = "drganjoo"
\ No newline at end of file
+author = "drganjoo"
+
+[[aws-sdk-rust]]
+message = "Manifest files for SDK crates now specify [tilde version requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#tilde-requirements) for runtime crates (i.e. those under `aws/rust-runtime` and `rust-runtime`), allowing customers to only update patch versions without bumping a minor version."
+references = ["smithy-rs#3009"]
```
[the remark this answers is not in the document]
ysaito1001: That makes sense. I have removed the entry in 153ba2d.

tools/ci-build/publisher/src/package.rs
ysaito1001: Addressed in 1597129.

tools/ci-build/publisher/src/package.rs
ysaito1001: Addressed in 1597129.

tools/ci-build/publisher/src/subcommand/fix_manifests.rs
ysaito1001: I like that and think it's safer in the long run (even if the function is currently private and only called from a call site where `package_version: &str` is derived from `semver::Version`).

Addressed in 49a067d.

tools/ci-build/publisher/src/package.rs
```diff
@@ -15,22 +15,90 @@ use smithy_rs_tool_common::package::PackageCategory;
...
+
+/// An enum that handles both semver as well as string values that do not adhere to the semver
+/// specification.
+///
+/// Most of the time, the `SemVer` variant will be used when manifest files are parsed into an
+/// in-memory data structure. The `Lenient` variant is used when the `fix_manifests` subcommand
+/// and the `generate_version_manifest` subcommand are executed, allowing the tilde version
+/// requirements to be rendered into the manifest files.
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub enum LenientVersion {
```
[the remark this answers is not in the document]
ysaito1001: Technically, the version of a package (the one appearing under `[package]`) and those under the dependency table ([the manifest format](https://doc.rust-lang.org/cargo/reference/manifest.html#the-manifest-format)) are both represented by this struct. If the struct were only for those under the dependency table, then I would use `VersionRequirement` right away, but that doesn't represent the package version, i.e. it's incorrect to put a tilde version requirement under `[package]`.

I'll sleep on an alternative name.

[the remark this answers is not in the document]
ysaito1001: Technically, the version of a package (the one appearing under `[package]`) and those under the dependency table ([the manifest format](https://doc.rust-lang.org/cargo/reference/manifest.html#the-manifest-format)) are both represented by this struct. If the struct were only for those under the dependency table, then I would use `VersionRequirement` right away, but that doesn't represent the package version, i.e. it's incorrect to put a tilde version requirement under `[package]` 🤔

tools/ci-build/publisher/src/subcommand/fix_manifests.rs
ysaito1001: Didn't realize there is already one existing! Fixed in bddc9e5.

tools/ci-build/publisher/src/package.rs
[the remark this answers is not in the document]
ysaito1001: I've decided to define our own `crate::package::Version` (couldn't think of a better name). Wherever `semver::Version` is used, the type name has been fully qualified for clarity. See 0a4ab75 and a9e122d.

tools/ci-build/publisher/src/package.rs
ysaito1001: Added a comment in 0a4ab75.

tools/ci-build/publisher/src/package.rs
ysaito1001: The function returning what we ask for actually helped get rid of [try_into_semver](https://github.com/awslabs/smithy-rs/pull/3009/commits/000950d7a0996d94b40f5ca5c174bbf0bba30f4c#diff-2cd24b049e142d74c42052e9ba595284814576c4bcf07fe9427621fdfe4349b7R60-R65), so I like it. Addressed in 0a4ab75.

tools/ci-build/publisher/src/sort.rs
```diff
@@ -34,27 +34,28 @@ pub fn dependency_order(packages: Vec<Package>) -> Result<Vec<Package>> {
...
 }
 
 fn dependency_order_visit(
     package_handle: &PackageHandle,
-    packages: &BTreeMap<PackageHandle, Package>,
-    stack: &mut BTreeSet<PackageHandle>,
-    visited: &mut BTreeSet<PackageHandle>,
+    packages: &BTreeMap<String, Package>,
+    stack: &mut BTreeSet<String>,
+    visited: &mut BTreeSet<String>,
```
jdisanti, line 45: Does this break anything if we depend on multiple versions of the same third-party crate?

tools/ci-build/publisher/src/subcommand/fix_manifests.rs
```diff
@@ -10,12 +10,12 @@
...
 
 use crate::fs::Fs;
-use crate::package::{discover_manifests, parse_version};
+use crate::package::{discover_manifests, parse_version, SemVer};
 use crate::SDK_REPO_NAME;
 use anyhow::{bail, Context, Result};
 use clap::Parser;
-use semver::Version;
 use smithy_rs_tool_common::ci::running_in_ci;
+use smithy_rs_tool_common::package::SDK_PREFIX;
```
jdisanti, line 18: I'm not sure why this constant is public... use `PackageCategory` instead:
```rust
PackageCategory::from_package_name(dep_name) == PackageCategory::AwsSdk
```

tools/ci-build/publisher/src/subcommand/fix_manifests.rs
```diff
@@ -210,6 +224,24 @@ fn update_dep(table: &mut Table, dep_name: &str, versions: &VersionView) -> Resu
...
+// Convert `package_version` into a tilde version requirement
+//
+// For instance, given `package_version` like `0.12.3`, the function returns `~0.12`.
+// The fact that this function takes a `semver::Version` means one can only convert a complete
+// semver `x.y.z` into a tilde version requirement, but not those like `0.21.0-alpha.1`.
+fn convert_to_tilde_requirement(package_version: &semver::Version) -> Result<String> {
+    // `package_version` is from the `semver` crate which requires versions to have 3 components,
+    // major, minor, and patch. So it is safe to assume its string value follows that format.
+    let package_version = package_version.to_string();
+    let package_version = match package_version.rfind('.') {
```
jdisanti, line 236: Consider adding an `assert!` that makes sure there are only two `.` characters so that it breaks instead of doing something weird for versions like `0.21.0-alpha.1`

tools/ci-build/publisher/src/sort.rs
ysaito1001: I don't think it does because the `publisher` is currently (and I think will continue to be) concerned with crates that we are allowed to publish. Whenever `publisher` discovers & validates packages, it deals with [such packages](https://github.com/awslabs/smithy-rs/blob/d800d33e283ff10d9d75ec81f933a5137ef14721/tools/ci-build/publisher/src/package.rs#L151).

So the third-party crates have not been in the picture for discovery & validation of packages. For those we can publish, multiple version checks have already been done by [validate_packages](https://github.com/awslabs/smithy-rs/blob/d800d33e283ff10d9d75ec81f933a5137ef14721/tools/ci-build/publisher/src/package.rs#L153C17-L153C17) before we determine the sort order.

ysaito1001: I don't think it does because the `publisher` is currently (and I think will continue to be) concerned with crates that we are allowed to publish. Whenever `publisher` discovers & validates packages, it deals with [such packages](https://github.com/awslabs/smithy-rs/blob/d800d33e283ff10d9d75ec81f933a5137ef14721/tools/ci-build/publisher/src/package.rs#L151).

So the third-party crates have not been in the picture for discovery & validation of packages. For those we can publish, multiple version checks have already been done by [validate_packages](https://github.com/awslabs/smithy-rs/blob/d800d33e283ff10d9d75ec81f933a5137ef14721/tools/ci-build/publisher/src/package.rs#L153C17-L153C17) before we determine the sort order.

Should the requirements for `publish` change in the future to handle third-party crate as well, an enhancement we can make is to define custom `Eq` and `PartialEq` traits for `crate::package::Version`.

tools/ci-build/publisher/src/subcommand/fix_manifests.rs
ysaito1001: Updated in b1ff4b6.

tools/ci-build/publisher/src/subcommand/fix_manifests.rs
ysaito1001: True, a semver can contain more than two dots. Added assert in 137822d.

jdisanti (approved):

ysaito1001:
Will run our internal release pipeline again before merging, just to be sure.

PR #3014, opened 2023-09-29
Remove ProviderConfig as the configuration source for AssumeRoleProvider

rcoh:
## Motivation and Context
`AssumeRoleProvider` currently uses `ProviderConfig` as a source of configuration, but that API is hard use and not intended for external consumption.

This fixes the Assume Role issue but only for `AssumeRoleProvider`

## Description
Update the API (see changelog) to be more ergonomic and derive configuration from `SdkConfig` instead.

## Testing
Existing tests + new unit tests

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/sts/assume_role.rs
+++ b/aws/rust-runtime/aws-config/src/sts/assume_role.rs
@@ -5,18 +5,18 @@
 
 //! Assume credentials for a role through the AWS Security Token Service (STS).
 
-use crate::connector::expect_connector;
-use crate::provider_config::ProviderConfig;
 use aws_credential_types::cache::CredentialsCache;
-use aws_credential_types::provider::{self, error::CredentialsError, future, ProvideCredentials};
+use aws_credential_types::provider::{
+    self, error::CredentialsError, future, ProvideCredentials, SharedCredentialsProvider,
+};
 use aws_sdk_sts::operation::assume_role::builders::AssumeRoleFluentBuilder;
 use aws_sdk_sts::operation::assume_role::AssumeRoleError;
 use aws_sdk_sts::types::PolicyDescriptorType;
 use aws_sdk_sts::Client as StsClient;
-use aws_smithy_client::erase::DynConnector;
 use aws_smithy_http::result::SdkError;
 use aws_smithy_types::error::display::DisplayErrorContext;
 use aws_types::region::Region;
+use aws_types::SdkConfig;
 use std::time::Duration;
 use tracing::Instrument;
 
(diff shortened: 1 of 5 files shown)


aws/rust-runtime/aws-config/src/sts/assume_role.rs
```diff
@@ -191,36 +208,57 @@ impl AssumeRoleProviderBuilder {
...
+        // ignore a credentials cache set from SdkConfig
+        conf = conf
+            .to_builder()
+            .credentials_cache(
+                self.credentials_cache
+                    .unwrap_or(CredentialsCache::no_caching()),
+            )
+            .build();
+
+        if conf.credentials_cache().is_none() {}
```
jdisanti, line 251:
```suggestion
```

aws/rust-runtime/aws-config/src/sts/assume_role.rs
```diff
@@ -287,40 +342,121 @@ impl ProvideCredentials for AssumeRoleProvider {
...
+            .sleep_impl(SharedAsyncSleep::new(TokioSleep::new()))
+            .time_source(StaticTimeSource::new(
+                UNIX_EPOCH + Duration::from_secs(1234567890 - 120),
+            ))
+            .http_connector(DynConnector::new(server))
+            .credentials_provider(SharedCredentialsProvider::new(provide_credentials_fn(
+                || async {
+                    panic!("don't call me — will be overridden");
+                    #[allow(unreachable_code)]
+                    Ok(Credentials::for_tests())
```
jdisanti, line 401:
```suggestion
                    panic!("don't call me — will be overridden")
```

aws/rust-runtime/aws-types/src/sdk_config.rs
```diff
@@ -602,4 +602,22 @@ impl SdkConfig {
     pub fn builder() -> Builder {
         Builder::default()
     }
+
+    /// Convert this [`SdkConfig`] back to a builder to enable modification
+    pub fn to_builder(self) -> Builder {
```
jdisanti, line 607: This is awesome!

The `to_builder()` on the generated configs takes `&self` rather than `self`. Should this be consistent?

rust-runtime/aws-smithy-async/src/time.rs
```diff
@@ -51,6 +51,11 @@ impl StaticTimeSource {
     pub fn new(time: SystemTime) -> Self {
         Self { time }
     }
+
+    /// Creates a new static time source from the provided number of seconds since the UNIX epoch
+    pub fn new_from_epoch(epoch_secs: u64) -> Self {
```
jdisanti, line 56: This is called `from_secs` on our DateTime. Should we use the same name here?

aws/rust-runtime/aws-config/src/sts/assume_role.rs
rcoh: you need this for inference

ysaito1001 (approved):
Less mental cost with `ProviderConfig` gone. Thanks for adding examples.

jdisanti (approved):

PR #3017, opened 2023-10-02
Add additional default semantics

rcoh:
> don't merge this until models are finalized with cleaned-up defaults.

Currently there are defaults that shouldn't exist which causes this change to have undesired behavior.

## Motivation and Context
Updates to the Smithy spec about default values, specifically:
- defaults should be set when the field is marked as required

## Description
An additional codepath was added to the builder generator to add default values for fields marked as required
## Testing
- New unit test
- [ ] consider codegen diff

## Checklist

----

### DIFF
--- a/codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGeneratorTest.kt
+++ b/codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGeneratorTest.kt
@@ -177,12 +177,24 @@ internal class BuilderGeneratorTest {
               @default(true)
               @required
               defaultDocument: Document
+
+              @clientOptional
+              @required
+              @default([])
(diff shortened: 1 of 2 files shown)


codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGeneratorTest.kt
```diff
@@ -207,6 +219,7 @@ internal class BuilderGeneratorTest {
                     assert_eq!(s.an_actually_required_field(), 5);
                     assert_eq!(s.no_default(), None);
                     assert_eq!(s.default_document().as_bool().unwrap(), true);
+                    assert_eq!(s.required_but_optional, Some(vec![]));
```
jdisanti, line 222: Missing an assertion on the new `reqOptInt` field?

PR #3021, opened 2023-10-03
Ensure identity resolver exists when a credentials provider is given only at operation level

ysaito1001:
## Motivation and Context
Fixes https://github.com/awslabs/aws-sdk-rust/issues/901

## Description
When a credentials provider is specified _only_ at the operation level (but not at the service config level), the code in the above PR fails on request dispatch, saying `NoMatchingAuthScheme`. This occurs today because if we [do not set a credentials provider at the service config level](https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/kms/src/config.rs#L757-L769), we will [not set the identity resolver for sigv4](https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/kms/src/config.rs#L811-L818). The same goes for configuring a `SigningRegion` when it is only supplied at the operation level.

This PR fixes the said issue so that `config_override` sets
- the identity resolver for sigv4 when a credentials provider is supplied only at the operation config level
- a `SigningRegion` when a `Region` is given only at the operation level

## Testing
Added a Kotlin integ test `test_specifying_credentials_provider_only_at_operation_level_should_work` based on the customer reported PR.

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -48,7 +48,7 @@ meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "all" }
 author = "jdisanti"
 
 [[aws-sdk-rust]]
-message = "Fix operation level configuration so that it now sets a) an `IdentityResolver` when a credentials provider is provided only at the operation level and b) a `SigningRegion` when a `Region` is provided only at the operation level."
+message = "Fix exclusively setting the credentials provider and/or region at operation config-override time. It's now possible to set the region or credentials when an operation is sent (via `.config_override()`), rather than at client-creation time."
 references = ["smithy-rs#3021", "aws-sdk-rust#901"]
 meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "ysaito1001"
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -46,3 +46,9 @@ message = "Fix code generation for union members with the `@httpPayload` trait."
 references = ["smithy-rs#2969", "smithy-rs#1896"]
 meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "all" }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Fix operation level configuration so that it now sets a) an `IdentityResolver` when a credentials provider is provided only at the operation level and b) a `SigningRegion` when a `Region` is provided only at the operation level."
+references = ["smithy-rs#3021", "aws-sdk-rust#901"]
+meta = { "breaking" = false, "tada" = false, "bug" = true }
+author = "ysaito1001"
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialCaches.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialCaches.kt
@@ -57,10 +57,17 @@ class CredentialCacheConfig(codegenContext: ClientCodegenContext) : ConfigCustom
     private val codegenScope = arrayOf(
         *preludeScope,
         "CredentialsCache" to AwsRuntimeType.awsCredentialTypes(runtimeConfig).resolve("cache::CredentialsCache"),
+        "CredentialsIdentityResolver" to AwsRuntimeType.awsRuntime(runtimeConfig)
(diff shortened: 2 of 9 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialCaches.kt
```diff
@@ -209,7 +216,14 @@ class CredentialCacheConfig(codegenContext: ClientCodegenContext) : ConfigCustom
                             #{Some}(credentials_cache),
                             #{Some}(credentials_provider),
                         ) => {
-                            resolver.config_mut().store_put(credentials_cache.create_cache(credentials_provider));
+                            let credentials_cache = credentials_cache.create_cache(credentials_provider);
+                            resolver.config_mut().store_put(credentials_cache.clone());
```
jdisanti, line 220: For my understanding, why does the cache also get written back to config?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4SigningDecorator.kt
```diff
@@ -120,6 +120,19 @@ class SigV4SigningConfig(
                     )
                 }
             }
+            is ServiceConfig.OperationConfigOverride -> {
+                if (runtimeMode.generateOrchestrator) {
+                    rustTemplate(
+                        """
+                        resolver.config_mut()
```
jdisanti, line 127: We should probably only do this if the SigningRegion isn't already set, since this could unintentionally overwrite another runtime plugin that set it.

CHANGELOG.next.toml
```diff
@@ -46,3 +46,9 @@ message = "Fix code generation for union members with the `@httpPayload` trait."
 references = ["smithy-rs#2969", "smithy-rs#1896"]
 meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "all" }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Fix operation level configuration so that it now sets a) an `IdentityResolver` when a credentials provider is provided only at the operation level and b) a `SigningRegion` when a `Region` is provided only at the operation level."
```
jdisanti, line 51:
```suggestion
message = "Fix exclusively setting the credentials provider and/or region at operation config override time. It's now possible to set the region or credentials when an operation is sent (via `.config_override()`), rather than at client-creation time."
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialCaches.kt
ysaito1001: Ah, good point. Line 220 can go away. I made it so based on my stale understanding of the cache, before the time runtime components had been introduced and the cache was picked up directly from the config bag. Now that we have runtime components, we don't need to write this back to config as long as we write back a new identity to runtime components.

ysaito1001: Updated in 3f09577.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4SigningDecorator.kt
ysaito1001: In this case, customers intentionally override
```
client
    .operation()
    .customize()
    .await
    .unwrap()
    .config_override(Config::builder().region(Region::new("us-east-1")))
    ... // and so on
```
Isn't this a surprise to them if `us-east-1` is not respected? Or are we saying they specify a `Region` from a different mechanism other than `config_override`?

jdisanti: Oh, is this only loading from the config override, and not from a combination of the client config and override config? If that's the case, then this should be fine.

ysaito1001: Ah, I see what you mean. Yes, this is prior to combining the client config and override config.

jdisanti (approved):

PR #3024, opened 2023-10-04
Merge 0.56.x back into main

rcoh:
----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/QuerySerializerGenerator.kt
```diff
@@ -128,6 +128,10 @@ abstract class QuerySerializerGenerator(private val codegenContext: CodegenConte
         TODO("AwsQuery doesn't support payload serialization")
     }
 
+    override fun unsetUnion(union: UnionShape): RuntimeType {
+        TODO("AwsQuery doesn't support payload serialization")
+    }
+
```
ysaito1001, line 134: This looks a duplicated definition as line 127-129, so we probably need to remove it.

PR #3026, opened 2023-10-05
Move `SdkBody` and `bytestream` into `aws-smithy-types`

ysaito1001:
## Motivation and Context
Takes care of the first part of https://github.com/awslabs/smithy-rs/issues/3053 (the remaining part is denoted as `TODO(runtimeCratesVersioningCleanup)` and until that and https://github.com/awslabs/smithy-rs/issues/3033 are done the issue will not be closed).

## Description
This PR moves from `aws-smithy-http` to `aws-smithy-types`:
- the `SdkBody` struct
- the `byte_stream` module

Within the origin crate, we leave "breadcrumbs" (i.e. reexports) for existing customers of the items above so they are not broken by the move.

We have just moved `SdkBody` to `aws-smithy-types` without renaming it to see how it looks there. However, as [TODO(naming)](https://github.com/awslabs/smithy-rs/pull/3026/files#diff-090c503b779024fdadb8ac97465c80438635df82e62c42b0d85b588a303d9a95R28) says, we can choose to rename it to just `Body`. Curious to hear what the reviewers think.

## Testing
Relied on the tests in CI.

----


rust-runtime/aws-smithy-types/src/futures_stream_adapter.rs
```diff
@@ -17,6 +17,7 @@ use std::task::{Context, Poll};
 /// new-type to enable the trait when it is required.
 ///
 /// This is meant to be used by codegen code, and users should not need to use it directly.
+#[derive(Debug)]
```
rcoh, line 20: can these be kept in aws-smithy-http? It's only used in unstable situations, right?

rcoh (commented):
overall the right direction, couple of small things to fix up

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
```diff
@@ -410,7 +410,7 @@ data class RuntimeType(val path: String, val dependency: RustDependency? = null)
         fun eventStreamSender(runtimeConfig: RuntimeConfig): RuntimeType =
             smithyHttp(runtimeConfig).resolve("event_stream::EventStreamSender")
         fun futuresStreamCompatByteStream(runtimeConfig: RuntimeConfig): RuntimeType =
```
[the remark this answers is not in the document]
ysaito1001: If the adapter lived in `aws-smithy-http`, then `aws-smithy-types` would have to depend on that, causing a circular dependency. Further, the adapter needs to tag along `ByteStream` wherever it goes, so the answer is probably no 🤔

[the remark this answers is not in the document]
ysaito1001: Alternatively, we can duplicate the wrapper in `aws-smithy-types`. The wrapper can be private and lives in `aws_smithy_types::byte_stream` (`ByteStream::into_async_read` needs it).

rust-runtime/aws-smithy-types/src/futures_stream_adapter.rs
ysaito1001: It's also used by `ByteStream::into_async_read`.

rust-runtime/aws-smithy-types/src/lib.rs
```diff
@@ -14,11 +14,19 @@
...
 pub mod config_bag;
 pub mod date_time;
 pub mod endpoint;
 pub mod error;
+// Marked as `doc(hidden)` because a type in the module is used both by this crate and by the code
+// generator, but not by external users. Also, by the module being `doc(hidden)` instead of it being
+// in `rust-runtime/inlineable`, each user won't have to pay the cost of running the module's tests
+// when compiling their generated SDK.
+#[doc(hidden)]
+pub mod futures_stream_adapter;
```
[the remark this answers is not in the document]
ysaito1001: Interesting. Originally, the adapter was created as an inlineable. Plus, the fact that we want to have `futures_stream_adapter` stay in smithy-http may mean that the adapter is turned back to an inlineable and that for `aws_smithy_types::byte_stream::ByteStream::into_async_read` can just define a small futures-stream-enabling-wrapper locally and use it.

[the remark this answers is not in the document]
ysaito1001: Interesting. Originally, the adapter was created as an inlineable. Now that we kinda want `futures_stream_adapter` to stay within `aws-smithy-http` means that
- the adapter can be turned back to an inlineable and the codegen just uses it from there.
- `aws_smithy_types::byte_stream::ByteStream::into_async_read` can define its own futures-stream-enabling-wrapper locally (within the module or even within the method) and use it.

rust-runtime/aws-smithy-http/src/body.rs
```diff
@@ -5,379 +5,6 @@
...
-
-    #[test]
-    fn sdk_body_is_send() {
-        fn is_send<T: Send>() {}
-        is_send::<SdkBody>()
-    }
-}
+//TODO(runtimeCratesVersioningCleanup): Re-point those who use the following reexport to
+// `aws_smithy_types` and remove the `body` module.
+pub use aws_smithy_types::body::{BoxBody, Error, SdkBody};
```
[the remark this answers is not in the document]
ysaito1001: 👍 Updated in ef08eea

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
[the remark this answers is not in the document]
jdisanti: Can the things that need it outside of `aws-smithy-types` use the `into_async_read` method, or do they need to name the type?

rust-runtime/aws-smithy-http/Cargo.toml
```diff
@@ -27,30 +27,20 @@ pin-project-lite = "0.2.9"
...
 
-# We are using hyper for our streaming body implementation, but this is an internal detail.
-hyper = "0.14.26"
-
-# ByteStream internals
+# For our event stream implementation
 futures-core = "0.3.14"
-tokio = { version = "1.23.1", optional = true }
-tokio-util = { version = "0.7", optional = true }
+hyper = "0.14.26"
```
jdisanti, line 32: Can hyper be an optional dependency now? If so, we should add a `hyper014` feature.

rust-runtime/aws-smithy-http/external-types.toml
```diff
@@ -12,20 +11,10 @@ allowed_external_types = [
...
-    "http::version::Version",
-    "http_body::Body",
-    "http_body::combinators::box_body::BoxBody",
-    "hyper::body::body::Body",
-
-    # TODO(https://github.com/awslabs/smithy-rs/issues/1193): Feature gate Tokio `AsyncRead`
-    "tokio::io::async_read::AsyncRead",
 
     # TODO(https://github.com/awslabs/smithy-rs/issues/1193): Switch to AsyncIterator once standardized
     "futures_core::stream::Stream",
```
jdisanti, line 16: What is this still needed for?

rust-runtime/aws-smithy-types/Cargo.toml
```diff
@@ -11,17 +11,30 @@ license = "Apache-2.0"
...
 serde-serialize = []
 serde-deserialize = []
 
 [dependencies]
 base64-simd = "0.8"
+bytes = "1"
+bytes-utils = "0.1"
+http = "0.2.3"
+http-body = "0.4.4"
+hyper = "0.14.26"
```
jdisanti, line 25: Why is hyper in here?

rust-runtime/aws-smithy-types/Cargo.toml
```diff
@@ -11,17 +11,30 @@ license = "Apache-2.0"
...
 itoa = "1.0.0"
 num-integer = "0.1.44"
+pin-project-lite = "0.2.9"
+pin-utils = "0.1.0"
```
jdisanti, line 29: We might be able to get rid of pin-utils at this point due to more recent stabilizations.

rust-runtime/aws-smithy-types/external-types.toml
```diff
@@ -1,2 +1,15 @@
 allowed_external_types = [
+    "bytes::bytes::Bytes",
+    "bytes::buf::buf_impl::Buf",
+
+    # TODO(https://github.com/awslabs/smithy-rs/issues/3033): Feature gate based on unstable versions
+    "http_body::Body",
+    "http_body::combinators::box_body::BoxBody",
+    "hyper::body::body::Body",
```
jdisanti, line 8: What requires exposing this?

rust-runtime/aws-smithy-http/external-types.toml
ysaito1001: For event-stream. It's worth updating the `TODO` comment to only use `Stream` with the `event-stream` feature.

rust-runtime/aws-smithy-types/external-types.toml
ysaito1001: - `impl http_body::Body for SdkBody` (src/body.rs:269:1)
- `pub type BoxBody = http_body::combinators::BoxBody<Bytes, Error>;` (src/body.rs:52:1)
- `impl From<hyper::Body> for SdkBody` (src/byte_stream.rs:471)

rust-runtime/aws-smithy-types/Cargo.toml
ysaito1001: For `impl From<hyper::Body> for SdkBody (src/byte_stream.rs:471)`

rust-runtime/aws-smithy-types/Cargo.toml
ysaito1001: Thanks for the tip! Since this PR only handles the move, we can clean them up in a subsequent PR.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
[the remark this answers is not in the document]
ysaito1001: The use cases of the adapters are [shown here](https://github.com/search?q=repo%3Aawslabs%2Fsmithy-rs%20FuturesStreamCompatByteStream&type=code), and there are only two of them. One use case is for implementing `ByteStream::into_async_read` and the other is for server's codegen.

To answer your question, server's codegen does not use the `into_async_read` method. They may not need the name the type.

The adapter was originally created as an inlineable, used by the server's codegen, but then `into_async_read` could also benefit the adapter, which was why the adapter ended up in `aws-smithy-http`. Technically speaking, both use cases do not need to share the adapter and each can define its own wrapper, if we're OK with duplicating small amount of code.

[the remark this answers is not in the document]
jdisanti: I think in both of those cases, you can just call `into_async_read()` rather than directly naming this adapter.

rust-runtime/aws-smithy-types/Cargo.toml
jdisanti: @rcoh, do we actually need this?

rust-runtime/aws-smithy-types/external-types.toml
jdisanti: How is BoxBody used? Can we get rid of it?

rust-runtime/aws-smithy-types/Cargo.toml
ysaito1001: Looks like it's at least used by the server (if we remove it, the server codegen fails as follows):
```
     --> rest_json/rust-server-codegen/src/operation.rs:35849:5
      |
35849 | /     ::aws_smithy_http_server::request::FromRequest<
35850 | |         ::aws_smithy_http_server::protocol::rest_json_1::RestJson1,
35851 | |         B,
35852 | |     > for crate::input::StreamingTraitsInput
      | |_____^     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
35857 |       B: Into<::aws_smithy_http::byte_stream::ByteStream>,
      |          ------------------------------------------------ unsatisfied trait bound introduced here
```

jdisanti: If it's only for a conversion in server, we can just change it to not use a From trait for that conversion.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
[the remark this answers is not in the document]
ysaito1001: I moved back `futures_stream_adapter` to `aws-smithy-http` in 82a04cb. For `ByteStream::into_async_read`, we have a local struct just to enable the `Stream` trait.

rust-runtime/aws-smithy-types/src/lib.rs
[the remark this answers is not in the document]
ysaito1001: Moved `futures_stream_adapter` back to `aws-smithy-http` in 82a04cb.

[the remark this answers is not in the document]
ysaito1001: Moved `futures_stream_adapter` back to `aws-smithy-http` in 82a04cb.

rust-runtime/aws-smithy-http/external-types.toml
ysaito1001: Updated in b8c6029.

rust-runtime/aws-smithy-types/Cargo.toml
ysaito1001: `impl From<hyper::Body> for SdkBody` will stay in `aws-smithy-types` but with a feature gate. By default, there is no conversion available for `hyper::Body` -> `SdkBody`; for those who want that conversion, they need to opt-in to enable it by that cargo feature. Once hyper goes 1.0, we can add a new conversion backwards-compatible manner.

We keep track via https://github.com/awslabs/smithy-rs/issues/3033, and will work on it right after this PR.

rust-runtime/aws-smithy-types/external-types.toml
ysaito1001: `smithy-checksums` uses it, and this needs to be feature-gated as well. Basically the same as https://github.com/awslabs/smithy-rs/pull/3026#discussion_r1351262625.

rust-runtime/aws-smithy-types/Cargo.toml
ysaito1001: I've searched for it, but did not find a likely feature. Let me know if you recall it.

rust-runtime/aws-smithy-http/Cargo.toml
ysaito1001: Made it an optional dependency in aae39be

rust-runtime/aws-smithy-http/Cargo.toml
```diff
@@ -11,8 +11,9 @@ license = "Apache-2.0"
 repository = "https://github.com/awslabs/smithy-rs"
 
 [features]
-rt-tokio = ["dep:tokio-util", "dep:tokio", "tokio?/rt", "tokio?/fs", "tokio?/io-util", "tokio-util?/io"]
-event-stream = ["aws-smithy-eventstream"]
+hyper014 = ["dep:hyper"]
+event-stream = ["aws-smithy-eventstream", 'hyper014']
```
jdisanti, line 15: I don't think event stream actually uses hyper, but it's pulling in `http-body`'s `Body` trait via the hyper crate. If you fix that import, you can get rid of it pulling in the hyper feature.

rust-runtime/aws-smithy-http/src/lib.rs
```diff
@@ -25,7 +25,11 @@
 #![allow(clippy::derive_partial_eq_without_eq)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-pub mod body;
+//TODO(runtimeCratesVersioningCleanup): Re-point those who use the following reexports to
+// directly depend on `aws_smithy_types` and remove the reexports below.
+pub use aws_smithy_types::body;
+pub use aws_smithy_types::byte_stream;
```
jdisanti, line 31: Should add `#[deprecated` to these with a note explaining where they moved to.

jdisanti (commented):
Looks good. Just two smaller things.

rust-runtime/aws-smithy-http/src/lib.rs
ysaito1001: Looks like deprecating re-exports is not supported (https://github.com/rust-lang/rust/issues/30827). Maybe changelog entries are a good alternative place?

rust-runtime/aws-smithy-http/Cargo.toml
ysaito1001: Addressed in c1cbe28.

rust-runtime/aws-smithy-http/src/lib.rs
ysaito1001: Type aliases work instead, e.g.
```
/// Stream of binary data
#[deprecated(note = "Moved to `aws_smithy_types::byte_stream::ByteStream`.")]
pub type ByteStream = aws_smithy_types::byte_stream::ByteStream;
```

However, adding this as part of this PR will add noise everywhere during compilation:
```
warning: use of deprecated type alias `byte_stream::ByteStream`: Moved to `aws_smithy_types::byte_stream::ByteStream`.
```

Will immediately follow-up on this PR, updating the existing places of moved types (similar to #3070).

PR #3029, opened 2023-10-05
Port connection poisoning tests to orchestrator and fix issues discovered

jdisanti:
Issues fixed:
- `ConnectionPoisoningInterceptor` should use the `read_after_deserialization` hook so that it's possible to poison modeled transient errors.
- aws-config should enable connection poisoning on its IMDS/ECS/HTTP clients.

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context/wrappers.rs
+++ b/rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context/wrappers.rs
@@ -173,14 +173,6 @@ impl<'a, I, O, E> BeforeDeserializationInterceptorContextMut<'a, I, O, E> {
     pub fn response_mut(&mut self) -> &mut Response {
         expect!(self, response_mut)
     }
-
-    #[doc(hidden)]
-    /// Downgrade this helper struct, returning the underlying InterceptorContext. There's no good
-    /// reason to use this unless you're writing tests or you have to interact with an API that
-    /// doesn't support the helper structs.
-    pub fn inner_mut(&mut self) -> &'_ mut InterceptorContext<I, O, E> {
-        self.inner
-    }
 }
 
 //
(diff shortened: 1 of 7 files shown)


rust-runtime/aws-smithy-runtime/tests/reconnect_on_transient_error.rs
```diff
@@ -0,0 +1,302 @@
...
+    feature = "client",
+    feature = "wire-mock",
+    feature = "connector-hyper-0-14-x",
+))]
+
+use ::aws_smithy_runtime::client::retries::classifier::{
+    HttpStatusCodeClassifier, SmithyErrorClassifier,
+};
+use ::aws_smithy_runtime_api::client::retries::RetryClassifiers;
+use aws_smithy_async::rt::sleep::TokioSleep;
```
rcoh, line 16: I'd expect the matching version of this test to be deleted?

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context/wrappers.rs
```diff
@@ -206,6 +198,14 @@ impl<'a, I, O, E> AfterDeserializationInterceptorContextRef<'a, I, O, E> {
...
         expect!(self, output_or_error)
     }
+
+    #[doc(hidden)]
+    /// Downgrade this helper struct, returning the underlying InterceptorContext. There's no good
+    /// reason to use this unless you're writing tests or you have to interact with an API that
+    /// doesn't support the helper structs.
+    pub fn inner(&self) -> &'_ InterceptorContext<I, O, E> {
+        self.inner
+    }
```
rcoh, line 208: clearly we actually need this so maybe it shouldn't be doc hidden?

rcoh (approved):

rust-runtime/aws-smithy-runtime/tests/reconnect_on_transient_error.rs
jdisanti: It will be in a future PR that more broadly cleans up middleware code.

rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context/wrappers.rs
jdisanti: Do you think we should add it to all the interceptor context wrappers then?

jdisanti: Went ahead and added it to all of them.

PR #3031, opened 2023-10-05
Enable the invoke_model_with_response_stream API

rcoh:
This API is a streaming API—I tested it manually, an E2E test an example are forthcoming. I also removed the unused `config` service from the smoketest and added bedrock-runtime.

## Motivation and Context
This API is suppressed by default because it's an event stream API

## Testing
https://gist.github.com/rcoh/ce61429303a56fefd65ba05bd36d28fc works

----


jdisanti (commented):
Why was `config.json` deleted?

rcoh:
>  I also removed the unused config service from the smoketest and added bedrock-runtime.

Velfi (approved):

jdisanti:
It's not unused. In fact, it's testing codegen for awsJson1_1: https://github.com/awslabs/smithy-rs/blob/main/aws/sdk/aws-models/README.md

jdisanti (approved):
Awesome! 🚀

ysaito1001 (approved):

PR #3032, opened 2023-10-05
Move `aws_smithy_http::operation::error::BuildError` into `aws-smithy-types`

ysaito1001:
## Motivation and Context
Takes care of the first part of https://github.com/awslabs/smithy-rs/issues/3054 (the remaining part is denoted as `TODO(runtimeCratesVersioningCleanup)` and until that's done the issue will not be closed).

## Description
This PR moves from `aws-smithy-http::operation::error::{BuildError, SerializationError}` to `aws-smithy-types::error::operation::{BuildError, SerializationError}`.

Within the origin crate, we leave "breadcrumbs" (i.e. reexports) for existing customers of the items above so they are not broken by the move. The breadcrumps will be removed by the part two of the issue.

As part of the move, `aws_smithy_http::endpoint::EndpointPrefix::new` now returns its `crate::endpoint::error::InvalidEndpointError` instead of `crate::operation::error::BuildError` for two reasons:
- `BuildError` is now in a stable crate and for an `InvalidUri` variant to be created, it needed to take a `http::uri::InvalidUri` from a unstable crate, which we cannot expose in public API.
- The new error is more closely related to the endpoint business.

## Testing
Relied on the tests in CI.

## Checklist

----


rcoh (commented):
@jdisanti do you think this makes more sense in `aws-smithy-runtime-api`? I guess we need to define a litmus test of what goes where.

ysaito1001:
> do you think this makes more sense in aws-smithy-runtime-api

Yeah, that distinction is blurry to me. For things like `BuildError` in particular, I don't have a guiding principle of which crate is a better home.

rust-runtime/aws-smithy-types/src/operation.rs
```diff
@@ -0,0 +1,166 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Types for representing the interaction between a service an a client, referred to as an "operation" in smithy.
+//! Clients "send" operations to services, which are composed of 1 or more HTTP requests.
+
+/// Errors for operations
+pub mod error {
```
jdisanti, line 10: This should go under the existing root-level `error` module: `crate::error::operation`

ysaito1001: That arrangement is better at the new home. I seem to have focused too much on preserving the original module path. Addressed in f6bb4ba.

rust-runtime/aws-smithy-http/src/endpoint/error.rs
```diff
@@ -69,23 +70,28 @@ pub struct InvalidEndpointError {
...
+    pub fn endpoint_must_have_scheme() -> Self {
         Self {
             kind: InvalidEndpointErrorKind::EndpointMustHaveScheme,
         }
     }
 
-    pub(super) fn failed_to_construct_authority(
+    /// Construct a build error for an invalid authority
+    pub fn failed_to_construct_authority(
+        authority: String,
```
jdisanti, line 82: `impl Into<String>`?

jdisanti (approved):

rust-runtime/aws-smithy-http/src/endpoint/error.rs
ysaito1001: Updated in 18e99d6.

PR #3037, opened 2023-10-06
Update `toSnakeCase` to better handle plurals, version numbers, and other pathological cases

rcoh:
## Motivation and Context
There are currently a lot of fields in the SDK that are clearly converted to snake-case wrong:
Screenshot 2023-10-09 at 12 00 43 PM
Screenshot 2023-10-09 at 12 02 48 PM

## Description
- Author a new splitWords algorithm
- Add snapshot test
- Add lots of individual tests.

## Testing
- snapshot testing, compared with current behavior.

## Checklist

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Strings.kt
```diff
@@ -8,18 +8,114 @@ package software.amazon.smithy.rust.codegen.core.util
...
 
 /**
  * Double quote a string, e.g. "abc" -> "\"abc\""
  */
 fun String.dq(): String = this.doubleQuote()
 
+private fun String.splitOnWordBoundaries(): List<String> {
+    val out = mutableListOf<String>()
+    // These are whole words but cased differently, e.g. `IPv4`, `MiB`, `GiB`, `TtL`
+    val completeWords = listOf("ipv4", "ipv6", "sigv4", "mib", "gib", "ttl")
```
drganjoo, line 24: nit: Will the algorithm work correctly if we add a word to the list that has a substring of it already in the list? For example, if "sigv4a" is added to the list?

drganjoo (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Strings.kt
rcoh: yeah should work (but sigv4a should probably work properly already because the issue is about when you get to the capital `V`)

jdisanti (approved):

PR #3052, opened 2023-10-10
Use tokio::process to allow timeouts to occur

rcoh:
## Motivation and Context
The existing credentials provider was a DOS risk and didn't obey timeout settings because it used `std::timeout::spawn` but relied on a async-based timeout mechanism.

## Description
- Use `tokio::process` instead
- introduce new cargo feature

## Testing
- unit test

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/credential_process.rs
+++ b/aws/rust-runtime/aws-config/src/credential_process.rs
@@ -3,56 +3,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#![cfg(feature = "credentials-process")]
+
 //! Credentials Provider for external process
 
 use crate::json_credentials::{json_parse_loop, InvalidJsonCredentials, RefreshableCredentials};
+use crate::sensitive_command::CommandWithSensitiveArgs;
 use aws_credential_types::provider::{self, error::CredentialsError, future, ProvideCredentials};
 use aws_credential_types::Credentials;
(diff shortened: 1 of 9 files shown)


Velfi (approved):

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -3,57 +3,20 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#![cfg(feature = "credentials-process")]
```
Velfi, line 6: question because I don't know: Is this strictly equivalent to adding the same flag to the `credential_process` `mod` declaration?

rust-runtime/aws-smithy-runtime/src/test_util/capture_test_logs.rs
```diff
@@ -83,7 +83,8 @@ where
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
         self.buf.lock().unwrap().extend_from_slice(buf);
         if !self.quiet {
-            self.inner.write(buf)
+            self.inner.write_all(buf)?;
```
Velfi, line 86: ugh, `write` vs. `write_all` is such an easy mistake to make.

aws/rust-runtime/aws-config/src/credential_process.rs
rcoh: yeah

PR #3055, opened 2023-10-11
New-type all the orchestrator futures

jdisanti:
This PR replaces the `Future` type-alias that was pointing to `NowOrLater` with a new-type for each trait that returns a future.

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime-api/src/client.rs
+++ b/rust-runtime/aws-smithy-runtime-api/src/client.rs
@@ -3,6 +3,68 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+macro_rules! new_type_future {
+    (
+        doc = $type_docs:literal,
+        pub struct $future_name:ident<$output:ty, $err:ty>,
+    ) => {
+        pin_project_lite::pin_project! {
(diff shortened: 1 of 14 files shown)


rust-runtime/aws-smithy-runtime-api/src/client.rs
```diff
@@ -3,6 +3,68 @@
...
+            pub fn new_boxed(
+                future: std::pin::Pin<
+                    Box<dyn std::future::Future<Output = Result<$output, $err>> + Send>,
+                >,
+            ) -> Self {
+                Self {
+                    inner: aws_smithy_async::future::now_or_later::NowOrLater::new(future),
+                }
+            }
+
```
rcoh, line 48: can we use the downcast trick here?

rcoh (approved):

rust-runtime/aws-smithy-runtime-api/src/client.rs
jdisanti: I don't think it would work so well here since the `Pin<Box<...>>` type would need to be just right, and the chances of that are slim.

PR #3061, opened 2023-10-11
Add lifetimes to async traits that take args by reference

jdisanti:
This PR adds lifetimes to the `IdentityResolver`, `DnsResolver` (renamed to `ResolveDns`), and `EndpointResolver` traits so that lifetime gymnastics aren't needed when implementing those traits. For example, `IdentityResolver::resolve_identity` takes `&ConfigBag` as an argument, which means you have to pull things out of the ConfigBag outside of any returned async block in order for the compiler to be satisfied. This change removes that consideration and makes implementing these traits a lot easier.

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/ecs.rs
+++ b/aws/rust-runtime/aws-config/src/ecs.rs
@@ -752,10 +752,7 @@ mod test {
     }
 
     impl ResolveDns for TestDns {
-        fn resolve_dns<'a>(&'a self, name: &'a str) -> DnsFuture<'a>
-        where
-            Self: 'a,
-        {
+        fn resolve_dns<'a>(&'a self, name: &'a str) -> DnsFuture<'a> {
             DnsFuture::ready(Ok(self.addrs.get(name).unwrap_or(&self.fallback).clone()))
         }
     }
@@ -763,10 +760,7 @@ mod test {
     #[derive(Debug)]
     struct NeverDns;
     impl ResolveDns for NeverDns {
-        fn resolve_dns<'a>(&'a self, _name: &'a str) -> DnsFuture<'a>
-        where
-            Self: 'a,
-        {
+        fn resolve_dns<'a>(&'a self, _name: &'a str) -> DnsFuture<'a> {
             DnsFuture::new(async {
                 Never::new().await;
                 unreachable!()
--- a/aws/rust-runtime/aws-config/src/ecs.rs
+++ b/aws/rust-runtime/aws-config/src/ecs.rs
@@ -50,7 +50,7 @@ use crate::http_credential_provider::HttpCredentialProvider;
 use crate::provider_config::ProviderConfig;
 use aws_credential_types::provider::{self, error::CredentialsError, future, ProvideCredentials};
 use aws_smithy_http::endpoint::apply_endpoint;
-use aws_smithy_runtime_api::client::dns::{DnsResolver, ResolveDnsError, SharedDnsResolver};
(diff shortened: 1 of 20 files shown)


aws/rust-runtime/aws-config/src/ecs.rs
```diff
@@ -751,16 +751,22 @@ mod test {
         }
     }
 
-    impl DnsResolver for TestDns {
-        fn resolve_dns(&self, name: String) -> DnsFuture {
-            DnsFuture::ready(Ok(self.addrs.get(&name).unwrap_or(&self.fallback).clone()))
+    impl ResolveDns for TestDns {
+        fn resolve_dns<'a>(&'a self, name: &'a str) -> DnsFuture<'a>
+        where
+            Self: 'a,
```
Velfi, line 757: If we say `&'a self`, then why do we also need the `where` bound? What's the difference between those two bounds?

ysaito1001 (commented):
At first glance, it seems that a trait method accepting multiple references, with each having a different lifetime, allows for more flexibility for callers, but this PR is saying that's not necessarily the case? Specifically, in cases where these trait methods are used, it's more ergonomic for users if all input references have the same lifetime?

> For example, IdentityResolver::resolve_identity takes &ConfigBag as an argument, which means you have to pull things out of the ConfigBag outside of any returned async block in order for the compiler to be satisfied.

Is it possible to elaborate on this with pseudo-code?

jdisanti:
Having separate lifetimes per argument definitely gives more flexibility. I think there's a balance to be struck between flexibility and usability: going too flexible results in a very lengthy and complex function declaration, while going too simple makes it hard to implement the function body (and potentially callers of it).

I could do something like this to maximize flexibility:
```rust
fn thing<'a, 'b, 'c>(&'a self, &'b SomeArgument) -> SomeFuture<'c>
where
    'a: 'c,
    'b: 'c,
    Self: 'c
```
But then customers have to have all that noise in their implementations, and for not much gain. Since `'a` and `'b` both live as long as `'c`, it's roughly equivalent to just using `'a`, but may grant some extra flexibility to the caller. Since we are the caller (in the orchestrator), I don't think that matters that much.

To give an example of what triggered me doing this in the first place, I had this in the LazyCache:
```rust
impl ResolveCachedIdentity for LazyCache {
    fn resolve_cached_identity(
        &self,
        resolver: SharedIdentityResolver,
        config_bag: &ConfigBag,
    ) -> IdentityFuture {
        // ...

        IdentityFuture::new(async move {
            // ...
        })
    }
}
```

With the above, if I try to use `config_bag` inside of the async block, I'll get an error saying it doesn't live long enough. Additionally, if I try to use `self` in the async block, it has the same error because `&self` won't live as long as the future.

By adding the single lifetime, I can now reference both `self` and `config_bag` inside the async block without any issue.

aws/rust-runtime/aws-config/src/ecs.rs
```diff
@@ -751,16 +751,16 @@ mod test {
         }
     }
 
-    impl DnsResolver for TestDns {
-        fn resolve_dns(&self, name: String) -> DnsFuture {
-            DnsFuture::ready(Ok(self.addrs.get(&name).unwrap_or(&self.fallback).clone()))
+    impl ResolveDns for TestDns {
```
ysaito1001, line 754: 👍 for verb + noun convention.

ysaito1001 (approved):
Thanks for [going into details](https://github.com/awslabs/smithy-rs/pull/3061#issuecomment-1760297862). Makes sense to me.

PR #3062, opened 2023-10-12
Implement an identity cache in aws-smithy-runtime

jdisanti:
This PR adds a `ResolveCachedIdentity` trait, and adapts the existing LazyCredentialsCache implementation to it. It does NOT make this cache used yet, as that will be done as a separate PR to keep code review smaller.

Notable differences from the credentials cache:
- Supports multiple different identity types and identity resolvers
- Applies cache partitioning to the `SharedIdentityResolver` so that operation config overrides of the identity resolver will result in the correct identity being resolved.

----


rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs
```diff
@@ -0,0 +1,671 @@
...
+    /// For example, if the identity is expiring in 15 minutes, the buffer time is 10 seconds,
+    /// and buffer time jitter fraction is 0.2, then buffer time is adjusted to 8 seconds.
+    /// Therefore, any requests made after 14 minutes and 52 seconds will load a new identity.
+    ///
+    /// Defaults to a randomly generated value between 0.0 and 1.0. This setter is for testing only.
+    #[cfg(feature = "test-util")]
+    pub fn buffer_time_jitter_fraction(mut self, buffer_time_jitter_fraction: fn() -> f64) -> Self {
+        self.set_buffer_time_jitter_fraction(Some(buffer_time_jitter_fraction));
+        self
+    }
```
rcoh, line 116: does this need to be pub, even feature gated?

rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs
```diff
@@ -0,0 +1,671 @@
...
+    pub fn build(self) -> SharedIdentityCache {
+        let default_expiration = self.default_expiration.unwrap_or(DEFAULT_EXPIRATION);
+        assert!(
+            default_expiration >= DEFAULT_EXPIRATION,
+            "default_expiration must be at least 15 minutes"
+        );
+        LazyCache::new(
+            self.time_source.unwrap_or_default(),
+            self.sleep_impl.unwrap_or_else(|| {
+                default_async_sleep().expect("no default sleep implementation available")
```
rcoh, line 170: should this come from a runtime plugin / runtime components?

rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs
```diff
@@ -0,0 +1,671 @@
...
+            // Another thread could have inserted the partition before we acquired the lock,
+            // so double check before inserting it.
+            partitions
+                .entry(key)
+                .or_insert_with(|| ExpiringCache::new(self.buffer_time));
+            drop(partitions);
+
+            partition = self.partitions.read().unwrap().get(&key).cloned();
+        }
+        partition.expect("inserted above if not present")
```
rcoh, line 209: because it's so easy to mess up RWLock caches, you may want to consider writing the code in a way where you explictly `drop()` the locks

rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs
```diff
@@ -0,0 +1,671 @@
...
+        if partition.is_none() {
+            let mut partitions = self.partitions.write().unwrap();
+            // Another thread could have inserted the partition before we acquired the lock,
+            // so double check before inserting it.
+            partitions
+                .entry(key)
+                .or_insert_with(|| ExpiringCache::new(self.buffer_time));
+            drop(partitions);
+
+            partition = self.partitions.read().unwrap().get(&key).cloned();
```
rcoh, line 207: since this pattern only works for append-only caches, it may be worth noting that entries are never removed.

rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs
```diff
@@ -0,0 +1,671 @@
...
+                                .mul_f64((self.buffer_time_jitter_fraction)());
+
+                            // Logging for cache miss should be emitted here as opposed to after the call to
+                            // `cache.get_or_load` above. In the case of multiple threads concurrently executing
+                            // `cache.get_or_load`, logging inside `cache.get_or_load` ensures that it is emitted
+                            // only once for the first thread that succeeds in populating a cache value.
+                            tracing::info!(
+                                "identity cache miss occurred; added new AWS identity (took {:?})",
+                                time.now().duration_since(start_time)
+                            );
```
rcoh, line 299: should probably note the impact of jitter on the buffer time in the docs above.

rust-runtime/aws-smithy-runtime/src/client/identity/cache.rs
```diff
@@ -0,0 +1,87 @@
...
+#[non_exhaustive]
+pub struct IdentityCache;
+
+impl IdentityCache {
+    /// Create an identity cache that does not cache any resolved identities.
+    pub fn no_cache() -> SharedIdentityCache {
+        NoCache.into_shared()
+    }
+
+    /// Configure a lazy identity cache.
```
rcoh, line 70: one more line that identities are lazily loaded when requested would probably be helpful here

rust-runtime/aws-smithy-runtime/src/client/identity/cache.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#[derive(Clone, Debug)]
+struct NoCache;
+
+impl ResolveCachedIdentity for NoCache {
+    fn resolve_cached_identity<'a>(
+        &'a self,
+        resolver: SharedIdentityResolver,
+        config_bag: &'a ConfigBag,
```
rcoh, line 83: do we want to be giving it the entire config bag? should this have just auth params instead?

rust-runtime/aws-smithy-runtime/src/expiring_cache.rs
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use std::future::Future;
```
rcoh, line 6: do we want this to be in runtime-api instead? I think we've hit other issues where this primitive wasn't accessible to the dependency tree.

rcoh (approved):
look good — some potential other ideas in line

rust-runtime/aws-smithy-runtime/src/client/identity/cache.rs
jdisanti: Identity resolution isn't supposed to have access to the auth parameters per the spec.

rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs
jdisanti: It doesn't look like this is even used in tests anymore. I'm going to make it private and `cfg(test)` in case its needed in the future.

rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs
jdisanti: Good call out. I refactored it so that these come in through a `RuntimeComponents` argument on `ResolveIdentity` so that they will work correctly with config overrides.

rust-runtime/aws-smithy-runtime/src/expiring_cache.rs
jdisanti: It looks like `aws-config` is the only other place that uses it, and that depends on `aws-smithy-runtime`. So I think it's fine where it is. We can always move/re-export it later if we need to.

PR #3065, opened 2023-10-12
Rename orchestrator traits to be verbs

jdisanti:
This PR makes the orchestrator traits more idiomatic by making them verbs.

----


jdisanti:
I didn't rename `HttpClient`, `HttpConnector`, `RuntimePlugin`, `AuthScheme`, or `RetryStrategy` since I couldn't think of good names for these, and was worried alternate names for them might make things less clear.

Velfi (approved):

PR #3068, opened 2023-10-13
Minimum throughput body timeouts Pt.1

Velfi:
## Motivation and Context

https://github.com/awslabs/smithy-rs/issues/1562

## Description

This change adds a new body wrapper: The minimum throughput limit wrapper. It tracks the rate that data is being streamed from itself. If that rate falls below some configurable limit, it emits an error instead of the next chunk. This protects users from requests that start quickly but then slow down considerably.

I'd like to get this merged and then figure out the codegen/docs/examples/config part in a separate PR.

## Testing

Tests are included.

----


aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
```diff
@@ -0,0 +1,142 @@
...
+    let body = body.map(move |body| {
+        let time_source = time_source.clone();
+        // Throw an error if the stream sends less than 1 byte per second
+        let minimum_throughput = (1u64, Duration::from_secs(1));
+        SdkBody::from_dyn(aws_smithy_http::body::BoxBody::new(
+            MinimumThroughputBody::new(time_source, body, minimum_throughput),
+        ))
+    });
+    // assert_eq!(255.0, time_source.seconds_since_unix_epoch());
+    // assert_eq!(Duration::from_secs(255), sleep.total_duration());
```
Velfi, line 94: I haven't yet figure out how to retain access to the `sleep_impl` and `time_source` such that I can run these asserts.

jdisanti: Should be able to clone them before calling `into_shared()`.

aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
```diff
@@ -0,0 +1,256 @@
...
+    assert_eq!(255.0, time_source.seconds_since_unix_epoch());
+    assert_eq!(Duration::from_secs(255), sleep.total_duration());
+    assert_eq!(*EXPECTED_BYTES, res);
+}
+
+// One mebibyte of zeroes.
+const LARGE_BODY: &[u8] = &[0u8; 1024 * 1024];
+
+#[tokio::test]
+async fn test_throughput_timeout_for_upload() {
```
Velfi, line 155: This is currently failing due to a hyper error that reminds me of when I was doing the concurrency tests.

```
dispatch failure: io error: error writing a body to connection: Connection reset by peer (os error 54): Connection reset by peer (os error 54) (DispatchFailure(DispatchFailure { source: ConnectorError { kind: Io, source: hyper::Error(BodyWrite, Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" }), connection: Unknown } }))
```

It's like it's dying when the inner body returns an error. I want to figure this out before merging this PR.

aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
```diff
@@ -0,0 +1,209 @@
...
+        SdkBody::from_dyn(aws_smithy_http::body::BoxBody::new(
+            MinimumThroughputBody::new(ts, body, minimum_throughput),
+        ))
+    });
+    let res = body.collect().await;
+
+    match res {
+        Ok(_) => panic!("Expected an error due to slow stream but no error occurred."),
+        Err(e) => panic!("{}", DisplayErrorContext(e)),
+    }
```
jdisanti, line 63: Both panic? What is the passing path?

aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
```diff
@@ -0,0 +1,209 @@
...
+use aws_types::region::Region;
+use aws_types::SdkConfig;
+use bytes::Bytes;
+use once_cell::sync::Lazy;
+use std::convert::Infallible;
+use std::time::{Duration, UNIX_EPOCH};
+
+#[should_panic = "minimum throughput was specified at 2 B/s, but throughput of 1.5 B/s was observed"]
+#[tokio::test]
+async fn test_throughput_timeout_less_than() {
```
jdisanti, line 29: Seems like these tests should be in minimum_throughput.rs. Doesn't look like they have anything to do with S3.

aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
```diff
@@ -0,0 +1,209 @@
...
+                None
+            } else {
+                sleep.sleep(Duration::from_secs(1)).await;
+                Some((
+                    Result::<_, Infallible>::Ok(Bytes::from(vec![state as u8])),
+                    state + 1,
+                ))
+            }
+        }
+    });
```
jdisanti, line 48: Make this a function to reduce duplication.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
```diff
@@ -0,0 +1,230 @@
...
+const LOG_WINDOW_SIZE: usize = 16;
+
+pin_project_lite::pin_project! {
+    /// A body-wrapper that will ensure that the wrapped body is emitting bytes faster than some
+    /// `minimum_throughput`.
+    pub struct MinimumThroughputBody<InnerBody> {
+        #[pin]
+        inner: InnerBody,
+        // A record of when and how much data was read
+        throughput_logs: VecDeque<(SystemTime, u64)>,
```
jdisanti, line 34: How large does this queue get in practice?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
```diff
@@ -0,0 +1,230 @@
...
+        // It's possible that this body won't be polled again, so we ensure it'll
+        // wake by spawning a task with a reference to the waker, forcing another
+        // poll.
+        let wake_after = self.minimum_throughput.per_time_elapsed;
+        let waker = cx.waker().clone();
+        let sleeper = self.async_sleep.clone();
+        let _task = tokio::task::spawn(async move {
+            sleeper.sleep(wake_after).await;
+            waker.wake();
+        });
```
jdisanti, line 114: Seems like this task shouldn't get spawned if we're going to return `Poll::ready`. Maybe split out most of this function into a separate function, and then here, call that function and examine the poll result before spawning this task.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
```diff
@@ -0,0 +1,230 @@
...
+        } else {
+            0
+        };
+
+        let now = this.time_source.now();
+        this.throughput_logs.push_back((now, bytes_read));
+        // When the number of logs exceeds the window size, toss the oldest log.
+        if this.throughput_logs.len() > LOG_WINDOW_SIZE {
+            this.throughput_logs.pop_front();
+        }
```
jdisanti, line 129: Seems like it should pop anything that's older than our time window, with the window size being a large upper bound. 16 is probably not high enough? Need to experiment with that number.

aws/sdk/integration-tests/s3/tests/throughput-timeout.rs
Velfi: I replaced this with something that should make more sense.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
Velfi: On my machine, a 100MB download with a tick rate of 1s produces 6700 logs. The download completes in only a few secs.

rust-runtime/aws-smithy-async/src/rt/sleep.rs
```diff
@@ -83,7 +83,7 @@ pub fn default_async_sleep() -> Option<SharedAsyncSleep> {
 /// Future returned by [`AsyncSleep`].
 #[non_exhaustive]
 #[must_use]
-pub struct Sleep(Pin<Box<dyn Future<Output = ()> + Send + 'static>>);
+pub struct Sleep(Pin<Box<dyn Future<Output = ()> + Send + Sync + 'static>>);
```
jdisanti, line 86: Futures shouldn't be `Sync`.

Velfi:
This PR is currently blocked on some HTTP Body changes that @ysaito1001 will soon PR.

Velfi:
The current CI failures are slightly related to Yuki's work to feature gate stuff, I'll resolve them once he gets that work merged.

rust-runtime/aws-smithy-async/src/rt/sleep.rs
Velfi: After talking with Carl, I decided that this future should in fact be sync. While we could wrap this future in a mutex or similar to make it `Sync`, I'm opting to just add the bound. The reasoning is that omitting the `Sync` bound will cause more user pain than including it.

ysaito1001: Requiring a future to be `Sync` can cause some issue when used with an async block ([lang forum](https://users.rust-lang.org/t/future-is-not-sync-when-calling-async-fn-of-dyn-trait-obj-in-another-async-fn/83285)), but I don't necessarily see how that bites us here 🤔

At least, I think this PR needs a label `breaking-change` because `Sleep::new` now fails to take an async block for certain cases ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=59ab7e7a4f8aead7788484a1b1fb339c) - derived from the above lang forum).

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -0,0 +1,262 @@
...
+impl PartialEq for Throughput {
+    fn eq(&self, other: &Self) -> bool {
+        if self.per_time_elapsed == other.per_time_elapsed {
+            // If both are using the same unit of time, compare bytes per second.
+            self.bytes_per_second() == other.bytes_per_second()
+        } else {
+            // Otherwise, normalize the unit of time and then compare adjusted bytes per second.
+            let normalized_self = self.normalize_to_per_one_second();
+            let normalized_other = other.normalize_to_per_one_second();
+            normalized_self.bytes_per_second() == normalized_other.bytes_per_second()
```
rcoh, line 73: I think we can compare `bytes_per_second` in all cases right? That's already been normalized and compares `per_time_elapsed`

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+}
+
+#[derive(Clone)]
+pub(super) struct ThroughputLogs {
+    max_length: usize,
+    min_elapsed_time: Duration,
+    inner: VecDeque<(SystemTime, u64)>,
+    floating_back: Option<(SystemTime, u64)>,
```
rcoh, line 122: what's floating_back?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -0,0 +1,262 @@
...
+            // If both are using the same unit of time, compare bytes per second.
+            self.bytes_per_second()
+                .partial_cmp(&other.bytes_per_second())
+        } else {
+            // Otherwise, normalize the unit of time and then compare adjusted bytes per second.
+            let normalized_self = self.normalize_to_per_one_second();
+            let normalized_other = other.normalize_to_per_one_second();
+            normalized_self
+                .bytes_per_second()
+                .partial_cmp(&normalized_other.bytes_per_second())
```
rcoh, line 90: same thing here—bytes_per_second is already normalized

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
```diff
@@ -0,0 +1,467 @@
...
+                this.throughput_logs.set_floating_back(now);
+            }
+            // We also schedule a wake up for current task to ensure that
+            // it gets polled at least one more time.
+            cx.waker().wake_by_ref();
+        };
+        this.sleep_fut.replace(sleep_fut);
+
+        // Calculate the current throughput and emit an error if it's too low.
+        let actual_throughput = this.throughput_logs.calculate_throughput()?;
```
rcoh, line 130: is it worth benchmarking this at some point? if we're getting a ton of data how much time do we spend calculating throughput? I wonder if we should _only_ calculate it when the sleep expires, e.g.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
```diff
@@ -0,0 +1,467 @@
...
+            }
+            Poll::Pending => Poll::Pending,
+            // If we've read all the data or an error occurred, then return that result.
+            res => return res,
+        };
+
+        // Check the sleep future to see if it needs refreshing.
+        let mut sleep_fut = this.sleep_fut.take().unwrap_or_else(|| {
+            this.async_sleep
+                .sleep(this.minimum_throughput.per_time_elapsed())
```
rcoh, line 170: this can be simplified—poll_trailers can only be called once. The only thing we need to do here is watch for a timeout, we don't actually need to calculate the size of the trailers. I wonder if we actually need _anything_ in this method at all—once the last read of the body is complete, we essentially want to schedule a timeout for how long it's OK to wait for the trailers to be polled.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
```diff
@@ -0,0 +1,467 @@
...
+        match self {
+            Self::ThroughputBelowMinimum { expected, actual } => {
+                write!(
+                    f,
+                    "minimum throughput was specified at {expected}, but throughput of {actual} was observed",
+                )
+            }
+            Self::TimeTravel(_) => write!(
+                f,
+                "negative time has elapsed while reading the inner body, this is a bug"
```
rcoh, line 221: it isn't really though—we're using SystemTime and not `Instant`. SystemTime doesn't guarantee monotonicity:
> Distinct from the Instant type, this time measurement is not monotonic. This means that you can save a file to the file system, then save another file to the file system, and the second file has a SystemTime measurement earlier than the first. In other words, an operation that happens after another operation in real time may have an earlier SystemTime!

https://doc.rust-lang.org/std/time/struct.SystemTime.html

We should probably try to handle this gracefully or add some sort of `InstantTimeSource` in our bag of tricks

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
```diff
@@ -0,0 +1,467 @@
...
+            ),
+        }
+    }
+}
+
+impl std::error::Error for Error {}
+
+#[cfg(test)]
+mod tests {
```
rcoh, line 230: do any tests attempt to test the trailer reading?

rcoh (commented):
Looks good—a few questions

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
Velfi: I'm delegating the trailer method so I won't test this.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
Velfi: ignore it, I'm removing it in favor of a simpler strategy.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
Velfi: Agreed, just not in this PR.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput.rs
Velfi: Agreed, just not in this PR.

rcoh (approved):

PR #3071, opened 2023-10-14
Simplify default config and add default async sleep

jdisanti:
While implementing identity caching, I noticed we don't have a default async sleep impl wired up for generic clients, which was causing caching to panic in many tests since it needs a sleep impl for timeouts. I likely need to figure out what to do other than panic if there's no sleep impl, but that's a problem for a different PR.

This PR adds a default sleep impl to generic clients, and also simplifies how default config works. Previously, the generated config `Builder::build` method set all the defaults with a series of "if not set, then set default" statements. In this PR, defaults are registered via default ordered runtime plugins.

Additionally, I cleaned up the standard retry strategy:
- The `TokenBucketPartition` didn't appear to be used at all, so I deleted it.
- `StandardRetryStrategy` was taking retry config at construction, which means it isn't possible to config override the retry config unless the strategy is recreated. It now doesn't take any config at all during construction.
- The adaptive retry client rate limiter was created at construction based on retry config at that point in time. This means config overrides wouldn't work with it, so it is also no longer set up at construction time.
- Removed some unused runtime plugins.

## Checklist

----


rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -0,0 +1,92 @@
...
+    let _default: Option<SharedHttpClient> = None;
+    #[cfg(feature = "connector-hyper-0-14-x")]
+    let _default = crate::client::http::hyper_014::default_client();
+
+    StaticRuntimePlugin::new()
+        .with_order(Order::Defaults)
+        .with_runtime_components(
+            RuntimeComponentsBuilder::new("default_http_client_plugin").with_http_client(_default),
+        )
+        .into_shared()
```
rcoh, line 38: nit: refactor our `fn default_plugin(runtime_components: RuntimeComponentsBuilder) { ... }`

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -0,0 +1,92 @@
...
+/// Runtime plugin that provides a default connector.
+pub fn default_http_client_plugin() -> SharedRuntimePlugin {
+    let _default: Option<SharedHttpClient> = None;
+    #[cfg(feature = "connector-hyper-0-14-x")]
+    let _default = crate::client::http::hyper_014::default_client();
+
+    StaticRuntimePlugin::new()
+        .with_order(Order::Defaults)
+        .with_runtime_components(
+            RuntimeComponentsBuilder::new("default_http_client_plugin").with_http_client(_default),
```
rcoh, line 36: I'd consider making this return `Option<Plugin>` instead—it may be a little confusing to have a runtime plugin configured that _doesn't set an HTTP client.

I'd also name the plugin to clarify that it's setting a hyper client.

rust-runtime/aws-smithy-runtime/src/client/retries/client_rate_limiter.rs
```diff
@@ -9,38 +9,11 @@
...
-        Self {
-            rate_limiter: ClientRateLimiter::new(seconds_since_unix_epoch),
-        }
-    }
-}
-
-impl RuntimePlugin for ClientRateLimiterRuntimePlugin {
-    fn config(&self) -> Option<FrozenLayer> {
-        let mut cfg = Layer::new("client rate limiter");
-        cfg.store_put(self.rate_limiter.clone());
```
rcoh, line 38: is this something a customer would want to use directly? Seems like it could be helpful.

rcoh (commented):
just skimmed, need to come back for detailed review

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
jdisanti: Fixed in ea3ae9620768fb652dbffc9c7eeef07cccfc3833

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
jdisanti: Fixed in ea3ae9620768fb652dbffc9c7eeef07cccfc3833

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -0,0 +1,104 @@
...
+    let _default: Option<SharedHttpClient> = None;
+    #[cfg(feature = "connector-hyper-0-14-x")]
+    let _default = crate::client::http::hyper_014::default_client();
+
+    _default.map(|default| {
+        default_plugin("default_http_client_plugin", |components| {
+            components.with_http_client(Some(default))
+        })
+        .into_shared()
+    })
```
rcoh, line 56:
```suggestion
    let default_client: Option<SharedHttpClient> = None;
    #[cfg(feature = "connector-hyper-0-14-x")]
    let default_client = crate::client::http::hyper_014::default_client();

    default_client.map(|client| {
        default_plugin("default_http_client_plugin", |components| {
            components.with_http_client(Some(client))
        })
        .into_shared()
    })
```

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/// Runtime plugin that provides a default time source.
+pub fn default_time_source_plugin() -> Option<SharedRuntimePlugin> {
+    Some(
+        default_plugin("default_time_source_plugin", |components| {
+            components.with_time_source(Some(SystemTimeSource::new()))
+        })
+        .into_shared(),
+    )
```
rcoh, line 76: nonblocking—I wonder if we could figure out at compile time if `SystemTimeSource` was going to work or not 🤔

rcoh (approved):

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/SensitiveOutputDecoratorTest.kt
```diff
@@ -16,6 +17,8 @@ import software.amazon.smithy.rust.codegen.core.testutil.integrationTest
 
 class SensitiveOutputDecoratorTest {
     private fun codegenScope(runtimeConfig: RuntimeConfig): Array<Pair<String, Any>> = arrayOf(
+        "capture_test_logs" to CargoDependency.smithyRuntimeTestUtil(runtimeConfig).toType()
```
ysaito1001, line 20: Thanks for starting to use this, which I did not when adding this test.

ysaito1001 (approved):

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
jdisanti: Probably need to look at the compilation target to determine that.

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
jdisanti: It definitely needs the underscore prefix to avoid compilation warnings.

PR #3072, opened 2023-10-16
Move default endpoint resolver into a runtime plugin

rcoh:
## Motivation and Context
In preparation for inlining the old endpoint traits, I'm cleaning up the endpoint resolver behavior.

## Description

1. The `Config` field of `service_runtime_plugin` was unused to I deleted it
2. Store the endpoint resolver in `RuntimeComponents` instead of in operation config. This allows a lot of complex logic around `ConfigOverride` to be removed.
3. Move the default endpoint resolver into a runtime plugin.
4. Don't have a fallback implementation for `EndpointResolver`, allow final construction of `RuntimeComponents` to fail
5. `ServiceRuntimePlugin` has been changed to `Defaults` so that it doesn't overwrite settings set by the service config

## Testing
Well covered by existing UT / IT
## Checklist

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationCustomization.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationCustomization.kt
@@ -95,8 +95,8 @@ sealed class OperationSection(name: String) : Section(name) {
         override val customizations: List<OperationCustomization>,
         val operationShape: OperationShape,
     ) : OperationSection("AdditionalRuntimePlugins") {
-        fun addServiceRuntimePlugin(writer: RustWriter, plugin: Writable) {
-            writer.rustTemplate(".with_service_plugin(#{plugin})", "plugin" to plugin)
+        fun addClientPlugin(writer: RustWriter, plugin: Writable) {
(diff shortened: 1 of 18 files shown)


jdisanti (approved):

ysaito1001 (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationCustomization.kt
```diff
@@ -95,8 +95,8 @@ sealed class OperationSection(name: String) : Section(name) {
         override val customizations: List<OperationCustomization>,
         val operationShape: OperationShape,
     ) : OperationSection("AdditionalRuntimePlugins") {
-        fun addServiceRuntimePlugin(writer: RustWriter, plugin: Writable) {
-            writer.rustTemplate(".with_service_plugin(#{plugin})", "plugin" to plugin)
+        fun addClientPlugin(writer: RustWriter, plugin: Writable) {
+            writer.rustTemplate(".with_client_plugin(#{plugin})", "plugin" to plugin)
         }
 
         fun addOperationRuntimePlugin(writer: RustWriter, plugin: Writable) {
```
jdisanti, line 102: Nit: rename this to be consistent with the client level function

jdisanti (approved):

PR #3073, opened 2023-10-16
Upgrade nightly used in CI

jdisanti:


jdisanti:
Looks like the CI failure is caused by new rustdoc warnings, which, paired with `-D warnings`, is causing cargo-check-external-types to fail.

aws/rust-runtime/aws-config/src/default_provider/retry_config.rs
```diff
@@ -81,7 +81,7 @@ impl Builder {
         self
     }
 
-    /// Attempt to create a [RetryConfig](aws_smithy_types::retry::RetryConfig) from following sources in order:
+    /// Attempt to create a [RetryConfig] from following sources in order:
```
ysaito1001, line 84:
```suggestion
    /// Attempt to create a [`RetryConfig`] from following sources in order:
```

ysaito1001 (approved):
Thanks for fixing a lot of doc links!

jdisanti:
It looks like https://github.com/smithy-lang/smithy-rs/pull/3298 was unintentionally merged into this PR before it was merged to main. So this PR included the MSRV upgrade as well.

PR #3076, opened 2023-10-17
Re-point those using `ByteStream` and `SdkBody` to smithy-types

ysaito1001:
# Motivation and Context
A follow-up on #3026

## Description
#3026 moved
- `aws_smithy_http::body::{BoxBody, Error, SdkBody}` to `aws_smithy_types::body::{BoxBody, Error, SdkBody}`
- `aws_smithy_http::byte_stream::{AggregatedBytes, ByteStream, error::Error}` to `aws_smithy_types::byte_stream::{AggregatedBytes, ByteStream, error::Error}`

and also left "breadcrumbs", so that customers could still consume updated types from `aws_smithy_http` after the move.

This PR turns breadcrumbs into deprecation messages (via `#[deprecated(...)]`) and updates existing places that used to use moved types from `aws_smithy_http` to directly depend on `aws_smithy_types`.

## Testing
Relied on tests in CI.

## Checklist

----


jdisanti (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -189,8 +189,8 @@ class ServerHttpBoundProtocolTraitImplGenerator(
         "OnceCell" to RuntimeType.OnceCell,
         "PercentEncoding" to RuntimeType.PercentEncoding,
         "Regex" to RuntimeType.Regex,
-        "SmithyHttp" to RuntimeType.smithyHttp(runtimeConfig),
         "SmithyHttpServer" to ServerCargoDependency.smithyHttpServer(runtimeConfig).toType(),
+        "SmithyTypes" to RuntimeType.smithyTypes(runtimeConfig),
```
drganjoo, line 193: nit: I am not 100% sure on how `.resolve` works. But I am assuming, that is the way a type should be referred to in the code generator instead of hard coding it like `#{SmithyTypes}::byte_stream::ByteStream`, which is how we've done in `streamingBodyTraitBounds`.

So instead of `SmithyTypes`, if we keep `"ByteStream" to RuntimeType.smithyTypes(rc).resolve("byte_stream::ByteStream")` in the codegenScope do you think that would resolve into the same generated code?

rcoh: yeah that results in the same generated code

drganjoo (approved):

PR #3080, opened 2023-10-18
Update SDK external types

rcoh:
## Motivation and Context
- https://github.com/awslabs/smithy-rs/issues/3019

## Description
Prune / split out allow external types

## Testing
- CI

----

### DIFF
--- a/aws/sdk/sdk-external-types.toml
+++ b/aws/sdk/sdk-external-types.toml
@@ -24,5 +24,5 @@ allowed_external_types = [
     "aws_smithy_http::result::*",
 
     # Pending future fix
-    "aws_smithy_http::event_stream::receiver::Receiver",
+    "aws_smithy_http::event_stream::*",
 ]
--- a/aws/sdk/sdk-external-types.toml
+++ b/aws/sdk/sdk-external-types.toml
@@ -22,7 +22,6 @@ allowed_external_types = [
 
     # Pending fix in open PRs:
     "aws_smithy_http::result::*",
-    "aws_smithy_http::endpoint::*",
 
     # Pending future fix
     "aws_smithy_http::event_stream::receiver::Receiver",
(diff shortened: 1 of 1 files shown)


aws/sdk/sdk-external-types.toml
```diff
@@ -1,21 +1,30 @@
...
+    "aws_smithy_async::rt::sleep::*",
+    "aws_smithy_async::time::*",
+    "aws_smithy_async::future::pagination_stream::PaginationStream",
+    # only allowed in from impl for presigned request
     "http::request::Request",
-    "http::error::Error",
-    "http::uri::Uri",
-    "http::method::Method",
+
+    ### All below this line need to be fixed:
```
jdisanti, line 14: The event stream one at the very bottom doesn't need to be fixed. It just needs a tooling upgrade.

aws/sdk/sdk-external-types.toml
```diff
@@ -1,21 +1,30 @@
...
-    "aws_smithy_runtime::*",
-    "aws_smithy_runtime_api::*",
     "aws_smithy_types::*",
+    "aws_credential_types::*",
     "aws_types::*",
-
-    "http::header::map::HeaderMap",
-    "http::header::value::HeaderValue",
+    "aws_smithy_runtime_api::*",
+    "aws_smithy_runtime::*",
```
jdisanti, line 7: Since this will be an unstable crate, we should be careful about what gets allowed here.

rcoh: oh good catch, bad copy-paste by me. It's actually unused.

aws/sdk/sdk-external-types.toml
rcoh: the crate won't go 1.0 though—we might want a newtype?

Velfi (approved):

PR #3082, opened 2023-10-19
Update release tooling to handle both stable and unstable crates

ysaito1001:
## Motivation and Context
This PR updates our smithy-rs release tooling to prepare for handling both stable and unstable crates. Merging this PR to `main` does not make any of the runtime crates stable, instead we will just have a supporting mechanism in place, ready to be turned on by implementing `TODO(GA)`.
 
## Description
At a high level, this PR will
- update the `Release smithy-rs` workflow UI to have an additional input `Stable semantic version`:
Screenshot 2023-10-24 at 10 19 07 PM
- split an existing gradle property `smithy.rs.runtime.crate.version` into `smithy.rs.runtime.crate.stable.version` and `smithy.rs.runtime.crate.unstable.version`.
- make `publisher upgrade-runtime-crates-version` take a new, optional (for backwards compatibility) command line argument `--stable-version` to use the said gradle property `smithy.rs.runtime.crate.stable.version`. This will allow the `publisher` to set a value passed via a workflow input `Stable semantic version` to `smithy.rs.runtime.crate.stable.version` in `gradle.properties` (and the same goes for unstable crates).
- update `fixRuntimeCrateVersions` so that it fixes up runtime crate versions based on properties `smithy.rs.runtime.crate.stable.version` and `smithy.rs.runtime.crate.unstable.version`.

**NOTE**
There is a guard in place. When this PR gets merged and then we enter a stable crate version in the `Stable semantic version` text box, it will be overwritten by a version in `Unstable semantic version`, so that 1.x.y version will NOT be published to `crates.io` before GA and that also replicates today's release workflow's behavior (only publishing unstable crates). Just make sure we specify a 0.x.y in the `Untable semantic version` text box because that does get shipped.

### What happens once `TODO(GA)` has been implemented?
Roughly, it will look like this. When we run the `Release smithy-rs` workflow (not a dry-run), providing a stable version (say 1.0.0) and a unstable version (say 0.57.0) in the workflow text boxes,
1. the workflow will create a new release branch `smithy-rs-release-1.x.y`.
2. the workflow will set 1.0.0 to `smithy.rs.runtime.crate.stable.version` and 0.57.0 to `smithy.rs.runtime.crate.unstable.version` in `gradle.properties`.
3. for whatever smithy runtime crates whose package metadata says
```
[package.metadata.smithy-rs-release-tooling]
stable = true
```
their `Cargo.toml` will include `version = 1.0.0` (and `version = 0.57.0` if `smithy-rs-release-tooling` is not specified or if its value is `stable = false`).

4. the workflow will publish smithy runtime crates accordingly to `crates.io`.
5. releasing `aws-sdk-rust` subsequently will render `version = 1.0.0` in `Cargo.toml` for stable AWS runtime crates (again as specified by `smithy-rs-release-tooling`), will render `version = 1.1.0` for SDK crates (we will not go into details here as to why it's not `1.0.0`), and will publish those crates to `crates.io`.
 
## Testing
In a [separate branch](https://github.com/awslabs/smithy-rs/tree/ysaito/stable-and-unstable-crates) that implemented `TODO(GA)`, I verified that
- our internal release pipeline was executed without errors
- a generated AWS SDK had following manifests (showing excerpts from arbitrary crates)
```
[package]
name = "aws-sdk-s3"
version = "1.1.0"
```
```
[package]
name = "aws-smithy-types"
version = "1.0.0"
```
```
[package]
name = "aws-smithy-http"
version = "0.57.0"
```
```
[package]
name = "aws-types"
version = "1.0.0"
```
```
[package]
name = "aws-config"
version = "1.0.0"
```
----
 

.github/workflows/release.yml
```diff
@@ -13,7 +13,7 @@ env:
   rust_version: 1.70.0
 
 name: Release smithy-rs
-run-name: ${{ github.workflow }} ${{ inputs.semantic_version }} (${{ inputs.commit_sha }}) - ${{ inputs.dry_run && 'Dry run' || 'Production run' }}
+run-name: ${{ github.workflow }} ${{ inputs.stable_semantic_version }} ${{ inputs.unstable_semantic_version }} (${{ inputs.commit_sha }}) - ${{ inputs.dry_run && 'Dry run' || 'Production run' }}
```
jdisanti, line 16:
```suggestion
run-name: ${{ inputs.dry_run && 'Dry run' || 'Prod run' }} - ${{ github.workflow }} ${{ inputs.stable_semantic_version }}/${{ inputs.unstable_semantic_version }} (${{ inputs.commit_sha }})
```
The dry run / prod run part always gets cut off in the GitHub UI, and is kind of important, so I suggest rearranging this. Doesn't matter if the commit hash is cutoff since only the first bit of it is needed.

.github/workflows/release.yml
```diff
@@ -23,8 +23,12 @@ on:
           You must use the non-abbreviated SHA (e.g. b2318b0 won't work!).
         required: true
         type: string
-      semantic_version:
-        description: The semver tag that you want to release (e.g. 0.52.1)
```
jdisanti, line 27: Can you update the description on `dry_run` so that it's less confusing in the GitHub UI? Maybe just add "Dry run." to the beginning of it. In fact, basically all of these should start with a field name since the description is used as the label for the form input.

aws/rust-runtime/aws-credential-types/Cargo.toml
```diff
@@ -35,3 +35,6 @@ all-features = true
 targets = ["x86_64-unknown-linux-gnu"]
 rustdoc-args = ["--cfg", "docsrs"]
 # End of docs.rs metadata
+
+[package.metadata.release-tooling]
+stable = true
```
jdisanti, line 40: I think we might be able to consolidate this crate into aws-types soon. Perhaps we don't want to stabilize it.

aws/sdk-codegen/build.gradle.kts
```diff
@@ -36,7 +36,7 @@ val generateAwsRuntimeCrateVersion by tasks.registering {
     val resourcesDir = "$buildDir/resources/main/software/amazon/smithy/rustsdk"
     val versionFile = file("$resourcesDir/sdk-crate-version.txt")
```
jdisanti, line 37: Is this file even used?

buildSrc/src/main/kotlin/CrateSet.kt
```diff
@@ -3,45 +3,57 @@
...
-        "aws-credential-types",
-        "aws-endpoint",
-        "aws-http",
-        "aws-hyper",
-        "aws-runtime",
-        "aws-runtime-api",
-        "aws-sig-auth",
-        "aws-sigv4",
-        "aws-types",
+        Crate("aws-config", STABLE_VERSION_PROP_NAME),
```
jdisanti, line 20: Maintaining stable/unstable in two places may be error prone. Since this is temporary, maybe we just use comments to have these point to each other.

rust-runtime/aws-smithy-async/Cargo.toml
```diff
@@ -31,3 +31,6 @@ all-features = true
 targets = ["x86_64-unknown-linux-gnu"]
 rustdoc-args = ["--cfg", "docsrs"]
 # End of docs.rs metadata
+
+[package.metadata.release-tooling]
```
jdisanti, line 35: we should probably call this `smithy-rs-release-tooling`

jdisanti (commented):
The overall approach looks good to me. I only have minor comments.

.github/workflows/release.yml
ysaito1001: Great suggestion! While the dry-run mode is a default, it's helpful for users to know in which mode a workflow will be executed without getting cut off in the UI.

.github/workflows/release.yml
ysaito1001: Great suggestion. Updated like so in 3c8e69f:

Screenshot 2023-10-24 at 10 19 07 PM

aws/rust-runtime/aws-credential-types/Cargo.toml
ysaito1001: Addressed in 3d090a2.

rust-runtime/aws-smithy-async/Cargo.toml
ysaito1001: Addessed in 913116e.

buildSrc/src/main/kotlin/CrateSet.kt
ysaito1001: 5eaaad0 added comments in manifest files to point to each other.

aws/sdk-codegen/build.gradle.kts
ysaito1001: Good eyes. No it isn't. In fact, the entire `generateAwsRuntimeCrateVersion` isn't used (ran our release pipeline and confirmed it went ok without it). Removed it in 73b1e81.

jdisanti (approved):

tools/ci-build/publisher/src/subcommand/upgrade_runtime_crates_version.rs
```diff
@@ -46,11 +84,9 @@ pub async fn subcommand_upgrade_runtime_crates_version(
 
 fn update_gradle_properties<'a>(
     gradle_properties: &'a str,
-    upgraded_version: &'a semver::Version,
+    upgraded_version: &semver::Version,
+    version_regex: &Regex,
```
drganjoo, line 88: Nit: Instead of accepting a regex from the caller, it may be safer to determine which regex should be used locally, thus minimizing the risk of an incorrect regex being passed in. I suggest introducing an `enum` with two variants, each containing a `semver::Version` as a parameter. Within the function, you can match on these variants to select the appropriate regex. Although the regex should remain static, it shouldn't be passed in by the caller.

drganjoo (approved):

buildSrc/src/main/kotlin/CrateSet.kt
ysaito1001: Since I do agree with you that it is risky, 4add629 added a unit test to ensure that places aware of crates' stabilities are in sync.

buildSrc/src/main/kotlin/CrateSet.kt
```diff
@@ -3,45 +3,57 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+data class Crate(val name: String, val versionPropertyName: String)
+
 object CrateSet {
+    const val STABLE_VERSION_PROP_NAME = "smithy.rs.runtime.crate.stable.version"
+    private const val UNSTABLE_VERSION_PROP_NAME = "smithy.rs.runtime.crate.unstable.version"
```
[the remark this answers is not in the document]
ysaito1001: > Does some part of the release process rely solely on the stable version?

Yes it does, except until I have added the unit in 4add629. After that commit, both of these variables are no longer private since they are used by the test.

tools/ci-build/publisher/src/subcommand/upgrade_runtime_crates_version.rs
```diff
@@ -24,17 +41,38 @@ pub struct UpgradeRuntimeCratesVersionArgs {
...
+        &upgraded_unstable_version,
+        &UNSTABLE_VERSION_REGEX,
+    )
+    .with_context(|| {
+        format!(
+            "Failed to extract the expected runtime crates version from `{:?}`",
+            &args.gradle_properties_path
+        )
+    })?;
+    let updated_gradle_properties = if let Some(stable_version) = &args.stable_version {
```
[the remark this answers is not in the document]
ysaito1001: I have added `TODO` in 49f951a, because erroring out right now makes it always error out due to 8d8fa61.

tools/ci-build/publisher/src/subcommand/upgrade_runtime_crates_version.rs
```diff
@@ -84,21 +120,43 @@ async fn update_gradle_properties_file(
...
+    }
+
+    #[test]
+    fn upgrading_works_with_dummy_stable_version() {
+        let gradle_properties = "smithy.rs.runtime.crate.stable.version=0.0.0-smithy-rs-head";
+        let version = semver::Version::new(1, 0, 3);
+        let updated =
+            update_gradle_properties(gradle_properties, &version, &STABLE_VERSION_REGEX).unwrap();
+        assert_eq!("smithy.rs.runtime.crate.stable.version=1.0.3", updated);
     }
```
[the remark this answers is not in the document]
ysaito1001: Added the test in 3fcea20

tools/ci-build/publisher/src/subcommand/upgrade_runtime_crates_version.rs
ysaito1001: I like the suggestion. Conveniently, we already have an enum `PackageStability` in `smithy_rs_tool_common::package`. The function can take that enum instead of a regex. Updated in 411b21d.

.github/workflows/release.yml
```diff
@@ -92,7 +99,8 @@ jobs:
       id: branch-push
       shell: bash
       env:
-        SEMANTIC_VERSION: ${{ inputs.semantic_version }}
+        # TODO(GA): specify ${{ inputs.stable_semantic_version }}
```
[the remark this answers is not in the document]
ysaito1001: The `SEMANTIC_VERSION` environment variable is used in a script `get-or-create-release-branch`. The script may create a new release branch based on that environment variable. For GA, we don't specify two environment variables for the script because it will create a stable release branch from `inputs.stable_semantic_version` and unstable crates are part of the stable release branch.

jdisanti (approved):

PR #3087, opened 2023-10-23
Fix bug where Custom Auth Schemes were not respected

### LINKED ISSUE #3034: Custom Authentication Scheme Issue
# Description
When creating a custom authentication workflow via the orchestrator using a `RuntimePlugin` per [this discussion post](https://github.com/awslabs/smithy-rs/discussions/2887), the authentication plugin is overwritten by the model default. In my case this default is no auth, so the authentication token headers were never used and my requests were rejected from the server. I am able to work around this using an `Interceptor` per [this snippet](https://github.com/awslabs/smithy-rs/blob/bc418284516975461dc0bee44ea999361f5c95be/aws/rust-runtime/aws-runtime/src/invocation_id.rs#L95-L129), but per office hours discussion this is not the idiomatic way and the issue with `RutimePlugin`s should be resolved.
# Context
## Reproduction
* Create some simple authentication middleware via `RuntimePlugin` per the discussion post linked above
* Using the following test, evaluate the trace logs and examine the authentication workflow and the final (lack of) headers:
```
    struct TestEndpointResolver;

    impl<T> ResolveEndpoint<T> for TestEndpointResolver {
        fn resolve_endpoint(&self, _params: &T) -> aws_smithy_http::endpoint::Result {
            Ok(::aws_smithy_types::endpoint::Endpoint::builder()
                .url(ENDPOINT_URL_STR)
                .property(
                    "authSchemes",
                    vec![::aws_smithy_types::Document::from({
                        let mut out = ::std::collections::HashMap::<
                            String,
                            ::aws_smithy_types::Document,
                        >::new();
                        out.insert("name".to_string(), "schemeid".to_string().into());
                        out
                    })],
                )
                .build())
        }
    }
    #[tokio::test]
    async fn proxy_test() {
        tracing_subscriber::fmt::init();
        let (server, request) = capture_request(None);
        let config = Config::builder()
            .http_connector(server)
            .endpoint_resolver(TestEndpointResolver {})
            .runtime_plugin(super::AuthRuntimePlugin::new())
            .build();
        let client = Client::from_conf(config);

        let _ = dbg!(client.get_client_configuration().send().await);
        let r = request.expect_request();
        dbg!(&r);
        assert!(r.headers().is_empty());
    }
```
Run with `RUST_LOGS=trace cargo test proxy_test`
## Smithy version
v0.56.1
## Relevant modifications
N/A
## Environment/Deployment Peculiarities
Code running on a custom embedded Linux-based stack

rcoh:
## Motivation and Context
- Fixes #3034
## Description
Because AuthSchemeOptions were being registered at the operation level, there was no way for them to be overridden by customer-provided runtime plugins. This moves them into a separate plugin that is added at Client/Default priority.

## Testing
- new unit test

## Checklist

----

### DIFF
--- a/codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpAuthDecoratorTest.kt
+++ b/codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpAuthDecoratorTest.kt
@@ -9,6 +9,7 @@ import org.junit.jupiter.api.Test
 import software.amazon.smithy.rust.codegen.client.testutil.clientIntegrationTest
 import software.amazon.smithy.rust.codegen.core.rustlang.Attribute
 import software.amazon.smithy.rust.codegen.core.rustlang.CargoDependency
+import software.amazon.smithy.rust.codegen.core.rustlang.raw
 import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
 import software.amazon.smithy.rust.codegen.core.smithy.RuntimeConfig
 import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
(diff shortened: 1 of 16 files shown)


Velfi (approved):

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpAuthDecoratorTest.kt
```diff
@@ -98,6 +99,171 @@ class HttpAuthDecoratorTest {
         }
     }
 
+    @Test
+    fun noAuthCustomAuth() {
```
ysaito1001, line 103: Nice!

ysaito1001 (approved):

rcoh:
not actually a breaking change but somehow broke the semver checker

PR #3088, opened 2023-10-23
Feature-gate public use of `http-body` and `hyper` within `aws-smithy-types`

ysaito1001:
## Motivation and Context
Implements https://github.com/awslabs/smithy-rs/issues/3033

## Description
This PR hides behind cargo features the third-party types from `http-body` and `hyper` crates that are used in`aws-smithy-types`' public API. Customers need to opt-in by enabling a cargo feature `http-body-0-4-x` in `aws-smithy-types` to create an `SdkBody` or `ByteStream` using those third-party types. For more details, please see [the upgrade guide](https://github.com/awslabs/smithy-rs/discussions/3089).

As can been seen from code changes, to reduce the surface area where we need to feature-gate things, we have fused the `aws_smithy_types::body::Inner::Streaming` enum variant into `aws_smithy_types::body::Inner::Dyn` variant, thereby removing `SdkBody::from_dyn`.

## Testing

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-types/src/body.rs
+++ b/rust-runtime/aws-smithy-types/src/body.rs
@@ -6,11 +6,10 @@
 //! Types for representing the body of an HTTP request or response
 
 use bytes::Bytes;
-use http::{HeaderMap, HeaderValue};
-use http_body::{Body, SizeHint};
 use pin_project_lite::pin_project;
 use std::error::Error as StdError;
 use std::fmt::{self, Debug, Formatter};
+use std::future::poll_fn;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
@@ -49,7 +48,10 @@ impl Debug for SdkBody {
 }
 
 /// A boxed generic HTTP body that, when consumed, will result in [`Bytes`] or an [`Error`].
-pub type BoxBody = http_body::combinators::BoxBody<Bytes, Error>;
+enum BoxBody {
+    #[cfg(feature = "http-body-0-4-x")]
+    HttpBody04(http_body_0_4::combinators::BoxBody<Bytes, Error>),
+}
 
 pin_project! {
     #[project = InnerProj]
@@ -59,14 +61,9 @@ pin_project! {
             inner: Option<Bytes>
         },
         // A streaming body
-        Streaming {
-            #[pin]
-            inner: hyper::Body
-        },
-        // Also a streaming body
         Dyn {
             #[pin]
-            inner: BoxBody
+            inner: BoxBody,
         },
 
         /// When a streaming body is transferred out to a stream parser, the body is replaced with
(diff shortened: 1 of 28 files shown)


rust-runtime/aws-smithy-types/src/body.rs
```diff
@@ -135,9 +140,14 @@ impl SdkBody {
         }
     }
 
-    fn poll_inner(
+    pub(crate) async fn next(&mut self) -> Option<Result<Bytes, Error>> {
+        let mut me = Pin::new(self);
+        poll_fn(|cx| me.as_mut().poll_next(cx)).await
+    }
+
+    pub(crate) fn poll_next(
```
Velfi, line 148: for `poll_next` and the `is_end_stream` methods below, I wonder if it might be better to define them in another impl block and feature gate the impl block like so:

```rust
#[cfg(any(
    feature = "feature-a",
    feature = "feature-b",
    feature = "feature-c",
    feature = "feature-d"
))]
fn my_fn() {
    todo!();
}
```

That way, we don't render it if it's unusable and we avoid having to turn of lints for specific lines.

Velfi: _(Although, I guess if we always have the `Once` variant then this suggestion doesn't work.)_

Velfi (approved):

rust-runtime/aws-smithy-types/src/body.rs
ysaito1001: Glad to hear you also consider having feature-gated code in a separate module (because I did during development). We will definitely have a new feature-gate when `http-body and `hyper` goes to 1.0.x, so moving `http-body-0-4-x` specific code to separate modules makes sense to me. Addressed in 70ab9d2.

drganjoo (commented):
**API stabilization clarifications:**

When we say "going 1.x," I understand this indicates a commitment to API stabilization. I'd like to clarify what this means in practical terms:

1. Does "API stabilization" imply that we will never remove a function from the API? If a removal does happen, would it necessitate the introduction of a new major version?
2. Are we allowed to add more functions to a crate post stabilization?
3. Is it possible to deprecate functions, or will we avoid that completely?
4. Can feature flags be added or removed after stabilization?
5. Could a current API be placed behind a feature flag in a subsequent minor version after going 1.x?
6. Are we bound to using the same underlying libraries indefinitely, or do we have the flexibility to switch them out? For example, if we're presently utilizing `rustls` internally, would we have the liberty to transition to `native-tls` down the line, as long as there isn't any disruption to the API?

**Specifics on this PR:**

Is it accurate to say that all functionalities associated with `http::Body` are now under the `http-body-0-4-x` flag? Hence, we have renamed functions. For instance, `ByteStream::from_path` to `ByteStream::from_path_0_4` due to the API's reliance on `http::Body`. Are there elements within `ByteStream` that function independently of `http::Body`?

**Minor questions regarding renaming:**

Am I correct in understanding that the reason for renaming `ByteStream::from_body(http::Body)` to `ByteStream::from_body_0_4(http::Body)` is due to concerns about future changes in `http::Body`? If the name remained unchanged, it would suggest our API can always create a `ByteStream` from `http::Body`. Is there a concern that certain features in the `http::Body` 1.x release might prevent us from maintaining this capability, thus making it impossible to produce a `ByteStream` from `http::Body` post 1.x?

1. If `http-body` releases a version 0.5, will we rename the function from `::from_body_0_4` to `::from_body_0_5`? Alternatively, do we anticipate that `http-body` will never release a `0.5`?
2. When `http-body` updates to 1.x, will the `http-body-0-4-x` feature flag remain in place?

Velfi:
## API stabilization clarifications:

> 1.     Does "API stabilization" imply that we will never remove a function from the API? If a removal does happen, would it necessitate the introduction of a new major version?

Removing a function would necessitate a new major version.

> 2.     Are we allowed to add more functions to a crate post stabilization?

Yes.

3.     Is it possible to deprecate functions, or will we avoid that completely?

It is possible to deprecate functions.

4.     Can feature flags be added or removed after stabilization?

They can be added but not removed.

5.     Could a current API be placed behind a feature flag in a subsequent minor version after going 1.x?

No, this would break users. Even it we made it a default feature, users with `no-default-features` would be broken. We could add a `#[cfg(any(feature = old-feature, feature = new-feature))]` if a feature gate already existed.

6.     Are we bound to using the same underlying libraries indefinitely, or do we have the flexibility to switch them out? For example, if we're presently utilizing rustls internally, would we have the liberty to transition to native-tls down the line, as long as there isn't any disruption to the API?

We are only bound in cases where external types are exposed in our public API. Transitioning to new defaults is breaking but we have a proposal from Russell on how to do that in a non-breaking way. Contact him for more info.

## Specifics on this PR

> Is it accurate to say that all functionalities associated with http::Body are now under the http-body-0-4-x flag?

Yes, based on the current version of `http_body` used in the runtime crates.

> Are there elements within ByteStream that function independently of http::Body?

Yes. Any code that doesn't need to interact with the inner body as a v0.4 HTTP body will not be feature-gated.

## Minor questions regarding renaming

> 1. If http-body releases a version 0.5, will we rename the function from ::from_body_0_4 to ::from_body_0_5? Alternatively, do we anticipate that http-body will never release a 0.5?

If we decide compatibility with a new version is important, we'll create functions for it. I don't anticipate a v0.5 version of http-body, but if one releases, we'll consider supporting it. Renaming the function would be a breaking change. We are committed indefinitely to supporting any API that exposes a 3rd-party type.

> When http-body updates to 1.x, will the http-body-0-4-x feature flag remain in place?

Yes, the feature flag will still exist.

## One last thing

The only thing that would cause us to remove the v0.4 stuff is if http-body goes 1.0 before our GA. We can't count on that so we must be prepared to support multiple versions of the API.

PR #3090, opened 2023-10-24
RustType renders types with wrong syntax

82marbag:
The correct syntax is `HashMap::<T, U>` instead of `HashMap<T, U>` and so on.

----


Velfi:
Both work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f530b9045d325bc05d925c7fb0fdda3

Did this formatting cause a specific problem for you?

82marbag:
> Both work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f530b9045d325bc05d925c7fb0fdda3
> 
> Did this formatting cause a specific problem for you?

Ok, I used the one in the Rust docs. Anyway, it doesn't work for maps: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d9cfd082eeb016be7ce251a526527a40

82marbag:
> Both work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f530b9045d325bc05d925c7fb0fdda3
> 
> Did this formatting cause a specific problem for you?

For example: [this test](https://github.com/awslabs/smithy-rs/blob/40f466213b038916f37528c1b51cc75cea11b02a/codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustTypeTest.kt#L77) is not [valid Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a2d8614350f8430a0a59d30e8625d19c)

Velfi:
Fair enough. I think previously we were only using this in the 'type assertion' position and that's why we never ran into the issue. Given that your version works in both places, you get my ✅.

Velfi (approved):

david-perez (approved):

PR #3094, opened 2023-10-25
ServerInstantiator without builder

82marbag:
* Allow `ServerInstantiator` to use writables instead of only static data.
* Build structures without the builder

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
@@ -138,7 +138,7 @@ open class Instantiator(
      * If the shape is optional: `Some(inner)` or `None`.
      * Otherwise: `inner`.
      */
-    private fun renderMember(writer: RustWriter, memberShape: MemberShape, data: Node, ctx: Ctx) {
+    protected fun renderMember(writer: RustWriter, memberShape: MemberShape, data: Node, ctx: Ctx) {
         val targetShape = model.expectShape(memberShape.target)
         val symbol = symbolProvider.toSymbol(memberShape)
         if (data is NullNode && !targetShape.isDocumentShape) {
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiator.kt
@@ -5,12 +5,17 @@
 
 package software.amazon.smithy.rust.codegen.server.smithy.generators
 
+import software.amazon.smithy.model.node.Node
+import software.amazon.smithy.model.node.ObjectNode
 import software.amazon.smithy.model.shapes.MemberShape
 import software.amazon.smithy.model.shapes.Shape
 import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.rust.codegen.core.rustlang.RustWriter
 import software.amazon.smithy.rust.codegen.core.rustlang.Writable
 import software.amazon.smithy.rust.codegen.core.rustlang.rust
 import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.core.rustlang.withBlock
+import software.amazon.smithy.rust.codegen.core.rustlang.withBlockTemplate
 import software.amazon.smithy.rust.codegen.core.rustlang.writable
 import software.amazon.smithy.rust.codegen.core.smithy.CodegenContext
 import software.amazon.smithy.rust.codegen.core.smithy.RustSymbolProvider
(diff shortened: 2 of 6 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiator.kt
```diff
@@ -68,15 +74,61 @@ class ServerBuilderKindBehavior(val codegenContext: CodegenContext) : Instantiat
...
+        val symbolProvider = codegenContext.symbolProvider
+        writer.withBlock("{", "}") {
+            for (member in shape.members()) {
+                val dataHasValue = data.getMember(member.memberName).isPresent
+                val hasValue = values.containsKey(member) || dataHasValue
+                check(member.isOptional || hasValue)
+                val value = writable {
+                    values[member]
+                        ?.let { it(this) }
+                        ?: super.renderMember(
```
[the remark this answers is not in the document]
david-perez: We want to not use builders when we're using `data` `ObjectNode`s too.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiatorTest.kt
```diff
@@ -233,4 +242,98 @@ class ServerInstantiatorTest {
...
+        val sut = ServerInstantiator(codegenContext)
+
+        val project = TestWorkspace.testProject(model)
+        inner.renderWithModelBuilder(model, symbolProvider, project)
+        nestedStruct.renderWithModelBuilder(model, symbolProvider, project)
+        project.moduleFor(inner) {
+            val nestedUnion = model.lookup<UnionShape>("com.test#NestedUnion")
+            UnionGenerator(model, symbolProvider, this, nestedUnion).render()
+
+            unitTest("writable_for_maps") {
```
david-perez, line 259: Not sure I follow why this test needs to be so complex (maps within maps) to test that we're interpolating a very simple writable `Some(result)`.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiatorTest.kt
```diff
@@ -233,4 +242,98 @@ class ServerInstantiatorTest {
...
+            unitTest("writable_preferred_over_node_data") {
+                val data = Node.parse(
+                    """
+                    {
+                        "str": "world",
+                        "num": 1
+                    }
+                    """,
+                )
+                val writables = mapOf(
```
david-perez, line 312: Again, I don't really follow why we need to create a struct to use a writable that pulls out its `.n` field for the `num` field of `NestedStruct`.

Like, shouldn't it be enough to interpolate a simple writable that e.g. writes a number to test that indeed we're using the writable?

david-perez (changes requested):

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiatorTest.kt
82marbag: Right, this is because I was still falling back to the previous behavior. I've removed it now

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerInstantiator.kt
```diff
@@ -68,15 +68,15 @@ class ServerBuilderKindBehavior(val codegenContext: CodegenContext) : Instantiat
...
-        customizations = listOf(ServerAfterInstantiatingValueConstrainItIfNecessary(codegenContext)),
-    )
+class ServerInstantiator(codegenContext: CodegenContext) : Instantiator(
+    codegenContext.symbolProvider,
+    codegenContext.model,
+    codegenContext.runtimeConfig,
+    ServerBuilderKindBehavior(codegenContext),
+    defaultsForRequiredFields = true,
+    customizations = listOf(ServerAfterInstantiatingValueConstrainItIfNecessary(codegenContext)),
+    constructPattern = InstantiatorConstructPattern.DIRECT,
```
david-perez, line 78: Nice.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -89,13 +90,19 @@ open class Instantiator(
...
     private val customizations: List<InstantiatorCustomization> = listOf(),
+    private val constructPattern: InstantiatorConstructPattern = InstantiatorConstructPattern.BUILDER,
 ) {
     data class Ctx(
         // The `http` crate requires that headers be lowercase, but Smithy protocol tests
         // contain headers with uppercase keys.
         val lowercaseMapKeys: Boolean = false,
     )
 
+    enum class InstantiatorConstructPattern {
```
david-perez, line 101: Docs would be nice.

david-perez (approved):

PR #3095, opened 2023-10-25
Build services with a derived configuration object

david-perez:
This is a pared-down version of #2809.

This is the code-generated version of the
david-perez/smithy-rs-service-config#2 POC.

This introduces a code-generated `PokemonServiceConfig` object which is
provided to the service builder constructor via
`PokemonService::builder(config: PokemonServiceConfig)`. This will displace the
current `builder_without_plugins` and `builder_with_plugins`, deprecating them.
We will eventually remove them.

The motivation is to have a single place where to register plugins and layers.
Smithy traits can leverage this object to inject methods that can apply plugins
and layers. For example, an `@authorization` trait implementer could vend a
smithy-rs decorator that pre-applied an authorization plugin inside the
`${serviceName}Config` object, possibly exposing a method to pass in any
runtime arguments needed to configure the middleware.

## Checklist

----


david-perez:
Upgrade guidance discussion: https://github.com/awslabs/smithy-rs/discussions/3096

david-perez:
Observations:

- We will make the config builder fallible when we inject methods to configure pre-applied middleware that can fail.
- Extensions traits on `HttpPlugins`/`ModelPlugins` are nice because they expose syntax sugar to push plugins. Said syntax sugar is not available in the config builder, because the config builder is code-generated.

david-perez:
> We will make the config builder fallible when we inject methods to configure pre-applied middleware that can fail.

Actually, I'll make the config builder always fallible to enforce that users register layers -> HTTP plugins -> model plugins in that order. When an injected method to configure pre-applied middleware is used, that method will return a `Result`.

---

There was also an idea by @hlbarber on how to enforce plugin dependencies at runtime we should look follow-up on: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=675785fd01852a155dced99d182f3f02

david-perez:
> Actually, I'll make the config builder always fallible to enforce that users register layers -> HTTP plugins -> model plugins in that order.

Injected methods may apply a combination of layers and plugins, which makes it tricky to say when in the chain they should appear. I guess the best option is to make them appear as early as possible (e.g. if a method applies a HTTP plugin and a model plugin, it should be invoked after the layers and before the model plugins).

82marbag (approved):

PR #3099, opened 2023-10-25
Small tweaks to the HTTP wrapper

rcoh:
## Motivation and Context
- small improvements / consistency updates to http wrapper
- fix bug where `.insert` was called but it should have been `append`

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
+++ b/rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
@@ -380,7 +380,7 @@ impl Headers {
         key: impl AsHeaderComponent,
         value: impl AsHeaderComponent,
     ) -> Result<Option<String>, HttpError> {
-        let key = header_name(key.into_maybe_static()?)?;
+        let key = header_name(key)?;
         let value = header_value(value.into_maybe_static()?)?;
         Ok(self
             .headers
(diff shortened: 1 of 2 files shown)


rust-runtime/aws-smithy-runtime/src/client/orchestrator/endpoints.rs
```diff
@@ -132,7 +132,7 @@ fn apply_endpoint(
     for (header_name, header_values) in endpoint.headers() {
         request.headers_mut().remove(header_name);
         for value in header_values {
-            request.headers_mut().insert(
+            request.headers_mut().append(
```
ysaito1001, line 135: would be nice to have a unit test for the use of `append` if we don't have one.

ysaito1001 (approved):

rust-runtime/aws-smithy-runtime-api/src/client/http/request.rs
```diff
@@ -404,8 +404,10 @@ impl Headers {
     /// Removes all headers with a given key
     ///
     /// If there are multiple entries for this key, the first entry is returned
```
Velfi, line 406: Should we be more explicit and add something like "all other entries are discarded" or "to get all values for a given key, keep calling this until it returns None"?

PR #3103, opened 2023-10-27
Move `result` and `connection` from `aws-smithy-http` to `aws-smithy-runtime-api`

ysaito1001:
## Motivation and Context
Implements #3092 

## Description
This PR moves the `connection` and  `result` modules from `aws-smithy-http` to `aws_smithy_runtime_api`. `SdkError` is primarily used within the context of the orchestrator, hence `aws_smithy_runtime_api` instead of `aws-smithy-types`.

Like the previous sibling PRs, type aliases for affected pub items are left in `aws_smithy_http` for backwards compatibility. However, due to lack of trait aliases, a trait `CreateUnhandledError` needs to be consumed from `aws_smithy_runtime_api`.

This PR also updates existing places that consumed those moved types so they now depend on `aws-smithy-runtime-api` to do so.

## Testing
Relied on the existing tests, which ensured no deprecated warnings of using moved types from the old place were issued.

## Checklist

----


rust-runtime/aws-smithy-runtime-api/src/client/result.rs
```diff
@@ -0,0 +1,657 @@
...
+    }
+
+    /// Return a reference to this error's raw response, if it contains one. Otherwise, return `None`.
+    pub fn raw_response(&self) -> Option<&R> {
+        match self {
+            Self::ServiceError(inner) => Some(inner.raw()),
+            Self::ResponseError(inner) => Some(inner.raw()),
+            _ => None,
+        }
+    }
```
Velfi, line 438: This is a random chore, feel free to ignore.
We're specifying variants in several of these methods differently. If you wouldn't mind, could you pick one style and update things?

Velfi (approved):

rust-runtime/aws-smithy-runtime-api/src/client/result.rs
ysaito1001: Thanks for pointing out! Yes, they look arbitrary. 8f5ebfe updated match arms in the file so they are consistently prefixed with `SdkError::`.

PR #3104, opened 2023-10-27
Allow custom writables in Instantiator

82marbag:
`Instantiator` currently renders only static data.
This PR allows it to customize this behavior to allow non-static data (e.g. data taken from other structs).

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
@@ -91,6 +91,7 @@ open class Instantiator(
     private val defaultsForRequiredFields: Boolean = false,
     private val customizations: List<InstantiatorCustomization> = listOf(),
     private val constructPattern: InstantiatorConstructPattern = InstantiatorConstructPattern.BUILDER,
+    private val customWritable: CustomWritable = CustomWritable(),
 ) {
     data class Ctx(
         // The `http` crate requires that headers be lowercase, but Smithy protocol tests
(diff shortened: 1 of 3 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -122,28 +123,50 @@ open class Instantiator(
         fun doesSetterTakeInOption(memberShape: MemberShape): Boolean
     }
 
+    /**
+     * Customize how each shape is rendered, instead of relying on static `Node` data
+     */
+    open class CustomWritable {
```
david-perez, line 129: Should this be an interface?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -122,28 +123,50 @@ open class Instantiator(
         fun doesSetterTakeInOption(memberShape: MemberShape): Boolean
     }
 
+    /**
+     * Customize how each shape is rendered, instead of relying on static `Node` data
```
david-perez, line 127:
```suggestion
     * Customize how each shape is rendered, instead of relying on static `Node` data.
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -122,28 +123,50 @@ open class Instantiator(
...
+     */
+    open class CustomWritable {
+        // Return `null` for the default behavior
```
david-perez, line 130:
```suggestion
        // Return `null` to rely on the default behavior, which uses the static `Node` data.
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -122,28 +123,50 @@ open class Instantiator(
...
+        open fun render(shape: Shape): Writable? = null
```
david-perez, line 131: We use `render` when we actually render code, so, when we have a `RustWriter` and we write it. We use `generate` or other words when we return a `Writable` that gets composed and rendered afterwards.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -284,13 +311,15 @@ open class Instantiator(
         shape: StructureShape,
         data: ObjectNode,
         headers: Map<String, String>,
+
```
david-perez, line 314:
```suggestion
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -309,6 +338,7 @@ open class Instantiator(
...
```
david-perez, line 341:
```suggestion
```

david-perez (approved):

PR #3105, opened 2023-10-27
Use re-exported types in examples instead of aws-smithy-types

drganjoo:
Some of the examples were using types from `aws-smithy-types` instead of the re-exported ones from the generated SDKs. As a guideline, it is preferred to use re-exported types rather than directly depending on underlying crates.


david-perez (approved):
Thanks!

PR #3106, opened 2023-10-29
Add lifetimes in structure impl

82marbag:
----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt
@@ -133,7 +133,8 @@ open class StructureGenerator(
         if (accessorMembers.isEmpty()) {
             return
         }
-        writer.rustBlock("impl $name") {
+        val lifetimes = lifetimeDeclaration()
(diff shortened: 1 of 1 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt
```diff
@@ -133,7 +133,8 @@ open class StructureGenerator(
         if (accessorMembers.isEmpty()) {
             return
         }
-        writer.rustBlock("impl $name") {
+        val lifetimes = lifetimeDeclaration()
```
drganjoo, line 136: nit: How about we call this the `genericParams`? Just to be in line with the official [specs](https://doc.rust-lang.org/stable/reference/items/implementations.html).

drganjoo (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt
82marbag: These are specifically lifetimes (from [the docs](https://doc.rust-lang.org/stable/reference/items/generics.html) you linked)

PR #3107, opened 2023-10-30
Test decorators in server tests

82marbag:
This change allows to test decorators by adding them to the CodegenContext.

----


david-perez (approved):
LGTM.

Shouldn't the same change be made to `serverTestSymbolProviders`?

I guess we should remove `ServerTestCodegenDecorator` too, it doesn't do anything.

PR #3108, opened 2023-10-30
RFC for forwards compatible errors

rcoh:
## Motivation and Context
This RFC describes a way to allow services to add model previously unmodeled variants in the future without breaking customers.

## Description
[Rendered](https://github.com/awslabs/smithy-rs/blob/errors-forward-compat-rfc/design/src/rfcs/rfc0039_forward_compatible_errors.md)

----


design/src/rfcs/rfc0039_forward_compatible_errors.md
```diff
@@ -0,0 +1,131 @@
...
+    ```rust,ignore
+        match err {
+            GetObjectError::NotFound => { ... },
+            err @ GetObject::Unhandled(_) if err.code() == Some("ValidationException") => { ... }
+        }
+    ```
+
+For `1`, we need to remove the `ProvideErrorMetadata` trait implementation from `Unhandled`. We would expose this isntead through a layer of indirection to enable code generated to code to still read the data.
+
+For `2`, we would deprecate the `Unhandled` variants with a message clearly indicating how this code should be written.
```
jdisanti, line 87: We should document here that this is different from how we do `Unknown` on enums because of the `ProvideErrorMetadata` trait. There's no equivalent for the generated enums.

Also, should we consider deprecating `Unknown` in the enum generator for the same reason? It would at least guide customers in the right direction better since right now it's just completely useless, and thus, confusing.

design/src/rfcs/rfc0039_forward_compatible_errors.md
```diff
@@ -0,0 +1,131 @@
...
+
+Generated code would this use `introspect` when supporting **top-level** `ErrorMetadata` (e.g. for [`aws_sdk_s3::Error`](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/enum.Error.html)).
+
+### Deprecating the Variant
+The `Unhandled` variant will be deprecated to prevent users from matching on it inadvertently.
+
+```rust,ignore
+enum GetObjectError {
+   NotFound(NotFound),
+   #[deprecated("Matching on `Unhandled` directly is a backwards compatibility hazard. Use `err if err.error_code() == ...` instead. See [here](<docs about using errors>) for more information.")]
```
jdisanti, line 116: Nice!

design/src/rfcs/rfc0039_forward_compatible_errors.md
rcoh: > Also, should we consider deprecating Unknown in the enum generator for the same reason?
absolutely. I was thinking the same thing. We can also use it to centralize some of the docs around that

jdisanti (approved):

PR #3109, opened 2023-10-30
Add new MRAP test to S3 canary

Velfi:
This PR also replaces the `2023_09_25` canary with current release *(`2023_10_26`)*

----


tools/ci-cdk/canary-lambda/src/latest/paginator_canary.rs
```diff
@@ -59,7 +59,7 @@ pub async fn paginator_canary(client: ec2::Client, page_size: usize) -> anyhow::
 
 #[cfg(test)]
 mod test {
-    use crate::latest::paginator_canary::paginator_canary;
+    use super::paginator_canary;
```
Velfi, line 62: This import seemed odd to me so I fixed it.

tools/ci-cdk/canary-lambda/src/latest/s3_canary.rs
```diff
@@ -82,38 +87,90 @@ pub async fn s3_canary(client: s3::Client, s3_bucket_name: String) -> anyhow::Re
...
+        .context("s3::PutObject[MRAP]")?;
+
+    // Get the test object and verify it looks correct
+    let output = client
+        .get_object()
+        .bucket(&s3_mrap_bucket_arn)
+        .key(&test_key)
+        .send()
+        .await
+        .context("s3::GetObject[MRAP]")?;
```
rcoh, line 149: should we override the region or something to make sure we're really testing multi-region?

rcoh (commented):
only thing that's annoying is it gets one step harder to run the canary locally since you need to set that env variable. I wonder if `local` mode should fill in nonsense defaults or something (at least so it doesn't panic)

tools/ci-cdk/canary-lambda/src/latest/s3_canary.rs
Velfi: Good idea. I've added that.

tools/ci-cdk/canary-lambda/src/latest/s3_canary.rs
```diff
@@ -82,39 +88,104 @@ pub async fn s3_canary(client: s3::Client, s3_bucket_name: String) -> anyhow::Re
...
-        .into_bytes();
-    if std::str::from_utf8(payload.as_ref()).context("s3 payload")? != "test" {
-        result = Err(CanaryError("S3 object body didn't match what was put there".into()).into());
-    }
+        .context("s3::DeleteObject")?;
+
+    // Return early if the result is an error
+    result?;
+
+    // We deliberately use a region that doesn't exist here so that we can
```
jdisanti, line 132: Nit: make a new function for this new part of the test

jdisanti (approved):

PR #3114, opened 2023-10-30
Avoid exposing `aws_smithy_http::event_stream::receiver::Receiver` in SDK's public API 

ysaito1001:
## Motivation and Context
Implements #3100 

## Description
Currently, we expose `aws_smithy_http::event_stream::Receiver` in generated SDKs, as shown in the following S3's example (see[ a generated diff](https://d2luzm2xt3nokh.cloudfront.net/codegen-diff/cc303ab1a07693ab02d5ec4f06101b628d1dbabe/1aa6a8da7d2b7669ba3ab7179a1fd72aadc03162/aws-sdk-ignore-whitespace/index.html) for `tmp-codegen-diff/aws-sdk/sdk/s3/src/operation/select_object_content/_select_object_content_output.rs`):
```
pub struct SelectObjectContentOutput {
    <p>The array of results.</p>
    pub payload: ::aws_smithy_http::event_stream::Receiver<
        crate::types::SelectObjectContentEventStream,
        crate::types::error::SelectObjectContentEventStreamError,
    >,
...
```

This PR wraps `Receiver` in a new-type, called `EventReceiver`, which then supports `pub async fn recv` method whose signature is the same as `aws_smithy_http::event_stream::Receiver::recv`.

## Testing
Relied on existing tests (e.g. `s3` and `transcribestreaming` integration tests cover uses cases affected by this change).

## Checklist

----


rust-runtime/inlineable/src/event_stream_receiver.rs
```diff
@@ -0,0 +1,30 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+use aws_smithy_http::event_stream::Receiver;
+use aws_smithy_runtime_api::box_error::BoxError;
+
+#[derive(Debug)]
+#[non_exhaustive]
```
jdisanti, line 10: non_exhaustive isn't needed since there's a private field

rust-runtime/inlineable/src/event_stream_receiver.rs
```diff
@@ -0,0 +1,30 @@
...
+/// Receives messages out of an Event Stream.
+pub struct EventStreamReceiver<T, E> {
```
jdisanti, line 12: What do you think about calling this `EventReceiver` since it receives individual events at a time rather than an entire stream at a time?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -254,15 +255,25 @@ class HttpBindingGenerator(
...
             "SdkBody" to RuntimeType.sdkBody(runtimeConfig),
             "unmarshallerConstructorFn" to unmarshallerConstructorFn,
+            "receiver" to writable {
+                if (codegenTarget == CodegenTarget.SERVER) {
+                    rust("${outputT.rustType().qualifiedName()}::new(unmarshaller, body)")
+                } else {
+                    rustTemplate(
+                        "#{Wrapper}::new(#{Receiver}::new(unmarshaller, body))",
+                        "Wrapper" to RuntimeType.eventStreamReceiverWrapper(runtimeConfig),
+                        "Receiver" to RuntimeType.eventStreamReceiver(runtimeConfig),
```
jdisanti, line 273: Nit: rename to match the Rust name for easier code searching

jdisanti:
Do you think we should also wrap the `EventStreamSender`? Is there an advantage to that?

It seems like one of the advantages to wrapping the receiver is that we get to properly hide methods such as `try_recv_initial`. Wrapping the sender would allow for adding such methods in the future.

rust-runtime/inlineable/src/event_stream_receiver.rs
ysaito1001: > since it receives individual events at a time

I like that the suggested name implies that. Updated the name in b3f4b86.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
ysaito1001: Addressed that in b3f4b86.

rust-runtime/inlineable/src/event_stream_receiver.rs
ysaito1001: Fixed in ad8bdb8.

ysaito1001:
> Do you think we should also wrap the `EventStreamSender`? Is there an advantage to that?
> 
> It seems like one of the advantages to wrapping the receiver is that we get to properly hide methods such as `try_recv_initial`. Wrapping the sender would allow for adding such methods in the future.

Yeah, I can see value in doing that, benefiting the event sender the same way it benefited the event receiver. Two things:
- It will be `MessageStreamAdapter` that'll be wrapped in a new-type `EventSender`, and `EventSender` will probably have a method `pub async send` whose implementation likely re-uses today's `MessageStreamAdapter::poll_next`.
- Since we currently do not expose `MessageStreamAdapter` in public API within a stable crate, it can be done later outside this PR in a non-breaking fashion.

So all in all, I think it's worth tracking it via an issue. Will create one.

jdisanti (approved):

PR #3126, opened 2023-11-01
Fix aws-sdk-rust#930 (PutSnapshotBlock)

rcoh:
## Motivation and Context
A regression in 0.56 caused `X-Amz-Content-sha256` to be omitted when the `UnsignedPayload` trait was present. This re-enables the old behavior and fixes this operation.

## Description
Enable the `x-amz-content-sha256` header whenever the unsigned payload trait is present.

Old code: https://github.com/awslabs/smithy-rs/blob/smithy-rs-release-0.54.x/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4SigningDecorator.kt#L168

## Testing
- [x] Manual integration test
- [x] Modeled protocol test that ensures that header is set

## Checklist

----


aws/sdk/aws-models/ebs-tests.smithy
```diff
@@ -0,0 +1,35 @@
+$version: "1.0"
+
+namespace com.amazonaws.ebs
+
+use smithy.test#httpRequestTests
+use aws.protocols#restJson1
+use com.amazonaws.ebs#PutSnapshotBlock
+apply PutSnapshotBlock @httpRequestTests([
```
jdisanti, line 8: Can we make a more minimal, focused, test for this to avoid pulling in an additional service? Perhaps a unit test for SigV4AuthDecorator?

jdisanti (approved):

PR #3139, opened 2023-11-02
Move `Message`, `Header`, `HeaderValue`, and `StrBytes` from `aws-smithy-eventstream` to `aws-smithy-types`

ysaito1001:
## Motivation and Context
Fixes the following statement in https://github.com/awslabs/smithy-rs/pull/3114

> A notable breaking change is the error type of the new recv method. Previously, it was [SdkError<E, RawMessage>>](https://docs.rs/aws-smithy-http/0.57.0/aws_smithy_http/event_stream/struct.Receiver.html#method.recv) but it is now a BoxError. This means that upon an event receive error, customers can only show what the error is, but not match on it and take some actions.

so that the `recv` method on a new-type wrapper `aws_smithy_http::event_stream::Receiver` can expose `SdkError<E, RawMessage>>` in public API. It is the responsibility of the above PR to move [RawMessage](https://github.com/awslabs/smithy-rs/blob/c7f0d5dc33937d65841d74b5b828cc77ed1d2db4/rust-runtime/aws-smithy-http/src/event_stream/receiver.rs#L92-L98) from `aws-smithy-http` to `aws-smithy-types` but doing so first requires `Message` to be moved to `aws-smithy-types`, hence this PR.

## Description
Moving `Message` to `aws-smithy-types` also requires moving the types it depends upon, namely `Header`, `HederValue`, and `StrBytes`. However, their serialization/deserialization methods (`read_from` and `write_to`) remain in `aws-smithy-eventstream` and they are converted to free functions, with the aim of moving things as minimally as possible.

`aws-smithy-eventstream` has fuzz testing. It previously implemented the `Arbitrary` trait for moved types, but since they are now in `aws-smithy-types`, we need to have newtypes for them to enable the trait due to the orphan rules. A newly added module `aws-smithy-eventstream::arbitrary` is exactly for that purpose.

## Testing
Relied on the existing tests.

## Checklist

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
```diff
@@ -79,4 +80,16 @@ fun pubUseSmithyPrimitives(codegenContext: CodegenContext, model: Model, rustCra
             "SdkBody" to RuntimeType.smithyTypes(rc).resolve("body::SdkBody"),
         )
     }
+    if (codegenContext.serviceShape.hasEventStreamOperations(model)) {
+        rustTemplate(
+            """
+            pub use #{Header};
+            pub use #{HeaderValue};
+            pub use #{Message};
```
ysaito1001, line 88: Do we also want to reexport `StrBytes` conditionally when a service supports event stream operations?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
```diff
@@ -79,4 +80,16 @@ fun pubUseSmithyPrimitives(codegenContext: CodegenContext, model: Model, rustCra
             "SdkBody" to RuntimeType.smithyTypes(rc).resolve("body::SdkBody"),
         )
     }
+    if (codegenContext.serviceShape.hasEventStreamOperations(model)) {
```
jdisanti, line 83: I think these should get their own `crate::primitives::event_stream` module to group them together.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
jdisanti: Seems reasonable.

rust-runtime/aws-smithy-eventstream/fuzz/fuzz_targets/round_trip.rs
```diff
@@ -4,25 +4,16 @@
...
-            | Error::InvalidHeaderNameLength
-            | Error::TimestampValueTooLarge(_),
-        ) => {}
-        Err(err) => panic!("unexpected error on write: {}", err),
-        Ok(_) => {
-            let mut data = &buffer[..];
-            let parsed = Message::read_from(&mut data).unwrap();
-            assert_eq!(message, parsed);
-        }
+    if write_message_to(&message, &mut buffer).is_ok() {
```
jdisanti, line 14: Why the change in error behavior here?

aws/rust-runtime/aws-sigv4/src/event_stream.rs
```diff
@@ -102,7 +103,7 @@ pub fn sign_message<'a>(
 ) -> Result<SigningOutput<Message>, SigningError> {
     let message_payload = {
         let mut payload = Vec::new();
-        message.write_to(&mut payload).unwrap();
+        write_message_to(message, &mut payload).unwrap();
```
drganjoo, line 106: nit: Just seeking further clarification: Had we introduced a `Writer` trait with a `write_to` method, and a parallel `Reader` trait with a `read_from` function in the `aws_smithy_eventstream::frame` module, could we have implemented these traits for both `Header` and `Message`? This way, had we been able to retain the familiar API calls like `message.write_to`, `header.write_to`, and `Message::from`?

rust-runtime/aws-smithy-eventstream/fuzz/fuzz_targets/round_trip.rs
```diff
@@ -4,25 +4,16 @@
...
+        let mut data = &buffer[..];
+        let parsed = read_message_from(&mut data).unwrap();
+        assert_eq!(message, parsed);
     }
```
drganjoo, line 18: Is this intentional that we no longer `panic!` on an error?

rust-runtime/aws-smithy-eventstream/src/frame.rs
```diff
@@ -22,6 +25,17 @@ const MESSAGE_CRC_LENGTH_BYTES: u32 = size_of::<u32>() as u32;
...
+pub(crate) const TYPE_FALSE: u8 = 1;
+pub(crate) const TYPE_BYTE: u8 = 2;
+pub(crate) const TYPE_INT16: u8 = 3;
+pub(crate) const TYPE_INT32: u8 = 4;
+pub(crate) const TYPE_INT64: u8 = 5;
+pub(crate) const TYPE_BYTE_ARRAY: u8 = 6;
+pub(crate) const TYPE_STRING: u8 = 7;
+pub(crate) const TYPE_TIMESTAMP: u8 = 8;
+pub(crate) const TYPE_UUID: u8 = 9;
+
```
drganjoo, line 38: Just a general question, so we do not allow a float / double as a type in HeaderValue? I guess, they can always be passed as a String.

rust-runtime/aws-smithy-eventstream/fuzz/fuzz_targets/round_trip.rs
jdisanti: I see. The intent of this match was to allow the types of errors that fuzzing would normally encounter by generating an invalid message, but disallow something that appears to be legitimately broken. If we can express that as `is_` functions on `Error`, then that could also work.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
drganjoo: If that's the only type for which a consumer will have to take a dependency on `aws-smithy-types` then we might as well export that too. Unless, that involves exporting a lot more stuff from the `aws-smithy-types::strbytes` mod.

rust-runtime/aws-smithy-types/src/event_stream.rs
```diff
@@ -0,0 +1,185 @@
...
+use crate::str_bytes::StrBytes;
+use bytes::Bytes;
+
+mod value {
+    use crate::str_bytes::StrBytes;
+    use crate::DateTime;
+    use bytes::Bytes;
+
+    /// Event Stream frame header value.
+    #[non_exhaustive]
```
drganjoo, line 17: I thought we only added the `#[non_exhaustive]` for errors? In future, if we introduce a new variant don't we want the consumer code to also break and have them handle the new variant?

Also, does adding a new variant to a non-exhaustive enum considered a breaking change? if yes, then I guess we will have to add `#[non_exhaustive]`.

drganjoo (approved):

aws/rust-runtime/aws-sigv4/src/event_stream.rs
ysaito1001: Yes. Using a verb-style name, if we had the `Write` trait and the `Read` trait, `Message`, `Header`, and `HeaderValue` can each implement their own `write_to` and `read_from`. However, defining the signature for `read_from` has a small twist because `read_header_from` also returns the number of bytes read, whereas `Message` and `HeaderValue` just return `Self`.
```
fn read_header_from<B: Buf>(mut buffer: B) -> Result<(Header, usize), Error>
```
Still, we could implement such traits, but that can be later given `aws-smithy-eventstream` is an unstable crate.

ysaito1001: Yes. Using a verb-style name, if we had the `Write` trait and the `Read` trait, `Message`, `Header`, and `HeaderValue` could each implement their own `write_to` and `read_from`. However, defining the signature for `read_from` has a small twist because `read_header_from` also returns the number of bytes read, whereas `Message` and `HeaderValue` just return `Self` (so we probably need a generic over a return type).
```
fn read_header_from<B: Buf>(mut buffer: B) -> Result<(Header, usize), Error>
```
Still, we could implement such traits, but that can be later given `aws-smithy-eventstream` is an unstable crate.

rust-runtime/aws-smithy-eventstream/fuzz/fuzz_targets/round_trip.rs
ysaito1001: Addressed in cc55893.

rust-runtime/aws-smithy-eventstream/fuzz/fuzz_targets/round_trip.rs
ysaito1001: It should've kept `panic!` on an error. See https://github.com/awslabs/smithy-rs/pull/3139#discussion_r1380583526.

rust-runtime/aws-smithy-types/src/event_stream.rs
ysaito1001: > does adding a new variant to a non-exhaustive enum considered a breaking change? if yes, then I guess we will have to add #[non_exhaustive].

Yes, It's for that reason.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
ysaito1001: Groped them together in c8c293b.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/SmithyTypesPubUseExtra.kt
ysaito1001: `StrBytes` added in reexports in c8c293b.

jdisanti (approved):
Looks great!

PR #3059, opened 2023-10-11
Http updates

rcoh:
## Motivation and Context
- [ ] Upgrade guidance
- [x] changelog

To allow seamless adoption of the incoming `http` crate, we're updating our client crates to use an intermediate type.

## Description

Update _client_ code to use `HttpRequest`. Some internal code still uses `http::Request` but those usages are limited.

## Testing
CI

## Checklist

- [x] I have updated `CHANGELOG.next.toml` if I made changes to the smithy-rs codegen or runtime crates
- [x] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3148, opened 2023-11-07
Create the HTTP `Response` wrapper types

jdisanti:
For the same reason that the request wrapper types were created in #3059, this PR establishes the response wrapper types and refactors existing code to use them.

----


ysaito1001 (approved):

rust-runtime/aws-smithy-runtime-api/src/http/response.rs
```diff
@@ -0,0 +1,245 @@
...
+    body: B,
+    extensions: Extensions,
+}
+
+impl<B> Response<B> {
+    /// Converts this response into an http 0.x response.
+    ///
+    /// Depending on the internal storage type, this operation may be free or it may have an internal
+    /// cost.
+    pub fn into_http02x(self) -> Result<http0::Response<B>, HttpError> {
```
drganjoo, line 78: nit: Wondering if we should call this `try_into_http02x`?

Actually, do we ever return a `HttpError`?

rust-runtime/aws-smithy-runtime-api/src/http/response.rs
```diff
@@ -0,0 +1,245 @@
...
+/// An HTTP Response Type
+#[derive(Debug)]
+pub struct Response<B = SdkBody> {
+    status: StatusCode,
+    headers: Headers,
+    body: B,
+    extensions: Extensions,
+}
+
+impl<B> Response<B> {
```
drganjoo, line 73: I'm uncertain about our handling of streaming responses; however, I presume this change won't impact that, correct?

rust-runtime/aws-smithy-runtime-api/src/http/response.rs
```diff
@@ -0,0 +1,245 @@
...
+
+impl fmt::Display for StatusCode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+/// An HTTP Response Type
+#[derive(Debug)]
+pub struct Response<B = SdkBody> {
```
drganjoo, line 66: Should we implement `Default` for this, or do you believe there is no need?

drganjoo (commented):
I've noticed the use of `expect` and `unwraps` in several places. While this might not pose a significant issue on the client side, it could be catastrophic if such code paths are executed on the server side. Are these instances confined to handling UTF-8 related headers only?

Additionally, what is our strategy for when `http` reaches version 1.x? Will we revert to using the types defined by `http` directly and convert our new `Request`/`Response` types into type aliases? Or do we plan on introducing `into_http1x`?

drganjoo (approved):

aws/rust-runtime/aws-config/src/imds/client.rs
```diff
@@ -636,11 +636,14 @@ pub(crate) mod test {
...
-            .body(SdkBody::from(token))
-            .unwrap()
+        HttpResponse::try_from(
+            http::Response::builder()
+                .status(200)
+                .header("X-aws-ec2-metadata-token-ttl-seconds", ttl)
+                .body(SdkBody::from(token))
+                .unwrap(),
+        )
+        .unwrap()
```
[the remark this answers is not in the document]
jdisanti: `try_from` fails when there a header value is not ASCII or when a header value is invalid UTF-8. My intention was for all these `try_from(...).unwrap()` conversions to only take place in unit tests. In the actual code paths, we should be doing proper error handling for it.

rust-runtime/aws-smithy-runtime-api/src/http/response.rs
jdisanti: This doesn't make any changes to streaming.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
```diff
@@ -34,9 +34,9 @@ class CustomizableOperationGenerator(
                 .resolve("CustomizableOperation"),
             "CustomizableSend" to ClientRustModule.Client.customize.toType()
                 .resolve("internal::CustomizableSend"),
-            "HttpRequest" to RuntimeType.smithyRuntimeApi(runtimeConfig)
+            "HttpRequest" to RuntimeType.smithyRuntimeApiClient(runtimeConfig)
                 .resolve("client::orchestrator::HttpRequest"),
-            "HttpResponse" to RuntimeType.smithyRuntimeApi(runtimeConfig)
+            "HttpResponse" to RuntimeType.smithyRuntimeApiClient(runtimeConfig)
```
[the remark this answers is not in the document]
david-perez: It appears like the types are also available without the `client` feature flag, under the `http` module. I wonder then why we don't use them directly instead of going through the `client` module?

rust-runtime/aws-smithy-http-server/src/protocol/aws_json/rejection.rs
```diff
@@ -26,6 +27,10 @@ pub enum RequestRejection {
     JsonDeserialize(#[from] aws_smithy_json::deserialize::error::DeserializeError),
     #[error("request does not adhere to modeled constraints: {0}")]
     ConstraintViolation(String),
+
+    /// Typically happens when the request has headers that are not valid UTF-8.
```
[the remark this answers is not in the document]
david-perez: We returned a `HeaderParse` rejection:

https://github.com/awslabs/smithy-rs/blob/d4208b2c8bba1c88c0fd3c7fc8f09f10c9346647/rust-runtime/aws-smithy-http-server/src/protocol/rest_json_1/rejection.rs#L129-L132

Which encompasses other failure scenarios, like expecting a header to be present and it isn't.

rust-runtime/aws-smithy-http/src/header.rs
```diff
@@ -132,8 +130,8 @@ fn read_many<T>(
 /// Read exactly one or none from a headers iterator
 ///
 /// This function does not perform comma splitting like `read_many`
-pub fn one_or_none<T: FromStr>(
-    mut values: ValueIter<'_, HeaderValue>,
+pub fn one_or_none<'a, T: FromStr>(
```
david-perez, line 133: (can't leave a comment on line 144). Shouldn't we now no longer need the `std::str::from_utf8` call?

david-perez: In general this file contains several `.as_bytes` calls, and code that works with `&[u8]` when I think it could directly work with `&str` all the way through?

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/http/ResponseBindingGeneratorTest.kt
```diff
@@ -95,21 +96,28 @@ class ResponseBindingGeneratorTest {
         val testProject = TestWorkspace.testProject(symbolProvider)
         testProject.renderOperation()
         testProject.withModule(symbolProvider.moduleForShape(outputShape)) {
-            unitTest(
```
david-perez, line 98: Are we forgoing the `unitTest` style of writing unit tests?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -722,7 +723,15 @@ class ServerHttpBoundProtocolTraitImplGenerator(
             inputShape.serverBuilderSymbol(codegenContext),
         )
         Attribute.AllowUnusedVariables.render(this)
-        rust("let (parts, body) = request.into_parts();")
+        rustTemplate(
+            """
+            let #{RequestParts} { uri, headers, body, .. } = #{Request}::try_from(request)?.into_parts();
+            """,
+            *preludeScope,
```
david-perez, line 730: Do we need `preludeScope` here? We should add it to `codegenScope` instead.

Better yet, I want `rustTemplate` to have the prelude scope baked in, but that's a separate PR.

david-perez: https://github.com/smithy-lang/smithy-rs/issues/3167

rcoh: yes I want this too!

jdisanti: I also want this, but I'll keep it out of this PR. Looks like I can remove the prelude from that rustTemplate call.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/http/ResponseBindingGeneratorTest.kt
jdisanti: At least in my opinion, the `unitTest` function just leads to deeper nesting and doesn't really buy me anything when I have a predetermined test name.

jdisanti:
> Additionally, what is our strategy for when http reaches version 1.x? Will we revert to using the types defined by http directly and convert our new Request/Response types into type aliases? Or do we plan on introducing into_http1x?

These wrapper types are forever, unfortunately. We will just change the insides of them to use http 1 at that time. You can read more in the RFC: https://github.com/smithy-lang/smithy-rs/blob/main/design/src/rfcs/rfc0037_http_wrapper.md

rust-runtime/aws-smithy-http-server/src/protocol/mod.rs
```diff
@@ -52,27 +67,39 @@ pub fn content_type_header_empty_body_no_modeled_input(headers: &HeaderMap) -> R
...
+    headers: &HeaderMap<HeaderValue>,
     expected_content_type: Option<&'static str>,
 ) -> Result<(), MissingContentTypeReason> {
-    if !headers.contains_key(http::header::CONTENT_TYPE) {
-        return Ok(());
+    if let Some(content_type) = headers.get(http::header::CONTENT_TYPE) {
+        let content_type = content_type.to_str().map_err(MissingContentTypeReason::ToStrError)?;
+        content_type_header_classifier(content_type, expected_content_type)
+    } else {
+        Ok(())
```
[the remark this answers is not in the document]
jdisanti: Good question. I don't know the answer to this.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationGenerator.kt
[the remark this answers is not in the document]
jdisanti: Yeah, this makes sense to me. We could probably even get rid of the HttpRequest/HttpResponse types from th eorchestrator module at this point. I think it's not so high priority though, and we can always deprecate them later.

examples/pokemon-service-client-usage/examples/retry-classifier.rs
```diff
@@ -53,7 +53,7 @@ impl ClassifyRetry for SampleRetryClassifier {
             .and_then(|err| err.downcast_ref::<GetServerStatisticsError>())
         {
             if let Some(response) = ctx.response() {
-                if response.status() == StatusCode::SERVICE_UNAVAILABLE {
+                if response.status().as_u16() == StatusCode::SERVICE_UNAVAILABLE.as_u16() {
```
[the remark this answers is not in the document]
jdisanti: Good call. Added in 08f60b0d1d765b406139a7ef080a44685ef1dd37.

rust-runtime/aws-smithy-http-server/src/protocol/aws_json/rejection.rs
```diff
@@ -26,6 +27,10 @@ pub enum RequestRejection {
...
+    #[error("failed to convert request: {0}")]
+    HttpConversion(#[from] HttpError),
```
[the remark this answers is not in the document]
jdisanti: Currently no test.

rust-runtime/aws-smithy-runtime-api/src/http/response.rs
jdisanti: For responses, the error isn't ever returned, but we keep the error as future proofing in the unlikely event that a http1 response can't easily be converted into a http02 response. I do like the idea of adding a `try_` prefix to it. I will change that for the request and the response.

jdisanti: Renamed in bc10bcc6680733f0aa2ed8a0796cc3c97838abe1.

rust-runtime/aws-smithy-http/src/header.rs
jdisanti: Oh, line 144 looks pretty bad. I'll fix that one at the very least. It'll take a good bit of effort to fix the rest of the file though, it looks like.

Touched up in 85af0354cf2a91e4a065a5232b9357ab40e6257f.

PR #3151, opened 2023-11-08
Add support for BehaviorMajorVersions

rcoh:
## Motivation and Context
See [rendered RFC](https://github.com/awslabs/smithy-rs/blob/df518bfb59c7498ab533fac4256ec0b977cee42c/design/src/rfcs/rfc0039_behavior_major_versions.md)

## Description
This add `BehaviorMajorVersions` to the SDK and wires them in up and down the stack.

## Testing
- [x] lots of ITs / UTs

## Checklist

----


design/src/rfcs/rfc0039_behavior_major_versions.md
```diff
@@ -0,0 +1,89 @@
...
+
+For new SDKs, a GA cutline presents difficult choices around timeline and features that can’t be added later without altering behavior.
+
+Both of these use cases are addressed by Behavior Major Versions.
+
+<!-- The "Terminology" section is optional but is really useful for defining the technical terms you're using in the RFC -->
+Terminology
+-----------
+
+- **Some Term**: A definition for that term
```
jdisanti, line 30: Delete it if you're not going to fill it in

design/src/rfcs/rfc0039_behavior_major_versions.md
```diff
@@ -0,0 +1,89 @@
...
+# Cargo.toml
+[dependencies]
+aws-config = { version = "1", features = ["behavior-version-latest"] }
+# OR
+aws-sdk-s3 = { version = "1", features = ["behavior-version-latest"] }
+```
+
+If no `BehaviorMajorVersion` is set, the client will panic during construction.
+
+`BehaviorMajorVersion` is an opaque struct with initializers like `::latest()`, `::v2023_11_09()`. Downstream code can check the version by calling methods like `::supports_v1()`
```
jdisanti, line 52: I don't think the `supports_*()` would work or be useful since the function wouldn't exist in a sufficiently old SDK. I think if we were to support something like this, we would need to have `from_str` and `supports_str` methods on the BMV.

design/src/rfcs/rfc0039_behavior_major_versions.md
```diff
@@ -0,0 +1,89 @@
...
+/// Over time, new best-practice behaviors are introduced. However, these behaviors might not be backwards
+/// compatible. For example, a change which introduces new default timeouts or a new retry-mode for
+/// all operations might be the ideal behavior but could break existing applications.
+#[derive(Debug, Clone)]
+pub struct BehaviorMajorVersion {
+    // currently there is only 1 MV so we don't actually need anything in here.
+}
+```
+
+We also will add `BehaviorVersionLatestRuntimePlugin`, a runtime plugin that is enabled when the `behavior-version-latest` feature is enabled in runtime crates.
```
jdisanti, line 72: Seems to me that these BMV plugins should be aggregate plugins that delegate to a bunch of other default RPs. Then the default RPs would no longer be configured on the client, but rather, just a single BMV plugin. That will cut down a lot on `Arc<dyn>` thrash since the BMV can directly refer to the RP type rather than making it into a trait object.

Separately, are there any use cases for registering a normal or override-level RP with BMVs?

rust-runtime/aws-smithy-runtime-api/src/client/behavior_version.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// This method will always return the latest major version.
+    ///
+    /// This is the recommend choice for customers who aren't reliant on extremely specific behavior
+    /// characteristics. For example, if you are writing a CLI app, the latest behavior major version
+    /// is probably the best setting for you.
+    ///
+    /// If, however, you're writing a service that is very latency sensitive, or that has written
+    /// code to tune Rust SDK behaviors, consider pinning to a specific major version.
+    ///
+    /// The latest version is currently [`BehaviorMajorVersion::v1`]
```
jdisanti, line 31:
```suggestion
    /// The latest version is currently [`BehaviorMajorVersion::v2023_11_09`]
```

rust-runtime/aws-smithy-runtime-api/src/client/behavior_version.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn latest() -> Self {
+        Self {}
+    }
+
+    /// This method returns the behavior configuration for November 9th 2023
```
jdisanti, line 36:
```suggestion
    /// This method returns the behavior configuration for November 9th, 2023
```

rust-runtime/aws-smithy-runtime-api/src/http/request.rs
```diff
@@ -337,7 +337,7 @@ mod test {
             .headers_mut()
             .try_insert("a\nb", "a\nb")
             .expect_err("invalid header");
-        let _ = req.headers_mut().insert("a\nb", "a\nb");
+        //let _ = req.headers_mut().insert("a\nb", "a\nb");
```
jdisanti, line 340: ?

jdisanti (commented):
Overall, I like the direction this is headed. Nice work!

jdisanti:
From offline conversation: We decided the BMV shouldn't be a runtime component nor in the config bag because there should be no way to change it from inside of orchestration. It needs to be a special field in the service config, separate from components and the layer. Then it can be passed in as a constructor arg to the runtime plugins that need it, and those runtime plugins can add things to the config bag to tell components further downstream how to behave.

design/src/rfcs/rfc0039_behavior_major_versions.md
```diff
@@ -0,0 +1,89 @@
...
+-----------
+
+- **Some Term**: A definition for that term
+
+<!-- Explain how users will use this new feature and, if necessary, how this compares to the current user experience -->
+The user experience if this RFC is implemented
+----------------------------------------------
+
+In the current version of the SDK, users can construct clients without indicating any sort of behavior major version.
+Once this RFC is implemented, there will be three ways to set a behavior major version:
```
ysaito1001, line 37:
```suggestion
Once this RFC is implemented, there will be two ways to set a behavior major version:
```

rust-runtime/aws-smithy-runtime-api/src/client/behavior_version.rs
```diff
@@ -0,0 +1,78 @@
...
+    ///
+    /// When a new behavior major version is released, this method will be deprecated.
+    pub fn v2023_11_09() -> Self {
+        Self {}
+    }
+
+    /// Returns whether V1 behaviors are enabled.
```
ysaito1001, line 43:
```suggestion
    /// Returns whether V2023_11_09 behaviors are enabled.
```

ysaito1001 (approved):
Love the idea of pinning when loading defaults.

design/src/rfcs/rfc0039_behavior_major_versions.md
rcoh: sufficently old SDKs won't have the newer code so there's no new versions to suppress

aws/rust-runtime/aws-config/src/default_provider/app_name.rs
```diff
@@ -117,7 +118,7 @@ mod tests {
     #[tokio::test]
     async fn profile_name_override() {
         let fs = Fs::from_slice(&[("test_config", "[profile custom]\nsdk_ua_app_id = correct")]);
-        let conf = crate::from_env()
+        let conf = crate::from_env_with_version(BehaviorMajorVersion::latest())
```
jdisanti, line 121: I think we should rename this to `defaults_with_version`, and `load_from_env_with_version` to `load_defaults_with_version`. In fact, maybe even remove the `_with_version` part.

design/src/rfcs/rfc0040_behavior_major_versions.md
```diff
@@ -0,0 +1,89 @@
...
+/// all operations might be the ideal behavior but could break existing applications.
+#[derive(Debug, Clone)]
+pub struct BehaviorMajorVersion {
+    // currently there is only 1 MV so we don't actually need anything in here.
+}
+```
+
+To help customers migrate, we are including `from_env` hooks that set `behavior-version-latest` that are _deprecated_. This allows customers to see that they are missing the required cargo feature and add it to remove the deprecation warning.
+
+Internally, `BehaviorMajorVersion` will become an additional field on `<client>::Config`. It is _not_ ever stored in the `ConfigBag` or in `RuntimePlugins`.
```
jdisanti, line 74: Would be nice to add how the BMV will be incorporated into RuntimePlugins via constructors and separate config bag values.

rcoh: done

PR #3156, opened 2023-11-08
Ensure identity resolver is set when a credentials provider is given only at operation level

ysaito1001:
## Motivation and Context
Fixes https://github.com/awslabs/aws-sdk-rust/issues/901

## Description
This PR is a rework of https://github.com/awslabs/smithy-rs/pull/3021 whose fix was inadvertently discarded during https://github.com/awslabs/smithy-rs/pull/3077. The way we fix the issue is slightly different. In this PR, we add an identity resolver to runtime components within `set_credentials_provider`, instead of using `ServiceConfig.OperationConfigOverride`.

## Testing
Added a Kotlin integration test to `CredentialProviderConfigTest.kt` based on the customer reported issue.

## Checklist

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialProviders.kt
```diff
@@ -103,16 +107,34 @@ class CredentialProviderConfig(codegenContext: ClientCodegenContext) : ConfigCus
                     *codegenScope,
                 )
 
-                rustTemplate(
+                rustBlockTemplate(
                     """
                     /// Sets the credentials provider for this service
-                    pub fn set_credentials_provider(&mut self, credentials_provider: #{Option}<#{SharedCredentialsProvider}>) -> &mut Self {
-                        self.config.store_or_unset(credentials_provider);
```
jdisanti, line 110: Where was the code that was moving this value from config into components for the service level config? That still should be deleted.

jdisanti (approved):
Looks great!

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialProviders.kt
ysaito1001: Thanks for catching! Deleted in 83125dd.

PR #3157, opened 2023-11-08
Fix panics in `Headers::try_insert` and `Headers::try_append`

jdisanti:
The `try_insert` and `try_append` methods on `Headers` would panic when passing a non-ASCII value for the key, or an invalid UTF-8 value for the value. This was due to them using the http crate's `from_static` method, which panics on these invalid values. This PR makes them always use `try_from` instead.

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime-api/src/http/headers.rs
+++ b/rust-runtime/aws-smithy-runtime-api/src/http/headers.rs
@@ -92,7 +92,7 @@ impl Headers {
     /// This will *replace* any existing value for this key. Returns the previous associated value if any.
     ///
     /// # Panics
-    /// If the key or value are not valid ASCII, this function will panic.
+    /// If the key is not valid ASCII, or if the value is not valid UTF-8, this function will panic.
     pub fn insert(
         &mut self,
         key: impl AsHeaderComponent,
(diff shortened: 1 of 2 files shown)


rust-runtime/aws-smithy-runtime-api/src/http/headers.rs
```diff
@@ -326,13 +330,19 @@ pub use header_value::HeaderValue;
...
+) -> Result<http0::HeaderName, HttpError> {
     name.repr_as_http02x_header_name().or_else(|name| {
         name.into_maybe_static().and_then(|cow| {
             if cow.chars().any(|c| c.is_uppercase()) {
                 return Err(HttpError::new("Header names must be all lower case"));
             }
             match cow {
+                Cow::Borrowed(s) if panic_safe => {
+                    http0::HeaderName::try_from(s).map_err(HttpError::invalid_header_name)
+                }
```
rcoh, line 345: ideally we wouldn't need this and we'd have a cheap way of pre-validating that `from_static` would work, I guess?

rust-runtime/aws-smithy-runtime-api/src/http/headers.rs
```diff
@@ -92,28 +92,31 @@ impl Headers {
     /// This will *replace* any existing value for this key. Returns the previous associated value if any.
     ///
     /// # Panics
-    /// If the key or value are not valid ascii, this function will panic.
+    /// If the key or value are not valid ASCII, this function will panic.
```
rcoh, line 95: the panic conditions are actually more subtle (although maybe we should eschew using `from_static` at all (or only use it if possible, but all the header to be made in the dynamic way?)

we could have the logic be:
- if we have a static value and it's all lower case and ready to go: use `from_static`
- otherwise try to go through the `HeaderName` constructor

seems like this would do the right thing for most use cases. We could emit a trace log if someone gave us a static header and we couldn't make use of it because, e.g. it was upper case.

rcoh (approved):

PR #3173, opened 2023-11-10
Clean up re-exports and improve 'BuildError' conversions

### LINKED ISSUE #3155: Need re-export of SharedCredentialsProvider and `ProvideCredentials`

### LINKED ISSUE #3171: Enable conversion from `BuildError` into `SdkError`
Improve customer experience by allowing `?` to used on `BuildError`

### LINKED ISSUE #3172: Re-export `BuildError`

rcoh:
## Motivation and Context
- Fixes #3171 
- Fixes #3155 
- Fixes #3172 

## Description
- Add `configReExport` to make it easy and consistent to re-export types into the config module when they are used!
- Add this for a bunch of types and clean up some (now) dead code
- Re export `BuildError`
- Enable converting from `BuildError` into `SdkError`

There are probably more places this can be used, but I'd like to keep this change small to facilitate easy backport to 0.57.

## Testing
CI

## Checklist

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientReExports.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientReExports.kt
@@ -1,0 +1,17 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.rust.codegen.client.smithy
+
+import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+
+/** Returns a symbol for a type re-exported into crate::config
+ *  Although it is not always possible to use this, this is the preferred method for using types in config customizations
+ *  and ensures that your type will be re-exported if it is used.
+ */
+fun configReexport(type: RuntimeType): RuntimeType = RuntimeType.forInlineFun(type.name, module = ClientRustModule.config) {
+    rustTemplate("pub use #{type};", "type" to type)
+}
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointConfigCustomization.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointConfigCustomization.kt
@@ -32,7 +32,6 @@ internal class EndpointConfigCustomization(
     private val codegenScope = arrayOf(
         *preludeScope,
         "Params" to typesGenerator.paramsStruct(),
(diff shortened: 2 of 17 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointConfigCustomization.kt
```diff
@@ -90,7 +89,6 @@ internal class EndpointConfigCustomization(
                                 ##[allow(deprecated)]
                                 self.set_endpoint_resolver(
                                     endpoint_url.map(|url| {
-                                        use #{IntoShared};
                                         #{StaticUriEndpointResolver}::uri(url).into_shared()
```
jdisanti, line 92: This compiles? How does it know where `into_shared` came from?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientReExports.kt
```diff
@@ -0,0 +1,17 @@
...
+package software.amazon.smithy.rust.codegen.client.smithy
+
+import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+
+/** Returns a symbol for a type re-exported into crate::config
+ *  Although it is not always possible to use this, this is the preferred method for using types in config customizations
+ *  and ensures that your type will be re-exported if it is used.
+ */
+fun configReexport(type: RuntimeType): RuntimeType = RuntimeType.forInlineFun(type.name, module = ClientRustModule.config) {
```
jdisanti, line 15: In an ideal world, I prefer to have all the re-exports in one place, but I understand the need to make them conditional and how that's easier with this approach. That said, I don't think we should use this for re-exports that are supposed to _always_ be present since then it could be easy to break one of these re-exports just by changing codegen in some other place.

jdisanti (commented):
This is awesome!

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointConfigCustomization.kt
rcoh: it's re-exported above

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientReExports.kt
rcoh: I see that but I think:
1. It's easier to miss items
2. We'll get a semver failure if we delete something by accident

rcoh: The other benefit is that in-situ, we're actually _using_ the rexport which I think is a benefit. We could make non-conditional re-exports be runtime types though, not opposed to that.

jdisanti (approved):

ysaito1001 (approved):

PR #3175, opened 2023-11-10
Fix compilation error in generate code caused by name collision

rcoh:
## Motivation and Context
If you had a model like this:
```smithy
   @http(uri: "/SomeOperation2", method: "GET")
            operation GetThing {
                // input: GetThingInput,
                output: GetThingOutput
            }
```

But then nested in some other API you did something like this:
```smithy
            list GetThings {
                member: GetThingOutput
            }
```

Code would fail to compile because we generated the same method signature for two different types.

## Testing
- [x] fixes minimal reproducer

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/NamingObstacleCourseTestModels.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/NamingObstacleCourseTestModels.kt
@@ -171,6 +171,11 @@ object NamingObstacleCourseTestModels {
         )
     }.toString().asSmithyModel()
 
+    /**
+     * This targets two bug classes:
+     * - operation inputs used as nested outputs
+     * - operation outputs used as nested outputs
+     */
     fun reusedInputOutputShapesModel(protocol: Trait) = """
         namespace test
         use ${protocol.toShapeId()}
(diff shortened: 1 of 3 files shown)


aws/sdk-codegen/src/test/kotlin/SdkCodegenIntegrationTest.kt
```diff
@@ -43,6 +43,30 @@ class SdkCodegenIntegrationTest {
             operation SomeOperation {
                 output: SomeOutput
             }
+
+            // re-use get thing output in a list & in an operation
+            @http(uri: "/SomeOperation2", method: "GET")
```
jdisanti, line 48: Organizationally speaking, this test belongs in codegen-client rather than sdk-codegen. Looks like we also need an equivalent test for inputs.

rcoh: oh right good point—when I started debugging I thought the SDK was involved

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/NamingObstacleCourseTestModels.kt
```diff
@@ -169,4 +170,56 @@ object NamingObstacleCourseTestModels {
...
+        service Service {
+            version: "2006-03-01",
+            operations: [GetThing, GetThingNested]
+        }
+
+        // re-use get thing output in a list & in an operation
+        @http(uri: "/SomeOperation2", method: "POST")
+        operation GetThing {
+            output: GetThingOutput
+            input: GetThingOutput
```
jdisanti, line 189:
```suggestion
            input: GetThingInput
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/NamingObstacleCourseTestModels.kt
```diff
@@ -169,4 +170,56 @@ object NamingObstacleCourseTestModels {
...
+        }
+
+        structure GetThingInput {
+            @required
+            meta: String
+        }
+
+        @http(uri: "/SomeOperation3", method: "POST")
+        operation GetThingNested {
+            input: GetThingADifferentWayOutput
```
jdisanti, line 204:
```suggestion
            input: GetThingADifferentWayInput
```

jdisanti (commented):
I think you intended to use the input structs?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/NamingObstacleCourseTestModels.kt
rcoh: it actually was on purpose but I've cleaned up the names and comments to clarify whats going on

jdisanti (approved):

PR #3179, opened 2023-11-10
Remove cargo doc from all-services check to speed up CI

rcoh:
## Motivation and Context
Running this check on all services significantly slows down CI

----


jdisanti (approved):
Thanks for fixing this!

Issue #3161, opened 2023-11-09
Fix `Debug`/`Display` impl on `DateTime`


rcoh:
Currently, `DateTime` lacks a display impl and the debug impl is basically useless since it logs seconds and subsecond nanos. We should add a `Display` impl that prints out the date in RFC-3339. This also removes the need for a direct dependency on aws_smithy_types to get basic formatting.

subhajit20:
Hey @rcoh I would like to take this,
Can you help out a little that in which .rs file I have to put this impl?

rcoh:
thank you so much! the file is here: https://github.com/awslabs/smithy-rs/blob/32655552c1c1cb825981a8193331c8a2d73ffc9f/rust-runtime/aws-smithy-types/src/date_time/mod.rs#L58

subhajit20:
> thank you so much! the file is here:
> 
> https://github.com/awslabs/smithy-rs/blob/32655552c1c1cb825981a8193331c8a2d73ffc9f/rust-runtime/aws-smithy-types/src/date_time/mod.rs#L58

Should I make a function with the name `Display` inside the DateTime impl?

rcoh:
You'll need to add an implementation of the `Display` trait. Docs here: https://doc.rust-lang.org/rust-by-example/hello/print/print_display.html

jdisanti:
Leaving open to enhance the `Debug` impl. The `Display` impl has been improved already.

mjn298:
I'll handle `Debug`! RFC-3339 for `Debug` as well?

PR #3183, opened 2023-11-12
Implement Display Trait for DateTime

HakanVardarr:
## Motivation and Context

It implements Display trait For DateTime

https://github.com/smithy-lang/smithy-rs/issues/3161

## Description

I implemented Display trait for DateTime

## Testing

I used this test
```rust
    fn test_display_formatting() {
        // Create a DateTime instance for testing
        let datetime = DateTime {
            seconds: 1636761600, // Example timestamp (replace with your actual timestamp)
            subsecond_nanos: 123456789, // Example subsecond nanos (replace with your actual value)
        };

        // Expected RFC-3339 formatted string
        let expected = "2021-11-13T00:00:00.123456789Z";

        // Format the DateTime using Display trait
        let formatted = format!("{}", datetime);

        // Assert that the formatted string matches the expected result
        assert_eq!(formatted, expected);
    }
```

## Checklist

----


HakanVardarr:
I don't know if I need to any updates on CHANGELOG.next.toml so I didn't do anything with it. If it is necessary can you help me on that?

rust-runtime/aws-smithy-types/src/date_time/mod.rs
```diff
@@ -329,6 +330,14 @@ impl Ord for DateTime {
     }
 }
 
+impl Display for DateTime {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let datetime =
+            chrono::NaiveDateTime::from_timestamp_opt(self.seconds, self.subsecond_nanos).unwrap();
+        let formatted = datetime.format("%Y-%m-%dT%H:%M:%S%.fZ");
+        write!(f, "{}", formatted)
```
rcoh, line 338: instead of using chrono, we should use the date time support already built into datetime: https://docs.rs/aws-smithy-types/latest/aws_smithy_types/date_time/struct.DateTime.html#method.fmt

make sure to also remove the chrono dependency

rcoh (changes requested):

rust-runtime/aws-smithy-types/src/date_time/mod.rs
rcoh: we'll also need at least one test

jdisanti:
> I don't know if I need to any updates on CHANGELOG.next.toml so I didn't do anything with it. If it is necessary can you help me on that?

You should add the following:

```
[[aws-sdk-rust]]
message = "Add `Display` impl for `DateTime`."
references = ["smithy-rs#3183"]
author = "HakanVardarr"

[[smithy-rs]]
message = "Add `Display` impl for `DateTime`."
references = ["smithy-rs#3183"]
meta = { "breaking" = false, "tada" = true, "bug" = false, "target" = "all" }
author = "HakanVardarr"
```

HakanVardarr:
Is this better?

rcoh (approved):
Looks good!

Issue #3182, opened 2023-11-11
Error struct member with default implementation cause rs clients to fail on build


codypenta:
When an error member has a default value, it will generate errors with smithy-rs client unless you manually specify `@required`

For example:

```smithy
// Valid Smithy, fails client creation
@error("client")
structure ReservedGroupUsedError {
    message: String = "Cannot use a reserved group name or any variation of"
}
```

```smithy
@error("client")
structure ReservedGroupUsedError {
    @required // This makes the client render again
    message: String = "Cannot use a reserved group name or any variation of"
}
```

I think the code is question is [here](https://github.com/smithy-lang/smithy-rs/blob/1331dc5443e6c3f5494eed2a44606062d3888446/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ProtocolParserGenerator.kt#L163)

I cloned and made this edit below and it worked for me

```kotlin
// If the message member is optional and wasn't set, we set a generic error message.
if (errorMessageMember != null) {
    if (errorMessageMember.isOptional && !errorMessageMember.hasNonNullDefault()) {
        rust(
            """
            if tmp.message.is_none() {
                tmp.message = _error_message;
            }
            """,
        )
    }
}
```

Happy to make a PR if this is the right approach

jdisanti:
To better assess the proposed change, can you provide what the error was before the change?

codypenta:
```rust
            if tmp.message.is_none() {
                tmp.message = _error_message;
            }
```

In the rust code this would throw `is_none()` does not exist on type `String`. And I think there is was some try accessing it as a String but it was an `Option<String>`

jdisanti:
I see. I think the proposed change will fix it in some cases, but potentially not all. To make it work in all cases, the `NullableIndex` must be used to determine if it's wrapped in an Option or not, or better yet, check if the member's symbol is marked as optional (which is determined by the `NullableIndex` earlier in the codegen). Something like:

```kotlin
if (errorMessageMember != null) {
    val symbol = symbolProvider.toSymbol(errorMessageMember)
    if (symbol.is_optional()) {
        ...
```

Feel free to open a PR.

PR #3190, opened 2023-11-13
Fix error struct with default impl

codypenta:
## Motivation and Context

When an error member has a default value, it will generate errors (specifically is_none not found on type String) with smithy-rs client unless you manually specify @required

More Here: #3182

## Description

Adds recomendation change to address error structures with a default implementation like so:

```kotlin
if (errorMessageMember != null) {
    val symbol = symbolProvider.toSymbol(errorMessageMember)
    if (symbol.isOptional()) {
        rust(
            """
            if tmp.message.is_none() {
                tmp.message = _error_message;
            }
            """,
        )
    }
}
..... 
```

```smithy
@error("client")
structure Error {
    @required
    requestId: String

    @required
    message: String

    code: String = "400"

    context: String
}
```

## Testing

Added the above and ran `./gradlew codegen-client-test:build` with a build successful

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -11,6 +11,14 @@
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
 
+[[smithy-rs]]
+message = """
+Fix rust rendering of @error structs when fields have default implementation
+"""
+references = ["smithy-rs#3182"]
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "all"}
+author = "codypenta"
+
 [[aws-sdk-rust]]
 message = "Change `ByteStream::into_async_read` to return `AsyncBufRead`"
 references = ["smithy-rs#3164"]
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -13,10 +13,10 @@
 
 [[smithy-rs]]
 message = """
-Fix rust rendering of @error structs when fields have default implementation
+Fix rendering of @error structs when fields have default values
 """
 references = ["smithy-rs#3182"]
-meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "all"}
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client"}
 author = "codypenta"
 
 [[aws-sdk-rust]]
--- a/codegen-client-test/model/error-correction-nullability-test.smithy
+++ b/codegen-client-test/model/error-correction-nullability-test.smithy
@@ -27,6 +27,10 @@ structure Error {
 
     @required
     message: String
+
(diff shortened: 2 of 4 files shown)


codypenta:
@jdisanti let me know if I need to make any changes, pretty new to the code base 😄

codegen-client-test/model/error-correction-nullability-test.smithy
```diff
@@ -27,6 +27,10 @@ structure Error {
 
     @required
     message: String
+
+    code: String = "400"
```
jdisanti, line 31: Thanks for thinking to add a test!

It would be nice for this to get its own isolated test case though so that it doesn't get lost in the future.

Would you mind creating a `ProtocolParserGeneratorTest.kt` file in `src/test/kotlin/... same path as ProtocolParserGenerator .../` that tests this in isolation? Here's an example of how to do a client integration test that runs the full client code generator against a test model and verifies it compiles:
https://github.com/smithy-lang/smithy-rs/blob/c4a14d7c28fa26fb7f89f9634933965bab561cac/codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/protocols/AwsQueryTest.kt#L12

CHANGELOG.next.toml
```diff
@@ -11,6 +11,14 @@
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
 
+[[smithy-rs]]
+message = """
+Fix rust rendering of @error structs when fields have default implementation
```
jdisanti, line 16:
```suggestion
Fix rendering of @error structs when fields have default values
```
Rust is implied

CHANGELOG.next.toml
```diff
@@ -11,6 +11,14 @@
...
+"""
+references = ["smithy-rs#3182"]
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "all"}
```
jdisanti, line 19:
```suggestion
meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client"}
```
This change only affects clients.

jdisanti (commented):
Thanks for taking the time to do this! ❤️ 

Just a couple minor things.

codegen-client-test/model/error-correction-nullability-test.smithy
codypenta: Sure can, would you like me to also revert the changes made in the current file or should I leave that in?

jdisanti: Yeah, please revert.

codypenta: Done and done 😄

jdisanti (approved):

rcoh (approved):

PR #3191, opened 2023-11-14
Revise unhandled error variant according to RFC-39

jdisanti:
This PR implements [RFC-39](https://github.com/smithy-lang/smithy-rs/blob/main/design/src/rfcs/rfc0039_forward_compatible_errors.md) with a couple slight deviations:
- No `introspect` method is added since `Error` already implements `ProvideErrorMetadata`.
- The same opaqueness and deprecation pointer is applied to the enum unknown variant for consistency.

----


aws/sdk/integration-tests/s3/tests/request_id.rs
```diff
@@ -91,7 +91,7 @@ async fn get_request_id_from_unmodeled_error() {
         .expect_err("status 500")
         .into_service_error();
     request.expect_request();
-    assert!(matches!(err, GetObjectError::Unhandled(_)));
+    assert!(err.is_unhandled());
```
rcoh, line 94: `is_unhandled` should _maybe_ be deprecated for the same reason

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/ClientEnumGenerator.kt
```diff
@@ -81,6 +82,10 @@ data class InfallibleEnumType(
 
     override fun additionalEnumMembers(context: EnumGeneratorContext): Writable = writable {
         docs("`$UnknownVariant` contains new variants that have been added since this code was generated.")
+        rust(
+            """##[deprecated(note = "Don't directly match on `$UnknownVariant`. Instead, match on `_` and use \
```
rcoh, line 86: I think this is easier if you match on `other if other.as_str() == "Other" { ... }` personally

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/OperationErrorGenerator.kt
```diff
@@ -103,12 +85,25 @@ class OperationErrorGenerator(
             rust(
                 """
                 /// An unexpected error occurred (e.g., invalid JSON returned by the service or an unknown error code).
+                ##[deprecated(note = "Don't directly match on `Unhandled`. Instead, match on `_` and use \
```
rcoh, line 88: should probably suggest `err` again—if you match on `_` and reach back up to the error, sometimes you'll hit lifetime issues

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/OperationErrorGenerator.kt
```diff
@@ -120,39 +115,46 @@ class OperationErrorGenerator(
...
+    }
+
+    private fun RustWriter.renderImplDisplay(errorSymbol: Symbol, errors: List<StructureShape>) {
+        rustBlock("impl #T for ${errorSymbol.name}", RuntimeType.Display) {
             rustBlock("fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::fmt::Result") {
-                delegateToVariants(errors) {
-                    writable { rust("_inner.fmt(f)") }
+                delegateToVariants(errors) { variantMatch ->
+                    when (variantMatch) {
+                        is VariantMatch.Unhandled -> writable { rust("f.write_str(\"unhandled error\")") }
```
rcoh, line 131: do we lose the code in this case? The unhandled error should still include the error code if we have it

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/OperationErrorGenerator.kt
```diff
@@ -182,60 +184,79 @@ class OperationErrorGenerator(
...
                     rust("matches!(self, Self::${errorVariantSymbol.name}(_))")
                 }
             }
+            rust(
+                """
+                /// True if this error is the `Unhandled` variant.
+                pub fn is_unhandled(&self) -> bool {
+                    matches!(self, Self::Unhandled(_))
+                }
+                """,
```
rcoh, line 240: this method should be deprecated or removed—it's an anti pattern

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/ServiceErrorGenerator.kt
```diff
@@ -191,7 +223,53 @@ class ServiceErrorGenerator(
                 rust("${sym.name}(#T),", sym)
             }
             docs("An unexpected error occurred (e.g., invalid JSON returned by the service or an unknown error code).")
+            rust(
+                "##[deprecated(note = \"Don't directly match on `Unhandled`. Instead, match on `_` and use " +
```
rcoh, line 227: same

rcoh: these deprecation messages should probably be centralized in codegen & include an explanation about why

aws/sdk/integration-tests/s3/tests/request_id.rs
jdisanti: I only added it for a unit test. I'll just remove it.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/OperationErrorGenerator.kt
jdisanti: The previous behavior only outputted "unhandled error", but it's trivial to add the code to it, so I just went ahead and did it.

ysaito1001 (approved):

examples/pokemon-service-client-usage/examples/handling-errors.rs
```diff
@@ -77,10 +77,6 @@ async fn main() {
                 GetStorageError::ValidationError(ve) => {
                     tracing::error!(error = %ve, "A required field has not been set.");
                 }
-                // An unexpected error occurred (e.g., invalid JSON returned by the service or an unknown error code).
-                GetStorageError::Unhandled(uh) => {
-                    tracing::error!(error = %uh, "An unhandled error has occurred.")
-                }
```
rcoh, line 83: if this is an example it should probably use `DisplayErrorContext`

rust-runtime/aws-smithy-types/src/error/unhandled.rs
```diff
@@ -58,6 +61,7 @@ impl Builder {
 /// [`DisplayErrorContext`](crate::error::display::DisplayErrorContext), use another
 /// error reporter library that visits the error's cause/source chain, or call
 /// [`Error::source`](std::error::Error::source) for more details about the underlying cause.
+#[deprecated(note = "This type is no longer used by errors.")]
```
rcoh, line 64: so is this type still used at all?

CHANGELOG.next.toml
```diff
@@ -90,3 +90,61 @@ message = "Remove deprecated error kind type aliases."
...
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = """
+Unhandled errors have been made opaque to ensure code is written in a future-proof manner. Where previously, you
+might have:
+```rust
+match service_error.err() {
+    GetStorageError::StorageAccessNotAuthorized(_) => { /* ... */ }
+    GetStorageError::Unhandled(unhandled) if unhandled.code() == "SomeUnmodeledErrorCode" {
```
rcoh, line 101: code is optional; `== Some(...)`, same for below

rcoh (approved):
approved with a few comments—I think we should also workshop that deprecation message but we can do that later

jdisanti:
While fixing examples for this, I realized we need a way to parse an enum while disallowing unknown variants. For example, this is what examples are currently doing:
```rust
    // parse resource type from user input
    let parsed = ResourceType::from(resource_type.as_str());
    if matches!(parsed, ResourceType::Unknown(_)) {
        panic!(
            "unknown resource type: `{}`. Valid resource types: {:#?}",
            &resource_type,
            ResourceType::values()
        )
    }
```
I'm going to add a `try_parse` method to the enums for this use-case that will error if the variant doesn't exist.

examples/pokemon-service-client-usage/examples/handling-errors.rs
jdisanti: Good call. Fixed.

rust-runtime/aws-smithy-types/src/error/unhandled.rs
jdisanti: It's no longer used. I left it in to be removed in the next release in case anyone was referencing it.

CHANGELOG.next.toml
jdisanti: Good catch!

jdisanti:
Also found that `StatusCode` was inaccessible while fixing examples. Added a fix here in 34b3dea55a461cafbe24fb36884cd4c029d35b7c.

rcoh:
LGTM. Suggested deprecation message:

```
#[deprecated(
    note = "Matching `Unhandled` directly is not forwards compatible. Instead, match using a 
    variable wildcard pattern and check `.code()`: 
    &nbsp;&nbsp;&nbsp;`err if err.code() == Some(\"SpecificExceptionCode\") => { /* handle the error */ }`.
    See [`ProvideErrorMetadata`](#impl-Unhandled-for-<NameOfStruct>) for more information."
)]
```
Screenshot 2023-11-15 at 10 46 29 AM

It renders like this (and the link works!)

rcoh (approved):

jdisanti:
Made a slight tweak to your suggested message:
```
See [`ProvideErrorMetadata`]($link) for what information is available for the error.
```
Instead of "for more information" since the trait doesn't give more information about this deprecation, so that could have been confusing.

PR #1422, opened 2022-05-30
Avoid unnecessary Cargo crate rebuilds

david-perez:
This commit modifies the Gradle buildscripts to avoid unnecessary Cargo
rebuilds of the generated crates, decreasing development iteration
cycles.

Prior to this commit, if you ran a crate-generating command _twice_ on
any of the `codegen-test`, `codegen-server-test`, `sdk-codegen-test`,
and `codegen-server-test:python` subprojects (without making any changes
to the codegen code), both invocations would take the same time: Cargo
would recompile the crate and its dependencies, even when the generated
crate is identical. This is because Gradle deletes everything under
`buildDir` when running the `generateSmithyBuild` task:

```
> Task :codegen-server-test:smithyBuildJar
Deleting stale output file: /local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/codegen-server-test/build/smithyprojections/codegen-server-test
```

So the files get recreated each time.

While developing, it is likely that only a small number of the generated
crate files are modified across rebuilds. [Cargo uses
`mtime`](https://github.com/rust-lang/cargo/issues/6529) (among other
factors) to determine whether it needs to recompile a unit. Indeed,
running with `CARGO_LOG=cargo::core::compiler::fingerprint=trace` yields
`err: current filesystem status shows we're outdated`. This commit adds
a Gradle task that compares the hashes of the newly generated files with
the (previously cached) old ones, and restores their `mtime`s if the
hashes coincide.

Another issue was causing unnecessary crate rebuilds. Prior to this
commit, we were sending `RUSTFLAGS=-D warnings` when invoking Cargo.
However, a common thing to do after generating a crate is to open its
contents in an editor. The editor's `rust-analyzer` would compile the
crate and its dependencies without the `RUSTFLAGS` we had used earlier.
The next time you rebuilt the crate, Cargo would claim `err: RUSTFLAGS
has changed: previously [], now ["-D", "warnings"]` and recompile
everything again. This commit refactors the Gradle tasks so as to not
send these flags when invoking Cargo, instead generating a
`.cargo/config.toml` containing these flags. This way, `rust-analyzer`
also picks them up and does not need to recompile the crates.

With both patches, Cargo avoids unnecessary crate rebuilds. All in all,
the second invocation of a `./gradlew --info -P modules='simple' -P
cargoCommands='test' codegen-server-test:build` command now takes 15
seconds less than the first invocation on my `c5.4xlarge` machine; Cargo
does not need to do _any_ work on the second invocation.

[...]

PR #3192, opened 2023-11-14
Rely on `.cargo/config.toml` instead of `RUSTFLAGS` in tests' generated crates

david-perez:
This is PR is similar to #1422. When generating a crate (either via
`Project.compileAndTest` or `{client,server}IntegrationTest`), we're
setting `RUSTFLAGS`. When developing, a common thing to do after
generating a crate is to open its contents in an editor, make
modifications, and run `cargo` again.

- the editor will kick off `rust-analyzer`, which will build the code
  again without the `RUSTFLAGS`; and
- the developer is unlikely to use the same `RUSTFLAGS` when running
  manual `cargo` commands; in fact, they'll most likely use none.

Indeed, using `CARGO_LOG=cargo::core::compiler::fingerprint=trace` on
the generated crate reveals that `cargo` is claiming it has to rebuild
the entire dependency closure because a different set of flags was used.
This causes unnecessary crate rebuilds during which the developer has to
wait.

Instead of relying on `RUSTFLAGS` in the command line invocation, we can
leverage `.cargo/config.toml`'s `[build.rustflags]` to persist the
settings in a file, like we did in #1422. That way, any plain `cargo`
command that runs on the same project will reuse the previous
compilation artifacts.

----


tools/ci-scripts/check-client-codegen-unit-tests
```diff
@@ -6,4 +6,5 @@
 
 set -eux
 cd smithy-rs
+printenv
```
david-perez, line 9: Weird, something in CI is setting `RUSTFLAGS=-D warnings`: https://github.com/smithy-lang/smithy-rs/actions/runs/6864720326/job/18667193959?pr=3192#step:3:356

david-perez: Oh, we bake these into the Docker image: https://github.com/smithy-lang/smithy-rs/blob/31d9821e0465218049c7345d14d23b97dd7bc011/tools/ci-build/Dockerfile#L194-L206

david-perez: I opened https://github.com/smithy-lang/smithy-rs/issues/3194.

jdisanti (approved):

PR #3204, opened 2023-11-15
Rename behavior versions and associated functions

jdisanti:
The previous names are just too verbose. This PR shortens them and tries to make things clearer.

----

### DIFF
--- a/design/src/SUMMARY.md
+++ b/design/src/SUMMARY.md
@@ -66,7 +66,7 @@
   - [RFC-0037: The HTTP Wrapper](./rfcs/rfc0037_http_wrapper.md)
   - [RFC-0038: User-configurable retry classification](./rfcs/rfc0038_retry_classifier_customization.md)
   - [RFC-0039: Forward Compatible Errors](./rfcs/rfc0039_forward_compatible_errors.md)
-  - [RFC-0040: Behavior Major Versions](./rfcs/rfc0040_behavior_major_versions.md)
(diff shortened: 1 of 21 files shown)


rcoh (approved):
looks good. please verify docs render as expected

design/src/SUMMARY.md
```diff
@@ -66,7 +66,7 @@
   - [RFC-0037: The HTTP Wrapper](./rfcs/rfc0037_http_wrapper.md)
   - [RFC-0038: User-configurable retry classification](./rfcs/rfc0038_retry_classifier_customization.md)
   - [RFC-0039: Forward Compatible Errors](./rfcs/rfc0039_forward_compatible_errors.md)
-  - [RFC-0040: Behavior Major Versions](./rfcs/rfc0040_behavior_major_versions.md)
+  - [RFC-0040: Behavior Major Versions](./rfcs/rfc0040_behavior_versions.md)
```
ysaito1001, line 69: I assume you deliberately left `RFC-0040: Behavior Major Versions` here as opposed to `RFC-0040: Behavior Versions` for a reason.

ysaito1001 (approved):
I personally like the new names `load_defaults` and `defaults`.

design/src/SUMMARY.md
jdisanti: Oops, I need to fix this one.

jdisanti:
This is a breaking change as far as cargo-semver-checks is concerned, but it's not actually breaking since we haven't released yet.

PR #3210, opened 2023-11-16
Fix AwsQuery target serialization

rcoh:
## Motivation and Context
- aws-sdk-rust#957

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -10,3 +10,9 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
+
+[[aws-sdk-rust]]
+message = "Fix broken serialization for SQS"
+references = ["smithy-rs#3210", "aws-sdk-rust#957"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
+author = "rcoh"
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/CodegenTestCommon.kt
```diff
@@ -137,6 +141,7 @@ fun Project.registerGenerateSmithyBuildTask(
     this.tasks.register("generateSmithyBuild") {
         description = "generate smithy-build.json"
         outputs.file(project.projectDir.resolve("smithy-build.json"))
+        // NOTE: This is not working.
```
Velfi, line 144: Should we create a TODO + issue for this?

Velfi (approved):

buildSrc/src/main/kotlin/CodegenTestCommon.kt
rcoh: filed

CHANGELOG.next.toml
```diff
@@ -9,4 +9,10 @@
...
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
-# author = "rcoh"
\ No newline at end of file
+# author = "rcoh"
+
+[[aws-sdk-rust]]
+message = "Fix broken serialization for SQS"
+references = ["smithy-rs#3210", "aws-sdk-rust#957"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
```
ysaito1001, line 17: Ah, this was actually `"bug" = true`

PR #3217, opened 2023-11-16
Remove invalid defaults for some services

milesziemer:
## Motivation and Context

Some services have generated types with properties that have a default value of zero. This can cause invalid requests if the service expects a value > 0.

## Description

Adds a customization that removes the default value for services with this issue.

## Testing

- Added a test for the customization to make sure defaults were removed.
- Generated clients for the impacted services and verified the expected types had the default value removed.

## Checklist

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/RemoveDefaultsDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/RemoveDefaultsDecorator.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.rustsdk.customize
+
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.rust.codegen.client.smithy.ClientRustSettings
+import software.amazon.smithy.rust.codegen.client.smithy.customize.ClientCodegenDecorator
+import software.amazon.smithy.rust.codegen.core.util.shapeId
+import java.util.logging.Logger
+
+/**
+ * Removes default values from certain shapes, and any member that targets those shapes,
+ * for some services where the default value causes serialization issues, validation
+ * issues, or other unexpected behavior.
+ */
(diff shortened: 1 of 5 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/RemoveDefaultsDecorator.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+class RemoveDefaultsDecorator : ClientCodegenDecorator {
+    override val name: String = "RemoveDefaults"
+    override val order: Byte = 0
+    private val logger: Logger = Logger.getLogger(javaClass.name)
+    // Service shape id -> Shape id of each root shape to remove the default from.
+    private val removeDefaults = mapOf(
+        "com.amazonaws.emrserverless#AwsToledoWebService".shapeId() to setOf(
+            // Service expects this to have a min value > 0
+            "com.amazonaws.emrserverless#WorkerCounts".shapeId()
```
rcoh, line 28: do we want to do it based on the range check? Or is hard coding it the right thing to do here?

rcoh (commented):
this _seems_ more service specific? (in which case, we'd put in it in the service-specific decorator) but doesn't super matter. We should make sure we get examples updated in case this causes examples compilation issues.

Code is fine, just want to make sure we audit the generated code before landing.

milesziemer:
> this _seems_ more service specific? (in which case, we'd put in it in the service-specific decorator) but doesn't super matter. We should make sure we get examples updated in case this causes examples compilation issues.
> 
> Code is fine, just want to make sure we audit the generated code before landing.

@rcoh This started out with quite a few more services, but by the time the PR was ready only one was left. I can update if need be though.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/RemoveDefaultsDecorator.kt
milesziemer: I thought about doing that. But my reasoning for not doing so was that since the range is wrong now, this will generate an `Option<T>`. Down the road, if the range check became not wrong, it would be a breaking change. So I think its best to just remove the default and always do it to be consistent.

rcoh (approved):
LGTM—please just add `TODO(<link to ticket in this repo): Remove this customizations after model updates`

PR #3228, opened 2023-11-17
Remove defaults for more services

milesziemer:
## Motivation and Context

Removing defaults from these shapes may not have a codegen impact since these are in top level input, but doing this just to be safe from future changes.

## Description

Adds more services/shapes to RemoveDefaultsDecorator, which will have their models changed in the future.  Also updates RemoveDefaults to allow for members to be present in the list.

## Testing

- Updated unit test for RemoveDefaults
- Generated clients and looked at the output code

## Checklist

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/RemoveDefaultsDecorator.kt
```diff
@@ -25,6 +25,32 @@ class RemoveDefaultsDecorator : ClientCodegenDecorator {
     // Service shape id -> Shape id of each root shape to remove the default from.
     // TODO(https://github.com/smithy-lang/smithy-rs/issues/3220): Remove this customization after model updates.
     private val removeDefaults = mapOf(
+        "com.amazonaws.amplifyuibuilder#AmplifyUIBuilder".shapeId() to setOf(
```
rcoh, line 28: for a little less visual noise I'd probably delete `.shapeId()` in all of these and do `.map { k, v -> k.shapeId() to v.shapeId() }` at the end.

rcoh (approved):

PR #3229, opened 2023-11-17
Avoid overflow in exponential backoff computation

rcoh:
## Motivation and Context
aws-sdk-rust#960

## Description
Use checked_pow to avoid overflow in backoff computation

## Testing
- unit test

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
+++ b/rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
@@ -286,7 +286,10 @@ fn check_rate_limiter_for_delay(
 }
 
 fn calculate_exponential_backoff(base: f64, initial_backoff: f64, retry_attempts: u32) -> f64 {
-    base * initial_backoff * 2_u32.pow(retry_attempts) as f64
+    2_u32
+        .checked_pow(retry_attempts)
+        .map(|backoff| (backoff as f64) * base * initial_backoff)
+        .unwrap_or(f64::MAX)
 }
 
 fn get_seconds_since_unix_epoch(runtime_components: &RuntimeComponents) -> f64 {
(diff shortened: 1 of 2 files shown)


rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
```diff
@@ -286,7 +286,10 @@ fn check_rate_limiter_for_delay(
 }
 
 fn calculate_exponential_backoff(base: f64, initial_backoff: f64, retry_attempts: u32) -> f64 {
-    base * initial_backoff * 2_u32.pow(retry_attempts) as f64
+    2_u32
+        .checked_pow(retry_attempts)
+        .map(|backoff| (backoff as f64) * base * initial_backoff)
+        .unwrap_or(f64::MAX)
```
jdisanti, line 292: f64::MAX is very different from u32::MAX. This doesn't seem like a sane fallback.

ysaito1001 (approved):

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
jdisanti: I wonder if we should compute exponential backoff at validation time and warn if the backoff is greater than some reasonable number... like, is a backoff of 1 week really useful?

rcoh: the next line after this is called mins it with a delay. We aren't falling back from u32 max to f64 max—we're falling back the entire retry delay–basically "if it's huge, just delay forever" but that will get capped

jdisanti: Ah, OK.

jdisanti (approved):

Issue #3232, opened 2023-11-17
Reexport http::Uri in aws_config


pacak:
For `endpoint` method on client buiilder (https://docs.rs/aws-config/0.100.0/aws_config/imds/client/struct.Builder.html#method.endpoint) to work `Uri` must match version used by aws-config - version must be semver compatible to 0.2.

Trying to add `http` dependency with `cargo add http` adds a different version (1.0) and code fails to compile. Adding required version manually is fragile.

A typical solution for such problem is for `aws_config` to reexport `Uri`:

```diff
-use http::Uri;
+pub use http::Uri;
```

jdisanti:
Closing this. See https://github.com/smithy-lang/smithy-rs/pull/3233#issuecomment-1816839602 for explanation.

PR #3233, opened 2023-11-17
reexport http::Uri

pacak:
## Motivation and Context
Fixes https://github.com/smithy-lang/smithy-rs/issues/3232

## Description
Reexport http::Uri

## Testing
No tests

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -11,6 +11,13 @@
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
 
+[[smithy-rs]]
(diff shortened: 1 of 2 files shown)


jdisanti:
Hey, thanks for contributing to the project!

We actually just changed the endpoint method on IMDS client to take a string instead of a URI, so this re-export should be unnecessary when that goes out in the next release. We don't want to expose the http crate at all since the SDK currently uses 0.2 and 1.0 just released. We need to be able to upgrade to 1.0 without it being a breaking change.

pacak:
> We actually just changed the endpoint method on IMDS client to take a string instead of a URI

That's even better :)

PR #3238, opened 2023-11-19
Fix stalled stream detection

rcoh:
A number of small issues in StalledStreamDetection came together to make it ineffective in practice:
1. We didn't push a throughput `0` on `Poll::Pending`. This means that if the stream goes to poll pending, the calculate throughput can never get to 0.
2. Because we always considered the _entire_ throughput log, an actually stalled stream creates a pathological case where most of the log is full of fast moving data, but we are very slow to evict it and detect slowness. With a 426 length log, it would effectively take 7 minutes for it to actually get to 0.

For these two issues, I introduced a check-window (default 1 second) in the throughput log. We only consider this much data when making the calculation. To avoid undeseriable interactions with the check interval, I reduced the check interval to 500ms. I think this is likely to be better in practice at detecting stalled streams.

As a coincidence, this makes all the math exact so the tests are now precise without floating epsilons.

Finally, in an unrelated issue, the `check_interval` wasn't actually being used in the call to sleep, I fixed that as well and added a test.

In doing this, I tried to also simplify sine wave test into a test that could be logically reasoned about. The previous test was producing nonsense data (integer input to `sine` which is in radians as one example). I replaced it with several tests that send data at one rate, then suddenly stop sending data, simulating a stalled stream. These have the nice property that it's easy to determine mathematically when the stream should stall.

Side note: It's possible these bugs are what was making it not work for request bodies? Needs more investigation. I didn't rerun that test.

After those issues were resolved, I was a little concerned about performance when maxing out the network connection. A quick benchmark showed there was a small regression. When data is flowing normally, the 426 item buffer is usually shorter than our default check window of 1 second. This allows for an optimization where we track the current total amount of data in the buffer, allowing us to return the calculated throughput in constant time.

During writing of the test, I discovered that we didn't actually expose `Throughput` so I took that as an opportunity to change it to use `u64` instead of `f64` for it's byte count representation.

## Testing
- [x] Manual test of downloading a file then turning off the wifi. Verify that the connection is aborted within the grace period.

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
+++ b/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
@@ -237,7 +237,7 @@ mod test {
             eight_byte_per_second_stream_with_minimum_throughput_timeout(minimum_throughput);
         let expected_err = Error::ThroughputBelowMinimum {
             expected: minimum_throughput,
-            actual: Throughput::new(8_f64, Duration::from_secs(1)),
+            actual: Throughput::new(8, Duration::from_secs(1)),
         };
         match res.await {
             Ok(_) => {
@@ -257,7 +257,7 @@ mod test {
 
     #[tokio::test]
     async fn test_throughput_timeout_less_than() {
-        let minimum_throughput = Throughput::new_bytes_per_second(9.0);
+        let minimum_throughput = Throughput::new_bytes_per_second(9);
         expect_error(minimum_throughput).await;
     }
 
@@ -278,18 +278,18 @@ mod test {
     async fn test_throughput_timeout_equal_to() {
         let (_guard, _) = capture_test_logs();
         // a tiny bit less. To capture 0-throughput properly, we need to allow 0 to be 0
-        let minimum_throughput = Throughput::new(31.0, Duration::from_secs(4));
+        let minimum_throughput = Throughput::new(31, Duration::from_secs(4));
         expect_success(minimum_throughput).await;
     }
 
     #[tokio::test]
     async fn test_throughput_timeout_greater_than() {
-        let minimum_throughput = Throughput::new(20.0, Duration::from_secs(3));
+        let minimum_throughput = Throughput::new(20, Duration::from_secs(3));
         expect_success(minimum_throughput).await;
     }
 
     // A multiplier for the sine wave amplitude; Chosen arbitrarily.
-    const BYTE_COUNT_UPPER_LIMIT: f64 = 1000.0;
+    const BYTE_COUNT_UPPER_LIMIT: u64 = 1000;
 
     /// emits 1000B/S for 5 seconds then suddenly stops
     fn sudden_stop(
@@ -307,7 +307,7 @@ mod test {
                 } else {
                     let mut bytes = BytesMut::new();
                     let bytes_per_segment =
-                        BYTE_COUNT_UPPER_LIMIT / (1f64 / sleep_dur.as_secs_f64());
+                        (BYTE_COUNT_UPPER_LIMIT as f64) * sleep_dur.as_secs_f64();
                     for _ in 0..bytes_per_segment as usize {
                         bytes.put_u8(0)
                     }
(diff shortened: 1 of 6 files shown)


rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
```diff
@@ -21,49 +21,69 @@ where
...
+            .sleep_fut
+            .take()
+            .unwrap_or_else(|| this.async_sleep.sleep(this.options.check_interval()));
         if let Poll::Ready(()) = pin!(&mut sleep_fut).poll(cx) {
+            tracing::trace!("sleep future triggered—triggering a wakeup");
             // Whenever the sleep future expires, we replace it.
-            sleep_fut = this
-                .async_sleep
-                .sleep(this.options.minimum_throughput().per_time_elapsed());
+            sleep_fut = this.async_sleep.sleep(this.options.check_interval());
```
Velfi, line 53: ty for catching this bug!

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
```diff
@@ -255,57 +276,95 @@ mod test {
...
+    }
+
+    #[tokio::test]
+    async fn test_slow_stream_detection() {
+        test_suddenly_stopping_stream(BYTE_COUNT_UPPER_LIMIT / 2, Duration::from_secs_f64(5.50))
+            .await
+    }
+
+    #[tokio::test]
+    async fn test_check_interval() {
```
Velfi, line 333: This is a cool test

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/options.rs
```diff
@@ -19,9 +19,17 @@ pub struct MinimumThroughputBodyOptions {
     /// If this is set to a positive value, whenever throughput is below the minimum throughput,
     /// a timer is started. If the timer expires before throughput rises above the minimum,
     /// an error is emitted.
+    ///
+    /// It is recommended to set this to a small value to avoid issues during stream-startup.
```
Velfi, line 23: We should quantify what a 'small' value means in this case.

rust-runtime/aws-smithy-runtime/tests/stalled_stream_performance.rs
```diff
@@ -0,0 +1,114 @@
...
+// TODO(postGA): convert this to an actual benchmark
+// This test evaluates streaming 1GB of data over the loopback with and without the body wrapper
+// enabled. After optimizations, the body wrapper seems to make minimal differences
+// NOTE: make sure you run this test in release mode to get a sensible result
+#[tokio::test]
+#[ignore]
+async fn stalled_stream_performance() {
+    // 1GB
+    let data_size = 1_000_000_000;
+    let block_size = 16384;
```
Velfi, line 37: We should note that this is derived from S3

rust-runtime/aws-smithy-runtime/tests/stalled_stream_performance.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+    println!(
+        "Average w/ wrapping: {}",
+        wrapping.iter().map(|it| it.as_millis() as f64).sum::<f64>() / runs as f64
+    );
+    println!(
+        "Average w/o wrapping: {}",
+        no_wrapping
+            .iter()
+            .map(|it: &Duration| it.as_millis() as f64)
```
Velfi, line 74: How come the one on line 68 needs no type hint?

rcoh: neither needs one, but I commented out the no-wrap test at one point (when you don't write to it, then it needs a hint)

ysaito1001:
Will merge once `CHANGELOG.next.toml` has been updated for the fix.

rcoh:
Ah, wasn't going to do a changelog since this brings the behavior in line with the change log we already have

Issue #102, opened 2021-01-05
Support optional URI elements


rhernandez35:
Currently, the Rust code generator assumes that [all members with the `httpLabel` trait are required](https://github.com/awslabs/smithy-rs/blob/main/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/SymbolVisitor.kt#L135-L136). This isn't always true. For example, S3's GetObjectRequest structure binds `partNumber` to a URI element, but it's necessarily optional since it's only accepted when retrieving a single part of a multi-part object. As is, the generated code forces this field to be specified in all requests.

I'd say "just examine the member for the `required` trait," but I sadly don't see many models using it. FWIW, [The AWS service models bundled with botocore](https://github.com/boto/botocore/blob/develop/botocore/data/) appear to correctly model all required members regardless of where they appear on the wire. Are these service definitions hand-maintained?

PR #3243, opened 2023-11-21
Serialize zero values

milesziemer:
## Motivation and Context

Generated clients currently don't serialize the value `0` and `false` for members that have a default value. It _seems_ like this logic was originally added to fix issues with [S3](https://github.com/smithy-lang/smithy-rs/issues/102) which are technically modeling issues and have since been fixed. Not serializing zero values has caused issues in the past, for example https://github.com/awslabs/aws-sdk-rust/issues/630, and in other SDKs that do it like Go v2: https://github.com/aws/aws-sdk-go-v2/issues/2162. Also the Smithy spec says clients should serialize default values, which generated clients currently do not do if that default is `0`/`false`.

## Description

Removes logic from RustWriter that was used to not serialize zero values, ie. 0 for numbers and false for booleans, if that was also the default value for the member. With this change, the only time the default value for a member is not serialized is if the member is also `@clientOptional` per [Smithy spec](https://smithy.io/2.0/spec/type-refinement-traits.html#default-value-serialization).

The Blob type was also updated to implement std::Default, since SymbolVisitor treats `""` as a RustDefault.

**Important**: This change impacts all the protocols and will cause many members to now be sent over the wire that previously weren't.

## Testing

- [x] Generated the AWS SDK and inspected the diff, looks like the expected changes, removing checks for default values in serializers.
- [ ] TODO: Testing with individual clients: This is a much larger task since there are so many clients that have been changed, will likely use a subset.

## Checklist

----


milesziemer:
Closing in favor of https://github.com/smithy-lang/smithy-rs/pull/3252, until we better test the impact of default value serialization.

PR #3246, opened 2023-11-21
Remove some defaults for s3 control

milesziemer:
## Motivation and Context

S3 Control's PublicAccessBlockConfiguration is incorrectly modeled, and all the members should be nullable. These members are not meant to have a default, as you are meant to be able to configure one member at a time - if you're just trying to turn BlockPublicAcls on, you don't want to accidentally turn off IgnorePublicAcls. This change makes these members nullable and is a break-fix.

## Description

Adds the members of PublicAccessBlockConfiguration to the RemoveDefaults customization. 
Also refactors the map of shapes to remove defaults from to avoid having to call .shapeId everywhere.

## Testing

- [x] Generated the client and looked at it

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -10,3 +10,9 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
+
+[[aws-sdk-rust]]
+message = "Make properties of S3Control PublicAccessBlockConfiguration optional. Previously, they defaulted to false, but this created invalid requests."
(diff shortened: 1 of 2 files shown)


rcoh:
We'll need to merge this directly to the release branch

rcoh (approved):
lgtm—does s3control need to be fully undefaulted or is this (hopefully), the only issue?

milesziemer:
> lgtm—does s3control need to be fully undefaulted or is this (hopefully), the only issue?

Their model doesn't have everything defaulted - ie. they do box things, so I don't think the entire model is just incorrect. But it is tough to say whether this is the only issue.

PR #3249, opened 2023-11-23
Lifetimes in builders

82marbag:
## Motivation and Context
We're not handling lifetimes in builders of structures.

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt
@@ -93,12 +93,9 @@ open class StructureGenerator(
      * When [SensitiveTrait] support is required, render a custom debug implementation to redact sensitive data
      */
     private fun renderDebugImpl() {
+        val lifetime = shape.lifetimeDeclaration(symbolProvider)
         writer.rustBlock(
-            "impl ${shape.lifetimeDeclaration(symbolProvider)} #T for $name ${
-                shape.lifetimeDeclaration(
-                    symbolProvider,
-                )
-            }",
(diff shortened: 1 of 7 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt
```diff
@@ -88,32 +88,19 @@ open class StructureGenerator(
...
-    }
-
     /**
      * Render a custom debug implementation
      * When [SensitiveTrait] support is required, render a custom debug implementation to redact sensitive data
      */
     private fun renderDebugImpl() {
-        writer.rustBlock("impl ${lifetimeDeclaration()} #T for $name ${lifetimeDeclaration()}", RuntimeType.Debug) {
+        writer.rustBlock(
+            "impl ${shape.lifetimeDeclaration(symbolProvider)} #T for $name ${
```
david-perez, line 97: Use multiline blocks `"""` if the string spans multiple lines.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -196,11 +198,11 @@ class ServerBuilderGenerator(
             it == RuntimeType.Debug || it == RuntimeType.Clone
         } + RuntimeType.Default
         Attribute(derive(builderDerives)).render(writer)
-        writer.rustBlock("${visibility.toRustQualifier()} struct Builder") {
+        writer.rustBlock("${visibility.toRustQualifier()} struct Builder $lifetime") {
```
david-perez, line 201: Codegen diff shows there's an extra space after `Builder` in these.

david-perez (approved):

PR #3252, opened 2023-11-24
Ignore actual default value instead of 0/false

milesziemer:
## Motivation and Context

Previously, there was logic used to ignore the default 0/false values for numbers/booleans when serializing members if they weren't boxed. This was to fix issues that occured when upstream models didn't properly box shapes that were meant to be optional, and for the most part worked because services would just fill in the default if it wasn't passed. However, with Smithy 2.0, models may have defaults != 0/false, but the codegenerator still ignores the zero value.

## Description

This commit makes it so that the actual default value for the member is ignored for booleans and numbers, instead of just 0/false. It also updates serialization for http bindings so that headers and query parameters with 0/false values aren't ignored if they are optional parameters.

## Testing

- [x] Generated AWS SDK and inspected diff. Only changes are not ignoring default value inside `if let Some() = ...` blocks, and ignoring default value instead of just 0 (only seems to effect nimble).
- [x] Added protocol tests for serializing 0/false in query params for restXml and restJson

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -111,13 +111,13 @@ meta = { "breaking" = false, "tada" = true, "bug" = false }
 author = "rcoh"
 
 [[smithy-rs]]
-message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false."
+message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false. See [changelog discussion](https://github.com/smithy-lang/smithy-rs/discussions/3312) for details."
 references = ["smithy-rs#3252", "smithy-rs#3312"]
-meta = { "breaking" = true, "tada" = false, "bug" = true, "target" = "client" }
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
 author = "milesziemer"
 
 [[aws-sdk-rust]]
-message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false."
+message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false. See [changelog discussion](https://github.com/smithy-lang/smithy-rs/discussions/3312) for details."
 references = ["smithy-rs#3252", "smithy-rs#3312"]
-meta = { "breaking" = true, "tada" = false, "bug" = true }
+meta = { "breaking" = false, "tada" = false, "bug" = true }
 author = "milesziemer"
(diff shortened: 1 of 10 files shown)


CHANGELOG.next.toml
```diff
@@ -109,3 +109,15 @@ message = "Add support for constructing [`SdkBody`] and [`ByteStream`] from `htt
 references = ["smithy-rs#3300", "aws-sdk-rust#977"]
 meta = { "breaking" = false, "tada" = true, "bug" = false }
 author = "rcoh"
+
+[[smithy-rs]]
+message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false."
```
jdisanti, line 114:
```suggestion
message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false. See [changelog discussion](https://github.com/smithy-lang/smithy-rs/discussions/3312) for details."
```

CHANGELOG.next.toml
```diff
@@ -109,3 +109,15 @@ message = "Add support for constructing [`SdkBody`] and [`ByteStream`] from `htt
...
+references = ["smithy-rs#3252", "smithy-rs#3312"]
+meta = { "breaking" = true, "tada" = false, "bug" = true, "target" = "client" }
+author = "milesziemer"
+
+[[aws-sdk-rust]]
+message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false."
```
jdisanti, line 120:
```suggestion
message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false. See [changelog discussion](https://github.com/smithy-lang/smithy-rs/discussions/3312) for details."
```

CHANGELOG.next.toml
```diff
@@ -109,3 +109,15 @@ message = "Add support for constructing [`SdkBody`] and [`ByteStream`] from `htt
...
+references = ["smithy-rs#3252", "smithy-rs#3312"]
+meta = { "breaking" = true, "tada" = false, "bug" = true }
```
jdisanti, line 122:
```suggestion
meta = { "breaking" = false, "tada" = false, "bug" = true }
```

CHANGELOG.next.toml
```diff
@@ -109,3 +109,15 @@ message = "Add support for constructing [`SdkBody`] and [`ByteStream`] from `htt
 references = ["smithy-rs#3300", "aws-sdk-rust#977"]
 meta = { "breaking" = false, "tada" = true, "bug" = false }
 author = "rcoh"
+
+[[smithy-rs]]
+message = "Serialize 0/false in query parameters, and ignore actual default value during serialization instead of just 0/false."
+references = ["smithy-rs#3252", "smithy-rs#3312"]
+meta = { "breaking" = true, "tada" = false, "bug" = true, "target" = "client" }
```
jdisanti, line 116:
```suggestion
meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
```

jdisanti (approved):
Looks great! Just some minor changelog changes.

PR #3257, opened 2023-11-26
Fix s3 tests

rcoh:
## Motivation and Context
Fix issue where S3 tests asserted on entire URI.

----


goabv (approved):
reviewed, necessary change to correct assert for url as url now can contain an optional additional parameter

PR #3262, opened 2023-11-28
Lazy initialize the default HTTP client

jdisanti:
This change initializes the TLS trust certs at base client validation time rather than upon creation of the default HTTP client runtime plugin so that if the default is overridden, that initialization doesn't need to take place. This is especially helpful on MacOS where that initialization takes approximately 100 milliseconds.

----


jdisanti:
We could potentially switch to the regex-lite crate to make the first initialization faster as well.

monken:
👀  You are working fast! Thank you!

jdisanti:
Split the regex changes out of this PR into https://github.com/smithy-lang/smithy-rs/pull/3269

jdisanti:
Since trusted cert initialization only occurs once per process and is only slow on MacOS, I'm currently of the opinion that we shouldn't merge this change. There may be some throw away work in the case where a client is overridden, but that amount is negligible. The primary benefit to optimizing it is for reducing AWS Lambda cold start time, but that is running Linux, and overriding the client there seems niche.

I could be persuaded otherwise though.

rcoh:
Even if this isn't required from a performance perspective, you still have the issue where loading native certs doesn't even work. For example, I suspect WASM may be broken by always initializing this connector?

I think lazy-loading may be not worth it but I think we do need a solution for not loading one if one is already set.

This is also going to make a big difference in running CI on macs as well.

jdisanti:
I believe WASM requires disabling the default hyper connector feature, so it actually still works in the present state.

I'll take a look at lazily loading the trusted certs without a behavior version change. It probably is worth it for the tests.

rust-runtime/aws-smithy-runtime/src/client/http/hyper_014.rs
```diff
@@ -535,10 +563,12 @@ impl HyperClientBuilder {
         self
     }
 
-    /// Create a [`HyperConnector`] with the default rustls HTTPS implementation.
+    /// Create a hyper client with the default rustls HTTPS implementation.
+    ///
+    /// This client will immediately load trusted certificates.
```
rcoh, line 568: doc comment is wrong

rcoh (approved):

ysaito1001 (approved):

PR #3269, opened 2023-11-28
Improve client init time by switching to regex-lite

jdisanti:
Each client initialization was taking between 1 and 2 milliseconds, regardless if the client had been constructed before or not. For example, if a customer wants five clients with different credentials providers, that could be 10 milliseconds of time spent in `Client::from_conf`. Approximately 98% of this time was spent compiling regular expressions for the endpoint partition resolver.

This change switches everything over to the regex-lite crate, which has faster regex compile times, and shouldn't have much of an impact on performance for our specific use-cases (small strings, only evaluated at client initialization).

The use of regex was entirely removed in aws-sigv4 since it was overkill for what it was being used for. 

----


jdisanti:
Whether or not the caching added in efd50328e10717878624fcc3277fac28e5bc1907 is justified is open for debate. The following are client creation times on MacOS with and without it.

With caching:
```
s3::Client::new - 215.5µs
s3::Client::new - 11.25µs
s3::Client::new - 8.791µs
s3::Client::new - 8.208µs
s3::Client::new - 8.708µs
```

Without caching:

```
s3::Client::new - 172.292µs
s3::Client::new - 27.875µs
s3::Client::new - 24.208µs
s3::Client::new - 23.583µs
s3::Client::new - 23.042µs
```

jdisanti:
Overall, this is an improvement over what's currently released to crates.io (again, measured on MacOS):
```
crates.io aws_config::load_defaults - 150.846792ms
crates.io s3::Client::new - 1.531208ms
crates.io s3::Client::new - 1.51575ms
crates.io s3::Client::new - 1.346666ms
crates.io s3::Client::new - 1.342542ms
crates.io s3::Client::new - 1.278125ms
crates.io list_buckets - 279.871625ms
crates.io list_buckets 2 - 66.796834ms

aws_config::load_defaults - 134.817875ms
s3::Client::new - 231.375µs
s3::Client::new - 23µs
s3::Client::new - 29.375µs
s3::Client::new - 11.125µs
s3::Client::new - 8.834µs
list_buckets - 262.579666ms
list_buckets 2 - 48.314125ms
```

The remaining large time chunks are being spent in default HTTP client initialization, which will be scrutinized in a separate PR.

jdisanti:
After more profiling, the remainder of the overhead is almost entirely the TLS handshake on first request on Linux. On Mac, loading trusted certs is slow. I think if we were to make the default HTTP client lazy, we could consider only doing it for Mac.

jdisanti:
Ran the previous release benchmark to make sure this didn't regress anything significantly:
```
compare/previous/S3 ListObjectsV2
                        time:   [34.145 µs 34.345 µs 34.527 µs]
compare/main/S3 ListObjectsV2
                        time:   [35.767 µs 35.963 µs 36.171 µs]
```
It seems to have increased request overhead by roughly 1 µs.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/rulesgen/StdLib.kt
```diff
@@ -65,19 +48,26 @@ class AwsPartitionResolver(runtimeConfig: RuntimeConfig, private val partitionsD
...
+            """partition_resolver: #{DEFAULT_PARTITION_RESOLVER}.clone()""",
             *codegenScope,
+            "DEFAULT_PARTITION_RESOLVER" to RuntimeType.forInlineFun("DEFAULT_PARTITION_RESOLVER", EndpointStdLib) {
+                rustTemplate(
+                    """
+                    // Loading the partition JSON is expensive since it involves many regex compilations,
+                    // so cache the result so that it only need to be paid for the first constructed client.
+                    pub(crate) static DEFAULT_PARTITION_RESOLVER: #{Lazy}<#{PartitionResolver}> =
+                        #{Lazy}::new(|| #{PartitionResolver}::new_from_json(b$json).expect("valid JSON"));
+                    """,
```
rcoh, line 67: we do eventually need to support runtime-loading of partitions.json but since this is all private, that seems fine

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
```diff
@@ -299,6 +299,7 @@ data class RuntimeType(val path: String, val dependency: RustDependency? = null)
         val PercentEncoding = CargoDependency.PercentEncoding.toType()
         val PrettyAssertions = CargoDependency.PrettyAssertions.toType()
         val Regex = CargoDependency.Regex.toType()
+        val RegexLite = CargoDependency.RegexLite.toType()
```
rcoh, line 302: actually unused, afaict

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
```diff
@@ -425,39 +425,37 @@ impl<'a> fmt::Display for CanonicalRequest<'a> {
...
+    } else {
+        // The normal trim function will trim non-breaking spaces and other various whitespace chars.
+        // S3 ONLY trims spaces so we use trim_matches to trim spaces only
+        Cow::Owned(
+            text.chars()
+                // Filter out consecutive spaces
+                .zip(text.chars().skip(1).chain(std::iter::once('!')))
+                .filter(|(a, b)| *a != ' ' || *b != ' ')
+                .map(|(a, _)| a)
+                .collect(),
```
rcoh, line 449: clever algorithm!

rcoh (approved):

ysaito1001 (approved):

PR #3270, opened 2023-11-29
Fix the previous release benchmark

jdisanti:
The benchmark was broken by the behavior major versions change.

----


rcoh (approved):
LGTM—should we at least be running cargo-check on this in `additional-ci`?

ysaito1001 (approved):

jdisanti:
> LGTM—should we at least be running cargo-check on this in additional-ci?

It's unfortunately not so trivial. I filed https://github.com/smithy-lang/smithy-rs/issues/3273 to track this as an ops improvement.

PR #3277, opened 2023-12-01
Improve error messaging for auth scheme selection

jdisanti:
This patch adds a stack-allocated list of explored auth schemes and the reason each didn't work to the auth orchestrator so that its error message is much easier to decipher.

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -32,3 +32,15 @@ result of the compilation."""
 references = ["aws-sdk-rust#975", "smithy-rs#3269"]
 meta = { "breaking" = false, "tada" = true, "bug" = false }
 author = "jdisanti"
+
+[[aws-sdk-rust]]
+message = "Improve the error messages for when auth fails to select an auth scheme for a request."
+references = ["aws-sdk-rust#979", "smithy-rs#3277"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
+author = "jdisanti"
+
+[[smithy-rs]]
+message = "Improve the error messages for when auth fails to select an auth scheme for a request."
+references = ["smithy-rs#3277"]
+meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client" }
+author = "jdisanti"
--- a/aws/sdk/integration-tests/dynamodb/tests/auth_scheme_error.rs
+++ b/aws/sdk/integration-tests/dynamodb/tests/auth_scheme_error.rs
@@ -6,6 +6,7 @@
 use aws_sdk_dynamodb::config::Region;
 use aws_sdk_dynamodb::error::DisplayErrorContext;
 use aws_sdk_dynamodb::{Client, Config};
+use aws_smithy_runtime::assert_str_contains;
 use aws_smithy_runtime::client::http::test_util::capture_request;
 
 #[tokio::test]
@@ -24,9 +25,8 @@ async fn auth_scheme_error() {
         .send()
         .await
         .expect_err("there is no credential provider, so this must fail");
-    let err = DisplayErrorContext(&err).to_string();
-    let expected = "\"sigv4\" wasn't an option because there was no identity resolver for it. Be sure to set an identity";
-    if !err.contains(expected) {
-        panic!("expected '{}'\nto contain '{}'", err, expected);
-    }
+    assert_str_contains!(
+        DisplayErrorContext(&err).to_string(),
+        "\"sigv4\" wasn't a valid option because there was no identity resolver for it. Be sure to set an identity"
+    );
 }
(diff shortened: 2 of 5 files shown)


aws/sdk/integration-tests/dynamodb/tests/auth_scheme_error.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    let err = client
+        .list_tables()
+        .send()
+        .await
+        .expect_err("there is no credential provider, so this must fail");
+    let err = DisplayErrorContext(&err).to_string();
+    let expected = "\"sigv4\" wasn't an option because there was no identity resolver for it. Be sure to set an identity";
+    if !err.contains(expected) {
+        panic!("expected '{}'\nto contain '{}'", err, expected);
```
rcoh, line 30: can we add `assert_contains` to test-util somewhere? 🤔

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -20,18 +20,78 @@ use std::error::Error as StdError;
...
+        if explored.items().count() == 0 {
+            return f
+                .write_str("no auth options are available. This is a bug. Please file an issue.");
+        }
+        if explored
+            .items()
+            .all(|explored| matches!(explored.result, ExploreResult::NoAuthScheme))
+        {
+            return f
+                .write_str("no auth schemes are registered. This is a bug. Please file an issue.");
```
rcoh, line 41: could it also be a problem in the model?

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -20,18 +20,78 @@ use std::error::Error as StdError;
...
+                        },
+                        ExploreResult::NoIdentityResolver => {
+                            try_add_identity = true;
+                            "there was no identity resolver for it."
+                        }
+                        ExploreResult::MissingEndpointConfig => {
+                            likely_bug = true;
+                            "it has auth config in its endpoint config, but this scheme wasn't listed in that endpoint config."
+                        },
+                        ExploreResult::NotExplored => unreachable!(),
```
rcoh, line 66: is this definitely unreachable? seems safer to write nothing/log an error than panic. could also debug_assert

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -20,18 +20,78 @@ use std::error::Error as StdError;
...
+                            likely_bug = true;
+                            "no auth scheme was registered for it."
+                        },
+                        ExploreResult::NoIdentityResolver => {
+                            try_add_identity = true;
+                            "there was no identity resolver for it."
+                        }
+                        ExploreResult::MissingEndpointConfig => {
+                            likely_bug = true;
+                            "it has auth config in its endpoint config, but this scheme wasn't listed in that endpoint config."
```
rcoh, line 64: I don't totally know what this means 😓 . What if we link the error message to a discussion on smithy-rs?

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -135,10 +202,66 @@ fn extract_endpoint_auth_scheme_config(
...
+    }
+}
+
+const MAX_EXPLORED_LIST_LEN: usize = 8;
+
+/// Stack allocated list of explored auth options for error messaging
+#[derive(Default)]
+struct ExploredList {
+    items: [ExploredAuthOption; MAX_EXPLORED_LIST_LEN],
+    len: usize,
```
rcoh, line 239: if you don't want to track this / test it, we could use `smallvec`. It's already in-tree so no compilation penalty.

rcoh (approved):
This is awesome! We should explore other places to use this pattern—I was hoping to build this generally with "error collector" but maybe we could find another way to generally build it into the orchestrator

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
jdisanti: I was thinking this would be a good idea, but `smallvec` will silently perform a heap allocation if the buffer size is exceeded. That seems undesirable for this use-case.

CHANGELOG.next.toml
```diff
@@ -33,6 +33,23 @@ references = ["aws-sdk-rust#975", "smithy-rs#3269"]
...
+        .await;
+}
+```
+
+This is designed for unit tests and using local mocks like DynamoDB Local and LocalStack with the SDK.
+"""
+meta = { "breaking" = false, "tada" = true, "bug" = false }
+author = "rcoh"
+references = ["smithy-rs#3279", "aws-sdk-rust#971"]
+
```
rcoh, line 52: I don't think you mean to have this changelog entry?

PR #3285, opened 2023-12-05
Add compile-time env var to add additional info to user agent header

jdisanti:
This PR adds a `AWS_SDK_RUST_BUILD_UA_METADATA` environment variable to allow additional metadata to be added to the SDK user agent header at compile time.

----


ysaito1001 (approved):

rcoh (commented):
We should try to lock this down/validate a little more probably — single word, alpha slashes allowed only?

jdisanti:
It's already validated by the `AdditionalMetadata::new` method, which returns an error if it's invalid.

PR #3289, opened 2023-12-06
Upgrade canary-runner to 1.x and fix bundle naming issue

jdisanti:
The SDK canary was failing due to the `.N` suffix on release tags (see https://github.com/awslabs/aws-sdk-rust/issues/993). This change upgrades the canary to the 1.x SDK and also fixes the bundle naming issue by removing periods.

----


ysaito1001 (approved):
Looks good.

PR #3292, opened 2023-12-07
Source sdk-partitions from the models instead of hardcoding

### LINKED ISSUE #2836: Use partitions.json from aws-models
Currently we use a hard coded file

rcoh:
## Motivation and Context
Fixes #2836 

This is required to gain access to hew partitions.

## Description
1. Update `sync-models` to sync from aws-sdk-rust. This location includes `sdk-partitions.json`
2. Update aws:sdk gradle file to properly set this variable in codegen.
3. Add an additional safeguard to ensure that the default hardcoded version is only used in tests and adhoc SDK generation.

## Testing
- Verify that adhoc assemble fails if you manually set the flag for a prod build

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkSettings.kt
```diff
@@ -25,16 +25,14 @@ class SdkSettings private constructor(private val awsSdk: ObjectNode?) {
...
-    val endpointsConfigPath: Path?
-        get() =
-            awsSdk?.getStringMember("endpointsConfigPath")?.orNull()?.value?.let { Paths.get(it) }
-
     /** Path to the `default-partitions.json` configuration */
     val partitionsConfigPath: Path?
         get() =
             awsSdk?.getStringMember("partitionsConfigPath")?.orNull()?.value?.let { Paths.get(it) }
 
+    val productionSdkBuild: Boolean
```
jdisanti, line 33: As this is being set to true in non-production builds, perhaps we should call it, "requirePartitionsJson"?

rcoh: I guess "production" isn't the right word. I really mean "we're generating an AWS SDK" — I think we may want this in other places, e.g. we can remove the requirement for endpoint rules based on this flag instead of requiring customers to manually set the codegen flag.

jdisanti: I think we were using "generateReadme" for that purpose previously.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkSettings.kt
```diff
@@ -47,11 +45,10 @@ class SdkSettings private constructor(private val awsSdk: ObjectNode?) {
 
     /** Whether to generate a README */
     val generateReadme: Boolean
-        get() =
-            awsSdk?.getBooleanMember("generateReadme")?.orNull()?.value ?: false
+        get() = awsSdkBuild && !(awsSdk?.getBooleanMember("suppressReadme")?.orNull()?.value ?: false)
 
     val requireEndpointResolver: Boolean
-        get() = awsSdk?.getBooleanMember("requireEndpointResolver")?.orNull()?.value ?: true
```
jdisanti, line 54: We should still look up these old values and log warnings that they're no longer read since they're used by a lot of people.

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/TestUtil.kt
```diff
@@ -55,8 +55,10 @@ fun awsIntegrationTestParams() = IntegrationTestParams(
             .withMember(
                 "awsSdk",
                 ObjectNode.builder()
-                    .withMember("generateReadme", false)
+                    .withMember("awsSdkBuild", true)
+                    .withMember("suppressReadme", true)
                     .withMember("integrationTestPath", "../sdk/integration-tests")
+                    .withMember("partitionsConfigPath", "../sdk/aws-models/sdk-endpoints.json")
```
jdisanti, line 61: Wrong config?

jdisanti (approved):

rcoh:
merged as bot because server changes are whitespace only

PR #3294, opened 2023-12-07
Allow lossless conversions from float into integral types

rcoh:
## Checklist

----


rust-runtime/aws-smithy-types/src/number.rs
```diff
@@ -333,6 +357,7 @@ mod test {
     #[test]
     fn to_i8() {
         to_signed_converter_tests!(i8);
+        i8::try_from(Number::Float(-3200000.0)).expect_err("overflow");
```
jdisanti, line 360: Should also have a test for i8 from `32.1`.

ysaito1001: And possibly a `try_from` test for its input being `Number::Float(std::f64::NAN)` ?

rcoh: I will add a manual test but this is also covered [here](https://github.com/smithy-lang/smithy-rs/pull/3294/files#diff-febb051ab1def278d41a98d95289db963c3664e01867bc1a0bdecdc79dc287d7R314)

jdisanti:
Looks good. Don't forget the changelog entry.

ysaito1001 (approved):

PR #3299, opened 2023-12-08
Stream implementation (wrapper) for PaginationStream

Ploppz:
## Motivation and Context
https://github.com/awslabs/aws-sdk-rust/discussions/995

## Description
I tried to implement futures::Stream for a wrapper struct around `PaginationStream`. I am unsure if I did it in the best way. After fighting with the borrow checker for a while I decided to try `Arc<Mutex<_>>` - is this the way to go or does there exist a better way?
Even then, does the code look correct? I used it in my project and my integration tests do pass but I am not 100% that these tests will catch any error in paginated ListObjectsV2.

I would appreciate any feedback so far.

## Testing
In progress while waiting for feedback on code

## Checklist


rcoh:
for a number of reasons, I think that _is_ the best you can do. Since you need to box the future, you then can't know for sure that the stream is alive long enough.

Luckily, there is a better way! We can expose `poll_next` on `PaginationStream` so that we can delegate to that. I've updated your branch with these changes.

I think there are a couple of remaining items we should do:

1. Add an extension trait so you can do something like:

```rust
s3.list_buckets().into_paginator().send().as_stream_03x().map(...) // and other stream trait useful functions
```

2. Add a changelog entry so folks know this new functionality is available!

Ploppz:
Thanks! Some questions:
- Why do you use `tokio-stream` rather than `futures`? I guess I will change the name of the module futures.rs
- What is `03x` in the function name?
- I assume at this point the code is so simply tests are not needed

Ploppz:
Does it look good now?

Ploppz:
Btw, if tokio-stream re-exports futures::Stream and we don't even use StreamExt for anything, then that means we simply bring in more dependencies than necessary. Shouldn't we just use `futures`?

abusch:
Nice, that would be useful to have! In the meantime we've done something pretty similar, which seems to work for us (plus an extension trait to add a `.into_stream()` method):

```rust
pub struct PaginationStreamAdapter<T>(PaginationStream<T>);

impl<T> futures::stream::Stream for PaginationStreamAdapter<T> {
    type Item = T;

    fn poll_next(
        mut self: Pin<&mut Self>,
        cx: &mut std::task::Context<'_>,
    ) -> std::task::Poll<Option<Self::Item>> {
        let fut = self.0.next();
        pin_mut!(fut);
        fut.poll(cx)
    }
}
```

Ploppz:
@abusch I'm not sure but isn't it wrong to call `next()` in every call to `poll_next`? 
Even if it somehow works, it seems wrong to me. I might be wrong though but here's my line of thought:
The async system will keep polling to check the status of whether the next item is ready, and every time it does that, internally, your impl will call the next() function which I believe (these semantics would make sense to me) should be waited for completion until it's called the next time. If you are lucky, it's always ready when polled, maybe that's why it works for you?

rcoh:
FYI I'm pretty sure that implementation is wrong — It may work due to a
coincidence of how our demand driven paginator works but I wouldn't
recommend relying on that

abusch:
Hmm, yeah, you might be right 😓

rcoh:
yeah that's fair. In that case, I think we should use `future_core` stream directly: [https://docs.rs/futures-core/latest/futures_core/stream/trait.Stream.html](futures_core::Stream) (futures::Stream) is just yet another re-export of this.

We should probably include a comment suggesting that tokio_stream is where most of the useful utilities live, though

rust-runtime/aws-smithy-types-convert/src/stream.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/// Trait to convert PaginationStream into implementor of `Stream`
+pub trait PaginationStreamExt<Item> {
+    /// Convert PaginationStream into implementor of `Stream`
+    fn as_stream_01x(self) -> PaginationStreamImplStream<Item>;
+}
+
+impl<Item> PaginationStreamExt<Item> for PaginationStream<Item> {
+    fn as_stream_01x(self) -> PaginationStreamImplStream<Item> {
```
rcoh, line 47: I wonder if this should be `into` instead since it consumes `self`. I don't have a strong opinion.

CHANGELOG.next.toml
```diff
@@ -79,3 +79,9 @@ message = "Expose local socket address from ConnectionMetadata."
 references = ["aws-sdk-rust#990"]
 meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client" }
 author = "declanvk"
+
+[[smithy-rs]]
+message = "Optional implementation of Stream on PaginationStream"
+references = ["smithy-rs#3299"]
+meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client"}
+author = "Erlend Langseth <erlend.langseth@openanalytics.eu>"
```
rcoh, line 87: we should probably also duplicate this under `aws-sdk-rust`, this will be relevant for folks using the SDK as well.

I think we should also change the message to be a little clearar about what folks can get and how, something like:

---

Add `PaginationStreamExt` extension trait to `aws-smithy-types-convert` behind the `convert-streams` feature. This makes it possible to treat a paginator as a [`futures_core::Stream`](https://docs.rs/futures-core/latest/futures_core/stream/trait.Stream.html), allowing customers to use stream combinators like [`map`](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.map) and [`filter`](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.filter).

Example: (double check this compiles)
```rust
use aws_smithy_types_convert::stream::PaginationStreamExt
s3_client.list_objects_v2().bucket("...").send().await.into_paginator().items().into_stream_03x()
```

---

rcoh (commented):
this is looking great! just a couple of suggestions then we'll get this merged. Thanks for working on this, I think customers are really going to appreciate it.

CHANGELOG.next.toml
Ploppz: can I write all that in `message` with markdown formatting? Or maybe I can add that code example in rustdocs instead?

rcoh: yep, use triple quotes to surround your message `"""` and paste it in.

rcoh: not sure if you can easily grab the markdown out of my message, if not here: https://gist.github.com/rcoh/b69e8a42509842e239cd460b68053e16

Ploppz:
I implemented the suggested changes, and rebased onto latest upstream `main` branch to fix conflicts. Also removed a trait bound that was unnecessary. Couldn't write very helpful code in the rustdoc unfortunately because we don't have a lot of dependencies to work with there.

CHANGELOG.next.toml
Ploppz: thanks, but a trick is to quote your message to get the markdown ;)

CHANGELOG.next.toml
```diff
@@ -109,3 +109,31 @@ message = "Add support for constructing [`SdkBody`] and [`ByteStream`] from `htt
...
+
+Example:
+ 
+```rust
+use aws_smithy_types_convert::stream::PaginationStreamExt
+let stream = s3_client.list_objects_v2().bucket("...").into_paginator().send().into_stream_03x();
+```
+"""
+references = ["smithy-rs#3299"]
+meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client"}
```
rcoh, line 138:
```suggestion
meta = { "breaking" = false, "tada" = false, "bug" = false }
```

CHANGELOG.next.toml
```diff
@@ -109,3 +109,31 @@ message = "Add support for constructing [`SdkBody`] and [`ByteStream`] from `htt
...
+author = "Erlend Langseth <erlend.langseth@openanalytics.eu>"
```
rcoh, line 125:
```suggestion
author = "Ploppz"
```

CHANGELOG.next.toml
```diff
@@ -109,3 +109,31 @@ message = "Add support for constructing [`SdkBody`] and [`ByteStream`] from `htt
...
+Example:
+ 
+```rust
+use aws_smithy_types_convert::stream::PaginationStreamExt
+let stream = s3_client.list_objects_v2().bucket("...").into_paginator().send().into_stream_03x();
+```
+"""
+references = ["smithy-rs#3299"]
+meta = { "breaking" = false, "tada" = false, "bug" = false }
+author = "Erlend Langseth <erlend.langseth@openanalytics.eu>"
```
rcoh, line 139:
```suggestion
author = "Ploppz"
```

rcoh (approved):

PR #3309, opened 2023-12-12
feat(aws-smithy-runtime): introduce `tls-rustls-webpki` feature

rvolosatovs:
## Motivation and Context
This allows (eventually) for downstream crates to not depend on native CA bundle, but rather use https://docs.rs/webpki-roots/latest/webpki_roots/

This is especially important, because on platforms where native CA bundle is missing, currently-used `hyper-rustls` panics at runtime (latest release uses version prior to https://github.com/rustls/hyper-rustls/commit/77154daea8663b9fd5263e70119ad145a050895e)

## Description

Provide a feature flag, which gives downstream developers a choice of the PKI bundle

## Testing
`cargo test` in respective crates with the new features (and old)

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -97,3 +97,15 @@ message = "[`Number`](https://docs.rs/aws-smithy-types/latest/aws_smithy_types/e
 references = ["smithy-rs#3294"]
 meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "all" }
 author = "rcoh"
+
+[[aws-smithy-runtime]]
+message = "`rustls-webpki` feature is now available to use [`webpki-roots`](https://docs.rs/webpki-roots/latest/webpki_roots/) CA bundle rather than native one"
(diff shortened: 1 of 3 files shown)


jdisanti:
Hey, thank you for taking the time to make this contribution!

Unfortunately, this isn't the approach we want to take for this. The problem with doing this via features is that the behavior becomes ambiguous in the case where both features are enabled, and other libraries could subtly break your TLS configuration depending on what features they enable. This kind of thing really needs to be explicit.

Addressing this will require some design. You're welcome to write a RFC in the design/rfcs directory if you still want to contribute this.

ysaito1001:
> because on platforms where native CA bundle is missing, currently-used hyper-rustls panics at runtime 

This has been addressed as of [release-2023-12-12](https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2023-12-12), i.e. loading native CA bundle will be delayed and only happen if the default HTTP client is not overridden.

PR #3314, opened 2023-12-13
Add daily credentials verification workflow

ysaito1001:
## Motivation and Context
Adds a daily credentials verification workflow

## Description
We rotate credentials manually. Those credentials are `RELEASE_AUTOMATION_BOT_CRATESIO_TOKEN` and `RELEASE_AUTOMATION_BOT_PAT`. While the validity of those credentials are checked during dry-runs of [the release workflow](https://github.com/smithy-lang/smithy-rs/blob/main/.github/workflows/release.yml) we've had instances where a dry-run failed because it was not idempotent and we nevertheless kicked off a production run, only to find out the token was invalid. This raises the need for daily credentials verification, and the PR adds one.

The workflow will check the validly of two credentials `RELEASE_AUTOMATION_BOT_CRATESIO_TOKEN` and `RELEASE_AUTOMATION_BOT_PAT`, each checked by a separate job. Upon failure, a job will notify us as follows:

Screenshot 2023-12-12 at 6 26 40 PM

## Testing
Manually triggered failures and got the messages in the above screenshot. Also verified a successful run with valid credentials.

----

### DIFF
--- a/.github/workflows/credentials-verification.yml
+++ b/.github/workflows/credentials-verification.yml
@@ -1,0 +1,51 @@
+name: Daily credentials verification
+on:
+  schedule:
+    # Runs 00:00 UTC every day
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+jobs:
+  # Verifies the token used by the bot to publish crates to crates.io
+  verify-crates-io-token:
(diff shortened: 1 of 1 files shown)


rcoh (approved):
Nice!

PR #3317, opened 2023-12-13
Allow `publisher` to put downgraded crates into `gradle.properties`

ysaito1001:
## Motivation and Context
We are currently in a circumstance where we need to publish `0.60.1` unstable runtime crates even though `0.61.0` were published to `crates.io` in the previous release.

## Description
This PR allows `publisher` to specify downgraded runtime crate versions in `gradle.properties` via the `--allow-downgrade` flag and updates a script `upgrade-gradle-properties` to use that flag. This is a bare minimum change in that we could update a release workflow to have a new checkbox whether we allow downgrade or not and thread through that value to the `upgrade-gradle-properties` script, instead of directly specifying like so.

## Testing
Added unit tests for `publisher` and manually ran `publisher upgrade-runtime-crates-version` with `--allow-downgrade` to check the functionality.

----


ysaito1001:
Closing this on behalf of PR3316

PR #3318, opened 2023-12-14
Fix subtle break of endpoint prefixes due to semver

jdisanti:
When we released smithy-rs release-2023-12-08, we introduced a silent failure for endpoint prefixes, and not in the newly released version, but in the previous releases. There is a subtle issue with semver that causes this. This PR addresses the endpoint prefix part of this problem. Other PRs will fix other parts that are broken by this semver issue.

The issue is that unstable (0.x) runtime crates are declaring types that get placed into the `ConfigBag`, and these types are referenced in the `ConfigBag` across crate boundaries. This by itself isn't a problem, but because our stable 1.x crates depend on the unstable crates, it becomes a problem. By releasing 1.1.0 that depends on 0.61, consumers of 1.x pull in both 0.60 and 0.61. The generated code pulls in 0.60, and the 1.1.x crates pull in 0.61. This is fine since two semver-incompatible crate versions can be in the dependency closure. Thus, the generated code which is using 0.60 is placing a 0.60 type in the `ConfigBag`, and the runtime crates that pull the type out of the `ConfigBag` are expecting a 0.61 version of it. This leads to the type not existing upon lookup, which then causes the silent break.

This PR fixes this by moving the `EndpointPrefix` type and its associated methods into `aws-smithy-runtime-api`, a stable crate. The `aws-smithy-http` unstable crate is updated to point to the new stable version so that a patch release of that crate will solve the issue for old versions going forward as well.

----


jdisanti:
I think cargo-semver-checks is wrong about this being breaking.

rust-runtime/aws-smithy-http/src/endpoint.rs
```diff
@@ -100,3 +82,23 @@ fn merge_paths<'a>(endpoint: &'a Uri, uri: &'a Uri) -> Cow<'a, str> {
...
+
+    /// An error that occurs when an endpoint is found to be invalid. This usually occurs due to an
+    /// incomplete URI.
+    #[deprecated(
+        since = "0.60.1",
+        note = "Use aws_smithy_runtime_api::client::endpoint::error::InvalidEndpointError instead."
+    )]
+    pub type InvalidEndpointError =
+        aws_smithy_runtime_api::client::endpoint::error::InvalidEndpointError;
+}
```
rcoh, line 104: `ResolveEndpointError` is used by generated code (and I don't see this CR updating that. I'm surprised tests didn't fail from the deprecation) (which makes me concerned)

rust-runtime/aws-smithy-http/src/endpoint.rs
```diff
@@ -5,58 +5,40 @@
...
 /// This method mutates `uri` by setting the `endpoint` on it
+#[deprecated(
+    since = "0.60.1",
+    note = "Use aws_smithy_runtime::client::endpoint::apply_endpoint instead."
+)]
 pub fn apply_endpoint(
     uri: &mut Uri,
     endpoint: &Uri,
     prefix: Option<&EndpointPrefix>,
 ) -> StdResult<(), InvalidEndpointError> {
```
rcoh, line 40: I think we can actually leave `ResolveEndpointError` and `InvalidEndpointError` in this crate unchanged and make a new version of it in the new crate that this function returns. Then we avoid any semver risks with these error types. This type can be removed in 0.61.

jdisanti: Good point. I restored them in cc1a98aff156f04da9aca7db3d61a8c3c94e2407.

rust-runtime/aws-smithy-http/src/endpoint.rs
jdisanti: I don't think it's used by generated code, but it is used by the orchestrator.

rcoh (approved):

ysaito1001 (approved):

PR #3325, opened 2023-12-15
Move `Metadata` config bag type into a stable crate

jdisanti:
This issue addresses a semver compatibility problem similar to the one described in #3318, except for the `aws_smithy_http::operation::Metadata` type.

----


ysaito1001 (commented):
LGTM. Will approve once `CHANGELOG.next.toml` has been updated.

ysaito1001 (approved):
It this kind of chicken-egg problem in terms of `cargo semver-checks`? We need move `Metadata` to a stable crate so we can eventually bump a minior version for `aws-smithy-http` from `0.60.1` to `0.61.0`. But to `cargo semver-checks`, this is a breaking change as a pub item has been moved, requiring us to have `0.61.0` for `aws-smithy-http` with this change. But I suppose we can override what the tool says for this particular instance and claim that this is necessary for the grand scheme of things.

rcoh (approved):

PR #3331, opened 2023-12-18
Fix BadSSL by building everything from source

rcoh:
## Motivation and Context
The TLS test broke. Why? Because Bad SSL stopped working.

Why did BadSSL stop working? Because it used an ancient version of ruby and it couldn't install packages anymore.

So I:
- Got it working on a newer version of ruby
- But that only work on Ubuntu 22.04.
- The version of nginx/openssl that you can install on 22.04 version actually serve these terrible certificates.

So instead, we compile nginx and openssl from source.

I also fixed things up so they won't fail silently in the future.

## Description
Mostly just sadness.

## Testing
The check passes again.

## Checklist

----

### DIFF
--- a/tools/ci-scripts/configure-tls/new-badssl-dockerfile
+++ b/tools/ci-scripts/configure-tls/new-badssl-dockerfile
@@ -1,0 +1,24 @@
+# Start with Ubuntu 16.04 (LTS), and build badssl.com up from there
+FROM ubuntu:22.04
+MAINTAINER April King <april@pokeinthe.io>
+EXPOSE 80 443
+RUN apt-get update && apt-get install -y apt-transport-https
+RUN apt-get install -y software-properties-common
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    git \
+    libffi-dev \
+    make \
+    nginx \
+    ruby3.0 \
+    ruby3.0-dev
+#RUN gem update --system
+RUN gem install jekyll
+
+# Install badssl.com
+ADD . badssl.com
+WORKDIR badssl.com
+RUN make inside-docker
+
+# Start things up!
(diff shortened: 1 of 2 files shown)


tools/ci-scripts/configure-tls/new-badssl-dockerfile
```diff
@@ -0,0 +1,88 @@
...
+# Update the nginx config to include the badssl configs.
+RUN head -n-1 /etc/nginx/nginx.conf > wip.conf
+RUN echo "# Virtual Host Configs\ninclude /var/www/badssl/_site/nginx.conf;\n}" >> wip.conf
+RUN mv wip.conf /usr/local/nginx/conf/nginx.conf
+RUN make inside-docker
+
+# Allow unsecure certs
+RUN sed -i 's/SECLEVEL=2/SECLEVEL=0/' /etc/ssl/openssl.cnf
+
+# Fix DH key that can't be generated...works in docker bug not on github. Who knows.
```
jdisanti, line 80:
```suggestion
# Fix DH key that can't be generated...works in docker but not on github. Who knows.
```

tools/ci-scripts/configure-tls/new-badssl-dockerfile
```diff
@@ -0,0 +1,88 @@
+# Why does this file exist?
+# badssl seems to be abandoned. The orginal Dockerfile was based on ubuntu 16.04 and all the bits were rotting.
```
jdisanti, line 2:
```suggestion
# badssl seems to be abandoned. The original Dockerfile (https://github.com/chromium/badssl.com/blob/master/Dockerfile)
# was based on ubuntu 16.04 and all the bits were rotting.
```

jdisanti (commented):
Looks good. Just some minor comment fixes.

jdisanti (approved):

PR #3332, opened 2023-12-20
Implement runtime-versioner audit tool

jdisanti:
This PR implements a build tool to audit runtime crate versions as part of CI and release. If a runtime crate doesn't have the special `0.0.0-smithy-rs-head` version number, then it is assumed to be independently versioned, and the audit tool will verify it is version bumped when any changes are made to it.

Given that there isn't a complete/reliable semver checking tool for Rust yet, this tool isn't smart. It will rely on devs to correctly bump the version number when that is done.

----


tools/ci-build/runtime-versioner/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "runtime-versioner"
+version = "0.1.0"
+authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>"]
+description = "Tool that managed runtime crate versions."
```
rcoh, line 5:
```suggestion
description = "Tool for managing runtime crate versions."
```

tools/ci-build/runtime-versioner/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Don't `git fetch` before auditing.
+    #[arg(long)]
+    no_fetch: bool,
+    /// Explicitly state the previous release's tag. Discovers it if not provided.
+    #[arg(long)]
+    previous_release_tag: Option<String>,
+    /// Path to smithy-rs. Defaults to current working directory.
+    #[arg(long)]
+    smithy_rs_path: Option<Utf8PathBuf>,
+    /// (For testing) Path to a fake crates.io index.
```
rcoh, line 33: are there instructions somewhere for how to create this?

tools/ci-build/runtime-versioner/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    smithy_rs_path: Option<Utf8PathBuf>,
+}
+
+#[derive(clap::Parser, Clone)]
+#[clap(author, version, about)]
+enum Command {
+    /// Audit the runtime crate versions in the smithy-rs repo at HEAD
+    ///
+    /// Requires a full clone of smithy-rs. Will not work against shallow clones.
+    Audit(Audit),
```
rcoh, line 51: can we add a little more description about what it's actually auditing?

tools/ci-build/runtime-versioner/src/command.rs
```diff
@@ -0,0 +1,53 @@
...
+/// Extension trait to make `Command` nicer to work with.
+pub trait CommandExt {
+    fn expect_success_output(&mut self, context: &str) -> Result<String>;
+
+    fn expect_statuses(&mut self, statuses: impl IntoIterator<Item = i32>) -> Result<i32>;
+}
+
+impl CommandExt for Command {
+    fn expect_success_output(&mut self, context: &str) -> Result<String> {
+        let output = self
```
rcoh, line 18: I feel like we have Command helpers in a few places? is there a common we can put it in? Or a 3p library that does this?

tools/ci-build/runtime-versioner/src/command.rs
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+use anyhow::{bail, Context, Result};
+use std::process::Command;
+
+/// Extension trait to make `Command` nicer to work with.
+pub trait CommandExt {
+    fn expect_success_output(&mut self, context: &str) -> Result<String>;
+
+    fn expect_statuses(&mut self, statuses: impl IntoIterator<Item = i32>) -> Result<i32>;
```
rcoh, line 13: from reading the code, I can see this is expecting that the status is one of a member of the list—didn't get that from name. Maybe a doc comment? Or `expect_status_one_of`

tools/ci-build/runtime-versioner/src/index.rs
```diff
@@ -0,0 +1,102 @@
...
+                            Crate::from_slice(&bytes).map_err(Error::from).map(Some)
+                        }
+                        StatusCode::NOT_FOUND => Ok(None),
+                        status => {
+                            let body = String::from_utf8_lossy(&bytes);
+                            Err(anyhow!(
+                                "request to crates.io index failed ({status}):\n{body}"
+                            ))
+                        }
+                    })
```
rcoh, line 59: is this going to be run by humans or in CI? may want retries at some layer—crates.io can often be down. (Or is this hitting GitHub, same thing...). 

if it's hitting GitHub, unauthenticated requests are often throttled very quickly

tools/ci-build/runtime-versioner/src/tag.rs
```diff
@@ -0,0 +1,131 @@
...
+use anyhow::{anyhow, bail, Context, Result};
+use smithy_rs_tool_common::release_tag::ReleaseTag;
+use std::str::FromStr;
+use tracing::warn;
+
+/// Discovers and returns the tag of the previous release for the revision at HEAD.
+pub fn previous_release_tag(
+    repo: &Repo,
+    release_tags: &[ReleaseTag],
+    hint: Option<&str>,
```
rcoh, line 16: what's `hint` here? it seems like it's a manually specified tag?

tools/ci-build/runtime-versioner/src/tag.rs
```diff
@@ -0,0 +1,131 @@
...
+// Function assumes the given tag is known to be suffixed.
+fn strip_describe_tags_suffix(tag: &str) -> Option<&str> {
+    // Example release tag with suffix: release-2023-12-01-42-g885048e40
+    let end_index = tag
+        .char_indices()
+        .rev()
+        .skip_while(|(_, c)| *c != '-')
+        .skip(1)
+        .find(|(_, c)| *c == '-')
+        .map(|(i, _)| i);
```
rcoh, line 68: `tag.rsplit('-')`?

tools/ci-build/runtime-versioner/src/tag.rs
```diff
@@ -0,0 +1,131 @@
...
+    let mut tags: Vec<_> = repo
+        .git(["tag"])
+        .expect_success_output("find the current ancestor release tag")?
+        .lines()
+        .flat_map(|tag| match ReleaseTag::from_str(tag) {
+            Ok(tag) => Some(tag),
+            Err(_) => {
+                if !tag.starts_with("v0.") {
+                    warn!("ignoring tag '{tag}': doesn't look like a release tag");
+                }
```
rcoh, line 87: we have old tags that look like this and `release-*` tags I guess?

tools/ci-build/runtime-versioner/src/audit.rs
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+    if errors.is_empty() {
+        println!("SUCCESS");
+        Ok(())
+    } else {
+        for error in errors {
+            eprintln!("{error}");
+        }
+        Err(anyhow!("there are audit failures in the runtime crates"))
```
rcoh, line 56: `bail!`?

rcoh (commented):
I haven't really dug into the weeds yet, here is a first pass

tools/ci-build/runtime-versioner/src/main.rs
jdisanti: It's a simple toml file of package name to version number. There are examples in the tests.

tools/ci-build/runtime-versioner/src/tag.rs
jdisanti: Yeah. The conditional avoids logging a bunch of useless warnings for our older tags.

tools/ci-build/runtime-versioner/src/index.rs
jdisanti: Good call. Will need to do some refactoring to pull out the retry logic in the publisher tool into common.

tools/ci-build/runtime-versioner/src/command.rs
jdisanti: We do, but I'm not so happy with them. The previous ones were designed to make unit testing with mocked or faked CLI commands easier, but I've been thinking lately that that's a harder path to take. It's easier to make fake git repositories and integration test against them.

I think the old ones still have a place for the publisher tool though, or at the very least, it would take a non-trivial amount of effort to change them. It's useful to mock out some of the cargo commands for that tool in testing.

I decided to move these extension functions into common and put in a comment noting when one should be used over the other.

tools/ci-build/runtime-versioner/src/audit.rs
```diff
@@ -0,0 +1,260 @@
...
+    let release_tags = release_tags(&repo)?;
+    let previous_release_tag =
+        previous_release_tag(&repo, &release_tags, args.previous_release_tag.as_deref())?;
+    if release_tags.first() != Some(&previous_release_tag) {
+        tracing::warn!("there are newer releases since '{previous_release_tag}'");
+    }
+
+    let next_crates = discover_runtime_crates(&repo.root).context("next")?;
+    let previous_crates = resolve_previous_crates(&repo, previous_release_tag.as_str())?;
+
```
rcoh, line 39: if you want to run this locally you could also stash and unstash

tools/ci-build/runtime-versioner/src/audit.rs
```diff
@@ -0,0 +1,260 @@
...
+                    rt_crate.next_release_version,
+                );
+            } else {
+                tracing::info!(
+                    "'{}' is a new crate (or wasn't independently versioned before) and will publish at {}",
+                    rt_crate.name,
+                    rt_crate.next_release_version,
+                );
+            }
+            Ok(())
```
rcoh, line 76: I wonder if it would be practical to invoke cargo semverchecks at this point...

tools/ci-build/runtime-versioner/src/audit.rs
```diff
@@ -0,0 +1,260 @@
...
+                toml::from_slice(&fs::read(&manifest_path).context("failed to read manifest")?)
+                    .context("failed to parse manifest")?;
+            let publish = manifest["package"]
+                .get("publish")
+                .and_then(|p| p.as_bool())
+                .unwrap_or(true);
+            let version = manifest["package"]["version"]
+                .as_str()
+                .expect("version is a string");
+            if publish && version != "0.0.0-smithy-rs-head" {
```
rcoh, line 193: I'm _slightly_ concerned about this check going in this direction. Is there anything that mandates those versions? Just want to make sure we don't have a crate with `0.0.0-smithy-rs` or something slightly different by accident.

Side note...what if we set this to `0.0.0-smithy-rs-stable`, `0.0.0-smithy-rs-unstable`....

tools/ci-build/runtime-versioner/src/audit.rs
```diff
@@ -0,0 +1,260 @@
...
+use anyhow::{anyhow, bail, Context, Result};
+use camino::{Utf8Path, Utf8PathBuf};
+use smithy_rs_tool_common::release_tag::ReleaseTag;
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fs,
+    process::Command,
+};
+
+pub fn audit(args: Audit) -> Result<()> {
```
rcoh, line 23: how does this scheme play with the sdk-versioner?

rcoh (approved):

tools/ci-build/runtime-versioner/src/tag.rs
jdisanti: Yeah, it's a manually specified tag. Renamed to be clearer.

tools/ci-build/runtime-versioner/src/tag.rs
jdisanti: Good suggestion. It's much cleaner as `tag.rsplitn(3, '-').skip(2).next()`.

tools/ci-build/runtime-versioner/src/audit.rs
jdisanti: Given its current inability to work across crates due to rustdoc limitations, it doesn't seem particularly useful.

tools/ci-build/runtime-versioner/src/audit.rs
jdisanti: I posted an update that checks out the previous revision to a temp directory so that it can work locally.

tools/ci-build/runtime-versioner/src/audit.rs
jdisanti: The runtime crate versions are set as part of the copy of runtime crates into the SDK (in the SDK's build.gradle.kts), so sdk-versioner doesn't touch those. It does, however, read them and update other crates to use those versions.

Similarly, the smithy-rs release sets the runtime crate versions as part of the copy in its build.gradle.kts.

I just tried setting aws-smithy-eventstream to 0.70.0 and generating a SDK and a smithy-rs release. The SDK generated and compiled successfully, but the smithy-rs release broke due to an overzealous validation in the fix-manifests tool. I'll work on correcting that.

tools/ci-build/runtime-versioner/src/audit.rs
jdisanti: > I'm slightly concerned about this check going in this direction. Is there anything that mandates those versions? Just want to make sure we don't have a crate with 0.0.0-smithy-rs or something slightly different by accident.

It's a fair concern, and if we were going to keep this situation forever, we might want to think of something to avoid it. But I think we should push for getting rid of the stable/unstable crate versions soon after independent runtime crate versions have been proven.

> Side note...what if we set this to 0.0.0-smithy-rs-stable, 0.0.0-smithy-rs-unstable....

It's an interesting idea. Maybe we should have done it that way to begin with a couple months back. I'm not sure it's worth the investment to refactor now though.

rcoh:
side note: We need to consider the edge case where we've tagged multiple releases but they haven't necessarily been released (at least the AWS SDK crates haven't)

tools/ci-build/runtime-versioner/src/audit.rs
rcoh: reasonable — maybe we should have a check that it's either 0.0.0-... or a real semver version

rcoh (changes requested):
0.70 eventstream version

rust-runtime/aws-smithy-eventstream/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-eventstream"
-version = "0.0.0-smithy-rs-head"
+version = "0.70.0"
```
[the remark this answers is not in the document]
jdisanti: Good catch. I definitely didn't intend to keep this in here. It was only for local testing.

aws/sdk/build.gradle.kts
```diff
@@ -359,11 +359,7 @@ tasks.register<ExecRustBuildTool>("fixManifests") {
 
     toolPath = publisherToolPath
     binaryName = "publisher"
-    arguments = mutableListOf("fix-manifests", "--location", outputDir.asFile.absolutePath).apply {
-        if (crateVersioner.independentVersioningEnabled()) {
-            add("--disable-version-number-validation")
-        }
-    }
+    arguments = mutableListOf("fix-manifests", "--location", outputDir.asFile.absolutePath)
```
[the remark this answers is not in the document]
jdisanti: Yeah, looks like it's safe to remove.

jdisanti:
> We need to consider the edge case where we've tagged multiple releases but they haven't necessarily been released (at least the AWS SDK crates haven't)

I think this should just work for the most part. There is one situation where it could pose a small amount of risk, and that's where a patch version is staged for release (but not yet published) for a runtime crate. In that case, it won't detect that the version number is already taken in a PR to main. I don't think there's a good way to detect that though without having it be aware of release branches and which one is current.

rcoh (approved):

PR #3337, opened 2023-12-21
Fix floating-point number default value instantiation

jdisanti:
When a floating-point shape has a default value without a decimal place, the instantiator instantiates it with an integer value, which causes compilation failure:

```
   --> sdk/connect/src/types/_user_proficiency.rs:109:49
    |
109 |             level: self.level.unwrap_or_else(|| 1),
    |                                                 ^
    |                                                 |
    |                                                 expected `f32`, found integer
    |                                                 help: use a float literal: `1.0`
```

## Checklist

----


rcoh:
fixed elsewhere, I'm pulling the new test into a separate PR though

PR #3343, opened 2023-12-29
- Uncommented several lines of commented-out code to make the example…

vronin:
- Uncommented several lines of commented-out code to make the example…

- I removed #[cfg(feature = "http0-compat")]. It's only useful if an example is built as part of the aws-sigv4 crate (that has a definition of this feature). However, there is no value in it when you copy it into a client app.

## Motivation and Context

I was trying to use aws-sigv4 for request signing and found an example of how to use it here.
https://docs.rs/aws-sigv4/1.1.1/aws_sigv4/http_request/index.html

The example needed to be completed (it didn't build).

I checked the code and found out that it looked like a formatting error (when somebody was converting it from commented-out code to documentation.

The change makes the example complete and buildable. 

## Description

The change
- Uncommented accidentally commented code
- Removes unnecessary #[cfg()] check 

## Testing

I included the sample in my client app and ensured it builds. 

Since the change in documentation of example, it can't affect any other areas of the code.

----

### DIFF
--- a/aws/rust-runtime/aws-sigv4/src/http_request.rs
+++ b/aws/rust-runtime/aws-sigv4/src/http_request.rs
@@ -10,10 +10,9 @@
 //! **Note**: This requires `http0-compat` to be enabled.
 //!
 //! ```rust
-//! # use aws_credential_types::Credentials;
+//! use aws_credential_types::Credentials;
 //! use aws_smithy_runtime_api::client::identity::Identity;
-//! # use aws_sigv4::http_request::SignableBody;
-//! #[cfg(feature = "http0-compat")]
+//! use aws_sigv4::http_request::SignableBody;
 //! fn test() -> Result<(), aws_sigv4::http_request::SigningError> {
 //! use aws_sigv4::http_request::{sign, SigningSettings, SigningParams, SignableRequest};
 //! use aws_sigv4::sign::v4;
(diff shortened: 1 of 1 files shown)


aws/rust-runtime/aws-sigv4/src/http_request.rs
```diff
@@ -10,10 +10,9 @@
 //! **Note**: This requires `http0-compat` to be enabled.
 //!
 //! ```rust
-//! # use aws_credential_types::Credentials;
+//! use aws_credential_types::Credentials;
 //! use aws_smithy_runtime_api::client::identity::Identity;
-//! # use aws_sigv4::http_request::SignableBody;
-//! #[cfg(feature = "http0-compat")]
```
rcoh, line 16: as the docs state above, this example requires the `http0-compat` feature to be enabled (that's why the example is gated on this feature)

vronin: Maybe I am missing something. Per my understanding, it requires the http0-compat feature to be enabled in Cargo.toml

`aws-sigv4 = { version = "1.1.1", default-features = false, features = ["http0-compat", "sign-http", "sign-eventstream", "sigv4a", "num-bigint"] }`

However, when I copy the example code into my client app,  #[cfg(feature = "http0-compat")] code will check whether http0-compact feature is enabled in my app (vs. on aws-sigv4 crate)

rcoh: oh I see what you're saying! Yeah the fix here is actually to add an additional `#` at the beginning of those lines to suppress them from docs.rs

We should suppress the `fn` test line too

The example itself needs the config gating so that it will compile even when this feature isn't enabled.

vronin: @rcoh Ah... I got your point. This example needs to be build as part of rustdocs and it should not be built when the feature is not enabled. I will update PR

PR #3351, opened 2024-01-10
Resolve runtime crate versions from manifests first

jdisanti:
This PR changes codegen-core's build-time runtime crate version resolution to pull versions from the actual runtime crate manifest files, only changing them to the version numbers in gradle.properties if they use the special `0.0.0-smithy-rs-head` version number. This is a prerequisite step in allowing independent crate versioning.

----


rcoh (approved):
LGTM—we should add a test target at some point that exercises this behavior. It isn't currently because codegen-client-test uses path dependencies.

PR #3355, opened 2024-01-11
Move `aws-http` types into `aws-runtime`

jdisanti:
This issue addresses a semver compatibility problem similar to the one described in https://github.com/smithy-lang/smithy-rs/pull/3318, except for the storable types in the aws-http crate. I opted to move all of aws-http into aws-runtime since there was so little in there.

----


rcoh:
how does this not break current clients that depend on aws-http? I guess we would MV or delete aws-http?

jdisanti:
> how does this not break current clients that depend on aws-http? I guess we would MV or delete aws-http?

I think this will need to get released as a new minor version of aws-http so that it's not picked up automatically.

rcoh:
we should carefully audit everything that's `pub` in here and consider if we want it put in `aws-runtime`. I suspect much of it doesn't need to be anymore since we removed the crate boundary?

aws/rust-runtime/aws-http/external-types.toml
```diff
@@ -1,11 +1,3 @@
 allowed_external_types = [
-    "aws_smithy_runtime_api::http::headers::Headers",
```
rcoh, line 2: is this actually true now? I thought the crate was empty

aws/rust-runtime/aws-http/Cargo.toml
```diff
@@ -2,24 +2,13 @@
 name = "aws-http"
 version = "0.0.0-smithy-rs-head"
```
rcoh, line 3: I think you need to change this version to some sort of EOL version? we need to be careful **not** to publish this with the next set of unstable crates

aws/rust-runtime/aws-http/external-types.toml
jdisanti: The crate re-exports the types that were moved into aws-runtime.

aws/rust-runtime/aws-http/Cargo.toml
jdisanti: Good call. Updated.

jdisanti:
> we should carefully audit everything that's pub in here and consider if we want it put in aws-runtime. I suspect much of it doesn't need to be anymore since we removed the crate boundary?

I took a pass over the types, and I don't see any opportunities here. The most we could do is choose not to remove/re-export something that's currently in aws-http, but everything looks to be used across crate boundaries.

rcoh (approved):

rcoh (approved):

PR #3359, opened 2024-01-11
Read the list of identity resolvers backwards

rcoh:
## Motivation and Context
- https://github.com/awslabs/aws-sdk-rust/issues/901

## Description
When merging RuntimeComponentBuilders, the vecs are appended. However, this means the most recently added auth options are at the end. If `sigv4` is provided multiple times, this means that only the original auth scheme will actually be used.

## Testing
New UT that fails without this change and passes after.

## Checklist

----


rcoh:
closed in favor of #3363

PR #3327, opened 2023-12-15
Add tool to simulate a release dry run

rcoh:
## Motivation and Context
As we discovered in #3318, it's possible to cause unexpected breakage by releasing runtime crates.
 
This tool automates setting up `aws-sdk-rust` (and manually, you can copy the patches into `~/.cargo/config.toml` for your entire system), to simulate the release of new runtime crates.

Screenshot 2023-12-15 at 12 58 41 PM

### Testing
With this tool (and a small hack—I had to simulate releasing 1.1.100 so that the versions became the latest), I correctly caught the breaking changes from the previous release.

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3369, opened 2024-01-17
Make SDK runtime patch tool that supports SDK runtime crates

jdisanti:
A tool to patch in new versions of smithy-rs runtime crates was introduced in #3327 so that semver breaks could better be detected for upcoming releases. However, this tool didn't support patching in SDK runtime crates, which is necessary for testing #3355.

This PR moves the tool into the runtime-versioner tool, and adjusts it to support both smithy-rs and SDK runtime crates.

----


tools/ci-build/runtime-versioner/src/command/patch.rs
```diff
@@ -0,0 +1,187 @@
...
+                args.unstable_crate_version
+            ));
+        } else {
+            new_props.push_str(line);
+        }
+        new_props.push('\n');
+    }
+    fs::write(&props_path, new_props).context("failed to write smithy-rs/gradle.properties")?;
+    Ok(())
+}
```
rcoh, line 113: can we provide this as `-P` instead? Then this will work with a dirty working tree.

tools/ci-build/runtime-versioner/Cargo.toml
```diff
@@ -18,6 +18,8 @@ anyhow = "1.0.75"
 camino = "1.1.6"
 clap = { version = "4.4.11", features = ["derive"] }
 crates-index = "2.3.0"
+fs_extra = "1.3.0"
```
rcoh, line 21: is `fs_extra` actually used? I removed it and everything still built / tested fine.

rcoh (approved):

rcoh (approved):
LGTM—Just remove the unused dep

PR #3371, opened 2024-01-17
Defer event stream semver hazard fix

jdisanti:
Defer fixing the event stream semver hazard.

----


rcoh (approved):
LGTM. We should add a todo during the next release to audit this version in the artifact.

PR #3378, opened 2024-01-19
Add tracing calls when an extension parameter cannot be constructed

drganjoo:
## Motivation and Context

When a user defines a handler function with additional parameters, these parameters are constructed using the `FromRequest` trait implementation specific to their types. If `FromRequest` fails to construct the parameter – for instance, if `Extension<State>` is expected by the user but `Extension<Arc<State>>` is actually added by the extension layer – the server currently does not log a message indicating this construction failure. Instead, it simply returns a 500 internal server error.

## Description

`tracing::{trace, error}` calls have been added.

## Breaking change

`FromParts<P>::Rejection` **must** implement `std::fmt::Display`.

----


drganjoo:
Since `tracing::Value` is sealed, we require that `Rejections` implement `std:error::Error` in the implementation of the `FromRequest` and `FromParts` traits.

david-perez (approved):

PR #3379, opened 2024-01-20
Implement support for `sso-session` in AWS config file

jdisanti:
This PR implements parsing support for `[sso-session name]` in the `~/.aws/config` file.

----


aws/rust-runtime/aws-config/src/profile/parser/normalize.rs
```diff
@@ -3,59 +3,103 @@
...
+    Default {
+        /// True when it is `[profile default]`
+        prefixed: bool,
+    },
+    /// `[profile name]` or `[name]`
+    Profile {
+        name: Cow<'a, str>,
+        /// True if prefixed with `profile`.
+        prefixed: bool,
+    },
```
ysaito1001, line 31: Much easier to read with these variants. I like them 👍

ysaito1001 (approved):
LGTM!

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -143,35 +167,151 @@ impl ProfileSet {
...
+        self.0.name()
     }
 
     /// Returns a reference to the property named `name`
     pub fn get(&self, name: &str) -> Option<&str> {
-        self.properties
-            .get(to_ascii_lowercase(name).as_ref())
-            .map(|prop| prop.value())
+        self.0.get(name)
+    }
```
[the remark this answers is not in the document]
jdisanti: Yeah, they are. However, removing them will be a breaking change since importing `Section` would be required. I added them to `SsoSession` to be consistent.

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -373,11 +527,28 @@ mod test {
...
+                    sso_sessions,
+                },
+            ) => {
+                if profiles != &actual_profiles {
+                    Err(format!(
+                        "mismatched profiles:\nExpected: {profiles:#?}\nActual: {actual_profiles:#?}",
+                    ))
+                } else if sso_sessions != &actual_sso_sessions {
+                    Err(format!(
+                        "mismatched sso_sessions:\nExpected: {actual_profiles:#?}\nActual: {actual_sso_sessions:#?}",
```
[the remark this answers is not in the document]
jdisanti: Good catch!

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -143,35 +167,151 @@ impl ProfileSet {
...
+    }
+
+    fn insert(&mut self, name: String, value: Property) {
+        self.0.insert(name, value)
+    }
+}
+
+/// A `[sso-session name]` section in the config.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct SsoSession(SectionInner);
```
rcoh, line 274: do we want this public? One might imagine a different API like:

```rust
let x: ProfileSet = ...;
let subsection: Option<&Profile> = x.get_extra_section("sso-session);
```

This keeps our public API narrower and doesn't hard code the concept of sso-sessions into the (currently) generic profile file parser.

Internally the profile parser _could_ still store the same repr, or we could store `extra_sections: HashMap<String, Profile>`

Maybe I'm missing something about the API that makes this impossible?

On a side note, we don't actually need to make this code public _at all_ since it's only needed internally.

aws/rust-runtime/aws-config/src/profile/mod.rs
```diff
@@ -16,7 +16,7 @@ mod parser;
 #[doc(inline)]
 pub use parser::ProfileParseError;
 #[doc(inline)]
-pub use parser::{load, Profile, ProfileFileLoadError, ProfileSet, Property};
+pub use parser::{load, Profile, ProfileFileLoadError, ProfileSet, Property, SsoSession};
```
rcoh, line 19: should carefully consider if we want to make this public. I don't think it's required to be right now.

aws/rust-runtime/aws-config/src/sso/token.rs
```diff
@@ -321,14 +321,10 @@ impl Builder {
         if self.sdk_config.is_none() {
             self.sdk_config = Some(crate::load_defaults(crate::BehaviorVersion::latest()).await);
         }
-        self.build_with(Env::real(), Fs::real())
+        self.build_sync(Env::real(), Fs::real())
     }
 
-    pub(crate) fn build_sync(self) -> SsoTokenProvider {
```
rcoh, line 327: `build_sync` is a little bit a confusing name since its actual functionality is "build without loading new credentials from the environment, not "build but use spawn_blocking" or something along those lines.

Maybe `build_as_configured` or something? Could also just go back to `build_with` as the function name

aws/rust-runtime/aws-config/test-data/profile-parser-tests.json
```diff
@@ -748,12 +850,48 @@
...
+      },
+      "output": {
+        "config": {
+          "profiles": {
+            "sso-sessionfoo": { "name": "value" },
+            "bar": {}
           },
-          "foo": {
-            "name": "value"
+          "sso_sessions": {
```
rcoh, line 894: should probably include a test of something like `sso-sesion blah` (a typo'd sso session key)

rcoh (changes requested):

rcoh (changes requested):
biggest thing is the panic, most of the other stuff is less blocking

CHANGELOG.next.toml
```diff
@@ -9,4 +9,10 @@
 # message = "Fix typos in module documentation for generated crates"
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
-# author = "rcoh"
\ No newline at end of file
+# author = "rcoh"
+
+[[aws-sdk-rust]]
+message = "Add support for `[sso-session]` in AWS config file for AWS Identity Center SSO credentials."
+references = ["aws-sdk-rust#703", "smithy-rs#3379"]
```
rcoh, line 16: would probably be helpful to include an example profile in the message. Since this doesn't support loading tokens yet (right?) it should probably call that out and link to the tracking issue.

aws/rust-runtime/aws-config/src/profile/parser/normalize.rs
```diff
@@ -3,59 +3,103 @@
...
+    /// `[sso-session name]`
+    SsoSession { name: Cow<'a, str> },
```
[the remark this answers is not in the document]
jdisanti: It's private to the crate, and I think having it explicit like this makes the code easier to read and reason about. Thus, I'm inclined to keep it this way for now. We can refactor as needed if anything other than `sso-session` is ever added later.

aws/rust-runtime/aws-config/src/profile/parser.rs
jdisanti: What makes it an "extra session" though? My understanding is that anything that is unrecognized becomes a profile name.

jdisanti: Hmmm. From what I can see in the profile parser tests, it actually seems to ignore anything with a space in it if it doesn't start with `profile`.

aws/rust-runtime/aws-config/src/profile/credentials/exec.rs
```diff
@@ -130,21 +130,33 @@ impl ProviderChain {
...
+                        // TODO(https://github.com/awslabs/aws-sdk-rust/issues/703): implement token provider support
+                        unimplemented!(
+                            "Using SSO tokens auth without resolving credentials via \
+                             assuming a role from a config file hasn't been implemented yet. \
+                             You can work around this for now by setting `sso_role_name` and \
+                             `sso_account_id`, but this won't work for services that use a \
+                             Builder ID, such as Code Catalyst. \
+                             Tracking in https://github.com/awslabs/aws-sdk-rust/issues/703"
+                        )
+                    }
```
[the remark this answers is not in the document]
jdisanti: Ah, good call. I didn't think of that.

For services that support access tokens, this code will be unreachable since the orchestrator will select the token identity provider which completely skips this code path. However, it should be a user-friendly error for services that don't support access tokens.

Fixed in 3ab8cfac3748f7e42e1707abe80e5425ffa0ba95

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
```diff
@@ -234,12 +238,15 @@ mod credential_process {
 
 const PROVIDER_NAME: &str = "ProfileFile";
 
-fn base_provider(profile: &Profile) -> Result<BaseProvider<'_>, ProfileFileError> {
+fn base_provider<'a>(
+    profile_set: &'a ProfileSet,
+    profile: &'a Profile,
+) -> Result<BaseProvider<'a>, ProfileFileError> {
```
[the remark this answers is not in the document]
jdisanti: Seems like that's the entire point of BaseProvider: putting all the relevant metadata in one place.

aws/rust-runtime/aws-config/src/profile/parser/normalize.rs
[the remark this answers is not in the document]
jdisanti: I made all the sso-session related code private to the crate, so this is a decision we can punt on now.

aws/rust-runtime/aws-config/src/profile/parser.rs
jdisanti: The parser is already hardcoded to understand profiles, so I don't see this as that big of a deal. If we were to start over again, a generic approach would probably be better. But given we need to maintain backwards compatibility, I think this is fine for now. I made the sso-session code private so that this can be a two-way door.

aws/rust-runtime/aws-config/src/profile/mod.rs
jdisanti: Fixed in 3ab8cfac3748f7e42e1707abe80e5425ffa0ba95

aws/rust-runtime/aws-config/src/sso/token.rs
jdisanti: Fixed in 3ab8cfac3748f7e42e1707abe80e5425ffa0ba95

aws/rust-runtime/aws-config/test-data/profile-parser-tests.json
jdisanti: Added in 3ab8cfac3748f7e42e1707abe80e5425ffa0ba95

CHANGELOG.next.toml
jdisanti: Opted not to include an example since the recommended approach for SSO is to use the AWS CLI to configure it (via `aws configure sso` and `aws sso login`). Added more information about the lack of Builder ID support.

aws/rust-runtime/aws-config/src/profile/parser.rs
rcoh: I guess I see the semantics a little different. Profiles being the fundamental unit because you're always reading the file in the "scope" of a given profile + additional global metadata accessible to all profiles.

But as you said, two way door, I'm fine with this.

rcoh (approved):

PR #3383, opened 2024-01-24
Add CI step to release workflow to check for semver hazards

jdisanti:
This CI step will check for subtle semver hazards with the `ConfigBag` during release. Once all the runtime crates are independent, then this check can be moved into normal CI.

----

### DIFF
--- a/tools/ci-scripts/check-semver-hazards
+++ b/tools/ci-scripts/check-semver-hazards
@@ -1,0 +1,41 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# This script patches the new runtime crates into an old AWS SDK and runs tests
+# to check for semver hazards, such as a `Storable` being in an unstable runtime crate.
+
+C_YELLOW='\033[1;33m'
+C_RESET='\033[0m'
+
+if [ "$#" -ne 2 ]; then
+  echo "Usage: check-semver-hazards <stable-crate-version> <unstable-crate-version>"
+  exit 1
+fi
+STABLE_CRATE_VERSION="$1"
+UNSTABLE_CRATE_VERSION="$2"
+echo "Stable crate version: ${STABLE_CRATE_VERSION}"
+echo "Unstable crate version: ${UNSTABLE_CRATE_VERSION}"
+
+# Need to allow warnings since there may be deprecations that the old SDK uses
+unset RUSTFLAGS
+
+set -eux
+
+echo -e "${C_YELLOW}# Patching SDK...${C_RESET}"
+runtime-versioner patch-runtime \
+  --sdk-path "$(pwd)/aws-sdk-rust" \
+  --smithy-rs-path "$(pwd)/smithy-rs" \
+  --stable-crate-version "${STABLE_CRATE_VERSION}" \
+  --unstable-crate-version "${UNSTABLE_CRATE_VERSION}"
+
+# Testing just a small subset of the full SDK to check for semver hazards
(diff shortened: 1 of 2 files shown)


jdisanti:
Ran a dry-run against this branch to confirm the new check works: https://github.com/smithy-lang/smithy-rs/actions/runs/7645480204/job/20832185070

tools/ci-scripts/check-semver-hazards
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# This script patches the new runtime crates into an old AWS SDK and runs tests
+# to check for semver hazards, such as a `Storable` being in an unstable runtime crate.
```
rcoh, line 8: should note how to set this up (notably, $PWD/aws-sdk-rust, and smithy-rs need to exist)

tools/ci-scripts/check-semver-hazards
```diff
@@ -0,0 +1,41 @@
...
+  --unstable-crate-version "${UNSTABLE_CRATE_VERSION}"
+
+# Testing just a small subset of the full SDK to check for semver hazards
+echo -e "${C_YELLOW}# Testing SDK...${C_RESET}"
+for sdk in dynamodb s3; do
+  echo -e "${C_YELLOW}# Testing ${sdk}...${C_RESET}"
+  pushd "aws-sdk-rust/sdk/${sdk}" &>/dev/null
+  cargo test --all-features
+  popd &>/dev/null
+done
```
rcoh, line 41: should we check a few more?

rcoh (approved):
have you tried creating an intentional semver break and then running release in preview mode?

jdisanti:
> have you tried creating an intentional semver break and then running release in preview mode?

I verified that locally by invoking the script manually. I also confirmed the script is working correctly in GitHub Actions in https://github.com/smithy-lang/smithy-rs/actions/runs/7645370946/job/20831801559, where you can see it correctly picked up the patched crate version and then failed to compile due to deprecations and `-D warnings` (the latter of which I removed in a later revision of this PR):
```
   Compiling aws-sdk-s3 v1.13.0 (/home/build/workspace/aws-sdk-rust/sdk/s3)
error: use of deprecated type alias `aws_http::user_agent::AwsUserAgent`: Use aws_runtime::user_agent::AwsUserAgent instead.
 --> sdk/s3/tests/request_information_headers.rs:6:27
  |
6 | use aws_http::user_agent::AwsUserAgent;
  |                           ^^^^^^^^^^^^
  |
  = note: `-D deprecated` implied by `-D warnings`

error: use of deprecated type alias `aws_http::user_agent::AwsUserAgent`: Use aws_runtime::user_agent::AwsUserAgent instead.
  --> sdk/s3/tests/request_information_headers.rs:50:29
   |
50 |             layer.store_put(AwsUserAgent::for_tests());
   |                             ^^^^^^^^^^^^

error: use of deprecated constant `aws_http::content_encoding::header_value::AWS_CHUNKED`: Use aws_runtime::content_encoding::header_value::AWS_CHUNKED instead.
   --> sdk/s3/tests/checksums.rs:229:76
    |
229 |         HeaderValue::from_static(aws_http::content_encoding::header_value::AWS_CHUNKED),
    |                                                                            ^^^^^^^^^^^
    |
    = note: `-D deprecated` implied by `-D warnings`
```

tools/ci-scripts/check-semver-hazards
jdisanti: I don't think cargo check will do a whole lot for us here since these breaks happen at runtime, unless you're thinking of having it detect backwards compatibility issues. These two services should compile enough runtime crates that I think it would catch most of those.

PR #3385, opened 2024-01-24
Make it possible to customize presigned requests

rcoh:
## Motivation and Context
This enables the following code:
```rust
    let req = client
        .get_object()
        .bucket("foo")
        .key("bar")
        .customize()
        .mutate_request(|req| {
            req.set_uri(req.uri().to_string() + "&a=b")
                .expect("failed to update URI")
        })
        .presigned(static_ps_config)
        .await
```
Previously, it wasn't possible to presign a request once customization had begun.

## Description
- Follow the same pattern as send by introducing an additional trait, CustomizeablePresign
- change InvokeWithStopPoint to actually _return_ the error if there is one.

## Testing
New integration test

## Checklist

- [x] Conditionally generate the presign method

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
@@ -157,15 +157,7 @@ class AwsPresigningDecorator internal constructor(
                         E: std::error::Error + #{Send} + #{Sync} + 'static,
                         B: #{CustomizablePresigned}<E>
                     {
-                        let mut config_override = self.config_override.unwrap_or_default();
-                        self.interceptors.into_iter().for_each(|interceptor| {
-                            config_override.push_interceptor(interceptor);
-                        });
-                        self.runtime_plugins.into_iter().for_each(|plugin| {
-                            config_override.push_runtime_plugin(plugin);
-                        });
-
-                        self.customizable_send.presign(config_override, presigning_config).await
+                        self.execute(move |sender, conf|sender.presign(conf, presigning_config)).await
                     }
                     """,
                     *codegenScope,
@@ -178,7 +170,7 @@ class AwsPresigningDecorator internal constructor(
         RuntimeType.forInlineFun("CustomizablePresigned", InternalTraitsModule) {
             rustTemplate(
                 """
-                pub trait CustomizablePresigned<E> {
+                pub trait CustomizablePresigned<E>: #{Send} + #{Sync} {
                     fn presign(self, config_override: crate::config::Builder, presigning_config: #{PresigningConfig}) -> BoxFuture<SendResult<#{PresignedRequest}, E>>;
                 }
 
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
@@ -8,6 +8,7 @@ package software.amazon.smithy.rustsdk
 import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.knowledge.HttpBinding
 import software.amazon.smithy.model.knowledge.HttpBindingIndex
+import software.amazon.smithy.model.knowledge.TopDownIndex
 import software.amazon.smithy.model.shapes.MemberShape
 import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.model.shapes.ServiceShape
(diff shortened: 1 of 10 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
```diff
@@ -128,21 +141,64 @@ class AwsPresigningDecorator internal constructor(
...
+                        E: std::error::Error + #{Send} + #{Sync} + 'static,
+                        B: #{CustomizablePresigned}<E>
+                    {
+                        let mut config_override = self.config_override.unwrap_or_default();
+                        self.interceptors.into_iter().for_each(|interceptor| {
+                            config_override.push_interceptor(interceptor);
+                        });
+                        self.runtime_plugins.into_iter().for_each(|plugin| {
+                            config_override.push_runtime_plugin(plugin);
+                        });
```
jdisanti, line 166: We should create a private function to share this logic between send/presign.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
```diff
@@ -128,21 +141,64 @@ class AwsPresigningDecorator internal constructor(
...
+            .any { presignableOperations.containsKey(it.id) }
+
+    override fun extraSections(codegenContext: ClientCodegenContext): List<AdHocCustomization> =
+        anyPresignedShapes(codegenContext).thenSingletonListOf {
+            adhocCustomization<CustomizableOperationSection.CustomizableOperationImpl> {
+                rustTemplate(
+                    """
+                    /// Sends the request and returns the response.
+                    ##[allow(unused_mut)]
+                    pub async fn presigned(mut self, presigning_config: #{PresigningConfig}) -> #{Result}<#{PresignedRequest}, crate::error::SdkError<E>> where
```
jdisanti, line 156: Can you add a test to verify the future returned from this function implements `Send`? Probably a good idea to add one to `send` if there isn't already as well.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
```diff
@@ -128,21 +141,64 @@ class AwsPresigningDecorator internal constructor(
...
+                    "CustomizablePresigned" to CustomizablePresigned,
+                )
+            }
+        }
+
+    private val CustomizablePresigned =
+        RuntimeType.forInlineFun("CustomizablePresigned", InternalTraitsModule) {
+            rustTemplate(
+                """
+                pub trait CustomizablePresigned<E> {
```
jdisanti, line 181: How come `CustomizableSend` has `Send + Sync` bounds, but this one doesn't?

aws/sdk/integration-tests/s3/tests/presigning.rs
```diff
@@ -195,3 +195,32 @@ async fn test_presigned_head_object() {
         presigned.uri().to_string(),
     );
 }
+
+#[tokio::test]
+async fn customized_presigning() {
```
jdisanti, line 200: It would be a good idea to add a test to Polly as well since it does request mutations in its presigning.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/CustomizableOperationDecorator.kt
```diff
@@ -5,12 +5,9 @@
 
 package software.amazon.smithy.rust.codegen.client.smithy.generators.client
 
-import software.amazon.smithy.rust.codegen.core.smithy.customize.NamedCustomization
-import software.amazon.smithy.rust.codegen.core.smithy.customize.Section
+import software.amazon.smithy.rust.codegen.core.smithy.customize.AdHocSection
 
-sealed class CustomizableOperationSection(name: String) : Section(name) {
+sealed class CustomizableOperationSection(name: String) : AdHocSection(name) {
```
jdisanti, line 10: You changed this to adhoc, but didn't actually make any calls to adhocCustomization anywhere. Should we just remove this entirely? I could also be misunderstanding how these work again... they're not very clear.

rcoh: there is a call to adhocCustomizations here:
https://github.com/smithy-lang/smithy-rs/blob/627a1a4cb76cdf4b955535c0fe03bf7e329143bb/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt#L151

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt
rcoh: no good reason—they aren't actually needed because we never make a trait object from it. I added them here for consistency.

jdisanti (approved):

PR #3386, opened 2024-01-25
Add placeholder types for S3 Express and enable control flow to be redirected for S3 Express use case

ysaito1001:
## Motivation and Context
This PR is the first in the series to support the S3 Express feature in the Rust SDK. The work will be done in the feature branch, and once it is code complete, the branch will be merged to main.

## Description
This PR adds placeholder types for S3 Express and enables control flow to be redirected for S3 Express use case. For instance, if we run the following example code against a generated SDK from this PR:
```rust
let shared_config = aws_config::from_env().region(aws_sdk_s3::config::Region::new("us-east-1")).load().await;
let client = aws_sdk_s3::Client::new(&shared_config);
client.list_objects_v2().bucket("testbucket--use1-az4--x-s3";).send().await.unwrap();
```
it will end up
```
thread 's3_express' panicked at 'not yet implemented', /Users/awsaito/src/smithy-rs/aws/sdk/build/aws-sdk/sdk/s3/src/s3_express.rs:104:13
```
which points to
```
impl ProvideCredentials for DefaultS3ExpressIdentityProvider {
    fn provide_credentials<'a>(&'a self) -> aws_credential_types::provider::future::ProvideCredentials<'a>
    where
        Self: 'a,
    {
        todo!() <---
    }
}
```

### Implementation decisions
- `DefaultS3ExpressIdentityProvider` has an accompanying identity cache. That identity cache cannot be configured by customers so it makes sense for the provider itself to internally own it. In that case, we do NOT want to use the identity cache stored in `RuntimeComponents`, since it interferes with the S3 Express's caching policy. To that end, I added an enum `CacheLocation` to `SharedIdentityResolver` (it already had the `cache_partition` field so it was kind of aware of caching).
  - Two reasons why `CacheLocation` is added to `SharedIdentityResolver`, but not to individual, concrete `IdentityResolver`s. One is `SharedIdentityResolver` was already cache aware, as mentioned above. The other is that it is more flexible that way; The cache location is not tied to a type of identity resolver, but we can select it when creating a `SharedIdentityResolver`.
  - I considered but did not add a field `cacheable` to `Identity` since I wanted to keep `Identity` as plain data, keeping the concept of "caching" somewhere outside. 
- I've added a separate `Config` method, `set_express_credentials_provider`, to override credentials provider for S3 Express. There are other SDKs (e.g. [Ruby](https://www.rubydoc.info/gems/aws-sdk-s3/Aws/S3/Client)) that follow this style and it makes it clear to the customers that this is the method to use when overriding the express credentials provider. The existing `set_credentials_provider`, given its input type, cannot tell whether a passed-in credentials provider is for a regular `sigv4` or for S3 Express.

## Testing
Only verified that control flow could be altered for an S3 Express use case, as shown above. Further testing will be added in subsequent PRs.

## Checklist
I am planning to include in `CHANGELOG.next.toml` a user guide for S3 Express once the feature branch `ysaito/s3express` is ready to be merged to main.

----


rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs
```diff
@@ -612,11 +612,23 @@ impl RuntimeComponentsBuilder {
...
         identity_resolver: impl ResolveIdentity + 'static,
+    ) -> &mut Self {
+        self.set_shared_identity_resolver(scheme_id, identity_resolver.into_shared())
+    }
+
+    /// Sets the identity resolver for a given `scheme_id`.
+    ///
+    /// This is effectively the same as [`Self::set_identity_resolver`] but takes a [`SharedIdentityResolver`]
+    /// instead.
+    pub fn set_shared_identity_resolver(
```
jdisanti, line 623: Can we use the `IntoShared` trait to modify the existing `set_identity_resolver` function in a backwards compatible manner?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
```diff
@@ -0,0 +1,210 @@
...
+                ).resolve("auth::SCHEME_ID"),
+        )
+
+    override fun section(section: ServiceConfig) =
+        writable {
+            when (section) {
+                ServiceConfig.BuilderImpl -> {
+                    rustTemplate(
+                        """
+                        /// Sets the credentials provider for S3 Express
```
jdisanti, line 169:
```suggestion
                        /// Sets the credentials provider for S3 Express One Zone
```
I think the name should be fully spelled out for the doc comments.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
```diff
@@ -0,0 +1,210 @@
...
+import software.amazon.smithy.rust.codegen.core.rustlang.rust
+import software.amazon.smithy.rust.codegen.core.rustlang.rustBlockTemplate
+import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.core.rustlang.writable
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType.Companion.preludeScope
+import software.amazon.smithy.rustsdk.AwsRuntimeType
+import software.amazon.smithy.rustsdk.InlineAwsDependency
+
+class S3ExpressDecorator : ClientCodegenDecorator {
```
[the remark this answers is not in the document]
ysaito1001: Yeah, consistency is important. The other SDKs call it just `S3Express` (e.g. [Java v2](https://github.com/search?q=repo%3Aaws%2Faws-sdk-java-v2%20S3Express&type=code)), so we should be good.

rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs
ysaito1001: Thanks for pointing it out! Turns out that this method is unnecessary because `set_identity_resolver` can already take a `SharedIdentityResolver` and `.into_shared()` is smart enough to not wrap a `SharedIdentityResolver` with another `SharedIdentityResolver` (as tested [here](https://github.com/smithy-lang/smithy-rs/blob/8873666b0ee0cf6dada9599778c96c77580f9042/rust-runtime/aws-smithy-runtime-api/src/shared.rs#L227-L231)).

Fixed in ef1e1be.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/// Supporting code for S3 Express auth
+pub mod auth {
```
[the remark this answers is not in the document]
ysaito1001: Updated in 38ce80f.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
```diff
@@ -0,0 +1,210 @@
...
+                            self.set_express_credentials_provider(#{Some}(#{SharedCredentialsProvider}::new(credentials_provider)));
+                            self
+                        }
+                        """,
+                        *codegenScope,
+                    )
+
+                    rustBlockTemplate(
+                        """
+                        /// Sets the credentials provider for S3 Express
```
jdisanti, line 180:
```suggestion
                        /// Sets the credentials provider for S3 Express One Zone
```

rust-runtime/aws-smithy-runtime-api/src/client/identity.rs
```diff
@@ -161,19 +161,45 @@ pub trait ResolveIdentity: Send + Sync + Debug {
...
+    /// Indicates the identity cache is internally managed by the identity resolver.
+    IdentityResolver,
+}
+
 /// Container for a shared identity resolver.
 #[derive(Clone, Debug)]
 pub struct SharedIdentityResolver {
     inner: Arc<dyn ResolveIdentity>,
     cache_partition: IdentityCachePartition,
+    cache_location: IdentityCacheLocation,
```
jdisanti, line 184: Shouldn't this defer to the wrapped `inner` identity resolver trait impl?

ysaito1001: Just wanna make sure you read the first white sub-bullet under `Implementation decisions` that says `Two reasons why CacheLocation is added to SharedIdentityResolver...` and then still have this question?

ysaito1001: Addressed in c2a2708

rust-runtime/aws-smithy-runtime-api/src/client/identity.rs
```diff
@@ -159,6 +159,14 @@ pub trait ResolveIdentity: Send + Sync + Debug {
     fn fallback_on_interrupt(&self) -> Option<Identity> {
         None
     }
+
+    /// Returns the location of an identity cache associated with this identity resolver.
+    ///
+    /// By default, identity resolvers will use the identity cache stored in runtime components.
+    /// Implementing types can change the cache location if they want to.
```
jdisanti, line 166: Would be good to state why an implementation might want to change the location.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
```diff
@@ -0,0 +1,195 @@
...
+                        ) {
+                            rustTemplate(
+                                """
+                                self.runtime_components.set_identity_resolver(#{S3_EXPRESS_SCHEME_ID}, credentials_provider);
+                                """,
+                                *codegenScope,
+                            )
+                        }
+                        rust("self")
+                    }
```
jdisanti, line 189: Nit: this style is less readable than a single large block quote.

jdisanti (approved):

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
ysaito1001: Addressed in f569231.

rust-runtime/aws-smithy-runtime-api/src/client/identity.rs
ysaito1001: Added an extra sentence in the comment (457f6b8).

PR #3387, opened 2024-01-26
Only create inner chain provider once in profile creds provider

jdisanti:
The `SsoCredentialsProvider` maintains an in-memory expiring cache of SSO tokens, and this cache is maintained within its instance. The `ProfileFileCredentialsProvider`, which uses `SsoCredentialsProvider` as a base/inner provider, is currently reconstructing the inner provider every time it loads credentials, which discards the SSO token cache entirely.

This PR refactors the `ProfileFileCredentialsProvider` to cache the inner provider so that it is only initialized once on first load of credentials. This is done via a new `ErrorTakingOnceCell` abstraction since the same init-caching mechanism will be needed for the `ProfileFileTokenProvider` when it is implemented.

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -28,3 +28,9 @@ message = "Add `try_into_http1x` and `try_from_http1x` to Request and Response c
 references = ["aws-sdk-rust#977", "smithy-rs#3365", "smithy-rs#3373"]
 meta = { "breaking" = false, "bug" = false, "tada" = false }
 author = "rcoh"
+
+[[aws-sdk-rust]]
+message = "Retain the SSO token cache between calls to `provide_credentials` when using IAM Identity Center SSO via the AWS config file."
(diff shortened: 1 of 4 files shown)


rcoh (commented):
changes themselves seem fine. I'm a little worried that with no test this behavior could regress.

Maybe we could construct a fake environment, mutate the file system, then requery credentials and validate that they change (or don't?)

jdisanti:
> Maybe we could construct a fake environment, mutate the file system, then requery credentials and validate that they change (or don't?)

Or, perhaps easier, we can register a named source in `NamedProviderFactory` that verifies the source is created exactly once.

jdisanti:
Actually looks like the way I proposed won't work because the factory holds arcs of already instantiated `ProvideCredentials` implementations, and SSO is separate from it entirely.

rcoh (approved):

PR #3388, opened 2024-01-29
Allow list-objects-v2 to run against an S3 Express bucket

ysaito1001:
## Motivation and Context
Adds an implementation spike to allow `list-objects-v2` (possibly others, haven't tested yet) to run against an S3 Express bucket.

## Description
This PR implements two ingredients, `S3ExpressIdentityProvider` and `S3ExpressSigner`. `S3ExpressIdentityProvider` uses an internal S3 client to obtain an S3 Express session token that is passed to `S3ExpressSigner`. `S3ExpressSigner` then signs a request with that token, using effectively sigv4 but with session token omitted and an extra header added instead, `x-amz-s3session-token`.

In addition, this PR supports presigning for S3 Express. Similarly to signing headers, presigning for S3 Express excludes a query param `X-Amz-Security-Token` and instead uses `X-Amz-S3session-Token` for the signing query params. The following screeshot shows that a presigned URL from `get_object` works for an S3 Express bucket:
<p align="center">
chain-provider-ext-timeout-2
</p>

Some implementation details:
- Since `S3ExpressIdentityProvider` passes an S3 Express bucket name for S3's `create_session` API to obtain an S3 Express session token, it needs to obtain the bucket name from somewhere. `S3ExpressIdentityProvider::ProvideCredentials` I put previously did not have enough arguments for us to figure this out, so I switched to `S3ExpressIdentityProvider::ResolveIdentity` that takes enough arguments.
- `SigV4Signer::sign_http_request` did not allow calling code to pass a configured `SigningSettings`; The signer needs to exclude a header `x-amz-security-token` and include `x-amz-s3session-token`. To make this happen, I made `sigv4::extract_operation_config` and `sigv4::settings` public APIs (previously private).
- One area I haven't quite figured out yet is how to configure the inner S3 client to call `create_session`. The changes in this PR inherits runtime components & config bag from the "outer" S3 client, but customers may want to configure the inner S3 client in a more flexible manner (e.g. operation timeout).

## Testing
To lock the behavior at this time, I added a connection recording test for `list-objects-v2`.

----


aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -91,10 +139,143 @@ pub(crate) mod identity_provider {
...
+                    .map(|ty| ty.0),
+            );
+
+            // inherits all runtime components from a current S3 operation but clears out
+            // out interceptors configured for that operation
+            let mut builder = runtime_components.to_builder();
+            builder.set_interceptors(std::iter::empty::<SharedInterceptor>());
+            config_builder.runtime_components = builder;
+
+            let client = crate::Client::from_conf(config_builder.build());
```
ysaito1001, line 267: Should create `client` only once during construction of `DefaultS3ExpressIdentityProvider` and make use of the `config_builder` variable through `.customize().config_override(config_builder)`.

ysaito1001: Oh I remember now. If `client` is created during construction of `DefaultS3ExpressIdentityProvider`, it goes into an infinite loop via repeated calls to `Client::from_conf`.

`DefaultS3ExpressIdentityProvider` is created in `ServiceRuntimePlugin` which is already in the middle of running `Client::from_conf`, so if `DefaultS3ExpressIdentityProvider` internally creates an S3 `client` via `Client::from_conf` this will be an infinite loop, causing stack overflow.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -91,10 +139,143 @@ pub(crate) mod identity_provider {
...
+        }
+
+        fn bucket_name<'a>(&'a self, config_bag: &'a ConfigBag) -> &'a str {
+            let params = config_bag
+                .load::<EndpointResolverParams>()
+                .expect("endpoint resolver params must be set");
+            let params = params
+                .get::<crate::config::endpoint::Params>()
+                .expect("`Params` should be wrapped in `EndpointResolverParams`");
+            params.bucket().expect("bucket name must be set")
```
jdisanti, line 198: Should this error instead of panic? Is it possible to reach this by not setting `bucket_name` on the fluent builder?

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -91,10 +139,143 @@ pub(crate) mod identity_provider {
...
+        }
+
+        async fn express_session_credentials<'a>(
+            &'a self,
+            bucket_name: &'a str,
+            runtime_components: &'a RuntimeComponents,
+            config_bag: &'a ConfigBag,
+        ) -> Result<SessionCredentials, BoxError> {
+            let mut config_builder = crate::Config::builder();
```
jdisanti, line 207: This seems a bit fragile. Should we generate a `pub(crate) fn from(config_bag: &ConfigBag)` for `Config` so that the copying of properties can be left to the config code generator?

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
```diff
@@ -141,28 +143,23 @@ impl SigV4Signer {
...
     }
-}
 
-impl Sign for SigV4Signer {
-    fn sign_http_request(
+    /// Signs the given `request`.
+    ///
+    /// This is a helper used by [`Sign::sign_http_request`] and will be useful if calling code
+    /// needs to pass a configured `settings`.
+    pub fn sign_http_request(
```
jdisanti, line 151: We should think about this function signature some more. A lot of customers want to do manual signing for use cases such as API Gateway invocations, and having a function that is easy to sign with would help a lot for that. If this requires `RuntimeComponents` or `ConfigBag`, it becomes more difficult to use directly.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
```diff
@@ -53,6 +53,7 @@ class S3Decorator : ClientCodegenDecorator {
     private val invalidXmlRootAllowList =
         setOf(
             // API returns GetObjectAttributes_Response_ instead of Output
+            ShapeId.from("com.amazonaws.s3#CreateSessionOutput"),
             ShapeId.from("com.amazonaws.s3#GetObjectAttributesOutput"),
```
jdisanti, line 57: Update comment

rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs
```diff
@@ -377,12 +377,37 @@ declare_runtime_components! {
     }
 }
 
+impl From<RuntimeComponents> for RuntimeComponentsBuilder {
+    fn from(rc: RuntimeComponents) -> Self {
+        RuntimeComponentsBuilder {
+            builder_name: "converted from RuntimeComponents",
```
jdisanti, line 383: Perhaps we should have a separate `from_components` function that takes a builder name rather than implementing the `From` trait.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
ysaito1001: I think it's possible depending on future requirements. Addressed in c4df9c7.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
ysaito1001: Addressed in e47157f

rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs
ysaito1001: Updated in 5a4e3b2.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
ysaito1001: Addressed it in 51518f9.

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
ysaito1001: Added `TODO` based on our offline discussion (573ef36). Will circle back!

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -56,15 +62,109 @@ pub(crate) mod auth {
...
     }
+
+    fn add_token_to_request(
+        express_credentials: &Credentials,
+        request: &mut HttpRequest,
+        settings: &mut SigningSettings,
+    ) -> Result<(), BoxError> {
+        match settings.signature_location {
+            SignatureLocation::Headers => {
+                let security_token_header = Cow::Borrowed("x-amz-security-token");
```
jdisanti, line 99: The STS session token must be excluded from signing, but doesn't it also need to be excluded from the request entirely? Shouldn't we be removing it from the headers instead of excluding it from signing here?

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -56,15 +62,109 @@ pub(crate) mod auth {
...
+                )
+                .unwrap();
+                value.set_sensitive(true);
+                request.headers_mut().insert(
+                    http::HeaderName::from_static("x-amz-s3session-token"),
+                    value,
+                );
+            }
+            SignatureLocation::QueryParams => {
+                let security_token_param = Cow::Borrowed("X-Amz-Security-Token");
```
jdisanti, line 123: Shouldn't this param be removed entirely?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/RegionDecorator.kt
```diff
@@ -211,6 +211,10 @@ class RegionProviderConfig(codegenContext: ClientCodegenContext) : ConfigCustomi
                     )
                 }
 
+                is ServiceConfig.BuilderFromConfigBag -> {
```
jdisanti, line 214: Side-note: it would be nice to have the type system ensure we implement all these config injections for each field that gets added to config. Seems like it will be easy to forget one in the future.

aws/sdk/integration-tests/s3/tests/express.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    let result = client
+        .list_objects_v2()
+        .bucket("s3express-test-bucket--usw2-az1--x-s3")
+        .send()
+        .await;
+    dbg!(result).expect("success");
+
+    http_client
+        .validate_body_and_headers(Some(&["x-amz-s3session-token"]), "application/xml")
```
jdisanti, line 43: Why exclude x-amz-s3session-token from validation? Isn't it the exact value that's returned from CreateSession?

rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs
```diff
@@ -380,12 +380,40 @@ declare_runtime_components! {
...
```
jdisanti, line 386: This one should be removed in favor of `from_runtime_components`

jdisanti (commented):
I got a little confused on the excluded headers/params special casing, as you'll find in my comments below. We should put some more though into how this exclusion is done.

Otherwise, just minor comments.

rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs
ysaito1001: Removed in 027c2cd.

aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
```diff
@@ -296,10 +296,15 @@ fn calculate_signing_params<'a>(
...
-        ));
+        if !params
+            .settings()
+            .param_excluded(param::X_AMZ_SECURITY_TOKEN)
+        {
+            signing_params.push((
+                param::X_AMZ_SECURITY_TOKEN,
+                Cow::Owned(security_token.to_string()),
+            ));
+        }
```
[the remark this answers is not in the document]
ysaito1001: Thanks for pointing that out! After several iterations, I was able to come up with the simplest approach that does what needs to be done, without contradicting existing code in `SigningSettings`. Addressed 830ef02.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
ysaito1001: Addressed 830ef02.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
ysaito1001: Addressed 830ef02.

aws/sdk/integration-tests/s3/tests/express.rs
ysaito1001: [validate_body_and_headers](https://github.com/smithy-lang/smithy-rs/blob/fd8aba4abe0f54caa91255fafdbfe7c9d7be4122/rust-runtime/aws-smithy-runtime/src/client/http/test_util/dvr/replay.rs#L97) does validate the passed-in headers, so it is included 🙂

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/RegionDecorator.kt
ysaito1001: Yeah, that's important to ensure. Given the scope, added a tracking issue in 97cecd3 to be worked on separately.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -95,10 +129,93 @@ pub(crate) mod identity_provider {
...
+            runtime_components: &'a RuntimeComponents,
+            config_bag: &'a ConfigBag,
+        ) -> Result<SessionCredentials, BoxError> {
+            let mut config_builder = crate::config::Builder::from_config_bag(config_bag);
+
+            // inherits all runtime components from a current S3 operation but clears out
+            // out interceptors configured for that operation
+            let mut rc_builder = runtime_components.to_builder();
+            rc_builder.set_interceptors(std::iter::empty::<SharedInterceptor>());
+            config_builder.runtime_components = rc_builder;
```
jdisanti, line 205: I think this will clear out some standard interceptors that we probably still want, such as the user agent interceptor. I understand needing to clear out any user-provided interceptors though... We should think about this some more.

jdisanti (commented):
Looks good. Just one concern.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
ysaito1001: If I'm not mistaken, the standard (i.e. non-user-provided interceptors) interceptors will be re-added by `ServiceRuntimePlugin` at line 207 during `from_conf` -> `base_client_runtime_plugins` -> `ServiceRuntimePlugin::new`, including `UserAgentInterceptor`?

jdisanti: Oh, right, because you create a new operation with the components rather than using them directly. That makes sense.

jdisanti (approved):
Nice work! 🚀

PR #3393, opened 2024-02-06
Allow S3 Express to use CRC32 for default request checksum

ysaito1001:
## Motivation and Context
Changes the default request checksum to CRC32 for S3 Express.

## Description
This PR modifies the default request checksum to CRC32 for S3 Express. Note that if a customer specifies a custom checksum algorithm via such as [checksum_algorithm](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/operation/put_object/struct.PutObjectInput.html#structfield.checksum_algorithm), then the PR does not affect anything and respects the customer-specified checksum algorithm.

Otherwise, if an operation is annotated with a `requestChecksumRequired` trait and set to true, the default checksum becomes relevant. The default algorithm is currently `MD5` and S3 Express requires it to be `CRC32`. The code changes in this PR reflect that requirement.

Implementation details include
- `RequestChecksumInterceptor` is a single-sourced place to handle request checksum, but since it is a common place for all services we cannot add S3-Express-specific logic in there to configure the default checksum. Instead, we have introduced a slight more abstract concept of overriding the default checksum via `DefaultRequestChecksumOverride`. `RequestChecksumInterceptor` grabs it from a `ConfigBag` and uses it (if found) to configure the default checksum.
- We deferred a point of adding a checksum to a request body to `RequestChecksumInterceptor::modify_before_signing` from `RequestChecksumInterceptor::modify_before_retry_loop`. This is because we won't know whether S3 Express checksum default actually kicks until an endpoint is fully resolved. `modify_before_retry_loop` occurs before `client::orchestrator::endpoints::orchestrate_endpoint` so we need to choose a different interceptor method that runs after it.
 
## Testing
Added integration tests for S3 Express default checksum to `integration-tests/s3/tests/express.rs`.

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
```diff
@@ -198,6 +212,67 @@ class S3ExpressIdentityProviderConfig(codegenContext: ClientCodegenContext) : Co
...
+                                    return original;
+                                }
+
+                                let endpoint = cfg
+                                    .load::<crate::config::endpoint::Endpoint>()
+                                    .expect("endpoint added to config bag by endpoint orchestrator");
+
+                                match endpoint.properties().get("backend") {
+                                    Some(#{Document}::String(backend)) if backend.as_str() == "S3Express" => {
+                                       #{customDefault:W}
```
jdisanti, line 252: If you decide to fork `RequestChecksumInterceptor` for S3, then you can just look up the request `Metadata` in the config bag to get the operation name. Or, alternatively, you can have the operation runtime plugin for that operation put the override value into the config bag.

aws/sdk/integration-tests/s3/tests/express.rs
```diff
@@ -98,3 +95,148 @@ async fn list_objects_v2_in_both_express_and_regular_buckets() {
...
+    let crc32_checksum = "AAAAAA==";
+    let crc32c_checksum = "AAAAAA==";
+    let sha1_checksum = "2jmj7l5rSw0yVb/vlWAYkK/YBwk=";
+    let sha256_checksum = "47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=";
+    for (checksum, value) in ChecksumAlgorithm::values().iter().zip(&[
+        crc32_checksum,
+        crc32c_checksum,
+        sha1_checksum,
+        sha256_checksum,
+    ]) {
```
jdisanti, line 185: This seems like a fragile way to pair these values. Would be better to do something like:
```rust
let checksums = &[
    (ChecksumAlgorithm::Crc32, "AAAAAA=="),
    (ChecksumAlgorithm::Crc32c, "AAAAAA=="),
    // ...
];
assert_eq!(ChecksumAlgorithm::values().len(), checksums.len(), "checksum enum changed; need to update test");
for (checksum, value) in checksums {
    // ...
}
```
Although, that assert may be a bad idea since it could hold up a release upon adding a new algorithm? Not sure on that one since a new algorithm would require new code on our end anyway.

jdisanti (commented):
This approach will work, but it may be simpler to create a separate interceptor just for S3 in this case.

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
```diff
@@ -60,6 +60,45 @@ impl Storable for RequestChecksumInterceptorState {
...
 }
 
+pub(crate) type CustomDefaultFn = Box<
+    dyn Fn(Option<ChecksumAlgorithm>, &ConfigBag) -> Option<ChecksumAlgorithm>
+        + Send
+        + Sync
+        + 'static,
+>;
+
+pub(crate) struct DefaultRequestChecksumOverride {
```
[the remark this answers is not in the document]
ysaito1001: I partially implemented that approach but abandoned it because I suspected that it was more complicated. The challenge is that if we have an S3-specific `RequestChecksumInterceptor`, then it needs to either
- disable [the general RequestChecksumInterceptor](https://github.com/smithy-lang/smithy-rs/blob/f7be5f801504e7edf5cb6b731f3400c37b298abc/aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs#L79-L100) entirely
- only change the checksum default and delegate the rest to the general `RequestChecksumInterceptor`.

For the former, all the logic in the general `RequestChecksumInterceptor` needs to be taken over by the S3-specific `RequestChecksumInterceptor`, but the general one must be disabled in runtime components, otherwise a request checksum related header will incorrectly be added twice (plus there is a problem of code duplication between the general `RequestChecksumInterceptor` and S3-specifc one).

For the latter, we need to have a decorating interceptor so the S3-specific `RequestChecksumInterceptor` wraps the current `RequestChecksumInterceptor`. But then again, we must disable the general `RequestChecksumInterceptor` for the same reason mentioned above.

To disable the general `RequestChecksumInterceptor` using [disable_interceptor](https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-runtime-api/src/client/interceptors.rs#L839), we need to remove a generic parameter `AP` from `RequestChecksumInterceptor` because we cannot enumerate all possible `AP` when specifying it in `disable_interceptor`.

Let me know if I'm overthinking and you have something simpler.

ysaito1001:
Will re-create a new PR due to many merge conflicts at this point. Will respond to review comments in the new PR.

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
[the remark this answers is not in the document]
jdisanti: I see. I neglected to consider the complex interplay between the Smithy/AWS codegen pieces. That makes sense.

aws/sdk/integration-tests/s3/tests/express.rs
ysaito1001: Yeah, I prefer that. I would probably remove the assert as you suspected, since if the service adds a new checksum algorithm, it could fail in our release pipeline. It's safer for us to react to it and update the code rather than the change is pushed upon us during a release.

PR #3394, opened 2024-02-07
Upgrade Smithy Gradle Plugin to 0.9

jdisanti:

### DIFF
--- a/aws/sdk-adhoc-test/build.gradle.kts
+++ b/aws/sdk-adhoc-test/build.gradle.kts
@@ -9,7 +9,9 @@ extra["moduleName"] = "software.amazon.smithy.rust.awssdk.adhoc.test"
 tasks["jar"].enabled = false
 
 plugins {
-    id("software.amazon.smithy")
+    java
+    id("software.amazon.smithy.gradle.smithy-base")
+    id("software.amazon.smithy.gradle.smithy-jar")
 }
 
 java {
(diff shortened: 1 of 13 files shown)


aws/sdk-adhoc-test/build.gradle.kts
```diff
@@ -79,7 +74,7 @@ project.registerGenerateSmithyBuildTask(rootProject, pluginName, allCodegenTests
 project.registerGenerateCargoWorkspaceTask(rootProject, pluginName, allCodegenTests, workingDirUnderBuildDir)
 project.registerGenerateCargoConfigTomlTask(layout.buildDirectory.dir(workingDirUnderBuildDir).get().asFile)
 
-tasks["smithyBuildJar"].dependsOn("generateSmithyBuild")
+tasks["jar"].dependsOn("generateSmithyBuild")
```
rcoh, line 77: do we have to name the task jar? is that the normal task name? it seems a little generic, but I guess it just means "the task of making the jar?"

build.gradle.kts
```diff
@@ -14,20 +14,16 @@ buildscript {
...
-}
 
-allprojects.forEach {
-    it.apply(plugin = "jacoco")
+    apply(plugin = "jacoco")
 
-    it.the<JacocoPluginExtension>().apply {
+    the<JacocoPluginExtension>().apply {
         toolVersion = "0.8.8"
         reportsDirectory.set(layout.buildDirectory.dir("jacoco-reports"))
```
rcoh, line 28: we don't actually use this, we should drop it

rcoh (approved):

aws/sdk-adhoc-test/build.gradle.kts
jdisanti: So my understanding is that the `jar` task name is what comes with the Gradle `java` plugin, and now Smithy's Gradle Plugin is hooking into that existing task instead of rolling its own.

PR #3407, opened 2024-02-13
Remove warning banner code

rcoh:
## Motivation and Context
Remove warning banner related code. The test is broken so this code is just causing a hassle and will never be used again.

- [ ] A generated code diff should be audited prior to merging.
----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCrateDocsDecorator.kt
```diff
@@ -126,7 +126,7 @@ internal class AwsCrateDocGenerator(private val codegenContext: ClientCodegenCon
             check(moduleVersion.isNotEmpty() && moduleVersion[0].isDigit())
 
             val moduleName = codegenContext.settings.moduleName
-            val stableVersion = !moduleVersion.startsWith("0.")
+            !moduleVersion.startsWith("0.")
```
Velfi, line 129: This line should be deleted too, or wrapped in an assert

Velfi (approved):

jdisanti (approved):

PR #2520, opened 2023-03-31
Feat/default wasi http

eduardomourar:
## Motivation and Context
https://github.com/awslabs/smithy-rs/issues/2087

This is a continuation of PR #2254, it will replace the mocked HTTP client with one based on the experimental HTTP support for WASI.

## Description
A default HTTP connector and a sleep for WASI will be automatically included in the AWS and Smithy relevant crates.

## Testing
We still need to wait for [this PR](https://github.com/bytecodealliance/wasmtime/pull/5929) to land in order to test using `cargo-wasi`.

## Checklist

- [ ] I have updated `CHANGELOG.next.toml` if I made changes to the smithy-rs codegen or runtime crates
- [ ] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3409, opened 2024-02-14
Add `aws-smithy-wasm` crate with WASI http client

landonxjames:
## Motivation and Context

This change adds a new crate, `aws-smithy-wasm`, that exports a SDK compatible WASI http client. This is a continuation of the work in #2520 using the now stabilized WASI 0.2.0 interfaces from the [wasi crate](https://crates.io/crates/wasi). This supports, but does not finalize the work for #2087 

## Description

Add a new crate, `aws-smithy-wasm` which exports a function `wasi_http_client` that will provide the user with a WASI compatible http client. This client is implemented by using the `wasi::http::outgoing_handler` [ref](https://docs.rs/wasi/0.12.0+wasi-0.2.0/wasi/http/outgoing_handler/index.html) along with some utility implementations of `TryFrom` to transform back and worth between the types from the `http` crate and the `wasi::http` types. It also exports a unit struct `WasmSleep` that impls the `AsyncSleep` trait needed by the SDK.

## Testing

This is tested via an integration test in `aws/sdk/integration-tests/webassembly` that uses the wasi http-client to vuild a config and an operation (that is not sent). It is further tested in a new canary (`wasm_canary`) that calls the S3 `list_objects_v2` API. 

## Checklist

----


aws/sdk/integration-tests/webassembly/src/list_objects.rs
```diff
@@ -0,0 +1,29 @@
...
+    use aws_sdk_s3::Client;
+
+    let shared_config = get_default_wasi_config().await;
+    let client = Client::new(&shared_config);
+    let operation = client
+        .list_objects_v2()
+        .bucket("nara-national-archives-catalog")
+        .delimiter("/")
+        .prefix("authority-records/organization/")
+        .max_keys(5);
```
jdisanti, line 19: Any tests that talk to real AWS services, ideally, would live in our canary. However, we don't quite have WASM set up there 😅 

We prefer to have traffic recorded tests in our integration tests.

Do you know if there's an easy way to run WASI in a Lambda?

rust-runtime/aws-smithy-wasm/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+  "AWS Rust SDK Team <aws-sdk-rust@amazon.com>",
+  "Eduardo Rodrigues <16357187+eduardomourar@users.noreply.github.com>",
+]
+description = "Smithy WebAssembly configuration for smithy-rs."
+edition = "2021"
+license = "Apache-2.0"
+repository = "https://github.com/awslabs/smithy-rs"
+
+[target.'cfg(all(target_family = "wasm", target_os = "wasi"))'.dependencies]
+# [dependencies]
```
jdisanti, line 14:
```suggestion
```

rust-runtime/aws-smithy-wasm/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "aws-smithy-wasm"
+version = "0.0.0-smithy-rs-head"
```
jdisanti, line 3: Our build/release system supports manual versioning of runtime crates now, and we'd like to move away from this auto-incrementing version string:
```suggestion
version = "0.1.0"
```

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+use bytes::{Bytes, BytesMut};
+use wasi::http::{
+    outgoing_handler,
+    types::{self as wasi_http, OutgoingBody, RequestOptions},
+};
+
+/// Creates a connector function that can be used during instantiation of the client SDK
+/// in order to route the HTTP requests through the WebAssembly host. The host must
+/// support the WASI HTTP proposal as defined in the Preview 2 specification.
+pub fn wasi_http_client() -> SharedHttpClient {
```
jdisanti, line 31: Will this client ever need any kind of configuration?

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+    shared::IntoShared,
+};
+use aws_smithy_types::body::SdkBody;
+use bytes::{Bytes, BytesMut};
+use wasi::http::{
+    outgoing_handler,
+    types::{self as wasi_http, OutgoingBody, RequestOptions},
+};
+
+/// Creates a connector function that can be used during instantiation of the client SDK
```
jdisanti, line 28:
```suggestion
/// Creates a HTTP client that can be used during instantiation of the client SDK
```

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/// HTTP client used in WASI environment
+#[derive(Debug, Clone)]
+pub struct WasiHttpClient {
+    connector: SharedHttpConnector,
+}
+
+impl WasiHttpClient {
+    /// Create a new Wasi HTTP client.
```
jdisanti, line 42:
```suggestion
    /// Create a new WASI HTTP client.
```

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+
+/// HTTP connector used in WASI environment
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct WasiHttpConnector;
+
+impl HttpConnector for WasiHttpConnector {
+    fn call(&self, request: HttpRequest) -> HttpConnectorFuture {
+        tracing::trace!("WasiHttpConnector: sending request {request:?}");
+        // TODO(wasi): add connect/read timeouts
```
jdisanti, line 74: Is this going to be a separate PR, and we'll just say don't use the WASI client in production yet?

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+
+            let sdk_res = Response::try_from(response)
+                .map_err(|err| ConnectorError::other(err.into(), None))?;
+
+            Ok(sdk_res)
+        })
+    }
+}
+
+/// Wasi http client containing the options to pass to the
```
jdisanti, line 103:
```suggestion
/// WASI HTTP client containing the options to pass to the
```

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+struct WasiDefaultClient {
+    options: WasiRequestOptions,
+}
+
+impl WasiDefaultClient {
+    /// Create a new Wasi HTTP client.
```
jdisanti, line 109:
```suggestion
    /// Create a new WASI HTTP client.
```

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+    fn new(options: WasiRequestOptions) -> Self {
+        Self { options }
+    }
+
+    /// Make outgoing http request in a Wasi environment
```
jdisanti, line 114:
```suggestion
    /// Make outgoing HTTP request in a WASI environment
```

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+    fn handle(&self, req: http::Request<Bytes>) -> Result<http::Response<Bytes>, ConnectorError> {
+        let req = WasiRequest::try_from(req).expect("Converting http request");
+
+        let res = outgoing_handler::handle(req.0, self.options.clone().0).expect("Http response");
```
jdisanti, line 118: Is it safe to `expect` here?

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,314 @@
...
+            let new_opts = RequestOptions::new();
+            new_opts
+                .set_between_bytes_timeout(opts.between_bytes_timeout())
+                .expect("Between bytes timeout");
+            new_opts
+                .set_connect_timeout(opts.connect_timeout())
+                .expect("Connect timeout");
+            new_opts
+                .set_first_byte_timeout(opts.first_byte_timeout())
+                .expect("First byte timeout");
```
jdisanti, line 160: So are all the expects due to each call out to a binding potentially failing due to a function not existing or something?

tools/ci-build/Dockerfile
```diff
@@ -121,17 +121,14 @@ ARG wasm_pack_version=0.11.0
...
 
 FROM install_rust AS wasmtime
-ARG wasmtime_precompiled_url=https://github.com/bytecodealliance/wasmtime/releases/download/v7.0.0/wasmtime-v7.0.0-x86_64-linux.tar.xz
-ARG wasmtime_precompiled_sha256=b8a1c97f9107c885ea73a5c38677d0d340a7c26879d366e8a5f3dce84cffec99
-RUN set -eux; \
-    curl "${wasmtime_precompiled_url}" -L -o wasmtime.xz; \
-    echo "${wasmtime_precompiled_sha256} wasmtime.xz" | sha256sum --check; \
-    tar xf wasmtime.xz; \
-    mv wasmtime-v*/wasmtime /opt;
+ARG cargo_wasmtime_version=17.0.0
```
jdisanti, line 124: If possible, it would be nice to use precompiled binaries to reduce image build time.

tools/ci-build/Dockerfile
```diff
@@ -49,11 +49,11 @@ RUN yum -y install --allowerasing \
...
     if [[ "$(uname -m)" == "aarch64" || "$(uname -m)" == "arm64" ]]; then \
-        curl https://static.rust-lang.org/rustup/archive/1.24.3/aarch64-unknown-linux-gnu/rustup-init --output rustup-init; \
-        echo "32a1532f7cef072a667bac53f1a5542c99666c4071af0c9549795bbdb2069ec1 rustup-init" | sha256sum --check; \
+    curl https://static.rust-lang.org/rustup/archive/1.24.3/aarch64-unknown-linux-gnu/rustup-init --output rustup-init; \
+    echo "32a1532f7cef072a667bac53f1a5542c99666c4071af0c9549795bbdb2069ec1 rustup-init" | sha256sum --check; \
     else \
-        curl https://static.rust-lang.org/rustup/archive/1.24.3/x86_64-unknown-linux-gnu/rustup-init --output rustup-init; \
-        echo "3dc5ef50861ee18657f9db2eeb7392f9c2a6c95c90ab41e45ab4ca71476b4338 rustup-init" | sha256sum --check; \
+    curl https://static.rust-lang.org/rustup/archive/1.24.3/x86_64-unknown-linux-gnu/rustup-init --output rustup-init; \
+    echo "3dc5ef50861ee18657f9db2eeb7392f9c2a6c95c90ab41e45ab4ca71476b4338 rustup-init" | sha256sum --check; \
```
jdisanti, line 56: Nit: leave these indents alone.

tools/ci-scripts/check-aws-sdk-standalone-integration-tests
```diff
@@ -39,4 +39,11 @@ find "${tmp_dir}"
 
 pushd "${tmp_dir}/aws/sdk/integration-tests"
 cargo check --tests --all-features
+
+# Running WebAssembly (WASI) specific integration tests
+pushd "${tmp_dir}/aws/sdk/integration-tests/webassembly" &>/dev/null
+cargo check --tests
+RUST_TEST_NOCAPTURE=1 cargo component test
```
jdisanti, line 46: This script isn't intended to actually run any tests since they run in a separate script. This one just checks that we didn't break the dev-only Cargo.toml files in aws/sdk/integration-tests, so it only needs `cargo check`

jdisanti (commented):
Thanks for taking this on! ❤️

CHANGELOG.next.toml
```diff
@@ -122,3 +122,9 @@ message = """Fix bug where timeout settings where not merged properly. This will
 references = ["smithy-rs#3405", "smithy-rs#3400", "smithy-rs#3258"]
 meta = { "bug" = true, "breaking" = true, tada = false }
 author = "rcoh"
+
+[[smithy-rs]]
+message = "Add aws-smithy-wasm crate to enable SDK use in wasi compliant environments"
```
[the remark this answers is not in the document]
landonxjames: For the moment everything in here is wasi related, but I can see a future where more wasm tooling is published and keeping both wasm/wasi functionality in the same crate makes sense to me

aws/sdk/integration-tests/webassembly/src/list_objects.rs
landonxjames: The `wasmtime` crate should allow you to run a WASI binary in a Rust Lambda, although I don't have any experience using it since we are calling the binaries from typescript and python. I can look into this if its a blocker for this PR, but will need some pointers around your canary setup.

rust-runtime/aws-smithy-wasm/src/wasi.rs
landonxjames: No, each of these `set_some_http_attribute` methods can fail if the string passed in can't be correctly parsed as the expected attribute. You can see the docs for those methods [here](https://docs.rs/wasi/0.12.0+wasi-0.2.0/wasi/http/types/struct.OutgoingRequest.html#method.set_path_with_query). Thankfully since this is a `Clone` we know that the values from the existing value should always be valid in this context.

aws/sdk/integration-tests/test.sh
```diff
@@ -11,6 +11,12 @@ for f in *; do
       echo
       echo "Testing ${f}..."
       echo "###############"
-      cargo test --manifest-path "${f}/Cargo.toml" --all-features
+      if [ "$f" != "webassembly" ]; then
+         cargo test --manifest-path "${f}/Cargo.toml" --all-features
+      else
+         # The webassembly tests use a custom runner set in config.toml that
+         # is not picked up when running the tests outside of the package
+         cd webassembly && cargo component test && cd ..
```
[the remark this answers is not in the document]
landonxjames: Addressed in https://github.com/smithy-lang/smithy-rs/pull/3409/commits/a38ef6ba579ea4dbcc018c2a125de7f8f7f2b8a9

aws/sdk/integration-tests/webassembly/src/list_objects.rs
landonxjames: Removed the network call in https://github.com/smithy-lang/smithy-rs/pull/3409/commits/0a87c1e4af5a41b6365830fc06301ffc134ce543, will add a full end to end test to the canary soon.

rust-runtime/aws-smithy-wasm/src/wasi.rs
landonxjames: As discussed offline I implemented an empty `WasiHttpClientBuilder` for this so it can be extended with more configuration in a backwards compatible way in the future.

rust-runtime/aws-smithy-wasm/src/wasi.rs
landonxjames: Discussed offline and I ended up taking the timeouts from the `HttpConnectorSettings` and baking them into the client, so this is taken care of

tools/ci-build/Dockerfile
landonxjames: Oops, autoformat thing I think, fixed.

tools/ci-scripts/check-aws-sdk-standalone-integration-tests
landonxjames: Removed the `test` invocation here

tools/ci-build/Dockerfile
landonxjames: Looks like wasmtime does offer [precompiled binaries](https://github.com/bytecodealliance/wasmtime/releases/tag/v17.0.1) for their tagged releases so this should be possible. I'll see if I have time to fit it in after updating the canary

landonxjames:
Ok pending this final CI run this should be all good to go. As requested I added a new canary `wasm_canary` which uses the `wasmtime` runtime to run a wasm file. The code for the wasm file lives in `tools/ci-cdk/canary-wasm` mostly because a separate crate was much easier to compile. The built wasm binary is bundled into the .zip thats deployed to the Lambda.

Currently the canary only tests a simple `list_objects_v2` with `no_credentials` set on the config. Note that I had to bump up the timeouts for the canary pretty substantially to get all of this to work. It was taking over a minute just to compile and instantiate the wasm binary. This could probably be improved with a beefier lambda instance, but I thought I would leave that choice up to the SDK team. 

Note: There is one CI task that continues to fail, the `check-aws-sdk-cargo-deny` task. It seems like something about the Cargo.toml in `aws-amithy-wasm` (I suspect the `target` statement for the dependencies) is causing the versions of the deps not to be resolved during the `:aws:sdk:assemble` step.

CHANGELOG.next.toml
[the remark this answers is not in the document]
landonxjames: I ended up adding an `AsyncSleep` impl in this crate too and that should work for any Wasm environment, WASI compliant or not. Its under a `wasm` module in the crate, so now I think keep the name definitely makes sense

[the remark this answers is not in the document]
landonxjames: I ended up adding an `AsyncSleep` impl in this crate too and that should work for any Wasm environment, WASI compliant or not. Its under a wasm module in the crate, so now I think keep the name definitely makes sense

rust-runtime/aws-smithy-wasm/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+authors = [
+  "AWS Rust SDK Team <aws-sdk-rust@amazon.com>",
+  "Eduardo Rodrigues <16357187+eduardomourar@users.noreply.github.com>",
+]
+description = "Smithy WebAssembly configuration for smithy-rs."
+edition = "2021"
+license = "Apache-2.0"
+repository = "https://github.com/awslabs/smithy-rs"
+
+[target.'cfg(all(target_family = "wasm", target_os = "wasi"))'.dependencies]
```
jdisanti, line 13: Is there any harm to making this crate have these dependencies for all targets? I think that would fix the CI issue.

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,345 @@
...
+use bytes::{Bytes, BytesMut};
+use wasi::http::{
+    outgoing_handler,
+    types::{self as wasi_http, OutgoingBody, RequestOptions},
+};
+
+/// Builder for [`WasiHttpClient`]. Currently empty, but allows for future
+/// config options to be added in a backwards compatible manner.
+#[derive(Default, Debug)]
+pub struct WasiHttpClientBuilder {}
```
jdisanti, line 31: So that it can't be constructed without calling `new` or `default`:
```suggestion
#[non_exhaustive]
pub struct WasiHttpClientBuilder {}
```

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,345 @@
...
+        let client = WasiHttpClient {};
+        client.into_shared()
+    }
+}
+
+/// An HTTP client that can be used during instantiation of the client SDK in
+/// order to route the HTTP requests through the WebAssembly host. The host must
+/// support the WASI HTTP proposal as defined in the Preview 2 specification.
+#[derive(Debug, Clone)]
+pub struct WasiHttpClient {}
```
jdisanti, line 50:
```suggestion
#[non_exhaustive]
pub struct WasiHttpClient {}
```

rust-runtime/aws-smithy-wasm/src/wasm.rs
```diff
@@ -0,0 +1,20 @@
...
+
+/// A struct implementing the AsyncSleep trait that can be used in
+/// Wasm environments
+#[derive(Debug, Clone)]
+pub struct WasmSleep;
+
+impl AsyncSleep for WasmSleep {
+    fn sleep(&self, duration: std::time::Duration) -> Sleep {
+        Sleep::new(Box::pin(async move {
+            tokio::time::sleep(duration).await;
```
jdisanti, line 17: Why have a separate `WasmSleep` if this is essentially identical to the `TokioSleep` in aws-smithy-async? Shouldn't customers also have the option to use a different async runtime in WASM?

tools/ci-cdk/canary-runner/src/build_bundle.rs
```diff
@@ -258,6 +263,11 @@ pub async fn build_bundle(opt: BuildBundleArgs) -> Result<Option<PathBuf>> {
     let crate_manifest_content = generate_crate_manifest(crate_source)?;
     fs::write(&manifest_path, crate_manifest_content).context("failed to write Cargo.toml")?;
 
+    let wasm_manifest_path = std::env::current_dir()
+        .expect("Current dir")
+        .join("../canary-wasm/Cargo.toml");
+
+    println!("MAINFEST PATH: {wasm_manifest_path:#?}");
```
jdisanti, line 270:
```suggestion
```

tools/ci-cdk/canary-runner/src/build_bundle.rs
```diff
@@ -271,6 +281,16 @@ pub async fn build_bundle(opt: BuildBundleArgs) -> Result<Option<PathBuf>> {
...
 
+        // Compile the wasm canary to a .wasm binary
+        let mut wasm_command = Command::new("cargo");
+        wasm_command
+            .arg("component")
+            .arg("build")
+            .arg("--release")
+            .arg("--manifest-path")
+            .arg(&wasm_manifest_path);
+        handle_failure("cargo build", &wasm_command.output()?)?;
```
jdisanti, line 292:
```suggestion
        handle_failure("cargo component build", &wasm_command.output()?)?;
```

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,345 @@
...
+
+impl From<&HttpConnectorSettings> for WasiRequestOptions {
+    fn from(value: &HttpConnectorSettings) -> Self {
+        //The WASI Duration is nanoseconds represented as u64
+        let connect_timeout = value
+            .connect_timeout()
+            .map(|dur| u64::try_from(dur.as_nanos()).unwrap_or(u64::MAX));
+        let read_timeout = value
+            .read_timeout()
+            .map(|dur| u64::try_from(dur.as_nanos()).unwrap_or(u64::MAX));
```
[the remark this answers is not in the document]
landonxjames: If they aren't called the default appears to be `600,000` milliseconds ([set here](https://github.com/bytecodealliance/wasmtime/blob/36fb62ca3bea7dd1456f12ead03084e2ceb48cda/crates/wasi-http/src/http_impl.rs#L23-L33)). But this is set in `wasmtime` and not by the WASI bindings, so it seems to be host dependent. To be consistent with that I can change the `u64::MAX` to match their 600 seconds default.

rust-runtime/aws-smithy-wasm/src/wasm.rs
landonxjames: Had no idea `TokioSleep` existed, will remove this in favor of that!

rust-runtime/aws-smithy-wasm/src/wasi.rs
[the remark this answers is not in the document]
landonxjames: Breaking this down a little more: 

The behavior of `set_*_timeout`:
* `set_*_timeout` is never called - then the default timeout (for all of them) is set to `600` seconds ([code](https://github.com/bytecodealliance/wasmtime/blob/36fb62ca3bea7dd1456f12ead03084e2ceb48cda/crates/wasi-http/src/http_impl.rs#L23-L33)). Note that we don't encounter this in our code since the timeouts are always set based on the `HttpConnectorSettings` values. 
* `set_*_timeout` is called with `None` - the timeout is set to 0 ([code](https://github.com/bytecodealliance/wasi/blob/8c5c76f94a35eb3f4022da2a81612d7aa86a573c/src/bindings.rs#L6069C25-L6072C27))
* `set_*_timeout` is called with `Some(u64)` - timeout is set to the `u64` value

The behavior of `WasiRequestOptions::from(HttpConnectorSettings)`:
* The `HttpConnectorSettings` contains a timeout value (`u128`) less than `u64::MAX` - Then the corresponding timeout value on `outgoing_handler::RequestOptions` is set to that value (cast to a `u64`)
* The `HttpConnectorSettings` contains a timeout value (`u128`) greater than `u64::MAX` - Then the value is clamped to `u64::MAX` and set to that

rust-runtime/aws-smithy-wasm/src/wasm.rs
landonxjames: Addressed in https://github.com/smithy-lang/smithy-rs/pull/3409/commits/c883aad007e0f5929794be8f5370854679a5f213

rust-runtime/aws-smithy-wasm/src/wasi.rs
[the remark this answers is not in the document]
landonxjames: Added comments explaining the clamping in https://github.com/smithy-lang/smithy-rs/pull/3409/commits/c883aad007e0f5929794be8f5370854679a5f213

rust-runtime/aws-smithy-wasm/Cargo.toml
landonxjames: Split out the `wasi` crate into its own targeted dependencies section in https://github.com/smithy-lang/smithy-rs/pull/3409/commits/c883aad007e0f5929794be8f5370854679a5f213

jdisanti (commented):
Code changes look good to me! Just need to get CI passing at this point.

rust-runtime/aws-smithy-wasm/src/wasi.rs
```diff
@@ -0,0 +1,350 @@
...
+        let options = WasiRequestOptions::from(settings);
+        let connector = WasiHttpConnector { options };
+
+        connector.into_shared()
+    }
+}
+
+/// HTTP connector used in WASI environment
+#[derive(Debug, Clone)]
+pub struct WasiHttpConnector {
```
jdisanti, line 69: Just one last change. It looks like this can be private, and since it's not possible to construct outside the crate, we should go ahead and do that to reduce the odds that we are constrained in the future due to backwards compatibility.
```suggestion
struct WasiHttpConnector {
```

jdisanti (approved):

PR #3412, opened 2024-02-14

ysaito1001:
## Motivation and Context
 
## Description
- through a profile file with a key `s3_disable_express_session_auth` (won't be supported yet in this PR until https://github.com/awslabs/aws-sdk-rust/issues/1060 is implemented)
 
#### `EndpointParamsInterceptor` now uses `read_before_serialization` instead of `read_before_execution`
This means the previous logic for `EndpointParamsInterceptor` will now kick in at a later point in the orchestrator execution. This is to support the following use case:
```
 
    let _ = client
        .list_objects_v2()
        .customize()
        .config_override(Config::builder().disable_s3_express_session_auth(false))
        .send()
        .await;
```
This code should end up using the S3 Express session auth since what's passed to the S3 client method should take the highest precedence (and the one set by the environment variable should be overruled). To support this, `EndpointParamsInterceptor` needs to "hold off" on grabbing `DisableS3ExpressSessionAuth` from a config bag so that the `DisableS3ExpressSessionAuth` fully reflects all three disabling places above, which is ensured by a newly added interceptor `DisableS3ExpressSessionAuthInterceptor`. However, the new interceptor needs to run after the config bag merges service-level and operation-level configurations (because there is a chance of `config_override` like above) but before the `EndpointParamsInterceptor`'s trait method runs.

The previous execution point for `EndpointParamsInterceptor` was [here](https://github.com/smithy-lang/smithy-rs/blob/1f7d70e81b6752c5c886fe5344b9d34d52504b3e/rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs#L188), but this left no room for `DisableS3ExpressSessionAuthInterceptor` to deterministically run before `EndpointParamsInterceptor` with a fully merged `ConfigBag`. Why? If `DisableS3ExpressSessionAuthInterceptor` were registered at a client level and implemented `read_before_execution`, it would run before `EndpointParamsInterceptor` but the config bag has not fully merged operation-level config. If `DisableS3ExpressSessionAuthInterceptor` were registered at an operation-level and implemented `read_before_execution`, both `EndpointParamsInterceptor` and that would be operation-level interceptors implementing the same `Intercept` trait method, no guarantee that which one runs first.

Because of this, the PR now implements `EndpointParamsInterceptor::read_before_serialization` instead of `EndpointParamsInterceptor::read_before_execution`, leaving room for `DisableS3ExpressSessionAuthInterceptor` to kick in and ensuring it has access to a fully merged `ConfigBag`.
 
#### `DisableS3ExpressSessionAuthInterceptor` is a client-level interceptor implementing `read_before_serialization`
This is because it needs to carry over a piece of information `s3_disable_express_session_auth` set by a profile file. Within `From<&SdkConfig> for Builder`, that piece of information is transferred to the interceptor (note that we don't call `builder.set_disable_s3_express_session_auth(...)` as doing so will collide with a customer setting a disable flag through the `disable_s3_express_session_auth` method on an S3 client). The interceptor then implements `read_before_serialization` so that it has access to a fully merged `ConfigBag` and runs before `EndpointParamsInterceptor::read_before_serialization`.

## Testing
Added unit tests for `DisableS3ExpressSessionAuthInterceptor` and integration tests for verifying disabling S3 Express session auth. Tests for a profile file will be TODO until the said work is implemented.
 
----
 
### DIFF
--- a/aws/rust-runtime/aws-inlineable/src/s3_express.rs
+++ b/aws/rust-runtime/aws-inlineable/src/s3_express.rs
@@ -731,3 +731,203 @@ pub(crate) mod identity_provider {
         }
     }
 }
+
+/// Supporting code for S3 Express interceptors
+pub(crate) mod interceptors {
+    use aws_smithy_runtime_api::box_error::BoxError;
+    use aws_smithy_runtime_api::client::interceptors::context::BeforeSerializationInterceptorContextRef;
+    use aws_smithy_runtime_api::client::interceptors::Intercept;
+    use aws_smithy_runtime_api::client::runtime_components::RuntimeComponents;
+    use aws_smithy_types::config_bag::{ConfigBag, Layer};
+    use aws_types::os_shim_internal::Env;
+
+    mod env {
+        pub(super) const S3_DISABLE_EXPRESS_SESSION_AUTH: &str =
+            "AWS_S3_DISABLE_EXPRESS_SESSION_AUTH";
+    }
+
+    // Interceptor to disable S3 Express session auth
+    //
+    // It is important for this interceptor to be registered as a client-level interceptor so that
+    // we ensure that it runs before `EndpointParamsInterceptor::read_before_serialization`, which
+    // is an operation-level interceptor.
+    // `DisableS3ExpressSessionAuth` is a single source of truth indicating whether the session auth
+    // is disabled. A customer can disable the session auth through a) s3 client b) an environment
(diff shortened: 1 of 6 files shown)


ysaito1001:
Internally got the feedback that will considerably change the way it is implemented. Instead of using an interceptor, we should implement it via a RuntimePlugin for S3 express that configures a value for `DisableS3ExpressSessionAuth`.

PR #3416, opened 2024-02-15
Support EKS Pod Identity credentials

jackkleeman:
## Motivation and Context
I would like to support EKS Pod Identity credentials in the Rust SDKs

## Description
This brings the ECS provider in line with other sdks (eg, Go) by supporting AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE as well as permitting http IPs to be non-loopback if they are the EKS pod identity IPs.

## Testing
I have added various new unit tests, and I have updated the existing integration test to also create pods with eks pod identity creds, which I have used to test in a real EKS cluster as well.

## Checklist

----


rcoh:
This looks great! Someone from our team will do a detailed review next week. The one missing piece is a e2e integration test. You can see examples here: https://github.com/smithy-lang/smithy-rs/tree/6e3e010e912874262b37727d900d7fae170efaed/aws/rust-runtime/aws-config/test-data/default-provider-chain/ecs_credentials_invalid_profile

jackkleeman:
done @rcoh

aws/rust-runtime/aws-config/src/ecs.rs
```diff
@@ -17,16 +17,25 @@
...
 //!
-//! **Next**: It wil check the value of `$AWS_CONTAINER_CREDENTIALS_FULL_URI`. This specifies the full
-//! URL to load credentials. The URL MUST satisfy one of the following two properties:
+//! **Next**: It will check the value of `$AWS_CONTAINER_CREDENTIALS_FULL_URI`. This specifies the full
+//! URL to load credentials. The URL MUST satisfy one of the following three properties:
 //! 1. The URL begins with `https`
-//! 2. The URL refers to a loopback device. If a URL contains a domain name instead of an IP address,
-//! a DNS lookup will be performed. ALL resolved IP addresses MUST refer to a loopback interface, or
-//! the credentials provider will return `CredentialsError::InvalidConfiguration`
+//! 2. The URL refers to an allowed IP addresses. If a URL contains a domain name instead of an IP address,
```
jdisanti, line 23:
```suggestion
//! 2. The URL refers to an allowed IP address. If a URL contains a domain name instead of an IP address,
```

aws/rust-runtime/aws-config/src/ecs.rs
```diff
@@ -93,15 +104,32 @@ impl EcsCredentialsProvider {
 
     /// Load credentials from this credentials provider
     pub async fn credentials(&self) -> provider::Result {
-        let auth = match self.env.get(ENV_AUTHORIZATION).ok() {
-            Some(auth) => Some(HeaderValue::from_str(&auth).map_err(|err| {
-                tracing::warn!(token = %auth, "invalid auth token");
+        let env_token = self.env.get(ENV_AUTHORIZATION_TOKEN).ok();
+        let env_token_file = self.env.get(ENV_AUTHORIZATION_TOKEN_FILE).ok();
+        let auth = if let Some(auth_token) = env_token {
```
jdisanti, line 109: `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` is supposed to take priority over `AWS_CONTAINER_AUTHORIZATION_TOKEN`, so these conditions will need to get swapped, and the doc comment updated. There should be a test that sets both to exercise the precedence too.

aws/rust-runtime/aws-config/src/ecs.rs
```diff
@@ -700,23 +868,30 @@ mod test {
...
         );
-        assert!(logs_contain(
-            "Address does not resolve to the loopback interface"
-        ));
-        validate_full_uri("http://localhost:8888/creds", dns)
+        assert!(logs_contain("Address does not resolve to an allowed IP"));
+        validate_full_uri("http://localhost:8888/creds", dns.clone())
             .await
             .expect("localhost is the loopback interface");
+        validate_full_uri("http://169.254.170.2.backname.io:8888/creds", dns.clone())
```
jdisanti, line 886: This definitely works, but I'm a little hesitant to have our tests rely on backname.io since it adds some level of risk to our release process if backname.io happens to go down for a period of time. Could you refactor the test to use a fake `ResolveDns` trait implementation instead of the real one?

jdisanti (commented):
This is a really high quality contribution, thank you! ❤️ 

There are just a couple of things that need to be changed before we can merge.

aws/rust-runtime/aws-config/src/ecs.rs
jackkleeman: this test is skipped because it already relies on dns resolution (of amazon.com)
see comment `// ignored by default because it relies on actual DNS resolution`
Is there some release process that is unskipping it?

jdisanti: Oh, nope. I just didn't see that comment. Looks good then.

jdisanti (approved):
Looks great. Thank you!

jackkleeman:
@jdisanti did this make it into the changelog?

jackkleeman:
I guess I did something that meant the automation didn't pick it up - probably should be in there?

https://github.com/smithy-lang/smithy-rs/commit/b515533c07b7eeb4e99c0a5f427bb6d00dec7ab4

PR #3421, opened 2024-02-16

ysaito1001:
## Motivation and Context

## Description
- through a profile file with a key `s3_disable_express_session_auth` (won't be supported yet in this PR until https://github.com/awslabs/aws-sdk-rust/issues/1060 is implemented)
 
Something to be aware of when setting the disable option through the environment variable. The environment variable is only checked during a client construction, meaning that if a customer sets it after the client has been created the SDK will not take the environment variable value into account, i.e. the following snippet will NOT disable the S3 Express session auth:
```

// Set the env variable to true after an S3 client has been created

let _ = client
    .list_objects_v2()
    .send()
    .await;
``` 

## Testing
Added unit tests for `S3ExpressRuntimePlugin` and integration tests for verifying disabling S3 Express session auth.
Tests for a profile file will be TODO until the said work is implemented.

----


aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -727,3 +727,124 @@ pub(crate) mod identity_provider {
...
+            let mut layer = Layer::new("S3ExpressRuntimePlugin");
+            if service_config
+                .load::<crate::config::DisableS3ExpressSessionAuth>()
+                .is_none()
+            {
+                match env.get(env::S3_DISABLE_EXPRESS_SESSION_AUTH) {
+                    Ok(value)
+                        if value.eq_ignore_ascii_case("true")
+                            || value.eq_ignore_ascii_case("false") =>
+                    {
```
jdisanti, line 762: Is the expected behavior that the value is just ignored if it's not true or false? E.g., `S3_DISABLE_EXPRESS_SESSION_AUTH=1` would result in it being ignored? Is it supposed to warn in that case?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
```diff
@@ -178,6 +179,17 @@ class S3ExpressIdentityProviderConfig(codegenContext: ClientCodegenContext) : Co
                     )
                 }
 
+                ServiceConfig.BuilderBuild -> {
+                    rustTemplate(
+                        """
+                        self.runtime_plugins.push(
```
jdisanti, line 185: I think this is probably a better place to hook into the runtime plugins since the ones in config are intended to be provided by customers:
https://github.com/smithy-lang/smithy-rs/blob/5761d1fe12e1a9235f9332122b867800db5d879a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt#L490

We would need to add a codegen customization hook in there, but we should probably have one anyway.

aws/sdk/integration-tests/s3/tests/express.rs
```diff
@@ -0,0 +1,249 @@
...
+
+#[tokio::test]
+async fn list_objects_v2() {
+    let _logs = capture_test_logs();
+
+    let http_client =
+        ReplayingClient::from_file("tests/data/express/list-objects-v2.json").unwrap();
+    let config = aws_config::from_env()
+        .http_client(http_client.clone())
+        .no_credentials()
```
rcoh, line 27: if this has no creds, how does it work?

aws/sdk/integration-tests/s3/tests/express.rs
```diff
@@ -0,0 +1,249 @@
...
+        .send()
+        .await;
+
+    let req = request.expect_request();
+    assert!(
+        !req.headers()
+            .get("authorization")
+            .unwrap()
+            .contains("x-amz-create-session-mode"),
+        "x-amz-create-session-mode should not appear in headers when S3 Express session auth is disabled"
```
rcoh, line 219: we probably want an S3 express section added to the canary. We'll need to update the CDK for the canary to create the bucket. That's where I'd probably smoke test all of the APIs

ysaito1001: Yeah, I was planning to add an S3 express test for canary, but at the same time I'd like to mainly exercise the happy path use cases. A failure in canary will block our release, so I want to reduce the amount of things we do in canary as much as possible. For testing a negative case like disabling S3 Express, I'd still want to keep in an integration test.

ysaito1001:
Will re-create a new PR due to many merge conflicts at this point. Will address review feedback in the new PR.

aws/sdk/integration-tests/s3/tests/express.rs
ysaito1001: A subsequent call to `.with_test_defaults()` will add one.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
ysaito1001: I thought about that too (including the need for a codegen customization hook there), but had a question whether once we're in `base_client_runtime_plugins`, a disabling value from method `Config::disable_s3_express_session_auth` and that in a profile file are stored in separate places? If they are consolidated into a single value, then there isn't a way for the environment variable to tell where that disabling value originally came from (i.e. if it comes from the method, then the env variable should not override it but if comes from a profile file, it should override it).

ysaito1001: That said, whatever we can do in `BuilderBuild`, we can also do in `base_client_runtime_plugins` such as accessing `pub(crate) config: crate::config::FrozenLayer,`. Doing so in `base_client_runtime_plugins` feels it's more right thing to do. Let me pursue that approach.

PR #3425, opened 2024-02-17
Fix credentials-verification workflow

jdisanti:


jdisanti:
Tested the fix by manually kicking it off against this branch: https://github.com/smithy-lang/smithy-rs/actions/runs/7938459322

rcoh (approved):
Weird this definitely used to work

jdisanti:
The Rust version isn't pinned down in this workflow, so perhaps something changed there.

ysaito1001:
> The Rust version isn't pinned down in this workflow, so perhaps something changed there.

Thanks for fixing!

PR #3429, opened 2024-02-19
Support for accessing arbitrary profile config data

Velfi:
I still have some TODOs and naming to think about but the main functionality is all there for review

## Motivation and Context

This is required to enable an upcoming service-specific env config feature

## Description

This PR adds support for accessing profile config data defined in any sort of section. It also supports sub-properties.

## Testing

I wrote a test

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/profile/parser.rs
+++ b/aws/rust-runtime/aws-config/src/profile/parser.rs
@@ -10,7 +10,7 @@ use aws_types::os_shim_internal::{Env, Fs};
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::error::Error;
-use std::fmt::{Display, Formatter};
+use std::fmt::{Debug, Display, Formatter};
 use std::path::PathBuf;
 use std::sync::Arc;
 
@@ -69,12 +69,34 @@ pub async fn load(
     Ok(ProfileSet::parse(source)?)
 }
 
+type SectionType = String;
+type SectionName = String;
+// One example of what this looks like in practice:
+// ```
+// {
+//   "services": { // section_type
+//     "foo": { // section_name
+//       "ec2": { // sub_properties_name
+//          "endpoint_url": "https://foo.example" // sub-properties
+//        },
+//       "s3": { "endpoint_url": "https://foo.example", "some_other_property": "a value" },
+//     },
+//     "bar": {
+//       "ec2": { "endpoint_url": "https://bar.example" },
+//     }
+// }
+// ```
+// Dictionaries in Rust man... No fun at all.
+type OtherSections =
+    HashMap<SectionType, HashMap<SectionName, HashMap<String, HashMap<String, String>>>>;
+
 /// A top-level configuration source containing multiple named profiles
 #[derive(Debug, Eq, Clone, PartialEq)]
 pub struct ProfileSet {
     profiles: HashMap<String, Profile>,
     selected_profile: Cow<'static, str>,
     sso_sessions: HashMap<String, SsoSession>,
+    other_sections: OtherSections,
 }
 
 impl ProfileSet {
@@ -153,6 +175,44 @@ impl ProfileSet {
         self.sso_sessions.get(name)
     }
 
+    /// Returns a list of all section names and their section type in tuple form e.g. `("profile", "user_a"), ("sso_session", "some_session"), ("services", "local_dev")`
+    ///
+    /// No sort order is guaranteed.
+    ///
+    /// # Example
+    ///
+    /// TODO add example
+    pub fn sections(&self) -> impl Iterator<Item = (&str, &str)> {
+        let other_sections = self
+            .other_sections
+            .iter()
+            .flat_map(|(k1, v)| v.keys().map(|k2| (k1.as_str(), k2.as_str())));
+        let profiles = self.profiles().map(|it| ("profile", it));
+        let sso_sessions = self.sso_sessions().map(|it| ("sso_session", it));
+
+        other_sections.chain(profiles).chain(sso_sessions)
+    }
+
+    /// Given a section type and name, retrieve the corresponding section, if it exists.
+    ///
+    /// # Example
+    ///
+    /// TODO Make this work for profile and sso_session
+    pub fn get_raw_section_with_properties<'a>(
+        &'a self,
+        section_type: &str,
+        section_name: &str,
+    ) -> Option<&HashMap<String, HashMap<String, String>>> {
+        match section_type {
+            "profile" => None,
+            "sso_session" => None,
+            _ => self
+                .other_sections
+                .get(section_type)
+                .and_then(|sections| sections.get(section_name)),
+        }
+    }
+
     fn parse(source: Source) -> Result<Self, ProfileParseError> {
         let mut base = ProfileSet::empty();
         base.selected_profile = source.profile;
@@ -168,6 +228,7 @@ impl ProfileSet {
             profiles: Default::default(),
             selected_profile: "default".into(),
             sso_sessions: Default::default(),
+            other_sections: Default::default(),
         }
     }
 }
@@ -389,7 +450,9 @@ mod test {
     };
     use crate::profile::profile_file::ProfileFileKind;
     use crate::profile::ProfileSet;
+    use crate::provider_config::ProviderConfig;
     use arbitrary::{Arbitrary, Unstructured};
+    use aws_types::os_shim_internal::{Env, Fs};
     use serde::Deserialize;
     use std::collections::HashMap;
     use std::error::Error;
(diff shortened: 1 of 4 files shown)


aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -69,12 +69,34 @@ pub async fn load(
...
+//       "s3": { "endpoint_url": "https://foo.example", "some_other_property": "a value" },
+//     },
+//     "bar": {
+//       "ec2": { "endpoint_url": "https://bar.example" },
+//     }
+// }
+// ```
+// Dictionaries in Rust man... No fun at all.
+type OtherSections =
+    HashMap<SectionType, HashMap<SectionName, HashMap<String, HashMap<String, String>>>>;
```
jdisanti, line 91: Should we adopt something similar to `serde_json::Value` or use our own `Document` type? Or better yet, perhaps we should make our own struct type for this use-case that just has lookup methods on it and uses the hash map nastiness internally?

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -153,6 +175,44 @@ impl ProfileSet {
...
+    /// Given a section type and name, retrieve the corresponding section, if it exists.
+    ///
+    /// # Example
+    ///
+    /// TODO Make this work for profile and sso_session
+    pub fn get_raw_section_with_properties<'a>(
+        &'a self,
+        section_type: &str,
+        section_name: &str,
+    ) -> Option<&HashMap<String, HashMap<String, String>>> {
```
jdisanti, line 205: I'm slightly concerned enforcing a pairing of section_type/section_name in the public API could cause issues in the future given we have sections such as `[default]`.

aws/rust-runtime/aws-config/src/profile/parser/normalize.rs
```diff
@@ -161,6 +183,42 @@ pub(super) fn merge_in(
...
+                    .or_default();
+                // get section by section name
+                let named_section = sections
+                    .entry(section_name.to_string())
+                    // if it doesn't exist, create new hashmap for that section name
+                    .or_default();
+
+                for (sub_properties_name, raw_sub_properties) in &raw_profile {
+                    tracing::info!("[{section_type} {section_name}].{sub_properties_name} = {raw_sub_properties}");
+                    // TODO do we need a more-specific validator for service IDs here?
```
jdisanti, line 205: I think aws-config definitely isn't the place to do that validation since someone could use an old version of aws-config with a newly launched SDK crate, and it should work in that case.

The SDK crates also can't validate it though since they only know about their own individual service. So, really, I don't think any kind of validation can be done here.

aws/rust-runtime/aws-config/src/profile/parser.rs
Velfi: In the case of `[default]`, I see that as just an unnamed `profile`-type section. I'm willing to budge if we really thing that we'll have 'typeless sections' in the future though.

aws/rust-runtime/aws-config/src/profile/parser.rs
rcoh: yeah we definitely need our own struct type here

aws/rust-runtime/aws-config/src/profile/parser/normalize.rs
rcoh: yeah we can only validate based on general profile file syntax

aws/rust-runtime/aws-config/src/profile/parser.rs
rcoh: I don't think that Value or our own document is the right thing because this is stringy all the way down

aws/rust-runtime/aws-config/src/profile/parser/normalize.rs
```diff
@@ -253,25 +222,25 @@ fn validate_identifier(input: &str) -> Result<&str, ()> {
...
+                .map(|it| it.trim_matches(WHITESPACE).to_owned());
+            let value = split
+                .next()
+                .map(|it| it.trim_matches(WHITESPACE).to_owned());
 
-    map
+            if let (Some(key), Some(value)) = (key, value) {
+                Some((key, value))
+            } else {
+                None
```
jdisanti, line 241: Should this case log a warning?

aws/rust-runtime/aws-config/src/profile/parser/section.rs
```diff
@@ -0,0 +1,314 @@
...
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn insert_at(&mut self, path: &[&str], value: String) {
+        if path.is_empty() {
+            return;
+        }
+
+        if path.len() != 4 {
```
jdisanti, line 59: Doesn't it also need to support 3 for the non-nested properties that are needed from the active profile section?

aws/rust-runtime/aws-config/src/profile/parser/section.rs
```diff
@@ -0,0 +1,314 @@
...
+    // [section-key section-name]
+    // sub-property-group-name =
+    //   sub-property-name = sub-property-value
+    inner: HashMap<
+        SectionKey,
+        HashMap<
+            SectionName,
+            HashMap<SubPropertyGroupName, HashMap<SubPropertyKey, SubPropertyValue>>,
+        >,
+    >,
```
jdisanti, line 46: It might be simpler to have a `HashMap<(SectionKey, SectionName, Key, Option<NestedKey), String>` instead of nesting the hash maps, or if the tuple is too much, make a struct for those fields in the key.

jdisanti (commented):
This is definitely an improvement. I think we could simplify the inner nested hashmap down to just a single hash map by creating a new key type though. Should greatly simplify the insertion and retrieval logic.

aws/rust-runtime/aws-config/src/profile/parser.rs
Velfi: I created a struct.

aws/rust-runtime/aws-config/src/profile/parser/section.rs
Velfi: I updated this to work with a simpler key type.

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -75,6 +78,7 @@ pub struct ProfileSet {
     profiles: HashMap<String, Profile>,
     selected_profile: Cow<'static, str>,
     sso_sessions: HashMap<String, SsoSession>,
+    other_sections: Properties,
```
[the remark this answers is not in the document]
Velfi: I was just going to avoid messing with this stuff at all. I'll reach out to you and we can discuss.

aws/rust-runtime/aws-config/src/standard_property.rs
```diff
@@ -3,23 +3,44 @@
...
         match self {
-            PropertySource::Environment { name } => write!(f, "environment variable `{}`", name),
-            PropertySource::Profile { name, key } => {
-                write!(f, "profile `{}`, key: `{}`", name, key)
+            PropertySource::Environment {
+                name,
+                service_specific,
+            } => {
+                #[rustfmt::skip]
+                let service_specific = if *service_specific { "service-specific " } else { "" };
```
rcoh, line 33: if you really want this on one line you could do 
`let service_specific = service_specific.then(|| "service-specific ").unwrap_or_default()`

aws/rust-runtime/aws-config/src/standard_property.rs
```diff
@@ -94,28 +125,238 @@ impl StandardProperty {
...
+            r#"[default]
+some_key = 4
+services = dev
+
+[services dev]
+service =
+  some_key = 5
+another_service =
+  some_key = 6
+"#,
```
rcoh, line 256: recommend adding a test with an additional services section

aws/rust-runtime/aws-config/src/standard_property.rs
```diff
@@ -94,28 +125,238 @@ impl StandardProperty {
...
+                            PropertySource::Profile {
+                                name: profile.selected_profile(),
+                                key: profile_key,
+                                service_specific: false,
+                            },
+                        )
+                    })
+                })
+        });
+
```
rcoh, line 163: there's probably some opportunity to cleanup / unify this code a little bit but no worries

rcoh (approved):
I have some cleanup comments but this looks good

aws/rust-runtime/aws-config/src/profile/parser.rs
```diff
@@ -153,6 +159,12 @@ impl ProfileSet {
         self.sso_sessions.get(name)
     }
 
+    /// Returns a struct allowing access to other sections in the profile config
+    #[allow(dead_code)] // Leaving this hidden for now.
+    pub(crate) fn other_sections(&self) -> &Properties {
```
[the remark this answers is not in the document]
Velfi: I'm going to expose it in a later PR.

Issue #3430, opened 2024-02-20
ConstraintViolation structure does not implement std::error::Error


drganjoo:
A `ConstraintViolation` structure is generated for each constrained type. However, it does not implement `std::error::Error`, making it difficult for client code to use the `?` operator.

PR #3431, opened 2024-02-20
Impl std::error::Error for a shape's ConstraintViolation struct

drganjoo:
## Motivation and Context
Each constrained shape has an associated `ConstraintViolation` structure, which at the moment does not implement `std::error::Error`.

## Description

All of the `ConstraintViolation` structures now implements `std::fmt::Display` and the `std::error::Error` marker trait. There is a difference between the message used for `pub(crate) fn as_validation_exception_field` and for `std::fmt::Display` as the latter does not know of the particular field which is being set in the structure. Hence, it uses the shape ID instead.

For example, for a constrained `int`, the following is the message used in `Display`: 

```
write!(f, "Value for `com.aws.example#MyNumberSensitive`failed to satisfy constraint: Member must be greater than or equal to 100")

```
and the following is used for `as_validation_exception_field`:

```
format!("Value at '{}' failed to satisfy constraint: Member must be greater than or equal to 100", &path),
```

## Testing

Tests have been added to ensure:

1. `std::fmt::Display` has all of the variants for the enum that represents the associated `Error` type in the `ConstraintViolation` struct.
2. `ConstraintViolation` implements `std::error::Error`.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/MapConstraintViolationGenerator.kt
```diff
@@ -84,8 +88,22 @@ class MapConstraintViolationGenerator(
                     ${if (keyConstraintViolationExists) "##[doc(hidden)] Key(#{KeyConstraintViolationSymbol})," else ""}
                     ${if (valueConstraintViolationExists) "##[doc(hidden)] Value(#{KeySymbol}, #{ValueConstraintViolationSymbol})," else ""}
                 }
+                
+                impl #{Display} for $constraintViolationName {
+                    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                        return match self {
+                            ${if (shape.hasTrait<LengthTrait>()) "#{LengthMatchingArm}" else ""}
+                            ${if ((keyConstraintViolationExists) || (valueConstraintViolationExists)) """other => write!(f, "{}", other),""" else ""}
```
jdisanti, line 96: Does this infinitely recurse? Basically self::display -> write -> self::display?

drganjoo: Oh good catch! Yes, it will recurse indefinitely. Hmm.. the tests should have caught this. Let me fix the tests as well. Thanks

CHANGELOG.next.toml
```diff
@@ -22,3 +22,9 @@ message = "The MSRV has been increase to 1.74.1"
 references = ["smithy-rs#3410"]
 meta = { "bug" = false, "breaking" = false, tada = false }
 author = "rcoh"
+
+[[smithy-rs]]
+message = "Implement std::error::Error for ConstraintViolation"
```
david-perez, line 27:
```sugestion
+message = "Implement `std::error::Error` for `ConstraintViolation`"
```
Why don't we implement `std::error::Error` for all operation error types, not only `ConstraintViolation`s?

I guess the reasoning is there's not an obvious `Display` representation for an arbitrary `@error` structure, but for `ConstraintViolation` there is? I don't like this asymmetry.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/EnumTraitValidationErrorMessage.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import software.amazon.smithy.model.shapes.Shape
+import software.amazon.smithy.model.traits.EnumTrait
+import software.amazon.smithy.rust.codegen.server.smithy.generators.idWithHashEscaped
+
+fun EnumTrait.validationErrorMessage() =
+    "Value at '{}' failed to satisfy constraint: Member must satisfy enum value set: [${enumValueSet()}]"
+
+fun EnumTrait.shapeConstraintViolationDisplayMessage(shape: Shape) =
+    "Value provided for '${shape.idWithHashEscaped()}' failed to satisfy constraint: Member must satisfy enum value set: [${enumValueSet()}]"
```
david-perez, line 16: I'd argue that it's not this function's responsibility to escape the hash, but rather the caller's, who is using `RustWriter`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/PatternTraitValidationErrorMessage.kt
```diff
@@ -0,0 +1,31 @@
...
+fun PatternTrait.validationErrorMessage() =
+    "Value at '{}' failed to satisfy constraint: Member must satisfy regular expression pattern: ${this.patternDescription()}"
+
+fun PatternTrait.shapeConstraintViolationDisplayMessage(shape: Shape) =
+    "Value provided for `${shape.idWithHashEscaped()}` failed to satisfy the constraint: Member must match the regular expression pattern: ${this.patternDescription()}"
+
+// A '#' character in the pattern must be replaced with "##" for the message to be usable
+// within `rustTemplate`, as it interpolates anything prefixed with '#'. Additionally,
+// passing the `toString()` representation of a regular expression directly to a function
+// that interpolates (e.g., println!, format!) can cause an error if the string contains
```
david-perez, line 21:
```suggestion
+// that interpolates (e.g., `println!`, `format!`) can cause an error if the string contains
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/CollectionConstraintViolationGenerator.kt
```diff
@@ -96,4 +112,20 @@ class CollectionConstraintViolationGenerator(
             }
         }
     }
+
+    private fun generateDisplayMessageForEachVariant(memberConstraintVariantPresent: Boolean) =
+        writable {
+            collectionConstraintsInfo.forEach {
```
david-perez, line 118: nit: not a fan of `forEach` if not already in an FP-style chain. A regular for loop is clearer.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedBlobGenerator.kt
```diff
@@ -143,9 +160,22 @@ class ConstrainedBlobGenerator(
...
+        }
 }
 
-data class BlobLength(val lengthTrait: LengthTrait) {
+// Each type of constraint that can be put on a Blob must implement the BlobConstraintGenerator
+// interface. This allows the
+interface BlobConstraintGenerator {
+    fun shapeConstraintViolationDisplayMessage(shape: Shape): Writable
+}
+
```
david-perez, line 177: Do we need this interface? I see this akin to `CollectionTraitInfo` in the case of collections, which is a `sealed class`.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/MapConstraintViolationGenerator.kt
```diff
@@ -84,8 +88,23 @@ class MapConstraintViolationGenerator(
                     ${if (keyConstraintViolationExists) "##[doc(hidden)] Key(#{KeyConstraintViolationSymbol})," else ""}
                     ${if (valueConstraintViolationExists) "##[doc(hidden)] Value(#{KeySymbol}, #{ValueConstraintViolationSymbol})," else ""}
                 }
+                
+                impl #{Display} for $constraintViolationName {
+                    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                        return match self {
```
david-perez, line 94:
```suggestion
+                        match self {
```
Isn't this caught by Clippy?

david-perez (commented):
My only concern with landing this is that we're introducing asymmetry. Why should we treat `ConstraintViolation` structs as special and implement `Display` and `Error` for them, but not for the rest of operation errors?

CHANGELOG.next.toml
drganjoo: We are already generating code to implement `std::fmt::Display` and `std::error::Error` on all `@error` structures.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedShapeGeneratorCommon.kt
```diff
@@ -26,3 +28,5 @@ fun rustDocsTryFromMethod(
 fun rustDocsInnerMethod(inner: String) = "Returns an immutable reference to the underlying [`$inner`]."
 
 fun rustDocsIntoInnerMethod(inner: String) = "Consumes the value, returning the underlying [`$inner`]."
+
+fun Shape.idWithHashEscaped() = this.id.toString().replace("#", "##")
```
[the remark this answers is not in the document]
drganjoo: Yes, I agree. This should be defined in a more central location. However, I am removing this for now. You make a good point that I should not be replacing '#' with '##' in a message since that is a requirement for `rustTemplate`. Therefore, the caller should do that; otherwise, the message would not be useful for any other purpose.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/MapConstraintViolationGenerator.kt
drganjoo: No, `Clippy` doesn't complain about this. But yes, I'm removing the return and the semicolon.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/UnconstrainedMapGeneratorTest.kt
```diff
@@ -162,5 +165,7 @@ class UnconstrainedMapGeneratorTest {
                 }
             }
         }
+
+        project.compileAndTest()
```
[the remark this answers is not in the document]
drganjoo: Yes for some reason we were not testing this one.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ConstrainedMapGeneratorTest.kt
```diff
@@ -150,6 +155,156 @@
...
+                        let constrained_res: Result<$rustShapeName, $rustShapeSnakeCaseName::ConstraintViolation> = map.try_into();
+                        let error = constrained_res.unwrap_err();
+                        let _error_display_works = format!("{error}");
+                        let _error_trait_is_implemented : &dyn std::error::Error = &error;
+                    """,
+                )
+            }
+
+            unitTest(
+                name = "try_constraint_enum_key",
```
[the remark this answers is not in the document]
drganjoo: Because we are testing whether the enum generated for a constraint contains the particular variant. But, I am renamed them now.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/EnumTraitValidationErrorMessage.kt
drganjoo: Yes, you have a valid point. Changing it to not escape it here.

CHANGELOG.next.toml
```diff
@@ -11,7 +11,7 @@
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
 
-[[aws-sdk-rust]]
+[[smithy-rs]]
```
rcoh, line 14: unintentional change of changelog entries

rcoh (changes requested):
please fix the changelog diff

david-perez:
Fixes #3430.

PR #3433, opened 2024-02-21

ysaito1001:
## Motivation and Context

## Description
- through a profile file with a key `s3_disable_express_session_auth` (won't be supported until https://github.com/awslabs/aws-sdk-rust/issues/1073 is addressed)
 
Something to be aware of when setting the disable option through the environment variable. The environment variable is only checked during a client construction, meaning that if a customer sets it after the client has been created the SDK will not take the environment variable value into account, i.e. the following snippet will NOT disable the S3 Express session auth:
```

// Set the env variable to true after an S3 client has been created

let _ = client
    .list_objects_v2()
    .send()
    .await;
``` 

## Testing
Added unit tests for `S3ExpressRuntimePlugin` and integration tests for verifying disabling S3 Express session auth.

----

### DIFF
--- a/aws/rust-runtime/aws-inlineable/src/s3_express.rs
+++ b/aws/rust-runtime/aws-inlineable/src/s3_express.rs
@@ -661,3 +661,124 @@ pub(crate) mod identity_provider {
         }
     }
 }
+
+/// Supporting code for S3 Express runtime plugin
+pub(crate) mod runtime_plugin {
+    use aws_smithy_runtime_api::client::runtime_plugin::RuntimePlugin;
+    use aws_smithy_types::config_bag::{CloneableLayer, FrozenLayer, Layer};
+    use aws_types::os_shim_internal::Env;
+
+    mod env {
+        pub(super) const S3_DISABLE_EXPRESS_SESSION_AUTH: &str =
+            "AWS_S3_DISABLE_EXPRESS_SESSION_AUTH";
+    }
+
+    #[derive(Debug)]
+    pub(crate) struct S3ExpressRuntimePlugin {
+        config: FrozenLayer,
+    }
+
+    impl S3ExpressRuntimePlugin {
+        pub(crate) fn new(service_config: &CloneableLayer) -> Self {
+            Self::new_with(service_config, Env::real())
+        }
+
+        fn new_with(service_config: &CloneableLayer, env: Env) -> Self {
+            let mut layer = Layer::new("S3ExpressRuntimePlugin");
+            if service_config
+                .load::<crate::config::DisableS3ExpressSessionAuth>()
+                .is_none()
+            {
+                match env.get(env::S3_DISABLE_EXPRESS_SESSION_AUTH) {
+                    Ok(value)
+                        if value.eq_ignore_ascii_case("true")
+                            || value.eq_ignore_ascii_case("false") =>
+                    {
+                        let value = value
+                            .to_lowercase()
+                            .parse::<bool>()
+                            .expect("just checked to be a bool-valued string");
+                        layer.store_or_unset(Some(crate::config::DisableS3ExpressSessionAuth(
+                            value,
+                        )));
+                    }
+                    _ => {
+                        // TODO(S3Express depending on aws-sdk-rust#1060): Transfer a value of
+                        //  `s3_disable_express_session_auth` from a profile file to `layer`
+                    }
+                }
+            }
+
+            Self {
+                config: layer.freeze(),
+            }
+        }
+    }
+
+    impl RuntimePlugin for S3ExpressRuntimePlugin {
+        fn config(&self) -> Option<FrozenLayer> {
+            Some(self.config.clone())
+        }
+    }
+
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+
+        #[test]
+        fn disable_option_set_from_service_client_should_take_the_highest_precedence() {
+            // Disable option is set from service client.
+            let mut layer = CloneableLayer::new("test");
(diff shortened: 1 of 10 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
```diff
@@ -183,6 +185,33 @@ class S3ExpressIdentityProviderConfig(codegenContext: ClientCodegenContext) : Co
...
+    override fun section(section: FluentClientSection): Writable =
+        writable {
+            when (section) {
+                is FluentClientSection.AdditionalBaseClientPlugins -> {
+                    rustTemplate(
+                        """
+                        .with_client_plugin(#{S3ExpressRuntimePlugin}::new(${section.config}.config.clone()))
+                        """,
+                        *codegenScope,
+                    )
```
rcoh, line 207: I would recommend changing the API of this section to be statements instead of chaining `.` calls onto the existing plugin builder. This is going to limiting (e.g. you can't call a method that accepts `&mut self`.

Alternatively, you could change the API to instead _solely_ contian a single client plugin and call `.with_client_plugin` at the root.

rcoh (commented):
Looks good—consider tweaking the new client plugins API.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -661,3 +661,124 @@ pub(crate) mod identity_provider {
...
+            let mut layer = Layer::new("S3ExpressRuntimePlugin");
+            if config
+                .load::<crate::config::DisableS3ExpressSessionAuth>()
+                .is_none()
+            {
+                match env.get(env::S3_DISABLE_EXPRESS_SESSION_AUTH) {
+                    Ok(value)
+                        if value.eq_ignore_ascii_case("true")
+                            || value.eq_ignore_ascii_case("false") =>
+                    {
```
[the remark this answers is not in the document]
ysaito1001: Thanks for catching this. Actually, no. Any other value other than "true" or "false" results in error (doesn't mention case-sensitiveness, though). Other SDKs behave differently, such as for Java, any value that case-insensitively does not equal to "false" is treated as "true". I think we should follow the requirement in that any value that case-insensitively does not equal to "true" or "false" should result in a runtime error.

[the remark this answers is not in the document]
ysaito1001: Thanks for caching. Actually no. If a value does not equal to "true" or "false", it should result in an error (although it does not mention case sensitivity). We should follow that and if a value is not case-sensitively equal to "true" or "false", it will be a runtime error (so "TRUE" and "FALSE" are also illegal).

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -661,3 +661,124 @@ pub(crate) mod identity_provider {
...
+            let cfg = sut.config().unwrap();
+            assert!(
+                cfg.load::<crate::config::DisableS3ExpressSessionAuth>()
+                    .unwrap()
+                    .0
+            );
+        }
+
+        // TODO(aws-sdk-rust#1073): Add a unit test that only sets `s3_disable_express_session_auth`
+        //  in a profile file
```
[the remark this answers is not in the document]
ysaito1001: Added a placeholder test in 20068b2

aws/rust-runtime/aws-inlineable/src/s3_express.rs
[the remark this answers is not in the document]
ysaito1001: Addressed in 3cdd811.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -690,18 +690,18 @@ pub(crate) mod runtime_plugin {
...
-                            .expect("just checked to be a bool-valued string");
+                            .expect("just checked to be a valid bool-valued string");
                         layer.store_or_unset(Some(crate::config::DisableS3ExpressSessionAuth(
                             value,
                         )));
                     }
+                    Ok(value) => panic!(
+                        "Must provide either `true` or `false` for an environment variable `{}`, but got {}",
+                        env::S3_DISABLE_EXPRESS_SESSION_AUTH,
+                        value),
```
jdisanti, line 704: Should we log a warning and ignore it instead?

ysaito1001: Switched to ignore and warn f2c3445. Our intention is that any unexpected value for an env var shouldn't cause an app to crash.

aws/rust-runtime/aws-inlineable/src/s3_express.rs
```diff
@@ -702,6 +702,11 @@ pub(crate) mod runtime_plugin {
                             value,
                         )));
                     }
+                    Ok(value) => {
```
jdisanti, line 705: The warning is going to prevent fallback to the profile file value once that's implemented. You can fix that later though.

jdisanti (approved):

aws/rust-runtime/aws-inlineable/src/s3_express.rs
ysaito1001: Yeah, whether specifying an unexpected value for the env variable is penalized or not (i.e. not falling back on the profile file value because you just specified the env variable) needs some discussion.

PR #3441, opened 2024-02-23
Add a builder for signing HTTP request

ysaito1001:
## Motivation and Context
https://github.com/smithy-lang/smithy-rs/pull/3388#discussion_r1480722375

## Description
This PR attempts to provide more ergonomic API for request signing. [As documented](https://docs.rs/aws-sigv4/latest/aws_sigv4/http_request/index.html#example-signing-an-http-request) in `aws_sigv4`, customers need to follow these steps after creating an `Identity` and a `SigningParams`:
```
// Convert the HTTP request into a signable request
let signable_request = SignableRequest::new(
    "GET",
    "https://some-endpoint.some-region.amazonaws.com",
    std::iter::empty(),
    SignableBody::Bytes(&[])
).expect("signable request");

let mut my_req = http::Request::new("...");
// Sign and then apply the signature to the request
let (signing_instructions, _signature) = sign(signable_request, &signing_params)?.into_parts();
signing_instructions.apply_to_request_http1x(&mut my_req);
```
This PR allows them to call `SignWith::sign_with` trait method with `SigningPackage` instead:
```
let signing_package = SigningPackage::builder()
    .signing_params(&signing_params)
    .build()
    .expect("required fields have been set");
my_req.sign_with(&signing_package);
```

## Testing

A semver failure in CI can be ignored; A public API `SigV4Signer::sign_http_request` has been removed that only existed in my feature branch.

## Checklist

----


aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
```diff
@@ -146,76 +145,47 @@ impl SigV4Signer {
...
-    /// needs to pass a configured `settings`.
+    /// This is a helper used by `(SigV4Signer as Sign).sign_http_request(...)` and will be useful
+    /// if calling code needs to pass a configured [`SigningSettings`].
     ///
-    /// TODO(S3Express): Make this method more user friendly, possibly returning a builder
-    ///  instead of taking these input parameters. The builder will have a `sign` method that
-    ///  does what this method body currently does.
+    /// If you already have an [`HttpRequest`] and a [`SigningParams`] at hand, consider using
+    /// [`SigningApplication`] directly.
     pub fn sign_http_request(
```
jdisanti, line 153: We can't change the arguments on this function since it's part of the public API.

aws/rust-runtime/aws-runtime/src/auth/signing_application.rs
```diff
@@ -0,0 +1,277 @@
...
+        &mut self,
+        payload_override: Option<SignableBody<'a>>,
+    ) -> &mut Self {
+        self.payload_override = payload_override;
+        self
+    }
+
+    #[cfg(feature = "event-stream")]
+    /// Sets the [`DeferredSignerSettings`](event_stream::DeferredSignerSettings) for the event stream signer (optional).
+    pub fn deferred_signer_settings(
```
jdisanti, line 94: Can all this event stream deferred signer stuff be made `pub(crate)`? Customers shouldn't need any of this when signing directly.

aws/rust-runtime/aws-runtime/src/auth/signing_application.rs
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    /// Sets the request (required).
+    pub fn set_request(&mut self, request: Option<&'a mut HttpRequest>) -> &mut Self {
+        self.request = request;
+        self
+    }
+
+    /// Sets the [`SigningParams`] for the builder (required).
+    pub fn signing_params(mut self, signing_params: &'a SigningParams<'_>) -> Self {
```
jdisanti, line 63: We should add an `impl From<Credentials>` for `Identity` where `Credentials` is defined to make these params easier to construct. Would be good to change the SigningParams builder to take `impl Into<Identity>` as an argument as well.

jdisanti (commented):
SigningApplication is definitely an improvement, but still seems a little non-intuitive since it is a builder pattern that has a side effect.

What do you think about creating a trait, maybe `SignWith`, that has a `sign_with` method taking the identity/settings args, and then implementing that trait for `http::Request`?

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
ysaito1001: Ah, this is a new function only existing in my feature branch. The semver failure in CI is both right and wrong; it's right because the arguments changed in this PR but wrong because it's not a semver failure w.r.t the `main` branch.

ysaito1001:
> SigningApplication is definitely an improvement, but still seems a little non-intuitive since it is a builder pattern that has a side effect.
> 
> What do you think about creating a trait, maybe `SignWith`, that has a `sign_with` method taking the identity/settings args, and then implementing that trait for `http::Request`?

That's an interesting approach 💡  Let's discuss further. I'm curious.

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
jdisanti: Oh, right. I forgot.

aws/rust-runtime/aws-runtime/src/auth/signing_application.rs
ysaito1001: Makes sense. Addressed in a81ab9b.

aws/rust-runtime/aws-runtime/src/auth/signing_application.rs
ysaito1001: Turns out `impl From<Credentials> for Identity` [already exists](https://github.com/smithy-lang/smithy-rs/blob/d42727e3315b0a8f7213c3e5f2e4ccf53b51bbe2/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs#L196-L201). Isn't updating the [SigningParams builder](https://github.com/smithy-lang/smithy-rs/blob/d42727e3315b0a8f7213c3e5f2e4ccf53b51bbe2/aws/rust-runtime/aws-sigv4/src/sign/v4.rs#L148-L149) to take `impl Into<Identity>` a breaking change with respect to `.identity(...)` and `.set_identity(...)`?

ysaito1001: Ah, I suppose no. `impl Into<Identity>` just makes use of the API more flexible so that the previous clients continue working.

rcoh:
should this target main instead? Is this related to S3 express?

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
```diff
@@ -76,12 +75,13 @@ impl SigV4Signer {
...
 
-    fn signing_params<'a>(
+    /// Creates a [`v4::SigningParams`] from the given arguments.
+    pub fn signing_params<'a>(
         settings: SigningSettings,
         identity: &'a Identity,
         operation_config: &'a SigV4OperationSigningConfig,
         request_timestamp: SystemTime,
-    ) -> Result<v4::SigningParams<'a, SigningSettings>, SigV4SigningError> {
+    ) -> Result<v4::SigningParams<'a, SigningSettings>, BoxError> {
```
jdisanti, line 84: I should have caught this in earlier reviews, but I think we should instead make `SigV4SigningError` an opaque/informative public error so that we can add methods to it in the future if we need to. See https://github.com/smithy-lang/smithy-rs/blob/main/design/src/rfcs/rfc0022_error_context_and_compatibility.md

It's pretty easy to do this. Just rename the existing error enum to "ErrorKind", and make a public struct with the previous name.

aws/rust-runtime/aws-runtime/src/sign.rs
```diff
@@ -0,0 +1,280 @@
...
+/// let signing_package = SigningPackage::builder()
+///     .signing_params(&signing_params)
+///     .build()
+///     .expect("required fields have been set");
+///
+/// let mut request = request();
+/// request.sign_with(&signing_package)
+/// # }
+/// ```
+pub trait SignWith {
```
jdisanti, line 41: What's your reasoning for putting this in aws-runtime instead of aws-sigv4?

ysaito1001: I was under the impression that moving forward, we'll be putting items in `aws-runtime-api` and `aws-runtime` crates, but not in constellation crates.

ysaito1001:
> should this target main instead? Is this related to S3 express?

It's just so happens that the S3 express feature branch was a convenient segue for this. But targeting main is actually better, since the S3 express feature branch doesn't need to hold onto these code changes. Will cherry-pick commits and create a different PR that targets main.

aws/rust-runtime/aws-runtime/src/auth/signing_application.rs
ysaito1001: Discussed offline that the suggestion is no longer relevant after the introduction of a `SignWith` trait.

aws/rust-runtime/aws-runtime/src/sign.rs
ysaito1001: I was under the impression that moving forward, we'll be putting items in `aws-runtime-api` and `aws-runtime` crates, but not in constellation crates, otherwise we can put it in `aws-sigv4` as well.

ysaito1001: I was under the impression that moving forward, we'll be putting items in `aws-runtime-api` and `aws-runtime` crates, but not in constellation crates, otherwise we can put it in `aws-sigv4` instead.

ysaito1001: `impl SignWith for HttpRequest` internally uses a module private `aws_runtime::auth::apply_signing_instructions`, so we need to move it to `aws-sigv4`.

ysaito1001:
Replaced by #3455

PR #3446, opened 2024-02-27
Add end-to-end throughput benchmarks for S3 Express

ysaito1001:
## Description
See [README](https://github.com/smithy-lang/smithy-rs/blob/73bf9b0a53f5e5615d8dc466274f5588a740ea54/aws/sdk/benchmarks/s3-express/README.md)

## Testing
Here are performance numbers measured on `Amazon Linux 2 x86_64 5.10` with `c5.4xlarge` in `us-west-2`, and benchmarks ran with `NUMBER_OF_ITERATIONS=1` to allow for quick execution

```
For 64KB objects:
PUT: [11.750 ms 12.025 ms 12.291 ms] / [67.737 ms 76.367 ms 85.531 ms]
GET: [16.071 ms 16.270 ms 16.461 ms] / [19.941 ms 22.622 ms 26.167 ms]
PUT + DELETE: [21.492 ms 22.132 ms 22.755 ms] / [75.608 ms 86.056 ms 98.329 ms]

For 1MB objects:
PUT [37.400 ms 39.130 ms 40.769 ms] / [144.30 ms 160.93 ms 180.93 ms]
GET [14.968 ms 15.193 ms 15.408 ms] / [24.872 ms 28.417 ms 32.984 ms]
PUT + DELETE: [52.106 ms 54.875 ms 57.503 ms] / [172.38 ms 185.00 ms 200.28 ms]
``` 
In each row, a tuple on the left of `/` is for execution time for an S3 Express bucket and that on the right is for a regular S3 bucket. Each tuple is a 99% confidence interval, containing `[<the lower end> <the mean> <the higher end>]`.

----


rcoh:
Is there a comparison with regular s3?

ysaito1001:
> Is there a comparison with regular s3?

You mean functionality to automatically run against both S3 express buckets and regular S3 buckets? If so, currently no. It needs to run separately depending on a bucket type.

aws/sdk/benchmarks/s3-express/README.md
```diff
@@ -0,0 +1,38 @@
+# S3 Express Benchmark
```
aajtodd, line 1: **Question**: Any reason this should not have just been made part of the existing `s3-throughput` benchmark test suite as a new type of test/input to that?

aws/sdk/benchmarks/s3-express/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This directory contains S3 Express One Zone benchmarks that measure end-to-end throughput when using the AWS Rust SDK to put, get, and delete objects to/from S3 Express One Zone buckets. We use [`Criterion`](https://github.com/bheisler/criterion.rs) for benchmarks.
+
+Performance numbers will vary depending on the benchmarking environment, but relative performance should still be accurate (i.e. regular S3 bucket vs. S3 Express bucket or comparing with a previous release of the Rust SDK).
+
+## Benchmark targets
+- `put_get_delete`: `PutObject`, `GetObject`, and `DeleteObject` using sequential invocations (20 by default) of operations across different buckets, switching buckets on every request and using both 64KB and 1MB objects.
```
aajtodd, line 8: **correctness**: I don't think we want to benchmark all three together in terms of throughput. We should separate it and probably just benchmark `put` vs `get`. 

This is important in particular because S3 express is [advertised](https://aws.amazon.com/s3/storage-classes/express-one-zone/) as

> purpose-built to deliver consistent single-digit millisecond data access 

If we combine all three operations into one in the benchmark it is harder to dive into whether put vs get is having issues on a regression.

aws/sdk/benchmarks/s3-express/benches/concurrent_put_get.rs
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+use aws_config::BehaviorVersion;
+use aws_sdk_s3::primitives::ByteStream;
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use s3_express::{confidence_level, number_of_iterations, sample_size};
+use tokio::runtime::Runtime;
+use tokio::task;
+
+pub fn concurrent_put_get(c: &mut Criterion) {
```
aajtodd, line 13: **question**: Where does this benchmark come from? What purpose is it serving (i.e. what are we measuring here)?

aws/sdk/benchmarks/s3-express/benches/put_get_delete.rs
```diff
@@ -0,0 +1,79 @@
...
+    let mut group = c.benchmark_group("put_get_delete");
+    group.sample_size(10).confidence_level(0.99);
+
+    const KB: usize = 1024;
+    for size in [64 * KB, KB * KB].iter() {
+        let object: Vec<u8> = vec![0; *size];
+        group.bench_with_input(BenchmarkId::new("size", size), size, |b, _| {
+            b.to_async(Runtime::new().unwrap()).iter(|| async {
+                for i in 0..number_of_iterations {
+                    let bucket = &buckets[i % buckets.len()];
```
aajtodd, line 41: **correctness**: This is round robin selecting a bucket each iteration which doesn't seem like what we want from trying to keep as many outside variables changing as possible when calculating the statistics.

aajtodd (changes requested):

aws/sdk/benchmarks/s3-express/README.md
ysaito1001: Because benchmark scenarios were prescribed by the internal benchmark doc, fitting it into the existing s3 benchmark suite didn't work smoothly, unfortunately.

aws/sdk/benchmarks/s3-express/README.md
ysaito1001: This combined operations, again, come from the internal benchmark doc. Let me add separate benchmarks, one for `put` and one for `get`.

aws/sdk/benchmarks/s3-express/benches/concurrent_put_get.rs
ysaito1001: It comes from the internal benchmark doc.

aws/sdk/benchmarks/s3-express/benches/put_get_delete.rs
ysaito1001: This round robin selection also comes from the internal benchmark doc.

aws/sdk/benchmarks/s3-express/README.md
ysaito1001: This combined operations, again, come from the internal benchmark doc. Let me add separate benchmarks, one for put and one for get.

ysaito1001: Added separate benchmarks in 2da2a76. I left the remaining ones; while your assertion about benchmarking combined operations is absolutely correct, I want to leave these as a proof for the internal benchmark doc that we've implemented them.

aws/sdk/benchmarks/s3-express/README.md
```diff
@@ -0,0 +1,40 @@
+# S3 Express Benchmark
```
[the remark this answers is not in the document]
ysaito1001: Great suggestion! Added in 4fda117

PR #3447, opened 2024-02-28
Replace the crates.io API with the sparse index

jdisanti:
The sparse index is preferred to the crates.io API for the checks we need, according to the [documentation](https://crates.io/data-access). When the tools were first implemented, the sparse index didn't exist, so the API was used.

----


ysaito1001 (approved):
Looks good!

PR #3450, opened 2024-02-28
Run runtime crate version audit as part of pre-commit

jdisanti:
This PR makes the runtime-versioner run via pre-commit so that devs will know if they need to version bump a runtime crate sooner than CI would otherwise tell them.

----

### DIFF
--- a/.pre-commit-hooks/runtime-versioner.sh
+++ b/.pre-commit-hooks/runtime-versioner.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+cd "$(git rev-parse --show-toplevel)/tools/ci-build/runtime-versioner" && cargo run -- audit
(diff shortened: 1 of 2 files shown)


rcoh:
since this runs every time, how long does it take to run?

.pre-commit-hooks/runtime-versioner.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+cd "$(git rev-parse --show-toplevel)/tools/ci-build/runtime-versioner" && cargo run -- audit
```
rcoh, line 8: this runs on every single runtime crate? I guess the tool itself has the crate list?

rcoh (approved):

.pre-commit-hooks/runtime-versioner.sh
jdisanti: That's correct.

jdisanti:
> since this runs every time, how long does it take to run?
It only runs when you change something in rust-runtime or aws/rust-runtime. The long part is compiling the tool. Otherwise, it's really fast.

PR #3455, opened 2024-03-01
Add a trait `SignWith` for signing HTTP requests

ysaito1001:
## Motivation and Context
https://github.com/smithy-lang/smithy-rs/pull/3388#discussion_r1480722375

## Description
This PR attempts to provide more ergonomic API for request signing. [As documented](https://docs.rs/aws-sigv4/latest/aws_sigv4/http_request/index.html#example-signing-an-http-request) in `aws_sigv4`, customers need to follow these steps after creating an `Identity` and a `SigningParams`:
```
// Convert the HTTP request into a signable request
let signable_request = SignableRequest::new(
    "GET",
    "https://some-endpoint.some-region.amazonaws.com",
    std::iter::empty(),
    SignableBody::Bytes(&[])
).expect("signable request");

let mut my_req = http::Request::new("...");
// Sign and then apply the signature to the request
let (signing_instructions, _signature) = sign(signable_request, &signing_params)?.into_parts();
signing_instructions.apply_to_request_http1x(&mut my_req);
```
This PR allows them to call `SignWith::sign_with` trait method with `SigningPackage` instead:
```
let signing_package = SigningPackage::builder()
    .signing_params(&signing_params)
    .build()
    .expect("required fields have been set");
my_req.sign_with(&signing_package);
```

## Testing

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
+++ b/aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
@@ -3,16 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-use crate::auth;
 use crate::auth::{
     extract_endpoint_auth_scheme_signing_name, extract_endpoint_auth_scheme_signing_region,
     SigV4OperationSigningConfig, SigV4SigningError,
 };
 use aws_credential_types::Credentials;
-use aws_sigv4::http_request::{
-    sign, SignableBody, SignableRequest, SigningParams, SigningSettings,
-};
-use aws_sigv4::sign::v4;
+use aws_sigv4::http_request::{SigningParams, SigningSettings};
+use aws_sigv4::sign::{v4, SignWith, SigningPackage};
 use aws_smithy_runtime_api::box_error::BoxError;
 use aws_smithy_runtime_api::client::auth::{
     AuthScheme, AuthSchemeEndpointConfig, AuthSchemeId, Sign,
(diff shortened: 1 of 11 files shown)


drganjoo (approved):

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
```diff
@@ -152,47 +158,33 @@ impl Sign for SigV4Signer {
...
-            )?;
-            sign(signable_request, &SigningParams::V4(signing_params))?
-        }
-        .into_parts();
+                .cloned(),
+        );
+
+        let (_, _signature) = request.sign_with(&signing_package.build()?)?.into_parts();
 
-        // If this is an event stream operation, set up the event stream signer
```
jdisanti, line 195: Why remove the comment?

aws/rust-runtime/aws-sigv4/src/sign.rs
```diff
@@ -3,32 +3,182 @@
...
+    /// Sign `self` with a [`SigningPackage`] and return the [`SigningOutput<SigningInstructions>`](SigningOutput)
+    /// that was used during the signing process.
+    ///
+    /// Most of the time, the return value can be ignored since it has already been applied to the
+    /// given request, but there are cases where calling code needs to have access to it even after
+    /// signing is done (e.g. deferring signing for event stream).
+    fn sign_with(
+        &mut self,
+        package: &SigningPackage<'_>,
+    ) -> Result<SigningOutput<SigningInstructions>, BoxError>;
```
jdisanti, line 63: Needs its own error type.

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
ysaito1001: Oh removal was by mistake. Will restore it.

rcoh:
Overall I think that a better API for signing is truly needed and this is directionally good. However, I think we need to work backwards from customer use cases. For example: We should work backwards from a customer who wants to sign an API Gateway request (and write the example!) and handle this that way.

If this code is needed urgently for S3 Express, we could consider putting it in `inlineable` to iterate on it.

I'm not totally sure about the trait API either. On the surface, it seems like it allows for an implementation on foreign types but I don't think it does in practice due to orphan trait rules. At the very least we should compare with other API designs.

ysaito1001:
As Russell commented, we need more iterations at a design level, and this PR will most likely be irrelevant when that design is settled. So closing this.

For S3 Express, there is [a PR ](https://github.com/smithy-lang/smithy-rs/pull/3457)created to inline `SigV4Signer::sign_http_request` into `S3ExpressSigner::sign_http_request`.

PR #3457, opened 2024-03-01
Drop `S3ExpressSigner` and override session token name via `RuntimePlugin`

ysaito1001:
## Motivation and Context
In response to https://github.com/smithy-lang/smithy-rs/pull/3455#issuecomment-1973771991, it's clear that we need to iterate more on the potential signing API.

This PR takes a different approach where `S3ExpressSigner` is dropped and a session token name override, if any, is placed into and retrieved from `ConfigBag`, which is used within `SigV4Signer::sign_http_request`.

## Testing
Existing tests in CI

A semver failure in CI can be ignored; A public API SigV4Signer::sign_http_request has been removed that only existed in the branch.

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/auth.rs
+++ b/aws/rust-runtime/aws-runtime/src/auth.rs
@@ -209,8 +209,7 @@ fn extract_field_from_endpoint_config<'a>(
         .and_then(|config| config.get(field_name))
 }
 
-/// Applies the instructions to the given `request`.
-pub fn apply_signing_instructions(
+fn apply_signing_instructions(
     instructions: SigningInstructions,
     request: &mut HttpRequest,
 ) -> Result<(), BoxError> {
--- a/aws/rust-runtime/aws-runtime/src/auth.rs
+++ b/aws/rust-runtime/aws-runtime/src/auth.rs
@@ -11,7 +11,7 @@ use aws_smithy_runtime_api::box_error::BoxError;
 use aws_smithy_runtime_api::client::auth::AuthSchemeEndpointConfig;
 use aws_smithy_runtime_api::client::identity::Identity;
 use aws_smithy_runtime_api::client::orchestrator::HttpRequest;
-use aws_smithy_types::config_bag::{Storable, StoreReplace};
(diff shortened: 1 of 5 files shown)


aws/rust-runtime/aws-runtime/src/auth.rs
```diff
@@ -199,7 +209,8 @@ fn extract_field_from_endpoint_config<'a>(
         .and_then(|config| config.get(field_name))
 }
 
-fn apply_signing_instructions(
+/// Applies the instructions to the given `request`.
+pub fn apply_signing_instructions(
```
ysaito1001, line 213: I attempted to use `signing_instructions.apply_to_request_http0x(request)` in `S3ExpressSigner::sign_http_request`, but the variable `request` was passed by a caller whose type is `&mut HttpRequest`.
Calling  `.try_into_http02x()` on `request` within `S3ExpressSigner::sign_http_request` would convert that to the underlying type, but `request` needed to continue being used, expecting it to be orchestrator's `HttpRequest`.

That's the rationale for making this function `pub`.

aws/rust-runtime/aws-runtime/src/auth.rs
```diff
@@ -75,6 +75,52 @@ impl Default for SigningOptions {
...
 
+pub(crate) type SessionTokenNameOverrideFn = Box<
+    dyn Fn(&SigningSettings, &ConfigBag) -> Result<Option<&'static str>, BoxError>
+        + Send
+        + Sync
+        + 'static,
+>;
+
+/// Custom config that provides the alternative session token name for [`SigningSettings`]
+pub struct SessionTokenNameOverride {
```
jdisanti, line 86: Nit: should prefix this with SigV4

jdisanti (approved):
Looks great!

aws/rust-runtime/aws-runtime/src/auth.rs
ysaito1001: Renamed in 0193d49

jdisanti (approved):

Issue #1925, opened 2022-10-31
Hyper Upgrade


jdisanti:
This tracks progress on the upgrade to Hyper 1.0

Many of these changes are [prototyped](https://github.com/awslabs/smithy-rs/compare/hyper-upgrade?expand=1) in this branch which can send an E2E request with RusTLS

### Upstream Changes
- [x] Backport `capture_connection` and `poison` to `hyper-util`
- [x] Create crate for 0.14 -> 1.x bridge. This allows us to use hyper-rustls and hyper-nativetls against the RC (or 1.0)

### SDK changes
- [ ] Backwards compatible solution for SegmentedBuf in `ByteStream`
- [ ] hyper::body::channel() replacement for `dvr`
- [ ] Careful upgrade + testing of `AwsChunkedBody`—the Frame API changes the flow here a little bit
- [ ] Update all tests
- [ ] Update Checksum Body
- [ ] Add `Spawn` to `AwsSmithyAsync`—bridge this with Hyper Executor (you now **must** provide an executor to create a Hyper builder)
- [ ] Event Stream Upgrade

## Testing Tasks
- [ ] Simple request/response with:
  - [ ] Native TLS 
  - [ ] Rustls
- [ ] Streaming request/response with: 
  - [ ] Native TLS 
  - [ ] Rustls
- [ ] Checksum validating response (S3 with checksums)
- [ ] Checksum calculating request (S3 with checksums)
- [ ] Unidirectional event stream operation (e.g., Transcribe Streaming and S3 Select):
  - [ ] HTTP/1.1
  - [ ] h2
  - [ ] Native TLS
  - [ ] Rustls
- [ ] Bidirectional event stream operation (Pokemon client/server test?):
  - [ ] h2
  - [ ] Native TLS
  - [ ] Rustls

jdisanti:
I spent some time working on this this week. I was able to get `aws-smithy-http` compiling with most tests passing, and some of `aws-smithy-checksums` (ended up stubbing out the checksum calculating body to move forward). My work is available on the https://github.com/awslabs/smithy-rs/tree/hyper-1x-experimentation-dec-7-2022 branch. Findings below.

Further work that is unblocked right now:
- [ ] Fix failing tests in `aws-smithy-http` after upgrade
- [ ] Reimplement checksum calculating body in `aws-smithy-checksums` to support new `Body` trait

Prerequisites before continuing:
[...]

PR #3461, opened 2024-03-05
Add Initial Support for Hyper 1.0

rcoh:
## Motivation and Context
- #1925 

## Description
This adds a minimal Hyper client, focusing on not exposing any unstable APIs. For this reason, the `Client::Builder` customization API is not exposed anymore. We do this because at some point in the future, we will likely move away from the hyper-util based Client.

The code for this was lifted directly from the Hyper 0.14 implementation but updated for new traits.

However, this does come with some new valuable pieces:
1. Support for aws-lc (no FIPS yet)
2. Support for providing a custom DNS resolver

## Testing
- E2E test with Hyper. A Canary should also be added (https://github.com/awslabs/aws-sdk-rust/issues/1089)

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-sigv4/Cargo.toml
+++ b/aws/rust-runtime/aws-sigv4/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-sigv4"
-version = "1.1.8"
+version = "1.1.9"
 authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "David Barsky <me@davidbarsky.com>"]
 description = "SigV4 signer for HTTP requests and Event Stream messages."
 edition = "2021"
@@ -28,7 +28,6 @@ hex = "0.4"
 hmac = "0.12"
 http0 = { version = "0.2", optional = true, package = "http" }
 http = { version = "1", optional = true }
-num-bigint = { version = "0.4.2", optional = true }
 once_cell = "1.8"
 p256 = { version = "0.11", features = ["ecdsa"], optional = true }
 percent-encoding = { version = "2.1", optional = true }
--- a/rust-runtime/aws-smithy-experimental/src/hyper_1_0.rs
+++ b/rust-runtime/aws-smithy-experimental/src/hyper_1_0.rs
@@ -172,7 +172,6 @@ mod build_connector {
         hyper_rustls::HttpsConnectorBuilder::new()
                .with_tls_config(
                 rustls::ClientConfig::builder_with_provider(Arc::new(restrict_ciphers(crypto_provider)))
-                    //.with_safe_default_kx_groups()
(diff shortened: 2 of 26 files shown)


rust-runtime/aws-smithy-experimental/src/hyper_1_0.rs
```diff
@@ -0,0 +1,1223 @@
...
+
+    pub(crate) fn make_tls<R>(
+        resolver: R,
+        crypto_provider: CryptoProvider,
+    ) -> HttpsConnector<HttpConnector<R>> {
+        use hyper_rustls::ConfigBuilderExt;
+        hyper_rustls::HttpsConnectorBuilder::new()
+               .with_tls_config(
+                rustls::ClientConfig::builder_with_provider(Arc::new(restrict_ciphers(crypto_provider)))
+                    //.with_safe_default_kx_groups()
```
jdisanti, line 175: ?

rust-runtime/aws-smithy-experimental/src/hyper_1_0.rs
```diff
@@ -0,0 +1,1223 @@
...
+    while let Some(err) = next {
+        if let Some(matching_err) = err.downcast_ref::<E>() {
+            return Some(matching_err);
+        }
+        next = err.source();
+    }
+    None
+}
+
+// TODO(https://github.com/awslabs/aws-sdk-rust/issues/1090): CacheKey must also include ptr equality to RuntimeComponents
```
jdisanti, line 515: I think pointer equality will end up ballooning the cached connectors into a memory leak if someone uses a config override on a operation. We'll have to come up with something else.

jdisanti (approved):
Looks great! Just a couple minor/non-blocking things.

rust-runtime/aws-smithy-experimental/src/hyper_1_0.rs
rcoh: ah yeah—that comment is badly worded. It should be the specific components of RC that are used.

rust-runtime/aws-smithy-experimental/src/hyper_1_0.rs
rcoh: ah they removed that — it is only safe now.

aws/rust-runtime/aws-sigv4/Cargo.toml
```diff
@@ -28,7 +28,6 @@ hex = "0.4"
 hmac = "0.12"
 http0 = { version = "0.2", optional = true, package = "http" }
 http = { version = "1", optional = true }
-num-bigint = { version = "0.4.2", optional = true }
```
jdisanti, line 31: Nice!

jdisanti (approved):

PR #3463, opened 2024-03-06
Updating timeout logic in the WASI HTTP client

landonxjames:
## Motivation and Context

Previously it was panicing if the timeouts were not correctly set. But one major environment where WASM/WASI is run (Node), currently doesn't support setting that timeout. So for the time being we now continue with the default timeout if setting it fails

## Description

Removed the `expect` from the `set_*_timeout` calls so that execution can continue with the default timeouts in environments where setting them is not supported. 

## Testing

This was tested locally by running a WASM binary in Node, an environment that does not yet support setting these timeouts.

## Checklist

----


landonxjames:
Hmm, I'm actually going to close this for now, although it might still be useful (but I always feel weird about silent failures like that). But it turns out the real issue I was running into was a bug in the wasm component code generator for javascript, https://github.com/bytecodealliance/jco/issues/394

PR #3464, opened 2024-03-06
bug: Fix comparison between Throughput with and without duration

iamjpotts:
## Motivation and Context

I observed a failure during a download from a bucket where the stalled stream detection complained that a throughput measurement of about 400kb over zero nanoseconds was below the minimum threshold of one byte per second.

Debug string of Err value:

```
ThroughputBelowMinimum { expected: Throughput { bytes_read: 1, per_time_elapsed: 1s }, actual: Throughput { bytes_read: 417792, per_time_elapsed: 0ns } }
```

The 417792 bytes read over a very short period of time shows that the stream is not stalled and that Err should not have been returned from the S3 api.

## Description

When a `Throughput` has a zero `Duration` but a non-zero `bytes_read` it should always compare `Ord::Greater` than any `Throughput` with a non-zero `Duration`.

## Testing

Added unit tests for `PartialEq` and `Eq` behavior.

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -28,3 +28,10 @@ message = "Make `BehaviorVersion` be future-proof by disallowing it to be constr
 references = ["aws-sdk-rust#1111", "smithy-rs#3513"]
 meta = { "breaking" = true, "tada" = false, "bug" = true, "target" = "client" }
 author = "Ten0"
+
+[[smithy-rs]]
+message = "Fixed comparison of Throughput when one or both have zero duration; prevents false positive Err for stalled stream detection"
+references = ["smithy-rs#3464"]
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client"}
+authors = ["iamjpotts"]
+
--- a/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
+++ b/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
@@ -64,20 +64,44 @@ impl Throughput {
 
 impl PartialEq for Throughput {
     fn eq(&self, other: &Self) -> bool {
-        self.bytes_per_second() == other.bytes_per_second()
+        if self.per_time_elapsed.is_zero() {
+            other.per_time_elapsed.is_zero() && self.bytes_read == other.bytes_read
+        } else {
+            self.bytes_per_second() == other.bytes_per_second()
+        }
     }
 }
 
 impl PartialOrd for Throughput {
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        self.bytes_per_second()
-            .partial_cmp(&other.bytes_per_second())
+        if self.bytes_read == 0 {
+            if other.bytes_read == 0 {
+                Some(std::cmp::Ordering::Equal)
+            } else {
+                Some(std::cmp::Ordering::Less)
+            }
+        } else if other.bytes_read == 0 {
+            Some(std::cmp::Ordering::Greater)
+        } else if self.per_time_elapsed.is_zero() {
+            if other.per_time_elapsed.is_zero() {
+                self.bytes_read.partial_cmp(&other.bytes_read)
+            } else {
+                // Zero duration of self is faster/greater than any non-zero duration throughput
+                Some(std::cmp::Ordering::Greater)
+            }
+        } else if other.per_time_elapsed.is_zero() {
+            // Zero duration of other is faster/greater than any non-zero duration throughput
+            Some(std::cmp::Ordering::Less)
+        } else {
+            self.bytes_per_second()
+                .partial_cmp(&other.bytes_per_second())
+        }
     }
 }
 
 impl fmt::Display for Throughput {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        // The default float formatting behavior will ensure the a number like 2.000 is rendered as 2
+        // The default float formatting behavior will ensure a number like 2.000 is rendered as 2
         // while a number like 0.9982107441748642 will be rendered as 0.9982107441748642. This
         // multiplication and division will truncate a float to have a precision of no greater than 3.
         // For example, 0.9982107441748642 would become 0.999. This will fail for very large floats
@@ -182,6 +206,13 @@ mod test {
     use super::{Throughput, ThroughputLogs};
     use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
+    fn zero_duration_throughput(bytes_read: u64) -> Throughput {
+        Throughput {
+            bytes_read,
+            per_time_elapsed: Duration::ZERO,
+        }
+    }
+
     #[test]
     fn test_throughput_eq() {
         let t1 = Throughput::new(1, Duration::from_secs(1));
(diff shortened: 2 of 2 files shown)


iamjpotts:
Thanks for the quick approval on the CI workflow run.

Should I fix the deprecation warning-as-error that's failing one of the CI jobs in this PR? It is not related to this bug fix.

```
error: use of deprecated associated function `chrono::NaiveDateTime::from_timestamp_opt`: use `DateTime::from_timestamp` instead
   --> aws-smithy-types-convert/src/date_time.rs:133:38
    |
133 |         match chrono::NaiveDateTime::from_timestamp_opt(self.secs(), self.subsec_nanos()) {
    |                                      ^^^^^^^^^^^^^^^^^^
    |
    = note: `-D deprecated` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(deprecated)]`

error: could not compile `aws-smithy-types-convert` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
```

iamjpotts:
Updates since first CI run:

1) Fixed cargo fmt issue
2) Added `#[allow(deprecated)]` to suppress warning for `chrono::NaiveDateTime::from_timestamp_opt`

rcoh:
can you backout the deprecation fix? I have a fix for that in another PR that replaces it with the non deprecated method. It should land this afternoon.

Let think on this fix. It seems right—I do wonder if in practice the stream _really_ was stalled though. The stalled stream protection won't kick in until after the grace period so if the time window was still 0ns that means you never got any other data.

iamjpotts:
> can you backout the deprecation fix? I have a fix for that in another PR that replaces it with the non deprecated method. It should land this afternoon.

Sure, i'll do it as soon as I see if CI passes green 

> Let think on this fix. It seems right—I do wonder if in practice the stream _really_ was stalled though. The stalled stream protection won't kick in until after the grace period so if the time window was still 0ns that means you never got any other data.

This was a part of an ETL job reading a CSV and then adding rows from a single file to another service in batches. I was tailing the log, and it was flying through the data when the "stall" error happened.

The process was using `csv_async` to read a gzip stream from `async_compression` and that stream was from `GetObjectOutput`. It would read rows up to a batch size limit, transform them, insert them into another service, and then continue reading the csv stream for more rows. The time between reading the last row in a batch and reading the first row in the next batch (from the same csv file) would be single digit seconds at worst, and typically sub-second.

iamjpotts:
Everything looks green with the chrono warning suppression included.

I can wait for your fix for that to land first, and then rebase and update my pr.

iamjpotts:
Rebased to include chrono deprecation fix and removed chrono deprecation warning suppression.

iamjpotts:
Rebased.

Also looks like https://github.com/smithy-lang/smithy-rs/pull/3469 is related.

jdisanti:
> Also looks like https://github.com/smithy-lang/smithy-rs/pull/3469 is related.

I'm working on adding stalled stream protection to request upload streams (where it's currently absent). As I've been doing that, I've been discovering a number of issues, so I'm planning to add a lot more testing. Hopefully I run into this issue here while working on that--would be great to reproduce it in a larger more comprehensive test.

iamjpotts:
Rebased to include newest additions from `main`; no other changes.

iamjpotts:
I can reproduce this issue just as easily with the recently released versions of the aws-sdk crates.

I also found I can workaround it by calling `ConfigLoader::stalled_stream_protection` and passing `StalledStreamProtectionConfig::disabled()` e.g.

```
aws_config::defaults(BehaviorVersion::v2023_11_09())
    .stalled_stream_protection(StalledStreamProtectionConfig::disabled());
```

jdisanti:
I would love to know if the issue is resolved with the changes in https://github.com/smithy-lang/smithy-rs/pull/3485

I ended up completely reworking the stalled stream protection implementation in there.

iamjpotts:
> I would love to know if the issue is resolved with the changes in #3485
> 
> I ended up completely reworking the stalled stream protection implementation in there.

I'm game.

My aws crate dependencies are:

```
aws-config = "1.1.8"
aws-sdk-s3 = "1.20"
aws-smithy-types = { version = "1.1.8", default-features = false, features = ["http-body-1-x"] }
aws-smithy-runtime-api = "1.2"
```

What is the most straightforward way to generate a compatible set of these four dependencies locally so I can reference them using a `cargo` path style dependency instead of from `crates.io` ?

jdisanti:
I went ahead and updated the aws-sdk-rust/next branch so that you don't need to generate anything yourself. You should be able to try it with these dependencies:
```
aws-config = { git = "https://github.com/awslabs/aws-sdk-rust", branch = "next" }
aws-sdk-s3 = { git = "https://github.com/awslabs/aws-sdk-rust", branch = "next" }
aws-smithy-types = { git = "https://github.com/awslabs/aws-sdk-rust", branch = "next", default-features = false, features = ["http-body-1-x"] }
aws-smithy-runtime-api = { git = "https://github.com/awslabs/aws-sdk-rust", branch = "next" }
```

iamjpotts:
Using the `next` branch I was able to run the failure scenario once without a repro of the failure (it succeeded). Previous to using `next` it would always fail, though which file it failed on was random.

> so that you don't need to generate anything yourself

Appreciate that.

I am curious though to run the failure scenario with the changes to the throughput comparison logic, but without the client-use-case change that is in `next`.

jdisanti:
> I am curious though to run the failure scenario with the changes to the throughput comparison logic, but without the client-use-case change that is in next.

Are you referring to the changes in this PR? I think for that, you'll have to generate a SDK using `./gradlew aws:sdk:assemble`. This will place generated crates in smithy-rs/aws/sdk/build/aws-sdk/sdk, and you should be able to use path dependencies in your Cargo.toml to point to them.

jdisanti:
Thanks a ton for testing that! That really helps me feel more confident about the changes. ❤️

iamjpotts:
Using a local build that includes the comparison logic change in this PR but that does not include the changes in https://github.com/smithy-lang/smithy-rs/pull/3485 also fixes the false positive stall errors about 400,000 bytes transferred in 0ns.

I even tried to force a stall by adding seven second sleeps in the loop that reads the streamed file from s3, and the stall protection was not triggered.

iamjpotts:
Rebased to current `main` as of an hour ago, which was https://github.com/smithy-lang/smithy-rs/commit/2d003567e50b117877538912d48327cafa96b2ea.

I can consistently reproduce the false-positive stream stall error using `main` and consistently get a successful outcome using the the changes in this PR.

iamjpotts:
Should I attempt to keep this PR up with `main` or do you think you will close it in lieu of another approach?

TBH I would feel a bit cautious about having stalled stream protection enabled with the current `Throughput` comparison logic.

jdisanti:
We're going to move forward with https://github.com/smithy-lang/smithy-rs/pull/3485

It's taking longer than I would like to get released due to some semver issues we uncovered, but it should hopefully be out soon.

jdisanti:
The stalled stream protection fix I had you test before [has been released now](https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2024-04-12). It's being published to crates.io right now and should be available in a matter of hours.

iamjpotts:
> The stalled stream protection fix I had you test before [has been released now](https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2024-04-12). It's being published to crates.io right now and should be available in a matter of hours.

Looks like its on `crates.io` now - thanks for the follow up.

PR #3466, opened 2024-03-06
Move `ProfileSet` to aws-runtime crate

Velfi:
I'm moving `ProfileSet` so that it can be accessed without requiring a dependency on aws-config. I did my best to move as little code as possible and to not break anything.

## Checklist

----


aws/rust-runtime/aws-runtime/Cargo.toml
```diff
@@ -33,6 +33,7 @@ tracing = "0.1"
 uuid = { version = "1" }
 
 [dev-dependencies]
+arbitrary = "1.3"
```
ysaito1001, line 36: Is `arbitrary` still used back in `aws-config`?

ysaito1001 (commented):
Looks good so far, needs to fix CI failures.

PR #3469, opened 2024-03-06
Add tick-advancing test sleep primitive

jdisanti:
This PR creates a new time/sleep primitive for testing the minimum throughput stall protection async logic. All the previous code in the `test_util` module was split out into sub-modules, and the new code is in the `test_util::tick_advance_sleep` module.

----


rust-runtime/aws-smithy-async/src/test_util/tick_advance_sleep.rs
```diff
@@ -0,0 +1,254 @@
...
+#[derive(Default, Debug)]
+struct Inner {
+    sleeps: Vec<QueuedSleep>,
+    /// Duration since `UNIX_EPOCH` that represents "now".
+    now: Duration,
+}
+
+impl Inner {
+    fn take_presenting_before(&mut self, time: Duration) -> Vec<QueuedSleep> {
+        self.sleeps.sort_by_key(|s| s.presents_at);
```
ysaito1001, line 72: Does `self.sleeps` need be a `Vec` if we need to sort elements upon `take_presenting_before` being called? Can something like `BTreeMap` be an option?

ysaito1001 (approved):
Cool functionality to add!

rcoh (approved):

rust-runtime/aws-smithy-async/src/test_util/tick_advance_sleep.rs
jdisanti: Added a comment about the collection choice in c492b4ee9ae5d0ea198dbe6979bc65fbbe920acd

rust-runtime/aws-smithy-async/src/test_util/tick_advance_sleep.rs
```diff
@@ -146,22 +151,25 @@ impl TickAdvanceTime {
...
+
+            // Make sure the time is always accurate for async code that runs
+            // after completing the sleep.
+            self.inner.get_mut().now = presenting.presents_at;
 
-            // Unblocked tasks could have queued some more sleeps within the duration,
-            // so we need to check those as well.
-            presenting = self.inner.get_mut().take_presenting_before(time);
+            // Notify the sleep, and then yield to let work blocked on that sleep to proceed
+            let _ = presenting.notify.take().unwrap().send(());
```
rcoh, line 171: so the only issue with this abstraction is that you don't have a guarantee that the code has _actually_ advanced past the sleep point (other than a "cross fingers") in `yield_now`. I suspect this MAY cause very occasionally flaky tests (or non flaky tests that rely on tokio implementation details)

rcoh (approved):

rust-runtime/aws-smithy-async/src/test_util/tick_advance_sleep.rs
```diff
@@ -0,0 +1,289 @@
...
+    ops::{Deref, DerefMut},
+    sync::{Arc, Mutex},
+    time::{Duration, SystemTime},
+};
+use tokio::sync::oneshot::Sender;
+
+#[derive(Debug)]
+struct QueuedSleep {
+    /// Duration since `UNIX_EPOCH` at which point the sleep is finished.
+    presents_at: Duration,
```
rcoh, line 59: consider `due_at` instead of `presents_at`

rust-runtime/aws-smithy-async/src/test_util/tick_advance_sleep.rs
jdisanti: Yeah, it does break down for very large numbers of sleeps. For example, I tried having 100 tasks blocked on 100 sleeps, and not all the tasks were visited after unblocking the 100 sleeps before the TickAdvanceTime proceeded with the next sleep.

In practice though, I haven't observed any issues with real world tests yet. It would be cool for tests to be able to wait until Tokio's task queue is completely blocked though.

PR #3470, opened 2024-03-06
Upgrade Smithy to 1.45

jdisanti:


jdisanti:
There's something weird going on with the `RestJsonZeroAndFalseQueryValues` (which was supposed to be fixed in https://github.com/smithy-lang/smithy/pull/2132). I noticed I forgot to remove it from the deny list in this PR, and when removing it, the test fails to pass with:

```
failures:

---- operation::server_all_query_string_types_test::rest_json_zero_and_false_query_values_request stdout ----
thread 'operation::server_all_query_string_types_test::rest_json_zero_and_false_query_values_request' panicked at rest_json/rust-server-codegen/src/operation.rs:1229:25:
assertion failed: `(left == right)`: Unexpected value for `query_params_map_of_string_list`

Diff < left / right > :
 Some(
     {
<        "Boolean": [
>        "queryBoolean": [
             "false",
         ],
<        "Integer": [
>        "queryInteger": [
             "0",
         ],
     },
 )


failures:
    operation::server_all_query_string_types_test::rest_json_zero_and_false_query_values_request

test result: FAILED. 724 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.12s
```

I think I fixed the test incorrectly, but now I'm puzzled why the fixed test in smithy-rs didn't break. I guess it must not be running for some reason.

jdisanti:
It appears the codegen-server-test subproject doesn't regenerate any code when changing the rest-json-extras model. That's unfortunate.

PR #3472, opened 2024-03-07
Fix `canary-wasm` not being able to locate generated crates

ysaito1001:
## Motivation and Context
Fixes the issue of `canary-wasm` not being able to locate generated crates in our release pipeline

## Description
The `canary-wasm` crate has [Cargo.toml](https://github.com/smithy-lang/smithy-rs/blob/main/tools/ci-cdk/canary-wasm/Cargo.toml) checked-in, whose SDK dependencies assume path locations with respect to build artifacts within the `smithy-rs` repository. However, those locations can be different in our release pipeline (the same reason why there's no manifest file checked-in for [canary-lambda](https://github.com/smithy-lang/smithy-rs/tree/main/tools/ci-cdk/canary-lambda) and it needs to be generated on the fly by `canary-runner`'s `build-bundle` subcommand).

To fix this, `canary-runner` now generates the manifest for `canary-wasm`.

## Testing
- [x] Added unit tests for `canary-runner`
- [x] Verified in our release pipeline that canary built and passed

----


tools/ci-cdk/canary-wasm/Cargo.toml
```diff
@@ -1,25 +0,0 @@
-
-# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jdisanti, line 2: These deleted Cargo.* files should be added to the .gitignore so they don't get added back unintentionally.

jdisanti (approved):

tools/ci-cdk/canary-wasm/Cargo.toml
ysaito1001: Added in bedff65

jdisanti (approved):

PR #3485, opened 2024-03-13
Overhaul stalled stream protection and add upload support

jdisanti:
This PR overhauls the existing stalled stream protection with a new algorithm, and also adds support for minimum throughput on upload streams. The new algorithm adds support for differentiating between the user or the server causing the stall, and not timing out if it's the user causing the stall. This will fix timeout issues when a customer makes remote service calls in between streaming pieces of information.

----


jdisanti:
Results of running the [stalled stream protection performance test](https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-runtime/tests/stalled_stream_performance.rs) in release mode.

```
Old impl without:  577.0 ± 2.9 ms
Old impl with:     584.1 ± 5.6 ms

New impl without:  577.1 ± 3.2 ms
New impl with:     583.3 ± 3.1 ms
```

These numbers are across six separate runs of the performance test each, averaged, and with a standard deviation applied. The performance is roughly the same, with maybe a miniscule amount of improvement.

The new implementation uses 80 bytes of memory for the log instead of 1 kilobyte though.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,264 @@ impl From<(u64, Duration)> for Throughput {
...
 }
 
-#[derive(Clone)]
-pub(super) struct ThroughputLogs {
-    max_length: usize,
-    inner: VecDeque<(SystemTime, u64)>,
-    bytes_processed: u64,
+/// Cell in a linear grid that represents a small chunk of time.
+#[derive(Copy, Clone, Debug)]
+enum Cell {
```
Velfi, line 102: _nitpick_

Should we avoid calling this `Cell` because of the connotations that that has for Rust programmers? Typically `Cell` means some sort of pointer type.

As an alternative, I offer the equally meaningless `Bin`

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,264 @@ impl From<(u64, Duration)> for Throughput {
...
+    /// Number of "bytes transferred" cells.
+    transferred: usize,
+    /// Number of "pending" cells.
+    pending: usize,
+}
+
+/// Underlying stack-allocated linear grid buffer for tracking
+/// throughput events for [`ThroughputLogs`].
+#[derive(Copy, Clone, Debug)]
+struct LogBuffer<const N: usize> {
```
Velfi, line 159: Is this our first usage of a GAT? If so, 🎊

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,264 @@ impl From<(u64, Duration)> for Throughput {
...
         }
     }
 
-    pub(super) fn push(&mut self, throughput: (SystemTime, u64)) {
-        // When the number of logs exceeds the max length, toss the oldest log.
-        if self.inner.len() == self.max_length {
-            self.bytes_processed -= self.inner.pop_front().map(|(_, sz)| sz).unwrap_or_default();
+    /// Mutably returns the tail of the buffer.
+    ///
+    /// The buffer MUST have at least one cell it it before this is called.
```
Velfi, line 177:
```suggestion
    /// ## Panics
    ///
    /// The buffer MUST have at least one cell it it before this is called or it will panic.
```

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,264 @@ impl From<(u64, Duration)> for Throughput {
...
+/// of events despite only receiving two: the third is "no polling". The poll
+/// functions cannot know when they're not being polled, so the log examines gaps
+/// in the event history to know when no polling took place.
+///
+/// The event logging is simplified down to a linear 10-cell grid, which each
+/// cell representing 1/10th the total time window. When an event is pushed,
+/// it is either merged into the current tail cell, or all the cells are rotated
+/// left to create a new empty tail cell, and then it is merged into that one.
+#[derive(Clone, Debug)]
+pub(super) struct ThroughputLogs {
```
Velfi, line 265: Great docs! These are really first class John.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,264 @@ impl From<(u64, Duration)> for Throughput {
...
+            _ => ThroughputReport::Incomplete,
+        }
     }
 }
 
 #[cfg(test)]
 mod test {
-    use super::{Throughput, ThroughputLogs};
-    use std::time::{Duration, SystemTime, UNIX_EPOCH};
+    use super::*;
```
Velfi, line 356: lol I'm always making these explicit so that we don't import things in the parent scope that only get used in the child scope. We're covering each other's paint.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,264 @@ impl From<(u64, Duration)> for Throughput {
...
+            transferred,
+            pending,
+        } = self.buffer.counts();
+
+        let bytes = self.buffer.bytes_transferred();
+        let time = self.resolution * (BUFFER_SIZE - empty) as u32;
+        let throughput = Throughput::new(bytes, time);
+
+        let half = BUFFER_SIZE / 2;
+        if empty >= half {
```
Velfi, line 342: Consider adding a comment saying why this mean the report is incomplete.

> If the buffer is less than half full of logs, then the report is considered incomplete.
> This is because...

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -192,92 +366,143 @@ mod test {
...
+        let start = SystemTime::UNIX_EPOCH;
+        let mut logs = ThroughputLogs::new(Duration::from_secs(1), start);
+
+        logs.push_pending(start + Duration::from_millis(50));
+        logs.push_pending(start + Duration::from_millis(150));
+        logs.push_pending(start + Duration::from_millis(250));
+        logs.push_pending(start + Duration::from_millis(450));
+        logs.push_pending(start + Duration::from_millis(650));
+        logs.push_pending(start + Duration::from_millis(750));
+        logs.push_pending(start + Duration::from_millis(850));
```
Velfi, line 503: alternatively,
```suggestion
        for i in (50..=850).step_by(100) {
                logs.push_pending(start + Duration::from_millis(i));
        }
```

Velfi: I didn't test this code so don't hit that commit button

rust-runtime/aws-smithy-runtime/tests/stalled_stream_common.rs
```diff
@@ -0,0 +1,113 @@
...
+    future::poll_fn,
+    mem,
+    pin::Pin,
+    sync::{Arc, Mutex},
+    task::{Context, Poll},
+    time::Duration,
+};
+pub use tracing::{info, Instrument as _};
+
+/// No really, it's 42 bytes long... super neat
```
Velfi, line 57: 😆

Velfi (approved):

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
jdisanti: I don't want to think about imports when writing tests 😆

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
jdisanti: 350 is skipped, so it's not quite equivalent. Perhaps it deserves a comment.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
jdisanti: Now that I think about it, I'm not really sure why I chose to compare against half. If any of the cells are empty, it should be incomplete. Revising that...

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
rcoh: assuming you mean "const generic", I think yes!

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
```diff
@@ -3,14 +3,58 @@
...
+        if violated {
+            tracing::debug!(
+                "current throughput: {throughput} is below minimum: {minimum_throughput}"
+            );
+        }
+        (violated, throughput)
+    }
+}
+
+impl<B> Body for MinimumThroughputDownloadBody<B>
```
jdisanti, line 57: Need to make sure size_hint is passed through.

aws/sdk/integration-tests/s3/tests/stalled-stream-protection.rs
```diff
@@ -4,58 +4,118 @@
...
+    // We spawn a faulty server that will stop all request processing after reading half of the request body.
     let (server, server_addr) = start_faulty_upload_server().await;
     let _ = tokio::spawn(server);
 
     let conf = Config::builder()
         .credentials_provider(Credentials::for_tests())
         .region(Region::new("us-east-1"))
         .endpoint_url(format!("http://{server_addr}"))
-        // .stalled_stream_protection(StalledStreamProtectionConfig::enabled().build())
+        .stalled_stream_protection(StalledStreamProtectionConfig::enabled().build())
```
rcoh, line 98: by default is it not enabled for upload? why is this required?

aws/sdk/integration-tests/s3/tests/stalled-stream-protection.rs
```diff
@@ -4,58 +4,118 @@
...
+
+    fn poll_data(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
+        loop {
+            let mut state = SlowBodyState::Taken;
+            std::mem::swap(&mut state, &mut self.state);
+            match state {
+                SlowBodyState::Wait(mut fut) => match fut.as_mut().poll(cx) {
```
rcoh, line 57: recommend using http_body 1.0 for bodies going forward

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/StalledStreamProtectionConfigCustomization.kt
```diff
@@ -124,7 +124,7 @@ class StalledStreamProtectionOperationCustomization(
                         // we can't count on hyper to poll a request body on wake.
                         rustTemplate(
                             """
-                            #{StalledStreamProtectionInterceptor}::new(#{Kind}::ResponseBody)
```
rcoh, line 127: do you think we need a BMV to enable it on request bodies?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
```diff
@@ -114,288 +145,45 @@ where
...
-                    "response succeeded instead of returning the expected error '{expected_err}'"
-                )
-            }
-            Err(actual_err) => {
-                assert_eq!(
-                    expected_err.to_string(),
-                    // We need to source this so that we don't get the streaming error it's wrapped in.
-                    actual_err.source().unwrap().to_string()
-                );
+    fn poll_data(
```
rcoh, line 155: could consider using http_body 1.0 here as well

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
rcoh: yeah, we may want an integration test that verifies that size hint and end-stream hint go all the way up and down the stack so we don't have any bodies that forget to delegate them.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
```diff
@@ -3,14 +3,58 @@
...
 use std::future::Future;
 use std::pin::{pin, Pin};
 use std::task::{Context, Poll};
 
-impl<B> Body for MinimumThroughputBody<B>
+const ZERO_THROUGHPUT: Throughput = Throughput::new_bytes_per_second(0);
+
+// Helper trait for interpretting the throughput report.
+trait DownloadReport {
+    fn minimum_throughput_violated(self, minimum_throughput: Throughput) -> (bool, Throughput);
```
rcoh, line 20: may want to consider returning an enum instead of a bool

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
```diff
@@ -46,44 +91,27 @@ where
...
-            .unwrap_or_else(|| this.async_sleep.sleep(this.options.check_interval()));
+            .unwrap_or_else(|| this.async_sleep.sleep(*this.resolution));
         if let Poll::Ready(()) = pin!(&mut sleep_fut).poll(cx) {
             tracing::trace!("sleep future triggered—triggering a wakeup");
             // Whenever the sleep future expires, we replace it.
-            sleep_fut = this.async_sleep.sleep(this.options.check_interval());
+            sleep_fut = this.async_sleep.sleep(*this.resolution);
 
             // We also schedule a wake up for current task to ensure that
             // it gets polled at least one more time.
```
rcoh, line 101: So, I would have thought sleep_fut wasn't required anymore since we are using the ThroughputCheckFuture externally. But if I remove it, download_stalls and download_too_slow fail. Do you understand why? 

Maybe it's just a matter of settings? I wonder if we could refactor that so we can remove that complexity from here.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
```diff
@@ -114,288 +145,45 @@ where
...
-        let (res, time_source, async_sleep) =
-            eight_byte_per_second_stream_with_minimum_throughput_timeout(minimum_throughput);
-        match res.await {
-            Ok(res) => {
-                assert_eq!(255.0, time_source.seconds_since_unix_epoch());
-                assert_eq!(Duration::from_secs(255), async_sleep.total_duration());
-                assert_eq!(*EXPECTED_BYTES, res.to_vec());
+            Poll::Pending => {
+                tracing::trace!("received poll pending");
+                this.throughput.push_pending(now);
```
rcoh, line 174: seems like we could slightly refactor the API to be something like:

`this.throughput.record_poll(poll_result)`

```
fn record_poll(&mut self, p: Poll<Bytes>) -> Poll<Bytes>
```

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/options.rs
```diff
@@ -136,19 +129,26 @@ impl MinimumThroughputBodyOptionsBuilder {
...
+    pub fn check_interval(self, _check_interval: Duration) -> Self {
         self
     }
 
-    /// Set the rate at which throughput is checked.
-    ///
-    /// Defaults to 1 second.
-    pub fn set_check_interval(&mut self, check_interval: Option<Duration>) -> &mut Self {
-        self.check_interval = check_interval;
+    /// No longer used.
```
rcoh, line 138: deprecation should probably refer to `check_window` since that appears to be the replacement?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,269 @@ impl From<(u64, Duration)> for Throughput {
...
+    NoPolling,
+
+    /// This many bytes were transferred during this bin.
+    TransferredBytes(u64),
+
+    /// The user/remote was not providing/consuming data fast enough during this bin.
+    ///
+    /// The number is the number of bytes transferred, if this replaced TransferredBytes.
+    Pending(u64),
+}
```
rcoh, line 116: So a bin transitions from `TransferredBytes` into `Pending` if the last thing that happened in the bin was `Poll::Pending`?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,269 @@ impl From<(u64, Duration)> for Throughput {
...
+    Incomplete,
+    /// The stream hasn't been polled for most of this time window.
+    NoPolling,
+    /// The stream has been waiting for most of the time window.
+    Pending,
+    /// The stream transferred this amount of throughput during the time window.
+    Transferred(Throughput),
+}
+
+const BUFFER_SIZE: usize = 10;
```
rcoh, line 248: this is the number of throughput windows?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,269 @@ impl From<(u64, Duration)> for Throughput {
...
+            buffer: LogBuffer::new(),
         }
     }
 
-    pub(super) fn push(&mut self, throughput: (SystemTime, u64)) {
-        // When the number of logs exceeds the max length, toss the oldest log.
-        if self.inner.len() == self.max_length {
-            self.bytes_processed -= self.inner.pop_front().map(|(_, sz)| sz).unwrap_or_default();
-        }
+    /// Returns the resolution at which events are logged at.
```
rcoh, line 291: need an example of some sort to understand what resolution means in this context. Is it the width of a bin?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,269 @@ impl From<(u64, Duration)> for Throughput {
...
+/// Underlying stack-allocated linear grid buffer for tracking
+/// throughput events for [`ThroughputLogs`].
+#[derive(Copy, Clone, Debug)]
+struct LogBuffer<const N: usize> {
+    entries: [Bin; N],
+    // The length only needs to exist so that the `fill_gaps` function
+    // can differentiate between `Empty` due to there not having been enough
+    // time to establish a full buffer worth of data vs. `Empty` due to a
+    // polling gap. Once the length reaches N, it will never change again.
+    length: usize,
```
rcoh, line 165: what if you add a third bin option `Untouched` or something? Then we can remove this length logic which makes me a little uncomfortable due to off-by-one risk

rcoh (commented):
I got through about half of it, will dig in more later

ysaito1001 (approved):
This is one of those wow PRs, excellent work!

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
ysaito1001: I think it is. A resolution is the amount of time each cell is responsible for.

rust-runtime/aws-smithy-runtime/tests/stalled_stream_common.rs
```diff
@@ -0,0 +1,113 @@
...
+    receiver: tokio::sync::mpsc::Receiver<Bytes>,
+}
+impl http_body_0_4::Body for ChannelBody {
+    type Data = Bytes;
+    type Error = Infallible;
+
+    fn poll_data(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
```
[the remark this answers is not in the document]
jdisanti: I don't think we can yet?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
jdisanti: This is the download body, which _does not_ use ThroughputCheckFuture since it's not the orchestrator that's waiting on it. Rather, the customer is waiting on it in their code directly since it is the response body.

rust-runtime/aws-smithy-runtime/src/test_util/capture_test_logs.rs
```diff
@@ -24,14 +24,25 @@ pub struct LogCaptureGuard(#[allow(dead_code)] DefaultGuard);
 pub fn capture_test_logs() -> (LogCaptureGuard, Rx) {
     // it may be helpful to upstream this at some point
     let (mut writer, rx) = Tee::stdout();
-    if env::var("VERBOSE_TEST_LOGS").is_ok() {
-        eprintln!("Enabled verbose test logging.");
+    let (enabled, level) = match env::var("VERBOSE_TEST_LOGS").ok().as_deref() {
+        Some("debug") => (true, Level::DEBUG),
```
rcoh, line 28: it would be nice if you could "loud" this from the API instead of needing to use an environment variable.

rust-runtime/aws-smithy-runtime/src/test_util/capture_test_logs.rs
```diff
@@ -24,14 +24,25 @@ pub struct LogCaptureGuard(#[allow(dead_code)] DefaultGuard);
...
     } else {
-        eprintln!("To see full logs from this test set VERBOSE_TEST_LOGS=true");
+        eprintln!(
+            "To see full logs from this test set VERBOSE_TEST_LOGS=true \
+                (or to a log level, e.g., trace, debug, info, etc)"
+        );
     }
     let subscriber = tracing_subscriber::fmt()
-        .with_max_level(Level::TRACE)
+        .with_max_level(level)
```
rcoh, line 45: should we just pass the contents to `with_env_filter` instead? Then you don't need to parse match etc.

rust-runtime/aws-smithy-runtime/tests/stalled_stream_common.rs
```diff
@@ -0,0 +1,115 @@
...
+    mem,
+    pin::Pin,
+    sync::{Arc, Mutex},
+    task::{Context, Poll},
+    time::Duration,
+};
+pub use tracing::{info, Instrument as _};
+
+/// No really, it's 42 bytes long... super neat
+pub const NEAT_DATA: Bytes = Bytes::from_static(b"some really neat data");
```
rcoh, line 60: I assume `NEAT` here just really means "test data"?

rust-runtime/aws-smithy-runtime/tests/stalled_stream_common.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+impl http_body_0_4::Body for ChannelBody {
+    type Data = Bytes;
+    type Error = Infallible;
+
+    fn poll_data(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
+        match self.receiver.poll_recv(cx) {
```
rcoh, line 98: we probably need a module of test bodies. we have a bunch of private ones everywhere

CHANGELOG.next.toml
```diff
@@ -22,3 +22,15 @@ message = "`DefaultS3ExpressIdentityProvider` now uses `BehaviorVersion` threade
 references = ["smithy-rs#3478"]
 meta = { "breaking" = false, "bug" = true, "tada" = false }
 author = "ysaito1001"
+
+[[smithy-rs]]
+message = "Stalled stream protection now supports request upload streams, and existing download stream protection no longer triggers when user-code is doing something other than downloading in the middle of the stream (for example, making a HTTP request to another service for each row in CSV file download from S3)."
```
rcoh, line 27: this is one super long sentence

aws/sdk/integration-tests/s3/tests/stalled-stream-protection.rs
jdisanti: Just being explicit.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/StalledStreamProtectionConfigCustomization.kt
jdisanti: Yeah, I'm not sure. I guess it depends on whether or not we consider this a bug fix?

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/options.rs
jdisanti: I'm not sure I want to make check_window actually configurable or not yet. I was leaning towards not exposing it until a use-case came up for it.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
jdisanti: Yeah, `Pending` is the highest priority in the `Bin::merge` function below.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,269 @@ impl From<(u64, Duration)> for Throughput {
...
+    fn merge(&mut self, other: Bin) {
+        use Bin::*;
+        // Assign values based on this priority order (highest priority higher up):
+        //   1. Pending
+        //   2. TransferredBytes
+        //   3. NoPolling
+        //   4. Empty
+        *self = match (other, *self) {
+            (Pending(other), this) => Pending(other + this.bytes()),
+            (TransferredBytes(other), this) => TransferredBytes(other + this.bytes()),
```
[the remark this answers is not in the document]
jdisanti: In your hypothetical case, it depends on what `other` is. `this` doesn't have a lot of bearing on what is chosen until the last few match arms. But the gist of it is what the comment above the match says: Pending > TransferredBytes > NoPolling > Empty.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
jdisanti: I had that approach originally, but it only moves the problem. Going that way, you have to track the time that the tracking started so that you know which cells can be flipped to NoPolling.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
jdisanti: Should rename this to `BIN_COUNT`.

rust-runtime/aws-smithy-runtime/tests/stalled_stream_common.rs
jdisanti: Pretty much 😄

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/options.rs
jdisanti: Offline discussion: Improve deprecation messaging to indicate that this configuration is not currently available and to file an issue if it's needed.

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs
```diff
@@ -97,90 +97,269 @@ impl From<(u64, Duration)> for Throughput {
...
+impl Bin {
+    fn merge(&mut self, other: Bin) {
+        use Bin::*;
+        // Assign values based on this priority order (highest priority higher up):
+        //   1. Pending
+        //   2. TransferredBytes
+        //   3. NoPolling
+        //   4. Empty
+        *self = match (other, *self) {
+            (Pending(other), this) => Pending(other + this.bytes()),
```
jdisanti, line 126:
```
(other, Pending(this)) => Pending(other.bytes() + this),
```

rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs
jdisanti: Fixed and integration tested in 73635049dc5bc5fabb603292fec92d5d803e9706

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/StalledStreamProtectionConfigCustomization.kt
jdisanti: Filed https://github.com/smithy-lang/smithy-rs/issues/3510 to track the BMV change.

rust-runtime/aws-smithy-runtime/src/test_util/capture_test_logs.rs
jdisanti: It's not a bad idea, but as I was attempting it, I realized we're kind of overloading the purpose of this test utility. We should probably add a separate `show_test_logs()` function that does the env filtering.

rust-runtime/aws-smithy-runtime/tests/stalled_stream_common.rs
jdisanti: Punting on this in this PR.

CHANGELOG.next.toml
jdisanti: Split into two changelog entries.

CHANGELOG.next.toml
```diff
@@ -9,4 +9,28 @@
...
+authors = ["jdisanti"]
+
+[[aws-sdk-rust]]
+message = "Stalled stream protection now supports request upload streams."
+references = ["smithy-rs#3485"]
+meta = { "breaking" = false, "tada" = true, "bug" = false }
+author = "jdisanti"
+
+[[smithy-rs]]
+message = "Stalled stream protection for download streams no longer triggers when user-code is taking a significant amount of time doing something other than downloading in the middle of the stream. As an example, previously, making a HTTP request to another service for each row in CSV file download from S3 would result in a stalled stream timeout."
```
rcoh, line 27:
```suggestion
message = "Stalled stream protection on downloads will now only trigger if the upstream source was too slow. Previously, stalled stream protection could be erroneously triggered if the customer was slowly consuming the stream slower than the minimum speed limit."
```

aws/sdk/integration-tests/s3/tests/body_size_hint.rs
```diff
@@ -0,0 +1,122 @@
...
+        Some(test_body_content.len() as u64),
+        test_body.size_hint().exact(),
+        "pre-condition check"
+    );
+
+    let http_client = TestClient {
+        response_body: Arc::new(Mutex::new(Some(test_body))),
+        ..Default::default()
+    };
+    let sdk_config = SdkConfig::builder()
```
rcoh, line 65: you can avoid the need to make a entire new struct by using `infallible_client_fn`

aws/sdk/integration-tests/s3/tests/body_size_hint.rs
```diff
@@ -0,0 +1,122 @@
...
+        .build();
+    let client = Client::new(&sdk_config);
+    let body = ByteStream::from_static(test_body_content);
+    assert_eq!(
+        (
+            test_body_content.len() as u64,
+            Some(test_body_content.len() as u64),
+        ),
+        body.size_hint(),
+        "pre-condition check"
```
rcoh, line 106: I think this can use `capture_request` (no need for adhoc struct)

rust-runtime/aws-smithy-runtime/src/test_util/capture_test_logs.rs
```diff
@@ -12,7 +12,37 @@ use tracing_subscriber::fmt::TestWriter;
...
+/// Setting it to "1" or "true" will enable trace logging.
+#[must_use]
+pub fn show_test_logs() -> LogCaptureGuard {
+    let (mut writer, _rx) = Tee::stdout();
+    let env_var = env::var("VERBOSE_TEST_LOGS").ok();
+    let env_filter = match env_var.as_deref() {
+        Some("true") | Some("1") => Some("trace"),
+        Some(filter) => Some(filter),
+        None => None,
+    };
```
rcoh, line 31: I think we should add a way to loudly show test logs without having to set an env var for convenience (e.g. when running tests w/ intellij)

Could make a `loud` argument to the variable, or add a boolean arg to the function?

Or keep `show_test_logs` as the loud one?

CHANGELOG.next.toml
```diff
@@ -12,13 +12,25 @@
...
+references = ["smithy-rs#3485"]
+meta = { "breaking" = false, "tada" = false, "bug" = true }
+authors = ["jdisanti"]
+
+[[aws-sdk-rust]]
```
rcoh, line 32: the changelog entries should also discuss _how_ to enable protection for uploads before it becomes a new BMV

rcoh (approved):

PR #3488, opened 2024-03-13
Add support for env-defined endpoint URLs

Velfi:
## Motivation and Context

Part of the endpoint config work I'm doing

## Description

This change does two things:
- add support for setting an endpoint URL from the env or profile file
- add support for ignoring endpoint URLs sourced from the env and profile file

## Testing

I wrote many unit tests.

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/default_provider/endpoint_url.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider/endpoint_url.rs
@@ -22,7 +22,7 @@ mod profile_key {
 /// 1. The environment variable `AWS_ENDPOINT_URL=http://localhost`
 /// 2. The profile key `endpoint_url=http://localhost`
 ///
-/// If invalid values are found, the provider will return None and an error will be logcged.
+/// If invalid values are found, the provider will return None and an error will be logged.
 pub async fn endpoint_url_provider(provider_config: &ProviderConfig) -> Option<String> {
     StandardProperty::new()
         .env(env::ENDPOINT_URL)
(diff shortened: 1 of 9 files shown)


rust-runtime/inlineable/Cargo.toml
```diff
@@ -32,7 +32,7 @@ once_cell = "1.16.0"
 percent-encoding = "2.2.0"
 pin-project-lite = "0.2"
 regex-lite = "0.1.5"
-url = "2.2.2"
+url = "2.5.0"
```
ysaito1001, line 35: Do we need to update version number [here](https://github.com/smithy-lang/smithy-rs/blob/2392a24ed611447fb4789d93873bf19e9e2a8e83/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/CargoDependency.kt#L249) as well?

ysaito1001 (approved):

rust-runtime/inlineable/Cargo.toml
jdisanti: What's the reason for upgrading it at all?

aws/rust-runtime/aws-config/src/default_provider/endpoint_url.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub(super) const ENDPOINT_URL: &str = "endpoint_url";
+}
+
+/// Load the value for an endpoint URL
+///
+/// This checks the following sources:
+/// 1. The environment variable `AWS_ENDPOINT_URL=http://localhost`
+/// 2. The profile key `endpoint_url=http://localhost`
+///
+/// If invalid values are found, the provider will return None and an error will be logcged.
```
jdisanti, line 25:
```suggestion
/// If invalid values are found, the provider will return None and an error will be logged.
```

jdisanti (approved):
LGTM

rust-runtime/inlineable/Cargo.toml
ysaito1001: To be in sync with what we specify in `inlineable/Cargo.toml`?

PR #3492, opened 2024-03-14
Fix invalid XML root in S3 GetObjectRetention

jdisanti:
This PR fixes the invalid XML root issue reported in https://github.com/awslabs/aws-sdk-rust/issues/1065.

----


ysaito1001 (approved):

rcoh:
The upstream fix for this will ship Monday

Issue #2523, opened 2023-03-31
Update Pokémon Service to use Smithy IDL v2


david-perez:
Smithy IDL v2 has long been released and there's good support for it in the server generator. Our example Pokémon Service should reflect standard best practices; we should give a good example and use the latest IDL.

timClicks:
We should also add some complexity to the service to demonstrate some more features.

david-perez:
@timClicks What things did you have in mind? We showcase pretty much everything we support: TLS, Lambda integration, data and event streaming, usual request-reply operations, sensitive data...

PR #3494, opened 2024-03-15
Update Pokémon models to Smithy IDL v2

djedward:
## Motivation and Context
Resolves [issue 2523](https://github.com/smithy-lang/smithy-rs/issues/2523) to migrate to Smithy IDL v2

## Description

I followed the [Smithy Migration Guide](https://smithy.io/2.0/guides/model-translations/migrating-idl-1-to-2.html) to convert the models. One change was that by moving the inputs/outputs into the operation, it did result in a renaming of the CapturePokemon structures. They went from CapturePokemonEvents(Input|Output) to CapturePokemon(Input|Output). From a brief look at the commit history, it seems the API used to be named different and the inputs/outputs weren't renamed at that time. So I decided to have them match the normal naming convention (it didn't seem to break any tests).

## Testing
As this was my first commit, I ran almost every target I could discover, but I think the key ones were:
* codegen-server-test
* codegen-client-test
* examples: `make test`

## Checklist

----

### DIFF
--- a/codegen-core/common-test-models/pokemon-awsjson.smithy
+++ b/codegen-core/common-test-models/pokemon-awsjson.smithy
@@ -1,4 +1,4 @@
-$version: "1.0"
+$version: "2"
 
 // TODO(https://github.com/smithy-lang/smithy-rs/issues/2215)
 // This is a temporary model to test AwsJson 1.0 with @streaming.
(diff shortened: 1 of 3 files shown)


codegen-core/common-test-models/pokemon.smithy
```diff
@@ -14,135 +14,131 @@ use com.aws.example#ResourceNotFoundException
...
+    masterball_unsuccessful: MasterBallUnsuccessful
 }
 
 structure CapturingEvent {
     @eventPayload
-    payload: CapturingPayload,
+    payload: CapturingPayload
 }
 
 structure CapturingPayload {
```
drganjoo, line 105: nit: Should we use [target elision syntax](https://smithy.io/2.0/guides/model-translations/migrating-idl-1-to-2.html#use-the-target-elision-syntax-sugar-to-reduce-boilerplate) to clarify that the `name` refers to pokemon's name?

codegen-core/common-test-models/pokemon.smithy
```diff
@@ -14,135 +14,131 @@ use com.aws.example#ResourceNotFoundException
 @title("Pokémon Service")
 @restJson1
 service PokemonService {
-    version: "2021-12-01",
-    resources: [PokemonSpecies, Storage],
+    version: "2021-12-01"
```
drganjoo, line 17: nit: Should we change this too?

drganjoo (approved):

codegen-core/common-test-models/pokemon-awsjson.smithy
```diff
@@ -17,80 +17,78 @@ use com.aws.example#CheckHealth
 @title("Pokémon Service")
 @awsJson1_0
 service PokemonService {
-    version: "2021-12-01",
+    version: "2021-12-01"
```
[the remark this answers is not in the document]
djedward: I'm so used to never touching these. Since this is an example, it probably doesn't hurt to move it forward.

codegen-core/common-test-models/pokemon.smithy
djedward: It felt like the feature complicated the model, but since it is an example/sample test, probably good to use as many features as possible. I'll get this updated

djedward:
I believe I've addressed the 3 comments now and resynced the branch.

rcoh (approved):

david-perez (approved):

PR #3497, opened 2024-03-19
Update GitHub actions to fix deprecation warning

Velfi:
Our workflows used outdated actions and that was causing deprecation warnings for Node v16. Those annoyed me so I updated everything. 

----


jdisanti (approved):

ysaito1001 (approved):
Thank you for going through and making them consistent!

PR #3498, opened 2024-03-19
Fix `sigv4-s3express` position in `auth_scheme_options`

ysaito1001:
## Motivation and Context
Fixes an S3-related bug introduced when s3 Express was added

## Description
This PR fixes a bug where s3 Express auth flow was incorrectly used when no auth schemes were available for an endpoint. This use case was observed in test settings and manifested itself as a confusing behavior.

To address this issue, the fix ensures that an auth scheme for S3 express will be registered after that for sigv4 has been registered.

## Testing
Added an integration test `s3_express_auth_flow_should_not_be_reached_with_no_auth_schemes`

## Checklist

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
@@ -39,6 +39,7 @@ import software.amazon.smithy.rust.codegen.core.util.hasEventStreamOperations
 import software.amazon.smithy.rust.codegen.core.util.hasTrait
 import software.amazon.smithy.rust.codegen.core.util.isInputEventStream
 import software.amazon.smithy.rust.codegen.core.util.thenSingletonListOf
+import software.amazon.smithy.rustsdk.customize.s3.S3ExpressDecorator
 
 internal fun ClientCodegenContext.usesSigAuth(): Boolean =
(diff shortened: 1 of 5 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
```diff
@@ -61,7 +62,9 @@ class SigV4AuthDecorator : ConditionalDecorator(
     delegateTo =
         object : ClientCodegenDecorator {
             override val name: String get() = "SigV4AuthDecorator"
-            override val order: Byte = 0
+
+            // This decorator must decorate before S3ExpressDecorator so that sigv4 appears before sigv4-s3express within auth_scheme_options
+            override val order: Byte = (S3ExpressDecorator.ORDER + 1).toByte()
```
jdisanti, line 67: It's a little strange for the generic sigv4 decorator to reference a specific S3 customization. Should be the other way around. Make the S3 decorator reference this one's order.

ysaito1001: I did the other around for the first time I implemented but I switched it to be the way it is, because companion object cannot be placed into a local class. But it didn't look as bad, so I will revert it to match your suggestion.

ysaito1001: Addressed in 726926d & 21f62c3

rcoh (approved):

PR #3503, opened 2024-03-20
Fix the runtime versioner and update tool deps

Velfi:
This issue was caused by a runtime crate referencing another runtime crate with a version dependency instead of a path dependency. I've added an error message that tells us how to fix this if it comes up again:

```txt
crate `aws-smithy-mocks-experimental` depends on crate `aws-smithy-types` crate by version instead of by path. Please update it to use path dependencies for all runtime crates.
```

----

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -1,0 +1,14 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Canary
+on:
+  workflow_dispatch:
+
+jobs:
+  canary:
+    name: Canary
+    runs-on: ubuntu-latest
+    steps:
+    - name: Invoke canary in aws-sdk-rust
+      run: echo "Hello World"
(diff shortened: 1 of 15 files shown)


ysaito1001:
Thanks for the fix! Maybe we can verify the fix by release dry-run using this feature branch and 8a16ddf as input to the release workflow if we haven't done so?

rcoh (approved):

Velfi:
I updated the lockfile b/c it was pulling in an outdated and incompatible version of `toml_edit`

jdisanti (approved):

Velfi:
This issue was caused by a runtime crate referencing another runtime crate with a version dependency instead of a path dependency. I've added an error message that tells us how to fix this if it comes up again:
```
crate `aws-smithy-mocks-experimental` depends on crate `aws-smithy-types` crate by version instead of by path. Please update it to use path dependencies for all runtime crates.
```

ysaito1001 (approved):

jdisanti (approved):

Issue #2598, opened 2023-04-19
Reject server operation inputs with unknown structure members


david-perez:
Currently, the deserializers the server uses carry over the behavior in the client whereby if an unknown structure member is received, it is simply skipped.

For example, given:

```smithy
structure SimpleStruct {
    blob: Blob
    boolean: Boolean
    string: String
}
```

We generate:

```rust
loop {
    match tokens.next().transpose()? {
        Some(aws_smithy_json::deserialize::Token::EndObject { .. }) => break,
        Some(aws_smithy_json::deserialize::Token::ObjectKey { key, .. }) => {
            match key.to_unescaped()?.as_ref() {
                "blob" => {
                    builder = builder.set_blob(
                        aws_smithy_json::deserialize::token::expect_blob_or_null(
                            tokens.next(),
                        )?,
                    );
                }
                "boolean" => {
                    builder = builder.set_boolean(
                        aws_smithy_json::deserialize::token::expect_bool_or_null(
                            tokens.next(),
                        )?,
                    );
                }
                "string" => {
                    builder = builder.set_string(
                        aws_smithy_json::deserialize::token::expect_string_or_null(
                            tokens.next(),
                        )?
                        .map(|s| s.to_unescaped().map(|u| u.into_owned()))
                        .transpose()?,
                    );
                }
                _ => aws_smithy_json::deserialize::token::skip_value(tokens)?, // <------------ Reject here.
            }
        }
        other => {
            return Err(
                aws_smithy_json::deserialize::error::DeserializeError::custom(format!(
                    "expected object key or end object, found: {:?}",
                    other
                )),
            )
        }
    }
}
```

Note how we `aws_smithy_json::deserialize::token::skip_value` instead of rejecting.

I think rejecting here is more in line with Smithy's spirit. Smithy expects that server SDKs are authoritative model consumers, meaning that they have been generated with the latest version of the Smithy model, and model changes roll out to service implementations before clients can consume them.

* It also matches the behavior we have when deserializing unknown `enum` variants and union shape members: the request is rejected (granted, in these scenarios there is no way to proceed since in servers we don't model an `Unknown` variant like the client does, which may be out of date).
* Rejecting here avoids a potential problem where a client might be misconfigured and was meaning to hit a "superset API" with entirely different semantics.

PR #3508, opened 2024-03-25
Changes for Protocol Tests

drganjoo:
## Motivation and Context
Protocol tests are passing.

### DIFF
--- a/buildSrc/src/main/kotlin/CodegenTestCommon.kt
+++ b/buildSrc/src/main/kotlin/CodegenTestCommon.kt
@@ -48,7 +48,7 @@ private fun generateSmithyBuild(
                             ${it.extraCodegenConfig ?: ""}
                         },
                         "service": "${it.service}",
-                        "module": "${it.module}",
+                        "module": "${it.module.lowercase()}",
                         "moduleVersion": "0.0.1",
                         "moduleDescription": "test",
                         "moduleAuthors": ["protocoltest@example.com"]
--- a/buildSrc/src/main/kotlin/CodegenTestCommon.kt
+++ b/buildSrc/src/main/kotlin/CodegenTestCommon.kt
@@ -29,6 +29,49 @@ fun generateImports(imports: List<String>): String =
         "\"imports\": [${imports.map { "\"$it\"" }.joinToString(", ")}],"
     }
 
+fun toRustCrateName(input: String): String {
+    val rustKeywords = setOf(
+        // Strict Keywords.
+        "as", "break", "const", "continue", "crate", "else", "enum", "extern",
+        "false", "fn", "for", "if", "impl", "in", "let", "loop", "match", "mod",
+        "move", "mut", "pub", "ref", "return", "Self", "self", "static", "struct",
+        "super", "trait", "true", "type", "unsafe", "use", "where", "while",
+
+        // Weak Keywords.
+        "dyn", "async", "await", "try",
+
+        // Reserved for Future Use.
+        "abstract", "become", "box", "do", "final", "macro", "override", "priv",
+        "typeof", "unsized", "virtual", "yield",
+
+        // Primitive Types.
+        "bool", "char", "i8", "i16", "i32", "i64", "i128", "isize",
+        "u8", "u16", "u32", "u64", "u128", "usize", "f32", "f64", "str",
+
+        // Additional significant identifiers.
+        "proc_macro"
+    )
+
+    // Then within your function, you could include a check against this set
+    if (input.isBlank()) {
+        throw IllegalArgumentException("Rust crate name cannot be empty")
+    }
+    val lowerCased = input.lowercase()
+    // Replace any sequence of characters that are not lowercase letters, numbers, or underscores with a single underscore.
+    val sanitized = lowerCased.replace(Regex("[^a-z0-9_]+"), "_")
+    // Trim leading or trailing underscores.
+    val trimmed = sanitized.trim('_')
+    // Check if the resulting string is empty, purely numeric, or a reserved name
+    val finalName = when {
+        trimmed.isEmpty() -> throw IllegalArgumentException("Rust crate name after sanitizing cannot be empty.")
+        trimmed.matches(Regex("\\d+")) -> "n$trimmed" // Prepend 'n' if the name is purely numeric.
+        trimmed in rustKeywords -> "${trimmed}_" // Append an underscore if the name is reserved.
+        else -> trimmed
+    }
+    return finalName
+}
+
+
 private fun generateSmithyBuild(
     projectDir: String,
     pluginName: String,
@@ -48,7 +91,7 @@ private fun generateSmithyBuild(
                             ${it.extraCodegenConfig ?: ""}
                         },
                         "service": "${it.service}",
-                        "module": "${it.module.lowercase()}",
+                        "module": "${toRustCrateName(it.module)}",
                         "moduleVersion": "0.0.1",
                         "moduleDescription": "test",
                         "moduleAuthors": ["protocoltest@example.com"]
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
@@ -66,6 +66,8 @@ import software.amazon.smithy.rust.codegen.core.util.hasTrait
 import software.amazon.smithy.rust.codegen.core.util.isTargetUnit
 import software.amazon.smithy.rust.codegen.core.util.letIf
 import java.math.BigDecimal
+import software.amazon.smithy.model.traits.DefaultTrait
+import software.amazon.smithy.rust.codegen.core.util.getTrait
 
 /**
  * Class describing an instantiator section that can be used in a customization.
@@ -448,12 +450,10 @@ open class Instantiator(
 
     /**
      * Returns a default value for a shape.
-     *
-     * Warning: this method does not take into account any constraint traits attached to the shape.
      */
     private fun fillDefaultValue(shape: Shape): Node =
         when (shape) {
-            is MemberShape -> fillDefaultValue(model.expectShape(shape.target))
+            is MemberShape -> shape.getTrait<DefaultTrait>()?.toNode() ?: fillDefaultValue(model.expectShape(shape.target))
 
             // Aggregate shapes.
             is StructureShape -> Node.objectNode()
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
@@ -450,6 +450,8 @@ open class Instantiator(
 
     /**
      * Returns a default value for a shape.
+     *
+     * Warning: this method does not take into account any constraint traits attached to the shape.
      */
     private fun fillDefaultValue(shape: Shape): Node =
         when (shape) {
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2.kt
@@ -7,10 +7,8 @@ package software.amazon.smithy.rust.codegen.core.smithy.protocols
 
 import software.amazon.smithy.codegen.core.CodegenException
 import software.amazon.smithy.model.Model
-import software.amazon.smithy.model.pattern.UriPattern
 import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.model.shapes.ToShapeId
-import software.amazon.smithy.model.traits.HttpTrait
 import software.amazon.smithy.model.traits.TimestampFormatTrait
 import software.amazon.smithy.rust.codegen.core.rustlang.CargoDependency
 import software.amazon.smithy.rust.codegen.core.rustlang.RustModule
@@ -21,7 +19,9 @@ import software.amazon.smithy.rust.codegen.core.smithy.protocols.parse.CborParse
 import software.amazon.smithy.rust.codegen.core.smithy.protocols.parse.StructuredDataParserGenerator
 import software.amazon.smithy.rust.codegen.core.smithy.protocols.serialize.CborSerializerGenerator
 import software.amazon.smithy.rust.codegen.core.smithy.protocols.serialize.StructuredDataSerializerGenerator
+import software.amazon.smithy.rust.codegen.core.smithy.traits.SyntheticOutputTrait
 import software.amazon.smithy.rust.codegen.core.util.PANIC
+import software.amazon.smithy.rust.codegen.core.util.expectTrait
 import software.amazon.smithy.rust.codegen.core.util.isStreaming
 import software.amazon.smithy.rust.codegen.core.util.outputShape
 
@@ -65,10 +65,14 @@ class RpcV2HttpBindingResolver(
     /**
      * > Responses for operations with no defined output type MUST NOT contain bodies in their HTTP responses.
      * > The `Content-Type` for the serialization format MUST NOT be set.
-     *
-     * TODO But the problem is we _always_ add input/output types in OperationNormalizer! This makes the `no_output_response` test fail.
      */
-    override fun responseContentType(operationShape: OperationShape): String = requestContentType(operationShape)
+    override fun responseContentType(operationShape: OperationShape): String? {
+        val syntheticOutputTrait = operationShape.outputShape(model).expectTrait<SyntheticOutputTrait>()
+        if (syntheticOutputTrait.originalId == null) {
+            return null
+        }
+        return requestContentType(operationShape)
+    }
 }
 
 /**
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2.kt
@@ -67,6 +67,8 @@ class RpcV2HttpBindingResolver(
      * > The `Content-Type` for the serialization format MUST NOT be set.
      */
     override fun responseContentType(operationShape: OperationShape): String? {
+        // When `syntheticOutputTrait.originalId == null` it implies that the operation had no output defined
+        // in the Smithy model.
         val syntheticOutputTrait = operationShape.outputShape(model).expectTrait<SyntheticOutputTrait>()
         if (syntheticOutputTrait.originalId == null) {
             return null
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
@@ -103,13 +103,13 @@ class CborParserGenerator(
             *codegenScope,
             "ListSymbol" to listSymbol,
         ) {
-            val deserializeMemberWritable = writable { deserializeMember(memberShape) }
+            val deserializeMemberWritable = deserializeMember(memberShape)
             if (isSparseList) {
                 rustTemplate(
                     """
                     let value = match decoder.datatype()? {
                         #{SmithyCbor}::data::Type::Null => {
-                            let _v = decoder.null()?;
+                            decoder.null()?;
                             None
                         }
                         _ => Some(#{DeserializeMember:W}?),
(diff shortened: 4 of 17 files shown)


buildSrc/src/main/kotlin/CodegenTestCommon.kt
```diff
@@ -48,7 +48,7 @@ private fun generateSmithyBuild(
                             ${it.extraCodegenConfig ?: ""}
                         },
                         "service": "${it.service}",
-                        "module": "${it.module}",
+                        "module": "${it.module.lowercase()}",
```
david-perez, line 51: This is still best effort. I suggest we keep this as-is unless we plan on coming up with a full mechanism to "massage" an arbitrary string into a valid crate name.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -448,12 +450,10 @@ open class Instantiator(
 
     /**
      * Returns a default value for a shape.
-     *
-     * Warning: this method does not take into account any constraint traits attached to the shape.
```
david-perez, line 452: This is still accurate: `Instantiator` doesn't instantiate with values that honor constraint traits.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/Instantiator.kt
```diff
@@ -448,12 +450,10 @@
...
      */
     private fun fillDefaultValue(shape: Shape): Node =
         when (shape) {
-            is MemberShape -> fillDefaultValue(model.expectShape(shape.target))
+            is MemberShape -> shape.getTrait<DefaultTrait>()?.toNode() ?: fillDefaultValue(model.expectShape(shape.target))
```
david-perez, line 456: Good catch.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2.kt
```diff
@@ -65,10 +65,14 @@ class RpcV2HttpBindingResolver(
...
-     * TODO But the problem is we _always_ add input/output types in OperationNormalizer! This makes the `no_output_response` test fail.
      */
-    override fun responseContentType(operationShape: OperationShape): String = requestContentType(operationShape)
+    override fun responseContentType(operationShape: OperationShape): String? {
+        val syntheticOutputTrait = operationShape.outputShape(model).expectTrait<SyntheticOutputTrait>()
+        if (syntheticOutputTrait.originalId == null) {
+            return null
+        }
+        return requestContentType(operationShape)
+    }
```
david-perez, line 75: Nice. Let's add a comment explaining that when `syntheticOutputTrait.originalId == null`, it implies that the original operation had no output; this is subtle.

Hot take: a breaking change I've always wanted to make is to *not* apply `OperationNormalizer` in the server, we derive no real benefit from it and it just makes our code more complex.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -103,13 +103,13 @@ class CborParserGenerator(
...
         ) {
-            val deserializeMemberWritable = writable { deserializeMember(memberShape) }
+            val deserializeMemberWritable = deserializeMember(memberShape)
             if (isSparseList) {
                 rustTemplate(
                     """
                     let value = match decoder.datatype()? {
                         #{SmithyCbor}::data::Type::Null => {
-                            let _v = decoder.null()?;
+                            decoder.null()?;
```
david-perez, line 112: I wrote the original because it better conveys that we're conciously ignoring the return type; it's a common Rust idiom.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -154,20 +154,20 @@
...
             )
-            val deserializeValueWritable = writable { deserializeMember(valueShape) }
+            val deserializeValueWritable = deserializeMember(valueShape)
             if (isSparseMap) {
                 rustTemplate(
                     """
                     let value = match decoder.datatype()? {
                         #{SmithyCbor}::data::Type::Null => {
-                            let _v = decoder.null()?;
+                            decoder.null()?;
```
david-perez, line 170: Ditto.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -206,33 +206,59 @@
...
-                                        withBlock("let v = ", "?;") {
-                                            deserializeMember(member)
+                        val callBuilderSetMemberFieldWritable = writable {
+                            withBlock("builder.${member.setterName()}(", ")") {
+                                conditionalBlock("Some(", ")", symbolProvider.toSymbol(member).isOptional()) {
+                                    val symbol = symbolProvider.toSymbol(member)
+                                    if (symbol.isRustBoxed()) {
+                                        rustBlock("") {
+                                            rustTemplate("let v = #{DeserializeMember:W}?;",
+                                                "DeserializeMember" to deserializeMember(member))
```
david-perez, line 219: style nit:

```suggestion
                                            rustTemplate(
                                                "let v = #{DeserializeMember:W}?;",
                                                "DeserializeMember" to deserializeMember(member)
                                            )
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -206,33 +206,59 @@
...
                                     }
-                                } else {
-                                    deserializeMember(member)
-                                    rust("?")
                                 }
                             }
                         }
+
+                        if (member.isOptional) {
+                            // Call `builder.set_member()` only if the value for the field on the wire is not null.
```
david-perez, line 239: Nice.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -206,33 +206,59 @@
...
 
-                // TODO Skip like in JSON or reject? I think we should reject in JSON too. Cut issue.
-                rust("_ => { todo!() }")
+                rust(
+                    """
+                    _ => { 
+                        decoder.skip()?;
+                        builder
+                    }
+                    """)
```
david-perez, line 261: Let's add a comment that we're explicitly skipping unrelated fields as per the spec.

I guess our current behavior in JSON protocols of skipping too is desirable then. I opened https://github.com/smithy-lang/smithy-rs/issues/2598 in the past but never followed up on it. It would be good to justify why we're skipping and link to the spec.

style nit:

```suggestion
                rust(
                    """
                    _ => { 
                        decoder.skip()?;
                        builder
                    }
                    """
                )
```

Doesn't ktlint catch these?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -256,7 +282,7 @@
                     val variantName = symbolProvider.toMemberName(member)
 
                     withBlock("${member.memberName.dq()} => #T::$variantName(", "?),", returnSymbolToParse.symbol) {
-                        deserializeMember(member)
+                        deserializeMember(member).invoke(this)
```
david-perez, line 285: Oh wow I didn't know you could call `invoke` on `Writable`s.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -423,13 +449,13 @@
         return structureParser(operationShape, symbolProvider.symbolForBuilder(inputShape), includedMembers)
     }
 
-    private fun RustWriter.deserializeMember(memberShape: MemberShape) {
+    private fun RustWriter.deserializeMember(memberShape: MemberShape) = writable {
```
david-perez, line 452: We now don't need this to be an extension function on `RustWriter`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/CborSerializerGenerator.kt
```diff
@@ -402,7 +402,7 @@ class CborSerializerGenerator(
         rust(
             """
             encoder.array(
-                (${context.valueExpression.asRef()}).len().try_into().expect("`usize` to `u64` conversion failed")
+                (${context.valueExpression.asValue()}).len().try_into().expect("`usize` to `u64` conversion failed")
```
david-perez, line 405: I don't think this is right. `context.valueExpression` should be a reference here (e.g. a `&Vec<T>`), there's no need to dereference it (with `*`) to call `len()` on it.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/CborSerializerGenerator.kt
```diff
@@ -421,7 +421,7 @@
         rust(
             """
             encoder.map(
-                (${context.valueExpression.asRef()}).len().try_into().expect("`usize` to `u64` conversion failed")
+                (${context.valueExpression.asValue()}).len().try_into().expect("`usize` to `u64` conversion failed")
```
david-perez, line 424: Same. Here we should be handling a `&HashMap<String, T>`, so `asRef()` should just print the variable name as is.

codegen-server-test/build.gradle.kts
```diff
@@ -45,8 +45,10 @@ val allCodegenTests = "../codegen-core/common-test-models".let { commonModels ->
             imports = listOf("$commonModels/naming-obstacle-course-structs.smithy"),
         ),
         CodegenTest("com.amazonaws.simple#SimpleService", "simple", imports = listOf("$commonModels/simple.smithy")),
-        CodegenTest("aws.protocoltests.rpcv2Cbor#RpcV2Protocol", "rpcv2_cbor"),
-        CodegenTest("com.amazonaws.simple#RpcV2Service", "rpcv2", imports = listOf("$commonModels/rpcv2.smithy")),
+        // CodegenTest("aws.protocoltests.restxml#RestXml", "restXml"),
+        CodegenTest("aws.protocoltests.rpcv2Cbor#RpcV2Protocol", "rpcV2"),
+        // Todo: change this to rpcv2extra
```
david-perez, line 50:
```suggestion
        // TODO: change this to rpcV2Extra
```

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocol.kt
```diff
@@ -357,5 +357,5 @@ class ServerRpcV2Protocol(
 
     override fun serverRouterRuntimeConstructor() = "rpc_v2_router"
 
-    override fun serverContentTypeCheckNoModeledInput() = true
+    override fun serverContentTypeCheckNoModeledInput() = false
```
david-perez, line 360: We need to update the comment in the `ServerProtocol` `interface` which calls out `rpcv2` and `restJson1`.

rust-runtime/aws-smithy-cbor/Cargo.toml
```diff
@@ -1,21 +1,32 @@
...
 license = "Apache-2.0"
 repository = "https://github.com/awslabs/smithy-rs"
 
 [dependencies.minicbor]
 version = "0.19.1" # TODO Update
-features = ["alloc"] # To write to a `Vec<u8>`: https://docs.rs/minicbor/latest/minicbor/encode/write/trait.Write.html#impl-Write-for-Vec%3Cu8%3E
+features = [
+    "alloc",
+    "half"  # To support reading `f16` to accomodate fewer bytes transmitted that fit the value.
+] # To write to a `Vec<u8>`: https://docs.rs/minicbor/latest/minicbor/encode/write/trait.Write.html#impl-Write-for-Vec%3Cu8%3E
```
david-perez, line 18:
```suggestion
features = [
    # To write to a `Vec<u8>`: https://docs.rs/minicbor/latest/minicbor/encode/write/trait.Write.html#impl-Write-for-Vec%3Cu8%3E
    "alloc",
    # To support reading `f16` to accomodate fewer bytes transmitted that fit the value.
    "half"
]
```

rust-runtime/aws-smithy-cbor/Cargo.toml
```diff
@@ -1,21 +1,32 @@
...
 
 [dependencies]
 aws-smithy-types = { path = "../aws-smithy-types" }
+criterion = "0.5.1"
```
david-perez, line 22: Shouldn't this be a `dev-dependency`? https://bheisler.github.io/criterion.rs/book/getting_started.html#step-1---add-dependency-to-cargotoml

rust-runtime/aws-smithy-cbor/benches/string.rs
```diff
@@ -0,0 +1,113 @@
...
+        0x6a, 0x74, 0x68, 0x69, 0x73, 0x49, 0x73, 0x41, 0x4b, 0x65, 0x79,
+    ];
+
+    // Indefinite length key `this`, `Is`, `A` and `Key`.
+    let indefinite_bytes = [
+        0x7f, 0x64, 0x74, 0x68, 0x69, 0x73, 0x62, 0x49, 0x73, 0x61, 0x41, 0x63, 0x4b, 0x65, 0x79,
+        0xff,
+    ];
+
+    c.bench_function("definite str_alt", |b| {
```
david-perez, line 16: nit: I'd order the benches so that `definite str_alt` comes right after `definite str` etc so they appear next to each other in the report and we can more easily compare.

rust-runtime/aws-smithy-cbor/benches/string.rs
```diff
@@ -0,0 +1,113 @@
...
+            let mut decoder = Decoder::new(&indefinite_bytes);
+            let x = black_box(decoder.str());
+            assert!(matches!(x.unwrap().as_ref(), "thisIsAKey"));
+        })
+    });
+}
+
+// We have two `string` implementations. One uses `collect` the other
+// uses `String::new` followed by `string::push`.
+pub fn string_benchmark(c: &mut Criterion) {
```
david-perez, line 49: This benchmark is using `str` and `str_alt` instead of `string` and `string_alt`.

rust-runtime/aws-smithy-cbor/benches/string.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    c.bench_function("indefinite string_alt()", |b| {
+        b.iter(|| {
+            let mut decoder = Decoder::new(&indefinite_bytes);
+            let _ = black_box(decoder.str_alt());
+        })
+    });
+}
+
+pub fn blob_benchmark(c: &mut Criterion) {
```
david-perez, line 90: I guess we should put this benchmark in a separate filename? Or maybe modify the filename and benchmark name.

rust-runtime/aws-smithy-cbor/benches/string.rs
```diff
@@ -0,0 +1,113 @@
...
+            let _ = black_box(decoder.blob());
+        })
+    });
+}
+
+criterion_group!(
+    benches,
+    string_benchmark,
+    // str_benchmark blob_benchmark,
+    // str_benchmark,
```
david-perez, line 110: Why ignore some of the benchmarks?

rust-runtime/aws-smithy-cbor/src/decode.rs
```diff
@@ -75,28 +77,85 @@ impl<'b> Decoder<'b> {
...
     }
 
-    // TODO An API to support both definite and indefinite strings could be:
-    //
-    //     pub fn str(&mut self, cap: u64) -> Result<&'b str, DeserializeError> {
-    //
-    pub fn str(&mut self) -> Result<&'b str, DeserializeError> {
-        self.decoder.str().map_err(DeserializeError::new)
+    // TODO-David: confirm benchmarks and keep either `str_alt` or `str`.
+    // The following seems to be a bit slower than the one we have kept.
```
david-perez, line 81: On my machine `str` is faster than `str_alt` for definite-length, and slower for indefinite-length.

I can see why `str` would be slower for indefinite length strings, since we'd be checking what data type we're decoding twice: once when attempting as a definite length string with `self.decoder.str()`, and another one when delegating to `self.string()`.

I don't understand why `str` would be faster than `str_alt` for _definite_ length strings though.

rust-runtime/aws-smithy-cbor/src/decode.rs
```diff
@@ -75,28 +77,85 @@
...
+            }
+        }
+    }
+
+    pub fn str(&mut self) -> Result<Cow<'b, str>, DeserializeError> {
+        match self.decoder.str() {
+            Ok(str_value) => Ok(Cow::Borrowed(str_value)),
+            Err(e) if e.is_type_mismatch() => {
+                // Move the position back by one element to the indefinite string marker.
+                self.decoder.set_position(self.decoder.position() - 1);
```
david-perez, line 111: We're relying on the fact that `minicbor` only needs to read one byte to determine the data type here. We should document this since it's an important assumption that affects correctness.

A more robust implementation would be to cache the position before we issue `str()` and restore it afterwards.

rust-runtime/aws-smithy-cbor/src/decode.rs
```diff
@@ -75,28 +77,85 @@
...
+                Ok(Cow::Owned(self.string()?))
+            }
+            Err(e) => Err(DeserializeError::new(e)),
+        }
+    }
+
+    // TODO-David: confirm benchmarks and keep either `string_alt` or `string` implementation.
+    // The following seems to be a bit slower than the one we have kept.
```
david-perez, line 119: On my machine `string` is faster than `string_alt` for definite-length and slower for indefinite-length. I honestly don't know why without diving deeper.

rust-runtime/aws-smithy-cbor/src/decode.rs
```diff
@@ -75,28 +77,85 @@
...
-            .map(String::from) // This allocates.
-            .map_err(DeserializeError::new)
+        let mut iter = self.decoder.str_iter().map_err(DeserializeError::new)?;
+        let head = iter.next();
+
+        let decoded_string = match head {
+            None => String::new(),
+            Some(head) => {
+                // The following is faster in benchmarks than using `Collect()` on a `String`.
+                let mut head = String::from(head.map_err(DeserializeError::new)?);
```
david-perez, line 137: Not a good variable name since it's no longer just the `head`.

rust-runtime/aws-smithy-cbor/src/decode.rs
```diff
@@ -185,3 +244,80 @@
             .map(|opt| opt.map_err(DeserializeError::new))
     }
 }
+
+pub fn set_optional<B, F>(builder: B, decoder: &mut Decoder, f: F) -> Result<B, DeserializeError>
```
david-perez, line 248: Docs would be nice.

rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.72.1"
+channel = "1.76.0"
```
david-perez, line 2: I guess this was changed because of:

```
error: package `clap_lex v0.7.0` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.72.1
```

```
clap_lex v0.7.0
└── clap_builder v4.5.2
    └── clap v4.5.3
        └── criterion v0.5.1
            └── aws-smithy-cbor v0.0.0-smithy-rs-head (/local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/rust-runtime/aws-smithy-cbor)
                ├── aws-smithy-http-server v0.0.0-smithy-rs-head (/local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/rust-runtime/aws-smithy-http-server)
                │   └── rpcv2 v0.0.1 (/local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/codegen-server-test/build/smithyprojections/codegen-server-test/rpcV2/rust-server-codegen)
                └── rpcv2 v0.0.1 (/local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/codegen-server-test/build/smithyprojections/codegen-server-test/rpcV2/rust-server-codegen)
```

Let's change it to 1.74.1, [which is what we currently have in `main`](https://github.com/smithy-lang/smithy-rs/blob/2d931fa8dc0da530fa0e17ec99d293895583bf0b/rust-toolchain.toml#L2), which should suffice. Rustc upgrades should go in a separate PR and usually require further changes.

david-perez (commented):
I'd like to spend more time diving deep on why the benchmarks exhibit the described behavior but unfortunately can't. Whichever implementation we choose, it'd be nice to keep the other one checked in in the `benches/` directory to reproduce in the future our design choice.

buildSrc/src/main/kotlin/CodegenTestCommon.kt
drganjoo: Let me do a better function than lowercase. If we keep it as-is it generates a warning.

PR #3513, opened 2024-03-25
Make BehaviorVersion as future-proof as the documentation and RFC specify

Ten0:
Fixes aws-sdk-rust#1111

This is *technically* a breaking change but since this was released recently and it's documented that this API shouldn't be used, this can probably be considered a bug, which is better fixed now than later.


ysaito1001:
Hi @Ten0, thank you for catching this and contributing the fix! Will get that merged once the PR has passed CI.

ysaito1001 (approved):

PR #3520, opened 2024-03-27
Make runtime versioner ignore any additional flags

rcoh:
## Motivation and Context
This is the first of two PRs to remove the concept of stable/unstable versions from the release process. This PR will cause a full docker rebuild, so I'm splitting it out to merge first to make iterating on the second PR easier.

----

### DIFF
--- a/tools/ci-scripts/check-semver-hazards
+++ b/tools/ci-scripts/check-semver-hazards
@@ -10,14 +10,9 @@
 C_YELLOW='\033[1;33m'
 C_RESET='\033[0m'
 
-if [ "$#" -ne 2 ]; then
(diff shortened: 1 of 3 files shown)


tools/ci-scripts/check-semver-hazards
```diff
@@ -28,8 +23,6 @@ echo -e "${C_YELLOW}# Patching SDK...${C_RESET}"
 runtime-versioner patch-runtime \
   --sdk-path "$(pwd)/aws-sdk-rust" \
   --smithy-rs-path "$(pwd)/smithy-rs" \
```
jdisanti, line 25:
```suggestion
  --smithy-rs-path "$(pwd)/smithy-rs"
```

jdisanti (approved):

PR #3540, opened 2024-03-29
Don't generate a SDK for runtime-versioner patching

jdisanti:
This PR removes the requirement of generating a SDK in order to patch smithy-rs runtime crates into an already released SDK version for testing. This is possible now that all the runtime crates are independently versioned. The tool also takes multiple paths as input for the runtime crate locations. This makes it possible to pass in the `rust-runtime` and `aws/rust-runtime` directories from smithy-rs (done by default by one of the subcommands).

Some of the publisher tool's package resolution code is moved into smithy-rs-tool-common so it can be better shared across the tools, and as part of this, the version in the package handle was made optional.

----


rcoh (approved):
consider removing the asserts by creating "UnpublishablePackage". We can keep version required except in the case where we need it to be optional.

My branch kinda went the wrong direction and make optional the default generic, but I think if we went the other direction and made non-optional verions the default we could remove a lot of asserts & make only small changes to the code.

tools/ci-build/smithy-rs-tool-common/src/package.rs
```diff
@@ -46,3 +55,291 @@ pub enum PackageStability {
...
+pub enum Publish {
+    Allowed,
+    NotAllowed,
+}
+
+/// Represents a crate (called Package since crate is a reserved word).
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+pub struct Package {
+    /// Package name and version information
+    pub handle: PackageHandle,
```
aajtodd, line 108: If it's just name, version, etc this feels more like `meta`, perhaps it should be `PackageMeta`

jdisanti: There's a lot more to meta than just name and version (authors, license, etc). This should only ever be just the name and version.

tools/ci-build/publisher/src/cargo/publish.rs
```diff
@@ -18,6 +20,10 @@ pub struct Publish {
 
 impl Publish {
     pub fn new(package_handle: PackageHandle, package_path: impl Into<PathBuf>) -> Publish {
+        assert!(
```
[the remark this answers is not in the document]
jdisanti: I tried it out, and while it works well for PackageHandle itself, it kind of breaks down for Package. Especially in the loading logic. I think it just makes things more complex to remove a few possible (yet easy to debug) panics.

Issue #3427, opened 2024-02-19
Credentials are loaded once for each client


rcoh:
Each `IdentityResolver` that is created has its own `IdentityPartitionCacheKey`. This means that we hit separate caches. We need to create the identity resolver in `SdkConfig` as part of the call to `build()`.

e.g:

```
--- a/aws/rust-runtime/aws-types/src/sdk_config.rs
+++ b/aws/rust-runtime/aws-types/src/sdk_config.rs
@@ -55,6 +55,7 @@ pub struct SdkConfig {
     app_name: Option<AppName>,
     identity_cache: Option<SharedIdentityCache>,
     credentials_provider: Option<SharedCredentialsProvider>,
+    credentials_identity_resolver: Option<SharedIdentityResolver>,
     region: Option<Region>,
     endpoint_url: Option<String>,
     retry_config: Option<RetryConfig>,
```

Logs:
```
2024-02-19T17:23:02.363590Z  INFO lazy_load_identity: aws_smithy_runtime::client::identity::cache::lazy: identity cache miss occurred; added new identity (took Ok(678.027ms)) new_expiration=SystemTime { tv_sec: 1708366982, tv_nsec: 0 } valid_for=Ok(3599.636442s) partition=IdentityCachePartition(1)
2024-02-19T17:23:03.600080Z  INFO lazy_load_identity: aws_smithy_runtime::client::identity::cache::lazy: identity cache miss occurred; added new identity (took Ok(733.768ms)) new_expiration=SystemTime { tv_sec: 1708366983, tv_nsec: 0 } valid_for=Ok(3599.399947s) partition=IdentityCachePartition(4)
```

jdisanti:
This was fixed in the [April 12th release](https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2024-04-12).

aajtodd:
Well to be clear it is now possible, you still have to pass an `identity_cache(..)` to the config loader yourself until #3578 is merged and made the default behavior.

PR #3544, opened 2024-04-02
rfc: change default identity cache and partitioning

aajtodd:
## Motivation and Context

See the motivation section of the RFC.

## Description

Proposes a new RFC that changes the default behavior of `SdkConfig` and `SharedCredentialsProvider`/`SharedTokenProvider` w.r.t caching and cache partitions. 

### DIFF
--- a/design/src/SUMMARY.md
+++ b/design/src/SUMMARY.md
@@ -69,6 +69,7 @@
   - [RFC-0040: Behavior Versions](./rfcs/rfc0040_behavior_versions.md)
   - [RFC-0041: Improve client error ergonomics](./rfcs/rfc0041_improve_client_error_ergonomics.md)
   - [RFC-0042: File-per-change changelog](./rfcs/rfc0042_file_per_change_changelog.md)
+  - [RFC-0043: Identity Cache Partitions](./rfcs/rfc0043_identity_cache_partitions.md)
 
 - [Contributing](./contributing/overview.md)
   - [Writing and debugging a low-level feature that relies on HTTP](./contributing/writing_and_debugging_a_low-level_feature_that_relies_on_HTTP.md)
--- a/design/src/rfcs/overview.md
+++ b/design/src/rfcs/overview.md
@@ -52,3 +52,6 @@
 - [RFC-0040: Behavior Versions](./rfc0040_behavior_versions.md)
 - [RFC-0041: Improve client error ergonomics](./rfc0041_improve_client_error_ergonomics.md)
 - [RFC-0042: File-per-change changelog](./rfc0042_file_per_change_changelog.md)
+- [RFC-0043: Identity Cache Partitions](./rfc0043_identity_cache_partitions.md)
+
+>>>>>>> 4a8757c23 (add RFC to fix identity cache partitioning and default cache behaviors)
--- a/design/src/rfcs/rfc0043_identity_cache_partitions.md
+++ b/design/src/rfcs/rfc0043_identity_cache_partitions.md
@@ -1,0 +1,295 @@
+RFC: Identity Cache Partitions
+===============================
+
+> Status: Accepted
+>
+> Applies to: AWS SDK for Rust
+
+Motivation
+-----------
+
+In the below example two clients are created from the same shared `SdkConfig` instance and each
+invoke a fictitious operation. Assume the operations use the same auth scheme relying on the same identity resolver.
+
+```rust,ignore
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+
+    let config = aws_config::defaults(BehaviorVersion::latest())
+        .load()
+        .await;
+
+    let c1 = aws_sdk_foo::Client::new(&config);
+    c1.foo_operation().send().await;
+
+    let c2 = aws_sdk_bar::Client::new(&config);
+    c2.bar_operation().send().await;
+
+    Ok(())
+}
+```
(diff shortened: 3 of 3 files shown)


rcoh (approved):
design LGTM. I think this is missing a change checklist item for actually _using_ the cache partition in the identity cache.

rcoh:
I think the RFC may also want to discuss the alternatives to address this issue

ysaito1001:
> I think this is missing a change checklist item for actually using the cache partition in the identity cache

My understanding is that the cache partition [is already used](https://github.com/smithy-lang/smithy-rs/blob/88405d638348283de32344dd783c72621e4f1165/rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs#L307-L308)? and it's just that a `SharedCredentialsProvider` previously returned `None`, forcing a new partition to be created. After implementing checklist items things should be wired up automatically, I think.

aajtodd:
> > I think this is missing a change checklist item for actually using the cache partition in the identity cache
> 
> My understanding is that the cache partition [is already used](https://github.com/smithy-lang/smithy-rs/blob/88405d638348283de32344dd783c72621e4f1165/rust-runtime/aws-smithy-runtime/src/client/identity/cache/lazy.rs#L307-L308)? and it's just that a `SharedCredentialsProvider` previously returned `None`, forcing a new partition to be created. After implementing checklist items things should be wired up automatically, I think.

That is coming from `SharedIdentityResolver`, I'll make it more explicit but we will need to wire it up in [SharedIdentityResolver::new](https://github.com/smithy-lang/smithy-rs/blob/release-2024-03-25/rust-runtime/aws-smithy-runtime-api/src/client/identity.rs#L197) and see if the given `resolver.cache_partition()` is set.

ysaito1001 (approved):

Velfi (approved):

design/src/rfcs/rfc0041_identity_cache_partitions.md
```diff
@@ -0,0 +1,293 @@
...
+behind a new behavior version.
+
+### Opting out
+
+Users can disable the shared identity cache by explicitly setting it to `None`. This will result in each client
+creating their own identity cache.
+
+```rust,ignore
+let config = aws_config::defaults(BehaviorVersion::latest())
+    .set_identity_cache(None)
```
aajtodd, line 115: This apparently doesn't exist on `ConfigLoader`, it only exists on the explicit client config. We'll probably instead need something like `no_identity_cache()` to match `no_credentials()`

design/src/rfcs/rfc0041_identity_cache_partitions.md
```diff
@@ -0,0 +1,293 @@
...
+
+There are two problems with this currently.
+
+
+1. The identity resolvers (e.g. `credentials_provider` for SigV4) are re-used but we end up with a different 
+[`IdentityCachePartition`](https://github.com/smithy-lang/smithy-rs/blob/release-2024-03-25/rust-runtime/aws-smithy-runtime-api/src/client/identity.rs#L41)
+each time a client is created. 
+    * More specifically this happens every time a `SharedIdentityResolver` is [created](https://github.com/smithy-lang/smithy-rs/blob/release-2024-03-25/rust-runtime/aws-smithy-runtime-api/src/client/identity.rs#L197). The conversion from [`From<SdkConfig>`](https://github.com/awslabs/aws-sdk-rust/blob/release-2024-04-01/sdk/sts/src/config.rs#L1207) 
+    sets the credentials provider which [associates](https://github.com/awslabs/aws-sdk-rust/blob/release-2024-04-01/sdk/sts/src/config.rs#L960) it as
+    the identity resolver for the auth scheme. Internally this is [converted](https://github.com/awslabs/aws-sdk-rust/blob/release-2024-04-01/sdk/aws-smithy-runtime-api/src/client/runtime_components.rs#L663) to `SharedIdentityResolver` which creates the new partition. The end result is the credentials
```
jdisanti, line 40: If it were already a `SharedIdentityResolver`, then the `IntoShared` implementation should detect that and wouldn't create a new instance. So that means it must be a `SharedCredentialsProvider` that's getting converted into a `SharedIdentityResolver`. This could be made clearer here.

design/src/rfcs/rfc0041_identity_cache_partitions.md
```diff
@@ -0,0 +1,293 @@
...
+    ...
+
+    fn cache_partition(&self) -> Option<IdentityCachePartition> {
+        Some(self.1)
+    }
+}
+
+```
+
+Additionally a new behavior version must be introduced that conditionally creates a default `IdentityCache` on `SdkConfig`
```
jdisanti, line 250: Most of the defaults are provided by runtime plugins later on. Why would this one get a default at this stage?

jdisanti (commented):
Nicely written RFC!

design/src/rfcs/rfc0041_identity_cache_partitions.md
aajtodd: If we don't then each client gets their own identity cache rather than sharing one. Even if we fix the cache partitioning the default behavior will still end up being that multiple clients created from `SdkConfig` don't actually share cached results.

This _can_ be worked around of course by explicitly providing a cache to `SdkConfig` but if the default behavior we _want_ is to enable caching it probably makes sense to just make it so.

Issue #2546, opened 2023-04-04
Use of Unit type in server event stream creates non-compiling code


dbgilley:
When attempting to generate server code that returns an event stream, an event type referencing `smithy.api#Unit` generates code that fails to compile, instead producing an error.

smithy-rs version appears to be 0.54.4

A reproducible example is:
```
$version: "2"
namespace com.example
use aws.protocols#restJson1
use smithy.framework#ValidationException


@restJson1 @title("Test Service") service TestService { version: "0.1", operations: [ TestOperation ] }

@http(uri: "/test", method: "POST")
operation TestOperation {
    input := { payload: String },
    output := {
        @httpPayload
        events: TestEvent,
    },
    errors: [ValidationException],
}

@streaming
union TestEvent {
    KeepAlive: Unit,
    Response: TestResponseEvent,
}

structure TestResponseEvent { data: String }
```

The error produced is:
```
---> test/src/event_stream_serde.rs:56:13
   |
56 |             Self::Input::KeepAlive(inner) => {
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not a tuple variant or struct
   ```

rcoh:
Note that this also applies to _any_ union, streaming or not

PR #3547, opened 2024-04-03
Fix errors for unions with unit target membershape

drganjoo:
## Motivation and Context

Unions that have a [unit target member shape](https://smithy.io/2.0/spec/model.html#unit-type) do not have an associated data in the generated Rust enum.

Closes Issue: [2546](https://github.com/smithy-lang/smithy-rs/issues/2546)

## Description

On the **server** side, when the union has constrained members, the code generated for the conversion from the `Unconstrained` type to the `Constrained` type incorrectly assumed that each Rust enum would have associated data.

```
rust-server-codegen/src/unconstrained.rs:31:129
  |
  |               crate::unconstrained::some_union_with_unit_unconstrained::SomeUnionWithUnitUnconstrained::Option1(unconstrained) => Self::Option1(
    |                                                                                                                                   -^^^^^^^^^^^^- help: consider using a semicolon here to finish the statement: `;`
    |  _________________________________________________________________________________________________________________________________|
  | |
  | |                 unconstrained
  | |             ),
  | |_____________- call expression requires function
    |
   ::: rust-server-codegen/src/model.rs:152:5
    |
    |       Option1,
    |       ------- `SomeUnionWithUnit::Option1` defined here
```

The marshaling code for event streams with unit target types incorrectly assumed that the variant would have associated data.

```
rust-server-codegen/src/event_stream_serde.rs

impl ::aws_smithy_eventstream::frame::MarshallMessage for TestEventMarshaller {
        fn marshal() {
            let payload = match input {
                Self::Input::KeepAlive(inner) => {
```

On the **client** side, the `event_stream_serde` code incorrectly assumes that a union member, which has the `@streaming` trait applied to it, takes a `model::Unit` type.

```
rust-client-codegen/src/event_stream_serde.rs:
    |
    |                     crate::types::TestEvent::KeepAlive(crate::types::Unit::builder().build()),
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^---------------------------------------
    |                     |
    |                     call expression requires function
    |
   ::: rust-client-codegen/src/types/_test_event.rs
    |
    |     KeepAlive,
    |     --------- `TestEvent::KeepAlive` defined here
```

## Testing

A unit test has been added that tests the following model:

```
            $version: "2"
            namespace com.example
            use aws.protocols#restJson1
            use smithy.framework#ValidationException
            
            @restJson1 @title("Test Service") 
            service TestService { 
                version: "0.1", 
                operations: [ 
                    TestOperation
                    TestSimpleUnionWithUnit
                ] 
            }
            
            @http(uri: "/testunit", method: "POST")
            operation TestSimpleUnionWithUnit {
                input := {
                    @required
                    request: SomeUnionWithUnit
                }
                output := {
                    result : SomeUnionWithUnit
                }
                errors: [
                    ValidationException
                ]
            }
            
            @length(min: 13)
            string StringRestricted
            
            union SomeUnionWithUnit {
                Option1: Unit
                Option2: StringRestricted
            }

            @http(uri: "/test", method: "POST")
            operation TestOperation {
                input := { payload: String }
                output := {
                    @httpPayload
                    events: TestEvent
                },
                errors: [ValidationException]
            }
            
            @streaming
            union TestEvent {
                KeepAlive: Unit,
                Response: TestResponseEvent,
            }
            
            structure TestResponseEvent { 
                data: String 
            }            

```


codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/UnionWithUnitTest.kt
```diff
@@ -0,0 +1,144 @@
...
+                KeepAlive: Unit,
+                Response: TestResponseEvent,
+            }
+            
+            structure TestResponseEvent { 
+                data: String 
+            }            
+            """.asSmithyModel()
+
+        serverIntegrationTest(model) { _, rustCrate ->
```
david-perez, line 77: Not sure why we need the tests below if all we want to check is that the model compiles correctly.

david-perez (approved):

david-perez:
Closes #2546.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/UnionWithUnitTest.kt
drganjoo: So that we can extend the same test to RpcV2Cbor (and possibly other protocols). Since event streams are not covered by the protocol tests, and none of the other tests were failing, I assume we do not have a test that actually checks whether data is being correctly serialized/deserialized. However, having said that, I think it is better to add comprehensive tests that use a client to serialize and then use the server to deserialize it. Currently, these tests directly use the inner data types/methods rather than invoking the server. Will remove these tests for now and we can add one in v2Cbor.

ysaito1001 (approved):
Reviewed primarily those in codegen-core

PR #3549, opened 2024-04-03
Fix release issues

jdisanti:
This PR fixes to issues with the release:
1. One of the DynamoDB tests was failing due to environment variables impacting its behavior.
2. The stalled stream protection changelog entries were added in error via merge conflict.

----

### DIFF
--- a/aws/sdk/integration-tests/dynamodb/tests/shared-config.rs
+++ b/aws/sdk/integration-tests/dynamodb/tests/shared-config.rs
@@ -9,6 +9,7 @@ use aws_sdk_dynamodb::config::{
 };
 use aws_smithy_runtime::client::http::test_util::capture_request;
 use http::Uri;
+use std::env;
 
 /// Iterative test of loading clients from shared configuration
 #[tokio::test]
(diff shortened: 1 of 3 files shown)


aws/sdk/integration-tests/dynamodb/tests/shared-config.rs
```diff
@@ -35,6 +36,10 @@ async fn shared_config_testbed() {
 async fn service_config_from_profile() {
     let _ = tracing_subscriber::fmt::try_init();
 
+    // Remove env vars that could interfere with the test
+    env::remove_var("AWS_DEFAULT_REGION");
```
aajtodd, line 40: Should we be more conservative and remove all `AWS_` environment variables?

Velfi (approved):

aajtodd (approved):

ysaito1001 (approved):

aws/sdk/integration-tests/dynamodb/tests/shared-config.rs
rcoh: we should probably do that in the test runner at the codebuild level, but this only impacts this single test so being surgical seems like the more reasonable approach. Maybe we should do these unsets in https://github.com/smithy-lang/smithy-rs/blob/742aae9560230f1398ddbcb68ee0810d3af71828/tools/ci-scripts/check-aws-sdk-services#L16

aajtodd: > but this only impacts this single test so being surgical seems like the more reasonable approach

Agreed I think this is more what I was suggesting is if we should consider this as a common test setup utility. Makes sense to unblock release at the moment.

jdisanti: I think the correct long-term fix is to make it possible to replace the environment via the ConfigLoader. This should work for now.

PR #3493, opened 2024-03-15
Sourcing service config from the environment.

Velfi:
## Motivation and Context

https://github.com/smithy-lang/smithy-rs/issues/2863
https://github.com/awslabs/aws-sdk-rust/issues/1060

## Description

This PR adds a new feature: the ability to source service-specific config from the environment.
This is **only** supported when creating a service config from an `SdkConfig`. I've posted [a guide](https://github.com/smithy-lang/smithy-rs/discussions/3537) to our discussions board.

[This also adds support for setting an endpoint URL in environment config.](https://github.com/smithy-lang/smithy-rs/issues/2863)

## Testing

I have written several tests ensuring config is extracted with the correct precedence.

## Checklist

- [x] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3561, opened 2024-04-05
Revert codegen for broken env config feature

Velfi:


rcoh:
This needs to target the release branch

Velfi:
OK, targeting release branch

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3ExpressDecorator.kt
```diff
@@ -83,25 +80,6 @@ class S3ExpressDecorator : ClientCodegenDecorator {
...
-    override fun extraSections(codegenContext: ClientCodegenContext): List<AdHocCustomization> {
-        return listOf(
-            adhocCustomization<SdkConfigSection.CopySdkConfigToClientConfig> { section ->
-                rust(
-                    """
-                    ${section.serviceConfigBuilder}.set_disable_s3_express_session_auth(
-                        ${section.sdkConfig}
-                            .service_config()
-                            .and_then(|conf| {
-                                let str_config = conf.load_config(service_config_key("AWS_S3_DISABLE_EXPRESS_SESSION_AUTH", "s3_disable_express_session_auth"));
```
rcoh, line 96: Do we want to disable these changes that have no global value? It seems like we could be strategic and only disable the region one. I'm fine with the larger revert though

Velfi: Good point, I'll leave this one in.

ysaito1001 (approved):
The latest codegen diff LGTM

ysaito1001:
Nit: [this](https://github.com/smithy-lang/smithy-rs/pull/3493/files#diff-722851a4ed160fa1d80699fd516f15194371ed7ab52cb80cd9add9d16d7b1e9f) can also be reverted probably, but not critical

PR #3527, opened 2024-03-27
Enable stalled stream protection for uploads behind new behavior version

jdisanti:
This PR creates a new behavior version that enables stalled stream protection for upload streams by default.

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3566, opened 2024-04-05
rfc43: fix identity cache partition

aajtodd:
## Motivation and Context

RFC: https://github.com/smithy-lang/smithy-rs/pull/3544

## Description

Fixes the `SharedCredentialsProvider` and `SharedTokenProvider` to re-use a consistent cache partition by default.
`SdkConfig` does not create an identity cache by default still, that will need to be a follow on PR that gates it behind a new behavior version. Ideally we do that with the stalled stream protection work in https://github.com/smithy-lang/smithy-rs/pull/3527

## Testing

Added new unit and integration tests.

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -10,6 +10,25 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
+[[aws-sdk-rust]]
+message = """
+Fixes the identity resolver types (`credentials_provider()` and `token_provider()`) from `SdkConfig` to have
+a consistent identity cache partition when re-used across different clients.
+"""
+references = ["smithy-rs#3427"]
+meta = { "breaking" = false, "tada" = false, "bug" = true }
+authors = ["aajtodd"]
+
+
+[[smithy-rs]]
+message = """
+`SharedIdentityResolver` now respects an existing cache partition when the `ResolveIdentity` implementation
+provides one already.
+"""
+references = ["smithy-rs#3427"]
+meta = { "breaking" = false, "tada" = false, "bug" = true }
+authors = ["aajtodd"]
+
 [[smithy-rs]]
 message = """
 Stalled stream protection now supports request upload streams. It is currently off by default, but will be enabled by default in a future release. To enable it now, you can do the following:
--- a/aws/sdk/integration-tests/s3/tests/identity-cache.rs
+++ b/aws/sdk/integration-tests/s3/tests/identity-cache.rs
@@ -1,0 +1,119 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use std::sync::atomic::{AtomicI32, Ordering};
+use std::sync::Arc;
+
+use aws_config::{identity::IdentityCache, BehaviorVersion, Region};
+use aws_credential_types::{
+    provider::{future::ProvideCredentials as ProvideCredentialsFuture, ProvideCredentials},
+    Credentials,
+};
+use aws_sdk_s3::Client;
+use aws_smithy_runtime::client::http::test_util::infallible_client_fn;
+
+// NOTE: These tests are _not_ S3 specific and would apply to any AWS SDK but due to the need to consume `aws-config`
+// (which depends on relocated runtime crates) we can't make this an `awsSdkIntegrationTest(..)`.
+
+#[tokio::test]
+async fn test_identity_cache_reused_by_default() {
+    let http_client =
+        infallible_client_fn(|_req| http::Response::builder().status(200).body("OK!").unwrap());
+
+    let provider = TestCredProvider::new();
+    let cache = IdentityCache::lazy().build();
+    let config = aws_config::defaults(BehaviorVersion::latest())
+        .http_client(http_client)
+        .credentials_provider(provider.clone())
+        // TODO(rfc-43) - remove adding a cache when this is the new default
+        .identity_cache(cache)
+        .region(Region::new("us-west-2"))
+        .load()
+        .await;
+
+    let c1 = Client::new(&config);
+    let _ = c1.list_buckets().send().await;
+    assert_eq!(1, provider.invoke_count.load(Ordering::SeqCst));
+
+    let c2 = Client::new(&config);
+    let _ = c2.list_buckets().send().await;
+    assert_eq!(1, provider.invoke_count.load(Ordering::SeqCst));
+}
+
+// TODO(rfc-43) - add no_identity_cache() to ConfigLoader and re-enable test
+// #[tokio::test]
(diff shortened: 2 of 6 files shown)


aws/sdk/integration-tests/s3/tests/identity-cache.rs
```diff
@@ -0,0 +1,116 @@
...
+use aws_config::{identity::IdentityCache, BehaviorVersion, Region};
+use aws_credential_types::{
+    provider::{future::ProvideCredentials as ProvideCredentialsFuture, ProvideCredentials},
+    Credentials,
+};
+use aws_sdk_s3::Client;
+use aws_smithy_runtime::client::http::test_util::infallible_client_fn;
+
+#[tokio::test]
+async fn test_identity_cache_reused_by_default() {
```
ysaito1001, line 18: If an integration test is service agnostic, it's a good practice we write it in Kotlin. Since the test uses `credentials_provider`, you can use `awsSdkIntegrationTest` ([example](https://github.com/smithy-lang/smithy-rs/blob/8bb11b8cc41d4c107d532c165ccc062c59780719/aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/EndpointsCredentialsTest.kt#L81)).

ysaito1001 (commented):
Great work (including RFC)! The only feedback I have is to write tests as Kotlin integration tests.

CHANGELOG.next.toml
```diff
@@ -10,6 +10,14 @@
...
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
+[[smithy-rs]]
+message = """
+Fixes the identity resolver types (`credentials_provider()` and `token_provider()`) from `SdkConfig` to have
+a consistent identity cache partition when re-used across different clients.
+"""
+references = ["smithy-rs#3427"]
+meta = { "breaking" = false, "tada" = false, "bug" = true }
+authors = ["aajtodd"]
```
ysaito1001, line 20: I would make this changelog entry `[[aws-sdk-rust]]` since `credentials_provider` and  `token_provider` are AWS-related concepts, rather than general smithy concepts.

In addition, we can also create a changelog entry for `[[smithy-rs]]` mentioned that `SharedIdentityResolver` now respects a cache partition reserved by the passed-in `IdentityResolver`.

rust-runtime/aws-smithy-runtime-api/src/client/identity.rs
```diff
@@ -222,6 +236,10 @@ impl ResolveIdentity for SharedIdentityResolver {
     fn cache_location(&self) -> IdentityCacheLocation {
         self.inner.cache_location()
     }
+
+    fn cache_partition(&self) -> Option<IdentityCachePartition> {
+        Some(self.cache_partition())
+    }
```
ysaito1001, line 242: Alternatively, if we delegate `cache_partition` to inner, i.e.
```
fn cache_partition(&self) -> Option<IdentityCachePartition> {
    self.inner.cache_partition()
}
```
then maybe we can remove the field `cache_partition` from `SharedIdentityResolver` and `SharedIdentityResolver::new()` will be unchanged?

aws/sdk/integration-tests/s3/tests/identity-cache.rs
aajtodd: Does this end up on every generated client? If so is that what we want here? 

It does use `credentials_provider` but with fake/test credentials. It really is testing the cache behavior and interaction between `SdkConfig` and generated client config.

CHANGELOG.next.toml
aajtodd: Sure I can update it as suggested but I want to take issue with this comment for a second:

> since credentials_provider and token_provider are AWS-related concepts, rather than general smithy concepts.


From Smithy and SRA standpoint they are not AWS SDK specific. They are "AWS" concepts in that they originate from AWS but they are not "AWS SDK" only concepts. A customer could use sigv4 in their own model if they wanted (or internally within Amazon/AWS). The way Rust has this organized now is not entirely correct. Sigv4 and token provider support should be in the Smithy runtime (the specific providers like `SsoTokenProvider` or `DefaultCredentialsChain` are of course AWS SDK specific though).

rust-runtime/aws-smithy-runtime-api/src/client/identity.rs
aajtodd: We could maybe do that but unfortunately there also already exists an accessor function on L221

```rust
    pub fn cache_partition(&self) -> IdentityCachePartition {
        self.cache_partition
    }
```

This is expected to always be present so relying on `inner` which can be `None` seemed like the wrong approach.

CHANGELOG.next.toml
ysaito1001: > From Smithy and SRA standpoint they are not AWS SDK specific

Good point. To be clear, what I meant was a tooling standpoint that's processing `CHANGELOG.next.toml`. When we write a changelog entry for `[smithy-rs]`, that entry only appears in the release notes in the `smithy-rs` repository, but not in the `aws-sdk-rust` repository. So when those who only use Smithy SDK, but not AWS SDK, see a changelog entry mentioning `credentials_provider` and `token_provider` they don't find the methods because they are defined in the AWS runtime crates & generated by AWS codegen decorators. It can confuse customers.

> The way Rust has this organized now is not entirely correct. Sigv4 and token provider support should be in the Smithy runtime

Interesting. Till we address it, I'd be consistent with the way it has been.

rust-runtime/aws-smithy-runtime-api/src/client/identity.rs
ysaito1001: Ah I see, good point.

aws/sdk/integration-tests/s3/tests/identity-cache.rs
ysaito1001: I don't think it does. `awsSdkIntegrationTest` is a test helper that tests codegen itself but does not run against generated AWS SDKs.

> It really is just testing the cache behavior and interaction between SdkConfig and generated client config. It doesn't make a real request.

That does sound like service-agnostic, and that's what `awsSdkIntegrationTest` is for

ysaito1001: Just throwing one last thing on the table before leaving tests in `integration-tests/s3`, can we directly construct `SdkConfig` in tests with necessary fields without going through `aws_config::defaults`? I'm assuming the tests won't need default configs in order to verify what they are supposed to test? If we don't use `aws-config`, then we may be able to avoid the problem we discussed offline.

ysaito1001 (approved):

aws/sdk/integration-tests/s3/tests/identity-cache.rs
aajtodd: We _can_ which would allow us to test some of this but not how different behavior versions behave in loading (which still require a subsequent PR to fully implement) if I understand correctly how it's all wired.

PR #3568, opened 2024-04-08
re-add support for sourcing endpoint URLs from service-specific env config

Velfi:
The service env config work I did was flawed in that it didn't respect the precedence of programmatic config. This PR
adds a way of tracking the precedence up to the point of converting the SdkConfig into a service config. Now, env config will only be resolved when config was not set programmatically. I added tests to `aws-config` for the origin tracking to ensure it works.


aws/rust-runtime/aws-runtime/src/env_config.rs
```diff
@@ -222,11 +262,19 @@ impl<'a> EnvConfigValue<'a> {
...
+            } else if let Some(v) = global_config {
+                tracing::trace!("(global env) {env_var} = {v:?}");
+                Some(v)
+            } else {
+                tracing::trace!("(env) no value set for {env_var}");
+                None
+            }
         });
 
+        tracing::trace!("loading profile config from {profiles:?}");
```
ysaito1001, line 277: Maybe we should return early if `env_value` is `Some` at line 276? The reason for this is that a trace log may look a bit weird if this line says "loading profile..." but at the end of this method `env_value.or(profile_value)` ends up using a value from `env_value`.

aws/rust-runtime/aws-types/src/sdk_config.rs
```diff
@@ -96,6 +98,7 @@ pub struct Builder {
     use_dual_stack: Option<bool>,
     behavior_version: Option<BehaviorVersion>,
     service_config: Option<Arc<dyn LoadServiceConfig>>,
+    config_origins: HashMap<&'static str, Origin>,
```
aajtodd, line 101: I don't actually hate this FWIW, it's purpose is narrowly scoped.

aws/rust-runtime/aws-runtime/src/env_config.rs
Velfi: I'll remove the two "loading ..." messages and keep the log-before-return messages.

aws/rust-runtime/aws-types/src/sdk_config.rs
```diff
@@ -822,6 +845,25 @@ impl SdkConfig {
         self.clone().into_builder()
     }
 
+    /// Set the origin of a setting.
+    ///
+    /// This is used internally to understand how to merge config structs while
+    /// respecting precedence of origins.
+    pub fn insert_origin(&mut self, setting: &'static str, origin: Origin) {
```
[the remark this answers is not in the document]
Velfi: Good question. I think so, but I'm not sure. For now, I'll make it so that:
- SdkConfig's Builder exposes an `insert_origin` method.
- SdkConfig itself only exposes a way to check origins.

We can add other stuff later if/when we need it.

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -950,6 +954,81 @@ mod loader {
...
+                        .build(),
+                )
+                .load()
+                .await;
+            assert_eq!(
+                Origin::shared_profile_file(),
+                loader.get_origin("endpoint_url")
+            );
+        }
+
```
ysaito1001, line 1031: The above tests are mutually exclusive in terms of where a config value comes from, which is great to have. Should we also add tests verifying the expected origin gets returned when config values are specified in multiple sources?

aws/rust-runtime/aws-types/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-types"
-version = "1.1.9"
+version = "1.1.10"
```
ysaito1001, line 3: yeah, I think a patch version bump is reasonable even though we've added a new `pub` item, `Origin`, since it's really meant to be used internally and not by customers.

aws/rust-runtime/aws-types/src/origin.rs
```diff
@@ -0,0 +1,253 @@
...
+        use Inner::*;
+
+        if self.is_unknown() || other.is_unknown() {
+            return None;
+        }
+
+        match self {
+            // IMDS is the lowest priority
+            Imds => Some(Ordering::Less),
+            // Programmatic is the second-lowest priority
```
ysaito1001, line 149:
```suggestion
            // ProfileFile is the second-lowest priority
```

ysaito1001 (commented):
The changes look great. I'll probably go through one more round to be sure.

aws/rust-runtime/aws-config/src/lib.rs
Velfi: I did this for 2/3 tests [^1]. As for the profile file test, I'm not sure how I would except by stubbing out some IMDS thing since that's the only thing with a lower precedence.

[^1]: The `test_origin_env` and `test_origin_programmatic` tests also set profile config, which is then overridden.

aws/rust-runtime/aws-types/src/origin.rs
Velfi: good catch

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SdkConfigDecorator.kt
```diff
@@ -59,6 +61,31 @@ object SdkConfigCustomization {
...
+        rustTemplate(
+            """
+            if ${section.sdkConfig}.get_origin("endpoint_url").is_client_config() {
+                ${section.serviceConfigBuilder}.set_$fieldName(${section.sdkConfig}.$fieldName()#{map});
+            } else {
+                ${section.serviceConfigBuilder}.set_$fieldName(
+                    ${section.sdkConfig}
+                        .service_config()
+                        .and_then(|conf| conf.load_config(service_config_key($envKey, $profileKey)).map(|it| it.parse().unwrap()))
+                        .or_else(|| ${section.sdkConfig}.$fieldName()#{map})
```
ysaito1001, line 82: If I understand it correctly, `shared_environment_variable` (a.k.a global env) takes precedence over `service_profile_file`. If a user configures a config value in those two places only, will this code respect `shared_environment_variable`?

Velfi: yeah

ysaito1001: That right, I misread the code. `${section.sdkConfig}.service_config()` returns `EnvServiceConfig` (not service-specific-only config), so it knows how to handle the said precedence.

ysaito1001:
Lastly, we probably need to a changelog entry that revises or corrects release notes that we put out in https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2024-04-08 and [the discussion topic](https://github.com/smithy-lang/smithy-rs/discussions/3537), since the support is currently for endpoint url.

Velfi:
> Lastly, we probably need to have a changelog entry that revises or corrects release notes that we put out in https://github.com/awslabs/aws-sdk-rust/releases/tag/release-2024-04-08 and [the discussion topic](https://github.com/smithy-lang/smithy-rs/discussions/3537), since the support is currently for endpoint url.

I updated the discussion. I think the Changelog entry is fine:
> 🎉 (https://github.com/smithy-lang/smithy-rs/issues/3493) Users may now set service-specific configuration in the environment. For more information, see https://github.com/smithy-lang/smithy-rs/discussions/3537.

ysaito1001 (approved):

PR #3571, opened 2024-04-10
Implement the waiter matcher generator

jdisanti:
This PR implements Smithy waiter matcher union codegen.

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGenerator.kt
@@ -36,7 +36,7 @@ private typealias Scope = Array<Pair<String, Any>>
 
 /**
  * Generates the Rust code for the Smithy waiter "matcher union".
- * See https://smithy.io/1.0/spec/waiters.html#matcher-union
+ * See https://smithy.io/2.0/additional-specs/waiters.html#matcher-union
  */
 class RustWaiterMatcherGenerator(
     private val codegenContext: ClientCodegenContext,
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGenerator.kt
@@ -1,0 +1,190 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.rust.codegen.client.smithy.generators.waiters
+
+import software.amazon.smithy.codegen.core.CodegenException
+import software.amazon.smithy.codegen.core.Symbol
+import software.amazon.smithy.jmespath.JmespathExpression
+import software.amazon.smithy.model.node.Node
+import software.amazon.smithy.model.shapes.Shape
+import software.amazon.smithy.rust.codegen.client.smithy.ClientCodegenContext
+import software.amazon.smithy.rust.codegen.client.smithy.ClientRustModule
+import software.amazon.smithy.rust.codegen.core.rustlang.RustModule
+import software.amazon.smithy.rust.codegen.core.rustlang.RustType
+import software.amazon.smithy.rust.codegen.core.rustlang.RustWriter
+import software.amazon.smithy.rust.codegen.core.rustlang.replaceLifetimes
+import software.amazon.smithy.rust.codegen.core.rustlang.rust
+import software.amazon.smithy.rust.codegen.core.rustlang.rustBlockTemplate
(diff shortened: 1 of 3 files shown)


codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGeneratorTest.kt
```diff
@@ -0,0 +1,349 @@
...
+                    """
+                    /// Make the unit test public and document it so that compiler
+                    /// doesn't complain about dead code.
+                    pub fn ${name}_test_case() {
+                        #{test}
+                    }
+                    ##[cfg(test)]
+                    ##[test]
+                    fn $name() {
+                        ${name}_test_case();
```
aajtodd, line 73: **question**: Why not just generate the test body `#{test}` here and avoid the need to make it public/documented?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGenerator.kt
```diff
@@ -0,0 +1,190 @@
...
+        val scope =
+            arrayOf(
+                *preludeScope,
+                "Input" to inputSymbol,
+                "Output" to outputSymbol,
+                "Error" to errorSymbol,
+                "ProvideErrorMetadata" to RuntimeType.provideErrorMetadataTrait(runtimeConfig),
+            )
+        return RuntimeType.forInlineFun(fnName, module) {
+            rustBlockTemplate("pub(crate) fn $fnName(_input: &#{Input}, _result: &#{Result}<#{Output}, #{Error}>) -> bool", *scope) {
```
aajtodd, line 65: **question**: So we're generating functions like `pub(crate) fn match_test_operation_1555d9862d8934ff9(...)`, can we generate any kind of provenance doc comment that ties it to the model (to aid in debugging/traceability)? 


In Kotlin these matcher functions are generated inline where they are used so they don't need an identifier. I think it's fine to generate them this way but I am wondering more generally what is the design for how these matches will be consumed? What will a generated waiter look like? How will these matchers be tied to an acceptor and retry strategy?

aajtodd (commented):
Overall looks good, couple questions.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGeneratorTest.kt
jdisanti: The compiler is unhappy about the `pub(crate)` function under test. Since its only used in a unit test, it considers it dead code, and since we have warnings as errors, this causes the test to fail.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGenerator.kt
jdisanti: I'll add a comment above the function with the matcher json, at the very least. I imagine it will be pretty easy to identify them if you're going from the direction of waiter impl -> matcher. The acceptor/retry strategy will be calling these functions directly.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGeneratorTest.kt
aajtodd: Ahh I see

aajtodd (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustWaiterMatcherGenerator.kt
```diff
@@ -0,0 +1,192 @@
...
+import software.amazon.smithy.waiters.Matcher.OutputMember
+import software.amazon.smithy.waiters.Matcher.SuccessMember
+import software.amazon.smithy.waiters.PathComparator
+import java.security.MessageDigest
+
+private typealias Scope = Array<Pair<String, Any>>
+
+/**
+ * Generates the Rust code for the Smithy waiter "matcher union".
+ * See https://smithy.io/1.0/spec/waiters.html#matcher-union
```
ysaito1001, line 39:
```suggestion
 * See https://smithy.io/2.0/spec/waiters.html#matcher-union
```

ysaito1001, line 39:
```suggestion
 * See https://smithy.io/2.0/additional-specs/waiters.html#matcher-union
```

jdisanti: Thanks! Didn't realize I was looking at the 1.0 spec 😅

PR #3575, opened 2024-04-10
Fix canary execution during prod release

jdisanti:
The new canary secrets weren't being propagated from the root prod release workflow into the CI workflow, so the canary was failing. This PR propagates those secrets, and also makes it possible to run the full CI workflow in a dry-run release so that it's possible to test in the future.

----


ysaito1001 (approved):

jdisanti:
Successful canary run: https://github.com/smithy-lang/smithy-rs/actions/runs/8639272994/job/23685586028

Issue #3471, opened 2024-03-07
@restJson1 and charset=utf8 in the header


x1a0:
Hi,

Probably a stupid question - I have a service defined with `@restJson1`. The incoming requests that have header `Content-Type: application/json` work fine.
However if they carry header `Content-Type: application/json; charset=utf8`, the server just returns:
```
415 Unsupported Media Type
```

Is there anyway to configure the header value matching somehow?
Thanks in advance.

x1a0:
Looks like `charset` should not be set for `application/json`. I will add a layer to remove it.

PR #3576, opened 2024-04-11
Allow expected content-type to ignore parameters

djedward:
## Motivation and Context
An issue was raised about a mobile client that appends "; charset=utf-8" to the Content-Type when using restJson1. The [latest RFC](https://www.rfc-editor.org/rfc/rfc8259) for "application/json" does not register a charset parameter, but indicates it is reasonable to accept it.

## Description
This change loosens the validation of the expected content type to allow all parameters. I figured if the type and subtype match, we can leave it up to the parsing of the data to figure out whether it can read it. it is just as easy for a client to specify "application/json" and then provide invalid json data.

## Testing
* Add an additional test for the use case
* ran the runtime and codegen tests
* Tested with the example server, passing requests with/without the charset set.
* Added a test for each protocol

## Checklist

----


CHANGELOG.next.toml
```diff
@@ -10,6 +10,15 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
+
+[[aws-sdk-rust]]
```
drganjoo, line 14: This should be `[[smithy-rs]]`. Could you please also add `"target" = "server"` to the meta tags. The change log tool will automatically list changes for the server under a separate heading in the release `CHANGELOG.md`.

Thanks.

CHANGELOG.next.toml
```diff
@@ -10,6 +10,15 @@
...
+message = """
+Content-Type header validation now ignores parameter portion of media types.
```
david-perez, line 16:
```suggestion
`Content-Type` header validation now ignores parameter portion of media types.
```

CHANGELOG.next.toml
```diff
@@ -10,6 +10,15 @@
...
+"""
+references = ["smithy-rs#3576"]
```
david-perez, line 18: @drganjoo found this old issue; apparently someone had identified this bug previously and we didn't notice because they adjusted the client-side! We should link it too: https://github.com/smithy-lang/smithy-rs/issues/3471

rust-runtime/aws-smithy-http-server/src/protocol/mod.rs
```diff
@@ -257,6 +257,13 @@ mod tests {
         assert!(matches!(result.unwrap_err(), MissingContentTypeReason::ToStrError(_)));
     }
 
+    #[test]
+    fn valid_params_content_type() {
```
david-perez, line 261: nit:

```suggestion
    fn valid_content_type_header_classifier_params() {
```

just to match the naming pattern of the other tests that also contain the name of the function being tested

rust-runtime/aws-smithy-http-server/src/protocol/mod.rs
```diff
@@ -107,7 +107,7 @@ fn content_type_header_classifier(
             .parse::<mime::Mime>()
             // `expected_content_type` comes from the codegen.
             .expect("BUG: MIME parsing failed, `expected_content_type` is not valid. Please file a bug report under https://github.com/smithy-lang/smithy-rs/issues");
-        if expected_content_type != found_mime {
+        if expected_content_type != found_mime && expected_content_type != found_mime.essence_str() {
```
david-perez, line 110: Nice, I didn't know about `essence_str()`. Can't we simplify this to just the latter condition though?

```suggestion
        if expected_content_type != found_mime.essence_str() {
```

We control `expected_content_type` and it's a `&str` with just the essence string.  So if the latter pair are not equal the first pair should also not be equal.

david-perez (commented):
We should add a protocol test for each of Smithy's protocols, given that `content_type_header_classifier` affects all protocols (i.e. `Content-Type` header comparison will now only take into account essence strings in all protocols).

We should contribute the test upstream to https://github.com/smithy-lang/smithy so all code generators benefit; to test it works against smithy-rs you can add it to `codegen-core/common-test-models/rest-json-extras.smithy` and confirm it passes locally with `./gradlew codegen-server-test:build -P modules='rest_json_extras'`.

rust-runtime/aws-smithy-http-server/src/protocol/mod.rs
djedward: I did that originally, but was wondering if someone would ever want to force a specific check. It seemed unlikely, but added the original check back in, just in case. Given the discussion on it today, I think just the essence_str check is enough.

rust-runtime/aws-smithy-http-server/src/protocol/mod.rs
djedward: I matched it to the ones above that were doing something_something_content_type, but happy to rename.

CHANGELOG.next.toml
djedward: ah, I missed that. Should have picked up on it when adding the "target" had failed validation.

drganjoo (approved):

CHANGELOG.next.toml
djedward: I've added this and the tests requested above. I would still need to upstream to smithy-rs, but figured I'd get them here first and these can be the model for upstream.

david-perez (approved):
We should make a note / reach out to Smithy team about upstreaming the protocol tests.

djedward:
I've written the upstream tests. I just need to reach out to the smithy team to figure out how the validate those tests across the implementations.

djedward:
Created a new pull request with just the necessary changes given upstream changes: https://github.com/smithy-lang/smithy-rs/pull/3724

PR #3578, opened 2024-04-11
create an IdentityCache by default when on latest BehaviorVersion

aajtodd:
## Motivation and Context
RFC: https://github.com/smithy-lang/smithy-rs/pull/3544

## Description
Make `ConfigLoader` create an `IdentityCache` by default when on the latest `BehaviorVersion`. This will allow SDK clients created from `SdkConfig` to share a cache by default instead of creating one per/client. 

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/lib.rs
+++ b/aws/rust-runtime/aws-config/src/lib.rs
@@ -214,6 +214,7 @@ mod loader {
     use aws_credential_types::Credentials;
     use aws_smithy_async::rt::sleep::{default_async_sleep, AsyncSleep, SharedAsyncSleep};
     use aws_smithy_async::time::{SharedTimeSource, TimeSource};
+    use aws_smithy_runtime::client::identity::IdentityCache;
     use aws_smithy_runtime_api::client::behavior_version::BehaviorVersion;
     use aws_smithy_runtime_api::client::http::HttpClient;
     use aws_smithy_runtime_api::client::identity::{ResolveCachedIdentity, SharedIdentityCache};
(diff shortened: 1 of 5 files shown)


aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -238,14 +239,14 @@ mod loader {
     use crate::provider_config::ProviderConfig;
 
     #[derive(Default, Debug)]
-    enum CredentialsProviderOption {
-        /// No provider was set by the user. We can set up the default credentials provider chain.
+    enum TriStateOption<T> {
```
jdisanti, line 242: We used to have a TriState in aws-smithy-types. You could consider resurrecting that if you want to, but doesn't matter to me either way since this is private.

aws/rust-runtime/aws-config/src/lib.rs
```diff
@@ -439,7 +440,29 @@ mod loader {
...
+        /// Turn off creating a default identity cache:
+        /// ```no_run
+        /// # async fn create_config() {
+        /// let config = aws_config::from_env()
+        ///     .no_identity_cache()
+        ///     .load()
+        ///     .await;
+        /// # }
+        /// ```
+        pub fn no_identity_cache(mut self) -> Self {
```
jdisanti, line 464: Does this behave differently from `loader.identity_cache(IdentityCache::no_cache())`?

aws/rust-runtime/aws-config/src/lib.rs
aajtodd: If we have need for another tri-state then we can move it but for now I think I'd keep it private.

aws/rust-runtime/aws-config/src/lib.rs
aajtodd: Yes. `IdentityCahe::no_cache()` is still an explicit cache being set whereas `no_identity_cache()` leaves it as "explicitly unset" so as to not enable the default cache. Open to bikeshedding on naming.

aajtodd: Discussed offline and decided to remove this API for now as it can always be added later and users can still control this by setting a different cache per/client.

jdisanti (approved):

PR #3581, opened 2024-04-11
Fix runtime-versioner audit edge case with SDK releases

jdisanti:
In the time gap between smithy-rs and SDK releases, the runtime-versioner can incorrectly allow changes to aws/rust-runtime crates without version bumping. This PR fixes this issue by detecting this edge case.

----


ysaito1001 (approved):
Producing the appropriate error message takes some effort 🚢

PR #3583, opened 2024-04-12
rfc to disable payload signing

aajtodd:
## Motivation and Context
Original issue: https://github.com/awslabs/aws-sdk-rust/issues/1087

## Description
RFC-44 proposes a new `disable_payload_signing()` operation customization that would allow users to disable Sigv4(a) payload signing for operation requests that target an `@streaming blob` shape.

----

### DIFF
--- a/design/src/rfcs/rfc0044_disable_payload_signing.md
+++ b/design/src/rfcs/rfc0044_disable_payload_signing.md
@@ -1,0 +1,86 @@
+RFC: Disable Payload Signing
+=============================
+
+> Status: RFC
+>
+> Applies to: client
+
+For a summarized list of proposed changes, see the [Changes Checklist](#changes-checklist) section.
+
+This RFC proposes a way for users to disable payload signing for requests with a streaming binary payload.
+
+
+The user experience if this RFC is implemented
+----------------------------------------------
+
+
+In the current version of the SDK, users are unable to control how a request is signed.
+Most requests must include a hash of the payload as part of the signature. However, requests with a
+[streaming](https://smithy.io/2.0/spec/streaming.html#smithy-api-streaming-trait) `blob` shape as the payload can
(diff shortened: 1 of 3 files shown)


design/src/rfcs/rfc0044_disable_payload_signing.md
```diff
@@ -0,0 +1,86 @@
...
+
+The user experience if this RFC is implemented
+----------------------------------------------
+
+
+In the current version of the SDK, users are unable to control how a request is signed.
+Most requests must include a hash of the payload as part of the signature. However, requests with a
+[streaming](https://smithy.io/2.0/spec/streaming.html#smithy-api-streaming-trait) `blob` shape as the payload can
+sometimes (e.g. S3 `PutObject`) choose to skip payload hashing and exclude it from the signature. When calculating the
+signature the body hash is instead set to the literal constant `UNSIGNED-PAYLOAD`. This has the advantage of not paying
```
jdisanti, line 21: What services does this actually work for? I was under the impression this only worked for S3.

jdisanti (commented):
RFC looks good. Just one question.

ysaito1001 (commented):
RFC looks good. Will the RFC briefly describe how `DisablePayloadSigningPlugin` is implemented?

design/src/rfcs/rfc0044_disable_payload_signing.md
jdisanti: Yeah, I think we should allow list this until we've verified it actually works in other services.

jdisanti (approved):

design/src/rfcs/rfc0044_disable_payload_signing.md
aajtodd: I'm good with that, will update the RFC to mention this.

aajtodd:
implemented in https://github.com/smithy-lang/smithy-rs/pull/3915

PR #3586, opened 2024-04-15
Add timeouts to CI jobs

jdisanti:
This PR adds reasonable timeouts to each CI job so that a PR isn't held up for six hours by a hung runner.

----

### DIFF
--- a/.github/workflows/ci-merge-queue.yml
+++ b/.github/workflows/ci-merge-queue.yml
@@ -22,6 +22,7 @@ jobs:
   # The login password is encrypted with the repo secret DOCKER_LOGIN_TOKEN_PASSPHRASE
   save-docker-login-token:
     name: Save a docker login token
+    timeout-minutes: 10
     outputs:
       docker-login-password: ${{ steps.set-token.outputs.docker-login-password }}
     permissions:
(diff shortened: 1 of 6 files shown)


Velfi (approved):

.github/workflows/ci-merge-queue.yml
```diff
@@ -51,6 +52,7 @@ jobs:
     name: Acquire Base Image
     needs: save-docker-login-token
     runs-on: ubuntu-latest
+    timeout-minutes: 10
```
ysaito1001, line 55: This timeout for acquire base image is shorter than those in `ci-main` or `ci-pr-forks`, because by the time a PR is merged to main, there's already a base image existing and we don't need to build it?

ysaito1001 (approved):
Thanks for adding timeout extensively! If a job starts failing due to a genuine, non-transient, reason, then these timeouts will help detect we did something wrong.

.github/workflows/ci-merge-queue.yml
jdisanti: I was thinking CI on main wouldn't ever need to rebuild the build image, but actually, that's not true since it will need to rebuild it for PRs from forks. I'll bring it up to the same value as the others.

jdisanti, line 55:
```suggestion
    timeout-minutes: 60
```

.github/workflows/ci-pr.yml
```diff
@@ -99,6 +101,7 @@ jobs:
   # The PR bot requires a Docker build image, so make it depend on the `acquire-base-image` job.
   pr_bot:
     name: PR Bot
+    timeout-minutes: 30
```
jdisanti, line 104:
```suggestion
```

PR #3587, opened 2024-04-15
Restructure codegen modules, Add JReleaser for publishing

haydenbaker:
## Motivation and Context
- We want to get the server codegen jars released in some form so that we can easily use rust server codegen in a tutorial
- Prior to publishing, the repo (specifically codegen) needs to be restructured such that it is consistent with how other codegen is organized
- NOTE: docs still need to be updated in this PR

## Description
- Adds [JReleaser](https://jreleaser.org) and associated configuration to easily release jars to maven-central (specifically for `codegen-server`)
- Restructures codegen modules (flattened)
- Refactor the root gradle build to correctly configure the codegen modules that should have publications:
  - `smithy-aws-rust-codegen`, `smithy-rust-codegen-core`, `smithy-rust-codegen-client`, `smithy-rust-codegen-server`

### E2E
- Run all of the commands given in the README (typescript fails, tested on main branch and it fails too), make sure pass
- Run and make sure pass: `./gradlew clean smithy-rust-codegen-core:test smithy-rust-codegen-client:test smithy-rust-codegen-client-test:test smithy-rust-codegen-server:test smithy-rust-codegen-server-test:test`

### Publications
- Run `./gradlew clean publish`, and the publications should be published under `build/staging` (list above)
- Ensure that each publication has jar, sources-jar, javadoc-jar, module-file, pom-file

### JReleaser
- Run `./gradlew jreleaserConfig`, and you'll see that it tries to configure itself
  - It will fail unless the right environment vars are set (e.g. tokens, gpg keys, etc.)
- The `jreleaserConfig` task should be run as a pre-req for releasing
- Releasing with the tool can be accomplished by running the `jreleaserFullRelease` task
- The trace-log for JReleaser is located under `build/jreleaser/trace.log`

## Checklist

----


codegen-server/build.gradle.kts
```diff
@@ -4,10 +4,12 @@
  */
 
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.jreleaser.model.Active
 
 plugins {
     kotlin("jvm")
     `maven-publish`
+    id("org.jreleaser") version "1.11.0"
```
hpmellema, line 12: i think we will also need to add and configure the `signing` plugin.

aajtodd: What is jreleaser giving us here?

aajtodd (changes requested):
There are two big issues I see at the moment:

1. The group name

The group name is currently set to `software.amazon.smithy.rust.codegen.server.smithy`

This is overly specific. It should follow the other code generators like `typescript`, `kotlin`, `ruby`, and `dafny` where the group is `software.amazon.smithy.<language>` and use module names appropriately. See examples [here](https://repo1.maven.org/maven2/software/amazon/smithy/)

e.g. we should probably end up with something like:

```
software.amazon.smithy.rust/              // group
     smithy-rs-codegen-server             // artifact name
     smithy-rs-codegen-client              // artifact name
     smithy-rs-codegen-core               // artifact name
```

We _could_ maybe leave the artifact names alone since they will be namespaced with `rust` in the group name but if we want to be more consistent across languages the artifact names would need updated as well.


2. Dependencies

Both `codegen-server` and `codegen-client` have a dependency on `codegen-core` which would need to be published as well. 

----

Some questions:

1. Who is going to be publishing this? Is this just a one off?
2. Why not commonize publishing setup and just publish the client as well?

codegen-server/build.gradle.kts
haydenbaker: We don't need it - JReleaser will sign the artifacts in staging

haydenbaker:
1. Yes, you're right about the group name, I totally missed that.
2. WRT publishing everything, it sounds like that will be what we need to do.

The goal of this PR is to add everything that is needed to publish the jars (easily). JReleaser makes it super straightforward, as you run a couple commands and it's published to maven (no sonatype UI, etc).

aajtodd:
Closing as I believe this is no longer being pursued and we have some concerns with how we land this.

PR #3465, opened 2024-03-06
Add support for S3 Express One Zone

ysaito1001:
## Motivation and Context
Allows the Rust SDK to use [S3 Express One Zone](https://aws.amazon.com/s3/storage-classes/express-one-zone/)

## Description
The PR adds the said S3-specific functionality to the Rust SDK. The code changes have already been reviewed by previous sub PRs, but it's worth going through them again as a whole:
- https://github.com/smithy-lang/smithy-rs/pull/3386
- https://github.com/smithy-lang/smithy-rs/pull/3388
- https://github.com/smithy-lang/smithy-rs/pull/3390
- https://github.com/smithy-lang/smithy-rs/pull/3432
- https://github.com/smithy-lang/smithy-rs/pull/3433
- https://github.com/smithy-lang/smithy-rs/pull/3459
- https://github.com/smithy-lang/smithy-rs/pull/3457
- https://github.com/smithy-lang/smithy-rs/pull/3462

In addition to the PRs above, commit eebe8af increases the canary lambda's memory size to 512MB from 128MB (also makes it configurable through a command line arg for `canary-runner`). By default, lambda's allowed memory size is 128MB but with the addition of `canary-wasm` in main, canary lambda's memory usage will be 152MB, causing the lambda to be killed by a signal during runtime. The commit addresses that issue.

## Testing
- Unit tests in [aws/rust-runtime/aws-inlineable/src/s3_express.rs](https://github.com/smithy-lang/smithy-rs/blob/7f8c28b7038372927ec6196eff88384452f908dd/aws/rust-runtime/aws-inlineable/src/s3_express.rs)
- Integration tests in [aws/sdk/integration-tests/s3/tests/express.rs](https://github.com/smithy-lang/smithy-rs/blob/7f8c28b7038372927ec6196eff88384452f908dd/aws/sdk/integration-tests/s3/tests/express.rs)
- Canary in smithy-rs#3462

## Checklist

- [x] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3590, opened 2024-04-18
Fix compiling S3 crate for wasm

ysaito1001:
## Motivation and Context
Running `cargo build --target wasm32-unknown-unknown --no-default-features` on an S3 crate has stopped working since https://github.com/smithy-lang/smithy-rs/pull/3465 with the following error:
```
error: the wasm*-unknown-unknown targets are not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /Users/[REDACTED]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.14/src/lib.rs:352:9
    |
352 | /         compile_error!("the wasm*-unknown-unknown targets are not supported by \
353 | |                         default, you may need to enable the \"js\" feature. \
354 | |                         For more information see: \
355 | |                         https://docs.rs/getrandom/#webassembly-support");
    | |________________________________________________________________________^
```

To address the issue, this PR updates an S3's dependency on `ahash` in a way that disables default features.

## Testing
Updated the existing test `integration-tests/webassembly` so that `check-aws-sdk-standalone-integration-tests` will run `cargo check` `aws-sdk-s3` against both `wasm32-wasi` and `wasm32-unknown-unknown` (the updated check would break if we removed `default-features = false` from the `ahash` dependency).

## Checklist

----

### DIFF
--- a/aws/sdk/integration-tests/test.sh
+++ b/aws/sdk/integration-tests/test.sh
@@ -16,6 +16,9 @@ for f in *; do
       else
          # The webassembly tests use a custom runner set in config.toml that
          # is not picked up when running the tests outside of the package
+         # The tests are written for `wasm32-wasi` but the manifest config also specifies
+         # `wasm32-unknown-unknown` so we can ensure the test build on that platform as well.
+         # For executing the tests, however, we explicitly choose a target `wasm32-wasi`.
          cd webassembly && cargo component test --all-features --target wasm32-wasi && cd ..
       fi
    fi
--- a/aws/sdk/integration-tests/test.sh
+++ b/aws/sdk/integration-tests/test.sh
@@ -16,7 +16,7 @@ for f in *; do
       else
          # The webassembly tests use a custom runner set in config.toml that
          # is not picked up when running the tests outside of the package
-         cd webassembly && cargo component test --all-features --all-targets && cd ..
+         cd webassembly && cargo component test --all-features --target wasm32-wasi && cd ..
       fi
    fi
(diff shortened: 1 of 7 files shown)


aws/sdk/integration-tests/test.sh
```diff
@@ -16,7 +16,7 @@ for f in *; do
       else
          # The webassembly tests use a custom runner set in config.toml that
          # is not picked up when running the tests outside of the package
-         cd webassembly && cargo component test --all-features --all-targets && cd ..
+         cd webassembly && cargo component test --all-features --target wasm32-wasi && cd ..
```
jdisanti, line 19: Shouldn't this still be all targets?

jdisanti (approved):

aws/sdk/integration-tests/test.sh
ysaito1001: I suspect "executing" `integration-tests/webassembly` is specifically for `wasm32-wasi`, but not `wasm32-unknown-unknown` that is mainly for browsers. Previously `all-targets` just included `wasm32-wasi` because that was the only target included in the cargo config for `integration-tests/webassembly`. But now that they are two of them, this line explicitly chooses `wasm32-wasi`. If that's true, maybe I should add a comment?

ysaito1001: Added explanatory comment in 8358943.

PR #3593, opened 2024-04-18
Fix several waiter path matcher issues

jdisanti:
While implementing waiters in another branch, I discovered and fixed a number of issues with the path matcher codegen logic. These issues were:

- Generated code for nested flatten projections failed to compile due to the first projection producing a `Vec<Vec<&T>>` instead of `Vec<&T>`.
- Path matchers that don't use input were taking input as an argument anyway, which results in an unnecessary clone of the input when used by the generated waiter logic.
- The comparisons generated by `RustWaiterMatcherGenerator` would fail to compile whenever comparing a string against an enum.

This PR fixes all these issues.

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustJmespathShapeTraversalGenerator.kt
```diff
@@ -790,7 +805,13 @@ private fun RustType.isNumber(): Boolean = this.dereference().let { it is RustTy
 
 private fun RustType.isBool(): Boolean = this.dereference() is RustType.Bool
 
-private fun RustType.isDoubleReference(): Boolean = this is RustType.Reference && this.member is RustType.Reference
+private fun RustType.isDoubleReference(): Boolean = this.refCount() > 1
+
+private fun RustType.refCount(): Int =
```
jdisanti, line 810: I can probably revert this change. I was going to use `refCount`, but then didn't end up going down that path.

aajtodd (approved):

ysaito1001 (approved):

PR #3594, opened 2024-04-19
Revert cargo-semver-checks version due to performance regression

jdisanti:
It seems upgrading cargo-semver-checks to 0.30 made it take significantly longer than before (going from less than 20 minutes up to between 1-4 hours). Reverting it back to 0.24.1 for now to keep CI faster.

----


ysaito1001 (approved):

ysaito1001:
Let's see if semver compliance check passes in this PR with the new timeout.

PR #3595, opened 2024-04-19
Implement waiters

jdisanti:
This PR implements waiters according to the [Smithy waiters spec](https://smithy.io/2.0/additional-specs/waiters.html).

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentBuilderGenerator.kt
```diff
@@ -31,10 +33,43 @@ import software.amazon.smithy.rust.codegen.core.smithy.rustType
...
+/**
+ * Codegen configuration for the fluent builder generator.
+ *
+ * Null values indicate "use the default", and defaults are oriented towards
+ * the main fluent builders used by the client.
+ *
+ * These can be overridden for secondary fluent builders, such as used by
+ * waiters.
+ */
+interface FluentBuilderConfig {
```
aajtodd, line 45: This is fine for now but I wonder if it would be nice if things like pagination and waiters could just be customizations and that the base fluent builder didn't need to know about such things (and instead just provided the hooks necessary).

aws/sdk/integration-tests/ec2/tests/waiters.rs
```diff
@@ -0,0 +1,84 @@
...
+        .await
+        .unwrap();
+
+    let waiter_task = tokio::spawn(
+        ec2.wait_for_instance_status_ok()
+            .instance_ids("i-09fb4224219ac6902")
+            .wait(Duration::from_secs(300)),
+    );
+
+    time_source.tick(Duration::from_secs(305)).await;
```
aajtodd, line 50: **question**: Why do we need to advance this far? I would assume the waiter finishes before this in the success case and I found this slightly odd when reading the test.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/ServiceErrorGenerator.kt
```diff
@@ -77,6 +78,30 @@ class ServiceErrorGenerator(
...
+            if (operations.any { it.hasTrait<WaitableTrait>() }) {
+                rustTemplate(
+                    """
+                    impl<O, E> #{From}<#{WaiterError}<O, E>> for Error
+                    where
+                        O: #{Debug} + #{Send} + #{Sync} + 'static,
+                        E: #{StdError} + #{Send} + #{Sync} + 'static,
+                    {
+                        fn from(err: #{WaiterError}<O, E>) -> Self {
+                            Error::Unhandled(#{Unhandled} {
```
aajtodd, line 91: **question**: It looks like `Error::Unhandled` is marked `deprecated` in the generated code and is documented as "not forwards compatible". Is this what we want to map these to? (I don't have a better suggestion but I am curious on the rationale on why this is the right choice).

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/WaitableGenerator.kt
```diff
@@ -0,0 +1,258 @@
...
+import software.amazon.smithy.rust.codegen.core.smithy.RustCrate
+import software.amazon.smithy.rust.codegen.core.util.getTrait
+import software.amazon.smithy.rust.codegen.core.util.orNull
+import software.amazon.smithy.rust.codegen.core.util.outputShape
+import software.amazon.smithy.rust.codegen.core.util.toPascalCase
+import software.amazon.smithy.rust.codegen.core.util.toSnakeCase
+import software.amazon.smithy.waiters.WaitableTrait
+import software.amazon.smithy.waiters.Waiter
+
+class WaitableGenerator(
```
aajtodd, line 34: docs

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/WaitableGenerator.kt
```diff
@@ -0,0 +1,258 @@
...
+                            "FluentBuilder" to spec.fluentBuilder,
+                        )
+                    }
+                }
+            }
+        }
+    }
+
+    private fun RustWriter.renderWaiterFnDeclaration(spec: WaiterSpec) {
+        val fnName = "wait_for_${spec.waiterName.toSnakeCase()}"
```
aajtodd, line 105: **discuss**

This name is fine I suppose (especially since there doesn't seem to be a recommended naming pattern) but I do wonder if it makes sense to align with other SDKs here.

It looks like [Go v1](https://github.com/aws/aws-sdk-go/blob/main/service/ec2/waiters.go#L16), [Java v2](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/waiters.html#id1waiters), [Swift](https://github.com/awslabs/aws-sdk-swift/blob/main/Sources/Services/AWSS3/Waiters.swift#L34), [Kotlin](https://docs.aws.amazon.com/sdk-for-kotlin/latest/developer-guide/waiters.html), [Ruby](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/using-waiters.html), [PHP](https://docs.aws.amazon.com/en_us/aws-sdk-php/guide/latest/guide/waiters.html), and possibly others use `wait until` instead of `wait for`.  Of course there are outliers (Python and Go v2, didn't look at others). My vote would be to align with `wait_until` though.

rust-runtime/aws-smithy-runtime-api/src/client/waiters.rs
```diff
@@ -0,0 +1,232 @@
...
+    };
+    use std::{fmt, time::Duration};
+
+    /// An error occurred while waiting.
+    ///
+    /// This error type is useful for distinguishing between the max wait
+    /// time being exceeded, or some other failure occurring.
+    #[derive(Debug)]
+    #[non_exhaustive]
+    pub enum WaiterError<O, E> {
```
aajtodd, line 25: nice

rust-runtime/aws-smithy-runtime/src/client/waiters.rs
```diff
@@ -0,0 +1,517 @@
...
+{
+    /// Orchestrates waiting via polling with jittered exponential backoff.
+    pub async fn orchestrate(
+        self,
+    ) -> Result<FinalPoll<O, SdkError<E, HttpResponse>>, WaiterError<O, E>> {
+        let start_time = self.time_source.now();
+        let mut attempt = 0;
+        let mut done_retrying = false;
+        loop {
+            tracing::debug!("executing waiter poll attempt #{}", attempt + 1);
```
aajtodd, line 96: **question**: What do the logs look like for a single waiter execution with multiple attempts? Each attempt will be a separate operation which (after we add a unique ID for each invoke) will result in each attempt being uncorrelated to the original waiter logs. Can we do better?

rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs
```diff
@@ -891,12 +891,41 @@ impl RuntimeComponentsBuilder {
...
+        }
+    }
+
     /// Wraps `v` in tracking associated with this builder
     fn tracked<T>(&self, v: Option<T>) -> Option<Tracked<T>> {
         v.map(|v| Tracked::new(self.builder_name, v))
     }
 }
 
+// TODO(waiters): Decide if this is the right approach for extracting sleep_impl/time_source
```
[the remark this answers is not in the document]
aajtodd: Whats the alternative? What are your concerns with this approach?

aajtodd (commented):
Overall looks great! Few questions and some discussion.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentBuilderGenerator.kt
jdisanti: The fluent builders do already have a hook to add additional impl methods, but we don't really have a good pattern for config (non-additive) customizations in smithy-rs. That's a problem that still needs to be solved. in addition, there's a difference between "all the fluent builders," and "just these additional fluent builders over here." The customization pattern doesn't really cover that well right now either.

I think config is probably the better approach for now until the customization patterns have been shored up.

aajtodd: Ya I wasn't suggesting we do anything about it right now (should have probably marked the comment as "ramblings" or something to make that more clear).

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/error/ServiceErrorGenerator.kt
jdisanti: That's to prevent customers from explicitly matching on it, not to prevent us from using it. If a customer matches `Unhandled` instead of `_`, then their code will break if a new error is modeled later.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/WaitableGenerator.kt
jdisanti: Ah. Good call. Thanks for looking at this. I'll rename them.

rust-runtime/aws-smithy-runtime/src/client/waiters.rs
jdisanti: I had a hunch you'd comment about this. I'll take a look at what we can do to correlate them.

rust-runtime/aws-smithy-runtime-api/src/client/runtime_components.rs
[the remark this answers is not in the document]
aajtodd: I don't think what you have is a one way door so I'm not _too_ concerned.

aws/sdk/integration-tests/ec2/tests/waiters.rs
ysaito1001: My guess is that `time_source` and `sleep_impl` returned by `tick_advance_time_and_sleep` are special, i.e. they are frozen unless `tick` is called.

jdisanti: Yuki is right. The responses in the test data require that the waiter wait around and poll a few times, so time has to get advanced. However, it doesn't need to get advanced a full 300 seconds for the success case, as Aaron points out. I had it advance that far so that in the case of failure, it would still complete (max time exceeded).

jdisanti: Added a comment in 637b4b6835607b6340cf3a15200662392034b12c

rust-runtime/aws-smithy-runtime/src/client/waiters.rs
jdisanti: I attached a correlation ID in e1e103fb672dc916f8f380fd50ec306b90a722ff.

Looks like the following in the logs:

```
2024-04-24T23:12:44.983787Z DEBUG waiter{waiter_id=5652710}: aws_smithy_runtime::client::waiters: executing waiter poll attempt #1
2024-04-24T23:12:44.984031Z DEBUG waiter{waiter_id=5652710}:invoke{service=ec2 operation=DescribeInstanceStatus}:apply_configuration: aws_smithy_runtime::client::orchestrator: timeout settings for this operation: TimeoutConfig { connect_timeout: Set(3.1s), read_timeout: Disabled, operation_timeout: Disabled, operation_attempt_timeout: Disabled }
...
2024-04-24T23:13:00.268964Z DEBUG waiter{waiter_id=5652710}: aws_smithy_runtime::client::waiters: delay calculated for attempt #2; elapsed (15.285171s); waiter will poll again in 24s
```

ysaito1001 (commented):
Still going through the PR, but looks great so far. One clarifying question. `.customize` and `.wait_XXX` on a fluent builder cannot be used together, i.e. the following is not supported?
```
client
    .operation()
    .customize()
    .config_override(...)
    .wait_until_XXX
    ...
```

ysaito1001 (approved):
Excellent work! In codegen diff, I see the doc on a waiter's fluent builder look like
```
impl ImageExistsFluentBuilder {
    /// Creates a new `DescribeImages`.
    pub(crate) fn new(handle: ::std::sync::Arc<crate::client::Handle>) -> Self {
        Self {
            handle,
            inner: ::std::default::Default::default(),
        }
    }
```
It says `/// Creates a new DescribeImages` but not `/// Creates a new ImageExistsFluentBuilder` or `/// Creates a new ImageExists`. Is that expected?

jdisanti:
> Still going through the PR, but looks great so far. One clarifying question. `.customize` and `.wait_XXX` on a fluent builder cannot be used together, i.e. the following is not supported?

I opted to not support this yet without better understanding the use-case for it. It won't be hard to add if a good reason comes up.

jdisanti:
> It says /// Creates a new DescribeImages but not /// Creates a new ImageExistsFluentBuilder or /// Creates a new ImageExists. Is that expected?

No, this is a bug. Fixed in 8a3d012780af033292789716ccb71bb37888dc34

PR #3598, opened 2024-04-23
Retro compatible for p256 0.10.0

rodoufu:
## Motivation and Context

Keeping it retro compatible with solana-sdk 1.16

## Description

Updating `p256` version from `0.11` to `0.10` will still allow version `0.11` but will include `0.10` which is necessary when building something that includes `solana-sdk` version `1.16`.

- solana-sdk https://crates.io/crates/solana-sdk/1.16.27/dependencies
  - curve25519-dalek https://crates.io/crates/curve25519-dalek/4.1.2/dependencies
    - zeroize ^1
-  p256 https://crates.io/crates/p256/0.10.1/dependencies
  - elliptic-curve https://crates.io/crates/elliptic-curve/0.11.12/dependencies
    - zeroize ^1
  
`p256 0.11` has another version of `zeroize` which is incompatible
-  p256 https://crates.io/crates/p256/0.11.0/dependencies
  - elliptic-curve https://crates.io/crates/elliptic-curve/0.12.3/dependencies
    - zeroize ^1.5

### DIFF
--- a/aws/rust-runtime/aws-sigv4/Cargo.toml
+++ b/aws/rust-runtime/aws-sigv4/Cargo.toml
@@ -29,7 +29,7 @@ hmac = "0.12"
 http0 = { version = "0.2", optional = true, package = "http" }
 http = { version = "1", optional = true }
 once_cell = "1.8"
-p256 = { version = "0.11", features = ["ecdsa"], optional = true }
+p256 = { version = "0.10", features = ["ecdsa"], optional = true }
 percent-encoding = { version = "2.1", optional = true }
 ring = { version = "0.17.5", optional = true }
 sha2 = "0.10"
--- a/aws/rust-runtime/aws-sigv4/Cargo.toml
+++ b/aws/rust-runtime/aws-sigv4/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-sigv4"
-version = "1.2.1"
+version = "1.2.2"
 authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "David Barsky <me@davidbarsky.com>"]
 description = "SigV4 signer for HTTP requests and Event Stream messages."
 edition = "2021"


jdisanti:
Hello. Can you explain the problem some more? I don't really understand. If the conflict is between zeroize `^1` and zeroize `^1.5`, it should just work. `^1` is compatible with `^1.5`, so it would just choose `^1.5`.

rodoufu:
> Hello. Can you explain the problem some more? I don't really understand. If the conflict is between zeroize `^1` and zeroize `^1.5`, it should just work. `^1` is compatible with `^1.5`, so it would just choose `^1.5`.

@jdisanti I'm sorry, I had it wrong in the description the first time I sent it.
I have fixed the description.

When I created it initially I got the wrong version of `curve25519-dalek`, the version that `cargo` is resolving is actually `3.2.1` which requires `zeroize >=1, <1.4` but `p256 0.11` requires `zeroize ^1.5`, that is the conflict.

I've also added a simple example of how to reproduce it with only a simple `Cargo.toml` file.

rodoufu:
> Instead of downgrading p256 used by `aws-sigv4` to 0.10.0, is it possible to ask the maintainers for `solana-sdk` to upgrade their dependency on `curve25519-dalek` (the latest of which, 4.1.2, does not have semver upper bound `< 1.4` on zeroize)? You can then use the upgraded `solana-sdk`.

@ysaito1001 updating `curve25519-dalek` to 4.1.2 breaks a lot of stuff in `solana-sdk`. I'm guessing that is why they are still using `3.2.1` in their latest version https://crates.io/crates/solana-sdk/1.18.11/dependencies

rodoufu:
@ysaito1001 to update solana one would also need to update its version of `rand` to `0.8` on `jsonrpc` from Parity for which I have an open PR as well https://github.com/paritytech/jsonrpc/pull/688

tarcieri:
Note: there's a PR to fix Solana here: https://github.com/anza-xyz/agave/pull/2252

I'd also note that the latest release of `p256` is v0.13, so `aws-sigv4` is a little behind, if anything.

aajtodd:
Closing as OBE at this point. Open an issue if still relevant.

PR #3601, opened 2024-04-25
Add MSRV to generated SDK Cargo.toml files

jdisanti:
This PR sets the `rust-version` property on generated Cargo.toml files for the AWS SDK crates. It doesn't attempt to place the property on the runtime crates for now since that will either require manual updating, or more machinery to automate.

## Checklist

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/CargoTomlGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/CargoTomlGenerator.kt
@@ -49,6 +49,7 @@ class CargoTomlGenerator(
     private val moduleDescription: String?,
     private val moduleLicense: String?,
     private val moduleRepository: String?,
+    private val minimumSupportedRustVersion: String?,
     private val writer: RustWriter,
     private val manifestCustomizations: ManifestCustomizations = emptyMap(),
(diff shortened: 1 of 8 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/CargoTomlGenerator.kt
```diff
@@ -90,6 +92,7 @@ class CargoTomlGenerator(
                         "edition" to "2021",
                         "license" to moduleLicense,
                         "repository" to moduleRepository,
+                        "rust-version" to minimumSupportedRustVersion,
```
aajtodd, line 95: **correctness**: I could be wrong but looking at this `minimumSupportedRustVersion` can be `null` here. Even though this is in a `listOfNotNull(...)` the type being added is a `Pair<String, String?>` which will never be `null`.

jdisanti: Ah, good point. It surprisingly worked despite this, but I'll correct it.

jdisanti: Fixed in f4dc57c92c3a0cf8b7cc1a0edfacc554f07ae631

aajtodd (approved):

PR #3608, opened 2024-04-26
Bump rustls and hyper-rustls versions. Replace aws-* with actuals

drganjoo:
There is a CVE: https://rustsec.org/advisories/RUSTSEC-2024-0336 related to `rustls`.  We want to back port the fix in `aws-smithy-client = 0.553`.


drganjoo:
No longer needed.

PR #3617, opened 2024-04-30
behavior version 2024-03-28

aajtodd:
## Motivation and Context
This PR is the combination of two previously reviewed PRs that both enable new behaviors behind a new Behavior Major Version (BMV):
* https://github.com/smithy-lang/smithy-rs/pull/3527
* https://github.com/smithy-lang/smithy-rs/pull/3578

## Description
* Enables stalled stream protection by default on latest behavior version.
* Enables creation of a default identity cache.

## Testing
All testing done on prior PRs. See for details.

## Checklist

- [x] I have updated `CHANGELOG.next.toml` if I made changes to the smithy-rs codegen or runtime crates
- [x] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3621, opened 2024-05-01
Fix panic when failing to create `Duration` for exponential backoff from a large float

ysaito1001:
## Motivation and Context
Avoids panic when [Duration for exponential backoff](https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs#L150) could not be created from a large float.

## Description
[Duration::from_secs_f64](https://doc.rust-lang.org/core/time/struct.Duration.html#method.from_secs_f64) may panic. This PR switches to use a fallible sibling `Duration::try_from_secs_f64` to avoid panics. If `Duration::try_from_secs_f64` returns an `Err` we fallback to `max_backoff` for subsequent retries. Furthermore, we learned from internal discussion that jitter also needs to be applied to `max_backoff`. This PR updates `calculate_exponential_backoff` to handle all the said business logic in one place. 

## Testing
- Added a unit test `should_not_panic_when_exponential_backoff_duration_could_not_be_created`
- Manually verified reproduction steps provided [in the original PR](https://github.com/awslabs/aws-sdk-rust/issues/1133)
<details> 
<summary>More details</summary>

```
#[tokio::test]
async fn repro_1133() {
    let config = aws_config::from_env()
        .region(Region::new("non-existing-region")) // forces retries
        .retry_config(
            RetryConfig::standard()
                .with_initial_backoff(Duration::from_millis(1))
                .with_max_attempts(100),
        )
        .timeout_config(
            TimeoutConfig::builder()
                .operation_attempt_timeout(Duration::from_secs(180))
                .operation_timeout(Duration::from_secs(600))
                .build(),
        )
        .load()
        .await;

    let client: Client = Client::new(&config);
    let res = client
        .list_objects_v2()
        .bucket("bucket-name-does-not-matter")
        .send()
        .await;

    dbg!(res);
}
```

Without changes in this PR:
```
---- repro_1133 stdout ----
thread 'repro_1133' panicked at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/time.rs:741:23:
can not convert float seconds to Duration: value is either too big or NaN
stack backtrace:
...

failures:
    repro_1133

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 9 filtered out; finished in 338.18s
```

With changes in this PR:
```
// no panic
---- repro_1133 stdout ----
res = Err(
    TimeoutError(
        TimeoutError {
            source: MaybeTimeoutError {
                kind: Operation,
                duration: 600s,
            },
        },
    ),
)
```
</details> 

## Checklist

## Appendix
<details> 
<summary>runtime-versioner bug fix</summary>

This PR also fixes a limitation in `runtime-versioner audit`. I included the fix in the PR because the issue occurred with special conditions, and we don't get to reproduce it every day. The way the issue manifests is this.

1. We have a branch from the main whose latest release tag at the time was `release-2024-04-30`
2. The main has moved ahead and a new `smithy-rs` release has been made `release-2024-05-08`
3. We perform `git merge main`, pre-commit hooks run, and we then get `audit` failures from `runtime-versioner`:
```
2024-05-10T16:54:36.434968Z  WARN runtime_versioner::command::audit: there are newer releases since 'release-2024-04-30'
aws-config was changed and version bumped, but the new version number (1.4.0) has already been published to crates.io. Choose a new version number.
aws-runtime was changed and version bumped, but the new version number (1.2.2) has already been published to crates.io. Choose a new version number.
aws-smithy-runtime-api was changed and version bumped, but the new version number (1.6.0) has already been published to crates.io. Choose a new version number.
aws-smithy-types was changed and version bumped, but the new version number (1.1.9) has already been published to crates.io. Choose a new version number.
aws-types was changed and version bumped, but the new version number (1.2.1) has already been published to crates.io. Choose a new version number.
Error: there are audit failures in the runtime crates
```
This happens because when the latest `main` is being merged to our branch, `runtime-versioner audit` should use `previous_release_tag` `release-2024-05-08` to perform audit but [pre-commit hooks run the tool](https://github.com/smithy-lang/smithy-rs/blob/main/.pre-commit-hooks/runtime-versioner.sh#L8) using the latest previous release tag that can be traced back from `HEAD` of our branch, which is `release-2024-04-30`. Hence the error.

The fix adds an environment variable `SMITHY_RS_RUNTIME_VERSIONER_AUDIT_PREVIOUS_RELEASE_TAG` to specify a previous release tag override, in addition to a `--previous-release-tag` command-line argument. 
Furthermore, the fix has relaxed certain checks in `audit`. Taking our example for instance, when `HEAD` is now behind `release-2024-05-08`, it's OK to fail even if `release-2024-05-08` is not the ancestor of `HEAD` (as stated above, `git merge-base --is-ancestor` does not know that while main is being merged) as long as `release-2024-04-28` (the latest release seen from `HEAD`) is the ancestor of `release-2024-05-08`.

```
   release-2024-04-28               release-2024-05-08           
            │                                │                   
────────────┼───────────────┬────────────────┼───────x─────► main
            │               │HEAD            │       x           
                            │ of                     x           
                            │branch                  x           
                            │                        x           
                            │                        x           
                            │              xxxxxxxxxxx           
                            │              x                     
                            │              x  git merge main     
                            │              x                     
                            │              ▼                     
                            │                                    
                            └──────────────────► feature branch  
```
To use the fix, set the environment variable with the new release tag and perform `git merge main`:
```
➜  smithy-rs git:(ysaito/fix-panic-in-exp-backoff) ✗ export SMITHY_RS_RUNTIME_VERSIONER_AUDIT_PREVIOUS_RELEASE_TAG=release-2024-05-08
➜  smithy-rs git:(ysaito/fix-panic-in-exp-backoff) ✗ git merge main
     ...
     Running `/Users/awsaito/src/smithy-rs/tools/target/debug/runtime-versioner audit`
2024-05-10T19:32:26.665578Z  WARN runtime_versioner::tag: expected previous release to be 'release-2024-04-30', but 'release-2024-05-08' was specified. Proceeding with 'release-2024-05-08'.
SUCCESS
```

</summary>

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime/Cargo.toml
+++ b/rust-runtime/aws-smithy-runtime/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-runtime"
-version = "1.4.0"
+version = "1.4.1"
 authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "Zelda Hessler <zhessler@amazon.com>"]
 description = "The new smithy runtime crate"
 edition = "2021"
--- a/rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
+++ b/rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
@@ -138,7 +138,7 @@ impl StandardRetryStrategy {
                     } else {
                         fastrand::f64()
                     };
-                    let backoff = calculate_exponential_backoff(
+                    Ok(calculate_exponential_backoff(
                         // Generate a random base multiplier to create jitter
                         base,
                         // Get the backoff time multiplier in seconds (with fractional seconds)
@@ -146,16 +146,9 @@ impl StandardRetryStrategy {
                         // `self.local.attempts` tracks number of requests made including the initial request
                         // The initial attempt shouldn't count towards backoff calculations, so we subtract it
                         request_attempts - 1,
-                    );
-                    match Duration::try_from_secs_f64(backoff) {
-                        Ok(duration) => Ok(duration.min(retry_cfg.max_backoff())),
-                        Err(e) => {
-                            tracing::warn!(
-                                "could not create `Duration` for exponential backoff: {e}"
-                            );
-                            Err(ShouldAttempt::No)
-                        }
-                    }
+                        // Maximum backoff duration as a fallback to prevent overflow when calculating a power
+                        retry_cfg.max_backoff(),
+                    ))
                 }
             }
             RetryAction::RetryForbidden | RetryAction::NoActionIndicated => {
@@ -296,11 +289,29 @@ fn check_rate_limiter_for_delay(
     None
 }
 
-fn calculate_exponential_backoff(base: f64, initial_backoff: f64, retry_attempts: u32) -> f64 {
-    2_u32
+fn calculate_exponential_backoff(
+    base: f64,
+    initial_backoff: f64,
+    retry_attempts: u32,
+    max_backoff: Duration,
+) -> Duration {
+    let result = match 2_u32
         .checked_pow(retry_attempts)
-        .map(|backoff| (backoff as f64) * base * initial_backoff)
-        .unwrap_or(f64::MAX)
+        .map(|power| (power as f64) * initial_backoff)
+    {
+        Some(backoff) => match Duration::try_from_secs_f64(backoff) {
+            Ok(result) => result.min(max_backoff),
+            Err(e) => {
+                tracing::warn!("falling back to {max_backoff:?} as `Duration` could not be created for exponential backoff: {e}");
+                max_backoff
+            }
+        },
+        None => max_backoff,
+    };
+
+    // Apply jitter to `result`, and note that it can be applied to `max_backoff`.
+    // Won't panic because `base` is either in range 0..1 or a constant 1 in testing (if configured).
+    result.mul_f64(base)
 }
 
 fn get_seconds_since_unix_epoch(runtime_components: &RuntimeComponents) -> f64 {
@@ -439,13 +450,15 @@ mod tests {
             ErrorKind::TransientError,
             // Greater than 32 when subtracted by 1 in `calculate_backoff`, causing overflow in `calculate_exponential_backoff`
             33,
-            RetryConfig::standard().with_max_attempts(100), // Any value greater than 33 will do
+            RetryConfig::standard()
+                .with_use_static_exponential_base(true)
+                .with_max_attempts(100), // Any value greater than 33 will do
         );
         let strategy = StandardRetryStrategy::new();
         let actual = strategy
             .should_attempt_retry(&ctx, &rc, &cfg)
             .expect("method is infallible for this use");
-        assert_eq!(ShouldAttempt::No, actual);
+        assert_eq!(ShouldAttempt::YesAfterDelay(MAX_BACKOFF), actual);
     }
 
     #[derive(Debug)]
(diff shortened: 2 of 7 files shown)


rust-runtime/aws-smithy-runtime/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-runtime"
-version = "1.4.0"
+version = "1.4.1"
```
ysaito1001, line 3: Will update this to `1.5.1` if #3617 is merged first.

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
```diff
@@ -147,7 +147,15 @@ impl StandardRetryStrategy {
...
                         request_attempts - 1,
                     );
-                    Ok(Duration::from_secs_f64(backoff).min(retry_cfg.max_backoff()))
+                    match Duration::try_from_secs_f64(backoff) {
+                        Ok(duration) => Ok(duration.min(retry_cfg.max_backoff())),
+                        Err(e) => {
+                            tracing::warn!(
+                                "could not create `Duration` for exponential backoff: {e}"
+                            );
+                            Err(ShouldAttempt::No)
```
aajtodd, line 156: This seems logical to me but just to be sure we've done our due diligence I'm going to poke at this.

1. Why does it make sense to not attempt instead of just capping the backoff to `retry_cfg.max_backoff()`?
2. I glanced through the SEP I don't see any guidance on this scenario, have we asked other SDKs what they do to ensure there is consistency? It may be worth updating the SEP if this is a grey area.


----

Separate question, if we are setting 100 attempts what does the token bucket look like here? I kind of would have thought we'd be at least getting close to exceeding our quotas.

Velfi: Yeah, I think this would be better:

`Duration::try_from_secs_f64(backoff).unwrap_or(Duration::MAX).min(retry_cfg.max_backoff())`

ysaito1001: After internal discussion, we learned it's still debatable whether we should continue making attempts. So we're not going to deliberately make that decision in this PR (and not making that decision doesn't make anythings worse than it is today). We can still make an improvement in that area once we've reached a consensus.

This PR allows customers to perform retries even when overflow occurs, which is updated in 54ad1f0.

ysaito1001: > Duration::try_from_secs_f64(backoff).unwrap_or(Duration::MAX).min(retry_cfg.max_backoff())

This one-liner is tempting, but there was some clarification in the design that jitter also needs to be applied to `max_backoff`. To make everything contained in one function (worrying about overflow, whether it should fallback to `max_backoff`, and for the ease of testing), I put things the way I did in the above commit. Let me know if you can think of a different way.

tools/ci-build/runtime-versioner/src/command/audit.rs
```diff
@@ -21,6 +21,9 @@ use std::{
     process::Command,
 };
 
+const SMITHY_RS_RUNTIME_VERSIONER_AUDIT_PREVIOUS_RELEASE_TAG: &str =
```
aajtodd, line 24: **question**: Why do we need an env variable when the arguments include [previous_release_tag](https://github.com/smithy-lang/smithy-rs/blob/69efe3a72867b0f269ee840f8042f5f477d7476a/tools/ci-build/runtime-versioner/src/main.rs#L34). 

Also can we not make use of the [env](https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes) attribute so that `clap` sources this for us into the `Audit` command line arguments?

aajtodd (approved):

tools/ci-build/runtime-versioner/src/command/audit.rs
ysaito1001: During development, we usually run `runtime-versioner` as [part of pre-commit hooks](https://github.com/smithy-lang/smithy-rs/blob/main/.pre-commit-hooks/runtime-versioner.sh#L8). Do you know if there is a way to specify `--previous-release-tag <some release tag>` to the above invocation done by pre-commit hooks?

> Also can we not make use of the [env](https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes) attribute

Haven't used this feature before. Let me try that.

PR #3622, opened 2024-05-03
Add (S3) transfer settings

brainstorm:
## Motivation and Context

The motivation attempting to fix https://github.com/awslabs/aws-sdk-rust/issues/968 and being able to implement and test the [big files copy S3 Batch Operations strategy outlined in this AWS blogpost](https://aws.amazon.com/blogs/storage/copying-objects-greater-than-5-gb-with-amazon-s3-batch-operations/) but using a Rust runtime lambda instead of Python/Boto3.

This feature is in turn needed for [our microservices, AWS EventBus based service, OrcaBus](https://github.com/umccr/orcabus/), which is a work in progress @umccr.

## Description

Just learned about `smithy-rs` this afternoon after reading docs, code and watching the excellent @Velfi's [Rust Linz youtube intro talk](https://youtu.be/N0XMjokwTIM)... so this is very much an early draft and attempt to understand where code for this feature would fit (and to possibly inform a RFC if needed).

Learning as I go and trying to implement it [by reading similar PRs](https://github.com/smithy-lang/smithy-rs/pull/3488/files). At present I'm following [boto3's transfer.py](https://github.com/boto/boto3/blob/develop/boto3/s3/transfer.py) names... all in all, feedback and [discussion](https://github.com/awslabs/aws-sdk-rust/discussions/964)/hints/guidance is very much welcome.

## Testing

None yet.

## Checklist

----

/cc @andrewpatto @mmalenic

### DIFF
--- a/aws/rust-runtime/aws-config/src/default_provider.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider.rs
@@ -57,3 +57,6 @@ pub mod ignore_configured_endpoint_urls;
 
 /// Default endpoint URL provider chain
 pub mod endpoint_url;
+
+/// Default transfer settings
+pub mod transfer;
--- a/aws/rust-runtime/aws-config/src/default_provider/transfer.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider/transfer.rs
@@ -1,0 +1,12 @@
+// Setting names borrowed from Boto3's transfer.py
+// https://github.com/boto/boto3/blob/096e45841545f24bbd572a62504cc0dbb62e6b07/boto3/s3/transfer.py#L299-L307
+
+#[derive(Default, Debug)]
+pub(crate) struct Settings {
+    pub(crate) multipart_threshold: usize,
+    pub(crate) max_concurrency: usize,
+    pub(crate) multipart_chunksize: usize,
+    pub(crate) num_download_attempts: usize,
+    pub(crate) max_io_queue: usize,
+    pub(crate) io_chunksize: usize,
+}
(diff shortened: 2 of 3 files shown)


aajtodd:
Thanks for taking the time to submit a PR and contribute!

What you're looking for is what we refer to as an S3 "transfer manager". It is actually a higher level library built on top of the base S3 client. This actively being discussed but we have no ETA to share at this time. Unfortunately we can't accept work related to this right now as there is actually a fair bit of nuance to getting it right as well as some (sadly) internal design specs that all languages need to adhere to when implementing this capability.


----

I did see your comment [here](https://github.com/awslabs/aws-sdk-rust/issues/1136#issuecomment-2092006836) about the need for this. I don't think you're necessarily limited to Rust here for performance though as boto3 has integrated the CRT based S3 client (a high performance [C based S3 client)](https://github.com/awslabs/aws-c-s3/). I can't comment on how to use this but I'm sure the Python folks can help you out if you have trouble. Just an option if you're waiting on Rust support for a transfer utility because while it's something we're actively discussing we don't have firm plans for delivering it at the moment.

brainstorm:
Thanks for the explanation! Yes I was aware of the CRT based Boto3 but since our FileManager component in OrcaBus is written in Rust I was looking for better integration, not just performance criteria.

Looking forward to a more public discussion about the transfer manager and seeing it on the public roadmap eventually!

brainstorm:
@aajtodd 

> I don't think you're necessarily limited to Rust here for performance though as boto3 has integrated the CRT based S3 client (a high performance [C based S3 client)](https://github.com/awslabs/aws-c-s3/).

Well, unfortunately (and surprisingly) there's no CRT-backed S3 boto3 `copy()`, it falls back to 'classic' python, see this snippet:

https://github.com/boto/boto3/blob/fb608de3453155578fd68a3a627e27b39f44647f/boto3/s3/inject.py#L437-L439

brainstorm:
Tracking issue https://github.com/awslabs/aws-sdk-rust/issues/1159, yay!

PR #3627, opened 2024-05-06
Add request compression runtime crate

Velfi:
This adds a new runtime crate. The new crate contains code related to compressing requests. Not included in this PR is everything needed to actually make use of the new crate.

----


rust-runtime/aws-smithy-checksums/README.md
```diff
@@ -1,4 +1,4 @@
-# aws-smithy-checksum-callbacks
```
Velfi, line 1: I noticed the README was wrong so I updated it. LMK if you want me to remove this from this PR.

rust-runtime/aws-smithy-compression/src/gzip.rs
```diff
@@ -0,0 +1,128 @@
...
+impl From<CompressionOptions> for Gzip {
+    fn from(options: CompressionOptions) -> Self {
+        Gzip {
+            compression: flate2::Compression::new(options.level),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct Error {
```
aajtodd, line 68: Seems odd to define an error here when we have an `error` module as well? Is this common style? Is the difference just public errors vs private? It makes me wonder if "compression" errors shouldn't be defined as well and instead of returning `BoxError` we return something more concrete 🤔 , what was the deciding factor for making `compress_bytes(..)` return `BoxError`?

rust-runtime/aws-smithy-compression/src/gzip.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    #[test]
+    fn test_gzip_compression() {
+        let gzip = Gzip::from(&CompressionOptions::default());
+        let mut compressed_output = Vec::new();
+        gzip.compress_bytes(gettysburg_address(), &mut compressed_output)
+            .expect("compression succeeds");
+
+        let uncompressed_expected = {
```
aajtodd, line 114: isn't uncompressed_expected just the input (`gettysburg_address()`)? Why do we need to use a decoder to decompress it?

rust-runtime/aws-smithy-compression/src/http.rs
```diff
@@ -0,0 +1,58 @@
...
+//! Checksum support for HTTP requests and responses.
+
+/// Support for the `http-body-0-4` and `http-0-2` crates.
+#[cfg(feature = "http-body-0-4-x")]
+pub mod http_body_0_4_x {
+    use crate::Compression;
+    use http_0_2::header::{HeaderMap, HeaderName, HeaderValue};
+
+    /// Implementors of this trait can be used to compress HTTP requests.
+    pub trait RequestCompressor: Compression + Send + Sync {
```
aajtodd, line 15: Isn't `Compression` already `Send + Sync`?

rust-runtime/aws-smithy-compression/Cargo.toml
```diff
@@ -0,0 +1,51 @@
+[package]
+name = "aws-smithy-compression"
+version = "0.0.1"
+authors = [
+  "AWS Rust SDK Team <aws-sdk-rust@amazon.com>",
+  "Zelda Hessler <zhessler@amazon.com>",
```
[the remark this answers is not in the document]
Velfi: The defacto process so far has been using the name of the actual person who created a crate as well as the team name.

We can change it if you like. It might hurt my vanity a bit though 😆

rust-runtime/aws-smithy-compression/src/gzip.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+impl Compression for Gzip {
+    fn compress_bytes(&mut self, bytes: &[u8], writer: &mut dyn Write) -> Result<(), BoxError> {
+        Gzip::compress_bytes(self, bytes, writer).map_err(Into::into)
+    }
+}
+
+mod http_body_0_4_x {
```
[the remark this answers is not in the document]
Velfi: I forgot them, thanks for noticing this.

rust-runtime/aws-smithy-compression/src/gzip.rs
Velfi: That's the only way I could think of to avoid cross-OS/Implementation/default zip algorithm. differences. Is GZIP deterministic?

rust-runtime/aws-smithy-compression/Cargo.toml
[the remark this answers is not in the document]
aajtodd: I don't care I was more asking if there was some additional rationale I'm missing. Keep your vanity 😆

rust-runtime/aws-smithy-compression/src/http.rs
Velfi: True, I suppose I don't need to include these.

rust-runtime/aws-smithy-compression/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub fn compression_algorithm(&self) -> &str {
+        &self.compression_algorithm
+    }
+}
+
+impl fmt::Display for UnknownCompressionAlgorithmError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            r#"unknown compression algorithm "{}", please pass a known algorithm name ("gzip")"#,
```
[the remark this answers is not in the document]
Velfi: I like to include calls to action whenever returning an "actionable" error to the user. I do agree that this is a potential spot we could overlook if we add new algorithms in the future. Let me think about it.

rust-runtime/aws-smithy-compression/src/gzip.rs
Velfi: I think this is just an oversight on my part but let me think about it for a day.

rust-runtime/aws-smithy-compression/src/http.rs
```diff
@@ -0,0 +1,58 @@
...
+        /// Return a map of headers that must be included when a request is compressed.
+        fn headers(self: Box<Self>) -> HeaderMap<HeaderValue> {
```
ysaito1001, line 17: Can you help better understand the API signature, i.e. the use of `Box<Self>` and `HeaderMap<HeaderValue>` as opposed to say `fn headers(&self) -> (&HeaderName, &HeaderValue)`?

ysaito1001 (commented):
Looks good so far, leaving minor questions.

rust-runtime/aws-smithy-compression/src/gzip.rs
Velfi: > Is GZIP deterministic?
[Survey says "Not necessarily".](https://unix.stackexchange.com/questions/570477/gzip-doesnt-produce-the-same-compressed-result-on-macos-vs-linux)

rust-runtime/aws-smithy-compression/src/body.rs
```diff
@@ -0,0 +1,219 @@
...
+        impl Body for CompressedBody<SdkBody, Box<dyn RequestCompressor>> {
+            type Data = bytes::Bytes;
+            type Error = aws_smithy_types::body::Error;
+
+            fn poll_frame(
+                mut self: Pin<&mut Self>,
+                cx: &mut Context<'_>,
+            ) -> Poll<Option<Result<Frame<Self::Data>, Self::Error>>> {
+                let this = self.as_mut().project();
+                Poll::Ready(match ready!(this.body.poll_frame(cx)) {
```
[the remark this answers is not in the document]
Velfi: The [`ready!` macro](https://doc.rust-lang.org/std/task/macro.ready.html) returns early if it isn't.

rust-runtime/aws-smithy-compression/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+
+impl Storable for CompressionOptions {
+    type Storer = StoreReplace<Self>;
+}
+
+/// An enum encompassing all supported compression algorithms.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum CompressionAlgorithm {
```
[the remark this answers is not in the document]
Velfi: This is an oversight on my part, I'll mark it as non exhaustive

rust-runtime/aws-smithy-compression/src/http.rs
Velfi: I refactored this code and related stuff to work more simply.

Headers now get appended like this:
```rust
fn wrap_request_body_in_compressed_body(
    request: &mut HttpRequest,
    request_compress_impl: Box<dyn CompressRequest>,
) -> Result<(), BuildError> {
    request.headers_mut().append(
        request_compress_impl.header_name(),
        request_compress_impl.header_value(),
    );
    let mut body = {
        let body = mem::replace(request.body_mut(), SdkBody::taken());
        body.map(move |body| {
            let body = CompressedBody::new(
                body,
                request_compress_impl.clone(),
            );

            SdkBody::from_body_0_4(body)
        })
    };
    mem::swap(request.body_mut(), &mut body);

    Ok(())
}
```

rust-runtime/aws-smithy-compression/src/http.rs
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Checksum support for HTTP requests and responses.
+
+/// Support for the `http-body-0-4` and `http-0-2` crates.
+#[cfg(feature = "http-body-0-4-x")]
+pub mod http_body_0_4_x {
```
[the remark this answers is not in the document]
Velfi: IMO yes, at least until hyper 1.0 support is the default.

rust-runtime/aws-smithy-compression/src/gzip.rs
Velfi: I decided to remove this b/c it wasn't necessary.

As for `why don't we return something more concrete?` I don't think that it would help users much. As I see it, the main benefit of returning an actual type is that users can then match on it. For this sort of user error, I don't expect them to address it programmatically. Rather, they should enter a supported algorithm and then re-compile.

rust-runtime/aws-smithy-compression/src/gzip.rs
aajtodd: The linked SO post says only the compressed representation may differ

> the standard just guarantees that every of these possible outputs will be decompressed to the same original data.

aajtodd (approved):
One suggestion but otherwise looks good.

rust-runtime/aws-smithy-compression/src/http.rs
```diff
@@ -0,0 +1,90 @@
...
+    use crate::Compress;
+    use http_0_2::header::{HeaderName, HeaderValue};
+
+    /// Implementors of this trait can be used to compress HTTP requests.
+    pub trait CompressRequest: Compress + CloneCompressRequest {
+        /// Clone this request compressor.
+        fn box_clone(&self) -> Box<dyn CompressRequest>;
```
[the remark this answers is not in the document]
Velfi: oops, that's a leftover. Thanks for catching that.

ysaito1001 (commented):
Ship it!

PR #3632, opened 2024-05-08
update AWS runtime crates for request compression

Velfi:
This PR includes all the changes necessary for AWS runtime crates to support request compression.

----

### DIFF
--- a/aws/rust-runtime/aws-config/src/default_provider/disable_request_compression.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider/disable_request_compression.rs
@@ -23,7 +23,7 @@ mod profile_key {
 /// 2. The profile key `disable_request_compression=true/false`
 ///
 /// If invalid values are found, the provider will return None and an error will be logged.
-pub async fn disable_request_compression_provider(
+pub(crate) async fn disable_request_compression_provider(
     provider_config: &ProviderConfig,
 ) -> Option<bool> {
     let env = provider_config.env();
@@ -86,6 +86,33 @@ mod test {
             .with_fs(Fs::from_slice(&[(
                 "conf",
                 "[default]\ndisable_request_compression = false",
+            )]));
+        assert_eq!(
+            disable_request_compression_provider(&conf).await,
+            Some(true)
+        );
+    }
+
+    #[tokio::test]
+    #[traced_test]
+    async fn profile_config_works() {
+        let conf = ProviderConfig::empty()
+            .with_profile_config(
+                Some(
+                    #[allow(deprecated)]
+                    ProfileFiles::builder()
+                        .with_file(
+                            #[allow(deprecated)]
+                            ProfileFileKind::Config,
+                            "conf",
+                        )
+                        .build(),
+                ),
+                None,
+            )
+            .with_fs(Fs::from_slice(&[(
+                "conf",
+                "[default]\ndisable_request_compression = true",
             )]));
         assert_eq!(
             disable_request_compression_provider(&conf).await,
--- a/aws/rust-runtime/aws-config/src/default_provider/disable_request_compression.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider/disable_request_compression.rs
@@ -1,0 +1,95 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use crate::environment::parse_bool;
+use crate::provider_config::ProviderConfig;
+use aws_runtime::env_config::EnvConfigValue;
+use aws_smithy_types::error::display::DisplayErrorContext;
+
+mod env {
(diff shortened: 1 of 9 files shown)


aws/rust-runtime/aws-config/src/default_provider/disable_request_compression.rs
```diff
@@ -0,0 +1,95 @@
...
+        assert_eq!(disable_request_compression_provider(&conf).await, None);
+        assert!(logs_contain(
+            "invalid value for `disable request compression` setting"
+        ));
+        assert!(logs_contain("AWS_DISABLE_REQUEST_COMPRESSION"));
+    }
+
+    #[tokio::test]
+    #[traced_test]
+    async fn environment_priority() {
```
ysaito1001, line 67: I believe `EnvConfigValue` itself has already been tested that way, but maybe we should also add a test making sure a value is correctly loaded from a profile config when disabling is only specified in a shared config file, but not set in the environment variable?

aws/rust-runtime/aws-config/src/default_provider/request_min_compression_size_bytes.rs
```diff
@@ -0,0 +1,100 @@
...
+        );
+        assert!(logs_contain(
+            "invalid value for `request minimum compression size bytes` setting"
+        ));
+        assert!(logs_contain("AWS_REQUEST_MIN_COMPRESSION_SIZE_BYTES"));
+    }
+
+    #[tokio::test]
+    #[traced_test]
+    async fn environment_priority() {
```
ysaito1001, line 72: The above comment regarding testing against a shared config file also applies here?

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
```diff
@@ -473,7 +473,7 @@ mod tests {
     use crate::sensitive_command::CommandWithSensitiveArgs;
     use serde::Deserialize;
 
-    #[cfg(feature = "test-utils")]
+    #[cfg(feature = "test-util")]
```
ysaito1001, line 476: Is this just a name change without any programmatic effects?

ysaito1001 (approved):

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
landonxjames: Beyond the name change, the `aws-config` crate doesn't seem to have a `test-util(s)` feature listed in `Cargo.toml` at all so I'm not sure how this is working. Does it transitively work with the `test-util` feature present in some of the other crates?

landonxjames (approved):
Looks good to me! Just a few clarifying questions as I get up to speed

aws/rust-runtime/aws-config/src/default_provider/disable_request_compression.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/// Load the value for "disable request compression".
+///
+/// This checks the following sources:
+/// 1. The environment variable `AWS_DISABLE_REQUEST_COMPRESSION=true/false`
+/// 2. The profile key `disable_request_compression=true/false`
+///
+/// If invalid values are found, the provider will return None and an error will be logged.
+pub async fn disable_request_compression_provider(
```
aajtodd, line 26: Seems odd to me these end up being public to the world but that seems to be the pattern here. I would have thought `pub crate`.

aajtodd (approved):

aws/rust-runtime/aws-config/src/environment/mod.rs
```diff
@@ -41,6 +41,25 @@ pub(crate) fn parse_bool(value: &str) -> Result<bool, InvalidBooleanValue> {
     }
 }
 
+#[derive(Debug)]
+pub(crate) struct InvalidUintValue {
```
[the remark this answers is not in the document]
Velfi: Correct. If it's not public then we don't have to worry about forwards compatibility.

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
Velfi: We never added these features to the cargo.toml so I'm pretty sure they were never usable. I changed the name to bring it in line with the other crates. I'm also going to add the feature to the `Cargo.toml`.

aws/rust-runtime/aws-config/src/default_provider/disable_request_compression.rs
Velfi: I was doing it to match the other providers but I don't mind making them `pub(crate)`. We can always undo that later if we decide to.

aws/rust-runtime/aws-config/src/profile/credentials/repr.rs
Velfi: I'm going to transfer this change to a separate PR because it's causing some test failures.

PR #3633, opened 2024-05-08
fix(docs: client construction): wrong type name

problame:
## Motivation and Context
Docs are wrong.

## Description
The docs are referring to the wrong type name (`Config`).
The example right below the change uses the right type name (`$crateName::config::Builder`).

## Testing
n/a

## Checklist

----


aajtodd:
Thanks for the PR!

The service client `Config` structs [do implement](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/struct.Config.html#impl-From%3C%26SdkConfig%3E-for-Config) `From<&SdkConfig>` (as does the [builder](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/config/struct.Builder.html#impl-From%3C%26SdkConfig%3E-for-Builder)). I believe this is correct and as intended.

aajtodd:
Ahh no I agree that makes more sense. Thanks for clarification.

Velfi (approved):

PR #3635, opened 2024-05-09
Allow AggregatedBytes to be constructed directly from segmented bytes

thedodd:
## Motivation and Context

This is useful for cases where I already have a segmented buffer of bytes, and I need to stream them to S3 as a single object.

## Description

Internal to this crate, the AggregatedBytes struct can be constructed directly from a SegmentedBytes buffer. 

There are cases where folks would like to perform this same thing. 

## Testing

I am currently using this for a project where a segmented buffer of bytes is streamed to S3 as a single object.

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-types/src/byte_stream.rs
+++ b/rust-runtime/aws-smithy-types/src/byte_stream.rs
@@ -492,6 +492,11 @@ impl From<Vec<u8>> for ByteStream {
 pub struct AggregatedBytes(SegmentedBuf<Bytes>);
 
 impl AggregatedBytes {
+    /// Construct a new instance from a segmented buffer of `Bytes`.
+    pub fn from_segmented(segments: SegmentedBuf<Bytes>) -> Self {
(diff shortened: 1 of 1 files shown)


ysaito1001:
Thank you for contributing! I suspect our CI may not be happy with the change because it now exposes in a stable crate (`aws-smithy-types`) a type [SegmentedBuf](https://docs.rs/bytes-utils/latest/bytes_utils/struct.SegmentedBuf.html) from an unstable crate (`bytes-utils` which is pre 1.0). The CI only allows for items from a stable crate to be exposed in a stable crate.

> Internal to this crate, the AggregatedBytes struct can be constructed directly from a SegmentedBytes buffer.

Since you mentioned "Internal to this crate", maybe you mean `pub(crate) fn from_segmented` ?

ysaito1001:
Thank you once again for your contribution. We can reopen this discussion once `bytes-utils` reaches version 1.x.

PR #3638, opened 2024-05-10
Request Compression PR 3/3 - Codegen

Velfi:
PR 3 of 3. Most of the compression tests happen here since we don't have a request-compression-supporting model in our set of "smoke test" service models.

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
@@ -109,10 +109,12 @@ fun Model.sdkConfigSetter(
     val builtIn = loadBuiltIn(serviceId, builtInSrc) ?: return null
     val fieldName = configParameterNameOverride ?: builtIn.name.rustName()
 
+    val builtinType = builtIn.type!!
     val map =
-        when (builtIn.type!!) {
+        when (builtinType) {
             ParameterType.STRING -> writable { rust("|s|s.to_string()") }
             ParameterType.BOOLEAN -> null
+            else -> PANIC("needs to handle unimplemented endpoint parameter builtin type: $builtinType")
         }
 
     return if (fieldName == "endpoint_url") {
--- a/aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpRequestCompressionDecoratorTest.kt
+++ b/aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpRequestCompressionDecoratorTest.kt
@@ -273,6 +273,8 @@ class HttpRequestCompressionDecoratorTest {
                             .with_test_defaults()
                             .http_client(http_client)
                             .disable_request_compression(false)
+                            // Since our streaming body is sized, we have to set this.
+                            .request_min_compression_size_bytes(128)
                             .build();
 
                             let client = $moduleName::Client::from_conf(config);
@@ -284,6 +286,7 @@ class HttpRequestCompressionDecoratorTest {
                             let body = ByteStream::read_from()
                                 .path(file.path())
                                 .buffer_size(1024)
+                                .length(#{Length}::Exact(UNCOMPRESSED_INPUT.len() as u64))
                                 .build()
                                 .await
                                 .unwrap();
@@ -311,6 +314,7 @@ class HttpRequestCompressionDecoratorTest {
                     "capture_request" to RuntimeType.captureRequest(rc),
                     "pretty_assertions" to CargoDependency.PrettyAssertions.toType(),
                     "tempfile" to CargoDependency.TempFile.toType(),
+                    "Length" to RuntimeType.smithyTypes(rc).resolve("byte_stream::Length"),
                 )
             }
         }
--- a/aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpRequestCompressionDecoratorTest.kt
+++ b/aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpRequestCompressionDecoratorTest.kt
@@ -1,0 +1,318 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.rustsdk
+
+import org.junit.jupiter.api.Test
+import software.amazon.smithy.rust.codegen.core.rustlang.CargoDependency
+import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType.Companion.preludeScope
+import software.amazon.smithy.rust.codegen.core.testutil.asSmithyModel
+import software.amazon.smithy.rust.codegen.core.testutil.integrationTest
+
+class HttpRequestCompressionDecoratorTest {
+    companion object {
+        // Can't use the dollar sign in a multiline string with doing it like this.
+        private const val PREFIX = "\$version: \"2\""
+        val model =
+            """
+            $PREFIX
+            namespace test
+
+            use aws.api#service
+            use aws.auth#sigv4
+            use aws.protocols#restJson1
+            use smithy.rules#endpointRuleSet
+
+            @service(sdkId: "dontcare")
+            @restJson1
+            @sigv4(name: "dontcare")
+            @auth([sigv4])
+            @endpointRuleSet({
+                "version": "1.0",
+                "rules": [{ "type": "endpoint", "conditions": [], "endpoint": { "url": "https://example.com" } }],
+                "parameters": {
+                    "Region": { "required": false, "type": "String", "builtIn": "AWS::Region" },
+                }
+            })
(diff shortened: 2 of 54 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientRustSettings.kt
```diff
@@ -86,8 +86,8 @@ data class ClientRustSettings(
  * [addMessageToErrors]: Adds a `message` field automatically to all error shapes
  */
 data class ClientCodegenConfig(
-    override val formatTimeoutSeconds: Int = defaultFormatTimeoutSeconds,
-    override val debugMode: Boolean = defaultDebugMode,
+    override val formatTimeoutSeconds: Int = DEFAULT_FORMAT_TIMEOUT_SECONDS,
+    override val debugMode: Boolean = DEFAULT_DEBUG_MODE,
```
Velfi, line 90: Some of the random fixes (like this one) are b/c IntelliJ complains a lot and I'm easily swayed.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
```diff
@@ -58,12 +58,12 @@ sealed class ServiceConfig(name: String) : Section(name) {
     /**
      * Additional documentation comments for the `Config` struct.
      */
-    object ConfigStructAdditionalDocs : ServiceConfig("ConfigStructAdditionalDocs")
+    data object ConfigStructAdditionalDocs : ServiceConfig("ConfigStructAdditionalDocs")
```
Velfi, line 61: This whole file is IntelliJ lints

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/Rust.kt
```diff
@@ -97,6 +97,7 @@ object TestWorkspace {
                 mapOf(
                     "workspace" to
                         mapOf(
+                            "resolver" to "2",
```
Velfi, line 100: We didn't use to specify this so we'd get the warning for "please use resolver = '2'" when generating tests. This fixes that.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/EndpointBuiltInsDecorator.kt
Velfi, line 1: I'm not sure why this is showing up in the diff. AFAICT, it's the same in `main` and `zhessler/request-compression`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/CargoDependency.kt
```diff
@@ -244,12 +257,11 @@ data class CargoDependency(
     companion object {
         // Forces AHash to be a later version that avoids
         // https://github.com/tkaitchuck/aHash/issues/200
-        val AHash: CargoDependency = CargoDependency("ahash", CratesIo("0.8.11"), defaultFeatures = false)
-        val OnceCell: CargoDependency = CargoDependency("once_cell", CratesIo("1.16"))
-        val Url: CargoDependency = CargoDependency("url", CratesIo("2.3.1"))
-        val Bytes: CargoDependency = CargoDependency("bytes", CratesIo("1.0.0"))
+        val AHash: CargoDependency = CargoDependency("ahash", CratesIo("0.8.11"))
```
aajtodd, line 260: **question**: Was `defaultFeatures = false` being dropped from `AHash` intentional?

rust-runtime/inlineable/src/client_request_compression.rs
```diff
@@ -0,0 +1,256 @@
...
+        if !options.is_enabled() {
+            tracing::trace!("request compression is disabled and will not be applied");
+            return Ok(());
+        }
+
+        // Don't wrap a body if it's below the minimum size
+        //
+        // Because compressing small amounts of data can actually increase its size,
+        // we check to see if the data is big enough to make compression worthwhile.
+        if let Some(known_size) = request.body().bytes().map(|b| b.len()) {
```
aajtodd, line 129: **question**: If I"m reading `ByteStream` and `SdkBody` type implementations correctly `SdkBody.bytes()` may not work correctly for file inputs and would restrict compression to only in-memory representations. Is this intentional as it seems like a possibly undesired limitation?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/CargoDependency.kt
Velfi: This was unintentional. I've re-added this.

rust-runtime/inlineable/src/client_request_compression.rs
Velfi:
```rust
// Don't wrap a body if it's below the minimum size
//
// Because compressing small amounts of data can actually increase its size,
// we check to see if the data is big enough to make compression worthwhile.
if let Some(known_size) = request.body().bytes().map(|b| b.len()) {
    if known_size < options.min_compression_size_bytes() as usize {
        tracing::trace!("request body is below minimum size and will not be compressed");
        return Ok(());
    }
    tracing::trace!("compressing non-streaming request body...")
} else {
    tracing::trace!("compressing streaming request body...");
}
```

The reasoning is:
1. When a body doesn't have a known length and compression is enabled, then we compress that body.
2. When a body does have a known length and compression is enabled, then we compress the body only if it's above the `min_compression_size_bytes`.

In the code, we're checking for an in-memory body, which is an indirect way of checking that it's streaming vs. non-streaming. In this case, we may end up compressing files that would be below the `min_compression_size_bytes`.

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpRequestCompressionDecoratorTest.kt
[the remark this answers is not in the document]
Velfi: Good job noticing this. I checked the internal spec and it says:

> For a request containing a member with the [streaming](https://smithy.io/2.0/spec/streaming.html#smithy-api-streaming-trait) trait without [requiresLength](https://smithy.io/2.0/spec/streaming.html#requireslength-trait), the request MUST be compressed regardless of the minimum compression threshold setting (if `DISABLE_REQUEST_COMPRESSION` is not true).

With our current implementation, we will always compress a streaming request, regardless of whether the length is known. Let me switch to using `SizeHint` instead of checking for an in-memory request body.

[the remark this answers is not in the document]
Velfi: Smithy doesn't allow `streaming` and `requiresLength` to be applied to the same thing. 

```smithy
@streaming
@requiresLength
blob KnownLengthStreamingBlob
```

_(Assuming request compression is enabled and the trait is applied to the operation in question)_
I updated the implementation so that we treat a sized streaming body like an in-memory body: both will be checked to ensure their length is greater than the minimum allowed. Un-sized streaming bodies will always be request.

rust-runtime/inlineable/src/client_request_compression.rs
Velfi: I changed this to use `SizeHint` so that sized streaming bodies will be treated the same as in-memory bodies. Streaming bodies without a length will always be compressed _(so long as request compression is enabled and appropriate for the operation.)_

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpRequestCompressionDecoratorTest.kt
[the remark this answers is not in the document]
ysaito1001: Ah got it, thanks for a deep dive! As long as double annotation (`streaming` and `requiresLength`) is prevented by the modeling tool `Smithy`, then I think we're OK to not protect ourselves from that in our runtime crate.

ysaito1001 (commented):
~~The changes themselves look good to me. We just need to remove unnecessary commits not part of the PR. I hope that can be resolved as simply as merging the latest main into the PR.~~

Looks like commit 1c30ef4 brought main into this PR, which is why it shows some diffs from main that the target branch `zhessler/request-compression` doesn't have. We can merge the latest main into `zhessler/request-compression` and then merge `zhessler/request-compression` into this PR to clean up those diffs, but I suppose it's ok to have extra diffs in this PR (as long as they are cleared when `zhessler/request-compression` is diffed against main).

rust-runtime/inlineable/src/client_request_compression.rs
```diff
@@ -0,0 +1,256 @@
...
+use std::borrow::Cow;
+use std::{fmt, mem};
+
+#[derive(Debug)]
+pub(crate) struct RequestCompressionRuntimePlugin {
+    runtime_components: RuntimeComponentsBuilder,
+}
+
+impl RequestCompressionRuntimePlugin {
+    pub(crate) fn new() -> Self {
```
ysaito1001, line 31: I see `RequestCompressionRuntimePlugin` can support `new_1_0` when we have another `RequestCompressionInterceptor` that's `http_body_1_x` aware, which is good.

ysaito1001 (approved):

PR #3639, opened 2024-05-10
fix const names in codegen

Velfi:
This renames stuff in codegen based on IntelliJ lints.

----


landonxjames (approved):
I wonder if we could use an automated linter/formatter that would keep these things from piling up? A quick search didn't turn up anything as authoritative as `rustfmt`, but there do seem to be some options out there. Beyond the scope of this PR, but might be a nice to have commit hook or something.

drganjoo:
This looks nice - thank you!

Should we also change `PropertyKey` to `PROPERTY_KEY` on line #45 of `codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/CargoDependency.kt`?

```
        private const val PropertyKey = "rustdep"
```

drganjoo (approved):

Issue #2891, opened 2023-07-31
Implement support for `@requestCompression`


jdisanti:
Add support for the `@requestCompression` trait added in Smithy 1.31, and then re-enable the protocol tests for that trait.

PR #3647, opened 2024-05-16
Feature: Request Compression for operations modeled with the `@requestCompression` trait

Velfi:
## Motivation and Context

#2891 

## Description

This PR contains all the changes from the last three compression PRs.

New in this PR:

- I re-enabled some GZIP protocol tests that were previously disabled.
- I added changelog entries.

## Checklist

----


rust-runtime/aws-smithy-compression/src/gzip.rs
```diff
@@ -0,0 +1,113 @@
...
+        encoder.write_all(bytes)?;
+        encoder.try_finish()?;
+
+        Ok(())
+    }
+}
+
+impl Compress for Gzip {
+    fn compress_bytes(&mut self, bytes: &[u8], writer: &mut dyn Write) -> Result<(), BoxError> {
+        Gzip::compress_bytes(self, bytes, writer).map_err(Into::into)
```
landonxjames, line 28: Is the `flate2` gzip implementation deterministic by default? Was reading and it looks like the GNU version is not unless you use `-n` since it includes a timestamp in the file. Note sure if that really matters for our use-case.

landonxjames (approved):
Looks good based on all the previous feature branch reviews. 

One question, just curious what was going on with the flaky test?

rust-runtime/aws-smithy-compression/src/gzip.rs
Velfi: The test was broken b/c I mounted the interceptor twice in codegen and I was doing snapshot testing. Once I removed the extra interceptor things worked fine. As far as I can tell, GZIP is deterministic in CI b/c the tests looking for a specific output are passing. If GZIP wasn't deterministic then we'd have to use a decoder to unzip and then check the contents. I think I do that in the compression crate somewhere.

I'm going to ignore it until it actually fails. LMK if you can think of a good test to add.

rust-runtime/inlineable/src/client_request_compression.rs
```diff
@@ -0,0 +1,256 @@
...
+#[derive(Debug, Copy, Clone)]
+pub(crate) struct RequestMinCompressionSizeBytes(pub(crate) u32);
+
+impl Default for RequestMinCompressionSizeBytes {
+    fn default() -> Self {
+        RequestMinCompressionSizeBytes(10240)
+    }
+}
+
+impl From<u32> for RequestMinCompressionSizeBytes {
```
ysaito1001, line 191: A trait implementation exposes a type in `pub` automatically. I see `RequestMinCompressionSizeBytes` stores `u32` as `pub(crate)`, which keeps `u32` private, but the `impl From` sort of undoes it, exposing `u32` to the world. If we can get way with a `pub(crate)` conversion, I think we should prefer that ([reference](https://youtu.be/qfknfCsICUM?si=qHwnvYZoEOh2BBnN&t=2780)). If not, let's stick to what we have here (we can add another `impl From<xxx>` in the future in an additive fashion.

The same goes for the above `impl From<bool> for DisableRequestCompression`.

ysaito1001 (approved):
Great work, thank you!

PR #3657, opened 2024-05-22
Fix: Only enforce content length for GET requests

rcoh:

### DIFF
--- a/aws/sdk/integration-tests/s3/tests/content-length-enforcement.rs
+++ b/aws/sdk/integration-tests/s3/tests/content-length-enforcement.rs
@@ -1,0 +1,99 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use aws_sdk_s3::{config::Region, error::DisplayErrorContext, Client, Config};
+use aws_smithy_runtime::client::http::test_util::dvr::ReplayingClient;
+
+#[tokio::test]
+async fn test_content_length_enforcement_is_not_applied_to_head_request() {
+    let http_client =
+        ReplayingClient::from_file("tests/data/content-length-enforcement/head-object.json")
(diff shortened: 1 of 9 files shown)


ysaito1001:
Thanks Russell. We just need to change the target branch to main for this PR, instead of the release branch. We discovered a couple of weeks ago that directly bumping a runtime crate in the release branch without cutting a release confused `runtime-versioner audit`

`runtime versioner` refers to the latest release in `smithy-rs` when detecting baseline crates versions and compares those in the main branch, but if a runtime crate version gets bumped "secretly" in the release branch without a new release tag and subsequently published to crates.io, then the tool gets confused and complains that the crate version is ahead of the latest release tag but somehow that version is already available on crates.io.

ysaito1001:
Thanks Russell. We just need to change the target branch to main for this PR, instead of the release branch. We discovered a couple of weeks ago that directly bumping a runtime crate in the release branch without cutting a release confused `runtime-versioner audit`

`runtime versioner` refers to the latest release in `smithy-rs` when detecting baseline crates versions and compares those in the current active branch. If a runtime crate version gets bumped "secretly" in the release branch without a new release tag and subsequently published to crates.io, then the tool gets confused and complains that the crate version is ahead of the latest release tag but somehow that version is already available on crates.io.

aws/sdk/integration-tests/s3/tests/content-length-enforcement.rs
```diff
@@ -0,0 +1,99 @@
...
+        .key("1000-lines.txt")
+        .bucket("dontcare")
+        .send()
+        .await
+        .unwrap();
+
+    // This will fail with a content-length mismatch error.
+    let content_length_err = output.body.collect().await.unwrap_err();
+
+    http_client.full_validate("").await.unwrap();
```
rcoh, line 60: the value you pass here is the content type `"application/text"` is probably clearer here

Velfi (approved):

PR #3662, opened 2024-05-23
Upgrade Smithy to 1.49

ysaito1001:
## Testing

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/QuerySerializerGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/QuerySerializerGenerator.kt
@@ -282,9 +282,17 @@ abstract class QuerySerializerGenerator(private val codegenContext: CodegenConte
         }
     }
 
-    private fun RustWriter.serializeCollection(
+    protected open fun RustWriter.serializeCollection(
         memberContext: MemberContext,
         context: Context<CollectionShape>,
+    ) {
+        serializeCollectionInner(memberContext, context, this)
+    }
+
+    protected fun serializeCollectionInner(
+        memberContext: MemberContext,
+        context: Context<CollectionShape>,
+        writer: RustWriter,
     ) {
         val flat = memberContext.shape.isFlattened()
         val memberOverride =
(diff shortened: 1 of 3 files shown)


landonxjames (approved):
LGTM!

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/QuerySerializerGenerator.kt
```diff
@@ -282,30 +282,38 @@ abstract class QuerySerializerGenerator(private val codegenContext: CodegenConte
...
-                serializeMemberValue(
+                writer.serializeMemberValue(
                     MemberContext(entryName, ValueExpression.Reference(itemName), context.shape.member),
                     targetShape,
                 )
             }
-            rust("$listName.finish();")
+            writer.rust("$listName.finish();")
         }
     }
```
Velfi, line 319: This looks to be doing the same thing. Why update it?

ysaito1001: This is working around Kotlin's limitation that an extension method cannot be overridden. `Ec2QuerySerializerGenerator.kt`, a subclass of `QuerySerializerGenerator.kt, needs to override `RustWriter.serializeCollection` so that it should only call it if an input list is NOT empty (otherwise `&ListArg=` would be rendered which ec2 query protocol doesn't like in Smithy 1.49). However, we currently couldn't do this in Kotlin if it were not for `serializeCollectionInner`:
```
override fun RustWriter.serializeCollection(
        memberContext: MemberContext,
        context: Context<CollectionShape>,
    ) {
        rustBlock("if !${context.valueExpression.asRef()}.is_empty()") {
            super.serializeCollection(memberContext, context) // cannot call super class's extension method like this
        }
```

ysaito1001: This is working around Kotlin's limitation that an extension method cannot be overridden. `Ec2QuerySerializerGenerator`, a subclass of `QuerySerializerGenerator`, needs to override `RustWriter.serializeCollection` so that it should only call it if an input list is NOT empty (otherwise `&ListArg=` would be rendered which ec2 query protocol doesn't like in Smithy 1.49). However, we currently couldn't do this in Kotlin if it were not for `serializeCollectionInner`:
```
override fun RustWriter.serializeCollection(
        memberContext: MemberContext,
        context: Context<CollectionShape>,
    ) {
        rustBlock("if !${context.valueExpression.asRef()}.is_empty()") {
            super.serializeCollection(memberContext, context) // cannot call super class's extension method like this
        }
```

PR #3665, opened 2024-05-28
Pin crc32c to avoid build.rs failure

ysaito1001:
## Motivation and Context
Works around https://github.com/zowens/crc32c/issues/67

----


landonxjames (approved):
Looks good. Discussed offline and the semver failure is likely due to the comparison against `aws-sdk-s3` which isn't impacted by this pinning

PR #3667, opened 2024-05-28
expose connector in user agent

Velfi:
## Motivation and Context

part of the `hyper` 1.0 upgrade

## Description

This PR adds information about the connector used to send a request to the metadata included in the user agent.

## Testing

I added an integration test.

## Checklist

----


rcoh (commented):
you might consider a more generalized system to get metadata into the user agent from interceptors 🤔

aws/rust-runtime/aws-runtime/src/user_agent/interceptor.rs
```diff
@@ -99,6 +112,18 @@ impl Intercept for UserAgentInterceptor {
                 if let Some(app_name) = maybe_app_name {
                     ua.set_app_name(app_name.clone());
                 }
+
+                let maybe_connector_metadata = runtime_components
+                    .http_client()
+                    .and_then(|c| c.connector_metadata());
+                if let Some(connector_metadata) = maybe_connector_metadata {
+                    let fm = FrameworkMetadata::new(
```
aajtodd, line 120: **fix**: This wouldn't be framework level metadata which is described as:

> For frameworks owned by the SDKs, e.g. `lib/chalice/1.2.3`

This is just additional `md/` pairs. e.g. `md/http#hyper-1.x`

rust-runtime/aws-smithy-experimental/src/hyper_1_0.rs
```diff
@@ -623,6 +614,10 @@ where
         let _ = (self.tcp_connector_fn)();
         Ok(())
     }
+
+    fn connector_metadata(&self) -> Option<ConnectorMetadata> {
+        Some(ConnectorMetadata::new("hyper-1", None))
```
aajtodd, line 619: **correctness**: Wouldn't this be `ConnectorMetadata::new("hyper", "1") or `"1.x"`

aajtodd (changes requested):

aws/sdk/integration-tests/s3/tests/user-agent.rs
ysaito1001, line 1: If it's not too much work, it's a good opportunity to convert them to use `awsSdkIntegrationTest` and delete this from `integration-tests` (since there aren't many tests in this file), unless these tests verify something s3-specific.

rust-runtime/aws-smithy-runtime-api/src/client/http.rs
```diff
@@ -165,6 +166,15 @@ pub trait HttpClient: Send + Sync + fmt::Debug {
         let _ = (runtime_components, cfg);
         Ok(())
     }
+
+    /// Provide metadata about the crate that this HttpClient uses to make connectors.
+    ///
+    /// If this is implemented and returns metadata, that metadata will be
+    /// inserted into the user agent string when making a request with this
+    /// client.
```
ysaito1001, line 174: To verify my understanding, is it still true when an SDK is code-generated without AWS codegen decorators?

rust-runtime/aws-smithy-experimental/src/hyper_1_0.rs
Velfi: I'll do "1.x"

aws/sdk/integration-tests/s3/tests/user-agent.rs
Velfi: good point, I'll work on that

rust-runtime/aws-smithy-runtime-api/src/client/http.rs
Velfi: It isn't b/c I don't think we have a generic user agent interceptor. I'll re-word this.

aws/rust-runtime/aws-runtime/src/user_agent/interceptor.rs
Velfi: Got it. I'll update that.

aws/rust-runtime/aws-runtime/src/user_agent/interceptor.rs
```diff
@@ -99,6 +112,15 @@ impl Intercept for UserAgentInterceptor {
                 if let Some(app_name) = maybe_app_name {
                     ua.set_app_name(app_name.clone());
                 }
+
+                let maybe_connector_metadata = runtime_components
+                    .http_client()
+                    .and_then(|c| c.connector_metadata());
+                if let Some(connector_metadata) = maybe_connector_metadata {
+                    let am = AdditionalMetadata::new(Cow::Owned(connector_metadata.to_string()))?;
```
aajtodd, line 120: I still don't think this follows the SEP the way we want. The SEP defines:

```
ua-pair              = ua-name ["#" ua-value]
ua-name              = token_no_hash
ua-value             = token
...
additional-metadata  = "md/" ua-pair
```

I would think we want this to show up as `md/http#hyper-1.x` or `md/http#hyper-0.x`, right now I think it shows up as just `md/hyper-1.x`.

aajtodd (approved):

PR #3670, opened 2024-05-29
bootstrap s3 transfer manager

aajtodd:
## Description

This PR bootstraps a new crate `aws-s3-transfer-manager` (feel free to bikeshed the crate name 🤷 ).

* Establishes a new crate and initial project setup
* Adds basic object discovery related functionality required for multi-part download/concurrent ranged gets.
* Adds initial error types (WIP and subject to change/feedback).

----

### DIFF
--- a/aws/hll/aws-s3-transfer-manager/Cargo.toml
+++ b/aws/hll/aws-s3-transfer-manager/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "aws-s3-transfer-manager"
+version = "0.1.0"
+edition = "2021"
+authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "Aaron Todd <todaaron@amazon.com>"]
+description = "S3 Transfer Manager"
+license = "Apache-2.0"
+repository = "https://github.com/smithy-lang/smithy-rs"
+publish = false
+
+
+[dependencies]
+aws-sdk-s3 = "1.22.0"
+aws-smithy-http = "0.60.7"
+tracing = "0.1"
+async-trait = "0.1.80"
+# FIXME - upgrade to hyper 1.x
+hyper = { version = "0.14.28", features = ["client"] }
+aws-smithy-types = "1.1.8"
+bytes = "1"
+thiserror = "1.0.58"
+aws-types = "1.2.0"
+tokio = { version = "1.37.0", features = ["rt-multi-thread", "io-util", "sync", "fs", "macros"] }
+aws-smithy-runtime-api = "1.4.0"
+async-channel = "2.2.1"
+
+[dev-dependencies]
+aws-config = { version = "1.2.0", features = ["behavior-version-latest"] }
+clap = { version = "4.5.4", default-features = false, features = ["derive", "std", "help"] }
+tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
--- a/aws/hll/aws-s3-transfer-manager/Cargo.toml
+++ b/aws/hll/aws-s3-transfer-manager/Cargo.toml
@@ -8,24 +8,23 @@ license = "Apache-2.0"
 repository = "https://github.com/smithy-lang/smithy-rs"
 publish = false
 
-
 [dependencies]
+async-channel = "2.2.1"
+async-trait = "0.1.80"
 aws-sdk-s3 = { version = "1.22.0", features = ["behavior-version-latest", "test-util"] }
 aws-smithy-http = "0.60.7"
-tracing = "0.1"
-async-trait = "0.1.80"
+aws-smithy-runtime-api = "1.4.0"
+aws-smithy-types = "1.1.8"
+aws-types = "1.2.0"
+bytes = "1"
 # FIXME - upgrade to hyper 1.x
 hyper = { version = "0.14.28", features = ["client"] }
-aws-smithy-types = "1.1.8"
-bytes = "1"
 thiserror = "1.0.58"
-aws-types = "1.2.0"
 tokio = { version = "1.37.0", features = ["rt-multi-thread", "io-util", "sync", "fs", "macros"] }
-aws-smithy-runtime-api = "1.4.0"
-async-channel = "2.2.1"
+tracing = "0.1"
 
 [dev-dependencies]
 aws-config = { version = "1.2.0", features = ["behavior-version-latest"] }
+aws-smithy-mocks-experimental = "0.2.1"
 clap = { version = "4.5.4", default-features = false, features = ["derive", "std", "help"] }
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
-aws-smithy-mocks-experimental = "0.2.1"
--- a/aws/hll/aws-s3-transfer-manager/Cargo.toml
+++ b/aws/hll/aws-s3-transfer-manager/Cargo.toml
@@ -10,7 +10,7 @@ publish = false
 
 
 [dependencies]
-aws-sdk-s3 = "1.22.0"
+aws-sdk-s3 = { version = "1.22.0", features = ["behavior-version-latest", "test-util"] }
 aws-smithy-http = "0.60.7"
 tracing = "0.1"
 async-trait = "0.1.80"
@@ -28,3 +28,4 @@ async-channel = "2.2.1"
 aws-config = { version = "1.2.0", features = ["behavior-version-latest"] }
 clap = { version = "4.5.4", default-features = false, features = ["derive", "std", "help"] }
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
+aws-smithy-mocks-experimental = "0.2.1"
--- a/aws/hll/aws-s3-transfer-manager/src/download.rs
+++ b/aws/hll/aws-s3-transfer-manager/src/download.rs
@@ -12,6 +12,9 @@ use aws_sdk_s3::operation::get_object::builders::{GetObjectFluentBuilder, GetObj
 
 use self::object_meta::ObjectMetadata;
 
+/// Request type for downloading a single object
+#[derive(Debug)]
+#[non_exhaustive]
 pub struct DownloadRequest {
     pub(crate) input: GetObjectInputBuilder,
 }
@@ -31,8 +34,17 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
     }
 }
 
-// FIXME - expose common fields?
+/// Response type for a single download object request.
+#[derive(Debug)]
+#[non_exhaustive]
 pub struct DownloadResponse {
     /// Object metadata
     pub object_meta: ObjectMetadata,
+}
+
+impl DownloadResponse {
+    /// Object metadata
+    pub fn object_meta(&self) -> &ObjectMetadata {
+        &self.object_meta
+    }
 }
--- a/aws/hll/aws-s3-transfer-manager/src/download.rs
+++ b/aws/hll/aws-s3-transfer-manager/src/download.rs
@@ -1,0 +1,38 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+mod discovery;
+mod handle;
+mod header;
+mod object_meta;
+
+use aws_sdk_s3::operation::get_object::builders::{GetObjectFluentBuilder, GetObjectInputBuilder};
+
+use self::object_meta::ObjectMetadata;
+
+pub struct DownloadRequest {
+    pub(crate) input: GetObjectInputBuilder,
+}
+
+// FIXME - should probably be TryFrom since checksums may conflict?
+impl From<GetObjectFluentBuilder> for DownloadRequest {
+    fn from(value: GetObjectFluentBuilder) -> Self {
+        Self {
+            input: value.as_input().clone(),
+        }
+    }
+}
+
+impl From<GetObjectInputBuilder> for DownloadRequest {
+    fn from(value: GetObjectInputBuilder) -> Self {
+        Self { input: value }
+    }
+}
+
+// FIXME - expose common fields?
+pub struct DownloadResponse {
+    /// Object metadata
+    pub object_meta: ObjectMetadata,
+}
--- a/aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
+++ b/aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
@@ -84,8 +84,10 @@ pub(super) async fn discover_obj(
         }
         ObjectDiscoveryStrategy::RangedGet(range) => {
             let byte_range = match range.as_ref() {
-                Some(r) => ByteRange::Inclusive(*r.start(), *r.start() + handle.target_part_size),
-                None => ByteRange::Inclusive(0, handle.target_part_size),
+                Some(r) => {
+                    ByteRange::Inclusive(*r.start(), *r.start() + handle.target_part_size - 1)
+                }
+                None => ByteRange::Inclusive(0, handle.target_part_size - 1),
             };
             let r = request
                 .input
@@ -153,9 +155,10 @@ async fn discover_obj_with_get(
 
     let meta: ObjectMetadata = resp.into();
 
+    // TODO - check content size matches range
     let remaining = match range {
         Some(range) => (*range.start() + data.remaining() as u64 + 1)..=*range.end(),
-        None => (data.remaining() as u64 + 1)..=meta.total_size(),
+        None => (data.remaining() as u64)..=meta.total_size(),
     };
 
     Ok(ObjectDiscovery {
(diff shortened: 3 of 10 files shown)


aws/hll/aws-s3-transfer-manager/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+publish = false
+
+
+[dependencies]
+aws-sdk-s3 = { version = "1.22.0", features = ["behavior-version-latest", "test-util"] }
+aws-smithy-http = "0.60.7"
+tracing = "0.1"
+async-trait = "0.1.80"
+# FIXME - upgrade to hyper 1.x
+hyper = { version = "0.14.28", features = ["client"] }
```
Velfi, line 18: It's possible to use hyper 1.0 right now. We should do so if possible.

aws/hll/aws-s3-transfer-manager/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+aws-smithy-types = "1.1.8"
+bytes = "1"
+thiserror = "1.0.58"
+aws-types = "1.2.0"
+tokio = { version = "1.37.0", features = ["rt-multi-thread", "io-util", "sync", "fs", "macros"] }
+aws-smithy-runtime-api = "1.4.0"
+async-channel = "2.2.1"
```
Velfi, line 25: consider sorting your deps

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -0,0 +1,38 @@
...
+    fn from(value: GetObjectInputBuilder) -> Self {
+        Self { input: value }
+    }
+}
+
+// FIXME - expose common fields?
+pub struct DownloadResponse {
+    /// Object metadata
+    pub object_meta: ObjectMetadata,
+}
```
Velfi, line 38: we should expose any of these fields directly if we ever plan to evolve this struct. Additionally, it should be marked with the `#[non_exhaustive]` attribute.

aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
```diff
@@ -0,0 +1,337 @@
...
+    let mut resp = resp.map_err(|e| error::DownloadError::DiscoverFailed(e.into()))?;
+
+    // take the body so we can convert the metadata
+    let empty_stream = ByteStream::new(SdkBody::empty());
+    let body = mem::replace(&mut resp.body, empty_stream);
+
+    let data = body
+        .collect()
+        .await
+        .map_err(|e| error::DownloadError::DiscoverFailed(e.into()))?;
```
Velfi, line 157: ideally this error's constructor would take an `impl Error` or whatever the `into` is doing. Then this line could be

```
.map_err(error::DownloadError::DiscoverFailed)?;
```

aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
```diff
@@ -0,0 +1,337 @@
...
+        download::{
+            discovery::{
+                discover_obj, discover_obj_with_get, discover_obj_with_head,
+                ObjectDiscoveryStrategy,
+            },
+            handle::DownloadHandle,
+            header::ByteRange,
+        },
+        MIN_PART_SIZE,
+    };
```
Velfi, line 185: I love doing imports like this but it's not in keeping with the style of imports in our other crates. Consider updating this to match our established style.

aws/hll/aws-s3-transfer-manager/src/download/handle.rs
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/// Shared handle used across a single download request
+#[derive(Debug, Clone)]
+pub(super) struct DownloadHandle {
+    pub(crate) client: aws_sdk_s3::Client,
```
Velfi, line 9: Does this need to be `Arc` so that the clone is cheap? Do we already have an `Arc` inside the client? I can't recall.

aws/hll/aws-s3-transfer-manager/src/download/header.rs
```diff
@@ -0,0 +1,149 @@
...
+                    )))
+                } else {
+                    let spec = ByteRange::from_str(range).map_err(|_| {
+                        error::invalid_meta_request(format!("invalid range header {}", s))
+                    })?;
+                    Ok(Range(spec))
+                }
+            }
+            _ => Err(error::invalid_meta_request(format!(
+                "unsupported byte range header format {}",
```
Velfi, line 61: If an error would make its way to a user, I think they should include calls to action that say what the user should do about their mistake.

aws/hll/aws-s3-transfer-manager/src/download/object_meta.rs
```diff
@@ -0,0 +1,147 @@
...
+ */
+
+use aws_sdk_s3::operation::{get_object::GetObjectOutput, head_object::HeadObjectOutput};
+
+// TODO(design): how many of these fields should we expose?
+// TODO(docs): Document fields
+
+/// Object metadata other than the body that can be set from either `GetObject` or `HeadObject`
+#[derive(Debug, Clone)]
+pub struct ObjectMetadata {
```
Velfi, line 13: Is exposing all of these fields intentional?

aws/hll/aws-s3-transfer-manager/src/download/object_meta.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub delete_marker: Option<bool>,
+    pub accept_ranges: Option<String>,
```
Velfi, line 15: Consider making this type a `Cow` so that it can avoid allocation in the case of the user providing a `&'static str`.

aws/hll/aws-s3-transfer-manager/src/error.rs
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use aws_smithy_types::byte_stream;
+use std::io;
+
+// TODO(design): revisit errors
```
Velfi, line 9: Whomever does look at this should read https://smithy-lang.github.io/smithy-rs/design/rfcs/rfc0022_error_context_and_compatibility.html

aws/hll/aws-s3-transfer-manager/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+//! # Crate Features
+//!
+//! - `test-util`: Enables utilities for unit tests. DO NOT ENABLE IN PRODUCTION.
+
+#![warn(
+    // TODO - re-enable missing_docs,
+    rustdoc::missing_crate_level_docs,
+    unreachable_pub,
+    rust_2018_idioms
+)]
```
Velfi, line 21: We should also set up `cargo-check-external-types` for this project.

aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
aajtodd: I think this is all auto imported...will have to sort out what IDE setting to change 🤔

aws/hll/aws-s3-transfer-manager/src/download/handle.rs
aajtodd: I believe clients are already cheap to clone. Will double check...

indeed they seem to be
```rust
#[derive(::std::clone::Clone, ::std::fmt::Debug)]
pub struct Client {
    handle: ::std::sync::Arc<Handle>,
}
```

aws/hll/aws-s3-transfer-manager/src/download/object_meta.rs
aajtodd: They are all the common fields between `HeadObject` and `GetObject` and would be public on the response types for those. This is certainly up for design/debate. I would expect though that we expose these somehow as the lowest level download primitive is still "download a single object...just do it intelligently/fast" (i.e. we still need to expose the metadata fields of the object somewhere).

aws/hll/aws-s3-transfer-manager/Cargo.toml
aajtodd: We will. Baby steps. I figure we'll get it working and then start optimizing and checking off fixmes/todos as it makes sense.

aws/hll/aws-s3-transfer-manager/Cargo.toml
aajtodd: Will something do it for me? Automation is the key to getting me to do things 😉

aws/hll/aws-s3-transfer-manager/src/download/header.rs
```diff
@@ -0,0 +1,149 @@
...
+
+use core::fmt;
+use std::str::FromStr;
+
+use crate::error;
+
+/// Representation of `Range` header.
+/// NOTE: S3 only supports a single bytes range this is a simplified representation
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct Range(pub(crate) ByteRange);
```
landonxjames, line 14: Since `Range` is already a type in `std` that seems like it might be used at some point in this crate should we call this `RangeHeader`?

aws/hll/aws-s3-transfer-manager/src/download/object_meta.rs
```diff
@@ -0,0 +1,147 @@
...
+                let total = range.split_once('/').map(|x| x.1).expect("content range total");
+                total.parse().expect("valid range total")
+            }
+            (Some(length), None) => length as u64,
+            (None, None) => panic!("total object size cannot be calculated without either content length or content range headers")
+        }
+    }
+}
+
+impl From<GetObjectOutput> for ObjectMetadata {
```
landonxjames, line 65: Before my last project wrapped up I was doing a lot of these `From` impls for types that shared a lot of fields with the same name/type and I really wanted a macro to do most of this for me. If we're going to be doing a lot of these might be a macro worth investing in now.

landonxjames (commented):
Awesome start, excited to see how fast we can get this to go!

aws/hll/aws-s3-transfer-manager/src/download/object_meta.rs
aajtodd: Yeah unsure at the moment exactly what to expose and how often we'll need this (but agree if we find ourselves doing this more we should consider that option).

aws/hll/aws-s3-transfer-manager/src/download/object_meta.rs
aajtodd: This is all returned by either `GetObject` or `HeadObject` which is also of type `Option<String>` for the corresponding fields. Users don't set these directly.

aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
aajtodd: we don't seem to have a rustfmt configuration defined anywhere, this could be controlled via https://rust-lang.github.io/rustfmt/?version=master&search=import#imports_granularity

aws/hll/aws-s3-transfer-manager/src/download/header.rs
aajtodd: it's _likely_ only used during discovery and we can disambiguate by either only referring to it as `header::Range` (which I believe we do know) or giving the import an alias (e.g. `use ... as <alias>`)

aajtodd: (that being said if you feel strongly I can change it makes no big difference to me)

aws/hll/aws-s3-transfer-manager/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[dependencies]
+aws-sdk-s3 = { version = "1.22.0", features = ["behavior-version-latest", "test-util"] }
+aws-smithy-http = "0.60.7"
+tracing = "0.1"
+async-trait = "0.1.80"
+# FIXME - upgrade to hyper 1.x
+hyper = { version = "0.14.28", features = ["client"] }
+aws-smithy-types = "1.1.8"
+bytes = "1"
+thiserror = "1.0.58"
```
ysaito1001, line 21: Do we use `thiserror`? IIRC, it incurs additional compilation time that some customers may not be happy with ([related thread](https://www.reddit.com/r/rust/comments/gj8inf/rust_structuring_and_handling_errors_in_2020/)). No need for us to make a decision as part of this PR, since it'll be merged into a feature branch, but we probably want to compare compilation time at some point in the future with/without using `thiserror`.

aajtodd: Yeah I think we will revisit the overall structure of errors and what we return from this library. I pulled in `thiserror` to make it easier to get started. I imagine we'll rip it out when we revisit errors more holistically.

ysaito1001: Hmm, although it's already in the dependency tree from `aws-smithy-protocol-test`, so I suppose using `thiserror` here isn't a big deal.
```
thiserror v1.0.61
├── aws-s3-transfer-manager v0.1.0 (/Users/awsaito/src/smithy-rs/aws/hll/aws-s3-transfer-manager)
└── aws-smithy-protocol-test v0.60.7
```

ysaito1001: Although, since it's already in the dependency tree from `aws-smithy-protocol-test`, I suppose using `thiserror` here isn't a big deal.
```
thiserror v1.0.61
├── aws-s3-transfer-manager v0.1.0 (src/smithy-rs/aws/hll/aws-s3-transfer-manager)
└── aws-smithy-protocol-test v0.60.7
    └── aws-smithy-runtime v1.5.4
        ├── aws-config v1.5.0
...
```

aajtodd: Probably profile it before doing anything but we'll be revisiting errors anyway once we get a feel for what errors should be exposed and which should be opaque. I wasn't sure yet so defaulted to exposing errors rather than wrapping them into something opaque.

aws/hll/aws-s3-transfer-manager/src/download/header.rs
ysaito1001: Yeah, maybe including an URL to the RFC (or just RFC number and the section name) for supported range formats could be helpful.

aws/hll/aws-s3-transfer-manager/src/download/header.rs
ysaito1001: As long as it's `pub(crate)`, it's a two-way door

aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
```diff
@@ -0,0 +1,328 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use std::ops::RangeInclusive;
+use std::str::FromStr;
+use std::{cmp, mem};
+
+use aws_sdk_s3::operation::get_object::builders::GetObjectInputBuilder;
+use aws_smithy_types::body::SdkBody;
+use aws_smithy_types::byte_stream::{AggregatedBytes, ByteStream};
```
ysaito1001, line 12: I think these are re-exported from `aws_sdk_s3::primitives`, so we should prefer using them wherever possible. That way, we don't need to depend on a core smithy runtime crate to get them.

Velfi (approved):

aws/hll/aws-s3-transfer-manager/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+#![warn(
+    missing_debug_implementations,
+    missing_docs,
+    rustdoc::missing_crate_level_docs,
+    unreachable_pub,
+    rust_2018_idioms
+)]
+
+pub(crate) const MEBI_BYTE: u64 = 1024 * 1024;
+pub(crate) const GIBI_BYTE: u64 = MEBI_BYTE * 1024;
```
graebm, line 25: trivial: feels weird to spell "mebi byte" as two words?
ignore if there's precedent for this in our libs already

aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
ysaito1001: Good question, I'd say it's preferable, as mentioned in [README](https://github.com/smithy-lang/smithy-rs/tree/50c825b5dbd3471bc731d7a8fbeaae78fe5a2321/rust-runtime/aws-smithy-types). Because an SDK crate is usually the one taking direct dependency on runtime crates, whoever depending on the SDK crate shouldn't need to depend on them directly (that's sort of bypassing dependency laying). John was a proponent to use re-exports as much as possible, and we've had issues/PRs to clean that up (like [this](https://github.com/smithy-lang/smithy-rs/pull/3105) or [a list of those](https://github.com/search?q=repo%3Asmithy-lang%2Fsmithy-rs+export&type=pullrequests)).

In this case, the types in question are already re-exported, so might get them from `aws_smithy_types` and remove a direct dependency on `aws_smithy_types`, if not used after cleanup.

ysaito1001: Good question, I'd say it's preferable, as mentioned in [README](https://github.com/smithy-lang/smithy-rs/tree/50c825b5dbd3471bc731d7a8fbeaae78fe5a2321/rust-runtime/aws-smithy-types). Because an SDK crate is usually the one taking direct dependency on runtime crates, whoever depending on the SDK crate shouldn't need to depend on them directly (that's sort of bypassing dependency layering). John was a proponent to use re-exports as much as possible, and we've had issues/PRs to clean that up (like [this](https://github.com/smithy-lang/smithy-rs/pull/3105) or [a list of those](https://github.com/search?q=repo%3Asmithy-lang%2Fsmithy-rs+export&type=pullrequests)).

In this case, the types in question are already re-exported, so might get them from `aws_smithy_types` and remove a direct dependency on `aws_smithy_types`, if not used after cleanup.

ysaito1001 (approved):
BTW, thanks for awesome work! This is officially the first high-level library (hll) 🚀

aws/hll/aws-s3-transfer-manager/src/download/discovery.rs
aajtodd: Ok I'll add a TODO for us to reconsider this. I don't (currently anyway) see the harm here as we're building up a new library as an abstraction on lower level parts (one of which is the `aws_sdk_s3` crate). I don't personally see any harm at the moment in re-using public types from the underlying runtime crates (especially types we _expect_ users to consume) but I also don't have a compelling reason to not follow your advice here.

aws/hll/aws-s3-transfer-manager/src/download/header.rs
```diff
@@ -58,7 +58,7 @@ impl FromStr for Range {
                 }
             }
             _ => Err(error::invalid_meta_request(format!(
-                "unsupported byte range header format {}",
+                "unsupported byte range header format {}; see https://www.rfc-editor.org/rfc/rfc9110.html#name-range for valid formats",
                 s
```
Velfi, line 62:
```suggestion
                "unsupported byte range header format {s}; see https://www.rfc-editor.org/rfc/rfc9110.html#name-range for valid formats"
```

PR #3671, opened 2024-05-29
Add invocation ID to root level “invoke” operation span

ysaito1001:
## Motivation and Context
The `invoke` debug span now includes invocation IDs for operation invocations. It is a random unique ID per client and shared across retries for the same operation.
 
Example debug trace output with the changes in this PR:
```
2024-05-29T17:50:22.230526Z DEBUG invoke{service=s3 operation=ListObjectsV2 sdk_invocation_id=4652364}:try_op:try_attempt: aws_smithy_runtime_api::client::interceptors::context: entering 'before deserialization' phase
```

## Description
Note that this invocation ID is currently for internal use only to improve debuggability when looking at logs (usually with customers). Note also that the invocation ID is not related to [that used in amz-sdk-invocation-id](https://github.com/smithy-lang/smithy-rs/pull/2626) (we [attempted to make it related](https://github.com/smithy-lang/smithy-rs/pull/3661) but decided to go for this two-way door approach).

In sum, this PR is meant to be simple, and if we are to augment the functionality (e.g. make the ID configurable or make it related to that in `amz-sdk-invocation-id`), we can do so without breaking anything.

## Testing
Relies on the existing tests in CI

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
+++ b/rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
@@ -139,6 +139,11 @@ pub async fn invoke_with_stop_point(
     runtime_plugins: &RuntimePlugins,
     stop_point: StopPoint,
 ) -> Result<InterceptorContext, SdkError<Error, HttpResponse>> {
+    let mut random_bytes = [0u8; 16];
+    fastrand::Rng::default().fill(&mut random_bytes);
+    // This is for internal use only, serving to correlate logs to a single operation.
+    // It always emits a random UUID per client and is currently not configurable.
+    let id = uuid::Builder::from_random_bytes(random_bytes).into_uuid();
     async move {
         let mut cfg = ConfigBag::base();
         let cfg = &mut cfg;
@@ -168,7 +173,7 @@ pub async fn invoke_with_stop_point(
         .maybe_timeout(operation_timeout_config)
         .await
     }
-    .instrument(debug_span!("invoke", service = %service_name, operation = %operation_name))
+    .instrument(debug_span!("invoke", service = %service_name, operation = %operation_name, sdk_invocation_id = %id))
     .await
 }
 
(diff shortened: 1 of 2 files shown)


rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -139,6 +139,11 @@ pub async fn invoke_with_stop_point(
     runtime_plugins: &RuntimePlugins,
     stop_point: StopPoint,
 ) -> Result<InterceptorContext, SdkError<Error, HttpResponse>> {
+    let mut random_bytes = [0u8; 16];
+    fastrand::Rng::default().fill(&mut random_bytes);
+    // This is for internal use only, serving to correlate logs to a single operation.
+    // It always emits a random UUID per client and is currently not configurable.
+    let id = uuid::Builder::from_random_bytes(random_bytes).into_uuid();
```
aajtodd, line 146: Any reason to not use something like what we do for [waiters](https://github.com/smithy-lang/smithy-rs/blob/5bd44f26973532e9d16e0d1b175e72e40e75f6ca/rust-runtime/aws-smithy-runtime/src/client/waiters.rs#L280)? 

UUID will obviously reduce the chance of re-use of a single ID but I would imagine in practice logs being separated enough to know they are different. Possible advantage being a shorter log line.

aajtodd (approved):

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
ysaito1001: I had looked at that in waiters and was on the fence, actually. But you mentioned a shorter log line, which makes me go for that in waiters. I believe it helps reduce cognitive cost when looking through logs.

ysaito1001: Updated in 3fee07f

PR #3673, opened 2024-05-30
implement http-body 1.0 for PathBody

Velfi:
## Motivation and Context

#1925 

## Description

Implements the v1 `http_body::Body` trait for `PathBody`. Part of the ongoing hyper v1 upgrade.
This also moves a pre-1.0 impl into its own module.

## Testing

I ported the tests too

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-types/Cargo.toml
+++ b/aws/rust-runtime/aws-types/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-types"
-version = "1.3.0"
+version = "1.3.1"
 authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "Russell Cohen <rcoh@amazon.com>"]
 description = "Cross-service types for the AWS SDK."
 edition = "2021"
@@ -27,7 +27,7 @@ hyper-rustls = { version = "0.24", optional = true, features = ["rustls-native-c
 [dev-dependencies]
 http = "0.2.4"
 tempfile = "3"
-tracing-test = "0.2.4"
+tracing-test = "=0.2.5"
 tokio = { version = "1", features = ["rt", "macros"] }
 
 [build-dependencies]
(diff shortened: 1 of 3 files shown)


rust-runtime/aws-smithy-types/src/byte_stream/bytestream_util.rs
```diff
@@ -238,407 +244,3 @@ enum State {
         bytes_left: u64,
     },
 }
-
-impl http_body_0_4::Body for PathBody {
```
Velfi, line 242: Moving this isn't breaking b/c it's only an impl and it already required the `http-body-0-4-x` feature.

aajtodd (approved):

rust-runtime/aws-smithy-types/src/byte_stream/bytestream_util/http_body_0_4_x.rs
```diff
@@ -0,0 +1,414 @@
...
+            .length(Length::Exact(file_length))
+            .build()
+            .await
+            .unwrap()
+            .into_inner();
+
+        // assert that the specified length is used as size hint
+        assert_eq!(body.content_length(), Some(file_length));
+
+        let mut body1 = body.try_clone().expect("retryable bodies are cloneable");
```
ysaito1001, line 122: Nit: `body1` existed before this PR, but maybe good to rename it now to something just `body` or `body_04` since this code is under the `body_0_4` module.

ysaito1001 (approved):

rust-runtime/aws-smithy-types/src/byte_stream/bytestream_util/http_body_0_4_x.rs
Velfi: Yeah, I think it was copied from a another test where the body was cloned several times and each clone had its own number. I've renamed this to body1 and the test still passes.

PR #2222, opened 2023-01-18
Bump Rust MSRV to 1.63.0

Nugine:
## Motivation and Context

I'm preparing a PR which upgrades `base64-simd` to v0.8.0. It needs Rust 1.63.0.

Previously: <https://github.com/awslabs/smithy-rs/pull/1938>

## Description

Bump Rust MSRV to 1.63.0

## Checklist

- [x] I have updated `CHANGELOG.next.toml` if I made changes to the smithy-rs codegen or runtime crates
- [x] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3674, opened 2024-05-30
Remove attributes annotation of Builder `struct`

edisongustavo:
The `struct` created for a Builder should not have attributes because these structs only derive from `Debug`, `PartialEq`, `Clone` and `Default`. None of these support attributes.

This becomes a problem if a plugin tries to add attributes to the `metadata` field to configure the generated code for the `struct`. In this case, the attributes will also be added to the `Builder` struct; which is wrong.

## Motivation and Context

I'm customizing client code generation by overriding the method:

```kotlin
override fun symbolProvider(base: RustSymbolProvider): RustSymbolProvider
```

I override the `symbol.expectRustMetadata().additionalAttributes` value and add extra values there. The generated code adds the correct attributes for the `struct`, however, it also adds these attributes to the `Builder`.

The `Builder` is restricted to deriving from `Debug`, `PartialEq`, `Clone` and `Default` (see [code](https://github.com/smithy-lang/smithy-rs/blob/a76dc184c2cf52d6027115f64fe78ab0c2a429e8/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt#L188)).

I believe this change was added by mistake in https://github.com/smithy-lang/smithy-rs/pull/2222/.

## Description

To solve the issue, I simply remove the offending line.

## Testing

I don't expect any existing code to break, given that this is a bug. It only manifests when someone is customizing the generated code.

## Checklist

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
@@ -189,6 +189,12 @@ class BuilderGenerator(
         metadata.derives.filter {
             it == RuntimeType.Debug || it == RuntimeType.PartialEq || it == RuntimeType.Clone
         } + RuntimeType.Default
+
+    // Filter out attributes
+    private val builderAttributes =
+        metadata.additionalAttributes.filter {
+            it == Attribute.NonExhaustive
+        }
     private val builderName = symbolProvider.symbolForBuilder(shape).name
 
     fun render(writer: RustWriter) {
@@ -306,8 +312,8 @@ class BuilderGenerator(
 
     private fun renderBuilder(writer: RustWriter) {
         writer.docs("A builder for #D.", structureSymbol)
-        metadata.additionalAttributes.render(writer)
         Attribute(derive(builderDerives)).render(writer)
+        this.builderAttributes.render(writer)
         writer.rustBlock("pub struct $builderName") {
             for (member in members) {
                 val memberName = symbolProvider.toMemberName(member)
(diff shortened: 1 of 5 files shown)


edisongustavo:
Ok, I see that we actually need Builders to have the `non_exhaustive` attribute. I'll work on fixing this.

david-perez (commented):
This LGTM but we'd want this change too to apply to the server. Make the change to these files too:

- `ServerBuilderGenerator`
- `ServerBuilderGeneratorWithoutPublicConstrainedTypes`

Bonus points if you make things DRYer and/or document the motivation for this change with either a comment or a small test.

edisongustavo:
Thanks @david-perez! I will change those files too, as you requested.

edisongustavo:
Hello @david-perez , I have added the tests as you requested, to both client and server generators.

Notice that I didn't need to make any changes to the server classes because they don't have the same bug as the client.

Please let me know if you'd like me to do anything else in this PR.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -189,6 +189,11 @@ class BuilderGenerator(
         metadata.derives.filter {
             it == RuntimeType.Debug || it == RuntimeType.PartialEq || it == RuntimeType.Clone
         } + RuntimeType.Default
+    // Filter out attributes
+    private val builderAttributes =
+        metadata.additionalAttributes.filter {
+            it == Attribute.NonExhaustive
+        }
```
david-perez, line 196:
```suggestion
    // Filter out any attribute that isn't `#[non_exhaustive]`.
    // `#[non_exhaustive]` is added to prevent users from instantiating the builder when it doesn't have any fields.
    private val builderAttributes =
        metadata.additionalAttributes.filter {
            it == Attribute.NonExhaustive
        }
```

david-perez (approved):
Thanks for the tests!

edisongustavo:
Hello, how could we expedite this to be merged and available in the next release?

aajtodd (approved):

edisongustavo:
I see some linting checks failed. I'll resolve these and update the branch

edisongustavo:
I'm sorry I didn't see the `./gradlew ktlint --format` option. I can run everything locally now, so the next set of tests should pass.

edisongustavo:
Hello, the failing check is ** ci / Ask maintainer to run the PR bot and canary workflows**.

PR #3678, opened 2024-06-03
Do not set `RUSTFLAGS` environment variable when invoking Cargo commands via Gradle tasks

david-perez:
Otherwise, if you generate a crate and compile it using Gradle, like for
example using the invocation:

```
./gradlew -P modules='simple' -P cargoCommands='test' codegen-server-test:build
```

And then manually run a `cargo` command within the generated crate
directory, or open the project using `rust-analyzer`, Cargo will
re-compile the project from scratch, with
`CARGO_LOG=cargo::core::compiler::fingerprint=trace` reporting that
flags have changed since the last compilation.

Instead, it's best if we persist these flags to `.cargo/config.toml`, so
all `cargo` invocations, either through Gradle, manually, or through
`rust-analyzer`, use the same set. This way, if no files were changed,
subsequent compilations since code generation will truly be no-ops, with
Cargo reusing all artifacts.

Note this commit fixes a regression that was introduced when `--cfg
aws_sdk_unstable` was introduced in #2614, since I fixed this the first
time back in #1422.

----

### DIFF
--- a/aws/sdk-adhoc-test/build.gradle.kts
+++ b/aws/sdk-adhoc-test/build.gradle.kts
@@ -20,7 +20,6 @@ java {
 }
 
 val smithyVersion: String by project
-val defaultRustDocFlags: String by project
 val properties = PropertyRetriever(rootProject, project)
 
 val pluginName = "rust-client-codegen"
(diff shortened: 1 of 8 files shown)


Velfi (approved):
Thanks for this fix David

david-perez:
I'll pull out the fixes for the Clippy warnings in the generated server code to a separate PR before merging this.

drganjoo (approved):

PR #3679, opened 2024-06-03
Fix S3 ListParts pagination infinite loop

landonxjames:
## Motivation and Context

The paginator for the S3 `ListParts` operation could loop forever: https://github.com/awslabs/aws-sdk-rust/issues/1143. This is because most paginated operations use an empty next token as an indication that pagination is exhausted. `ListParts` instead sets the final next token as `0` causing the pagination to loop back to the first page and loop forever. Instead of an empty next token `ListParts` uses `IsTruncated =  false` to indicate that pagination has been exhausted.  

## Description

* Added a new trait `isTruncatedPaginatorTrait`
* Add that trait to the S3 `ListPartsOutput` shape
* Use the presence of that trait to vary the logic setting the `is_empty` value in the paginator.
  * If the trait is absent it looks for an empty next token as always 
  * if the trait is present the value is set based on the value of the response's `is_truncated` field

## Testing

* Added integration test confirming that pagination terminates when a response contains `is_truncated = false`

**Note:** I'm still working on turning this into a model test rather than an S3 specific integration test, but I wanted to get some feedback on the actual fix while I'm figuring that out)

## Checklist

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
```diff
@@ -64,6 +65,7 @@ class S3Decorator : ClientCodegenDecorator {
         setOf(
             ShapeId.from("com.amazonaws.s3#CopyObject"),
         )
+    private val operationsWithIsTruncatedPaginator = setOf(ShapeId.from("com.amazonaws.s3#ListPartsOutput"))
```
aajtodd, line 68: **suggestion**: Processing the model and adding the trait could be it's own codegen decorator/integration. It isn't specific to S3 other than S3 has the only operation we know of right now with this, it could possibly be that other operations in the service fabric use something similar.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/PaginatorGenerator.kt
```diff
@@ -211,9 +223,37 @@ class PaginatorGenerator private constructor(
...
+     * In this case we use a false value of isTruncated as the only indicator that
+     * the pagination is exhausted.
+     * */
+    private fun isEmptySetter() =
+        writable {
+            if (isTruncatedPaginator) {
+                rustTemplate(
+                    """
+                    // Pagination is exhausted when `is_truncated` is false
+                    let is_empty = !resp.is_truncated.unwrap_or(false);
```
aajtodd, line 244: **nit**: This assumes the output shape member name is `is_truncated` which while likely true we could probably be more precise by having the trait include the member shape ID of the truncation member.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/PaginatorGenerator.kt
```diff
@@ -211,9 +223,37 @@ class PaginatorGenerator private constructor(
...
 
+    /** Generate code to calculate the value of is_empty. For most paginators this
+     * is indicated by the next token being the empty string. But for paginators
+     * with the isTruncatedPaginator trait the next token is not necessarily empty.
+     * (ex: for s3 ListParts the final next token is "0" when pagination is complete,
+     * causing the paginator to go back to the first page and loop forever)
+     * In this case we use a false value of isTruncated as the only indicator that
+     * the pagination is exhausted.
+     * */
+    private fun isEmptySetter() =
```
aajtodd, line 238: If you wanted to go even more down the rabbit hole you could declare a [section](https://github.com/smithy-lang/smithy-rs/blob/fe36727499a7a74e42a2beca7e76d0b433871181/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/Customization.kt#L30) for this bit of code and then register a customization for that section that does something different. You would then basically be able to move all of this into a single `TruncatedPaginatorDecorator` that handles both the model transform and the actual codegen logic.

aajtodd (approved):
Left a few things to consider but nothing blocking. Looks good overall.

CHANGELOG.next.toml
```diff
@@ -27,4 +27,16 @@ author = "Velfi"
...
+message = "Fix S3 ListParts API paginator infinite loop."
+references = ["aws-sdk-rust#1143"]
+meta = { "breaking" = false, "tada" = false, "bug" = true }
+author = "landonxjames"
+
+[[smithy-rs]]
+message = "Fix S3 ListParts API paginator infinite loop."
+references = ["aws-sdk-rust#1143"]
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client"}
+author = "landonxjames"
```
ysaito1001, line 42: Nit: since this work is specific to AWS SDK for Rust, I think this entry for `smithy-rs` (i.e. a generic SDK code-generated from smithy) should be removed.

ysaito1001 (approved):
Thanks for the effort put into adding codegen tests!

CHANGELOG.next.toml
landonxjames: Removed in https://github.com/smithy-lang/smithy-rs/pull/3679/commits/a14a2eacf112eae221c3ca3abc47b34af6f360bb

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
landonxjames: Refactored to its own decorator in https://github.com/smithy-lang/smithy-rs/pull/3679/commits/a14a2eacf112eae221c3ca3abc47b34af6f360bb

landonxjames:
Merging this for now to make sure it goes in with the next smithy-rs release since it fixes an actual customer bug. Going to connect with @aajtodd on Monday about the unaddressed suggestions to hopefully get those in as well.

Issue #139, opened 2021-01-13
[Machine learning customization] Endpoint customization


rcoh:
For Operation Predict, replace request endpoint with value from PredictEndpoint input member.

jdisanti:
This might be solved by endpoints 2.0. Adding the `retriage` tag to confirm later.

rcoh:
This will be solved by EP2 if it isn't already

aajtodd:
As of 6/5/2024 the machine learning model does not include endpoint rules for this yet.

PR #3682, opened 2024-06-05
additional service specific protocol tests

aajtodd:
## Motivation and Context
There are several service specific http request/response tests from Smithy. These are similar to the protocol tests but apply only to a specific service. We were tasked with ensuring some of the S3 URI related tests were captured.

## Description

We currently were running the API gateway related tests as part of the `aws/sdk-adhoc-test` package. There are also tests for [machinelearning](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/awsJson1_1/services/machinelearning.smithy), [glacier](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/services/glacier.smithy), and [s3](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restXml/services/s3.smithy). I attempted to add all of these but hit issues.

1. `machinelearning` can be enabled but we have yet to actual implement the [customization](https://github.com/smithy-lang/smithy-rs/issues/139) that would allow this test to pass.
2. `s3` pulls in `aws-config` as a dev dependency for _reasons_ and this causes conflicts between relocated and non-relocated runtime crates. The `sdk-adhoc-test` package does not implement any of the runtime relocation and Cargo.toml processing that the `sdk` build does (nor do I want to pursue that).

For now I've enabled the `glacier` tests and copied the S3 specific tests into `s3-tests.smithy` so that they are built and tested with the actual S3 model. In the future we can either (1) fix the build issues allowing us to remove these tests in favor of the ones defined in Smithy or (2) ideally these tests get implemented in the real upstream service models and we don't have to do anything special to get them.

## Testing
Tested locally and CI

----

### DIFF
--- a/aws/sdk-adhoc-test/build.gradle.kts
+++ b/aws/sdk-adhoc-test/build.gradle.kts
@@ -73,14 +73,12 @@ val allCodegenTests = listOf(
         "com.amazonaws.glacier#Glacier",
         "glacier",
     ),
+    // TODO(https://github.com/smithy-lang/smithy-rs/issues/139) - we assume this will be handled by EP2.0 rules but
+    //  the machinelearning service model has yet to be updated to include rules that handle the expected customization
     // baseTest(
     //     "com.amazonaws.machinelearning#AmazonML_20141212",
     //     "machinelearning",
     // ),
-    // baseTest(
-    //     "com.amazonaws.s3#AmazonS3",
-    //     "s3"
-    // )
 )
 
 project.registerGenerateSmithyBuildTask(rootProject, pluginName, allCodegenTests)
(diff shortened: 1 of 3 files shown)


ysaito1001 (approved):
[This](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restXml/services/s3.smithy#L54) doesn't need to be copied to `s3-tests.smithy` either because it's already existing or irrelevant (just to better understand)?

aajtodd:
> [This](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restXml/services/s3.smithy#L54) doesn't need to be copied to `s3-tests.smithy` either because it's already existing or irrelevant (just to better understand)?

These are all endpoint related tests that predate EP2.0. There should be EP2.0 tests in S3's model that cover these and more at this point.

Issue #3194, opened 2023-11-14
Deny warnings in generated crates


david-perez:
We already `--deny warnings` in:

- integration tests (crates generated by the `codegen-{core,client,server}-test` Gradle subprojects),
- `{client,server}IntegrationTest`s; and
- `RustWriter.compileAndTest`, via the `RUSTFLAGS` [we bake in the CI Docker image](https://github.com/smithy-lang/smithy-rs/blob/0f9ada693d7dbafef6a34400399821a6a27e9e68/tools/ci-build/Dockerfile#L194-L206) (note though that we should delete usages of `RustWriter.compileAndTest` https://github.com/smithy-lang/smithy-rs/issues/3193).

However, we do not deny warnings in `TestWriterDelegator.compileAndTest`; we generate crates with tests that have warnings.

It seems worthwhile to modify the tests so that they don't generate warnings to be consistent across the board. If a test can't be modified so as not to emit warnings, that specific test should opt-in into allowing warnings.

david-perez:
For the record, as of writing (cfd89422e639f53b380f8f99641df6792f7848d1), there's 210 tests in the server that fail to compile if we `--deny warnings`.
[failing-tests.txt](https://github.com/smithy-lang/smithy-rs/files/13740773/failing-tests.txt)

PR #3684, opened 2024-06-06
Squash assorted Clippy and Rust doc warnings in codegen integration tests

david-perez:
This commit gets rid of some Clippy and Rust doc warnings we produce in
generated code, which are surfaced in our codegen integration tests. The
aim is that eventually we will be able to deny Clippy and Rust doc
warnings in #3678 (despite us baking in `RUSTDOCFLAGS="-D warnings"` and
`RUSTFLAGS="-D warnings"` in our CI Docker image, we curently clear
these in codegen integration tests). Note that denying compiler warnings
is separately tracked in #3194.

The commit contains fixes for the following:

- Unconditionally referring to the `crate::types` module in Rust docs when
  the Smithy model is such that it is not generated.
- Broken intra-crate doc links for client documentation.
- Incorrectly escaping Rust reserved keywords when referring to
  operation names in Rust docs.
- An unnecessary semi-colon when rendering additional client
  plugins.
- Not escaping brackets in text when rendering Rust docs containing
  HTML, that are interpreted as broken links.
- A broken link to unit variants of unions.
- Using `TryFrom` instead of `From` when infallibly converting from
  `&str` to `String` when deserializing an `@httpPayload` in the server.
- Using a redundant closure with `unwrap_or_else` when falling back to a
  `@default` boolean or number value for a `document` shape, instead of
  `unwrap_or`.
- Not opting into `clippy::enum_variant_names` when rendering the
  `Operation` enum in the server (since we render the operation names
  that were modeled as-is).

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
@@ -8,6 +8,7 @@ package software.amazon.smithy.rust.codegen.core.rustlang
 import org.intellij.lang.annotations.Language
 import org.jsoup.Jsoup
 import org.jsoup.nodes.Element
+import org.jsoup.select.Elements
 import software.amazon.smithy.codegen.core.CodegenException
 import software.amazon.smithy.codegen.core.Symbol
 import software.amazon.smithy.codegen.core.SymbolDependencyContainer
@@ -336,10 +337,10 @@ fun <T : AbstractCodeWriter<T>> T.docsOrFallback(
             docs("_Note: ${it}_")
         }
     } else if (autoSuppressMissingDocs) {
+        // Otherwise, suppress the missing docs lint for this shape since
+        // the lack of documentation is a modeling issue rather than a codegen issue.
         rust("##[allow(missing_docs)] // documentation missing in model")
     }
-    // Otherwise, suppress the missing docs lint for this shape since
-    // the lack of documentation is a modeling issue rather than a codegen issue.
     return this
 }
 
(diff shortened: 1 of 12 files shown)


aajtodd (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt
```diff
@@ -439,16 +440,56 @@ fun RustWriter.deprecatedShape(shape: Shape): RustWriter {
...
+ *     <p>Text without brackets</p>
+ *     <div>Some text with \[brackets\]</div>
+ *     <span>Another \[example, 1\]</span>
+ * </body>
+ * ```
+ *
+ * This is useful when rendering Rust docs, since text within brackets might get misinterpreted as broken Markdown doc
+ * links (`[link text](https://example.com)`).
+ **/
+private fun Element.escapeBracketsInText() {
```
drganjoo, line 480: Nit: Just a thought: Should we do this *only if* the box brackets are not followed by text in parentheses? This would allow Smithy model authors to write linkable text in their documentation. For example, `[this is clickable](https://example.com)` should not be escaped.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorCommon.kt
```diff
@@ -108,15 +110,17 @@ fun generateFallbackCodeToDefaultValue(
...
-            is NumberShape, is EnumShape, is BooleanShape -> {
-                writer.rustTemplate(".unwrap_or(#{DefaultValue:W})", "DefaultValue" to defaultValue)
-            }
-            // Values for the Rust types of the rest of the shapes require heap allocations, so we calculate them
-            // in a (lazily-executed) closure for slight performance gains.
-            else -> {
-                writer.rustTemplate(".unwrap_or_else(|| #{DefaultValue:W})", "DefaultValue" to defaultValue)
-            }
+        val node = member.expectTrait<DefaultTrait>().toNode()!!
+        if ((targetShape is DocumentShape && (node is BooleanNode || node is NumberNode)) ||
```
drganjoo, line 114: Thinking out loud: Would defining a function, such as `isBasicRustType(targetShape)`, provide any advantages, or is this the only place where we check for this?

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriterTest.kt
```diff
@@ -99,6 +99,21 @@ class RustWriterTest {
...
+        val output =
+            normalizeHtml(
+                """
+                <a>Link without href attribute</a>
+                <div>Some text with [brackets]</div>
+                <span>] mismatched [ is escaped too</span>
+                """,
+            )
+        output shouldContain "<code>Link without href attribute</code>"
+        output shouldContain "Some text with \\[brackets\\]"
```
drganjoo, line 113: Nit: Should we use `"""Some text with \[brackets\]"""` for increased readability?

drganjoo (approved):

Issue #3672, opened 2024-05-30
Unable to build `aws-smithy-checksums` in MacOS ARM using nightly toolchains post 14th May 


navasvarela:
When trying to compile a project in MacOS ARM (Sonoma), using any nightly toolchain after 2024-05-14, the following error occurs:
```
error[E0635]: unknown feature `stdarch_arm_crc32`
  --> USER_HOME/.cargo/registry/src/index.crates.io-6f17d22bba15001f/crc32c-0.6.5/src/lib.rs:23:60
   |
23 | #![cfg_attr(all(target_arch = "aarch64", nightly), feature(stdarch_arm_crc32))]
   |                                                            ^^^^^^^^^^^^^^^^^
```

`aws-smithy-checksums` depends on crc32c version 0.6.5. This version of crc32c uses a feature that was removed on 14th May and causes compilation errors (see https://github.com/rust-lang/stdarch/pull/1573). This problem has been fixed in subsequent releases of crc32c. It seems that the fix for this should be relatively simple.

aajtodd:
Pinning crc32c was done [here](https://github.com/smithy-lang/smithy-rs/pull/3665) to work around build failures and ensure the SDK can continue to be released.

The underlying issue doesn't appear to be resolved/released and so until then we'll be pinned to 0.6.5.

orlp:
@aajtodd That does mean anything that uses `aws-smithy-checksums`, which includes `aws-sdk-s3` simply does not build on the latest nightly on ARM, which includes all developers on Apple M* laptops.

Andrewp2:
I think that commit has made it into 0.6.8
https://github.com/zowens/crc32c/issues/70
https://crates.io/crates/crc32c/0.6.8

ysaito1001:
We just released a new [aws-smithy-checksums](https://crates.io/crates/aws-smithy-checksums/0.60.10/dependencies) that depends on `crc32c` version 0.6.8. Could you try compiling your project and see if it works now

Andrewp2:
My project is working, which compiles to AWS Lambda ARM. I don't have a mac device to check there.

navasvarela:
I can confirm that this works now for me in my Mac

PR #3692, opened 2024-06-12
Pin crc32c to a newer working version

ysaito1001:
## Motivation and Context
https://github.com/smithy-lang/smithy-rs/issues/3672

## Description
Pin to a newer working version of `crc32c` (will still pin to a particular version to avoid a breaking change, until the lockfile protection is available)

## Testing
- Verified locally build of `aws-smithy-checksums` with the updated `crc32c`

----

### DIFF
--- a/rust-runtime/aws-smithy-checksums/Cargo.toml
+++ b/rust-runtime/aws-smithy-checksums/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-checksums"
-version = "0.60.9"
+version = "0.60.10"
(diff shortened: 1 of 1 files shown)


aajtodd (approved):
Do we want to loosen this to `0.6.8` rather than `=0.6.8` given pending lockfile adoption?

ysaito1001:
> Do we want to loosen this to `0.6.8` rather than `=0.6.8` given pending lockfile adoption?

Will keep `=` just in case until lockfile adoption is fully in place.

PR #3695, opened 2024-06-13
exclude invoke operations from SSP

Velfi:
## Motivation and Context

aws-sdk-rust#1166

## Description

SSP shouldn't be enabled for long-running operations that transfer no data.

## Testing

Checked the generated lambda SDK to see that SSP has been disabled for the operations specified.

## Checklist

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -10,3 +10,27 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
+
+[[aws-sdk-rust]]
+message = """
+Stalled stream protection will no longer be applied to the following Lambda operations: [Invoke], [InvokeAsync], [InvokeWithResponseStream].
+
+[Invoke]: https://docs.rs/aws-sdk-lambda/latest/aws_sdk_lambda/client/struct.Client.html#method.invoke
+[InvokeAsync]: https://docs.rs/aws-sdk-lambda/latest/aws_sdk_lambda/client/struct.Client.html#method.invoke_async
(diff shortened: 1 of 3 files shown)


aajtodd (approved):
**suggestion**: I wonder if a more generalized `DisableStalledStreamProtectionDecorator` that centralizes this wouldn't be better. We'd add the existing S3 operation and the new Lambda ones to a single decorator.

I'm not going to block on this as this works but an alternative to consider that may be better for long term maintenance and discoverability.

CHANGELOG.next.toml
```diff
@@ -9,4 +9,28 @@
...
+message = """
+Stalled stream protection will no longer be applied to the following Lambda operations: [Invoke], [InvokeAsync], [InvokeWithResponseStream].
+
+[Invoke]: https://docs.rs/aws-sdk-lambda/latest/aws_sdk_lambda/client/struct.Client.html#method.invoke
+[InvokeAsync]: https://docs.rs/aws-sdk-lambda/latest/aws_sdk_lambda/client/struct.Client.html#method.invoke_async
+[InvokeWithResponseStream]: https://docs.rs/aws-sdk-lambda/latest/aws_sdk_lambda/client/struct.Client.html#method.invoke_with_response_stream
+"""
+references = ["aws-sdk-rust#1166", "smithy-rs#3639"]
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client"}
+author = "Velfi"
```
ysaito1001, line 36: nit: this entry for `[[smithy-rs]]` probably should go away since when generic SDK users see `smithy-rs` release notes referring to `Lambda`, they might go "?"

ysaito1001 (approved):
Thanks for the quick fix!

CHANGELOG.next.toml
Velfi: ah, true

PR #3697, opened 2024-06-14
s3-tm naive multipart download

aajtodd:
## Description

Adds a "naive" implementation of a multipart download as well as an example to exercise and play with the API. I haven't had much time to play around with it but the initial performance seems off (slow) and requires further investigation.

The basic idea in this initial implementation:
* do initial object discovery (possibly fetching first chunk of data with it)
* spin up a worker to distribute work, this worker simply takes the remaining total range of the object and chunks it by the part size and puts each chunk range to fetch onto a bounded worker channel. 
* spin up N workers that read off the "work to do" channel. They pull requests off the channel for a single range, fetch it, and place the result onto completion channel
* consumer reads from a new `Body` type that handles sequencing the chunks

I fully expect the `Body` type to change. Using `AggregatedBytes` seems limiting and we don't have much control over it. It's a very thin wrapper on `SegmentedBuf` from `bytes-util`. 

### DIFF
--- a/aws/hll/aws-s3-transfer-manager/examples/cp.rs
+++ b/aws/hll/aws-s3-transfer-manager/examples/cp.rs
@@ -30,9 +30,14 @@ pub struct Args {
     /// Destination to copy to <S3Uri | Local>
     #[arg(required = true)]
     dest: TransferUri,
+
     /// Number of concurrent uploads/downloads to perform.
     #[arg(long, default_value_t = 8)]
     concurrency: usize,
+
+    /// Part size to use
+    #[arg(long, default_value_t = 8388608)]
+    part_size: u64,
 }
 
 #[derive(Clone, Debug)]
@@ -114,6 +119,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let tm = Downloader::builder()
         .sdk_config(config)
         .concurrency(args.concurrency)
+        .target_part_size(args.part_size)
         .build();
 
     let (bucket, key) = args.source.expect_s3().parts();
--- a/aws/hll/aws-s3-transfer-manager/examples/cp.rs
+++ b/aws/hll/aws-s3-transfer-manager/examples/cp.rs
@@ -17,7 +17,7 @@ use tokio::io::AsyncWriteExt;
 
 type BoxError = Box<dyn Error + Send + Sync>;
 
-const ONE_GIGABYTE: u64 = 1024 * 1024 * 1024;
+const ONE_MEBIBYTE: u64 = 1024 * 1024;
 
 #[derive(Debug, Clone, clap::Parser)]
 #[command(name = "cp")]
@@ -102,16 +102,14 @@ fn invalid_arg(message: &str) -> ! {
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
     tracing_subscriber::fmt::init();
-    let args = Args::parse();
-    println!("{:#?}", args);
+    let args = dbg!(Args::parse());
 
+    use TransferUri::*;
     match (&args.source, &args.dest) {
-        (TransferUri::Local(_), TransferUri::S3(_)) => todo!("upload not implemented yet"),
-        (TransferUri::Local(_), TransferUri::Local(_)) => {
-            invalid_arg("local to local transfer not supported")
-        }
-        (TransferUri::S3(_), TransferUri::Local(_)) => (),
-        (TransferUri::S3(_), TransferUri::S3(_)) => invalid_arg("s3 to s3 transfer not supported"),
+        (Local(_), S3(_)) => todo!("upload not implemented yet"),
+        (Local(_), Local(_)) => invalid_arg("local to local transfer not supported"),
+        (S3(_), Local(_)) => (),
+        (S3(_), S3(_)) => invalid_arg("s3 to s3 transfer not supported"),
     }
 
     let config = aws_config::from_env().load().await;
@@ -143,19 +141,13 @@ async fn main() -> Result<(), Box<dyn Error>> {
         }
     }
 
-    let elapsed_secs = start.elapsed().as_secs_f64();
-
+    let elapsed = start.elapsed();
     let obj_size = handle.object_meta.total_size();
-    let obj_size_gigabytes = obj_size as f64 / ONE_GIGABYTE as f64;
-    let obj_size_gigabits = obj_size_gigabytes * 8.0;
+    let obj_size_mebibytes = obj_size as f64 / ONE_MEBIBYTE as f64;
 
     println!(
-        "downloaded {} bytes ({} GiB) in {} seconds; GiB/s: {}; Gbps: {}",
-        obj_size,
-        obj_size_gigabytes,
-        elapsed_secs,
-        obj_size_gigabytes / elapsed_secs,
-        obj_size_gigabits / elapsed_secs
+        "downloaded {obj_size} bytes ({obj_size_mebibytes} MiB) in {elapsed:?}; MiB/s: {}",
+        obj_size_mebibytes / elapsed.as_secs_f64(),
     );
 
     Ok(())
--- a/aws/hll/aws-s3-transfer-manager/examples/cp.rs
+++ b/aws/hll/aws-s3-transfer-manager/examples/cp.rs
@@ -1,0 +1,156 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+use std::error::Error;
+use std::path::PathBuf;
+use std::str::FromStr;
+use std::{mem, time};
+
+use aws_s3_transfer_manager::download::Downloader;
+
+use aws_s3_transfer_manager::download::body::Body;
+use aws_sdk_s3::operation::get_object::builders::GetObjectInputBuilder;
+use clap::{CommandFactory, Parser};
+use tokio::fs;
+use tokio::io::AsyncWriteExt;
+
+type BoxError = Box<dyn Error + Send + Sync>;
+
+const ONE_GIGABYTE: u64 = 1024 * 1024 * 1024;
+
+#[derive(Debug, Clone, clap::Parser)]
+#[command(name = "cp")]
+#[command(about = "Copies a local file or S3 object to another location locally or in S3.")]
+pub struct Args {
+    /// Source to copy from <S3Uri | Local>
+    #[arg(required = true)]
+    source: TransferUri,
+
+    /// Destination to copy to <S3Uri | Local>
+    #[arg(required = true)]
+    dest: TransferUri,
+    /// Number of concurrent uploads/downloads to perform.
+    #[arg(long, default_value_t = 8)]
+    concurrency: usize,
+}
+
+#[derive(Clone, Debug)]
+enum TransferUri {
+    /// Local filesystem source/destination
+    Local(PathBuf),
+
+    /// S3 source/destination
+    S3(S3Uri),
+}
+
+impl TransferUri {
+    fn expect_s3(&self) -> &S3Uri {
+        match self {
+            TransferUri::S3(s3_uri) => s3_uri,
+            _ => panic!("expected S3Uri"),
+        }
+    }
+
+    fn expect_local(&self) -> &PathBuf {
+        match self {
+            TransferUri::Local(path) => path,
+            _ => panic!("expected Local"),
+        }
+    }
+}
+
+impl FromStr for TransferUri {
+    type Err = BoxError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let uri = if s.starts_with("s3://") {
+            TransferUri::S3(S3Uri(s.to_owned()))
+        } else {
+            let path = PathBuf::from_str(s).unwrap();
+            TransferUri::Local(path)
+        };
+        Ok(uri)
+    }
+}
+
+#[derive(Clone, Debug)]
+struct S3Uri(String);
+
+impl S3Uri {
+    /// Split the URI into it's component parts '(bucket, key)'
+    fn parts(&self) -> (&str, &str) {
+        self.0
+            .strip_prefix("s3://")
+            .expect("valid s3 uri prefix")
+            .split_once('/')
+            .expect("invalid s3 uri, missing '/' between bucket and key")
+    }
+}
+
+fn invalid_arg(message: &str) -> ! {
+    Args::command()
+        .error(clap::error::ErrorKind::InvalidValue, message)
+        .exit()
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    tracing_subscriber::fmt::init();
+    let args = Args::parse();
+    println!("{:#?}", args);
+
+    match (&args.source, &args.dest) {
+        (TransferUri::Local(_), TransferUri::S3(_)) => todo!("upload not implemented yet"),
+        (TransferUri::Local(_), TransferUri::Local(_)) => {
+            invalid_arg("local to local transfer not supported")
+        }
+        (TransferUri::S3(_), TransferUri::Local(_)) => (),
+        (TransferUri::S3(_), TransferUri::S3(_)) => invalid_arg("s3 to s3 transfer not supported"),
+    }
+
+    let config = aws_config::from_env().load().await;
+
+    let tm = Downloader::builder()
+        .sdk_config(config)
+        .concurrency(args.concurrency)
+        .build();
+
+    let (bucket, key) = args.source.expect_s3().parts();
+    let input = GetObjectInputBuilder::default().bucket(bucket).key(key);
+
+    let mut dest = fs::File::create(args.dest.expect_local()).await?;
+    println!("dest file opened, starting download");
+
+    let start = time::Instant::now();
+
+    // TODO(aws-sdk-rust#1159) - rewrite this less naively,
+    //      likely abstract this into performant utils for single file download. Higher level
+    //      TM will handle it's own thread pool for filesystem work
+    let mut handle = tm.download(input.into()).await?;
+    let mut body = mem::replace(&mut handle.body, Body::empty());
+
+    while let Some(chunk) = body.next().await {
+        let chunk = chunk.unwrap();
+        for segment in chunk.into_segments() {
+            dest.write_all(segment.as_ref()).await?;
+        }
+    }
+
+    let elapsed_secs = start.elapsed().as_secs_f64();
+
+    let obj_size = handle.object_meta.total_size();
+    let obj_size_gigabytes = obj_size as f64 / ONE_GIGABYTE as f64;
+    let obj_size_gigabits = obj_size_gigabytes * 8.0;
+
+    println!(
+        "downloaded {} bytes ({} GiB) in {} seconds; GiB/s: {}; Gbps: {}",
+        obj_size,
+        obj_size_gigabytes,
+        elapsed_secs,
+        obj_size_gigabytes / elapsed_secs,
+        obj_size_gigabits / elapsed_secs
+    );
+
+    Ok(())
+}
--- a/aws/hll/aws-s3-transfer-manager/src/download.rs
+++ b/aws/hll/aws-s3-transfer-manager/src/download.rs
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-mod body;
+pub mod body;
 mod context;
 mod discovery;
 mod handle;
@@ -12,15 +12,15 @@ mod object_meta;
 mod worker;
 
 use crate::download::body::Body;
-use crate::download::discovery::discover_obj;
+use crate::download::discovery::{discover_obj, ObjectDiscovery};
 use crate::download::handle::DownloadHandle;
-use crate::download::worker::{chunk_downloader, distribute_work};
+use crate::download::worker::{chunk_downloader, distribute_work, ChunkResponse};
 use crate::error::TransferError;
 use crate::{MEBIBYTE, MIN_PART_SIZE};
 use aws_sdk_s3::operation::get_object::builders::{GetObjectFluentBuilder, GetObjectInputBuilder};
 use aws_types::SdkConfig;
 use context::DownloadContext;
-use std::cmp;
+use std::{cmp, mem};
 use tokio::sync::mpsc;
 use tokio::task::JoinSet;
 use tracing::Instrument;
@@ -130,6 +130,11 @@ pub struct Downloader {
 }
 
 impl Downloader {
+    /// Create a new [Builder]
+    pub fn builder() -> Builder {
+        Builder::new()
+    }
+
     /// Download a single object from S3.
     ///
     /// A single logical request may be split into many concurrent ranged `GetObject` requests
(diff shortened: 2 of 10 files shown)


aws/hll/aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -0,0 +1,162 @@
...
+    Args::command()
+        .error(clap::error::ErrorKind::InvalidValue, message)
+        .exit()
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    tracing_subscriber::fmt::init();
+    let args = Args::parse();
+    println!("{:#?}", args);
```
Velfi, line 106:
```suggestion
    let args = dbg!(Args::parse());
```

aws/hll/aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    match (&args.source, &args.dest) {
+        (TransferUri::Local(_), TransferUri::S3(_)) => todo!("upload not implemented yet"),
+        (TransferUri::Local(_), TransferUri::Local(_)) => {
+            invalid_arg("local to local transfer not supported")
+        }
+        (TransferUri::S3(_), TransferUri::Local(_)) => (),
+        (TransferUri::S3(_), TransferUri::S3(_)) => invalid_arg("s3 to s3 transfer not supported"),
+    }
```
Velfi, line 115: Consider this:

```suggestion
    use TransferUri::*;
    match (&args.source, &args.dest) {
        (Local(_), S3(_)) => todo!("upload not implemented yet"),
        (Local(_), Local(_)) => {
            invalid_arg("local to local transfer not supported")
        }
        (S3(_), Local(_)) => (),
        (S3(_), S3(_)) => invalid_arg("s3 to s3 transfer not supported"),
    }
```

aws/hll/aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -0,0 +1,162 @@
...
+    let obj_size_gigabits = obj_size_gigabytes * 8.0;
+
+    println!(
+        "downloaded {} bytes ({} GiB) in {} seconds; GiB/s: {}; Gbps: {}",
+        obj_size,
+        obj_size_gigabytes,
+        elapsed_secs,
+        obj_size_gigabytes / elapsed_secs,
+        obj_size_gigabits / elapsed_secs
+    );
```
Velfi, line 159: Consider this:
```suggestion
    let elapsed = start.elapsed();
    let obj_size = handle.object_meta.total_size();
    let obj_size_gigabytes = obj_size as f64 / ONE_GIGABYTE as f64;
    let obj_size_gigabits = obj_size_gigabytes * 8.0;

    println!(
        "downloaded {obj_size} bytes ({obj_size_gigabytes} GiB) in {elapsed:?}; GiB/s: {}; Gbps: {}",
        obj_size_gigabytes / elapsed_secs.as_secs_f64(),
        obj_size_gigabits / elapsed_secs.as_secs_f64()
    );
```

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -34,17 +49,178 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
...
-    /// Object metadata
-    pub object_meta: ObjectMetadata,
+/// Fluent style builder for [Downloader]
+#[derive(Debug, Clone)]
+pub struct Builder {
+    target_part_size_bytes: u64,
+    checksum_validation_enabled: bool,
+    // TODO(design): should we instead consider an enum here allows for not only explicit but also
+    // an "Auto" mode that allows us to control the concurrency actually used based on overall transfer and part size?
+    concurrency: usize,
```
Velfi, line 59: I think yes, but it doesn't matter unless we have to expose this. Unexposed, we're free to change the impl whenever.

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -34,17 +49,178 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
...
+    /// Size of parts the object will be downloaded in, in bytes.
+    /// The minimum part size is 5 MiB and any value given less than that will be rounded up.
+    /// Defaults is 8 MiB.
+    pub fn target_part_size(mut self, size_bytes: u64) -> Self {
+        self.target_part_size_bytes = cmp::min(size_bytes, MIN_PART_SIZE);
+        self
+    }
+
+    /// Set whether checksum validation is enabled.
+    /// Default is true.
```
Velfi, line 82: When writing doc comments, it's preferred to leave a summary line at the top, then an empty newline, and then whatever else you want to convey. This is b/c of how RustDoc will expose this doc to rust-analyzer and other things.
```suggestion
    /// Set whether checksum validation is enabled.
    ///
    /// Default is true.
```

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -34,17 +49,178 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
...
+                    .instrument(tracing::debug_span!("chunk-downloader", worker = i));
+                tasks.spawn(worker);
+            }
+        }
+
+        // drop our reference to completion sender such that when all workers drop theirs it's closed
+        drop(comp_tx);
+
+        let handle = DownloadHandle {
+            object_meta: discovery.meta,
```
aajtodd, line 200: TODO: this needs fixed, the initial object discovery is for a range/first-part which will not have the right metadata w.r.t content-length, etc. I'm not sure what we want to do here or if there are any other metadata fields that would vary when doing a "full" get object vs ranged GET(s). Presumably CRT team already knows answer to this.

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -34,17 +49,178 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
...
+        self
+    }
+
+    /// Set the concurrency level this component is allowed to use. This sets the maximum
+    /// number of concurrent in-flight requests.
+    /// Default is 8.
+    pub fn concurrency(mut self, concurrency: usize) -> Self {
+        self.concurrency = concurrency;
+        self
+    }
```
Velfi, line 100: TODO Create a breadcrumb leading users to a guide explaining how to choose the right amount of concurrency.

Velfi: Possibly worth explaining the difference between concurrency and parallelism somewhere.

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -34,17 +49,178 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
...
+) -> u64 {
+    let mut start_seq = 0;
+    if let Some(initial_data) = mem::replace(&mut discovery.initial_chunk, None) {
+        let chunk = ChunkResponse {
+            seq: start_seq,
+            data: Some(initial_data),
+            object_meta: Some(discovery.meta.clone()),
+        };
+        completed.send(Ok(chunk)).await.expect("initial chunk");
+        start_seq += 1;
```
Velfi, line 223: No iteration happening here, prefer assignment:
```suggestion
        start_seq = 1;
```

aws/hll/aws-s3-transfer-manager/src/download/body.rs
```diff
@@ -0,0 +1,223 @@
...
+ */
+use crate::download::worker::ChunkResponse;
+use crate::error::TransferError;
+use aws_smithy_types::byte_stream::AggregatedBytes;
+use std::cmp;
+use std::cmp::Ordering;
+use std::collections::BinaryHeap;
+use tokio::sync::mpsc;
+
+/// Stream of binary data representing an object's contents.
```
Velfi, line 13: What does `object` refer to? Is this body used by both requests and responses?

aws/hll/aws-s3-transfer-manager/src/download/body.rs
```diff
@@ -0,0 +1,223 @@
...
+
+            let chunk = self.inner.next().await;
+            if chunk.is_none() {
+                break;
+            }
+
+            match chunk? {
+                Ok(chunk) => self.sequencer.push(chunk),
+                Err(err) => return Some(Err(err)),
+            }
```
Velfi, line 68:
```suggestion
            match self.inner.next().await {
                Ok(None) => break;
                Ok(Some(chunk)) => self.sequencer.push(chunk),
                Err(err) => return Some(Err(err)),
            }
```

aws/hll/aws-s3-transfer-manager/src/download/body.rs
```diff
@@ -0,0 +1,223 @@
...
+        if next.is_none() {
+            return false;
+        }
+
+        next.unwrap().seq == self.next_seq
+    }
+
+    fn peek(&self) -> Option<&ChunkResponse> {
+        self.chunks.peek().map(|c| &c.0 .0)
+    }
```
Velfi, line 119: Using an `into` would be nicer imo.

aws/hll/aws-s3-transfer-manager/src/download/body.rs
```diff
@@ -0,0 +1,223 @@
...
+impl UnorderedBody {
+    fn new(chunks: Option<BodyChannel>) -> Self {
+        Self { chunks }
+    }
+
+    /// Pull the next chunk of data off the stream.
+    ///
+    /// Returns [None] when there is no more data.
+    /// Chunks returned from an [UnorderedBody] are not guaranteed to be sequenced
+    /// in the right order. Consumers are expected to sequence the data themselves
```
Velfi, line 163: I think we should prefer the term `sorted` to `sequenced` b/c `sorted` feels more common to me.

aws/hll/aws-s3-transfer-manager/src/download/handle.rs
```diff
@@ -2,10 +2,32 @@
...
+pub struct DownloadHandle {
+    /// Object metadata
+    pub object_meta: ObjectMetadata,
+
+    /// The object content
+    pub body: Body,
+
+    /// All child tasks spawned for this download
+    pub(crate) tasks: task::JoinSet<()>,
+}
```
Velfi, line 21: Consider linking to the type for "download object request"

aws/hll/aws-s3-transfer-manager/src/download/worker.rs
```diff
@@ -0,0 +1,182 @@
...
+    /// Size of this chunk request in bytes
+    pub(super) fn size(&self) -> u64 {
+        self.range.end() - self.range.start() + 1
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct ChunkResponse {
+    // the seq number
+    pub(crate) seq: u64,
```
Velfi, line 38: I think it would be preferable to always hide this and impl `PartialOrd` instead.

aws/hll/aws-s3-transfer-manager/src/download/worker.rs
```diff
@@ -0,0 +1,182 @@
...
+    // chunk data
+    pub(crate) data: Option<AggregatedBytes>,
+    // object metadata
+    pub(crate) object_meta: Option<ObjectMetadata>,
```
Velfi, line 42: What does it mean for a `ChunkResponse` to have no data or metadata?

aws/hll/aws-s3-transfer-manager/examples/cp.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Source to copy from <S3Uri | Local>
+    #[arg(required = true)]
+    source: TransferUri,
+
+    /// Destination to copy to <S3Uri | Local>
+    #[arg(required = true)]
+    dest: TransferUri,
+
+    /// Number of concurrent uploads/downloads to perform.
+    #[arg(long, default_value_t = 8)]
```
sbiscigl, line 35: should we make the default [`available_parallelism`](https://doc.rust-lang.org/std/thread/fn.available_parallelism.html)? there is some [precedent of this being done](https://github.com/libcpr/cpr/blob/master/include/cpr/threadpool.h#L16).

aws/hll/aws-s3-transfer-manager/src/download/body.rs
```diff
@@ -0,0 +1,223 @@
...
+    pub(crate) fn unordered(self) -> UnorderedBody {
+        self.inner
+    }
+
+    /// Pull the next chunk of data off the stream.
+    ///
+    /// Returns [None] when there is no more data.
+    /// Chunks returned from a [Body] are guaranteed to be sequenced
+    /// in the right order.
+    pub async fn next(&mut self) -> Option<Result<AggregatedBytes, TransferError>> {
```
graebm, line 52: Seems weird to return an optional result. Would this make more sense?
```suggestion
    pub async fn next(&mut self) -> Result<Option<AggregatedBytes>, TransferError> {
```

aws/hll/aws-s3-transfer-manager/examples/cp.rs
graebm: In async Rust, you can have 1 thread handling **MANY** concurrent HTTP requests.

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -34,17 +49,178 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
...
+    ///
+    ///     let handle = client.download(request).await?;
+    ///     // process data off handle...
+    ///     Ok(())
+    /// }
+    /// ```
+    pub async fn download(&self, req: DownloadRequest) -> Result<DownloadHandle, TransferError> {
+        // if there is a part number then just send the default request
+        if req.input.get_part_number().is_some() {
+            todo!("single part download not implemented")
```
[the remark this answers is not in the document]
aajtodd: Can you expand on what the concern is? I imagine we'll end up with another way to construct a request but it makes sense for Rust SDK users to at least provide some compatibility. 

Note these are internal usage of those types which is inevitable at some point since we're making the request through aws_sdk_s3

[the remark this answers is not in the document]
graebm: When we "split" the request, there are some options like [if_match](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/operation/get_object/builders/struct.GetObjectInputBuilder.html#method.if_match) we should copy verbatim into every sub-request.

There are other like [range](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/operation/get_object/builders/struct.GetObjectInputBuilder.html#method.range) that would require special handling: "Oh we can split this range into even smaller ranges, but still need to fall within the original range".

For something like PutObject, there's stuff like [checksum_crc32](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/operation/put_object/builders/struct.PutObjectInputBuilder.html#method.checksum_crc32) where ... if the user passes that for a large object what do we do? If we split it, then the user's checksum won't end up as the whole-object-checksum. Reject the request? Send without splitting it so the user's checksum gets used? Split it anyway, but validate that all the bytes we sent did checksum to that value? 

So... every option needs to be considered. If we do something like: "add special handling for X, Y, and Z options but copy everything else over verbatim". Then what if a new option is added to the model, and it needs special handling, but we didn't catch it and terrible things happen as a result.

I'm not dead-set against using the request options from aws_sdk_s3, but if we have that we should have some allow-list of options that we let through, and reject anything we don't know about

aws/hll/aws-s3-transfer-manager/src/download/body.rs
Velfi: I would say it's not so strange in Rust for a iteration method to return an option

aws/hll/aws-s3-transfer-manager/src/download.rs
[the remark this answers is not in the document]
aajtodd: Right, the checksum issue is one of the reasons I put a TODO that perhaps we should be using `TryFrom` (fallible version of `From`). 

> I'm not dead-set against using the request options from aws_sdk_s3, but if we have that we should have some allow-list of options that we let through, and reject anything we don't know about

We're definitely going to use these types since at some point we are invoking `aws_sdk_s3` methods requiring them BUT the external request type need not match. You'll note we don't directly take these types, only c convenience for converting from them to `DownloadRequest`. I think we'll offer this kind of convenience for aws-sdk-rust/aws_sdk_s3 users (e.g. `TryFrom<GetObjectFluentBuilder> for DownloadRequest`) but what we expose via FFI of course won't be that. 

I understand the concern and I think right now it's not as clear since we aren't providing an alternative to `GetObjectInputFluentBuilder/GetObjectInput` but I expect we will.

aws/hll/aws-s3-transfer-manager/src/download/body.rs
```diff
@@ -0,0 +1,223 @@
...
+        // TODO(aws-sdk-rust#1159, design) - do we want ChunkResponse (or similar) rather than AggregatedBytes? Would
+        //  make additional retries of an individual chunk/part more feasible (though theoretically already exhausted retries)
```
ysaito1001, line 54: Question: as "already exhausted retries" says, fetching individual chunks already benefits retries [as part of this call](https://github.com/smithy-lang/smithy-rs/pull/3697/files#diff-b91e2788fd16643b8df16c99701cd4cade84f33318fd08d063230fe1aba4400bR74-R78), right? (and if fetching one chunk fails after retries, we eventually error out at [a later point](https://github.com/smithy-lang/smithy-rs/pull/3697/files#diff-1bd049795ee294cf2e26a25903e7eef0ac42b813f0afd68652244f75bcaf2e57R140), in which case `unwrap` will panic)

aws/hll/aws-s3-transfer-manager/src/download/body.rs
```diff
@@ -0,0 +1,223 @@
...
+        sequencer.push(chunk_resp(0, None));
+        assert_eq!(sequencer.pop().unwrap().seq, 0);
+    }
+
+    #[tokio::test]
+    async fn test_body_next() {
+        let (tx, rx) = mpsc::channel(2);
+        let mut body = Body::new(rx);
+        tokio::spawn(async move {
+            for i in 0..3 {
```
ysaito1001, line 205: Can be in a separate test but is it possible we shuffle the order of elements (e.g. [2, 0, 1] or [1, 2, 0] and make sure `body.next()` works as expected? (it may help include the actual order of elements in the diagnostic message of `assert_eq!` in case the test fails and we wish to reproduce the failure)

ysaito1001, line 205: This is sort of covered by `test_sequencer` but is it possible we randomly shuffle the order of elements (e.g. [2, 0, 1] or [1, 2, 0] and make sure `body.next()` works as expected? (it may help include the actual order of elements in the diagnostic message of `assert_eq!` in case the test fails and we wish to reproduce the failure)

aws/hll/aws-s3-transfer-manager/src/download/body.rs
```diff
@@ -0,0 +1,223 @@
...
+                let data = Bytes::from(format!("chunk {i}"));
+                let aggregated = ByteStream::from(data).collect().await.unwrap();
+                let chunk = chunk_resp(i as u64, Some(aggregated));
+                tx.send(Ok(chunk)).await.unwrap();
```
ysaito1001, line 209: This is sort of covered by `test_sequencer` but is it possible we randomly shuffle the order of elements (e.g. [2, 0, 1] or [1, 2, 0] and make sure `body.next()` works as expected? (it may help include the actual order of elements in the diagnostic message of `assert_eq!` in case the test fails and we wish to reproduce the failure)

Maybe we can also add a test where a channel sender sends the `Err` variant and make sure `body.next()` correctly errors out.

aws/hll/aws-s3-transfer-manager/src/download/worker.rs
aajtodd: I wondered that as well and started to change it but we can't actually construct `AggregatedBytes` directly as it turns out. Only other advantage is it makes testing a little easier to not have to set things like `ObjectMetadata`. 

I think this could probably change as we move away from `AggregatedBytes`

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -34,17 +49,178 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
...
+            checksum_validation_enabled: true,
+            concurrency: 8,
+            sdk_config: None,
+        }
+    }
+
+    /// Size of parts the object will be downloaded in, in bytes.
+    /// The minimum part size is 5 MiB and any value given less than that will be rounded up.
+    /// Defaults is 8 MiB.
+    pub fn target_part_size(mut self, size_bytes: u64) -> Self {
```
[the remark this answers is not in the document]
aajtodd: Good point. Any preferences here or leave at 5MiB min?

aws/hll/aws-s3-transfer-manager/src/download/body.rs
aajtodd: Exactly, this is commonly used for iteration.

aws/hll/aws-s3-transfer-manager/src/download/body.rs
aajtodd: > we eventually error out at [a later point](https://github.com/smithy-lang/smithy-rs/pull/3697/files#diff-1bd049795ee294cf2e26a25903e7eef0ac42b813f0afd68652244f75bcaf2e57R140), in which case unwrap will panic

Right, this is just in the example driver program though.

I think this TODO comment was made thinking about pause/resume as well as if we are downloading a massive file and a few chunks fail including retries. Having the flexibility to potentially keep going vs what we have now where it would be very difficult to retry for just that missing chunk. We'll see how it plays out.

aws/hll/aws-s3-transfer-manager/src/download.rs
aajtodd: I don't know that we know this yet but agreed we'll want to revisit all docs anyway (I am trying to document as we go though to the extent that makes sense).

aws/hll/aws-s3-transfer-manager/src/download/worker.rs
aajtodd: I'll add this as a TODO for now. I had something like this and can't recall what I didn't like or wasn't optimal about it off top of my head.

aws/hll/aws-s3-transfer-manager/src/download/body.rs
aajtodd: I was going to add more here but I expect this will ultimately be covered by tests involving checksums of data. I can add a simple test though for sanity sake.

The assertion has the chunk numbers in it.

sbiscigl (approved):

aws/hll/aws-s3-transfer-manager/src/download.rs
```diff
@@ -34,17 +49,183 @@ impl From<GetObjectInputBuilder> for DownloadRequest {
...
-#[derive(Debug)]
-#[non_exhaustive]
-pub struct DownloadResponse {
-    /// Object metadata
-    pub object_meta: ObjectMetadata,
+/// Fluent style builder for [Downloader]
+#[derive(Debug, Clone)]
+pub struct Builder {
+    target_part_size_bytes: u64,
+    checksum_validation_enabled: bool,
```
graebm, line 56: aws-c-s3 also started with options like this on the "Client", but when integrating with Java and Python SDKs we ended up needing to make them configurable per "meta-request" as well.

I know this PR is already merged. I just remembered...

PR #3712, opened 2024-06-20
Unpin pinned dependencies

landonxjames:
## Motivation and Context

Due to numerous dependency issues over the past few weeks several of our dependencies were pinned to specific minor versions. Now that we have lockfiles for all of our workspaces those pinned versions can be undone.

## Description

I unpinned those dependencies and generated new lockfiles after that change. 

## Testing

I ran the gradle `assemble` and `sdkTest` tasks locally and they both passed. I ran a full set of preview/release e2e tests against this change in my local stack and those all passed. And since we copy the Cargo.lock into the test workspaces the CI run for this PR will also test the changes.

## Checklist

No functional changes, only dependency changes.

----

### DIFF
--- a/aws/sdk/Cargo.lock
+++ b/aws/sdk/Cargo.lock
@@ -329,7 +329,7 @@ checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
 
 [[package]]
 name = "aws-config"
-version = "1.5.2"
+version = "1.5.3"
 dependencies = [
  "aws-credential-types 1.2.0",
  "aws-runtime 1.3.0",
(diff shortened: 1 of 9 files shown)


aws/sdk/Cargo.lock
```diff
@@ -443,6 +443,30 @@ dependencies = [
  "paste",
 ]
 
+[[package]]
+name = "aws-runtime"
+version = "1.2.2"
```
ysaito1001, line 448: Do we know why `aws-runtime` with its source crates.io-index has been downgraded to `1.2.2` from `1.3.0`?

landonxjames: Good catch! I think (although crates.io doesn't seem to preserve timestamps so hard to say for sure) that when I actually generated these lockfiles that 1.3.0 might not have been published yet since it is only 3 days old. Took me a couple days to get through all the e2e-testing with the unpinned versions. I regenerated and that bit disappeared so should be good after another CI run.

SimenB:
Any particular reason why this hasn't been published? 👀 

E.g. latest https://crates.io/crates/aws-smithy-runtime is still 1.6.0

PR #3726, opened 2024-06-28
Improve broken protocol test generation

david-perez:
We currently "hotfix" a broken protocol test in-memory, but there's no
mechanism that alerts us when the broken protocol test has been fixed
upstream when updating our Smithy version. This commit introduces such a
mechanism by generating both the original and the fixed test, with a
`#[should_panic]` attribute on the former, so that the test fails when
all its assertions succeed.

With this change, in general this approach of fixing tests in-memory
should now be used over adding the broken test to `expectFail` and
adding the fixed test to a `<protocol>-extras.smithy` Smithy model,
which is substantially more effort.

----


ysaito1001 (approved):
Great mechanism to add! [This](https://github.com/smithy-lang/smithy/pull/2341) will be a broken test for a client response test in Smithy 1.50, but it'll probably be fixed in the next Smithy release. Do you still recommend adding that test to `brokenTests` in `ClientProtoclTestGenerator` or is that overkill?

drganjoo (approved):

david-perez:
> https://github.com/smithy-lang/smithy/pull/2341 will be a broken test for a client response test in Smithy 1.50, but it'll probably be fixed in the next Smithy release. Do you still recommend adding that test to brokenTests in ClientProtoclTestGenerator or is that overkill?

@ysaito1001 Not overkill, that's precisely what `brokenTests` is for.

Issue #3339, opened 2023-12-21
Server SDKs should serialize unset values for shapes that have default values


david-perez:
This is a SHOULD according to [the spec](https://smithy.io/2.0/spec/type-refinement-traits.html#default-value-serialization):

> All effective default values SHOULD be serialized. This ensures that messages are unambiguous and do not change during deserialization if the default value for a member changes after the message was serialized.

So, to expand on the problem, consider this order of events:

1. Server responds without setting the default value A for a shape.
2. Response is stored somewhere.
3. Model update is rolled out to servers and clients; default value changes to B.
4. Client deserializes response from 2, sees no value for shape, so it defaults to B.

---

If server user code uses builders, default values are always set and are hence serialized in the response. However, user code might construct the struct directly, and omit values for shapes that have `@default`, in which case the default value is _not_ serialized.

As @rcoh suggested, a good fix for this would be to code-generate a pass in the serializer that recursively set default values prior to performing serialization.

PR #3728, opened 2024-06-28
Upgrade Smithy to 1.50.0

ysaito1001:
## Motivation and Context
This PR upgrades Smithy to 1.50.0. The majority of the changes follow `TODO` added in https://github.com/smithy-lang/smithy-rs/pull/3690. Other than that, a few adjustments needed to be made:
- for the client
  - added two failing tests `RestJsonClientPopulatesDefaultValuesInInput` and `RestJsonClientUsesExplicitlyProvidedMemberValuesOverDefaults` to known failing tests for the same reason [here](https://github.com/smithy-lang/smithy-rs/blob/main/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ClientProtocolTestGenerator.kt#L72)
  - added one broken test (i.e. the upstream test definition is incorrect but our implementation is correct) to known broken tests per ([smithy#2341](https://github.com/smithy-lang/smithy/pull/2341), [smithy-rs#3726](https://github.com/smithy-lang/smithy-rs/pull/3726#issuecomment-2199833659))
- for the server
  - removed `rest-xml-extras.smithy` since `RestXmlMustSupportParametersInContentType` is now available upstream Smithy 1.50.0
  - added the following to known failing tests (since the `awsJson1_0` counterparts are already in the list, but we need the server team to verify this assumption & provide additional `TODO` comments if necessary)
    - `RestJsonServerPopulatesDefaultsWhenMissingInRequestBody`
    - `RestJsonServerPopulatesDefaultsInResponseWhenMissingInParams`,
    - `RestJsonServerPopulatesNestedDefaultValuesWhenMissingInInResponseParams` 

## Testing
Existing tests in CI

----


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -141,12 +138,15 @@ class ServerProtocolTestGenerator(
                 // These tests are broken because they are missing a target header.
                 FailingTest.RequestTest(AWS_JSON_10, "AwsJson10ServerPopulatesNestedDefaultsWhenMissingInRequestBody"),
                 FailingTest.RequestTest(AWS_JSON_10, "AwsJson10ServerPopulatesDefaultsWhenMissingInRequestBody"),
+                FailingTest.RequestTest(REST_JSON, "RestJsonServerPopulatesDefaultsWhenMissingInRequestBody"),
```
david-perez, line 141: `RestJsonServerPopulatesDefaultsWhenMissingInRequestBody` fails because `defaultNullDocument` has a `null` default, so the `hasNonNullDefault` check here:

https://github.com/smithy-lang/smithy-rs/blob/9af72f5f2a04e941dd7f0f17a292e4396efdfda6/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt#L537

is `false` and so [`aws_smithy_types::Document::Null`](https://docs.rs/aws-smithy-types/latest/aws_smithy_types/enum.Document.html#variant.Null) is not used by the instantiator.

But even if we improved the check and special-cased `document` member shapes, there's an additional problem: the same check is used by Smithy itself [here](https://github.com/smithy-lang/smithy/blob/efcb39cf19f4cc71be65b64d2337802d628b8571/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java#L142) when determining whether the structure member should be optional or not. Since the node for the `@default` value of the member shape is `null`, Smithy says the structure member should be optional. But the generated code for default value setting expects the member to be non-`Option`al, since there's a default value.

In general, this begs the question of whether a required (non-`Option`al) `document` shape makes sense at all. The user can always set it to [`aws_smithy_types::Document::Null`](https://docs.rs/aws-smithy-types/latest/aws_smithy_types/enum.Document.html#variant.Null) and convey the same semantics as if shape were nullable and `None` had been set for it. To belabor the point, consider this model:

```
structure OperationInput {
    @required
    document: Document
}
```

This gets generated in Rust as:

```
struct OperationInput {
    document: aws_smithy_types::Document
}
```

The user can instantiate this as:

```
let op_input = OperationInput::builder()
    .document(aws_smithy_types::Document::Null)
    .build();
```

Which will get serialized on the wire as `{ document: null }` or `{ }`, depending on whether the protocol serializes nulls or not, respectively. Either way, we're allowing the user to not provide a value for a required member shape, violating the model.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
```diff
@@ -141,12 +138,15 @@ class ServerProtocolTestGenerator(
...
                 // Response defaults are not set when builders are not used https://github.com/smithy-lang/smithy-rs/issues/3339
                 FailingTest.ResponseTest(AWS_JSON_10, "AwsJson10ServerPopulatesDefaultsInResponseWhenMissingInParams"),
                 FailingTest.ResponseTest(
                     AWS_JSON_10,
                     "AwsJson10ServerPopulatesNestedDefaultValuesWhenMissingInInResponseParams",
                 ),
+                FailingTest.ResponseTest(REST_JSON, "RestJsonServerPopulatesDefaultsInResponseWhenMissingInParams"),
+                FailingTest.ResponseTest(REST_JSON, "RestJsonServerPopulatesNestedDefaultValuesWhenMissingInInResponseParams"),
```
david-perez, line 149: `RestJsonServerPopulatesDefaultsInResponseWhenMissingInParams` and `RestJsonServerPopulatesNestedDefaultValuesWhenMissingInInResponseParams` fail because of https://github.com/smithy-lang/smithy-rs/issues/3339. The protocol test generator does not use the builder to instantiate the struct, but instead uses direct struct instantiation with an expression, hence the `@default` values don't get filled out.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
ysaito1001: Thanks for the detailed explanation! Moved the test to the bottom of the list and added a link to this discussion for more context.

Velfi (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
david-perez: Can you add a TODO above these two linking to #3339?

david-perez (approved):

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt
ysaito1001: Actually, a TODO with 3339 is already at line 141. I'll update that TODO to explicitly state that the TODO applies to the following 4 tests.

PR #3734, opened 2024-07-02
add detailed error explanations page

Velfi:
I was speaking to several customers that found our connection poisoning terminology confusing. To address this, I've changed the messaging and I've also created a new section in the docs that explains some errors in greater detail. To direct users to these explanations, I've added a link to the `tracing::info!` call.

I propose that we consider doing the same any time we identify error messaging that
- must be kept succinct to avoid logging too much info
- must be more understandable to our users.

----

### DIFF
--- a/design/src/client/detailed_error_explanations.md
+++ b/design/src/client/detailed_error_explanations.md
@@ -1,30 +1,35 @@
 # Detailed Error Explanations
 
-This page collects detailed explanations for some errors. If you encounter an error
-and are interested in learning more about what it means and why it occurs, check here.
+This page collects detailed explanations for some errors. If you encounter an
+error and are interested in learning more about what it means and why it occurs,
+check here.
 
-**If you can't find the explanation on this page, please file an issue asking for it to be added.**
+**If you can't find the explanation on this page, please file an issue asking
+for it to be added.**
 
 ## "Connection encountered an issue and should not be re-used. Marking it for closure"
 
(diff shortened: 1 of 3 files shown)


design/src/client/detailed_error_explanations.md
```diff
@@ -0,0 +1,31 @@
+# Detailed Error Explanations
+
+This page collects detailed explanations for some errors. If you encounter an error
+and are interested in learning more about what it means and why it occurs, check here.
+
+**If you can't find the explanation on this page, please file an issue asking for it to be added.**
+
+## "Connection encountered an issue and should not be re-used. Marking it for closure"
+
+The SDK clients each maintain their own connection pool. By the convention of some services,
```
aajtodd, line 10: This isn't strictly true, in fact most of the time connection pools are shared across clients. Depends on how the client is constructed though.

design/src/client/detailed_error_explanations.md
```diff
@@ -0,0 +1,31 @@
...
+when a request fails due to a [transient error](#transient-errors), that connection should not be re-used for a
+retry. Instead, it should be dropped and a new connection created instead. This prevents
+clients from repeatedly sending requests to a service endpoint that's currently overloaded
```
aajtodd, line 13: prevents clients from repeatedly sending requests over a failed connection?

landonxjames (approved):

PR #3742, opened 2024-07-07
Support `stringArray` type in endpoints params

landonxjames:
## Description

Updated our endpoint rule generation to support the new `stringArray` parameter type

## Testing

Updated the existing `EndpointsDecoratorTest` with tests for `stringArray` 

## Checklist

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
@@ -5,6 +5,7 @@
 
 package software.amazon.smithy.rust.codegen.client.smithy.endpoint.generators
 
+import software.amazon.smithy.model.node.ArrayNode
 import software.amazon.smithy.model.node.BooleanNode
 import software.amazon.smithy.model.node.Node
 import software.amazon.smithy.model.node.StringNode
@@ -156,6 +157,11 @@ class EndpointParamsInterceptorGenerator(
             when (node) {
                 is StringNode -> rust("Some(${node.value.dq()}.to_string())")
                 is BooleanNode -> rust("Some(${node.value})")
+                is ArrayNode -> {
+                    val elms = node.elements.map { "\"$it\".to_string()" }.joinToString(",")
+                    rust("Some(vec![$elms])")
+                }
+
                 else -> PANIC("unsupported default value: $node")
             }
         }
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
@@ -158,7 +158,8 @@ class EndpointParamsInterceptorGenerator(
                 is StringNode -> rust("Some(${node.value.dq()}.to_string())")
                 is BooleanNode -> rust("Some(${node.value})")
                 is ArrayNode -> {
-                    val elms = node.elements.map { "\"$it\".to_string()" }.joinToString(",")
+                    // Cast the elements to a StringNode so this will fail if non-string values are provided
+                    val elms = node.elements.map { "\"${(it as StringNode).value}\".to_string()" }.joinToString(",")
                     rust("Some(vec![$elms])")
                 }
 
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
@@ -159,7 +159,7 @@ class EndpointParamsInterceptorGenerator(
                 is BooleanNode -> rust("Some(${node.value})")
                 is ArrayNode -> {
                     // Cast the elements to a StringNode so this will fail if non-string values are provided
-                    val elms = node.elements.map { "\"${(it as StringNode).value}\".to_string()" }.joinToString(",")
+                    val elms = node.elements.map { "${(it as StringNode).value.dq()}.to_string()" }.joinToString(",")
                     rust("Some(vec![$elms])")
                 }
 
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/ClientCodegenIntegrationTest.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/ClientCodegenIntegrationTest.kt
@@ -12,6 +12,7 @@ import software.amazon.smithy.rust.codegen.client.smithy.ClientCodegenContext
 import software.amazon.smithy.rust.codegen.client.smithy.RustClientCodegenPlugin
 import software.amazon.smithy.rust.codegen.client.smithy.customize.ClientCodegenDecorator
 import software.amazon.smithy.rust.codegen.core.smithy.RustCrate
+import software.amazon.smithy.rust.codegen.core.smithy.generators.protocol.ProtocolTestGenerator
 import software.amazon.smithy.rust.codegen.core.testutil.IntegrationTestParams
 import software.amazon.smithy.rust.codegen.core.testutil.codegenIntegrationTest
 import java.nio.file.Path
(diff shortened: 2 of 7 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -156,6 +157,11 @@ class EndpointParamsInterceptorGenerator(
             when (node) {
                 is StringNode -> rust("Some(${node.value.dq()}.to_string())")
                 is BooleanNode -> rust("Some(${node.value})")
+                is ArrayNode -> {
+                    val elms = node.elements.map { "\"$it\".to_string()" }.joinToString(",")
```
Velfi, line 161: The two above use `node.value` instead. Is that load bearing? It feels odd that this one would just be `$it`

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/ClientCodegenIntegrationTest.kt
```diff
@@ -37,6 +38,13 @@ fun clientIntegrationTest(
                 ) {
                     test(codegenContext, rustCrate)
                 }
+
+                override fun protocolTestGenerator(
+                    codegenContext: ClientCodegenContext,
+                    baseGenerator: ProtocolTestGenerator,
+                ): ProtocolTestGenerator {
+                    return super.protocolTestGenerator(codegenContext, baseGenerator)
+                }
```
Velfi, line 47: This override can be deleted, right?

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecoratorTest.kt
```diff
@@ -184,6 +188,12 @@ class EndpointsDecoratorTest {
                                             .a_bool_param(false)
                                             .a_string_param("hello".to_string())
                                             .region("us-east-2".to_string())
+                                            .a_string_array_param(
+                                                vec!["a", "b", "c"]
+                                                    .iter()
+                                                    .map(|i| i.to_string())
+                                                    .collect::<Vec<String>>()
```
Velfi, line 195: you may even be able to do the slightly more idiomatic:
```suggestion
                                                    .map(ToString::to_string)
                                                    .collect::<Vec<_>>()
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/ClientCodegenIntegrationTest.kt
landonxjames: Without it I was getting failures trying to run the Endpoints test locally, not entirely sure why it was working before or what about my changes would have necessitated adding this override though.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecoratorTest.kt
```diff
@@ -63,7 +63,8 @@ class EndpointsDecoratorTest {
                 "BuiltInWithDefault": { "required": true, "type": "String", "builtIn": "AWS::DefaultBuiltIn", "default": "some-default" },
                 "BoolBuiltInWithDefault": { "required": true, "type": "Boolean", "builtIn": "AWS::FooBar", "default": true },
                 "AStringParam": { "required": false, "type": "String" },
-                "ABoolParam": { "required": false, "type": "Boolean" }
+                "ABoolParam": { "required": false, "type": "Boolean" },
+                "AStringArrayParam": { "required": false, "type": "stringArray" }
```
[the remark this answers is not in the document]
landonxjames: You would think, but changing it produces the following error:
```
software.amazon.smithy.model.SourceException: Unexpected parameter type `StringArray`. Expected `string`, `boolean`, or `stringArray`. (test.smithy [42, 67])
when parsing endpoint ruleset
while parsing the parameter `AStringArrayParam`
  at test.smithy:42
while parsing the parameter type
Unexpected parameter type `StringArray`. Expected `string`, `boolean`, or `stringArray`.
```

It looks like lowercase `boolean` and `string` do work though so I could change those so they all match.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/ClientCodegenIntegrationTest.kt
landonxjames: Huh I must have been doing something weird locally at some point in my debug cycle because removing this doesn't cause the error to reoccur.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
landonxjames: I couldn't easily get the value because `$it` is just a generic `Node` type. I will figure out an idiomatic way to coerce it to `StringNode` so I can extract the value more explicitly.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecoratorTest.kt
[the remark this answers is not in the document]
landonxjames: Reference to Smithy code for this confirming the difference: https://github.com/smithy-lang/smithy/blob/2cd3579c7fad1322798ebb62dd9f47b13ca4beb3/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/parameters/ParameterType.java#L46-L60

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/testutil/ClientCodegenIntegrationTest.kt
landonxjames: Addressed in https://github.com/smithy-lang/smithy-rs/pull/3742/commits/d1cae4a89017a0bde9fb725329ef405329148ade

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
landonxjames: Added an (unsafe) cast in https://github.com/smithy-lang/smithy-rs/pull/3742/commits/d1cae4a89017a0bde9fb725329ef405329148ade which has the added benefit that this will fail if a non-string type is provided to the array. So we get a quick extra safety check out of it

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecoratorTest.kt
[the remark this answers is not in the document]
landonxjames: Made them all lowercase in https://github.com/smithy-lang/smithy-rs/pull/3742/commits/d1cae4a89017a0bde9fb725329ef405329148ade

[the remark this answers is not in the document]
landonxjames: And created a smithy issue about the discrepancy: https://github.com/smithy-lang/smithy/issues/2347

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -158,7 +158,8 @@ class EndpointParamsInterceptorGenerator(
                 is StringNode -> rust("Some(${node.value.dq()}.to_string())")
                 is BooleanNode -> rust("Some(${node.value})")
                 is ArrayNode -> {
-                    val elms = node.elements.map { "\"$it\".to_string()" }.joinToString(",")
+                    // Cast the elements to a StringNode so this will fail if non-string values are provided
+                    val elms = node.elements.map { "\"${(it as StringNode).value}\".to_string()" }.joinToString(",")
```
Velfi, line 162:
```suggestion
                    val elms = node.elements.map { "${(it as StringNode).value.dq()}.to_string()" }.joinToString(",")
```

PR #3744, opened 2024-07-08
Fix panic when merging a stalled stream log to an empty LogBuffer

Velfi:
## Motivation and Context

this fixes an issue submitted by an internal user.

## Description

If the first event pushed into the log underlying stalled stream protection, then the log manager would panic. This is because it was expecting to merge the new event in but there was nothing to merge it with.

To fix this, I updated the code to `push` the event when the log manager is empty, instead of merging it.

## Testing

ran existing tests and wrote a new test exercising the previous-panicking code.

## Checklist

----


landonxjames (approved):
LGTM!

PR #3753, opened 2024-07-11
Add initial implementation of a Serde Decorator

rcoh:
## Motivation and Context
Customers want to be able to use `serde` with Smithy models. For details, see the [RFC](https://github.com/smithy-lang/smithy-rs/blob/configurable-serde/design/src/rfcs/rfc0045_configurable_serde.md)

## Description

Implementation of `serde::Serialize` for smithy-rs code generators. This takes the approach of a trait, `SerializeConfigured` which _returns_ a type that implements `Serialize`. This allows customers to control serde behavior for their use case.

```rust
/// Trait that allows configuring serialization
/// **This trait should not be implemented directly!** Instead, `impl Serialize for ConfigurableSerdeRef<T>`**
pub trait SerializeConfigured {
    /// Return a `Serialize` implementation for this object that owns the object. This is what you want
    /// If you need to pass something that `impl`s serialize elsewhere.
    fn serialize_owned(self, settings: SerializationSettings) -> impl Serialize;

    /// Return a `Serialize` implementation for this object that borrows from the given object
    fn serialize_ref<'a>(&'a self, settings: &'a SerializationSettings) -> impl Serialize + 'a;
}
```

This can be used as follows:
```rust
serde_json::to_string(&my_struct.serialize_ref(&SerializationSettings::redact_sensitive_fields());
```

Currently, this codegen plugin is not used by anything. It can be enabled by bringing it in scope during code generation & adding the `@serde` trait to the model for the fields where serialization is desired. This will allow the SDK (if they choose to) roll it out only for specific types or services that have customer demand.

There are a number of follow on items required:
- [x] Ensure the generated impls work for `Server` codegen
- [ ] Generate `Deserialize` implementations
- [x] Test the implementation all Smithy protocol test models (ensure it compiles)

## Testing
Unit test testing serialization with a kitchen-sink model

## Checklist

----


codegen-serde/build.gradle.kts
```diff
@@ -0,0 +1,94 @@
...
+val smithyVersion: String by project
+
+dependencies {
+    implementation(project(":codegen-core"))
+    implementation(project(":codegen-client"))
+    implementation(project(":codegen-server"))
+}
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_11
```
Velfi, line 29: Aren't we on v17 now?

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/KotlinClientSerdeDecorator.kt
```diff
@@ -0,0 +1,654 @@
...
+
+val SerdeFeature = Feature("serde", false, listOf("dep:serde"))
+val Module =
+    RustModule.public(
+        "serde_impl",
+        additionalAttributes = listOf(Attribute.featureGate(SerdeFeature.name)),
+        documentationOverride = "Implementations of `serde` for model types",
+    )
+
+class KotlinClientSerdeDecorator : ClientCodegenDecorator {
```
Velfi, line 61: What is the significance of including `Kotlin` in the name? Does this imply that we'll have a Java version some day?

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/KotlinClientSerdeDecorator.kt
```diff
@@ -0,0 +1,654 @@
...
+            val wrapper =
+                when {
+                    deps != null -> null
+                    shape is MapShape -> serializeMap(shape)
+                    shape is CollectionShape -> serializeList(shape)
+                    // Need to figure out the best default here.
+                    else -> serializeWithTodo(shape)
+                }
+            if (wrapper != null && applyTo != null) {
+                rustTemplate("&#{wrapper}(#{applyTo})", "wrapper" to wrapper, "applyTo" to applyTo)
```
Velfi, line 128: I think you can also do this:
```suggestion
                rust("&#1T(#2T)", wrapper, applyTo)
```

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/KotlinClientSerdeDecorator.kt
```diff
@@ -0,0 +1,654 @@
...
+            } else {
+                deps?.toSymbol().also { addDependency(it) }
+                applyTo?.invoke(this)
+            }
+        }
+    }
+
+    private fun serializeWithTodo(shape: Shape): RuntimeType =
+        serializeWithWrapper(shape) { _ ->
+            // PANIC("cant serialize $shape")
```
Velfi, line 138: leftover?

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/KotlinClientSerdeDecorator.kt
```diff
@@ -0,0 +1,654 @@
...
+        }
+
+    /**
+     * Serialize a type that already implements `Serialize` directly via `value.serialize(serializer)`
+     */
+    private fun directSerde(shape: Shape): RuntimeType {
+        return RuntimeType.forInlineFun(codegenContext.symbolProvider.toSymbol(shape).rustType().toString(), Module) {
+            implSerializeConfigured(codegenContext.symbolProvider.toSymbol(shape)) {
+                val baseValue =
+                    writable { rust("self.value") }.letIf(shape.isStringShape) { it.plus(writable(".as_str()")) }
```
Velfi, line 191: This is super cool, I didn't realize we could conditionally extend writables like this!

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/KotlinClientSerdeDecorator.kt
```diff
@@ -0,0 +1,654 @@
...
+                    if (codegenContext.symbolProvider.toSymbol(member).isOptional()) {
+                        rustTemplate(
+                            "if let Some($field) = &inner.$fieldName { #{serializeField} }",
+                            "serializeField" to fieldSerialization,
+                        )
+                    } else {
+                        rustTemplate(
+                            "let $field = &inner.$fieldName; #{serializeField}",
+                            "serializeField" to fieldSerialization,
+                        )
```
Velfi, line 310:
```suggestion
                        rust("let $field = &inner.$fieldName; #T", fieldSerialization)
```

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/KotlinClientSerdeDecorator.kt
```diff
@@ -0,0 +1,654 @@
...
+                         #{Document}::Number(#{Number}::Float(value)) => value.serialize(serializer),
+                         #{Document}::Number(#{Number}::PosInt(value)) => {
+                             value.serialize(serializer)
+                         },
+                         #{Document}::Number(#{Number}::NegInt(value)) => {
+                             value.serialize(serializer)
+                         },
+                         #{Document}::Bool(b) => b.serialize(serializer),
+                         #{Document}::Null => serializer.serialize_none(),
+                     }
```
Velfi, line 398: Could this be implemented as part of a runtime crate so that we generate less code?

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/Traits.kt
```diff
@@ -0,0 +1,59 @@
...
+                    .withMember("content", content)
+            return builder.build()
+        }
+
+        class Provider : AbstractTrait.Provider(ID) {
+            override fun createTrait(
+                target: ShapeId,
+                value: Node,
+            ): Trait {
+                val serialize = value.expectObjectNode().getBooleanMember("serialize").map { it.value }.orElse(true)
```
Velfi, line 39: Can we just do the check once? It would shorten the code a bit.
```suggestion
                value: ObjectNode,
            ): Trait {
                val serialize = value.getBooleanMember("serialize").map { it.value }.orElse(true)
```

codegen-serde/build.gradle.kts
rcoh: oh dono, I copied this from codegen core, can change it to whatever.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/KotlinClientSerdeDecorator.kt
rcoh: yeah you can (this was actually originally the only way!). I don't necessarily think its clearer (but if it's clearer to you, that's certainly enough reason to change it.)

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/KotlinClientSerdeDecorator.kt
rcoh: No. `SerializeConfigured` is crate local (and I think it needs to be for coherence). I can experiment with making it _not_ crate local but I would probably start by defining it in the crate.

The total amount of generated code is only dozens of lines which is totally dwarfed by the SDK code though, so I don't think there is a huge impact.

rcoh: We _could_ put it in an inlinable if you prefer though.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/Traits.kt
rcoh: good idea, I refactored this entire file

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
```diff
@@ -0,0 +1,397 @@
...
+        val symbolProvider = codegenContext.symbolProvider
+        val unionSymbol = symbolProvider.toSymbol(shape)
+
+        return writable {
+            implSerializeConfigured(unionSymbol) {
+                rustTemplate(
+                    """
+                    """,
+                    *SupportStructures.codegenScope,
+                )
```
Velfi, line 266: Is this meant to write an empty line or is it a leftover thing?

Velfi (approved):
Good idea separating the implementation of serialization and deserialization.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
rcoh: leftover thanks

ysaito1001 (approved):
Brilliant piece of work!

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecorator.kt
```diff
@@ -0,0 +1,80 @@
...
+import software.amazon.smithy.rust.codegen.core.util.hasTrait
+import software.amazon.smithy.rust.codegen.server.smithy.ServerCodegenContext
+import software.amazon.smithy.rust.codegen.server.smithy.customize.ServerCodegenDecorator
+
+val SerdeFeature = Feature("serde", false, listOf("dep:serde"))
+val SerdeModule =
+    RustModule.public(
+        "serde",
+        additionalAttributes = listOf(Attribute.featureGate(SerdeFeature.name)),
+        documentationOverride = "Implementations of `serde` for model types. NOTE: These implementations are NOT used for wire serialization as part of a smithy protocol and WILL NOT match the wire format. They are provided for convenience only.",
```
david-perez, line 26:
```suggestion
        documentationOverride = "Implementations of `serde` for model types. NOTE: These implementations are NOT used for wire serialization as part of a Smithy protocol and WILL NOT match the wire format. They are provided for convenience only.",
```

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
```diff
@@ -0,0 +1,392 @@
...
+            }
+        }
+    }
+
+    private fun serializeMap(shape: MapShape): RuntimeType =
+        serializeWithWrapper(shape) { value ->
+            writable {
+                rustTemplate(
+                    """
+                    use #{serde}::ser::SerializeMap;
```
david-perez, line 95: nit: could use FQS instead of having these `use` statements: https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
```diff
@@ -0,0 +1,392 @@
...
+     */
+    private fun serializeMember(
+        shape: MemberShape,
+        memberRef: String,
+    ): Writable {
+        val target = codegenContext.model.expectShape(shape.target)
+        return writable {
+            serializerFn(target) {
+                rust("&$memberRef")
+            }.plus { rust(".serialize_ref(&self.settings)") }(this)
```
david-perez, line 206: Hmm Clippy should catch this, there's no need for `&` here. Which makes me think that we're not running `cargo clippy --all-features` in CI.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecorator.kt
```diff
@@ -0,0 +1,80 @@
...
+import software.amazon.smithy.rust.codegen.core.rustlang.Feature
+import software.amazon.smithy.rust.codegen.core.rustlang.RustModule
+import software.amazon.smithy.rust.codegen.core.smithy.CodegenContext
+import software.amazon.smithy.rust.codegen.core.smithy.RustCrate
+import software.amazon.smithy.rust.codegen.core.util.hasTrait
+import software.amazon.smithy.rust.codegen.server.smithy.ServerCodegenContext
+import software.amazon.smithy.rust.codegen.server.smithy.customize.ServerCodegenDecorator
+
+val SerdeFeature = Feature("serde", false, listOf("dep:serde"))
+val SerdeModule =
```
david-perez, line 22: Chucking everything into a single module is going to make the `src/serde.rs` a huge file if `@serde` targets the service shape.

We should mirror the structure we've got with `src/protocol_serde/`, one file per shape.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
```diff
@@ -0,0 +1,392 @@
...
+                    """,
+                    *SupportStructures.codegenScope,
+                    "value" to value,
+                    "member" to serializeMember(shape.member, "v"),
+                )
+            }
+        }
+
+    /**
+     * Serialize a type that already implements `Serialize` directly via `value.serialize(serializer)`
```
david-perez, line 129: Hmm, well, with a `as_str` call that I assume is there for enums.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
```diff
@@ -0,0 +1,392 @@
...
+        RuntimeType.forInlineFun("SerializeDocument", SerdeModule) {
+            implSerializeConfigured(codegenContext.symbolProvider.toSymbol(shape)) {
+                rustTemplate(
+                    """
+                    match self.value {
+                         #{Document}::String(v) => serializer.serialize_str(v),
+                         #{Document}::Object(v) => {
+                             use #{serde}::ser::SerializeMap;
+                             let mut map = serializer.serialize_map(Some(v.len()))?;
+                             for (k, v) in v.iter() {
```
david-perez, line 322: Are these `iter()` needed?

codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecoratorTest.kt
```diff
@@ -0,0 +1,302 @@
...
+        }
+        """.replace("\\s".toRegex(), "")
+
+    @Test
+    fun generateSerializersThatWorkClient() {
+        clientIntegrationTest(simpleModel, params = IntegrationTestParams(cargoCommand = "cargo test --all-features")) { ctx, crate ->
+            val codegenScope =
+                arrayOf(
+                    "crate" to RustType.Opaque(ctx.moduleUseName()),
+                    "serde_json" to CargoDependency("serde_json", CratesIo("1")).toDevDependency().toType(),
```
david-perez, line 235: Can we also test with `serde_cbor` for good measure? To also check that Smithy blobs get serialized as bytes and not base64-encoded.

david-perez (changes requested):
I like the solution explained in the RFC.

Main issue with current implementation is that it won't handle constrained shapes in the server, for which we generate Rust newtypes.

Something that would make us gain confidence in the implementation is if we integration tested against the `constraints.smithy` model (or any large model like the `restJson1` one), post transforming it to add `@serde` to the service shape.

Have a look at this test: https://github.com/smithy-lang/smithy-rs/blob/c63e79245430c98b2725627566b80a0fbf8e9619/codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/serialize/CborSerializerAndParserGeneratorSerdeRoundTripIntegrationTest.kt#L116

codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecoratorTest.kt
```diff
@@ -0,0 +1,302 @@
...
+            operations: [SayHello, SayGoodBye],
+            version: "1"
+        }
+        @optionalAuth
+        operation SayHello {
+            input: TestInput
+            errors: [ValidationException]
+        }
+        @serde
+        structure TestInput {
```
landonxjames, line 39: I think it would be worthwhile to have at least one recursive type in the test. I know it shows up in DDB's `AttributeValue` type so probably worth testing.

settings.gradle.kts
```diff
@@ -9,6 +9,7 @@ include(":codegen-core")
 include(":codegen-client")
 include(":codegen-client-test")
 include(":codegen-server")
+include(":codegen-serde")
```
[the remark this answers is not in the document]
rcoh: so that you can add a separate dependency on it. Adding the dependency is what brings the decorator into scope and causes it to run during codegen

design/src/rfcs/rfc0045_configurable_serde.md
```diff
@@ -0,0 +1,212 @@
...
+    pub redact_sensitive_fields: bool,
+}
+```
+
+We MAY add additional configuration options in the future, but will keep the default behavior matching current behavior. Future options include:
+- Serialize `null` when a field is unset (the current default is to skip serializing that field)
+- Serialize blobs via a list of numbers instead of via base64 encoding
+- Change the default format for datetimes (current `HttpDate`)
+
+No objects actually implement `SerializeConfigured`. Instead, the crate defines two private structs:
```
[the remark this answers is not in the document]
rcoh: They are actually private structs allowing us to evolve the internal representation in the future

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SupportStructures.kt
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+    private fun serializationSettings() =
+        RuntimeType.forInlineFun("SerializationSettings", supportModule) {
+            // TODO(serde): Consider removing `derive(Default)`
+            rustTemplate(
+                """
+                /// Settings for use when serializing structures
+                ##[non_exhaustive]
+                ##[derive(Copy, Clone, Debug, Default)]
```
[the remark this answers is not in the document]
rcoh: yeah that's a fair point—I figured this would probably be boolean / enum flags. I can remove

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
rcoh: no, removed.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecorator.kt
rcoh: done

codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecoratorTest.kt
rcoh: done

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecorator.kt
```diff
@@ -0,0 +1,80 @@
...
+import software.amazon.smithy.rust.codegen.core.rustlang.Attribute
+import software.amazon.smithy.rust.codegen.core.rustlang.Feature
+import software.amazon.smithy.rust.codegen.core.rustlang.RustModule
+import software.amazon.smithy.rust.codegen.core.smithy.CodegenContext
+import software.amazon.smithy.rust.codegen.core.smithy.RustCrate
+import software.amazon.smithy.rust.codegen.core.util.hasTrait
+import software.amazon.smithy.rust.codegen.server.smithy.ServerCodegenContext
+import software.amazon.smithy.rust.codegen.server.smithy.customize.ServerCodegenDecorator
+
+val SerdeFeature = Feature("serde", false, listOf("dep:serde"))
```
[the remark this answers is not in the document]
rcoh: not currently, but this would be an easy follow-on

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
rcoh: fixed clippy lints in the generated code

codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecoratorTest.kt
rcoh: done

design/src/rfcs/rfc0045_configurable_serde.md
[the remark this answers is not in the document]
david-perez: Their fields are `pub(crate)`. Defining them in a separate common runtime crate wouldn't preclude us from evolving them in the future if we add `#[doc(hidden)]`. We have many examples of `#[doc(hidden)]` items in our runtime crates to make our lives easier and not have to generate them in every SDK / make them an `inlineable`.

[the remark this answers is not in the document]
david-perez: And the `SerializeConfigured` trait is public, and is what the user interacts with. That could also live in a separate common runtime crate.

settings.gradle.kts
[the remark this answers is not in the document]
david-perez: Ok, internally we'll have to create a separate package that publishes only this JAR.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
```diff
@@ -0,0 +1,500 @@
...
+                is UnionShape -> RuntimeType.forInlineFun(name, serdeSubmodule(shape), serializeUnionImpl(shape))
+                is TimestampShape -> serializeDateTime(shape)
+                is BlobShape ->
+                    if (shape.hasTrait<StreamingTrait>()) {
+                        serializeByteStream(shape)
+                    } else {
+                        serializeBlob(shape)
+                    }
+
+                is StringShape, is NumberShape, is BooleanShape -> directSerde(shape)
```
david-perez, line 95: These shapes could be constrained and as such we can't just delegate to `directSerde`.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
```diff
@@ -0,0 +1,500 @@
...
+                    #{serializer}
+                    """,
+                    "Shape" to codegenContext.symbolProvider.toSymbol(shape),
+                    "serializer" to serialization,
+                )
+            }
+        return wrapperStruct
+    }
+
+    private fun Shape.unwrapConstraints(): Boolean =
```
david-perez, line 260: I think you'll also need to take into account whether the `publicConstrainedTypes` setting is disabled. If it is then the user is not exposed to the wrapper newtypes and it doesn't matter whether the shapes are constrained or not, they'll work with types generated as if the constraints hadn't been modeled.

david-perez (changes requested):
The PR still doesn't properly handle constrained shapes. Make sure to add an integration test to `SerdeProtocolTestTest` against the `constraints.smithy` model, which is very thorough, and run it with both `publicConstrainedTypes` set to `true` (the default) and to `false`, like here, so we can be confident.

https://github.com/smithy-lang/smithy-rs/blob/c63e79245430c98b2725627566b80a0fbf8e9619/codegen-server-test/build.gradle.kts#L53-L68

You can read the model as a `String` without having to add a dependency on `codegen-core`:

```kotlin
val filePath = "../codegen-core/common-test-models/constraints.smithy"
var model = File(filePath).readText().asSmithyModel()
```

design/src/rfcs/rfc0045_configurable_serde.md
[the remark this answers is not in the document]
rcoh: you can't have them both be centralized/public because you'll run into coherence issues, I believe. But we could have one of them be public.

settings.gradle.kts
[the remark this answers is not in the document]
rcoh: yeah true. Another option is opting into it via a codegen setting I guess? That may be easier.

codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeDecoratorTest.kt
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+        operation SayGoodBye {
+            input: NotSerde
+        }
+        structure NotSerde {}
+        """.asSmithyModel(smithyVersion = "2")
+
+    @Test
+    fun generateSerializersThatWorkServer() {
```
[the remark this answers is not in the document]
rcoh: they are subtly different because the servers and clients have different module layouts. I had tried briefly but I don't think its worth it.

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
rcoh: done

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
rcoh: yeah correct. I've cleaned this up.

codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeProtocolTestTest.kt
```diff
@@ -0,0 +1,37 @@
...
+            model.expectShape(serviceShapeId, ServiceShape::class.java).toBuilder().addTrait(
+                SerdeTrait(true, false, null, null, SourceLocation.NONE),
+            ).build()
+        model =
+            ModelTransformer.create().mapShapes(model) { serviceShape ->
+                serviceShape.letIf(serviceShape.id == serviceShapeId) {
+                    service
+                }
+            }
+        clientIntegrationTest(model, IntegrationTestParams(service = ServiceShapeId.REST_JSON, cargoCommand = "cargo test --all-features")) { clientCodegenContext, rustCrate ->
```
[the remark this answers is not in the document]
rcoh: yes, done.

design/src/rfcs/rfc0045_configurable_serde.md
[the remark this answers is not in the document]
rcoh: I think using `#[doc(hidden)]` isn't actually a great idea—you still need to follow semver since the generated code can rely on it so you can't actually use it to make breaking changes.

david-perez (approved):

codegen-serde/src/main/kotlin/software/amazon/smithy/rust/codegen/serde/SerializeImplGenerator.kt
```diff
@@ -0,0 +1,550 @@
...
+
+                        false -> rust("map.serialize_entry(k.as_str(), &#T)?;", member)
+                    }
+                }
+            writable {
+                rustTemplate(
+                    """
+                    use #{serde}::ser::SerializeMap;
+                    let mut map = serializer.serialize_map(Some(#{value}.len()))?;
+                    for (k, v) in #{value}.iter() {
```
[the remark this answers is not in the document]
rcoh: these are load bearing

PR #3755, opened 2024-07-11
Add support for `operationContextParams` Endpoints trait

landonxjames:
## Motivation and Context

We have to support the new [`operationContextParams` trait](https://smithy.io/2.0/additional-specs/rules-engine/parameters.html#smithy-rules-operationcontextparams-trait) for endpoint resolution. This trait specifies JMESPath expressions for selecting parameter data from the operation's input type.

## Description

* Add codegen support for the [JMESPath `keys`](https://jmespath.org/specification.html#keys) function (required by the trait [spec](https://smithy.io/2.0/additional-specs/rules-engine/parameters.html#smithy-rules-operationcontextparams-trait))
* Add codegen support for the trait itself. This is achieved by generating `get_param_name` functions for each param specified in `operationContextParams`. These functions pull the data out of the input object and it is added to the endpoint params in the `${operationName}EndpointParamsInterceptor`

## Testing

Updated the existing test suite for JMESPath codegen to test the `keys` function. Updated the existing EndpointsDecoratorTest with an `operationContextParams` trait specifying one param of each supported type to test the codegen.

## Checklist

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustJmespathShapeTraversalGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustJmespathShapeTraversalGenerator.kt
@@ -446,11 +446,12 @@ class RustJmespathShapeTraversalGenerator(
                     output =
                         writable {
                             arg.output(this)
-                            val out = arg.outputShape.shape
-                            when (out) {
+                            val outputShape = arg.outputShape.shape
+                            when (outputShape) {
                                 is StructureShape -> {
                                     // Can't iterate a struct in Rust so source the keys from smithy
-                                    val keys = out.allMembers.keys.map { "${it.dq()}.to_string()" }.joinToString(",")
+                                    val keys =
+                                        outputShape.allMembers.keys.joinToString(",") { "${it.dq()}.to_string()" }
                                     rust("let $ident = vec![$keys];")
                                 }
 
@@ -459,7 +460,7 @@ class RustJmespathShapeTraversalGenerator(
                                 }
 
                                 else ->
-                                    throw UnsupportedJmesPathException("The shape type for an input to the keys function must be a struct or a map, got ${out?.type}")
+                                    throw UnsupportedJmesPathException("The shape type for an input to the keys function must be a struct or a map, got ${outputShape?.type}")
                             }
                         },
                 )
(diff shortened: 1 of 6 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustJmespathShapeTraversalGenerator.kt
```diff
@@ -429,6 +431,40 @@ class RustJmespathShapeTraversalGenerator(
...
+                }
+                GeneratedExpression(
+                    identifier = ident,
+                    outputType = RustType.Vec(RustType.String),
+                    outputShape = TraversedShape.Array(null, TraversedShape.String(null)),
+                    output =
+                        writable {
+                            arg.output(this)
+                            val out = arg.outputShape.shape
+                            when (out) {
```
Velfi, line 450: `out` is a special word in Kotlin per https://kotlinlang.org/docs/keyword-reference.html#modifier-keywords
We should change it to `output` or similar.

Velfi (approved):

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecoratorTest.kt
```diff
@@ -194,6 +216,8 @@ class EndpointsDecoratorTest {
                                                     .map(ToString::to_string)
                                                     .collect::<Vec<_>>()
                                             )
+                                            .jmes_path_param_string_array(vec!["key2".to_string(), "key1".to_string()])
```
aajtodd, line 219: **question**: are these `jmes_path_param_*` functions internal or public?

aajtodd (approved):

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecoratorTest.kt
ysaito1001: Probably `pub` since they're methods on service-specific endpoint params builder.

landonxjames: Correct, they are `pub` since they are part of the param builder. Normally they won't have the `jmes_path_param_` prefix, thats just because I named them like that in the model:

```smithy
@operationContextParams(
            JmesPathParamString: {
                path: "nested.field",
            }
            JmesPathParamBoolean: {
                path: "nested.boolField",
            }
            JmesPathParamStringArray: {
                path: "keys(nested.mapField)",
            }
        )
```

We could make them private with the assumption that these should only be set by the generated JMESPath getters, but we allow overriding other values set in the model (mostly those on `staticContextParams`) so it seemed most consistent to keep that behavior.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/waiters/RustJmespathShapeTraversalGenerator.kt
landonxjames: Fixed in https://github.com/smithy-lang/smithy-rs/pull/3755/commits/f964801602a10eb7de34244b28544b32fc35ffaf

PR #3756, opened 2024-07-12
add io related abstractions to support s3 tm uploads

aajtodd:
## Description

Adds a new `io` module to S3 transfer manager HLL. This is supporting functionality for implementing uploads.

High level description of new components:

* **InputStream** - this is a new abstraction to replace `ByteStream` for use in transfer manager uploads. The primary driver behind this is once we create a `ByteStream` we have very little control over optimizing how it can be consumed. We lose whether it came from a file or in-memory or elsewhere. Internally this will of course eventually map to a `ByteStream` when we call `UploadPart` or `PutObject` but we want to have control over how bytes are read from the stream (e.g. being able to read a file concurrently). 
    * In the future we may add additional variants to support e.g. unknown streams (e.g. `impl io::Read`) or giving users control over checksums for each part, etc. This is an area of exploration still.
* **PartReader / ReadPart** - internal abstractions over `InputStream` that workers will eventually consume. The idea is we know what `InputStream` is and we convert it to something that reads "parts" for multipart upload in the most efficient way possible for whatever the underlying stream type is. 
    * In a subsequent PR workers share a single part reader and read from it concurrently. Workers don't have to care about anything other than asking for the next "part" of data and uploading it.
* **PathBodyBuilder / PathBody** - replacements for `FsBuilder` from smithy runtime / SDK. Notable difference is we don't currently allow specifying either an already open file or control over buffer sizes. 
    * We can of course support an open file but the implementation will require additional locking to ensure bytes are read sequentially

----

### DIFF
--- a/aws/hll/aws-s3-transfer-manager/src/io/error.rs
+++ b/aws/hll/aws-s3-transfer-manager/src/io/error.rs
@@ -13,7 +13,7 @@ pub(crate) enum ErrorKind {
     UpperBoundSizeHintRequired,
     OffsetGreaterThanFileSize,
     TaskFailed(JoinError),
-    IOError(StdIoError),
+    IoError(StdIoError),
 }
 
 /// An I/O related error occurred
@@ -35,7 +35,7 @@ impl From<ErrorKind> for Error {
 
 impl From<StdIoError> for Error {
     fn from(err: StdIoError) -> Self {
-        ErrorKind::IOError(err).into()
+        ErrorKind::IoError(err).into()
     }
 }
 
@@ -50,8 +50,8 @@ impl fmt::Display for Error {
                 f,
                 "offset must be less than or equal to file size but was greater than"
             ),
-            ErrorKind::IOError(_) => write!(f, "I/O error"),
-            ErrorKind::TaskFailed(_) => write!(f, "internal task failed to complete"),
+            ErrorKind::IoError(_) => write!(f, "I/O error"),
+            ErrorKind::TaskFailed(_) => write!(f, "task failed"),
         }
     }
 }
@@ -61,7 +61,7 @@ impl StdError for Error {
         match &self.kind {
             ErrorKind::UpperBoundSizeHintRequired => None,
             ErrorKind::OffsetGreaterThanFileSize => None,
-            ErrorKind::IOError(err) => Some(err as _),
+            ErrorKind::IoError(err) => Some(err as _),
             ErrorKind::TaskFailed(err) => Some(err as _),
         }
     }
--- a/aws/hll/aws-s3-transfer-manager/src/io/error.rs
+++ b/aws/hll/aws-s3-transfer-manager/src/io/error.rs
@@ -1,0 +1,79 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+use std::error::Error as StdError;
+use std::fmt;
+use std::fmt::Formatter;
+use std::io::{Error as StdIoError, ErrorKind as StdIoErrorKind};
+use tokio::task::JoinError;
+
+#[derive(Debug)]
+pub(crate) enum ErrorKind {
+    UpperBoundSizeHintRequired,
+    OffsetGreaterThanFileSize,
+    TaskFailed(JoinError),
+    IOError(StdIoError),
+}
+
+/// An I/O related error occurred
+#[derive(Debug)]
+pub struct Error {
+    kind: ErrorKind,
+}
+
+impl Error {
+    pub(crate) fn upper_bound_size_hint_required() -> Error {
+        ErrorKind::UpperBoundSizeHintRequired.into()
+    }
+}
+impl From<ErrorKind> for Error {
+    fn from(kind: ErrorKind) -> Self {
+        Self { kind }
+    }
+}
+
+impl From<StdIoError> for Error {
+    fn from(err: StdIoError) -> Self {
+        ErrorKind::IOError(err).into()
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match &self.kind {
+            ErrorKind::UpperBoundSizeHintRequired => write!(
+                f,
+                "size hint upper bound (SizeHint::upper) is required but was None"
+            ),
+            ErrorKind::OffsetGreaterThanFileSize => write!(
+                f,
+                "offset must be less than or equal to file size but was greater than"
+            ),
+            ErrorKind::IOError(_) => write!(f, "I/O error"),
+            ErrorKind::TaskFailed(_) => write!(f, "internal task failed to complete"),
+        }
+    }
+}
+
+impl StdError for Error {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        match &self.kind {
+            ErrorKind::UpperBoundSizeHintRequired => None,
+            ErrorKind::OffsetGreaterThanFileSize => None,
+            ErrorKind::IOError(err) => Some(err as _),
+            ErrorKind::TaskFailed(err) => Some(err as _),
+        }
+    }
+}
+impl From<Error> for StdIoError {
+    fn from(err: Error) -> Self {
+        StdIoError::new(StdIoErrorKind::Other, err)
+    }
+}
+
+impl From<JoinError> for Error {
+    fn from(value: JoinError) -> Self {
+        ErrorKind::TaskFailed(value).into()
+    }
+}
--- a/aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
+++ b/aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
@@ -1,0 +1,336 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+use std::cmp;
+use std::ops::DerefMut;
+use std::sync::Mutex;
+
+use bytes::{Buf, Bytes, BytesMut};
+
+use crate::io::error::Error;
+use crate::io::path_body::PathBody;
+use crate::io::stream::RawInputStream;
+use crate::io::InputStream;
+use crate::MEBIBYTE;
+
+/// Builder for creating a `ReadPart` implementation.
+#[derive(Debug)]
+pub(crate) struct PartReaderBuilder {
+    stream: Option<RawInputStream>,
+    part_size: usize,
+}
+
+impl PartReaderBuilder {
+    pub(crate) fn new() -> Self {
+        Self {
+            stream: None,
+            part_size: 5 * MEBIBYTE as usize,
+        }
+    }
+
+    /// Set the input stream to read from.
+    pub(crate) fn stream(mut self, stream: InputStream) -> Self {
+        self.stream = Some(stream.inner);
+        self
+    }
+
+    /// Set the target part size that should be used when reading data.
+    ///
+    /// All parts except for possibly the last one should be of this size.
+    pub(crate) fn part_size(mut self, part_size: usize) -> Self {
+        self.part_size = part_size;
+        self
+    }
+
+    pub(crate) fn build(self) -> impl ReadPart {
+        let stream = self.stream.expect("input stream set");
+        match stream {
+            RawInputStream::Buf(buf) => {
+                PartReader::Bytes(BytesPartReader::new(buf, self.part_size))
+            }
+            RawInputStream::Fs(path_body) => {
+                PartReader::Fs(PathBodyPartReader::new(path_body, self.part_size))
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+enum PartReader {
+    Bytes(BytesPartReader),
+    Fs(PathBodyPartReader),
+}
+
+impl ReadPart for PartReader {
+    async fn next_part(&self) -> Result<Option<PartData>, Error> {
+        match self {
+            PartReader::Bytes(bytes) => bytes.next_part().await,
+            PartReader::Fs(path_body) => path_body.next_part().await,
+        }
+    }
+}
+
+/// Data for a single part
+pub(crate) struct PartData {
+    pub(crate) data: Bytes,
+    pub(crate) part_number: u64,
+}
+
+/// The `ReadPart` trait allows for reading data from an `InputStream` and packaging the raw
+/// data into `PartData` which carries additional metadata needed for uploading a part.
+pub(crate) trait ReadPart {
+    /// Request the next "part" of data.
+    ///
+    /// When there is no more data readers should return `Ok(None)`.
+    /// NOTE: Implementations are allowed to return data in any order and consumers are
+    /// expected to order data by the part number.
+    fn next_part(
+        &self,
+    ) -> impl std::future::Future<Output = Result<Option<PartData>, Error>> + Send;
+}
+
+#[derive(Debug)]
+struct PartReaderState {
+    // current start offset
+    offset: u64,
+    // current part number
+    part_number: u64,
+    // total number of bytes remaining to be read
+    remaining: u64,
+}
+
+impl PartReaderState {
+    /// Create a new `PartReaderState`
+    fn new(content_length: u64) -> Self {
+        Self {
+            offset: 0,
+            part_number: 1,
+            remaining: content_length,
+        }
+    }
+
+    /// Set the initial offset to start reading from
+    fn with_offset(self, offset: u64) -> Self {
+        Self { offset, ..self }
+    }
+}
+
+/// [ReadPart] implementation for in-memory input streams.
+#[derive(Debug)]
+struct BytesPartReader {
+    buf: Bytes,
+    part_size: usize,
+    state: Mutex<PartReaderState>,
+}
+
+impl BytesPartReader {
+    fn new(buf: Bytes, part_size: usize) -> Self {
+        let content_length = buf.remaining() as u64;
+        Self {
+            buf,
+            part_size,
+            state: Mutex::new(PartReaderState::new(content_length)),
+        }
+    }
+}
+
+impl ReadPart for BytesPartReader {
+    async fn next_part(&self) -> Result<Option<PartData>, Error> {
+        let mut state = self.state.lock().expect("lock valid");
+        if state.remaining == 0 {
+            return Ok(None);
+        }
+
+        let start = state.offset as usize;
+        let end = cmp::min(start + self.part_size, self.buf.len());
+        let data = self.buf.slice(start..end);
+        let part_number = state.part_number;
+        state.part_number += 1;
+        state.offset += data.len() as u64;
+        state.remaining -= data.len() as u64;
+        let part = PartData { data, part_number };
+        Ok(Some(part))
+    }
+}
+
+/// [ReadPart] implementation for path based input streams
+#[derive(Debug)]
+struct PathBodyPartReader {
+    body: PathBody,
+    part_size: usize,
+    state: Mutex<PartReaderState>,
+}
+
+impl PathBodyPartReader {
+    fn new(body: PathBody, part_size: usize) -> Self {
+        let offset = body.offset;
+        let content_length = body.length;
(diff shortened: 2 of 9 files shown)


aws/hll/aws-s3-transfer-manager/src/io/error.rs
```diff
@@ -0,0 +1,79 @@
...
+use std::fmt::Formatter;
+use std::io::{Error as StdIoError, ErrorKind as StdIoErrorKind};
+use tokio::task::JoinError;
+
+#[derive(Debug)]
+pub(crate) enum ErrorKind {
+    UpperBoundSizeHintRequired,
+    OffsetGreaterThanFileSize,
+    TaskFailed(JoinError),
+    IOError(StdIoError),
```
Velfi, line 16: even tho it look like `lo` I think we should use this casing per our convention:
```suggestion
    IoError(StdIoError),
```

aws/hll/aws-s3-transfer-manager/src/io/error.rs
```diff
@@ -0,0 +1,79 @@
...
+            ErrorKind::UpperBoundSizeHintRequired => write!(
+                f,
+                "size hint upper bound (SizeHint::upper) is required but was None"
+            ),
+            ErrorKind::OffsetGreaterThanFileSize => write!(
+                f,
+                "offset must be less than or equal to file size but was greater than"
+            ),
+            ErrorKind::IOError(_) => write!(f, "I/O error"),
+            ErrorKind::TaskFailed(_) => write!(f, "internal task failed to complete"),
```
Velfi, line 54: Will tasks have specific failure modes in the future? We should consider using language similar to the `IoError` ones, i.e. more generic messaging like:
```suggestion
            ErrorKind::TaskError(_) => write!(f, "task error"),
```

Velfi: Or maybe
```suggestion
            ErrorKind::JoinError(_) => write!(f, "join error"),
```

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+
+use crate::io::error::Error;
+use crate::io::path_body::PathBody;
+use crate::io::stream::RawInputStream;
+use crate::io::InputStream;
+use crate::MEBIBYTE;
+
+/// Builder for creating a `ReadPart` implementation.
+#[derive(Debug)]
+pub(crate) struct PartReaderBuilder {
```
Velfi, line 19: I have a preference for giving stuff like this a generic `Builder` name and then referring to it with the module.
```suggestion
pub(crate) struct Builder {
```
We'd then say `part_reader::Builder` to refer to it.

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+            RawInputStream::Fs(path_body) => {
+                PartReader::Fs(PathBodyPartReader::new(path_body, self.part_size))
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+enum PartReader {
+    Bytes(BytesPartReader),
```
Velfi, line 61: Does the `Bytes` enum member always refer to bytes in memory or does this also apply to byte streams that can't be "restarted"?

aajtodd: right now we only represent two input streams (1) in-memory (Bytes and all it's various conversions to it) and (2) path based. 

I expect we'll expand to at least support `dyn io::Read` or something for any abstract stream as well as some others Micheal mentioned to me around users wanting precise control over how memory is handled or checksums but for now I just implemented these two. I do wonder what a generic approach you and I were discussing might look like still but haven't dug into it.

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+            PartReader::Fs(path_body) => path_body.next_part().await,
+        }
+    }
+}
+
+/// Data for a single part
+pub(crate) struct PartData {
+    pub(crate) data: Bytes,
+    pub(crate) part_number: u64,
+}
```
Velfi, line 78: I would simplify this as:
```suggestion
/// Data for a single part
pub(crate) struct Part {
    pub(crate) data: Bytes,
    pub(crate) number: u64,
}
```
but maybe that's too far? One other thing to consider with this struct is the [memory layout](https://doc.rust-lang.org/nomicon/repr-rust.html). I think the `part_number` field should be first, otherwise there may end up being 32 bits of padding between `data` and `part_number` (although probably not b/c everything is 64bit now.)

aws/hll/aws-s3-transfer-manager/src/io/error.rs
aajtodd: I don't know given we are currently using `tokio` but we may want to be more runtime agnostic in the future. Hard to say what it should look like.

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+        Ok(Some(part))
+    }
+}
+
+/// [ReadPart] implementation for path based input streams
+#[derive(Debug)]
+struct PathBodyPartReader {
+    body: PathBody,
+    part_size: usize,
+    state: Mutex<PartReaderState>,
```
Velfi, line 162: For ease of reading struct defs, would you mind adding a comment saying what variety of Mutex it is? That way I don't have to scroll back to the `use` statements.
```suggestion
    state: Mutex<PartReaderState>, // std Mutex
```

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+            (offset, part_number, part_size)
+        };
+        let path = self.body.path.clone();
+        let handle = tokio::task::spawn_blocking(move || {
+            // TODO(aws-sdk-rust#1159) - replace allocation with memory pool
+            let mut dst = BytesMut::with_capacity(part_size as usize);
+            // we need to set the length so that the raw &[u8] slice has the correct
+            // size, we are guaranteed to read exactly part_size data from file on success
+            // FIXME(aws-sdk-rust#1159) - can we get rid of this use of unsafe?
+            unsafe { dst.set_len(dst.capacity()) }
```
Velfi, line 201: I feel like there should be a `min` in here somewhere
```suggestion
            unsafe { dst.set_len(dst.len().min(dst.capacity())) }
```
but IDK. As long as we test the three cases (>, ==, and <)

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+        &self,
+    ) -> impl std::future::Future<Output = Result<Option<PartData>, Error>> + Send;
+}
+
+#[derive(Debug)]
+struct PartReaderState {
+    // current start offset
+    offset: u64,
+    // current part number
+    part_number: u64,
```
Velfi, line 98: We should note that this is 1-indexed

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+            return Ok(None);
+        }
+
+        let start = state.offset as usize;
+        let end = cmp::min(start + self.part_size, self.buf.len());
+        let data = self.buf.slice(start..end);
+        let part_number = state.part_number;
+        state.part_number += 1;
+        state.offset += data.len() as u64;
+        state.remaining -= data.len() as u64;
```
Velfi, line 151: Should we be sticking these behind methods or is that too much?
```suggestion
        state.increment_part_number();
        state.increment_offset(data.len());
        state.decrement_remaining(data.len());
```

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
aajtodd: `dst.len()` is 0 after calling `with_capacity()`, that was the original issue here is that we are allocating a chunk and trying to fill the underlying slice but that slice is of length zero to start with because nothing has been written to `BytesMut` yet.

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
aajtodd: I'm still waiting to see how much of this we keep when we get to benchmarking and optimizations 😆

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,336 @@
...
+
+    async fn path_reader_test(limit: Option<usize>, offset: Option<usize>) {
+        let part_size = 5;
+        let mut tmp = NamedTempFile::new().unwrap();
+        let mut data = Bytes::from("a lep is a ball, a tay is a hammer, a flix is a comb");
+        tmp.write_all(data.chunk()).unwrap();
+
+        let mut builder = InputStream::read_from().path(tmp.path());
+        if let Some(limit) = limit {
+            data.truncate(limit);
```
Velfi, line 294: Why do we truncate here? To simulate reading from an offset?

Velfi: never mind, I think I understand now that I've seen the invocations of this fn.

aws/hll/aws-s3-transfer-manager/src/io/path_body.rs
```diff
@@ -0,0 +1,210 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use crate::io::error::{Error, ErrorKind};
+use crate::io::stream::InputStream;
+use crate::io::stream::RawInputStream;
```
Velfi, line 8:
```suggestion
use crate::io::stream::{ InputStream, RawInputStream };
```

aws/hll/aws-s3-transfer-manager/src/io/path_body.rs
```diff
@@ -0,0 +1,210 @@
...
+    ///
+    /// By pre-specifying the length, this API skips an additional call to retrieve the size from file-system metadata.
+    ///
+    /// When used in conjunction with [`offset`](PathBodyBuilder::offset), allows for reading a single "chunk" of a file.
+    ///
+    /// <div class="warning">
+    /// Setting the length manually will trigger no validation related to any offset provided or the actual size of
+    /// the file. This is an advanced setting mainly used to avoid an additional syscall if you know the
+    /// size of the file already.
+    /// </div>
```
Velfi, line 79: so fancy

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
aajtodd: yes it's for simulating reading up to a certain number of bytes, I'm basically abusing `Bytes` ability to (1) start reading from an offset, (2) read up to a certain amount, and (3) chunk by a certain size where all chunks except the last will be of that size (which is the exact behavior we want part readers to have).

aws/hll/aws-s3-transfer-manager/src/io/path_body.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /// Returns a [`InputStream`] from this builder.
+    pub fn build(self) -> Result<InputStream, Error> {
+        let path = self.path.expect("path set");
+        let offset = self.offset.unwrap_or_default();
+
+        let length = match self.length {
+            None => {
+                // TODO(aws-sdk-rust#1159, design) - evaluate if we want build() to be async and to use tokio for stat() call (bytestream FsBuilder::build() is async)
```
Velfi, line 92: Maybe we could just have two build methods? `build` and `build_blocking`?

aws/hll/aws-s3-transfer-manager/src/io/path_body.rs
aajtodd: I feel like I expect too much out of rustfmt 🤷‍♂️

aws/hll/aws-s3-transfer-manager/src/io/path_body.rs
aajtodd: the fanciest

aws/hll/aws-s3-transfer-manager/src/io/path_body.rs
```diff
@@ -0,0 +1,210 @@
...
+        let stream = PathBodyBuilder::new()
+            .path(tmp.path())
+            .length(64)
+            .build()
+            .unwrap();
+
+        let body = path_body(&stream);
+        assert_eq!(0, body.offset);
+        // we don't validate this
+        assert_eq!(64, body.length);
```
Velfi, line 160: Could you also note "why not?", I'm curious

aws/hll/aws-s3-transfer-manager/src/io/stream.rs
```diff
@@ -0,0 +1,117 @@
...
+use crate::io::path_body::PathBodyBuilder;
+use crate::io::size_hint::SizeHint;
+
+/// Source of binary data.
+///
+/// `InputStream` wraps a stream of data for ease of use.
+#[derive(Debug)]
+pub struct InputStream {
+    pub(super) inner: RawInputStream,
+}
```
Velfi, line 22: What does `Input` mean here? This implies the existence of an `OutputStream` too, right?

aws/hll/aws-s3-transfer-manager/src/io/path_body.rs
aajtodd: When we get to implementing an actual transfer manager and are traversing a directory I expect we'll already have this metadata available and we can avoid any additional syscall. We could make this `pub(crate)` potentially.

aws/hll/aws-s3-transfer-manager/src/io/stream.rs
```diff
@@ -0,0 +1,117 @@
...
+            inner: RawInputStream::Buf(value),
+        }
+    }
+}
+
+impl From<Vec<u8>> for InputStream {
+    fn from(value: Vec<u8>) -> Self {
+        Self::from(Bytes::from(value))
+    }
+}
```
Velfi, line 117: consider adding some of the other [`From` impls that `Bytes` supports](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#impl-From%3C%26%5Bu8%5D%3E-for-Bytes), namely `&'static [u8]`

aws/hll/aws-s3-transfer-manager/src/io/stream.rs
aajtodd: yeah it implies it but I don't think we'll implement it. Open to naming but essentially we don't want to use `ByteStream` as the input or output type for upload or download respectively (for different reasons). I'm not sure what we want to call them so open to bikeshedding.

`InputStream` is a common name for something you can read from which is what this is but I don't expect it to be re-used for both inputs and outputs the way `ByteStream` is.

Velfi (approved):

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,337 @@
...
+
+/// The `ReadPart` trait allows for reading data from an `InputStream` and packaging the raw
+/// data into `PartData` which carries additional metadata needed for uploading a part.
+pub(crate) trait ReadPart {
+    /// Request the next "part" of data.
+    ///
+    /// When there is no more data readers should return `Ok(None)`.
+    /// NOTE: Implementations are allowed to return data in any order and consumers are
+    /// expected to order data by the part number.
+    fn next_part(
```
graebm, line 89: Just need to warn, this API, where the uploader "pulls/reads" data from the user, worked great in aws-c-s3 until we hit the use case where a user starts tons of uploads, but doesn't actually control when the data will arrive for any particular file. We ended up adding an alternate "push/write" API to cover this use case. See the Issue description in this PR: https://github.com/awslabs/aws-c-s3/pull/418

An API that covered both cases would probably work something like:
- 1 async call where user requests permission to submit another part. The async call completes when the uploader is ready to accept more data (maybe it returns a buffer from a pool, for them to fill in?).
- another call to submit the filled-in buffer

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,337 @@
...
+        }
+    }
+}
+
+/// Data for a single part
+pub(crate) struct PartData {
+    // 1-indexed 
+    pub(crate) part_number: u64,
+    pub(crate) data: Bytes,
+}
```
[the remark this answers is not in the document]
aajtodd: Yeah this is internal facing, I think we when we get to checksums and existing use cases we know about we need to consider this and whether we do something like make this more public facing. We want to retain as much control as we can though so it's a tradeoff.

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,337 @@
...
+            PartReader::Fs(path_body) => path_body.next_part().await,
+        }
+    }
+}
+
+/// Data for a single part
+pub(crate) struct PartData {
+    // 1-indexed 
+    pub(crate) part_number: u64,
+    pub(crate) data: Bytes,
```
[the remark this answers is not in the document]
aajtodd: Yes we still need to investigate memory pooling options, there is a TODO comment around this already.

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
```diff
@@ -0,0 +1,337 @@
...
+/// data into `PartData` which carries additional metadata needed for uploading a part.
+pub(crate) trait ReadPart {
+    /// Request the next "part" of data.
+    ///
+    /// When there is no more data readers should return `Ok(None)`.
+    /// NOTE: Implementations are allowed to return data in any order and consumers are
+    /// expected to order data by the part number.
+    fn next_part(
+        &self,
+    ) -> impl std::future::Future<Output = Result<Option<PartData>, Error>> + Send;
```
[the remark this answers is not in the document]
aajtodd: All good points. This is an internal only API/abstraction for now so we could make changes here to support that or we may throw this entire idea away depending on what optimizations we need/want to pursue.

aws/hll/aws-s3-transfer-manager/src/io/part_reader.rs
aajtodd: Yeah we talked about this the other day briefly where you mentioned this use case. I think we need to flesh that out more and consider this design incomplete until we cover it.

Right now this is basically the view/abstraction that a single uploader task will see (be more clear in a future PR but you can see in-branch [here](https://github.com/smithy-lang/smithy-rs/blob/atodd/s3-mpu/aws/hll/aws-s3-transfer-manager/src/upload.rs#L232-L237) and [here](https://github.com/smithy-lang/smithy-rs/blob/atodd/s3-mpu/aws/hll/aws-s3-transfer-manager/src/upload.rs#L320-L325)). This is obviously simplistic as it doesn't cover users needing/wanting to control checksums nor the ability for users to control which uploads have data ready.

aws/hll/aws-s3-transfer-manager/src/io/size_hint.rs
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/// A body size hint
+#[derive(Debug, Clone, Default)]
+pub struct SizeHint {
+    lower: u64,
+    upper: Option<u64>,
```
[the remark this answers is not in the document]
graebm: If there's precedent for this and Rust users expect it, that's a good argument. It does have the advantage for MPU, where user can say "I don't know the size, but small enough that we shouldn't do MPU"

graebm (approved):

PR #3758, opened 2024-07-12
v2 Smoketest codegen

Velfi:
This PR adds codegen for service-defined smoketests.

----


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
```diff
@@ -0,0 +1,172 @@
...
+            val useMultiRegionAccessPoints = node.getBooleanMemberOrDefault("useMultiRegionAccessPoints", true)
+
+            region?.let { writer.rust(".region(crate::config::Region::new(${it.dq()}))") }
+//            sigv4aRegionSet?.let { writer.rust("._($it)") }
+//            useAccountIdRouting?.let { writer.rust("._($it)") }
+            useDualstack?.let { writer.rust(".use_dual_stack($it)") }
+            useFips?.let { writer.rust(".use_fips($it)") }
+            uri?.let { writer.rust(".endpoint_url($it)") }
+            useAccelerate?.let { writer.rust(".accelerate_($it)") }
+//            useGlobalEndpoint?.let { writer.rust(".use_global_endpoint_($it)")}
```
Velfi, line 133: afaik, we never use the global endpoint

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
```diff
@@ -0,0 +1,172 @@
...
+            val uri = node.getStringMemberOrDefault("uri", null)
+            val useAccelerate = node.getBooleanMemberOrDefault("useAccelerate", false)
+//            val useGlobalEndpoint = node.getBooleanMemberOrDefault("useGlobalEndpoint", false)
+            val forcePathStyle = node.getBooleanMemberOrDefault("forcePathStyle", false)
+            val useArnRegion = node.getBooleanMemberOrDefault("useArnRegion", true)
+            val useMultiRegionAccessPoints = node.getBooleanMemberOrDefault("useMultiRegionAccessPoints", true)
+
+            region?.let { writer.rust(".region(crate::config::Region::new(${it.dq()}))") }
+//            sigv4aRegionSet?.let { writer.rust("._($it)") }
+//            useAccountIdRouting?.let { writer.rust("._($it)") }
```
Velfi, line 128: We don't support this one yet

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
```diff
@@ -0,0 +1,172 @@
...
+    private val operationInput = operation.inputShape(codegenContext.model)
+
+    override fun section(section: OperationSection): Writable {
+        if (testCases.isEmpty()) return emptySection
+
+        return when (section) {
+            is OperationSection.UnitTests ->
+                writable {
+                    testCases.forEach { testCase ->
+                        Attribute(cfg("v2SmokeTests")).render(this)
```
aajtodd, line 68: **question**: Why `v2SmokeTests`? Why not just `smoketests`?

tools/ci-scripts/check-aws-sdk-services
```diff
@@ -7,9 +7,16 @@
...
+if [ -n "$ENABLE_V2_SMOKETESTS" ]; then
+    RUN_V2_SMOKETESTS="--cfg v2SmokeTests"
+else
+    RUN_V2_SMOKETESTS=""
+fi
+
 # Invoking `cargo test` at the root directory implicitly checks for the validity
 # of the top-level `Cargo.toml`
-cargo test --all-features
+cargo test --all-features -- "$RUN_V2_SMOKETESTS"
```
aajtodd, line 19: **correctness**: I don't see any `--cfg` flag for cargo, this looks like a rustc flag only. We can probably just remove this and integrate running these into that future PR/work.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationUnitTestGenerator.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun render(
+        writer: RustWriter,
+        operationShape: OperationShape,
+        customizations: List<OperationCustomization>,
+    ) {
+        // TODO(smithy-rs#3759) don't generate this if there are no customizations adding unit tests.
+        writer.rustTemplate(
+            """
+            ##[cfg(test)]
+            mod test {
```
aajtodd, line 31: This just randomly adds a `mod test` in the [middle of the operation files.](https://d2luzm2xt3nokh.cloudfront.net/codegen-diff/e4a58c360856bbbaa5c53e5db9ff25c1acaca0b9/14be1a5a285f614fb34b51d5541e1cd87c968a26/aws-sdk/index.html#file-0-0).

Can we not generate all of these into a dedicated module?

```rust
# lib.rs

#[cfg(smoketest)]
mod smoketests;
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
```diff
@@ -0,0 +1,172 @@
...
+    SmokeTestsBuilderKindBehavior(codegenContext),
+) {
+    fun renderConf(
+        writer: RustWriter,
+        data: Optional<ObjectNode>,
+    ) {
+        writer.rust("let conf = crate::config::Builder::new()")
+        writer.indent()
+        writer.rust(".behavior_version(crate::config::BehaviorVersion::latest())")
+        data.orNull()?.let { node ->
```
aajtodd, line 107: I don't think you need to parse all of these yourself, you should be able to use [AwsVendorParams](https://github.com/smithy-lang/smithy/blob/6466fe77c65b8a17b219f0b0a60c767915205f95/smithy-aws-smoke-test-model/src/main/java/software/amazon/smithy/aws/smoketests/model/AwsVendorParams.java#L14) (or the S3VendorParams) that comes from `smithy-aws-smoke-test-model` package.

```kt
// made up API for Optional/ObjectNode but you get the idea
val params = AwsVendorParams(data.getOr(ObjectNode::empty())

writer.rust(".use_dual_stack(${params.useDualStack()}")
...
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    fun renderExpectation(
+        writer: RustWriter,
+        expectation: Expectation,
+    ) {
+        if (expectation.isSuccess) {
+            writer.rust("""res.expect("request should succeed");""")
+        } else if (expectation.isFailure) {
+            writer.rust("""res.expect_err("request should fail");""")
```
aajtodd, line 169: [Smoke test trait ](https://smithy.io/2.0/additional-specs/smoke-tests.html#failureexpectation-struct) error expectation has the expected shape ID, can we assert the error is the correct type?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
Velfi: `AwsVendorParams` can't be constructed afaict. `init` is private.

aajtodd: You're right, init is private, looks like you can use [this](https://github.com/smithy-lang/smithy/blob/6466fe77c65b8a17b219f0b0a60c767915205f95/smithy-aws-smoke-test-model/src/main/java/software/amazon/smithy/aws/smoketests/model/AwsSmokeTestModel.java) to create an instance though 🤷‍♂️

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
Velfi: We have multiple conceptions of what a smoke test is in the SDK and I decided to be specific to make things less ambiguous.

Previously, we've referred to the models included in smithy-rs as "smoke test models" and I wanted to associate this feature with the related internal designation instead.

aajtodd: I think my issue is that these are associated with the [smokeTests trait](https://smithy.io/2.0/additional-specs/smoke-tests.html#smithy-test-smoketests-trait) which has no connection to a "v2SmokeTests". 

I also don't see how "v2SmokeTests" improves the status quo since it implies there was a v1 at some point. I think we could solve this pretty easily with some docs on the generated module if you take the suggestion to group them all into a common module.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
ysaito1001: I was thinking maybe we should add `TODO` for commented lines, but am curious whether you can think of a better way so that when account id based routing is supported for instance, it helps us remember to revisit here and comment in corresponding lines?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationUnitTestGenerator.kt
ysaito1001: Or as long as it's not worse than what procotol tests get rendered ([this](https://github.com/smithy-lang/smithy-rs/blob/main/aws/sdk/aws-models-extra/s3-tests.smithy#L172) gets rendered in the `put_object_test` mod in `put_object.rs`), maybe we are ok?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
Velfi: I changed it to `smoketests`

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationUnitTestGenerator.kt
Velfi: These tests are now generated into their own file in the `tests` folder.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
Velfi: I added a "throw error" for these so we'll know when we start getting test cases that require setting these.

tools/ci-scripts/check-aws-sdk-services
```diff
@@ -7,9 +7,14 @@
 set -eux
 cd aws-sdk
 
-# Invoking `cargo test` at the root directory implicitly checks for the validity
-# of the top-level `Cargo.toml`
-cargo test --all-features
+# Check if the $ENABLE_SMOKETESTS environment variable is set
+if [ -n "$ENABLE_SMOKETESTS" ]; then
```
landonxjames, line 11: Will we ever want to run these smoketests in our CI check? My understanding was that we wanted to keep all network calls out of that portion of our testing? This came up when I contributed the wasm crate because I initially put the tests in CI and was asked to move them to the Canary.

landonxjames: Ref: https://github.com/smithy-lang/smithy-rs/pull/3409#discussion_r1489802497

landonxjames (approved):
Looks good! One question about our strategy around making network calls in our CI as opposed to in the canary.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
```diff
@@ -0,0 +1,217 @@
...
+import java.util.logging.Logger
+
+class SmokeTestsDecorator : ClientCodegenDecorator {
+    override val name: String = "SmokeTests"
+    override val order: Byte = 0
+    private val logger: Logger = Logger.getLogger(javaClass.name)
+
+    private fun isSmokeTestSupported(smokeTestCase: SmokeTestCase): Boolean {
+        AwsSmokeTestModel.getAwsVendorParams(smokeTestCase)?.orNull()?.let { vendorParams ->
+            if (vendorParams.sigv4aRegionSet.isPresent) {
```
aajtodd, line 49: **Question**: Why can't we support sigv4a tests?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
```diff
@@ -0,0 +1,217 @@
...
+                logger.warning("skipping smoketest `${smokeTestCase.id}` with unsupported vendorParam `sigv4aRegionSet`")
+                return false
+            }
+            if (vendorParams.useAccountIdRouting()) {
```
aajtodd, line 53: **Suggestion**: Can we add a TODO here that links to issue for supporting account ID routing so we don't forget to enable this later?

aajtodd (approved):

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
Velfi: We don't have a way to configure region sets right now unless you manually sign a request.

tools/ci-scripts/check-aws-sdk-services
Velfi: We run these checks in Catapult too IIRC. I wanted to use the same script in both cases, but differ in whether smoketests are enabled.

ysaito1001: Related to an env var `ENABLE_SMOKETESTS`, it fails when the script runs in a container
```
./smithy-rs/tools/ci-scripts/check-aws-sdk-services: line 11: ENABLE_SMOKETESTS: unbound variable
```
we need to make sure that the script runs fine in our relase pipeline (and compiling smoke tests)

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecorator.kt
Velfi: Additionally, I can't find any service using it: https://github.com/search?q=repo%3Aaws%2Faws-models%20sigv4aRegionSet&type=code

tools/ci-scripts/check-aws-sdk-services
Velfi: I'll remove the script change for now so that we can get this merged. We can integrate it in a separate PR

ysaito1001 (approved):

Issue #3757, opened 2024-07-12
`@sensitive` doesn't work when applied to `list` and `map` targets


rcoh:
Consider the following model:

```smithy
map Timestamps {
    key: String
    value: SensitiveTimestamp 
}

@sensitive
timestamp SensitiveTimestamp
```

Because the debug printer currently [operates on the field level](https://github.com/smithy-lang/smithy-rs/blob/fe36727499a7a74e42a2beca7e76d0b433871181/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/StructureGenerator.kt#L107) it won't catch that the value in this map are sensitive.

The same issue also applies to `list` targets.

PR #3765, opened 2024-07-16
Fix maps/lists with sensitive members not redacted

landonxjames:
## Motivation and Context

Lists and maps with sensitive members were not being properly redacted. Closes https://github.com/smithy-lang/smithy-rs/issues/3757

## Description

Updated `Shape.shouldRedact` method to properly check for list and map shapes with sensitive members. Also updated the test definition so it would actually run since previously the test code was generated in a nested function inside a no-op function and never run.

## Testing

Added test cases for lists with sensitive members, maps with sensitive keys, and maps with sensitive values. 

## Checklist

----


aajtodd (commented):
This has [a lot of changes](https://d2luzm2xt3nokh.cloudfront.net/codegen-diff/0b059fb5e501d331dc688ee23fdc68581cf0bb0b/5a651004c2aecc83c1dd65f43bc60a12348b70b7/aws-sdk/index.html) that actually seem to relax debug/display implementations. Have you verified this is the desired/correct behavior change? Were we over redacting before?

landonxjames:
> This has [a lot of changes](https://d2luzm2xt3nokh.cloudfront.net/codegen-diff/0b059fb5e501d331dc688ee23fdc68581cf0bb0b/5a651004c2aecc83c1dd65f43bc60a12348b70b7/aws-sdk/index.html) that actually seem to relax debug/display implementations. Have you verified this is the desired/correct behavior change? Were we over redacting before?

Good catch! That was a bug (same one causing the CI failures) and should be fixed in https://github.com/smithy-lang/smithy-rs/pull/3765/commits/4b398c752e573dd4fa55abb9ad6fa802b1d1263a

aajtodd (approved):

ysaito1001 (approved):

ysaito1001:
Probably this warrants a  `CHANGELOG` entry with a bug-fix being true.

david-perez:
This needs to be ported over to `ServerBuilderGenerator` too.

It also looks like we never added `Debug` implementation to `ServerBuilderGeneratorWithoutPublicConstrainedTypes`.

PR #3767, opened 2024-07-17
Add client-support for RPC v2 CBOR

ysaito1001:
## Motivation and Context
Follow-up on https://github.com/smithy-lang/smithy-rs/pull/2544 to add client-side support for the protocol

## Description
The client implementation mainly focuses on a sub-section [Requests](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html#requests) in the spec. To that end, this PR addresses `TODO` for the client to fill in the blanks and  includes additional adjustments/refactoring to pass client protocol tests. 

## Testing
- Existing tests in CI
- Upstream protocol test `rpcv2Cbor`
- Our handwritten protocol test `rpcv2Cbor-extras.smithy`

----

### DIFF
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -10,3 +10,11 @@
 # references = ["smithy-rs#920"]
 # meta = { "breaking" = false, "tada" = false, "bug" = false, "target" = "client | server | all"}
 # author = "rcoh"
+
+[[smithy-rs]]
+message = """
+Add support for <a href="https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html">the Smithy RPC v2 CBOR protocol</a> both on the client and on the server.
+"""
+references = ["smithy-rs#2544", "smithy-rs#3767"]
+meta = { "breaking" = false, "tada" = true, "bug" = false, "target" = "all" }
+author = ["david-perez", "jjant", "ysaito1001"]
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -22,3 +22,9 @@ message = "Fix incorrect redaction of `@sensitive` types in maps and lists."
 references = ["smithy-rs#3765",  "smithy-rs#3757"]
 meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
 author = "landonxjames"
+
+[[smithy-rs]]
+message = "Fix client error correction to properly parse structure members that target a `Union` containing that structure recursively."
+references = ["smithy-rs#3767"]
+meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
+author = "ysaito1001"
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -25,7 +25,7 @@ author = "landonxjames"
 
 [[smithy-rs]]
 message = """
-Add support for <a href="https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html">the Smithy RPC v2 CBOR protocol</a> both on the client and on the server.
+Add initial support for [the Smithy RPC v2 CBOR protocol](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html) both on the client and on the server. The remaining work, including event streams, is being tracked under [this umbrella issue](https://github.com/smithy-lang/smithy-rs/issues/3573).
 """
 references = ["smithy-rs#2544", "smithy-rs#3767"]
 meta = { "breaking" = false, "tada" = true, "bug" = false, "target" = "all" }
--- a/CHANGELOG.next.toml
+++ b/CHANGELOG.next.toml
@@ -22,11 +22,3 @@ message = "Fix incorrect redaction of `@sensitive` types in maps and lists."
 references = ["smithy-rs#3765",  "smithy-rs#3757"]
 meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
 author = "landonxjames"
-
-[[smithy-rs]]
-message = """
-Add initial support for [the Smithy RPC v2 CBOR protocol](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html) both on the client and on the server. The remaining work, including event streams, is being tracked under [this umbrella issue](https://github.com/smithy-lang/smithy-rs/issues/3573).
-"""
-references = ["smithy-rs#2544", "smithy-rs#3767"]
-meta = { "breaking" = false, "tada" = true, "bug" = false, "target" = "all" }
-author = ["david-perez", "jjant", "ysaito1001"]
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/ErrorCorrection.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/ErrorCorrection.kt
@@ -87,7 +87,18 @@ private fun ClientCodegenContext.errorCorrectedDefault(member: MemberShape): Wri
 
             target is TimestampShape -> instantiator.instantiate(target, Node.from(0)).some()(this)
             target is BlobShape -> instantiator.instantiate(target, Node.from("")).some()(this)
-            target is UnionShape -> rust("Some(#T::Unknown)", targetSymbol)
+            target is UnionShape ->
+                rustTemplate(
+                    "Some(#{unknown})", *preludeScope,
+                    "unknown" to
+                        writable {
+                            if (memberSymbol.isRustBoxed()) {
+                                rust("Box::new(#T::Unknown)", targetSymbol)
+                            } else {
+                                rust("#T::Unknown", targetSymbol)
+                            }
+                        },
+                )
         }
     }
 }
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ClientProtocolTestGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ClientProtocolTestGenerator.kt
@@ -31,6 +31,7 @@ import software.amazon.smithy.rust.codegen.core.smithy.generators.protocol.Proto
 import software.amazon.smithy.rust.codegen.core.smithy.generators.protocol.ServiceShapeId
 import software.amazon.smithy.rust.codegen.core.smithy.generators.protocol.ServiceShapeId.AWS_JSON_10
 import software.amazon.smithy.rust.codegen.core.smithy.generators.protocol.ServiceShapeId.REST_JSON
+import software.amazon.smithy.rust.codegen.core.smithy.generators.protocol.ServiceShapeId.RPC_V2_CBOR
 import software.amazon.smithy.rust.codegen.core.smithy.generators.protocol.TestCase
 import software.amazon.smithy.rust.codegen.core.util.PANIC
 import software.amazon.smithy.rust.codegen.core.util.dq
@@ -78,6 +79,8 @@ class ClientProtocolTestGenerator(
                 FailingTest.RequestTest(AWS_JSON_10, "AwsJson10ClientPopulatesDefaultValuesInInput"),
                 FailingTest.RequestTest(REST_JSON, "RestJsonClientPopulatesDefaultValuesInInput"),
                 FailingTest.RequestTest(REST_JSON, "RestJsonClientUsesExplicitlyProvidedMemberValuesOverDefaults"),
+                FailingTest.RequestTest(RPC_V2_CBOR, "RpcV2CborClientPopulatesDefaultValuesInInput"),
+                FailingTest.RequestTest(RPC_V2_CBOR, "RpcV2CborClientUsesExplicitlyProvidedMemberValuesOverDefaults"),
             )
 
         private val BrokenTests:
@@ -268,6 +271,7 @@ class ClientProtocolTestGenerator(
             """,
             RT.sdkBody(runtimeConfig = rc),
         )
+        val mediaType = testCase.bodyMediaType.orNull()
         rustTemplate(
             """
             use #{DeserializeResponse};
@@ -280,19 +284,19 @@ class ClientProtocolTestGenerator(
             let parsed = de.deserialize_streaming(&mut http_response);
             let parsed = parsed.unwrap_or_else(|| {
                 let http_response = http_response.map(|body| {
-                    #{SdkBody}::from(#{copy_from_slice}(body.bytes().unwrap()))
+                    #{SdkBody}::from(#{copy_from_slice}(&#{decode_body_data}(body.bytes().unwrap(), #{MediaType}::from(${(mediaType ?: "unknown").dq()}))))
                 });
                 de.deserialize_nonstreaming(&http_response)
             });
             """,
             "copy_from_slice" to RT.Bytes.resolve("copy_from_slice"),
-            "SharedResponseDeserializer" to
-                RT.smithyRuntimeApiClient(rc)
-                    .resolve("client::ser_de::SharedResponseDeserializer"),
-            "Operation" to codegenContext.symbolProvider.toSymbol(operationShape),
+            "decode_body_data" to RT.protocolTest(rc, "decode_body_data"),
             "DeserializeResponse" to RT.smithyRuntimeApiClient(rc).resolve("client::ser_de::DeserializeResponse"),
+            "MediaType" to RT.protocolTest(rc, "MediaType"),
+            "Operation" to codegenContext.symbolProvider.toSymbol(operationShape),
             "RuntimePlugin" to RT.runtimePlugin(rc),
             "SdkBody" to RT.sdkBody(rc),
+            "SharedResponseDeserializer" to RT.smithyRuntimeApiClient(rc).resolve("client::ser_de::SharedResponseDeserializer"),
         )
         if (expectedShape.hasTrait<ErrorTrait>()) {
             val errorSymbol = codegenContext.symbolProvider.symbolForOperationError(operationShape)
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2Cbor.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2Cbor.kt
@@ -144,10 +144,9 @@ open class RpcV2Cbor(val codegenContext: CodegenContext) : Protocol {
     override fun parseEventStreamErrorMetadata(operationShape: OperationShape): RuntimeType =
         TODO("rpcv2Cbor event streams have not yet been implemented")
 
-    // Unlike other protocols, RpcV2Cbor also needs to set the `Content-Length` header when the input is an empty structure, see
+    // Unlike other protocols, the `rpcv2Cbor` protocol requires that `Content-Length` is always set
+    // unless there is no input or if the operation is an event stream, see
     // https://github.com/smithy-lang/smithy/blob/6466fe77c65b8a17b219f0b0a60c767915205f95/smithy-protocol-tests/model/rpcv2Cbor/empty-input-output.smithy#L106
-    // Since we generate a synthetic empty input when the input of an operation is Unit / no input, this method can
-    // return true unconditionally.
     // TODO(https://github.com/smithy-lang/smithy-rs/issues/3772): Do not set `Content-Length` for event stream operations
-    override fun needsRequestContentLength(operationShape: OperationShape) = true
+    override fun needsRequestContentLength(operationShape: OperationShape) = operationShape.input.isPresent
 }
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2Cbor.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2Cbor.kt
@@ -144,9 +144,10 @@ open class RpcV2Cbor(val codegenContext: CodegenContext) : Protocol {
     override fun parseEventStreamErrorMetadata(operationShape: OperationShape): RuntimeType =
         TODO("rpcv2Cbor event streams have not yet been implemented")
 
-    override fun needsRequestContentLength(operationShape: OperationShape): Boolean =
-        // Unlike other protocols, RpcV2Cbor also needs to set the `Content-Length` header when the input is an empty structure, see
-        // https://github.com/smithy-lang/smithy/blob/6466fe77c65b8a17b219f0b0a60c767915205f95/smithy-protocol-tests/model/rpcv2Cbor/empty-input-output.smithy#L106
-        super.needsRequestContentLength(operationShape) ||
-            !operationShape.inputShape.hasMember()
+    // Unlike other protocols, RpcV2Cbor also needs to set the `Content-Length` header when the input is an empty structure, see
+    // https://github.com/smithy-lang/smithy/blob/6466fe77c65b8a17b219f0b0a60c767915205f95/smithy-protocol-tests/model/rpcv2Cbor/empty-input-output.smithy#L106
+    // Since we generate a synthetic empty input when the input of an operation is Unit / no input, this method can
+    // return true unconditionally.
+    // TODO(https://github.com/smithy-lang/smithy-rs/issues/3772): Do not set `Content-Legnth` for event stream operations
+    override fun needsRequestContentLength(operationShape: OperationShape) = true
 }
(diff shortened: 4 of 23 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -619,7 +653,18 @@ class CborParserGenerator(
                     if (returnSymbolToParse.isUnconstrained) {
                         rust("Ok(builder)")
                     } else {
-                        rust("Ok(builder.build())")
+                        if (codegenTarget == CodegenTarget.SERVER) {
```
aajtodd, line 656: having conditional checks littered for client vs server feels not ideal. Is this common in codegen today?

aajtodd (approved):
Overall looks good.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
Velfi: Less common than it used to be 😄

Velfi (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -216,7 +247,7 @@ class CborParserGenerator(
                         val callBuilderSetMemberFieldWritable =
                             writable {
                                 withBlock("builder.${member.setterName()}(", ")") {
-                                    conditionalBlock("Some(", ")", symbolProvider.toSymbol(member).isOptional()) {
+                                    conditionalBlock("Some(", ")", codegenTarget != CodegenTarget.SERVER || symbolProvider.toSymbol(member).isOptional()) {
```
david-perez, line 250: Can't you use the other setter client builders generate? The one prefixed with `set_` (e.g. `set_dense_struct_map`) always takes `Option`s, while the one unprefixed (e.g. `dense_struct_map`) takes the actual type as modeled, so you wouldn't need `codegenTarget` check here.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -404,9 +435,12 @@ class CborParserGenerator(
                 """
                 pub(crate) fn $fnName(value: &[u8], mut builder: #{Builder}) -> #{Result}<#{Builder}, #{Error}> {
                     #{StructurePairParserFn:W}
-                    
+
                     let decoder = &mut #{Decoder}::new(value);
-                    
+                    if decoder.position() == value.len() {
+                        return #{Ok}(builder);
+                    }
```
david-perez, line 442: Isn't this equivalent to `0 == value.len()`?

Is this check necessary?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -496,7 +530,7 @@ class CborParserGenerator(
                     if (this@CborParserGenerator.returnSymbolToParse(target).isUnconstrained) {
                         rust("decoder.string()")
                     } else {
-                        rust("#T::from(u.as_ref())", symbolProvider.toSymbol(target))
+                        rust("decoder.string().map(|s| s.as_str().into())")
```
david-perez, line 533: This path would only be hit by the server if the symbol is constrained. I prefer the previous codegen since with `#T::from()` it's apparent the type you want to convert into.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
david-perez: I think this one can be avoided like we do in `JsonParserGenerator` abstracting with `BuilderInstantiator`.

https://github.com/smithy-lang/smithy-rs/blob/c63e79245430c98b2725627566b80a0fbf8e9619/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/JsonParserGenerator.kt#L524-L532

rust-runtime/aws-smithy-cbor/src/decode.rs
```diff
@@ -223,8 +230,14 @@ impl<'b> Decoder<'b> {
...
             )))
         } else {
+            // Values that are more granular than millisecond precision SHOULD be truncated to fit
+            // millisecond precision for epoch-seconds:
+            // https://smithy.io/2.0/spec/protocol-traits.html#timestamp-formats
             let epoch_seconds = self.decoder.f64().map_err(DeserializeError::new)?;
-            Ok(DateTime::from_secs_f64(epoch_seconds))
+            let mut result = DateTime::from_secs_f64(epoch_seconds);
+            let subsec_nanos = result.subsec_nanos();
+            result.set_subsec_nanos(subsec_nanos / 1_000_000 * 1_000_000);
```
david-perez, line 239: Isn't this always 0? `subsec_nanos` is a `u32`. `u32::MAX` is smaller than `10^6 * 10^6`.

Also, I would add a Rust unit test.

rust-runtime/aws-smithy-protocol-test/src/lib.rs
```diff
@@ -474,6 +475,17 @@ actual body in base64 (useful to update the protocol test):
     }
 }
 
+pub fn decode_body_data(body: &[u8], media_type: MediaType) -> Cow<'_, [u8]> {
```
david-perez, line 478: Can you change the server protocol test generator to also rely on this function? I basically code-generated the same thing when generating the tests:

https://github.com/smithy-lang/smithy-rs/blob/c63e79245430c98b2725627566b80a0fbf8e9619/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/protocol/ServerProtocolTestGenerator.kt#L458-L462

rust-runtime/inlineable/src/cbor_errors.rs
```diff
@@ -0,0 +1,70 @@
...
+fn sanitize_error_code(error_code: &str) -> &str {
+    // Trim a trailing URL from the error code, beginning with a `:`
+    let error_code = match error_code.find(':') {
+        Some(idx) => &error_code[..idx],
+        None => error_code,
+    };
+
+    // Trim a prefixing namespace from the error code, beginning with a `#`
+    match error_code.find('#') {
+        Some(idx) => &error_code[idx + 1..],
```
david-perez, line 20: Won't this panic if the error code ends with `#`?

rust-runtime/inlineable/src/cbor_errors.rs
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+use aws_smithy_cbor::decode::DeserializeError;
+use aws_smithy_cbor::Decoder;
+use aws_smithy_runtime_api::http::Headers;
+use aws_smithy_types::error::metadata::{Builder as ErrorMetadataBuilder, ErrorMetadata};
+
+fn sanitize_error_code(error_code: &str) -> &str {
+    // Trim a trailing URL from the error code, beginning with a `:`
+    let error_code = match error_code.find(':') {
```
david-perez, line 13: Since we're looking for trailing `:`, don't you want https://doc.rust-lang.org/std/primitive.str.html#method.rfind?

CHANGELOG.next.toml
```diff
@@ -22,3 +22,11 @@ message = "Fix incorrect redaction of `@sensitive` types in maps and lists."
 references = ["smithy-rs#3765",  "smithy-rs#3757"]
 meta = { "breaking" = false, "tada" = false, "bug" = true, "target" = "client" }
 author = "landonxjames"
+
+[[smithy-rs]]
+message = """
+Add support for <a href="https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html">the Smithy RPC v2 CBOR protocol</a> both on the client and on the server.
```
david-perez, line 28: Do we want to already advertise this? We should at least test event streams work correctly before that. Not saying that they don't already, but I didn't add a test in my PR checking so and this one doesn't either.

david-perez (changes requested):

CHANGELOG.next.toml
ysaito1001: 840f61a added disclaimer in the release notes.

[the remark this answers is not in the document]
ysaito1001: Fixed in 840f61a

rust-runtime/aws-smithy-cbor/src/decode.rs
ysaito1001: `subsec_nanos / 1_000_000` is performed first and then multiplication. But added parentheses for clarify and added unit test in 4d0b0b5.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
ysaito1001: Thanks for catching, client protocol tests used to panic and be unable to decode single bytes without this when my branch was off of an early commit in https://github.com/smithy-lang/smithy-rs/pull/2544, but looks like no longer needed. Removed the check in 21e1ab2.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/ErrorCorrection.kt
```diff
@@ -87,7 +87,18 @@ private fun ClientCodegenContext.errorCorrectedDefault(member: MemberShape): Wri
 
             target is TimestampShape -> instantiator.instantiate(target, Node.from(0)).some()(this)
             target is BlobShape -> instantiator.instantiate(target, Node.from("")).some()(this)
-            target is UnionShape -> rust("Some(#T::Unknown)", targetSymbol)
+            target is UnionShape ->
```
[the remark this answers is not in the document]
ysaito1001: I believe this is a latest bug that's been exposed by `rpcv2Cbor_extras`. Without the change, we'd get
```
   --> rpcv2Cbor_extras/rust-client-codegen/src/serde_util.rs:21:38
    |
21  |         builder.complex_union = Some(crate::types::ComplexUnion::Unknown)
    |                                 ---- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Box<ComplexUnion>`, found `ComplexUnion`
    |                                 |
    |                                 arguments to this enum variant are incorrect
    |
    = note: expected struct `Box<ComplexUnion>`
                 found enum `ComplexUnion`
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2Cbor.kt
```diff
@@ -111,11 +120,16 @@ open class RpcV2Cbor(val codegenContext: CodegenContext) : Protocol {
...
+        RuntimeType.cborErrors(runtimeConfig).resolve("parse_error_metadata")
 
     // TODO(https://github.com/smithy-lang/smithy-rs/issues/3573)
     override fun parseEventStreamErrorMetadata(operationShape: OperationShape): RuntimeType =
         TODO("rpcv2Cbor event streams have not yet been implemented")
+
+    override fun needsRequestContentLength(operationShape: OperationShape): Boolean =
+        // Unlike other protocols, RpcV2Cbor also needs to set the `Content-Length` header when the input is an empty structure, see
+        // https://github.com/smithy-lang/smithy/blob/6466fe77c65b8a17b219f0b0a60c767915205f95/smithy-protocol-tests/model/rpcv2Cbor/empty-input-output.smithy#L106
+        super.needsRequestContentLength(operationShape) ||
```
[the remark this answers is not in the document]
ysaito1001: Not sure if I understand the suggestion fully (I do partially).

> when the input is Unit / no input. In those cases the protocol does not require you to set Content-Length right?

If the input is Unit, then the protocol test requires us to set `Content-Length`, and that' what [this test does](https://github.com/smithy-lang/smithy/blob/26ff08f366a8556a6d69a83442f2efba19f9642b/smithy-protocol-tests/model/rpcv2Cbor/empty-input-output.smithy#L113-L115) whose input is [EmptyStructure](https://github.com/smithy-lang/smithy/blob/4cdef73b014f1ae76f7d1e61bc185a2777d48c8f/smithy-protocol-tests/model/rpcv2Cbor/main.smithy#L30-L32).

I do understand that the use of `operationShape.inputShape.hasMember` could be incorrect and that we might need a more targeted method or logic that doesn't look at synthetic operation input.

rust-runtime/inlineable/src/cbor_errors.rs
ysaito1001: No, it won't, yielding an empty string code for error code. This function `sanitize_error_code` is a copy-paste from [json_errors::sanitize_error_code](https://github.com/smithy-lang/smithy-rs/blob/1223f61e10c82b5b8172ad9e18cd6cd558203751/rust-runtime/inlineable/src/json_errors.rs#L18), and I admit it's not ideal to duplicate code, but this is in `inlineable`. I see two options:
- still duplicate the function like above (with a comment saying this has been copied from `json_errors` and even duplicate the unit test
- move `sanitize_error_code` to a Smithy runtime crate and make it `pub`, maybe in `aws-smithy-http`?

rust-runtime/inlineable/src/cbor_errors.rs
ysaito1001: If we used that, then [this unit test](https://github.com/smithy-lang/smithy-rs/blob/1223f61e10c82b5b8172ad9e18cd6cd558203751/rust-runtime/inlineable/src/json_errors.rs#L167-L174) would fail (assuming we use unaltered version of [json_errors::sanitize_error_code](https://github.com/smithy-lang/smithy-rs/blob/1223f61e10c82b5b8172ad9e18cd6cd558203751/rust-runtime/inlineable/src/json_errors.rs#L18) in `cbor_errors`).

rust-runtime/inlineable/src/cbor_errors.rs
ysaito1001: It won't, instead, will yield an empty string for error code. This function `sanitize_error_code` is a copy-paste from [json_errors::sanitize_error_code](https://github.com/smithy-lang/smithy-rs/blob/1223f61e10c82b5b8172ad9e18cd6cd558203751/rust-runtime/inlineable/src/json_errors.rs#L18), and I admit it's not ideal to duplicate code, but this is in `inlineable`. I see two options:
- still duplicate the function like above (with a comment saying this has been copied from `json_errors`) and even duplicate the unit test
- move `sanitize_error_code` to a Smithy runtime crate and make it `pub`, maybe in `aws-smithy-http`?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -80,13 +82,49 @@ class CborParserGenerator(
...
+        if (isSparse) {
+            rustTemplate(
+                """
+                decoder.null()?;
+                #{None}
+                """,
+                *codegenScope,
+            )
+        } else {
+            codegenTarget.ifServer {
```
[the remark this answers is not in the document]
ysaito1001: Addressed in 6699669

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -80,13 +82,49 @@ class CborParserGenerator(
...
         arrayOf(
+            *preludeScope,
             "SmithyCbor" to smithyCbor,
             "Decoder" to smithyCbor.resolve("Decoder"),
             "Error" to smithyCbor.resolve("decode::DeserializeError"),
             "HashMap" to RuntimeType.HashMap,
             *preludeScope,
         )
 
+    private fun handleNull(
```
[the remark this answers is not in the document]
ysaito1001: Renamed to `handleNullForCollection` in 6699669

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
ysaito1001: The short answer is no AFAIK. For the client, unprefixed setters for collections don't behave as you described, unfortunately. For instance, here is what `default_map` and `set_default_map` look like:
```
    /// Adds a key-value pair to `default_map`.
    ///
    /// To override the contents of this collection use [`set_default_map`](Self::set_default_map).
    ///
    pub fn default_map(mut self, k: impl ::std::convert::Into<::std::string::String>, v: impl ::std::convert::Into<::std::string::String>) -> Self {
        let mut hash_map = self.default_map.unwrap_or_default();
        hash_map.insert(k.into(), v.into());
        self.default_map = ::std::option::Option::Some(hash_map);
        self
    }
    pub fn set_default_map(
        mut self,
        input: ::std::option::Option<::std::collections::HashMap<::std::string::String, ::std::string::String>>,
    ) -> Self {
        self.default_map = input;
        self
    }
```
As you can see, the unprefixed setter only inserts a new key-value and is not a sibling method of the prefixed counterpart (this has been done [here](https://github.com/smithy-lang/smithy-rs/blob/1223f61e10c82b5b8172ad9e18cd6cd558203751/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt#L405) on the client side for a long time).

ysaito1001: 4d24ece will make this behavior configurable through a `CborParserGenerator`'s constructor argument.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
ysaito1001: Good to know. Incorporated in ae02348.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ClientProtocolTestGenerator.kt
```diff
@@ -268,6 +271,7 @@ class ClientProtocolTestGenerator(
             """,
             RT.sdkBody(runtimeConfig = rc),
         )
+        val mediaType = testCase.bodyMediaType.orNull()
```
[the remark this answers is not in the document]
ysaito1001: Addressed in 6ef41eb

rust-runtime/aws-smithy-protocol-test/src/lib.rs
ysaito1001: Addressed in 6ef41eb

rust-runtime/inlineable/src/cbor_errors.rs
drganjoo: No need to move `sanitize_error_code` to `aws-smithy-http`. I think David's question was about the slice starting at `idx + 1`, when `idx == len() - 1`. Since that won't panic, the implementation is okay. The conversation can be resolved.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2Cbor.kt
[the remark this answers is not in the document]
ysaito1001: Thanks for clarifying! That makes sense. 6258fc4 now returns `true` from the method unconditionally.

rust-runtime/inlineable/src/cbor_errors.rs
ysaito1001: c9886c0 added a comment to `cbor_errors::sanitize_error_code` for clarification

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
ysaito1001: Love the long explanation above, thanks a ton! Addressed David's original suggestion in 801af00

david-perez (approved):
Looks good!

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/ErrorCorrection.kt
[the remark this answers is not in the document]
ysaito1001: Good suggestion. Mentioned in 9acffc8.

rust-runtime/inlineable/src/cbor_errors.rs
ysaito1001: Added supplementary comment in 0e153b3.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/RpcV2Cbor.kt
```diff
@@ -96,26 +103,51 @@ open class RpcV2Cbor(val codegenContext: CodegenContext) : Protocol {
...
     // TODO(https://github.com/smithy-lang/smithy-rs/issues/3573)
     override fun parseEventStreamErrorMetadata(operationShape: OperationShape): RuntimeType =
         TODO("rpcv2Cbor event streams have not yet been implemented")
+
+    // Unlike other protocols, RpcV2Cbor also needs to set the `Content-Length` header when the input is an empty structure, see
+    // https://github.com/smithy-lang/smithy/blob/6466fe77c65b8a17b219f0b0a60c767915205f95/smithy-protocol-tests/model/rpcv2Cbor/empty-input-output.smithy#L106
+    // Since we generate a synthetic empty input when the input of an operation is Unit / no input, this method can
+    // return true unconditionally.
+    // TODO(https://github.com/smithy-lang/smithy-rs/issues/3772): Do not set `Content-Length` for event stream operations
+    override fun needsRequestContentLength(operationShape: OperationShape) = true
```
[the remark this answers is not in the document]
ysaito1001: Incorporated in 4e70ca2

rust-runtime/aws-smithy-cbor/src/decode.rs
```diff
@@ -338,4 +356,20 @@ mod tests {
...
+        // extracting portion representing a timestamp value.
+        let bytes = [
+            0xc1, 0xfb, 0x41, 0xcc, 0x37, 0xdb, 0x38, 0x0f, 0xbe, 0x77, 0xff,
+        ];
+        let mut decoder = Decoder::new(&bytes);
+        let timestamp = decoder.timestamp().expect("should decode timestamp");
+        // `timestamp` == 2000-01-02T20:34:56.123Z
+        assert_eq!(
+            timestamp,
+            aws_smithy_types::date_time::DateTime::from_secs_and_nanos(946845296, 123000000)
```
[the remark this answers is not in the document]
ysaito1001: Updated in 54faaa7.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/CborParserGenerator.kt
```diff
@@ -70,6 +70,8 @@ class CborParserGenerator(
     private val returnSymbolToParse: (Shape) -> ReturnSymbolToParse = { shape ->
         ReturnSymbolToParse(codegenContext.symbolProvider.toSymbol(shape), false)
     },
+    private val handleNullForNonSparseCollection: (String) -> Writable,
+    private val shouldWrapBuilderMemberSetterInputWithOption: (MemberShape) -> Boolean = { _ -> true },
```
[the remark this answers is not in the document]
ysaito1001: Added in 56d39f4.

PR #3770, opened 2024-07-19
Remove unused hyper dependency in aws-config

jackkleeman:
This allows end-users to keep hyper 0.x out of their tree

## Motivation and Context
This dependency appears unused and means that I have a 0.14 import in my tree

## Description
Removes the dependency

## Testing
Builds locally, if CI passes I think its fair to say this is a no-op. I can't find any usages of this lib.

## Checklist
I'm not sure this change warrants a changelog as there is no user facing change

----


landonxjames:
Canary run: https://github.com/smithy-lang/smithy-rs/actions/runs/10011184228

landonxjames (approved):

PR #3779, opened 2024-08-01
Improve HTTP header errors

Velfi:
This improves error messaging for HTTP header related errors.
----


rust-runtime/aws-smithy-runtime-api/src/http/headers.rs
```diff
@@ -181,12 +181,11 @@ impl TryFrom<http_02x::HeaderMap> for Headers {
...
     fn try_from(value: http_02x::HeaderMap) -> Result<Self, Self::Error> {
-        if let Some(e) = value
+        if let Some(not_utf8) = value
             .values()
-            .filter_map(|value| std::str::from_utf8(value.as_bytes()).err())
-            .next()
+            .find(|value| std::str::from_utf8(value.as_bytes()).is_err())
         {
-            Err(HttpError::header_was_not_a_string(e))
+            Err(HttpError::not_utf8(not_utf8.as_bytes()))
```
rcoh, line 188: seems like including the actual header here (including the key!) would be more helpful

rust-runtime/aws-smithy-runtime-api/src/http/error.rs
```diff
@@ -16,48 +16,113 @@ use std::str::Utf8Error;
...
+enum Kind {
+    InvalidExtensions,
+    InvalidHeaderName,
+    InvalidHeaderValue,
+    InvalidStatusCode,
+    InvalidUri,
+    InvalidUriParts,
+    MissingAuthority,
+    MissingScheme,
+    NotUtf8(Vec<u8>),
```
rcoh, line 34: why not also include the decoding error here instead of reconstructing it during display?

rcoh (commented):
strongly recommend adding some tests of invalid headers and ensuring that you get errors that help to trace back to the source.

landonxjames (approved):

rust-runtime/aws-smithy-runtime-api/src/http/headers.rs
Velfi: We can't always include the key because some header-related methods focus solely on the value. For cases where the key _is_ available, I've added it to the error message.

Velfi:
> strongly recommend adding some tests of invalid headers and ensuring that you get errors that help to trace back to the source.

I still don't fully understand where the error is coming from. I'd like to release what I have because I expect it to help and then I'll work on tracking down the source afterwards.

PR #3781, opened 2024-08-02
Allow User Agent to accept business metrics

ysaito1001:
## Motivation and Context
This PR lays the groundwork for emitting and tracking business metrics related to features like waiters, request compression, and S3 express within the Rust SDKs. It introduces `BusinessMetric` to the `user_agent` module and updates `AwsUserAgent` to support its integration.

## Description
While establishing the groundwork for the said motivation, this PR does not send metrics from features yet.
The PR deprecates existing feature and config metadata, which are now superseded by the new business metrics.

Related tasks include
- implementing versioning for the user agent string format, e.g. `ua/2.0` (note that including a user agent version string to `x-amz-user-agent` could cause many connection recording tests to fail and may also break semver check in CI just like the [content-length enforcement test](https://github.com/smithy-lang/smithy-rs/pull/3620) needed to be disabled to work around it)
- sending metrics from recently implement features, maybe starting with request compression and RPC V2 Cbor (this will also modify `x-amz-user-agent` in connection recording tests, so the argument above applies)

## Testing
- Added unit tests for `BusinessMetric` and `AwsUserAgent`
- Existing tests in CI

## Checklist

----


aajtodd (commented):
Overall looks good, couple questions.

> could cause many connection recording tests to fail and may also break semver check in CI just like the https://github.com/smithy-lang/smithy-rs/pull/3620 needed to be disabled to work around it)

1. What's our strategy here for predictable UA strings in tests?


2. Features and metrics are only really known at runtime, how is the UA string manipulated dynamically? e.g. how would an interceptor or other feature set metrics? It might be helpful to think through for all of the metrics we currently could support to see how they would be able to do this.

Velfi (approved):

ysaito1001:
> Overall looks good, couple questions.
> 
> > could cause many connection recording tests to fail and may also break semver check in CI just like the #3620 needed to be disabled to work around it)
> 
>     1. What's our strategy here for predictable UA strings in tests?

Good question. As we briefly discussed offline, it's likely we add a new tests verification utility to not compare actual vs. expected for user-agent strings for all of the connection recording tests (maybe do so only for Kotlin integ tests that do not have AWS codegen decorators).

>     2. Features and metrics are only really known at runtime, how is the UA string manipulated dynamically? e.g. how would an interceptor or other feature set metrics? It might be helpful to think through for all of the metrics we currently could support to see how they would be able to do this.

An interceptor that's relevant to the feature in question most likely extracts `AwsUserAgent` out of the config bag and calls `.add_business_metric` on it. `AwsUserAgent` is currently constructed by default in the [modify_before_signing](https://github.com/smithy-lang/smithy-rs/blob/bb1971a1ba8401fce4cea5556109e626a978da43/aws/rust-runtime/aws-runtime/src/user_agent/interceptor.rs#L91) method, but if other interceptors need to call a method on it, we may need to move the default-construction of `AwsUserAgent` to another method that's executed earlier than `modify_before_signing` during the execution in the orchestrator.

Issue #3523, opened 2024-03-27
content length enforcement would retroforwardly break tests


rcoh:
## The problem
this is a "real" semver failure and it's kindof a mind bender: https://github.com/smithy-lang/smithy-rs/actions/runs/8457663420/job/23170159614#step:4:774

A journey:
1. Content length enforcement is implemented via *default runtime plugins*. Default runtime plugins come in directly from the runtime crates. This is a cool emergent property that we can patch runtime behavior of the SDK without even releasing a new version of the SDK.
2. I merged the content length enforcement PR after we cut the release. This caught a lot of tests setting the wrong content length including this one from Dynamo: https://github.com/smithy-lang/smithy-rs/blob/745d48a74e8ffbdb0e39c86ce3db10bdfd6ccf74/aws/sdk/integration-tests/dynamodb/tests/retries-with-client-rate-limiting.rs#L29
3. We are running the tests with the released SDK against the soon-to-be-released runtime libraries. Those tests [have not been updated](https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/dynamodb/tests/retries-with-client-rate-limiting.rs#L28). Those tests fail against the new runtime libraries. 

## The way out
Disable this behavior. Release again to release all the fixed tests. Re-enable this behavior. This issue tracks re-enabling this behavior.

ysaito1001:
Addressed and concluded by the #3618

PR #3786, opened 2024-08-10
Make connection recording tests less senstive to semver hazards

ysaito1001:
## Motivation and Context
A preparatory PR that relaxes test verification of the connection recording tests.

## Description
This PR is a preparatory step for upcoming changes to the `UserAgent`, which will introduce new header values in `x-amz-user-agent`, such as `ua/2.0` (user agent metadata) and `m/A` (business metrics). 

However, the introduction of new header values will cause the following pain points:
- we have to update many connection recording tests to make them pass again (i.e. the very places updated in this PR)
- check for semver hazards [fail to pass](https://github.com/smithy-lang/smithy-rs/actions/runs/10209305234/job/28247956895). This is much the same as we encountered in [content length enforcement tests](https://github.com/smithy-lang/smithy-rs/issues/3523). This creates a chicken-and-egg problem: tests need to be updated for the PRs to pass CI, but the "released SDKs" in the `aws-sdk-rust` repository won't implement the new `UserAgent` header values until the PRs are merged and released.

To prevent recurring issues with headers affecting connection recording tests (hence semver checks), this PR preemptively updates the connection recording tests. Specifically, it adjusts them to ignore certain headers, ensuring that updates to the `x-amz-user-agent` header do not trigger semver hazards in subsequent PRs.

**Questions**:
- This PR modifies the connection recording tests to skip verification of the `x-amz-user-agent` and `authorization` headers. Consequently, we no longer test the SigV4 signature match in `aws/sdk/integration-tests`. Although we continue to run canary tests in CI, it would be beneficial to maintain at least one integration test for verifying the correctness of the SigV4 signature. This helps in detecting potential bugs affecting SigV4 signature correctness early on.
To address this, I’ve added [an awsSdkIntegrationTest](https://github.com/smithy-lang/smithy-rs/blob/f513b924dc0e624d9810889b7177cb4eda6709d2/aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecoratorTest.kt#L72) that excludes the `UserAgentInterceptor` and checks the `Signature` value in the `authorization` header. The question is, do we want to keep this test? If future header updates cause semver hazards to fail, this test would also be affected. We would then need to repeat the process we are going through with this PR: update the test, release the change to aws-sdk-rust, and only then can we make subsequent changes without breaing semver hazards.

- I've removed the commented-out tests and their associated connection recording files from `request_information_headers.rs` as part of cleanup, since there were no explanatory comments. Let me know if we want to restore these tests, and I will do so along with a comment explaining their purpose.

## Testing
- Existing tests in CI

----


aajtodd (approved):
Overall looks good, no blocking comments.


> The question is, do we want to keep this test? If future header updates cause semver hazards to fail, this test would also be affected. We would then need to repeat the process we are going through with this PR: update the test, release the change to aws-sdk-rust, and only then can we make subsequent changes without breaking semver hazards.

This seems like we want to avoid this state right? What are we verifying with this test that isn't already verified as part of our signing test suite?

> I've removed the commented-out tests and their associated connection recording files from request_information_headers.rs as part of cleanup, since there were no explanatory comments. Let me know if we want to restore these tests, and I will do so along with a comment explaining their purpose.

Do they serve a purpose? If so maybe consider keeping them, if you find they are redundant or not "carrying their weight" then remove them.

---

Only other thought that came up is most invocations involve ignoring the same set of headers (e.g. `validate_body_and_headers_except(&["x-amz-user-agent", AUTHORIZATION.as_str()], ...)` 


I wonder if that shouldn't be the default or have a convenience method like:

```rs
pub fn full_validate(...) -> ... // everything

// convenience with defaults, allows us to modify the default relaxed headers over time without adjusting every invocation in the future
pub fn relaxed_validate(media_type: &str) -> Result<...>{
     self.validate_body_and_headers_except(DEFAULT_RELAXED_HEADERS, media_type)
}

pub fn validate_body_and_headers_except(...) -> // explicit for cases where relaxed isn't the right default

```

ysaito1001:
> This seems like we want to avoid this state right? What are we verifying with this test that isn't already verified as part of our signing test suite?

Yeah, I came to this conclusion after a few days of thinking. I'll go ahead and remove it, since it doesn't add much value.

> Do they serve a purpose? If so maybe consider keeping them, if you find they are redundant or not "carrying their weight" then remove them.

Right, I probably went too far. I'll resurrect them.


> I wonder if that shouldn't be the default or have a convenience method like:
> // convenience with defaults, allows us to modify the default relaxed headers over time without adjusting every invocation in the future
> pub fn relaxed_validate(media_type: &str) -> Result<...>{
>      self.validate_body_and_headers_except(DEFAULT_RELAXED_HEADERS, media_type)
> }

I like that, given that many call sites repeatedly specify those parameters.

PR #3793, opened 2024-08-17
Provide business metrics for RPC V2 CBOR, Gzip request compression, paginator, and waiter

ysaito1001:
## Motivation and Context
Version `User-Agent` header string and begin tracking business metrics in that header

## Description
This PR versions `User-Agent` string and the version is set to `2.1`. Furthermore, we track business metrics for SDK features in `User-Agent` header. Specifically, we now track the following metrics in the User-Agent header:
- RPC V2 CBOR (M)
- Gzip request compression (L)
- paginator (C)
- waiter (B)
 
Each letter corresponds to a metric value defined [in the specification](https://github.com/smithy-lang/smithy-rs/blob/3f3c874c9f16ad65e80e5dfb7a6b8076d6342149/aws/rust-runtime/aws-runtime/src/user_agent/metrics.rs#L207-L226)).

#### Overall implementation strategy ####
Since business metrics is an AWS-specific concept, the `aws-smithy-runtime` crate cannot directly reference it or call `AwsUserAgent::add_business_metric`. Instead, the crate tracks Smithy SDK features using the config bag with the `StoreAppend` mode. During the execution of `UserAgentInterceptor::modify_before_signing`, this method retrieves the SDK features from the config bag and converts them into business metrics. This implies that any SDK features—whether specific to Smithy or AWS—that we intend to track must be added to the config bag prior to the invocation of the `modify_before_signing` method.

## Testing
- Added a test-only utility function, `assert_ua_contains_metric_values`, in the `aws-runtime` crate to verify the presence of metric values in the `User-Agent` string. Since the position of metric values in the `business-metrics` string may change as new metrics are introduced (e.g., previously `m/A` but now `m/C,A,B`), it is essential that this function accounts for potential variations and does not rely solely on substring matching.
- Added unit and integration tests to verify tracking of the business metrics introduced in this PR: RPC V2 CBOR, Gzip request compression, paginator, and waiter.

----

### DIFF
--- a/rust-runtime/inlineable/src/client_request_compression.rs
+++ b/rust-runtime/inlineable/src/client_request_compression.rs
@@ -6,6 +6,7 @@
 use aws_smithy_compression::body::compress::CompressedBody;
 use aws_smithy_compression::http::http_body_0_4_x::CompressRequest;
 use aws_smithy_compression::{CompressionAlgorithm, CompressionOptions};
+use aws_smithy_runtime::client::sdk_feature::SmithySdkFeature;
 use aws_smithy_runtime_api::box_error::BoxError;
 use aws_smithy_runtime_api::client::interceptors::context::{
     BeforeSerializationInterceptorContextRef, BeforeTransmitInterceptorContextMut,
@@ -76,10 +77,9 @@ impl Intercept for RequestCompressionInterceptor {
         "RequestCompressionInterceptor"
     }
 
-    fn read_before_serialization(
+    fn read_before_execution(
         &self,
         _context: &BeforeSerializationInterceptorContextRef<'_>,
-        _runtime_components: &RuntimeComponents,
         cfg: &mut ConfigBag,
     ) -> Result<(), BoxError> {
         let disable_request_compression = cfg
@@ -103,7 +103,7 @@ impl Intercept for RequestCompressionInterceptor {
         Ok(())
     }
 
-    fn modify_before_signing(
+    fn modify_before_retry_loop(
         &self,
         context: &mut BeforeTransmitInterceptorContextMut<'_>,
         _runtime_components: &RuntimeComponents,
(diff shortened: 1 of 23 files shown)


rust-runtime/inlineable/src/client_request_compression.rs
```diff
@@ -76,10 +77,9 @@ impl Intercept for RequestCompressionInterceptor {
         "RequestCompressionInterceptor"
     }
 
-    fn read_before_serialization(
+    fn read_before_execution(
```
aajtodd, line 80: **Question**: Why did this change? (not that it's wrong just curious)

aajtodd (approved):

landonxjames (approved):
Nice work!

rust-runtime/inlineable/src/client_request_compression.rs
ysaito1001: Ah good question. This change wasn't really necessary, but my thinking was
- `CompressionOptions` can be created at an early point in the sequence of interceptor execution, so we don't have to wait till a later interceptor method
-  Creating `CompressionOptions` doesn't require `runtime_components` so it can be done in an earlier interceptor method
- By creating CompressionOptions earlier, we allow for a larger gap between its creation and the modify_before_retry_loop, which means we can potentially run more interceptors in between in the future.

ysaito1001: Ah good question. This change wasn't really necessary, but my thinking was
- `CompressionOptions` can be created at an early point in the sequence of interceptor execution, so we don't have to wait till a later interceptor method
-  Creating `CompressionOptions` doesn't require `runtime_components` so it can be done in an earlier interceptor method
- By creating `CompressionOptions` earlier, we allow for a larger gap between its creation and the `modify_before_retry_loop`, which means we can potentially run more interceptors in between in the future

PR #3392, opened 2024-02-06
Implement and test retry classifier sorting

Velfi:
## Motivation and Context

#3322

## Description

I gave classifiers priorities but never actually sorted them. This PR fixes that.

## Testing

I wrote a test

## Checklist

- [x] I have updated `CHANGELOG.next.toml` if I made changes to the AWS SDK, generated SDK code, or SDK runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #3798, opened 2024-08-20
Fix execution order of `read_before_serialization` and `modify_before_serialization`

ysaito1001:
## Motivation and Context
Fixes the execution order of [modify_before_serialization](https://docs.rs/aws-smithy-runtime-api/latest/aws_smithy_runtime_api/client/interceptors/trait.Intercept.html#method.modify_before_serialization) and [read_before_serialization](https://docs.rs/aws-smithy-runtime-api/latest/aws_smithy_runtime_api/client/interceptors/trait.Intercept.html#method.read_before_serialization) in the orchestrator.

## Description
The fix ensures that the execution order of these interceptor methods aligns with our orchestrator design. While customers may see a behavior change as a result, we don't treat this as a new `BehaviorVersion` since it is intended as a bug fix.

## Testing
Existing tests in CI

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
+++ b/rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
@@ -217,8 +217,8 @@ async fn try_op(
 ) {
     // Before serialization
     run_interceptors!(halt_on_err: {
-        read_before_serialization(ctx, runtime_components, cfg);
         modify_before_serialization(ctx, runtime_components, cfg);
+        read_before_serialization(ctx, runtime_components, cfg);
     });
 
     // Serialization
--- a/.changelog/1724255475.md
+++ b/.changelog/1724255475.md
@@ -1,0 +1,13 @@
+---
+applies_to:
+- client
+- aws-sdk-rust
+authors:
+- ysaito1001
+references:
+- smithy-rs#3798
+breaking: false
+new_feature: false
+bug_fix: true
+---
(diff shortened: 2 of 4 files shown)


rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -217,8 +217,8 @@ async fn try_op(
 ) {
     // Before serialization
     run_interceptors!(halt_on_err: {
-        read_before_serialization(ctx, runtime_components, cfg);
         modify_before_serialization(ctx, runtime_components, cfg);
+        read_before_serialization(ctx, runtime_components, cfg);
```
landonxjames, line 221: This seems like it could be a breaking change if someone was relying on the read values not yet having been modified? I do think we should make the change since correctness > backwards compatibility IMO, but maybe it could use a new `BehaviorVersion`? That also kind of seems like overkill for a change this small, but we should probably discuss

aajtodd: When we discussed yesterday we weren't going to pursue a new `BehaviorVersion`. One of the main arguments (I have anyway) against it is it isn't "valid" to have the SDK work both ways, hence it isn't really opt-in behavior difference we would want to support. I think we treat it as a bug and make the change (perhaps bumping minor versions where appropriate).

ysaito1001: Admittedly, whether or not we introduce a new `BehaviorVersion` is a calculated risk. I code searched GitHub and internal and didn't find code that depended on the prior order of `read_before_serialization` -> `modify_before_serialization`. I believe it's safe to make this change without introducing a new `BehaviorVersion`.

> perhaps bumping minor versions where appropriate

Yeah, we can bump a minor version of `aws-smithy-runtime` in this PR.

ysaito1001: The nature of the fix is the same as https://github.com/smithy-lang/smithy-rs/pull/3392  (a new `BehaviorVersion` wasn't introduced there either).

IMHO, if the design is correct, but the implementation didn't align with it, we should first consider not introducing a new `BehaviorVersion`. If the design is incorrect/sub-optimal, the implementation aligns with it, and we later need to update both design & implementation, then we should first consider introducing a new `BehaviorVersion`.

landonxjames (approved):

PR #3826, opened 2024-09-11
Add support for converting to/from jiff::Timestamp

snaggen:
This Provides easy convertion from/from DateTime to/from jiff::Timestamp by adding simmilar convertion functions as exists for time and chrono

### DIFF
--- a/rust-runtime/Cargo.lock
+++ b/rust-runtime/Cargo.lock
@@ -794,6 +794,7 @@ dependencies = [
  "aws-smithy-types 1.2.6",
  "chrono",
  "futures-core",
+ "jiff",
  "time",
 ]
 
@@ -2063,6 +2064,12 @@ name = "itoa"
 version = "1.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "49f1f14873335454500d59611f1cf4a4b0f786f9ac11f4312a78e4cf2566695b"
+
+[[package]]
+name = "jiff"
+version = "0.1.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a45489186a6123c128fdf6016183fcfab7113e1820eb813127e036e287233fb"
 
 [[package]]
 name = "jobserver"
--- a/rust-runtime/aws-smithy-types-convert/Cargo.toml
+++ b/rust-runtime/aws-smithy-types-convert/Cargo.toml
@@ -8,6 +8,7 @@ license = "Apache-2.0"
 repository = "https://github.com/smithy-lang/smithy-rs"
 
 [features]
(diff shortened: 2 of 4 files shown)


landonxjames:
Hi snaggen, thanks for the contribution! We are always very hesitant to tie our public APIs to the APIs of other crates, especially ones like `jiff` that are not yet 1.0 since we would be tied to supporting that indefinitely. Due to that we unfortunately will not be able to accept this contribution.

BurntSushi:
@landonxjames How do y'all deal with other dependencies that aren't 1.0 yet? e.g., `chrono` and `time`.

landonxjames:
@BurntSushi the answer to that is a two parter, we recently changed our tactics on supporting things like this:

Previously we just accepted them behind feature flags and moved on. That caused some problems. Even though the dependency was behind a feature flag breaking changes could still impact our customers because the flag says nothing about which version of the dependency it is gating. Since the [unstable attribute](https://github.com/rust-lang/rfcs/issues/1491) isn't yet available to library authors we don't have a great way to telling users that this feature could break at any point (besides the parent crate just not being 1.0, but we still strive to minimize breaking changes in our pre-1.0 libs).

Now we have changed our strategy a bit. For external libraries that are pre-1.0 that have to be part of our public API (mostly Hyper and the Http types) we have a feature flag that is tied to the version of the library (see the `http-02x` and `http-1x` flags in [aws-smithy-runtime-api](https://docs.rs/crate/aws-smithy-runtime-api/1.7.2/features) for an example). This is safer and provides a more reasonable upgrade path for users, but is also a lot of work and maintenance on our end. Due to this new strategy we are trying to minimize the number of pre-1.0 crates we support to lower our own burden. 

Hopefully all this makes sense. We are a small team and have to be very careful about the maintenance burden we take on.

BurntSushi:
Thanks for the response! That all makes sense, and I would probably do the same. :)

FalkWoldmann:
Hi, since `chrono` is now [soft-deprecated](https://github.com/chronotope/chrono/issues/1768), would you be more open to supporting `jiff` 0.2 or do you want to wait for the 1.0 release?

landonxjames:
We would still prefer waiting for the 1.0 release so we don't end up in a similar situation with `jiff` as we are with `chrono`. Seems like `jiff` was slated to [go 1.0 last summer](https://github.com/rust-lang/cargo/pull/15293#discussion_r1996099608), that kind of delay makes me a bit nervous from a maintenance perspective.

FalkWoldmann:
Unfortunately, @BurntSushi is facing some serious health issues right now (https://burntsushi.net/encephalitis/), so the stable release could be delayed (wishing a good recovery!!). This paired with the `chrono` deprecation finds us in a slightly awkward predicament

PR #3831, opened 2024-09-15
Remove stalled stream protection from transcribestreaming

landonxjames:
## Motivation and Context

https://github.com/awslabs/aws-sdk-rust/issues/1181

## Checklist

----

### DIFF
--- a/.changelog/removessptranscribestreaming.md
+++ b/.changelog/removessptranscribestreaming.md
@@ -1,0 +1,12 @@
+---
+applies_to:
+  - aws-sdk-rust
+authors:
+  - landonxjames
+references: ["aws-sdk-rust#1181"]
+breaking: false
+new_feature: false
+bug_fix: true
+---
+
+Remove stalled stream protection from transcribe-streaming operations.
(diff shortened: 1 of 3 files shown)


aajtodd (approved):
This is fine for now and follows existing conventions.

I think we really want to change how we structure customizations though. This removal of stalled stream protection is [scattered](https://github.com/smithy-lang/smithy-rs/blob/5ebf4395f5b415fc24e1cb4771644762e77a6c65/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt#L90) around instead of being defined in one place. This isn't the only offender either. We should probably define it as a customization describing what it does and apply it to as many services/operations there.

e.g.

```
/// Disable stalled stream protection for specific operations that are not compatible
/// (e.g. where waiting for the service to respond is expected). 
class DisableStalledStreamProtection : ClientCodegenDecorator {
      private val appliesTo = setOf( ... )
}
```

In Kotlin (and other smithy code generators) this is pretty common (examples: [common customization](https://github.com/awslabs/aws-sdk-kotlin/blob/main/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/BackfillOptionalAuth.kt), [service specific](https://github.com/awslabs/aws-sdk-kotlin/blob/main/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/glacier/GlacierAddVersionHeader.kt))

ysaito1001 (approved):

landonxjames:
>I think we really want to change how we structure customizations though. This removal of stalled stream protection is [scattered](https://github.com/smithy-lang/smithy-rs/blob/5ebf4395f5b415fc24e1cb4771644762e77a6c65/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt#L90) around instead of being defined in one place. This isn't the only offender either. We should probably define it as a customization describing what it does and apply it to as many services/operations there.

Agree I thought of that while I was doing this, but it was the weekend and I wasn't going to spend too much time on it. I created https://github.com/smithy-lang/smithy-rs/issues/3835 to track this and I'll hopefully get around to it when oncall/flexible checksums/observability are all mopped up. Adding it to our tracking board as well.

PR #3836, opened 2024-09-16
Update Smoketest codegen to be endpoint param aware

landonxjames:
## Motivation and Context

Smoketests for service were failing because the smoke test assumed that `use_dual_stack` would be present. But the service's endpoint rules did not use that bulit-in so it was not.

## Description

Check if the service actually uses the fips or dual stack built-ins before adding them to the test.

## Testing

Did not add tests for this because the current `SmokeTestsDecoratorTest` isn't actually working and fixing it is going to be a bit of work. Want to get this in to unblock customer ASAP and will go back and update the tests once they are unblocked. 

## Checklist

----

### DIFF
--- a/.changelog/smoketestcodegen.md
+++ b/.changelog/smoketestcodegen.md
@@ -1,0 +1,12 @@
+---
+applies_to:
+  - smithy-rs
+authors:
+  - landonxjames
+references: []
+breaking: false
+new_feature: false
+bug_fix: true
+---
+
+Update Smoketest codegeneration to be endpoint built-in aware.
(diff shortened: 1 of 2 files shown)


.changelog/smoketestcodegen.md
```diff
@@ -0,0 +1,12 @@
+---
+applies_to:
+  - smithy-rs
+authors:
+  - landonxjames
+references: []
```
ysaito1001, line 6:
```suggestion
references: [smithy-rs#3836]
```

ysaito1001 (approved):

.changelog/smoketestcodegen.md
```diff
@@ -0,0 +1,12 @@
+---
+applies_to:
+  - smithy-rs
```
ysaito1001, line 3:
```suggestion
  - aws-sdk-rust
```

PR #3839, opened 2024-09-18
`aws_smithy_http_server::*` should be re-exported from generated crates

drganjoo:
Many customers have reported the need to explicitly depend on the `aws-smithy-http-server` crate in their service handler code. When they re-generate the crate, they often encounter version mismatches between the `aws-smithy-http-server` crate used in the generated code and the one they use in their service handler. This version discrepancy leads to compilation errors, requiring them to manually adjust the crate versions, which adds friction to their development workflow.

To resolve this issue, we now re-export all relevant types from `aws-smithy-http-server` within the generated crates. By doing so, customers can use these re-exported types directly, eliminating the need to depend on `aws-smithy-http-server` in their handler code.

Additionally, the generated crates no longer have the `aws-lambda` feature flag enabled by default. This prevents the `aws-lambda` feature from being automatically enabled in `aws-smithy-http-server` when the SDK is not intended for AWS Lambda.


david-perez (approved):
LGTM!

david-perez:
We should announce this in the changelog.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/ServerRequiredCustomizations.kt
```diff
@@ -51,11 +51,19 @@ class ServerRequiredCustomizations : ServerCodegenDecorator {
         rustCrate.mergeFeature(
             Feature(
                 "aws-lambda",
-                true,
+                false,
```
david-perez, line 54: Why this change?

drganjoo: I changed the default feature flag to `false` so that if the generated SDK is not intended for AWS Lambda, it won't automatically enable the `aws-lambda` feature on the `aws-smithy-http-server` crate.

PR #3840, opened 2024-09-18
Update `SmokeTestDecoratorTest`

landonxjames:
## Motivation and Context

Add tests the were skipped in https://github.com/smithy-lang/smithy-rs/pull/3836 in favor of merging and unblocking customer. 

## Description

Testing our smoketest generator with the `UseDualStack` and `UseFips` endpoint built-ins. Previously the smoketest generator assumed that these were always available, but they are only available if those built-ins are included in the endpoints rule set of the model. We now test the smoke test with both of these built-ins, with each by itself, and with neither. 

## Checklist
Just a test change, no changelog

----


aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecoratorTest.kt
```diff
@@ -109,7 +227,7 @@ class SmokeTestsDecoratorTest {
             // setting enabled by default in `.crate/config.toml` in test workspaces.
             // To work around this issue, we unset `RUSTFLAGS` to allow unused and dead code.
```
ysaito1001, line 228: As discussed offline, can we add a clarifying comment something along the lines of

> to perform a compilation-only test, we don't need to set `mapOf(Pair("RUSTFLAGS", "--cfg smoketests"))`?

I think it's worth clarifying.

Velfi (approved):

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecoratorTest.kt
```diff
@@ -94,11 +189,34 @@ class SmokeTestsDecoratorTest {
 
             @error("server")
             structure FooException { }
-            """.asSmithyModel(smithyVersion = "2")
+        """
+        val model = (imports + traitsWithAllBuiltIns + serviceDef).asSmithyModel(smithyVersion = "2")
+        val modelWithNoDualStack = (imports + traitsWithNoDualStack + serviceDef).asSmithyModel(smithyVersion = "2")
+        val modelWithNoFips = (imports + traitsWithNoFips + serviceDef).asSmithyModel(smithyVersion = "2")
+        val modelWithNeitherBuiltIn = (imports + traitsWithNeither + serviceDef).asSmithyModel(smithyVersion = "2")
```
ysaito1001, line 196: 👍

ysaito1001 (approved):
Thanks for updating the test!

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/SmokeTestsDecoratorTest.kt
landonxjames: Added clarifying comment in https://github.com/smithy-lang/smithy-rs/pull/3840/commits/c20b7b8d1d3f6eeb0ad03e4ccfc4420eff777413

PR #3845, opened 2024-09-22
Flexible Checksums Updatess

landonxjames:
## Motivation and Context

Updates to flexible checksum functionality to align with the upcoming spec updates. 

**Note:** this PR is to a feature branch since some of the changes here would be breaking before service deployments to align with the updated spec.

## Description

* Add two new `SdkConfig` fields `response_checksum_validation` and `request_checksum_calculation` that indicate the user's preferred checksum behavior
* Use `Crc32` as the default checksum algorithm
* Update the checksum calculating/validating interceptors to respect the new `SdkConfig` settings and default checksum
* Add two new interceptors (one for requests and one for responses) that each mutate the input to set the default checksum values. Note that these cannot be done in the existing inlineable interceptors because they require access to both runtime values and values from the modeled checksum trait

## Testing

Added new suite of tests as modeled in the SEP in `HttpChecksumTest.kt`

There are a few CI tests failing an all of those are currently expected:
* **Check PR semver compliance:** I marked an enum in `aws-smithy-checksums` as `non-exhaustive`. That crate is not yet 1.0 so this change should be fine (and is necessary since we will likely support different checksum types in the future).
* **Canary:** This fails on MPU since the `upload_part` contains the default `crc32` checksum but `create_multipart_upload` does not specify a checksum. S3 will fix this behavior before we deploy.
* ~~**check-aws-sdk-smoketest-unit-tests:** This fails on an S3 protocol test with `missing required header: 'content-md5'`. This should be fixed by bumping to the newest smithy version (in a separate PR) since MD5 is now deprecated.~~

## Checklist

----


aws/rust-runtime/aws-inlineable/src/http_response_checksum.rs
```diff
@@ -60,9 +61,9 @@ where
         "ResponseChecksumInterceptor"
     }
 
-    fn read_before_serialization(
+    fn modify_before_serialization(
```
ysaito1001, line 64: Do we need `modify_before_serialization` since I'm trying to see where `context` is being mutated after the change...

rust-runtime/aws-smithy-checksums/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-checksums"
-version = "0.60.12"
+version = "0.62.0"
```
ysaito1001, line 3: Are we skipping `0.61.0`?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt
```diff
@@ -43,7 +43,9 @@ val DECORATORS: List<ClientCodegenDecorator> =
             UserAgentDecorator(),
             SigV4AuthDecorator(),
             HttpRequestChecksumDecorator(),
+            HttpRequestChecksumMutationInterceptorGenerator(),
             HttpResponseChecksumDecorator(),
+            HttpResponseChecksumMutationInterceptorGenerator(),
```
ysaito1001, line 48: Should they be called `***Decorator` rather than `***Generator` since they inherit from `ClientCodegenDecorator`?

ysaito1001 (commented):
Great work on this, leaving comments accumulated so far. Will circle back.

rust-runtime/aws-smithy-types/src/checksum_config.rs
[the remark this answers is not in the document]
landonxjames: That was a last minute change done in https://github.com/smithy-lang/smithy-rs/pull/3845/commits/3175d0c73aea426356372281dd13af4ec7e23177 (so should be easy to revert if necessary). The motivations was the the `aws-smithy-checksum` crate is not 1.0, so our CI was throwing errors about stable crates exporting non-stable types. 

This PR was already big enough that I didn't want to audit and `non-exhaustive` everything in `aws-smithy-checksums` and add more noise so I moved them to `aws-smithy-types` to make our CI happy. I agree that those types living in `-checksums` makes more sense, and I think that crate could probably 1.0 fairly easily, but if we are going that route I would like to do that in another standalone PR, rebase this PR once that has merged to main, and then revert the commit I referenced above.

rust-runtime/aws-smithy-checksums/Cargo.toml
landonxjames: `0.61.0` was already published and yanked so to avoid any confusion I made this `0..62.0` instead of `0.61.1`, but I don't have a strong opinion here and am happy to change it.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt
landonxjames: Good catch, those names predated me finding the `OperationSection.RuntimePluginSupportingTypes` where these now live. Originally the codegen for these was based on the `EndpointParamsInterceptorGenerator` and I adopted the name from there.

landonxjames: Updated in https://github.com/smithy-lang/smithy-rs/pull/3845/commits/0c6059b5a5a5aed884fe927b22c3670ef3cd57d6

aws/rust-runtime/aws-inlineable/src/http_response_checksum.rs
landonxjames: Good catch! I think this was a leftover of when I was trying to set the default value on the input before I realized it needed to be in a codegened interceptor instead of an inlineable one. Fixed in https://github.com/smithy-lang/smithy-rs/pull/3845/commits/0c6059b5a5a5aed884fe927b22c3670ef3cd57d6

aws/rust-runtime/aws-config/src/default_provider/checksums.rs
```diff
@@ -0,0 +1,239 @@
...
+             |err| tracing::warn!(err = %DisplayErrorContext(&err), "invalid value for response_checksum_validation setting"),
+         )
+         .unwrap_or(None);
+
+    // response_checksum_validation should always have a non-None value and the
+    // default is WhenSupported
+    loaded.or(Some(ResponseChecksumValidation::WhenSupported))
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
landonxjames: Added these tests in https://github.com/smithy-lang/smithy-rs/pull/3845/commits/0c6059b5a5a5aed884fe927b22c3670ef3cd57d6

rust-runtime/aws-smithy-types/src/checksum_config.rs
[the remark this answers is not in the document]
landonxjames: Discussed offline and we are going to leave this as is for now. If we want to `1.0` the `aws-smithy-checksums` crate in the future we can move the types there and re-export them from `aws-smithy-types` to maintain the current behavior.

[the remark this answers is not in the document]
landonxjames: Added explanatory comment in https://github.com/smithy-lang/smithy-rs/pull/3845/commits/be5e2d4a45149092eb51c9a50cad60b2c7fc022b

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpRequestChecksumDecorator.kt
```diff
@@ -39,20 +47,49 @@ internal fun RuntimeConfig.awsInlineableHttpRequestChecksum() =
             CargoDependency.smithyHttp(this),
             CargoDependency.smithyRuntimeApiClient(this),
             CargoDependency.smithyTypes(this),
+            AwsCargoDependency.awsSigv4(this),
```
aajtodd, line 50: Why do we need awsSigv4 crate?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpRequestChecksumDecorator.kt
```diff
@@ -148,9 +193,111 @@ class HttpRequestChecksumCustomization(
...
+            else -> emptySection
+        }
+    }
+}
+
+/**
+ * Determine if the current service contains any operations with the HttpChecksum trait
+ */
+fun serviceHasHttpChecksumOperation(codegenContext: ClientCodegenContext) =
+    codegenContext.serviceShape.allOperations
```
aajtodd, line 299: **correctness**: You usually want to leverage one of the knowledge indexes to ensure you traverse the model correctly. IIRC this would travers literal operation shapes and miss resources defined on the model. In this case use `TopDownIndex` most likely.

landonxjames: Updated to use knowledge index in https://github.com/smithy-lang/smithy-rs/pull/3845/commits/618240137f2b03f7bd7a068693390bc0c487a322

ysaito1001:
> Add two new interceptors (one for requests and one for responses) that each mutate the input to set the default checksum values. Note that these cannot be done in the existing inlineable interceptors because they require access to both runtime values and values from the modeled checksum trait

I'm thinking whether we can somehow avoid adding new interceptors, since we'd want all related logic to be bundled in one place. What if we augment existing interceptors in `aws-inlineable/src/http_request_checksum.rs` by introducing another generic parameter for mutating checksum, i.e.,
```
pub(crate) struct RequestChecksumInterceptor<AP, CM> {
    algorithm_provider: AP,
    checksum_mutator: CM, // this is new, similar to the above `AP`
}
```
In the `impl` block, we call the `CM` closure:
```
impl<AP, CM> Intercept for RequestChecksumInterceptor<AP, CM>
where
    AP: Fn(&Input) -> Result<Option<ChecksumAlgorithm>, BoxError> + Send + Sync,
    CM: Fn(&mut Input) -> Result<(), BoxError> + Send + Sync, // this is new
{
...
    fn modify_before_serialization(
        &self,
        context: &mut BeforeSerializationInterceptorContextMut<'_>,
        _runtime_components: &RuntimeComponents,
        _cfg: &mut ConfigBag,
    ) -> Result<(), BoxError> {
        (self.checksum_mutator)(context.input_mut())
    }
...
}
```
To wire things up, where [we instantiate this struct in codegen](https://github.com/smithy-lang/smithy-rs/blob/37c1cc0b31fa85057bd294707bdc88d7bc91d627/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpRequestChecksumDecorator.kt#L130-L135), give [this code](https://github.com/smithy-lang/smithy-rs/blob/1f25cb4f6c9b78b28db8d13ff27b08018a2fa0c0/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpRequestChecksumMutationInterceptorGenerator.kt#L132-L155) to the closure for `CM` (we can borrow the idea from how to pass the argument for the closure `AP`). At that place, we should have access things like `checksumTrait.isRequestChecksumRequired`.

The same argument goes for `aws-inlineable/src/http_response_checksum.rs`.

I haven't verified if this idea compiles or works. But the net benefit is that we don't have to render new interceptor structs per operation.

landonxjames:
Good suggestion @ysaito1001, implemented in https://github.com/smithy-lang/smithy-rs/pull/3845/commits/d682c8df6ebfbf170e2ef1afb29c4e95b001223d and https://github.com/smithy-lang/smithy-rs/pull/3845/commits/6c9c3da7632f14697c7499317adffbea75f62afa

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpResponseChecksumDecorator.kt
```diff
@@ -86,24 +129,56 @@ class HttpResponseChecksumCustomization(
...
+                                            .load::<#{ResponseChecksumValidation}>()
+                                            .unwrap_or(&#{ResponseChecksumValidation}::WhenSupported);
+
+                                        // If validation setting is WhenSupported (or unknown) we enable response checksum
+                                        // validation. If it is WhenRequired we do not enable (since there is no way to
+                                        // indicate that a response checksum is required).
+                                        ##[allow(clippy::wildcard_in_or_patterns)]
+                                        match response_checksum_validation {
+                                            #{ResponseChecksumValidation}::WhenRequired => {}
+                                            #{ResponseChecksumValidation}::WhenSupported | _ => {
```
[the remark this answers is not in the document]
landonxjames: The `{Response | Request}ChecksumValidation` enums are marked `non-exhaustive` as a safety measure so have to handle that case somehow. Since these should always be set before this point that case shouldn't actually be encountered, but if it is we treat it the same as the default (`WhenSupported`).

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpResponseChecksumDecorator.kt
```diff
@@ -65,19 +77,50 @@ class HttpResponseChecksumDecorator : ClientCodegenDecorator {
...
+            listOf(
+                adhocCustomization<SdkConfigSection.CopySdkConfigToClientConfig> { section ->
+                    rust(
+                        """
+                        ${section.serviceConfigBuilder}.set_response_checksum_validation(${section.sdkConfig}.response_checksum_validation());
+                        """,
+                    )
+                },
+            )
+        }
```
Velfi, line 102: obviously it's too late to update this PR now, but in the future we should avoid `if not`s and instead define this and things like it like this:

```
        if (serviceHasHttpChecksumOperation(codegenContext)) {
            listOf(
                adhocCustomization<SdkConfigSection.CopySdkConfigToClientConfig> { section ->
                    rust("""
                        ${section.serviceConfigBuilder}
                            .set_response_checksum_validation(
                                ${section.sdkConfig}.response_checksum_validation()
                            );
                    """)
                },
            )
        } else {
            listOf()
        }
```

aws/sdk/integration-tests/s3/tests/express.rs
```diff
@@ -183,7 +183,12 @@ async fn presigning() {
         ][..],
         &query_params
     );
-    assert_eq!(presigned.headers().count(), 0);
+    // Presigned request has one header and that is the x-amz-checksum-mode
+    // header with default value ENABLED
+    assert_eq!(presigned.headers().count(), 1);
+    let headers = presigned.headers().collect::<Vec<(&str, &str)>>();
+    assert_eq!(headers.get(0).unwrap().0, "x-amz-checksum-mode");
+    assert_eq!(headers.get(0).unwrap().1, "ENABLED");
```
Velfi, line 191: For tuples, imo just compare them to tuples:
```rust
    assert_eq!(headers.get(0).unwrap(), ("x-amz-checksum-mode", "ENABLED"));
```

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
```diff
@@ -99,40 +103,48 @@ impl DefaultRequestChecksumOverride {
...
     ) -> Result<(), BoxError> {
+        let _ = (self.checksum_mutator)(context.input_mut(), cfg);
         let checksum_algorithm = (self.algorithm_provider)(context.input())?;
-
         let mut layer = Layer::new("RequestChecksumInterceptor");
-        layer.store_put(RequestChecksumInterceptorState { checksum_algorithm });
+        layer.store_put(RequestChecksumInterceptorState {
+            checksum_algorithm: checksum_algorithm.0,
+            request_checksum_required: checksum_algorithm.1,
+        });
```
Velfi, line 147: prefer destructuring and concise struct field assignment:
```rust
        let _ = (self.checksum_mutator)(context.input_mut(), cfg);
        let (checksum_algorithm, request_checksum_required) = (self.algorithm_provider)(context.input())?;

        let mut layer = Layer::new("RequestChecksumInterceptor");

        layer.store_put(RequestChecksumInterceptorState {
            checksum_algorithm,
            request_checksum_required,
        });
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpResponseChecksumDecorator.kt
landonxjames: Updated in https://github.com/smithy-lang/smithy-rs/commit/b80f97148c9d943509d1a8936ce5637da5ab525e

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
landonxjames: Updated in https://github.com/smithy-lang/smithy-rs/commit/b80f97148c9d943509d1a8936ce5637da5ab525e

aws/sdk/integration-tests/s3/tests/express.rs
landonxjames: Updated in https://github.com/smithy-lang/smithy-rs/commit/b80f97148c9d943509d1a8936ce5637da5ab525e

PR #3847, opened 2024-09-24
Updating Smithy to 1.51.0

landonxjames:
## Motivation and Context

Updating Smithy to 1.51.0 to get the updated `httpChecksum` trait changes

----

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,31 +2,23 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-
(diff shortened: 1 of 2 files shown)


gradle.properties
ysaito1001, line 1: nit: do you know why line breaks are gone?

ysaito1001 (approved):

gradle.properties
landonxjames: I do not, I believe either IntelliJ or our pre-commit linting hooks did that modification?

ysaito1001: No problem, I don't have strong preference so we can just keep the new revision.

PR #3848, opened 2024-09-25
add codegen test for sigv4a EPR

Velfi:
exactly what it says on the tin. Let me know if we should have any other tests verifying the function of sigv4a auth schemes.

----

### DIFF
--- a/aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecoratorTest.kt
+++ b/aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecoratorTest.kt
@@ -155,7 +155,9 @@ class SigV4AuthDecoratorTest {
                         let client = $moduleUseName::Client::from_conf(config);
                         let _ = client.some_operation().something(#{ByteStream}::from_static(b"Hello, world!")).send().await;
 
-                        http_client.assert_requests_match(&[]);
+                        http_client.assert_requests_match(&["authorization"]);
+                        let auth_header = http_client.actual_requests().next().unwrap().headers().get(http::header::AUTHORIZATION).unwrap();
+                        assert!(auth_header.contains("AWS4-ECDSA-P256-SHA256"));
                         """,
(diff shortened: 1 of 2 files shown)


aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecoratorTest.kt
```diff
@@ -56,10 +62,113 @@ class SigV4AuthDecoratorTest {
...
+                        let config = $moduleUseName::Config::builder()
+                            .http_client(http_client.clone())
+                            .endpoint_url("https://example.com")
+                            .behavior_version_latest()
+                            .with_test_defaults()
+                            .build();
+                        let client = $moduleUseName::Client::from_conf(config);
+                        let _ = client.some_operation().something(#{ByteStream}::from_static(b"Hello, world!")).send().await;
+
+                        http_client.assert_requests_match(&[]);
```
ysaito1001, line 158: A bit concerned the use of `assert_requests_match` since the verification tries matching `Signature`, which is subject to change as we add new business metrics moving forward (new business metrics may be included in `x-amz-user-agent` by default in SDK codegen decorators, so it can affect a Kotlin test like this one). When new business metrics were introduced in a PR, that PR would probably fail this test as part of semver hazards checks (which is why we introduced `relaxed_validate` in [this PR](https://github.com/smithy-lang/smithy-rs/pull/3786) to work around the problem).

Long story short, to verify the expected output of the test, I'd probably get the `authorization` header and check if it contains `AWS4-ECDSA-P256-SHA256`.

landonxjames (approved):

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecoratorTest.kt
Velfi: OK, sounds good.

ysaito1001 (approved):

PR #3851, opened 2024-09-27
Adding user-agent metrics for flexible checksums

landonxjames:
## Motivation and Context

Add user-agent metrics for checksum configs/algorithms.

## Description

Add user-agent metrics for checksum configs/algorithms.

## Testing

Add tests to `HttpChecksumTest.kt` to ensure the the user-agent features are set for configs and the selected algorithm. Note that the canary is still failing for this PR for the same reasons that it is in #3845 

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/user_agent/metrics.rs
+++ b/aws/rust-runtime/aws-runtime/src/user_agent/metrics.rs
@@ -126,7 +126,17 @@ iterable_enum!(
     AccountIdModeDisabled,
     AccountIdModeRequired,
     Sigv4aSigning,
-    ResolvedAccountId
+    ResolvedAccountId,
+    FlexibleChecksumsReqCrc32,
+    FlexibleChecksumsReqCrc32c,
+    FlexibleChecksumsReqCrc64,
+    FlexibleChecksumsReqSha1,
+    FlexibleChecksumsReqSha256,
+    FlexibleChecksumsReqWhenSupported,
+    FlexibleChecksumsReqWhenRequired,
+    FlexibleChecksumsResWhenSupported,
+    FlexibleChecksumsResWhenRequired,
+    DdbMapper
 );
 
 pub(crate) trait ProvideBusinessMetric {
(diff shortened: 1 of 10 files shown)


aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpChecksumTest.kt
```diff
@@ -427,6 +460,128 @@ internal class HttpChecksumTest {
             )
         }
     }
+
+    /**
+     * Generate miscellaneous tests
+     */
+    private fun createMiscellaneousTests(context: ClientCodegenContext): Writable {
```
ysaito1001, line 467: Great addition! Can this function somehow use `test-util` only [assert_ua_contains_metric_values](https://github.com/smithy-lang/smithy-rs/blob/abd28bc71ee91a3e59b8afc801364b909436bb94/aws/rust-runtime/aws-runtime/src/user_agent/test_util.rs#L19) (or improve it if necessary)?

ysaito1001 (commented):
Looks great overall, leaving a minor question

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpChecksumTest.kt
landonxjames: Good call, missed those test-utils earlier. Updated to use them in https://github.com/smithy-lang/smithy-rs/pull/3851/commits/d714eaa67f02859192622eefa309e5facfca81b0.

Refactored the test-util slightly and split out the extraction from the assertion since I wanted access to the actual underlying items. This is so that I could assert on the absence of some UA headers (the headers for the various checksum configs should be mutually exclusive), but that use case seems uncommon enough that it probably doesn't warrant its own util.

aws/rust-runtime/aws-runtime/src/user_agent/metrics.rs
```diff
@@ -141,6 +151,23 @@ impl ProvideBusinessMetric for SmithySdkFeature {
             Paginator => Some(BusinessMetric::Paginator),
             GzipRequestCompression => Some(BusinessMetric::GzipRequestCompression),
             ProtocolRpcV2Cbor => Some(BusinessMetric::ProtocolRpcV2Cbor),
+            FlexibleChecksumsReqCrc32 => Some(BusinessMetric::FlexibleChecksumsReqCrc32),
```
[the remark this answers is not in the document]
landonxjames: That could be helpful, but I am torn. It would certainly eliminate boilerplate, but without most dev tools having some kind of easy macro expansion utility I think it also hides some of the details. Given that it is only really used for UA things at the moment I would leave it as is, but if we were using this pattern in multiple places a macro would likely be worth it.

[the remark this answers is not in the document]
ysaito1001: Yeah, that sounds good.

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpChecksumTest.kt
ysaito1001: Cool refactoring!

ysaito1001 (approved):

PR #3857, opened 2024-10-03
Add allowlist for audit errors to `sdk-lockfiles`

ysaito1001:
## Motivation and Context
Unblocks smithy-rs#3856

## Description
See [README](https://github.com/smithy-lang/smithy-rs/blob/d09d1213971f698cffa11caec0ccd3c1911bc12c/tools/ci-build/sdk-lockfiles/README.md#false-positives)

For those who may wonder why `sdk-lockfiles audit` returns `SUCCESS` with the lockfiles in the current main branch. That's because `pin-project` [is listed in the SDK lockfile](https://github.com/smithy-lang/smithy-rs/blob/3871e9aed7ed6fbecdc45a76c7a5b75f17a941b5/aws/sdk/Cargo.lock#L3256-L3263) due to the dependency chain: `hyper-util` -> `tower` -> `pin-project`. However, since the last `cargo update` in PR3856, a more recent version of `hyper-util` no longer depends on `pin-project`, so `pin-project` does not appear in the SDK lockfile.

## Testing
- Added unit tests for the tool
- Confirmed that this PR unblocks the target PR

----

### DIFF
--- a/tools/ci-build/sdk-lockfiles/Cargo.lock
+++ b/tools/ci-build/sdk-lockfiles/Cargo.lock
@@ -933,16 +933,18 @@ dependencies = [
 
 [[package]]
 name = "sdk-lockfiles"
-version = "0.1.0"
+version = "0.1.2"
 dependencies = [
  "anyhow",
  "cargo-lock",
  "clap",
  "itertools",
+ "once_cell",
  "petgraph",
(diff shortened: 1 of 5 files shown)


landonxjames (approved):
Nice solution!

PR #3859, opened 2024-10-03
Compare query strings based on set of components in ec2 integration tests

ysaito1001:
## Motivation and Context
Avoids comparing raw strings in ec2 integration tests

## Description
With a updated service model for ec2, we have run into the following test failures in its integration tests, 
```
----- paginators_handle_unset_tokens stdout ----
body did not match. left=expected, right=actual
Diff < left / right > :
<Action=DescribeSpotPriceHistory&Version=2016-11-15&·[1;48;5;52;31mAvailabilityZone=eu-north-1a&InstanceType.1=g5.48xlarge&ProductDescription.1=Linux%2FUNIX
>Action=DescribeSpotPriceHistory&Version=2016-11-15&InstanceType.1=g5.48xlarge&ProductDescription.1=Linux%2FUNIX·[1;48;5;22;32m&AvailabilityZone=eu-north-1a
 
thread 'paginators_handle_unset_tokens' panicked at sdk/aws-smithy-runtime/src/client/http/test_util/replay.rs:98:43:

---- paginators_handle_empty_tokens stdout ----
body did not match. left=expected, right=actual
Diff < left / right > :
<Action=DescribeSpotPriceHistory&Version=2016-11-15&·[1;48;5;52;31mAvailabilityZone=eu-north-1a&InstanceType.1=g5.48xlarge&ProductDescription.1=Linux%2FUNIX
>Action=DescribeSpotPriceHistory&Version=2016-11-15&InstanceType.1=g5.48xlarge&ProductDescription.1=Linux%2FUNIX·[1;48;5;22;32m&AvailabilityZone=eu-north-1a

thread 'paginators_handle_empty_tokens' panicked at sdk/aws-smithy-runtime/src/client/http/test_util/replay.rs:98:43:
```

We don't know exactly how a generated ec2 SDK built up query strings in a different order from what it is today, but whatever the root cause is, the ultimate fix remains the same. Comparing raw query strings can be unreliable, so this PR will fix that by comparing sets of strings derived from query strings. 

## Testing
Ran the edited tests against the generated ec2 SDK in question and it passed (without this PR, it did fail).

----

### DIFF
--- a/aws/sdk/integration-tests/ec2/tests/paginators.rs
+++ b/aws/sdk/integration-tests/ec2/tests/paginators.rs
@@ -5,11 +5,10 @@
 
 use aws_runtime::user_agent::test_util::assert_ua_contains_metric_values;
 use aws_sdk_ec2::{config::Credentials, config::Region, types::InstanceType, Client, Config};
-use aws_smithy_runtime::client::http::test_util::{
-    capture_request, ReplayEvent, StaticReplayClient,
-};
+use aws_smithy_runtime::client::http::test_util::capture_request;
 use aws_smithy_runtime_api::client::http::HttpClient;
 use aws_smithy_types::body::SdkBody;
+use std::collections::HashSet;
 
 fn stub_config(http_client: impl HttpClient + 'static) -> Config {
     Config::builder()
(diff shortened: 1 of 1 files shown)


aws/sdk/integration-tests/ec2/tests/paginators.rs
```diff
@@ -19,28 +18,28 @@ fn stub_config(http_client: impl HttpClient + 'static) -> Config {
         .build()
 }
 
+fn validate_query_string(expected: &str, actual: &str) {
+    let expected = expected.split('&').collect::<HashSet<&str>>();
+    let actual = actual.split('&').collect::<HashSet<&str>>();
+    assert_eq!(expected, actual);
```
landonxjames, line 24: I think this would miss if the same param appeared in  multiple times in one and not in the other? Shouldn’t actually impact the behavior, but might be worth asserting on the matching lengths as well to catch duplications?

landonxjames (approved):
One comment, not a blocker

aws/sdk/integration-tests/ec2/tests/paginators.rs
ysaito1001: Yeah, that crossed my mind. Maybe I should collect into sorted Vecs and compare them.

ysaito1001: 1dbf37a added assertion on matching lengths

PR #3867, opened 2024-10-07
Add RDS URL signer

Velfi:
## Motivation and Context

[aws-sdk-rust/951](https://github.com/awslabs/aws-sdk-rust/issues/951)

## Description

Adds a struct for generating signed URLs for logging in to RDS. See [this doc](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.html) for more info.

## Testing

I wrote a test.

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-inlineable/Cargo.toml
+++ b/aws/rust-runtime/aws-inlineable/Cargo.toml
@@ -19,9 +19,10 @@ http_1x = ["dep:http-1x", "dep:http-body-1x", "aws-smithy-runtime-api/http-1x"]
 # https://github.com/tkaitchuck/aHash/issues/200
 # when built with `cargo update -Z minimal-versions`
 ahash = { version = "0.8.11", default-features = false }
-aws-credential-types = { path = "../aws-credential-types" }
+aws-credential-types = { path = "../aws-credential-types", features = ["test-util"] }
 aws-runtime = { path = "../aws-runtime", features = ["http-02x"] }
 aws-sigv4 = { path = "../aws-sigv4" }
+aws-types = { path = "../aws-types" }
 aws-smithy-async = { path = "../../../rust-runtime/aws-smithy-async", features = ["rt-tokio"] }
 aws-smithy-checksums = { path = "../../../rust-runtime/aws-smithy-checksums" }
 aws-smithy-http = { path = "../../../rust-runtime/aws-smithy-http" }
@@ -41,6 +42,7 @@ ring = "0.17.5"
 sha2 = "0.10"
 tokio = "1.23.1"
 tracing = "0.1"
+url = "2.5.2"
 
 [dev-dependencies]
 aws-smithy-async = { path = "../../../rust-runtime/aws-smithy-async", features = ["test-util"] }
--- a/aws/rust-runtime/aws-inlineable/src/rds_signer.rs
+++ b/aws/rust-runtime/aws-inlineable/src/rds_signer.rs
@@ -1,0 +1,242 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Code related to creating signed URLs for logging in to RDS.
+//!
+//! For more information, see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.html
+
+use aws_credential_types::provider::{ProvideCredentials, SharedCredentialsProvider};
+use aws_sigv4::http_request;
+use aws_sigv4::http_request::{SignableBody, SignableRequest, SigningSettings};
+use aws_sigv4::sign::v4;
+use aws_smithy_runtime_api::box_error::BoxError;
+use aws_smithy_runtime_api::client::identity::Identity;
+use aws_types::region::Region;
+use std::fmt::Debug;
+use std::time::Duration;
+
+const ACTION: &str = "connect";
+const SERVICE: &str = "rds-db";
+
+/// A signer that generates an auth token for a database.
+#[derive(Debug)]
+pub struct Signer {
+    config: SignerConfig,
+}
+
+impl Signer {
+    /// Given a `SignerConfig`, create a new RDS database login URL signer.
+    pub fn new(config: SignerConfig) -> Self {
+        Self { config }
+    }
+
+    /// Return a signed URL usable as an auth token.
+    pub async fn get_auth_token(
+        &self,
+        config: &aws_types::sdk_config::SdkConfig,
+    ) -> Result<String, BoxError> {
+        let credentials = self
+            .config
+            .credentials()
+            .or(config.credentials_provider().as_ref())
+            .ok_or("credentials are required to create an RDS login URL")?
+            .provide_credentials()
+            .await?;
+        let identity: Identity = credentials.into();
+        let region = self
+            .config
+            .region()
+            .or(config.region())
+            .cloned()
+            .unwrap_or_else(|| Region::new("us-east-1"));
+        let time = config.time_source().ok_or("a time source is required")?;
+
+        let mut signing_settings = SigningSettings::default();
+        signing_settings.expires_in = Some(Duration::from_secs(900));
+        signing_settings.signature_location = http_request::SignatureLocation::QueryParams;
+
+        let signing_params = v4::SigningParams::builder()
+            .identity(&identity)
+            .region(region.as_ref())
+            .name(SERVICE)
+            .time(time.now())
+            .settings(signing_settings)
+            .build()?;
+
+        let url = format!(
+            "https://{}:{}/?Action={}&DBUser={}",
+            self.config.hostname(),
+            self.config.port(),
+            ACTION,
+            self.config.username()
+        );
+
+        let signable_request =
+            SignableRequest::new("GET", &url, std::iter::empty(), SignableBody::empty())
+                .expect("signable request");
+
+        let (signing_instructions, _signature) =
+            http_request::sign(signable_request, &signing_params.into())?.into_parts();
+
+        let mut url = url::Url::parse(&url).unwrap();
+        for (name, value) in signing_instructions.params() {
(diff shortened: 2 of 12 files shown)


aws/rust-runtime/aws-inlineable/Cargo.toml
```diff
@@ -19,9 +19,10 @@ http_1x = ["dep:http-1x", "dep:http-body-1x", "aws-smithy-runtime-api/http-1x"]
 # https://github.com/tkaitchuck/aHash/issues/200
 # when built with `cargo update -Z minimal-versions`
 ahash = { version = "0.8.11", default-features = false }
-aws-credential-types = { path = "../aws-credential-types" }
+aws-credential-types = { path = "../aws-credential-types", features = ["test-util"] }
```
aajtodd, line 22: **fix**: We shouldn't enable `test-util` as part of `[dependencies]`

rust-runtime/aws-smithy-checksums/src/lib.rs
```diff
@@ -220,7 +220,7 @@ impl Checksum for Sha1 {
 }
 
 #[derive(Debug, Default)]
-struct Sha256 {
+pub struct Sha256 {
```
aajtodd, line 223: I may have missed it but didn't see this used anywhere, why public?

aws/rust-runtime/aws-inlineable/src/rds_signer.rs
```diff
@@ -0,0 +1,242 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Code related to creating signed URLs for logging in to RDS.
+//!
+//! For more information, see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.html
```
aajtodd, line 8: **comment/suggestion**: Since this is inlined for a single service it would be useful to provide an example of how to use this API in conjunction with the actual rds SDK.

aws/rust-runtime/aws-inlineable/src/rds_signer.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// Given a `SignerConfig`, create a new RDS database login URL signer.
+    pub fn new(config: SignerConfig) -> Self {
+        Self { config }
+    }
+
+    /// Return a signed URL usable as an auth token.
+    pub async fn get_auth_token(
+        &self,
+        config: &aws_types::sdk_config::SdkConfig,
+    ) -> Result<String, BoxError> {
```
aajtodd, line 39: Probably YAGNI but do we want a `String` as the result type or do we want to protect ourselves with a wrapper struct type like `struct AuthToken` that derefs to `String` or has an accessor? I'm just thinking about possible future requirements that may come along one day. Probably not worth it but just raising the question.

rust-runtime/aws-smithy-checksums/src/lib.rs
Velfi: This was used in a past version and I forgot to remove `pub` once I stopped needing this.

aws/rust-runtime/aws-inlineable/Cargo.toml
Velfi: oops

aws/rust-runtime/aws-inlineable/src/rds_signer.rs
Velfi: I couldn't figure out how to get this to work. I'm going to ask the internal team that requested this to write an example and submit a PR.

aws/rust-runtime/aws-inlineable/src/rds_signer.rs
Velfi: I think it's worth being safe. I'll do this.

aws/rust-runtime/aws-inlineable/src/rds_auth_token.rs
```diff
@@ -0,0 +1,284 @@
...
+    inner: String,
+}
+
+impl Deref for AuthToken {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
```
ysaito1001, line 66: I'm thinking whether we need to implement `Deref` _NOW_ since customers can get `&str` anyway from the `Display` trait. Implementing `Deref` feels a bit exposing the inner type and we cannot change the `Target` type in the future since it's part of public API.

ysaito1001, line 66: I'm thinking whether we need to implement `Deref` _NOW_ since customers can get `&str` anyway from the `Display` trait. Implementing `Deref` feels a bit exposing the inner type and we cannot change the `Target` type in the future since it's part of public API.

Essentially, implementing `Deref` now seems to be a one-way door. Not implementing `Deref` now gives us a two-way door and cutstomers can still get `&str` by the `Display` trait.

aajtodd: I agree `Deref` is a one way door but using `Display` trait is kind of an abuse/irregular use of that trait since customers won't be using it for display/formatting purposes. IMO I'd just offer an `as_str(): &str` method. If we ever implement deref it can forward to that 🤷

ysaito1001: Ah, actually [we get `to_string` for free](https://doc.rust-lang.org/std/string/trait.ToString.html#tymethod.to_string) once we've implemented the `Display` trait.

aws/rust-runtime/aws-inlineable/src/rds_auth_token.rs
```diff
@@ -0,0 +1,284 @@
...
+
+impl Config {
+    /// Create a new `SignerConfigBuilder`.
+    pub fn builder() -> ConfigBuilder {
+        ConfigBuilder::default()
+    }
+
+    /// The AWS credentials to sign requests with.
+    pub fn credentials(&self) -> Option<&SharedCredentialsProvider> {
```
ysaito1001, line 163: Ah, I think this needs to be `Option<SharedCredentialsProvider>` because `SharedCredentialsProvider` is already a shared pointer and whoever that receives this return value must participate in bumping up the reference count. Adding `&` will dilute that semantic.

I remember John fixing that a long time ago, and if we [code search this repo](https://github.com/search?q=repo%3Asmithy-lang%2Fsmithy-rs%20%22Option%3CShared%22&type=code), `SharedXXX` in the return value won't have a `&`.

aws/rust-runtime/aws-inlineable/src/rds_auth_token.rs
```diff
@@ -0,0 +1,284 @@
...
+        let region = self
+            .config
+            .region()
+            .or(config.region())
+            .cloned()
+            .unwrap_or_else(|| Region::new("us-east-1"));
+        let time = config.time_source().ok_or("a time source is required")?;
+
+        let mut signing_settings = SigningSettings::default();
+        signing_settings.expires_in = Some(Duration::from_secs(900));
```
ysaito1001, line 102: For `900` seconds, does this need to be configurable?

aws/rust-runtime/aws-inlineable/src/rds_auth_token.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// The region the database is located in. Uses the region inferred from the runtime if omitted.
+    region: Option<Region>,
+
+    /// The database username to login as.
+    username: Option<String>,
+}
+
+impl ConfigBuilder {
+    /// Set the AWS credentials to sign requests with. Uses the default credential provider chain if not specified.
+    pub fn credentials(mut self, credentials: SharedCredentialsProvider) -> Self {
```
ysaito1001, line 209: Should the type of `credentials` be `impl ProvideCredentials` ([like this](https://github.com/smithy-lang/smithy-rs/blob/e6b154b03e9ab1d5b8ee723d6629724fb8f6f225/aws/rust-runtime/aws-config/src/lib.rs#L475-L477)) to not require callers to have shared ownership of it?

aws/rust-runtime/aws-inlineable/src/rds_auth_token.rs
```diff
@@ -0,0 +1,284 @@
...
+                .hostname("prod-instance.us-east-1.rds.amazonaws.com")
+                .port(3306)
+                .region(Region::new("us-east-1"))
+                .username("peccy")
+                .build()
+                .unwrap(),
+        );
+
+        let signed_url = signer.auth_token(&sdk_config).await.unwrap();
+        assert_eq!(signed_url.deref(), "prod-instance.us-east-1.rds.amazonaws.com:3306/?Action=connect&DBUser=peccy&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=akid%2F20240826%2Fus-east-1%2Frds-db%2Faws4_request&X-Amz-Date=20240826T220000Z&X-Amz-Expires=900&X-Amz-SignedHeaders=host&X-Amz-Signature=dd0cba843009474347af724090233265628ace491ea17ce3eb3da098b983ad89");
```
ysaito1001, line 282: Good that we can verify `Signature` here because no business metrics are involved, which may potentially alter `Signature` in the future, when testing just the signer 👍

ysaito1001, line 282: `&signed_url.to_string()` since it already implements the `Display` trait?

ysaito1001 (commented):
Overall structure utilizing `aws-inlineable` looks good 👍 Left a few comments.

aws/rust-runtime/aws-inlineable/src/rds_auth_token.rs
Velfi: I'll make it configurable

aws/rust-runtime/aws-inlineable/src/rds_auth_token.rs
Velfi: yes

PR #3869, opened 2024-10-08
Generate MSRV in Kotlin file

david-perez:
Currently, when generating Rust crates, the MSRV is read from
`gradle.properties` by determining the project root using `git
rev-parse`. However, some build environments smithy-rs needs to build on
don't have `git` available.

This commit uses Gradle's `rootDir` to determine the project root, and
generates a Kotlin file exposing the MSRV as part of a Gradle build task
that is then published as part of `codegen-core.jar`. This also has the
benefit that consumers of smithy-rs can read the MSRV from there.

----


david-perez:
Related: https://github.com/smithy-lang/smithy-rs/pull/3841

Velfi (approved):

PR #3871, opened 2024-10-08
Make `TransferredBytes` be the top of the list in `BinLabel`

ysaito1001:
## Motivation and Context
https://github.com/awslabs/aws-sdk-rust/issues/1202

## Description
The issue above demonstrated the incorrect [BinLabel](https://github.com/smithy-lang/smithy-rs/blob/07fe426697cc30ad613568902528c305f953deb1/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs#L100-L119) ordering in [LogBuffer](https://github.com/smithy-lang/smithy-rs/blob/07fe426697cc30ad613568902528c305f953deb1/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs#L173-L183), the underlying data structure we use for stall stream protection.

The following trace logs are generated from executing the reproduction steps in the issue above. In the file labeled "no_sleep," we have commented out `std::thread::sleep(std::time::Duration::from_millis(120));` from the reproducer so the updated code can be tested as the happy path.

[s3_throughput_min_repro_no_sleep.log](https://github.com/user-attachments/files/17299373/s3_throughput_min_repro_no_sleep.log)

[s3_throughput_min_repro_with_sleep.log](https://github.com/user-attachments/files/17299447/s3_throughput_min_repro_with_sleep.log)

In both files, it’s important to note that `Bin`s assigned `TransferredBytes` can be overwritten by `Pending` due to [`ThroughputLogs::push`](https://github.com/smithy-lang/smithy-rs/blob/07fe426697cc30ad613568902528c305f953deb1/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs#L346). Once a `Bin` is labeled as `Pending`, it cannot be re-labeled.

When this occurs, the only way to avoid the stall stream protection check going into the grace period is for time to advance beyond the current `Bin`'s resolution, the `LogBuffer` pushes a new `Bin` during [`catch_up`](https://github.com/smithy-lang/smithy-rs/blob/07fe426697cc30ad613568902528c305f953deb1/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/throughput.rs#L355) , and this new `Bin` hopefully [gets assigned](https://github.com/smithy-lang/smithy-rs/blob/07fe426697cc30ad613568902528c305f953deb1/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs#L78-L79) a `TransferredBytes`. However, this new `Bin` could also be overwritten by Pending in a subsequent call to [`MinimumThroughputDownloadBody::poll_data`](https://github.com/smithy-lang/smithy-rs/blob/07fe426697cc30ad613568902528c305f953deb1/rust-runtime/aws-smithy-runtime/src/client/http/body/minimum_throughput/http_body_0_4_x.rs#L84), which can trigger the the grace period if the overall `LogBuffer` looks like it's violated the stall stream protection check.

The reproducer without sleep does not fail the stall stream protection obviously because the execution completes way  before the grace period ends, but more importantly because the execution periodically assigns new `TransferredBytes` `Bin`s in the throughput logs. This effectively resets the grace period for the stall stream protection (search for `throughput recovered; exiting grace period` in the `s3_throughput_min_repro_no_sleep.log`). However, with sleep, `Bin`s labeled as `TransferredBytes` are frequently (and almost immediately) overwritten by `Pending`. This results in the execution being unable to exit the grace period, ultimately leading to a stall stream protection error.

To resolve this, we make `TransferredBytes` be the top priority in `BinLabel`. This means once a new `Bin` has earned `TransferredBytes`, it's green for that time resolution and that it should not be revoked by `Pending` overwriting it to make it look like no bytes transferred during that time.

## Testing
- Existing tests in CI
- Added unit tests for `BinLabel` ordering and for `ThroughputLogs`
- Passed the customer's reproduction step
- To confirm the stall stream protection for download still works, I switched off WiFi while running the customer's reproducer (with sleep) and it successfully failed with the stall stream protection error:
```
---- s3_throughput_min_repro stdout ----
2024-10-08T23:29:24.999477Z DEBUG aws_smithy_runtime::client::http::body::minimum_throughput::http_body_0_4_x: current throughput: 0 B/s is below minimum: 1 B/s
2024-10-08T23:29:24.999513Z TRACE aws_smithy_runtime::client::http::body::minimum_throughput::http_body_0_4_x: received poll pending
2024-10-08T23:29:24.999530Z DEBUG aws_smithy_runtime::client::http::body::minimum_throughput::http_body_0_4_x: current throughput: 0 B/s is below minimum: 1 B/s
2024-10-08T23:29:25.081811Z TRACE aws_smithy_runtime::client::http::body::minimum_throughput::http_body_0_4_x: received poll pending
2024-10-08T23:29:25.081938Z DEBUG aws_smithy_runtime::client::http::body::minimum_throughput::http_body_0_4_x: current throughput: 0 B/s is below minimum: 1 B/s
test s3_throughput_min_repro ... FAILED
...
called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: Error { kind: StreamingError(ThroughputBelowMinimum { expected: Throughput { bytes_read: 1, per_time_elapsed: 1s }, actual: Throughput { bytes_read: 0, per_time_elapsed: 1s } }) } }
```

## Checklist

----

### DIFF
--- a/.changelog/1728489433.md
+++ b/.changelog/1728489433.md
@@ -1,0 +1,14 @@
+---
+applies_to:
+- client
+- aws-sdk-rust
+authors:
+- ysaito1001
+references:
+- smithy-rs#3871
+- aws-sdk-rust#1202
+breaking: false
+new_feature: false
+bug_fix: true
+---
+Fix minimum throughput detection for downloads to avoid incorrectly raising an error while the user is consuming data at a slow but steady pace.
--- a/rust-runtime/Cargo.lock
+++ b/rust-runtime/Cargo.lock
@@ -203,7 +203,7 @@ dependencies = [
  "aws-smithy-async 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "aws-smithy-eventstream 0.60.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "aws-smithy-http 0.60.11 (registry+https://github.com/rust-lang/crates.io-index)",
- "aws-smithy-runtime 1.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "aws-smithy-runtime 1.7.1",
  "aws-smithy-runtime-api 1.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "aws-smithy-types 1.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "aws-types",
@@ -233,7 +233,7 @@ dependencies = [
  "aws-smithy-eventstream 0.60.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "aws-smithy-http 0.60.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "aws-smithy-json 0.60.7 (registry+https://github.com/rust-lang/crates.io-index)",
- "aws-smithy-runtime 1.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "aws-smithy-runtime 1.7.1",
  "aws-smithy-runtime-api 1.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "aws-smithy-types 1.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "aws-smithy-xml 0.60.9 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -410,7 +410,7 @@ name = "aws-smithy-experimental"
 version = "0.1.4"
 dependencies = [
  "aws-smithy-async 1.2.1",
- "aws-smithy-runtime 1.7.1",
+ "aws-smithy-runtime 1.7.2",
  "aws-smithy-runtime-api 1.7.2",
  "aws-smithy-types 1.2.7",
  "h2 0.4.6",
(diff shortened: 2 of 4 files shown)


landonxjames (approved):
Looks good! One non-blocking request, would it be possible to create a test that replicates the reproduction from https://github.com/awslabs/aws-sdk-rust/issues/1202? Might be too difficult since we would need some kind of mocked server to talk to? In that case I would prioritize getting this out since it is currently causing customer issues.

Velfi (approved):

ysaito1001:
> Looks good! One non-blocking request, would it be possible to create a test that replicates the reproduction from [awslabs/aws-sdk-rust#1202](https://github.com/awslabs/aws-sdk-rust/issues/1202)? Might be too difficult since we would need some kind of mocked server to talk to? In that case I would prioritize getting this out since it is currently causing customer issues.

I think it would, using test utilities John originally wrote (it's in [integration tests](https://github.com/smithy-lang/smithy-rs/tree/main/rust-runtime/aws-smithy-runtime/tests) of `aws-smithy-runtime`).

> In that case I would prioritize getting this out since it is currently causing customer issues.

Since I've verified the customer repro and added relevant unit tests, let's get it out first to customers' hands. I'll work on adding an integ test in the next PR.

PR #3875, opened 2024-10-11
Allow multiple releases per day in the smithy-rs repository

ysaito1001:
## Motivation and Context
Currently, we can only make one `smithy-rs` release per day, and this restricts our ability to respond to urgent issues. This PR lifts that limitation, allowing us to make multiple releases per day.

## Description
The core of this change is in the `render` subcommand of `changelogger`. When generating a date-based release tag, it now checks for existing tags on the same day. If a tag already exists, the `render` subcommand will append a numerical suffix to ensure the new tag is unique.

In fact, appending a numerical suffix to make a release tag unique has been a workaround in our release pipeline (outside the `smithy-rs` repository) for quite some time. With the changes in this PR, we can eliminate that temporary solution from the release pipeline.

Now that `changelogger` requires access to previous tags, CI steps that run `generate-smithy-rs-release` need to checkout the `smithy-rs` repository with all tags (`fetch-depth: 0` is for that purpose).

## Testing
- [x] Added unit tests for `changelogger`
- [x] Successfully bumped the release tag in [dry-run](https://github.com/smithy-lang/smithy-rs/actions/runs/11356509152/job/31588857360#step:8:26) (based on [this dummy change](https://github.com/smithy-lang/smithy-rs/commit/cb19b31eaa506d66d9869261c920784ccc311b15) to trick `changelogger` into thinking that it has to bump a release tag)
- [x] Successfully bumped the release tag in the release pipeline (without the temporary hack we placed last year)

----


tools/ci-build/changelogger/src/render.rs
```diff
@@ -131,10 +131,11 @@ fn date_based_release_metadata(
     ReleaseMetadata {
         title: date_title(&now),
         tag: format!(
-            "release-{year}-{month:02}-{day:02}",
+            "release-{year}-{month:02}-{day:02}-{epoch}",
```
aajtodd, line 134: Is there anyway to make only releases after the first include an epoch? I expect it's infrequent to release twice in a day so it seems useful to keep the simple format as the default and only tack on the epoch if necessary

ysaito1001: I see the argument. We'd have to have some sort of hack in place (like the one we already have in the release pipeline where if there is already an existing tag, then append a sequential number. We are currently cloning the `aws-sdk-rust` repo with depth 20 to support this hack). We'd have to have the same hack in the `smithy-rs` repository as well to support what's suggested.

Rationale for this approach is the following:
- We don't need hack anymore, not in the smithy-rs or in the release pipeline to append a sequential number.
- I'd assume people usually see the release _TITLE_ in the release pages like `October 9th, 2024`. That won't be affected by this PR (I don't recall the last time I had to read the release tags to do something, so I assumed it was acceptable if it included an epoch suffix in it). 
- Even with the change, the release tags can be sorted by chronological and people can just read "the date portion" of the release tags, if they wish to human-read the release tags.

I'll sleep on it.

.github/workflows/release.yml
```diff
@@ -202,7 +208,7 @@ jobs:
           # to retry a release action execution that failed due to a transient issue.
           # In that case, we expect the commit to be releasable as-is, i.e. the changelog should have already
           # been processed.
-          git fetch --unshallow
+          git fetch
```
ysaito1001, line 211: Specifying `--unshallow` in a repository with `fetch-depth: 0` will fail with the following error:
```
fatal: --unshallow on a complete repository does not make sense
```

ysaito1001, line 211: FYI, specifying `--unshallow` in a repository with `fetch-depth: 0` will fail with the following error:
```
fatal: --unshallow on a complete repository does not make sense
```

tools/ci-build/changelogger/src/render.rs
```diff
@@ -97,18 +101,76 @@ pub fn subcommand_render(args: &RenderArgs) -> Result<()> {
             .unwrap_or(current_dir.as_path()),
     )
     .context("failed to find smithy-rs repo root")?;
-    let smithy_rs = GitCLI::new(&repo_root)?;
 
+    let current_tag = {
+        let cli_for_tag = if let Some(aws_sdk_rust_repo_root) = &args.aws_sdk_rust_location {
```
landonxjames, line 106: Presumably a prerequisite of running this is that the `aws-sdk-rust` repo is on `main` and that the local is up to date with `origin/main`? If there a way we can enforce that and fail if it isn't true?

.github/workflows/release.yml
```diff
@@ -181,12 +181,18 @@ jobs:
...
       uses: ./smithy-rs/.github/actions/docker-build
       with:
         action: generate-smithy-rs-release
     - name: Download all artifacts
       uses: ./smithy-rs/.github/actions/download-all-artifacts
+    # This step is not idempotent, as it pushes release artifacts to the `smithy-rs-release-1.x.y` branch. However,
+    # if this step succeeds but a subsequent step fails, retrying the release workflow is "safe" in that it does not
+    # create any inconsistent states; this step would simply fail because the release branch would be ahead of `main`
+    # due to previously pushed artifacts.
+    # To successfully retry a release, revert the commits in the release branch that pushed the artifacts.
```
landonxjames, line 195: Does this mean revert all of the commits to the release branch so it is on the same commit as main, or only the commits from the failed execution of the release? We should probably document this in our internal ops docs. 

And could we get some of the context from this comment `echo`ed from the script if it does fail, I think it would be beneficial for it to be output with the failure so whoever runs the release doesn't have to go searching through the action definition to find this comment.

landonxjames (commented):
Mostly looks good, two questions around if some validations are possible and one about improving an error message

tools/ci-build/changelogger/src/render.rs
ysaito1001: No there isn't, but only the release pipeline renders changelog for `aws-sdk-rust` in the release pipeline. It clones the `aws-sdk-rust` repo and checks out from `origin/main`. It's been like that over a couple of years now. So I'd keep what we have here.

.github/workflows/release.yml
ysaito1001: > Does this mean revert all of the commits to the release branch so it is on the same commit as main, or only the commits from the failed execution of the release?

Yes

> We should probably document this in our internal ops docs.

We documented it in our ops docs two weeks ago.

> And could we get some of the context from this comment echoed from the script if it does fail

Good idea, given the amount of changes, we'll do that in a subsequent PR (this error does happen today and this PR is not making it worse on that front).

ysaito1001: > Does this mean revert all of the commits to the release branch

Yes

> We should probably document this in our internal ops docs

We documented that in our internal docs two weeks ago (thanks @drganjoo)

> And could we get some of the context from this comment echoed from the script if it does fail

That's a reasonable idea (we need to ensure that the error is displayed no matter how the script errors out)

ysaito1001: > Does this mean revert all of the commits to the release branch

Yes

> We should probably document this in our internal ops docs

We documented that in our internal docs two weeks ago (thanks @drganjoo)

> And could we get some of the context from this comment echoed from the script if it does fail

Hmm, the error message must be placed in all subsequent scripts after this step (putting the error message in this step won't be as helpful because if this step fails, no harm has NOT been done). This error happens today, and this PR is on par with where we are (at least we have this explanatory comment). We'll keep eye on the need to echo some messages, but I'd defer it to a future PR.

ysaito1001: > Does this mean revert all of the commits to the release branch

Yes

> We should probably document this in our internal ops docs

We documented that in our internal docs two weeks ago (thanks @drganjoo)

> And could we get some of the context from this comment echoed from the script if it does fail

Hmm, the error message must be placed in all subsequent scripts after this step (putting the error message in this step won't be as helpful because if this step fails, no harm has been done). This error happens today, and this PR is on par with where we are (at least we have this explanatory comment). We'll keep eye on the need to echo some messages, but I'd defer it to a future PR.

Velfi (approved):

landonxjames (approved):

david-perez (approved):

PR #3877, opened 2024-10-15
chore: add Into conversions for Blob to [u8]

ajewellamz:
## Motivation and Context
Often I have a &[u8] or a Vec<u8>, but the sdk function takes a Blob.
It would be convenient to pass the my_vec instead of aws_smithy_types::Blob::new(my_vec)

## Description
In blob.rs I added
Into<Vec<u8>> for Blob
Into<Blob> for Vec<u8>
Into<Blob> for &[u8]

## Testing
Added a test to blob.rs to test the various scenarios.

----

### DIFF
--- a/rust-runtime/aws-smithy-types/Cargo.toml
+++ b/rust-runtime/aws-smithy-types/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-types"
-version = "1.2.7"
+version = "1.2.8"
 authors = [
     "AWS Rust SDK Team <aws-sdk-rust@amazon.com>",
     "Russell Cohen <rcoh@amazon.com>",
--- a/rust-runtime/aws-smithy-types/src/blob.rs
+++ b/rust-runtime/aws-smithy-types/src/blob.rs
@@ -31,21 +31,21 @@ impl AsRef<[u8]> for Blob {
     }
 }
 
-impl Into<Vec<u8>> for Blob {
-    fn into(self) -> Vec<u8> {
-        self.into_inner()
+impl From<Vec<u8>> for Blob {
+    fn from(value: Vec<u8>) -> Self {
+        Blob::new(value)
     }
 }
 
-impl Into<Blob> for Vec<u8> {
-    fn into(self) -> Blob {
-        Blob::new(self)
+impl From<Blob> for Vec<u8> {
+    fn from(value: Blob) -> Self {
+        value.into_inner()
     }
 }
 
-impl Into<Blob> for &[u8] {
-    fn into(self) -> Blob {
-        Blob::new(self)
+impl From<&[u8]> for Blob {
+    fn from(value: &[u8]) -> Self {
+        Blob::new(value)
     }
 }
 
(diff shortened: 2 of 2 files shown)


ysaito1001:
Thanks for contributing. I have two questions.
Q1.
> but the sdk function takes a Blob.

Is it possible that you could point us to an example function, just to be on the same page?

Q2. What's the reason for also adding
```
impl From<Blob> for Vec<u8> {
    fn from(value: Blob) -> Self {
        value.into_inner()
    }
}
```
when `Blob` already has [`.into_inner`](https://github.com/smithy-lang/smithy-rs/blob/6172d3b0091c653fe1aabca4d0fb05aa3daac5e5/rust-runtime/aws-smithy-types/src/blob.rs#L23-L25)? I see the other two `impl From` make it more ergonomic when passing `Vec<u8>` and `&[u8]` to a function taking `impl Into<Blob>`. Do you also have a use case where there is a function that takes `impl Into<Vec<u8>>` and you need to pass a `Blob`?

ajewellamz:
> Thanks for contributing. I have two questions. Q1.
> 
> > but the sdk function takes a Blob.
> 
> Is it possible that you could point us to an example function, just to be on the same page?

How about `aws_sdk_dynamodb::types::AttributeValue::B`?


> Q2. What's the reason for also adding `impl From<Blob> for Vec<u8>`

I dunno. Completeness I guess, or symmetry . I'll delete it, if it's a problem.

ysaito1001:
> How about `aws_sdk_dynamodb::types::AttributeValue::B`?
> 
> Better yet, look at the docs for aws-sdk-kms and search for `Blob` as a parameter. You'll find many many things that take blobs as input and return blobs as output.

Hmm in `aws-sdk-kms`, I don't see methods that take specifically `impl Into<Blob>` for some reason.
```
➜  kms git:(main) pwd
/Users/awsaito/src/aws-sdk-rust/sdk/kms
➜  kms git:(main) rg Blob | wc -l  
     329
➜  kms git:(main) rg Blob | tail -n 10
src/operation/generate_mac/_generate_mac_input.rs:    pub fn get_message(&self) -> &::std::option::Option<::aws_smithy_types::Blob> {
src/operation/generate_mac/builders.rs:    pub fn message(mut self, input: ::aws_smithy_types::Blob) -> Self {
src/operation/generate_mac/builders.rs:    pub fn set_message(mut self, input: ::std::option::Option<::aws_smithy_types::Blob>) -> Self {
src/operation/generate_mac/builders.rs:    pub fn get_message(&self) -> &::std::option::Option<::aws_smithy_types::Blob> {
src/types/_recipient_info.rs:    pub attestation_document: ::std::option::Option<::aws_smithy_types::Blob>,
src/types/_recipient_info.rs:    pub fn attestation_document(&self) -> ::std::option::Option<&::aws_smithy_types::Blob> {
src/types/_recipient_info.rs:    pub(crate) attestation_document: ::std::option::Option<::aws_smithy_types::Blob>,
src/types/_recipient_info.rs:    pub fn attestation_document(mut self, input: ::aws_smithy_types::Blob) -> Self {
src/types/_recipient_info.rs:    pub fn set_attestation_document(mut self, input: ::std::option::Option<::aws_smithy_types::Blob>) -> Self {
src/types/_recipient_info.rs:    pub fn get_attestation_document(&self) -> &::std::option::Option<::aws_smithy_types::Blob> {
➜  kms git:(main) rg Blob | rg Into | wc -l
       0
➜  kms git:(main) rg Blob | rg impl | wc -l
       0
```
(I don't either in `/Users/awsaito/src/aws-sdk-rust/sdk/kms/dynamodb`)

Can you send us a URL to the doc for a function/method that takes `impl Into<Blob>`, otherwise even after providing  `impl Into<Vec> for Blob` or `impl From<&[u8]> for Blob`, callers still have to call `.into()` at the call sites before calling the method in question (which seems as ergonomic as calling `Blob::new`, IMO).

ajewellamz:
> Can you send us a URL to the doc for a function/method that takes `impl Into<Blob>`, otherwise even after providing `impl Into<Vec> for Blob` or `impl From<&[u8]> for Blob`, callers still have to call `.into()` at the call sites before calling the method in question (which seems as ergonomic as calling `Blob::new`, IMO).

I was only referring to functions that that a Blob, not Into<Blob>

There's a chicken and egg problem. Why support an Into<Blob> parameter type, if there aren't any types that support Into<Blob> yet.

Personally, I think .into() is much preferable to Blob::new, because it's more idiomatic and more discoverable. The compiler will tell you "use .into()" whereas without it, one must dig through the documentation to find Blob::new.

ysaito1001 (approved):

PR #3882, opened 2024-10-17
`codegen-serde`: traverse operations of resources

david-perez:
The current implementation doesn't traverse a resource's operations if
the serialization root is a service shape; the added test fails without
this fix.

----


rcoh (approved):
LGTM, good catch

Velfi (approved):

PR #3887, opened 2024-10-18
Bump smithy version to 1.52

landonxjames:
## Motivation and Context

Bumping smithy to 1.52 for an `httpChecksum` trait bugfix

A few fixes had to be made to protocol tests for this release:
* Update document to initialize as empty `{}` instead of null
* Update our header serialization to serialize present but empty headers to `""` (Note that this change does not apply to server header serialization)
* Add `accept: application/cbor` header to add rpcV2Cbor requests
* Skip two broken tests. These tests were fixed in https://github.com/smithy-lang/smithy/pull/2423 and that change was included in the 1.52 release notes, but the actual change did not make it into the build artifact. The changes should make it into the 1.52.1 release in the next few days: https://github.com/smithy-lang/smithy/pull/2428


OKEAMAH (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -669,20 +703,36 @@ class HttpBindingGenerator(
...
-                """,
-                "HeaderValue" to RuntimeType.Http.resolve("HeaderValue"),
-                "invalid_field_error" to renderErrorMessage("header_value"),
-            )
+
+                    """,
+                    "HeaderValue" to RuntimeType.Http.resolve("HeaderValue"),
+                    "invalid_field_error" to renderErrorMessage("header_value"),
+                )
+            } else {
```
ysaito1001, line 720: We might be able to use [conditionalBlock](https://github.com/smithy-lang/smithy-rs/blob/42751e5dbf4d51c06c085e4193bf013a7333a6f5/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustWriter.kt#L125-L137) to consolidate these branches (may need to hoist `let $safeName = $formatted;`)

landonxjames: Good call, did not know about `conditionalBlock`, updated both of the conditional sections in HttpBindingGenerator.kt to use it in https://github.com/smithy-lang/smithy-rs/pull/3887/commits/c2ecd75c7ad61b893e44082dfe9faf91d6b93693

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/http/HttpBindingGenerator.kt
```diff
@@ -617,17 +625,29 @@ class HttpBindingGenerator(
...
-            )
+        // Conditionally wrap the header generation in a block that handles empty header values if
+        // `serializeEmptyHeaders` is true
+        conditionalBlock(
+            """
+            // Empty vec in header is serialized as an empty string
+            if ${context.valueExpression.name}.is_empty() {
+                builder = builder.header("$headerName", "");
+            } else {""",
+            "}", conditional = serializeEmptyHeaders,
```
ysaito1001, line 636: Nice!

ysaito1001 (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/http/RequestBindingGenerator.kt
```diff
@@ -91,7 +91,7 @@ class RequestBindingGenerator(
      */
     fun renderUpdateHttpBuilder(implBlockWriter: RustWriter) {
         uriBase(implBlockWriter)
-        val addHeadersFn = httpBindingGenerator.generateAddHeadersFn(operationShape)
+        val addHeadersFn = httpBindingGenerator.generateAddHeadersFn(operationShape, serializeEmptyHeaders = true)
```
david-perez, line 94: Why does this only affect clients?

david-perez (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/http/RequestBindingGenerator.kt
landonxjames: The new tests for this only apply to clients: https://github.com/smithy-lang/smithy/pull/2403/files

There are similar tests for the server, but they still indicate that empty headers should not be serialized by the server: https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L455

I asked internally about this discrepancy and there didn't seem to be much of a reason for it, so it might be updated in the future to align the two.

landonxjames: I made a PR to smithy to update the server tests to align: https://github.com/smithy-lang/smithy/pull/2433

PR #3888, opened 2024-10-21
migrate http test utils to new http-client crate

aajtodd:
## Motivation and Context

Previous PR(s):
    * https://github.com/smithy-lang/smithy-rs/pull/3866

## Description
This PR migrates all of the HTTP related `test-util` feature to the new `aws-smithy-http-client` crate.

* Re-export the relocated types from `aws-smithy-runtime`. I've tried to ensure the feature gates line up correctly to what was there before. All existing code depending on the `test-util` feature compiles and runs still in `smithy-rs`.
* Updated the internals of all `test-util` code to be based on `hyper-1 / http-1`. 
* There were two functions `legacy_infallible::infallible_client_fn` and `legacy_capture_request` that I had to just duplicate with the legacy types to avoid a breaking change to them. These will be re-exported from `aws-smithy-runtime` as the original types but otherwise not be used going forward.

Next steps:
* Deprecate the old `test-util` APIs and update all of our usage of `test-util` to be based on the new API's in `aws-smithy-http-client`. I may ask for team to help in this effort as it should be relatively straight forward migration work.

## Checklist

----


OKEAMAH (approved):

landonxjames (approved):
LGTM! The only CI failure that still seems relevant is the external types failures

rust-runtime/aws-smithy-runtime/src/client/http.rs
```diff
@@ -6,8 +6,38 @@
 /// Interceptor for connection poisoning.
 pub mod connection_poisoning;
 
+#[deprecated = "Direct HTTP test utility support from `aws-smithy-runtime` crate is deprecated. Please use the `test-util` feature from `aws-smithy-http-client` instead"]
```
ysaito1001, line 9: Note deprecating re-exports is currently not supported (https://github.com/rust-lang/rust/issues/30827). To do so, we need to use type aliases, e.g.,
```
#[deprecated = "...")]
pub type ... = ...;
```
But I think `pub use` is intended here because some of the deprecated items are functions and modules, to which we cannot add `pub type`.

rust-runtime/aws-smithy-http-client/src/test_util/wire.rs
```diff
@@ -199,60 +194,115 @@ pub struct WireMockServer {
...
-                            .unwrap()
-                            .push(RecordedEvent::Response(next_event.clone()));
-                        if next_event == ReplayedEvent::Timeout {
-                            tracing::info!("{} is poisoned", remote_addr);
-                            poisoned_conns.lock().unwrap().insert(remote_addr);
-                        }
-                        tracing::debug!("replying with {:?}", next_event);
-                        let event = generate_response_event(next_event).await;
-                        dbg!(event)
+            loop {
```
ysaito1001, line 247: Just curious, what caused you to work harder (i.e. the code in the `loop` block)? Is it to support graceful shutdown in hyper1.x?

ysaito1001 (approved):

rust-runtime/aws-smithy-runtime/src/client/http.rs
aajtodd: Good point though since we are deprecating the module I would assume anything in that module including re-exports is deprecated 🤔

rust-runtime/aws-smithy-http-client/src/test_util/wire.rs
aajtodd: Yeah the hyper 1.x landscape changed for servers as well and utility functions like `Sender` and `Graceful` were either removed or relocated. I had to spend a bit of time figuring out how to use the new APIs while retaining the previous behaviors and not changing the code too much.

rust-runtime/aws-smithy-runtime/src/client/http.rs
ysaito1001: Yeah exactly, it'd be nice if it were supported.

rust-runtime/aws-smithy-http-client/src/test_util/wire.rs
ysaito1001: Awesome job BTW moving test utils into the new crate.

PR #3897, opened 2024-10-30
Extract `smithy.rust#serde` trait to a new `smithy-shapes` Gradle subproject

david-perez:
----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/RustClientCodegenPlugin.kt
```diff
@@ -72,6 +73,7 @@ class RustClientCodegenPlugin : ClientDecoratableBuildPlugin() {
                 IdempotencyTokenDecorator(),
                 StalledStreamProtectionDecorator(),
                 StaticSdkFeatureTrackerDecorator(),
+                ClientSerdeDecorator(),
```
rcoh, line 76: so this turns it on by default for all clients? I'm a little worried this would accidentally start generating serializers for the Rust SDK if a team ever exported the `serde` trait in their model. I think we need to make this opt-in in some way, e.g. via a codegen setting if we are going to bundle it like this.

codegen-client/src/test/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/serde/SerdeDecoratorTest.kt
```diff
@@ -0,0 +1,77 @@
...
+import software.amazon.smithy.rust.codegen.core.testutil.asSmithyModel
+import software.amazon.smithy.rust.codegen.core.testutil.integrationTest
+import software.amazon.smithy.rust.codegen.core.testutil.unitTest
+
+/**
+ * The serde decorators for the client and the server are _identical_, but they live in separate Gradle projects.
+ * There's no point in testing everything twice, since the code is the same. Hence this file just contains a simple
+ * smoke test to ensure we don't disable the client decorator _somehow_; all tests testing the decorator's logic should
+ * live in `SerdeDecoratorTest` of the `codegen-server` project instead.
+ */
```
rcoh, line 24: this isn't really true. The code they generate is identical but client and server shapes are not the same—we really need to test against both.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/serde/SerdeDecorator.kt
```diff
@@ -26,35 +19,22 @@ val SerdeModule =
...
-        rustCrate: RustCrate,
-    ) = extrasCommon(codegenContext, rustCrate)
-}
-
-// Just a common function to keep things DRY.
-private fun extrasCommon(
+/**
+ * The entrypoint to both the client and server decorators.
+ */
+fun extrasCommon(
```
rcoh, line 25: can we rename this something related to serde—just because it's being called from an external context.

`serdeExtras` maybe?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customizations/serde/SerializeImplGenerator.kt
```diff
@@ -57,10 +56,11 @@ import software.amazon.smithy.rust.codegen.core.util.isEventStream
 import software.amazon.smithy.rust.codegen.core.util.isTargetUnit
 import software.amazon.smithy.rust.codegen.core.util.letIf
 import software.amazon.smithy.rust.codegen.core.util.toPascalCase
-import software.amazon.smithy.rust.codegen.server.smithy.ServerRustSettings
-import software.amazon.smithy.rust.codegen.server.smithy.hasConstraintTrait
 
-class SerializeImplGenerator(private val codegenContext: CodegenContext) {
+class SerializeImplGenerator(
+    private val codegenContext: CodegenContext,
+    private val unwrapConstraints: (Shape) -> Writable,
```
rcoh, line 62: this this is now a public name, we should probably name it better. It's not actually "unwrapping" constraints in that it doesn't fail if the constraints are violated. Maybe `constrainedShapeToInner` or something?

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/serde/SerdeDecoratorTest.kt
```diff
@@ -364,8 +379,8 @@ class SerdeDecoratorTest {
         "e": "A",
         "nested": {
           "int": 5,
-          "float": "-Infinity",
```
rcoh, line 367: are there any remaining tests of the stringified floats?

rcoh (changes requested):
- need to figure out the opt-in story (and how we ensure the Rust SDK isn't opted in by accident)
- Need to figure out the testing story for the clients. I don't think its sufficient to only smoke test the client given the shape-generation divergence.

PR #3898, opened 2024-10-30
switch integration tests to new test-utils

aajtodd:
## Motivation and Context
Follow up PR to https://github.com/smithy-lang/smithy-rs/pull/3888

## Description
* Update `aws/sdk/integration-tests` to use the `test-util` feature(s) directly from `aws-smithy-http-client` rather than `aws-smithy-runtime`
* ~Refactor `aws-smithy-http` to use `http-1.x~
* Fixed an erroneous error message from `imds` credentials provider in passing while tracking down a test failure
* Fixed a few tests that were not immune to environment variables (e.g. on dev desktop `AWS_EC2_METADATA_ENABLED=false` is set by default and was causing some test failures)

I started pulling at some threads and found them to be quite load bearing (e.g. `inlineable/serialization_settings.rs`) and would require updating all of codegen to use `http-1.x`. ~This PR is a bit of a half step with some stop gaps put into place (e.g. the new `compat` feature of `aws-smithy-http` or inlining a function) to prevent the need to update everything in a single PR. Future PR's will continue where I left off here. ~

One lesson I learned the hard way was that our integration test `Cargo.toml` have to reflect codegen. I had updated them to all use `http = "1"` and got it all working only to realize codegen generates `http = "0.2.9"` and `http-1x = { package = "http", version = "1"}` and thus the integration tests have to look the same as that.

### Future

* ~Continue to refactor `aws-smithy-http` to upgrade to `http-body-1.x`. ~
* Refactor codegen to update to `http-1.x`

UPDATE: I reverted the update to http 1.x for `aws-smithy-http`. This needs to be done and planned with the server and I don't want it to unnecessarily block hyper 1. We will pursue the remaining hyper 1 changes and then make a decision on internal use of http and http-body pre 1.x usage. 

----


aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
```diff
@@ -169,48 +169,48 @@ impl SigningInstructions {
...
+            )
+            .expect("unreachable: URI is valid");
             for (name, value) in new_query {
                 query.insert(name, &value);
             }
-            *request.uri_mut() = query.build_uri();
+            *request.uri_mut() = query
+                .build_uri()
+                .to_string()
+                .parse()
```
Velfi, line 191: This looks weird. is this a 0.4 to 1.0 conversion? Why do we create a URI, stringify it, and then parse it as a URI?

rust-runtime/aws-smithy-checksums/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-checksums"
-version = "0.60.13"
+version = "0.70.0"
```
Velfi, line 3: Not that it really matters, but I would expect the next version to be `v0.61.0`

Velfi (approved):

rust-runtime/aws-smithy-checksums/Cargo.toml
landonxjames: I hit this with the flexibile checksums work, `0.61.0` was already published and yanked: https://crates.io/crates/aws-smithy-checksums/versions

But I would probably expect `0.62.0`

aws/rust-runtime/aws-inlineable/Cargo.toml
```diff
@@ -12,7 +12,7 @@ publish = false
 repository = "https://github.com/smithy-lang/smithy-rs"
 
 [features]
-http_1x = ["dep:http-1x", "dep:http-body-1x", "aws-smithy-runtime-api/http-1x"]
+http_1x = ["dep:http-body-1x", "aws-smithy-runtime-api/http-1x"]
```
ysaito1001, line 15: FYI, the feature name was updated to `http-1x` in [the main branch](https://github.com/smithy-lang/smithy-rs/blob/main/aws/rust-runtime/aws-inlineable/Cargo.toml#L15), as the updated nightly compiler told us that those who used this feature were specifying `http-1x` (and we tend to use a hyphen in feature names in our crates).

rust-runtime/aws-smithy-checksums/Cargo.toml
aajtodd: Good point.

ysaito1001 (approved):
Thanks for going through so many places and updating each site carefully. I know CI is currently failing on server-related tests, but LGTM.

PR #3903, opened 2024-11-05
increase windows CI timeout from 20 to 30 minutes

Velfi:
This is intended to avoid the timeout issue happening in the private staging repo

----


aws/rust-runtime/aws-config/Cargo.lock
ysaito1001, line 1: Maybe we can exclude this file from this PR? Some dependencies got even downgraded :)

ysaito1001 (approved):

aws/rust-runtime/aws-config/Cargo.lock
Velfi: oops

PR #3912, opened 2024-11-11
Add `aws-smithy-observability` and `aws-smithy-observability-otel` crates

landonxjames:
## Motivation and Context

We would like to have a consistent way to measure the performance of the SDK going forward so we can evaluate how updates change the performance over time. The changes in this PR are a first step towards that, adding the interfaces (and one implementation) that we will use to instrument our runtime crates. 

## Description

Add two new crates
* `aws-smithy-observability` - contains traits for our observability solution and a `global` module for managing the global telemetry provider
* `aws-smithy-observability-otel` - contains an OpenTelemetry based implementation of the traits

## Testing

Added new tests in both crates

## Checklist
Note: leaving out a changelog entry for now since these crates are somewhat useless until we instrument our runtime crates

----


rust-runtime/aws-smithy-observability-otel/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[dependencies]
+aws-smithy-observability = { path = "../aws-smithy-observability" }
+opentelemetry = {version = "0.26.0", features = ["metrics"]}
+# The below dependencies are transitive, but must be pinned until we can generate a new lockfile
+# lockfile generation currently blocked by aws-lc-sys build issue
+value-bag = "1.10.0"
+async-global-executor = "2.4.1"
+async-task = "=4.7.1"
+
+# This crate cannot beused on powerpc
```
Velfi, line 21:
```suggestion
# This crate cannot be used on powerpc
```

Velfi (approved):

rust-runtime/aws-smithy-observability/README.md
```diff
@@ -0,0 +1,7 @@
+# aws-smithy-observability
+
+This crate contains traits allowing for the implementation of `TelemetryProvider`s for the AWS SDK for Rust. I also contains a `global` module for setting and interacting with the current `GlobalTelemetryProvider`.
```
aajtodd, line 3: I -> It also contains

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,118 @@
...
+//! Metrics are used to gain insight into the operational performance and health of a system in
+//! real time.
+
+use crate::attributes::{Attributes, Context};
+use crate::error::ObservabilityError;
+
+/// Provides named instances of [Meter].
+pub trait MeterProvider {
+    /// Get or create a named [Meter].
+    fn get_meter(&self, scope: &'static str, attributes: Option<&Attributes>) -> Box<dyn Meter>;
```
aajtodd, line 15: **discuss** Should this be `-> impl Meter`?

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    /// Optional method to flush the metrics pipeline, default is noop
+    fn flush(&self) -> Result<(), ObservabilityError> {
```
aajtodd, line 18: **correctness** Flush and shutdown aren't part of SRA, what purpose do these serve in practice? Why do we need them?

Also they don't feel like they belong here anyway as these are concerns of the consumer usually not of the library doing the instrumenting (which is the primary purpose/use of this API). In other words the instrumented code should not be responsible for flush/shutdown call.s

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,118 @@
...
+pub trait Meter {
+    /// Create a new Gauge.
+    #[allow(clippy::type_complexity)]
+    fn create_gauge(
+        &self,
+        name: String,
+        callback: Box<dyn Fn(&dyn AsyncMeasurement<Value = f64>) + Send + Sync>,
+        units: Option<String>,
+        description: Option<String>,
+    ) -> Box<dyn AsyncMeasurement<Value = f64>>;
```
aajtodd, line 38: Same question should these be `Box<T> `or should they be `impl -> T`?

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+/// The entry point to creating instruments. A grouping of related metrics.
+pub trait Meter {
+    /// Create a new Gauge.
+    #[allow(clippy::type_complexity)]
+    fn create_gauge(
+        &self,
+        name: String,
```
aajtodd, line 34: These should probably be more flexible like `impl Into<String>` or `impl Into<Cow<'static, str>>`, etc.

rust-runtime/aws-smithy-observability/src/provider.rs
```diff
@@ -0,0 +1,82 @@
...
+// compatibilty with what we have today.
+/// A builder for [TelemetryProvider].
+#[non_exhaustive]
+pub struct TelemetryProviderBuilder {
+    meter_provider: Box<dyn MeterProvider + Send + Sync>,
+}
+
+impl TelemetryProviderBuilder {
+    /// Set the [MeterProvider].
+    pub fn meter_provider(mut self, meter_provider: Box<dyn MeterProvider + Send + Sync>) -> Self {
```
aajtodd, line 52: should probably take `impl MeterProvider` or something more flexible.

rust-runtime/aws-smithy-observability/src/global.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use crate::provider::{GlobalTelemetryProvider, TelemetryProvider};
+
+// Statically store the global provider
+static GLOBAL_TELEMETRY_PROVIDER: Lazy<RwLock<GlobalTelemetryProvider>> =
+    Lazy::new(|| RwLock::new(GlobalTelemetryProvider::new(TelemetryProvider::default())));
+
+/// Set the current global [TelemetryProvider]. If [None] is supplied then a noop provider is set.
+/// The previous [TelemetryProvider] is returned in an [Arc] so appropriate cleanup can be done if necessary.
+pub fn set_global_telemetry_provider(
```
aajtodd, line 22: Given the namespacing of this in module `global` we can probably drop `_global_` from the name (same for `get`)

rust-runtime/aws-smithy-observability/src/global.rs
```diff
@@ -0,0 +1,82 @@
...
+    new_provider: Option<TelemetryProvider>,
+) -> Arc<TelemetryProvider> {
```
aajtodd, line 24: I see the comment on why returning the old one is done but I'm not sure I agree/understand why it's necessary in practice. I would think as a customer it would be set and forget.

rust-runtime/aws-smithy-observability/src/global.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::provider::{GlobalTelemetryProvider, TelemetryProvider};
+
+// Statically store the global provider
+static GLOBAL_TELEMETRY_PROVIDER: Lazy<RwLock<GlobalTelemetryProvider>> =
+    Lazy::new(|| RwLock::new(GlobalTelemetryProvider::new(TelemetryProvider::default())));
+
+/// Set the current global [TelemetryProvider]. If [None] is supplied then a noop provider is set.
+/// The previous [TelemetryProvider] is returned in an [Arc] so appropriate cleanup can be done if necessary.
+pub fn set_global_telemetry_provider(
+    new_provider: Option<TelemetryProvider>,
```
aajtodd, line 23: If `TelemetryProvider::default()` is no-op or we offer `TelemetryProvider::noop()` (which we probably should anyway as it will be less ambiguous) then we don't need `Option<T>` here just take in the telemetry provider directly.

rust-runtime/aws-smithy-observability/src/provider.rs
```diff
@@ -0,0 +1,82 @@
...
+
+//! Definitions of high level Telemetry Providers.
+
+use std::sync::Arc;
+
+use crate::{meter::MeterProvider, noop::NoopMeterProvider};
+
+/// A struct to hold the various types of telemetry providers.
+#[non_exhaustive]
+pub struct TelemetryProvider {
```
aajtodd, line 14: Any reason to not make it a trait? What was your rationale for or against?

rust-runtime/aws-smithy-observability/src/provider.rs
```diff
@@ -0,0 +1,82 @@
...
+impl TelemetryProvider {
+    /// Returns a builder struct for [TelemetryProvider]
+    pub fn builder() -> TelemetryProviderBuilder {
+        TelemetryProviderBuilder {
+            meter_provider: Box::new(NoopMeterProvider),
+        }
+    }
+
+    /// Get the set [MeterProvider]
+    pub fn meter_provider(&self) -> &(dyn MeterProvider + Send + Sync) {
```
aajtodd, line 27: should this return `impl MeterProvider + Send + Sync`?

rust-runtime/aws-smithy-observability/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+    rustdoc::missing_crate_level_docs,
+    unreachable_pub,
+    rust_2018_idioms
+)]
+
+//! Smithy Observability
+//TODO(smithyobservability): once we have finalized everything and integrated metrics with our runtime
+// libraries update this with detailed usage docs and examples
+
+pub mod attributes;
```
aajtodd, line 20: Several of these modules _probably_ should be private and just re-export their type instead.

```rs
use aws_smithy_observability::Error;
use aws_smithy_observability::Attributes`
```

vs

```rs
use aws_smithy_observability::error::Error;
use aws_smithy_observability::attributes::Attributes`
```

In other words consider how you'd like the API to be exposed and consumed.

Also missing docs on public modules.

rust-runtime/aws-smithy-observability/src/attributes.rs
```diff
@@ -0,0 +1,83 @@
...
+impl Attributes {
+    /// Create a new empty instance of [Attributes].
+    pub fn new() -> Self {
+        Self {
+            attrs: HashMap::new(),
+        }
+    }
+
+    /// Set an attribute.
+    pub fn set(&mut self, key: String, value: AttributeValue) {
```
aajtodd, line 42: Should probably take more flexible types like `impl Into<String>` or `impl Into<AttributeValue>`

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,118 @@
...
+        description: Option<String>,
+    ) -> Box<dyn AsyncMeasurement<Value = u64>>;
+
+    /// Create a new [Histogram].
+    fn create_histogram(
+        &self,
+        name: String,
+        units: Option<String>,
+        description: Option<String>,
+    ) -> Box<dyn Histogram>;
```
aajtodd, line 82: We seem to be forcing `Box<dyn T>` everywhere and I'm wondering if this is the right tradeoff/choice. We probably want to be leveraging generics a bit more here.

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/// Collects a set of events with an event count and sum for all events.
+pub trait Histogram {
```
aajtodd, line 86: Do we need to consider whether we want to force (or need) instrument types to implement `Clone`?

aajtodd (changes requested):
Good start but I have some questions and concerns to be discussed.

ysaito1001 (commented):
Thanks for the work! Will circle back, leaving initial comments...

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/// A counter that can increase or decrease.
+pub trait UpDownCounter {
+    /// Increment or decrement a counter by a fixed amount.
+    fn add(&self, value: i64, attributes: Option<&Attributes>, context: Option<&dyn Context>);
+}
+
+/// A measurement that can be taken asynchronously.
+pub trait AsyncMeasurement {
```
ysaito1001, line 104: I acknowledge there are many traits introduced in this PR, so I don't know how strictly we should enforce the trait naming convention in the codebase, i.e. verb + noun. For instance, `AsyncMeasurement` is probably on the "easy" side for renaming: `AsyncMeasure`. For others like `Histogram`, it's probably harder to rename. Maybe it's not straightforward to make names of new traits consistent, but this is just bringing awareness.

rust-runtime/aws-smithy-observability-otel/src/meter.rs
```diff
@@ -0,0 +1,495 @@
...
+/// An OpenTelemetry based implementation of the AWS SDK's [MeterProvider] trait
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct AwsSdkOtelMeterProvider {
+    meter_provider: OtelSdkMeterProvider,
+}
+
+impl AwsSdkOtelMeterProvider {
+    /// Create a new [AwsSdkOtelMeterProvider] from an [OtelSdkMeterProvider].
+    pub fn new(otel_meter_provider: OtelSdkMeterProvider) -> Self {
```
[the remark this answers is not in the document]
landonxjames: Yeah, the intention here is that the `aws-smithy-observability-otel` crate will not GA before the actual OTel ecosystem does. They are tracking their path towards GA here: https://github.com/open-telemetry/opentelemetry-rust/issues/1678

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: I stuck with the names directly from the spec to align with other SDKs. I'm not strongly opposed to changing them, but we should discuss if breaking that alignment is worth it.

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: I don't think that would work. Returning `impl Meter` would make `MeterProvider` not object safe and so we couldn't pass it around as a `dyn` everywhere. 

From [docs](https://doc.rust-lang.org/reference/items/traits.html#object-safety):

- All associated functions must either be dispatchable from a trait object or be explicitly non-dispatchable:
  - Dispatchable functions must:
    - Not have an opaque return type; that is,
      - Not have a return position impl Trait type (fn example(&self) -> impl Trait).

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: I initially left these out of the trait and `impl`ed them on `AwsSdkOtelMeterProvider`. But at the end of your run when you do want to shutdown/flush the meter provider you first have to get it out of the `GlobalTelemetryProvider` and the only type information you get is `&dyn MetricProvider`. So to clear it you first have cast it to `Any`, then `downcast_ref` to the concrete type. All of that feels quite messy and not user friendly so I moved them to the actual trait. 

It seems likely that most metric solutions will provide some way to clear their buffers/shutdown gracefully (see the [clear method on buckets](https://docs.rs/metrics-util/0.5.0/metrics_util/struct.AtomicBucket.html#method.clear) in metrics.rs). 

So the addition of these turns this:
```rs
let otel_tp = set_global_telemetry_provider(None);
let foo = otel_tp
      .meter_provider()
      .as_any()
      .downcast_ref::<AwsSdkOtelMeterProvider>()
      .unwrap();
foo.shutdown().unwrap();
```

into:
```rs
let otel_tp = set_global_telemetry_provider(None);
otel_tp.meter_provider().shutdown().unwrap();
```

I'm not strongly opposed to removing them, but I opted for the more user friendly API to start.

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: See response above, using `impl T` would make these not object safe
https://github.com/smithy-lang/smithy-rs/pull/3912#discussion_r1841211316

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: I tried that, but once again it makes the `Meter` trait no longer object safe

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: `Sized` is a supertrait of `Clone` and `Sized` supertraits [break object safety](https://doc.rust-lang.org/std/clone/trait.Clone.html)

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+//! Metrics are used to gain insight into the operational performance and health of a system in
+//! real time.
+
+use crate::attributes::{Attributes, Context};
+use crate::error::ObservabilityError;
+
+/// Provides named instances of [Meter].
+pub trait MeterProvider {
```
ysaito1001, line 13: Traits in heavily used runtime components typically have super traits like `Send + Sync + Debug` (ex. [HttpClient](https://github.com/smithy-lang/smithy-rs/blob/8cf9ebddd8cf6793e80a52e9b7c883cc3c1e89d6/rust-runtime/aws-smithy-runtime-api/src/client/http.rs#L135)) to make consuming types be sent & shared among threads. Curious to know about new traits introduced in this PR in that regard 🤔

rust-runtime/aws-smithy-observability/src/meter.rs
ysaito1001: For that, we've already broken alignments for the trait names in the SRA because the doc was mostly written with Java in mind (ex. the doc specifies `CredentialsProvider` as an interface but we have the `ProvideCredentials` trait to align with the verb + noun Rust ecosystem).

Given [John worked on a PR](https://github.com/smithy-lang/smithy-rs/pull/3065) just to rename trait names, personally I'd like to carry that spirit. If it's ok, let's make an effort to rename the obvious ones like `AsyncMeasure`, `MeterProvider` to verb + none to make them align with the codebase as much as possible. There are always exceptions like `HttpClient` trait in the existing code, and for them we can just leave them as they are (such as `Histogram`, `Meter`).

ysaito1001:
I see that the design specifies an async API such as `create_guage` returns an async measurement handle. Do we need to support that, not necessarily in this PR but before the `aws-smithy-observability` goes 1.0 (in which case, like many traits do in the smithy runtime crates, a trait needs to return a struct and that struct implements the `Future` trait)?

rust-runtime/aws-smithy-observability-otel/src/attributes.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+impl From<AttributesWrap> for Vec<KeyValue> {
+    fn from(value: AttributesWrap) -> Self {
+        value
+            .attributes()
+            .iter()
+            .map(|(k, v)| {
+                KeyValue::new(
+                    k.clone(),
```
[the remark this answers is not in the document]
landonxjames: Good suggestion! Added in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/71cd2efa666cb8e949ca18f8ff409ee14649f1c7

rust-runtime/aws-smithy-observability/README.md
landonxjames: Fixed in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/0559e8b54fb41bcf04125e2e3cedeec87b24a9f9

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Removed in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/0559e8b54fb41bcf04125e2e3cedeec87b24a9f9

rust-runtime/aws-smithy-observability/src/provider.rs
landonxjames: Done in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/0559e8b54fb41bcf04125e2e3cedeec87b24a9f9

rust-runtime/aws-smithy-observability/src/global.rs
landonxjames: Done in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/0559e8b54fb41bcf04125e2e3cedeec87b24a9f9

rust-runtime/aws-smithy-observability/src/global.rs
landonxjames: Removed this behavior in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/0559e8b54fb41bcf04125e2e3cedeec87b24a9f9

rust-runtime/aws-smithy-observability/src/global.rs
landonxjames: Added an explicit `noop` method for `TelemetryProvider` in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/0559e8b54fb41bcf04125e2e3cedeec87b24a9f9

rust-runtime/aws-smithy-observability/src/lib.rs
landonxjames: Added some `pub use`s for these types in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/0559e8b54fb41bcf04125e2e3cedeec87b24a9f9

The public modules are already documented, but the docs are in the module file instead of in `lib.rs`. I prefer keeping them there because I think it is cleaner than putting them all in `lib.rs`, but will move them if that is a consensus.

rust-runtime/aws-smithy-observability/src/attributes.rs
landonxjames: Done in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/0559e8b54fb41bcf04125e2e3cedeec87b24a9f9

rust-runtime/aws-smithy-observability/src/provider.rs
landonxjames: Mostly because it is not specified in the SRA and so isn't really an interface we would expect users to be implementing themselves. Since the `TelemetryProvider` just exists to store and return references to the `MeterProvider` (and eventually maybe a `TracerProvider` and `LogProvider`) it doesn't seem like something a user would need to implement any custom functionality on.

rust-runtime/aws-smithy-observability/src/provider.rs
landonxjames: I don't think you can return a reference to an `impl Trait`, and I didn't want to hand ownership of the `MeterProvider` to the caller, so settled for the `&dyn` that is there now.

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Ack, I'll rearrange these to follow our grammatical convention in another commit since we have already committed to it with other SRA interfaces.

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Good call, initially I was going to hold off on adding these and add them as-needed just to minimize what the user might need to implement. But looking at your example and others it seems we tend to be fairly proactive with those supertraits so I added them in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/fe50a9dd499faa70ec4be9ad3bcc43bd13785b4c

ysaito1001 (approved):
Thanks for the great work! Left the final round of comments.

rust-runtime/aws-smithy-observability/src/attributes.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn into_attributes(self) -> impl Iterator<Item = (String, AttributeValue)> {
+        self.attrs.into_iter()
+    }
+}
+
+impl Default for Attributes {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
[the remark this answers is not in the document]
landonxjames: Addressed in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/4e11e53a2360c050313d71c067143d8db8ac873a

rust-runtime/aws-smithy-observability/src/global.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Get an [Arc] reference to the current global [TelemetryProvider].
+///
+/// This can panic if called when another thread is calling [set_telemetry_provider].
+pub fn get_telemetry_provider() -> Arc<TelemetryProvider> {
+    // TODO(smithyObservability): would probably be nicer to return a Result here, but the Guard held by the error from
+    // .try_read is not Send so I struggled to build an ObservabilityError from it
+    GLOBAL_TELEMETRY_PROVIDER
+        .try_read()
+        .expect("GLOBAL_TELEMETRY_PROVIDER RwLock Poisoned")
```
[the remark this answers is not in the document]
landonxjames: I'm not terribly worried about `get_telemetry_provider` panicing. `RwLock` allows multiple readers, so `try_read` only errors if:

1. A write/exclusive lock is already held (so in our case if someone was concurrently calling `set_telemetry_provider`)
2. The lock is poisoned (which only happens if a panic occurs while and write/exclusive lock is held)

But I don't like having the panic in there anyway so I implemented your suggestion in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/4e11e53a2360c050313d71c067143d8db8ac873a

Do you think `set_telemetry_provider` should return a `Some<()>` to mirror this? I do want to spend some more time trying to make a reasonable `Error` out of the `TryLockResult` type before we GA this.

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Moved to `Arc` in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/4e11e53a2360c050313d71c067143d8db8ac873a

rust-runtime/aws-smithy-observability/src/global.rs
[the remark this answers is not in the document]
ysaito1001: > Do you think set_telemetry_provider should return a Some<()> to mirror this?

Good question. `set_telemetry_provider` might not lead to a panic, as I assume we have full control over when it's called—such as during runtime initialization or when configuring default runtime plugins in the SDK?

> I do want to spend some more time trying to make a reasonable Error out of the TryLockResult type before we GA this.

Good idea. If we decide to return a `Result`, we probably don't need to propagate the exact error from `TryLockResult`. Instead, we can match on it, map it to an error type that we control (one that implements `Send`), and return that from the function.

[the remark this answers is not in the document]
landonxjames: This PR has been taking longer to merge than anticipated so I went ahead and refactored the `get/set_telemetry_provider` functions to return `Result<>` in https://github.com/smithy-lang/smithy-rs/pull/3912/commits/b96423135773d437e94e93f0f21e9cf1e579ab89

landonxjames:
Updated in https://github.com/smithy-lang/smithy-rs/pull/3986 to an Associated Type based implementation instead of the `dyn Trait` one featured here.

PR #3914, opened 2024-11-15
refactor hyper1 client builder to allow multiple TLS providers and add new config

aajtodd:
## Motivation and Context
Yet another follow up to: https://github.com/smithy-lang/smithy-rs/pull/3866

## Description
* Refactor the hyper 1 implementation into multiple modules to make it (slightly) easier to navigate (separating TLS, DNS, and timeout concerns into new/separate modules)
* Refactors the hyper 1 implementation to allow for additional TLS provider implementations (e.g. s2n). Before we just exposed a `CryptoMode` but that is only applicable to rustls. I've swapped this for a new `TlsProvider` concept
* Refactored how connectors are built. Before we were taking a TLS connector and wrapping it (via `Connector`). But this is structurally awkward since our `Connector` is analogous to hyper_utils' [`HttpConnector`](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/connect/struct.HttpConnector.html#) (or supposed to be anyway). The updated structure forces our `Connector` to build the `HttpConnector` (which is the lowest layer/TCP connector) and then (if enabled) wrap it in TLS, and then finally wrap it with the SDK timeouts. 
```
// build function returns a connector that looks like this as far as layering goes:
Connector<TlsProviderSelected>.build() -> SdkTimeouts(TlsConnector(HttpConnector))
```

* Exposed additional builder setting for `TCP_NODELAY` (and defaulted it to true).
* Exposed additional builder setting for binding to a specific NIC
* Fixed feature flags (powerset build was failing locally due to invalid or misconfigured feature flag application)
* Renamed TLS feature flags to be associated with the underlying TLS lib (e.g. `crypto-aws-lc` -> `rustls-aws-lc`)

**NOTE**:  Apologies in advance, the diff didn't quite come out as I wanted. `client.rs` replaced `hyper_1.rs` (several modules of `hyper_1.rs` were split out into new modules). Everything in `timeout.rs` and `dns.rs` did not change.

----


rust-runtime/aws-smithy-http-client/src/lib.rs
```diff
@@ -20,24 +21,61 @@
...
+pub use client::{default_client, tls, Builder, Connector, ConnectorBuilder};
 
 #[cfg(feature = "test-util")]
 pub mod test_util;
+
+#[allow(unused_macros, unused_imports)]
+#[macro_use]
+pub(crate) mod cfg {
+    /// Any TLS provider enabled
+    macro_rules! cfg_tls {
```
ysaito1001, line 51: Is this intentionally the same as `cfg_rustls!`?

ysaito1001 (approved):
Nicely refactored!

rust-runtime/aws-smithy-http-client/src/lib.rs
aajtodd: Right now it is yes. `cfg_tls` is meant for _any_ TLS provider enabled (with the expectation we'll be adding at least opt-in support for S2n in the near future) whereas `cfg_rustls` is targeted to if rustls is enabled.

rust-runtime/aws-smithy-http-client/src/client.rs
```diff
@@ -0,0 +1,945 @@
...
+}
+
+/// Builder for [`Connector`].
+#[derive(Default, Debug)]
+pub struct ConnectorBuilder<Tls = TlsUnset> {
+    connector_settings: Option<HttpConnectorSettings>,
+    sleep_impl: Option<SharedAsyncSleep>,
+    client_builder: Option<hyper_util::client::legacy::Builder>,
+    enable_tcp_nodelay: bool,
+    interface: Option<String>,
```
Velfi, line 102: What is this?

landonxjames (approved):

PR #3915, opened 2024-11-19
Add operation customization for disabling payload signing

Velfi:
## Motivation and Context

https://github.com/smithy-lang/smithy-rs/pull/3583

## Description

This PR adds the ability for users to disable payload signing with an operation customization.

## Testing

This PR includes tests.

## Checklist

----


aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
```diff
@@ -190,11 +190,8 @@ fn add_checksum_for_request_body(
         // Body is streaming: wrap the body so it will emit a checksum as a trailer.
         None => {
             tracing::debug!("applying {checksum_algorithm:?} of the request body as a trailer");
-            if let Some(mut signing_config) = cfg.load::<SigV4OperationSigningConfig>().cloned() {
-                signing_config.signing_options.payload_override =
-                    Some(SignableBody::StreamingUnsignedPayloadTrailer);
-                cfg.interceptor_state().store_put(signing_config);
-            }
+            cfg.interceptor_state()
```
aajtodd, line 193: Why does this not need gated anymore?

aws/sdk/integration-tests/s3/tests/signing-it.rs
```diff
@@ -40,3 +42,78 @@ async fn test_signer() {
...
+    let request = request.expect_request();
+    let x_amz_content_sha256 = request
+        .headers()
+        .get("x-amz-content-sha256")
+        .expect("x-amz-content-sha256 is set")
+        .to_owned();
+    assert_eq!("UNSIGNED-PAYLOAD", x_amz_content_sha256);
+}
+
+// This test ensures that the interceptor's payload signing setting
```
aajtodd, line 75: Which interceptor? Also maybe add context on why this is important.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt
```diff
@@ -65,6 +65,7 @@ val DECORATORS: List<ClientCodegenDecorator> =
             TokenProvidersDecorator(),
             ServiceEnvConfigDecorator(),
             HttpRequestCompressionDecorator(),
+            DisablePayloadSigningDecorator(),
```
aajtodd, line 68: **correctness**: Should we only be applying this to S3 (and possibly only specific S3 operations)? 

I'm guessing most operations are going to require signing and won't work if the payload is unsigned but you'd have to test it. I also can't imagine it matters much for most operations to need unsigned so I'd rather not support this for more than we need to.

aajtodd (commented):
Couple questions but overall looks good.

This looks congruent with the proposed [rfc](https://github.com/smithy-lang/smithy-rs/pull/3583) unless I'm mistaken. If theres no material differences we should update that and get it merged with this work.

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
Velfi: Previously this info was stored within something in the config bag. Now I don't need to load anything so I therefore don't need an `if let`.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt
Velfi: I'm going to store the allowlist in the plugin itself. Currently, it only targets two S3 operations.

aajtodd (approved):

.changelog/1732034799.md
```diff
@@ -0,0 +1,28 @@
+---
+applies_to: ["client"]
```
ysaito1001, line 2: nit: given code changes are mostly under `aws`, the release note should probably show up in a `aws-sdk-rust` release. This should be either `["client", "aws-sdk-rust"]` or just `["aws-sdk-rust"]`

PR #3917, opened 2024-11-19
Avoid logging signable body by default whose data can be very large

ysaito1001:
## Motivation and Context
While investigating a connect timeout issue for uploading object(s) in [`aws-s3-transfer-manager-rs`](https://github.com/awslabs/aws-s3-transfer-manager-rs), we saw that the size of trace log was about 70 GB and that the last 1 GB only had 30 lines, with each line having couple MB's body to be logged (due to [this location](https://github.com/awslabs/aws-sdk-rust/blob/953cd6c7af04f02938a0dcf36f793ebe7a06cc57/sdk/aws-sigv4/src/http_request/sign.rs#L224)).

## Description
This PR disables logging the actual body data in `SignableBody` by default. Customers can set the `LOG_SIGNABLE_BODY` environment variable to log the body data if they want to, as described in the comment within the `Debug` implementation.

## Testing
- Added a small unit test
- Tests in CI

----

### DIFF
--- a/aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
+++ b/aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
@@ -95,7 +95,7 @@ pub enum SignableBody<'a> {
     StreamingUnsignedPayloadTrailer,
 }
 
-/// Formats the value using the given formatter. To print the body data, set `LOG_SIGNABLE_BODY=true`.
+/// Formats the value using the given formatter. To print the body data, set the environment variable `LOG_SIGNABLE_BODY=true`.
 impl<'a> Debug for SignableBody<'a> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         let should_log_signable_body = std::env::var(LOG_SIGNABLE_BODY)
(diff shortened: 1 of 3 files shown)


aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
```diff
@@ -93,6 +95,35 @@ pub enum SignableBody<'a> {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let should_log_signable_body = std::env::var(LOG_SIGNABLE_BODY)
+            .map(|v| v.eq_ignore_ascii_case("true"))
+            .unwrap_or_default();
+        match self {
+            Self::Bytes(arg0) => {
+                if should_log_signable_body {
+                    f.debug_tuple("Bytes").field(arg0).finish()
+                } else {
+                    f.debug_tuple("Bytes").finish()
```
landonxjames, line 109: Should we log something like `"body redacted"` or a pointer to the `LOG_SIGNABLE_BODY` env var in place of the actual value?

landonxjames (approved):
Good catch!

aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
ysaito1001: Interestingly, that's what a function [log_sensitive_body](https://github.com/smithy-lang/smithy-rs/blob/df77d5ff33f143abaea3f8bf5a9cdeec05353ece/rust-runtime/aws-smithy-runtime/src/client/orchestrator/http.rs#L50-L51) does in the trace log message. Probably a pointer to `LOG_SIGNABLE_BODY` would be preferred in this case. Will make that adjustment.

aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
```diff
@@ -93,6 +95,39 @@ pub enum SignableBody<'a> {
...
+        let should_log_signable_body = std::env::var(LOG_SIGNABLE_BODY)
+            .map(|v| v.eq_ignore_ascii_case("true"))
+            .unwrap_or_default();
+        match self {
+            Self::Bytes(arg0) => {
+                if should_log_signable_body {
+                    f.debug_tuple("Bytes").field(arg0).finish()
+                } else {
+                    f.debug_tuple("Bytes")
+                        .field(&"** REDACTED **. To print, set LOG_SIGNABLE_BODY=true")
```
aajtodd, line 110: Suggest logging the total size of the body if possible

Suggest: `set environment variable LOG_SIGNABLE_BODY=true`

aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
```diff
@@ -93,6 +95,39 @@ pub enum SignableBody<'a> {
...
+                        .finish()
+                }
+            }
+            Self::UnsignedPayload => write!(f, "UnsignedPayload"),
+            Self::Precomputed(arg0) => {
```
aajtodd, line 115: I think you can log precomputed as the value should just be a checksum.

aajtodd (approved):
Some small suggestions but looks good.

aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
ysaito1001: Addressed in bba1a52

aws/rust-runtime/aws-sigv4/src/http_request/sign.rs
ysaito1001: Good suggestions, incorporated in bba1a52

PR #3918, opened 2024-11-20
Add codegen helper for defining operation-specific `CustomizableOperation` methods.

Velfi:
This PR lays the groundwork for defining operation-specific `CustomizableOperation` methods. 

## Codegen Example

If we were to define pre-signable ops this way, code like the following would be emitted for each op that supported presigning.

```rust
impl<E, B> CustomizableOperation<crate::operation::put_object::PutObject, E, B> {
    /// Sends the request and returns the response.
    #[allow(unused_mut)]
    pub async fn presigned(
        mut self,
        presigning_config: crate::presigning::PresigningConfig,
    ) -> ::std::result::Result<crate::presigning::PresignedRequest, crate::error::SdkError<E>>
    where
        E: std::error::Error + ::std::marker::Send + ::std::marker::Sync + 'static,
        B: crate::client::customize::internal::CustomizablePresigned<E>,
    {
        self.execute(move |sender, conf| sender.presign(conf, presigning_config)).await
    }
}
```

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/Customization.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/Customization.kt
@@ -80,3 +80,13 @@ fun <T : Section> RustWriter.writeCustomizationsOrElse(
         orElse(this)
     }
 }
+
+fun <T : Section> allCustomizationsAreEmpty(
+    customizations: List<NamedCustomization<T>>,
+    section: T,
+): Boolean {
+    val test = RustWriter.root()
+    test.writeCustomizations(customizations, section)
+    // If they're not dirty, then they're empty.
+    return !test.dirty()
+}
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
+version = "0.2.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
+checksum = "45862d1c77f2228b9e10bc609d5bc203d86ebc9b87ad8d5d5167a6c9abf739d9"
 
 [[package]]
 name = "anes"
@@ -69,7 +69,7 @@ checksum = "721cae7de5c34fbb2acd27e21e6d2cf7b886dce0c27388d46c4e6c47ea4318dd"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.87",
+ "syn 2.0.89",
 ]
 
 [[package]]
@@ -138,7 +138,7 @@ dependencies = [
 
 [[package]]
 name = "aws-runtime"
-version = "1.4.3"
+version = "1.4.4"
 dependencies = [
  "arbitrary",
  "aws-credential-types",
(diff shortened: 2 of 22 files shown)


ysaito1001:
Thanks for the PR. This gives us a good starting point for discussion.

> We've been generating the presign method for all S3 ops, even though it's only guaranteed to work for a subset. Before merging this, we'll probably want to undo the changes to presigning while keeping the underlying new way of defining customizable op impls.

Talked about this briefly offline, and it might be better to keep the presign method for all operations so we don’t break things for people. There's also [an issue](https://github.com/awslabs/aws-sdk-rust/issues/488) (linked off from [this TODO](https://github.com/smithy-lang/smithy-rs/blob/aac9becfd469e2479f68c61fd4c8074ddf755482/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsPresigningDecorator.kt#L74)) that says `All S3 requests should be presignable—this is the behavior of the Ruby SDK.` so that's something to keep in mind.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/Customization.kt
```diff
@@ -80,3 +80,13 @@ fun <T : Section> RustWriter.writeCustomizationsOrElse(
         orElse(this)
     }
 }
+
+fun <T : Section> allCustomizationsAreEmpty(
```
aajtodd, line 84: Why is this necessary? If `customizations` is empty then how would `writeCustomizations` end up with a "dirty" writer?

Velfi: Great question!

How can you tell if the following function will write when called through `RustWriter.writeCustomizations`?

```kotlin
    override fun operationCustomizations(
        codegenContext: ClientCodegenContext,
        operation: OperationShape,
        baseCustomizations: List<OperationCustomization>,
    ): List<OperationCustomization> {
        return baseCustomizations +
            object : OperationCustomization() {
                override fun section(section: OperationSection): Writable {
                    return writable {
                        when (section) {
                            is OperationSection.CustomizableOperationImpl -> {
                                    rust("// Writing occurred!")
                                }
                            }

                            else -> {}
                        }
                    }
                }
            }
    }
```

It just returns a list of customizations with method implementations. You can't know anything about the implementations, including what sections they actually care about. There's no way to know if this will write anything without actually calling it and trying to write something.

So why is it important to know if something was written? It's important in cases where you want to avoid empty wrappers. Imagine that you want to generate a block of code with functions for converting a type. If you always have `n+1` customizations, then it's always OK to write the enclosing block. But if you only have `n` customizations then you'll write an empty block and clippy will be mad at you.

Enter `allCustomizationsAreEmpty`. With this function, we can see if anything would actually be written, giving us the power to avoid writing an enclosing block when it would be empty.

Does that make sense?

aajtodd: It does, one of my concerns was triggering unrelated/additional side effects (e.g. imports that trigger dependencies being added, etc). We saw this in Kotlin though for a similar but slightly different reason.

Would be nice if we had a cleaner way to do this but I suppose it's fine for now.

aajtodd (approved):

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/Customization.kt
ysaito1001: Maybe adding source code comment with an example Rust snippet that this function is trying to solve is useful?

PR #3920, opened 2024-11-26
Add missing lifetime annotation to `EscapedStr::as_escaped_str`

drganjoo:
`JsonTokenIterator` parses and returns `Token`s from a provided u8 slice. `EscapedStr` is used as a member field in several token variants, such as `ObjectKey` and `ValueString`. While `as_escaped_str` returns a reference to the underlying `&'a str`, the lifetime annotation was missing from its return type.

This PR adds the missing lifetime annotation.

### DIFF
--- a/rust-runtime/aws-smithy-json/Cargo.toml
+++ b/rust-runtime/aws-smithy-json/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-json"
-version = "0.60.7"
+version = "0.61.1"
 authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "John DiSanti <jdisanti@amazon.com>"]
 description = "Token streaming JSON parser for smithy-rs."
 edition = "2021"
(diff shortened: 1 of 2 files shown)


david-perez (approved):
Right, so `EscapedStr` is valid only if the inner `&str` is valid, so, `'a` must outlive `'1`, i.e. `'1 < 'a`. According to [lifetime elision rules](https://doc.rust-lang.org/reference/lifetime-elision.html#lifetime-elision-in-functions):

```rust
pub fn as_escaped_str(&self) -> &str
```

is the same thing as

```rust
pub fn as_escaped_str(&'1 self) -> &'1 str
```

So we were giving out a _shorter_ lifetime than `'a`.

But we might want to use the inner `&str` after `EscapedStr` is dropped, so this code only compiles if we give out `'a`:

```rust
struct EscapedStr<'a>(&'a str);

impl<'a> EscapedStr<'a> {
    // Doesn't work, would work if we give out `'a`.
    pub fn as_escaped_str(&self) -> &str {
        self.0
    }
}

fn main() {
    let foo = "".to_owned();
    let escaped = EscapedStr(&foo);

    let as_escaped = escaped.as_escaped_str();

    drop(escaped);

    println!("{}", as_escaped);
}
```

PR #3921, opened 2024-11-26
Add feature ID for S3 Transfer Manager

ysaito1001:
## Motivation and Context
This PR introduces the `AwsSdkFeature` enum, which will be used for AWS SDK-specific feature identifiers. For now, it includes a single ID for the S3 Transfer Manager. This approach is similar to the existing [SmithySdkFeature](https://github.com/smithy-lang/smithy-rs/blob/aac9becfd469e2479f68c61fd4c8074ddf755482/rust-runtime/aws-smithy-runtime/src/client/sdk_feature.rs#L10) in terms of its module and crate structure (that is for generic client but `AwsSdkFeature` is AWS specific); the `sdk_feature` module is hidden as it is intended for internal use only.

## Testing
Added an integration test to verify tracking a business metric for Transfer Manager. Since Transfer Manager is a high-level library, we do not track its business metric directly within the `smithy-rs` repository. Instead, [an s3 client](https://github.com/awslabs/aws-s3-transfer-manager-rs/blob/main/aws-s3-transfer-manager/src/config.rs#L21) used in [aws-s3-transfer-manager-rs](https://github.com/awslabs/aws-s3-transfer-manager-rs) can configure the s3 client as demonstrated in the integration test to track the metric.

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/lib.rs
+++ b/aws/rust-runtime/aws-runtime/src/lib.rs
@@ -38,6 +38,7 @@ pub mod invocation_id;
 /// Supporting code for request metadata headers in the AWS SDK.
 pub mod request_info;
 
+/// AWS SDK feature identifies.
 #[doc(hidden)]
 pub mod sdk_feature;
 
--- a/aws/rust-runtime/aws-runtime/src/lib.rs
+++ b/aws/rust-runtime/aws-runtime/src/lib.rs
@@ -38,6 +38,9 @@ pub mod invocation_id;
 /// Supporting code for request metadata headers in the AWS SDK.
 pub mod request_info;
 
+#[doc(hidden)]
+pub mod sdk_feature;
+
 /// Interceptor that determines the clock skew between the client and service.
 pub mod service_clock_skew;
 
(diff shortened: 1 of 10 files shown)


aws/rust-runtime/aws-runtime/src/lib.rs
```diff
@@ -38,6 +38,9 @@ pub mod invocation_id;
 /// Supporting code for request metadata headers in the AWS SDK.
 pub mod request_info;
 
+#[doc(hidden)]
+pub mod sdk_feature;
+
```
Velfi, line 43: Though hidden, I still think we should document this for ourselves.

aws/rust-runtime/aws-runtime/src/sdk_feature.rs
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use aws_smithy_types::config_bag::{Storable, StoreAppend};
+
+/// IDs for the features that may be used in the AWS SDK
+#[non_exhaustive]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum AwsSdkFeature {
+    /// Indicates that an operation called using the S3 Transfer Manager
```
Velfi, line 12:
```suggestion
    /// Indicates that an operation was called by the S3 Transfer Manager
```

Velfi (approved):

aws/rust-runtime/aws-runtime/src/lib.rs
ysaito1001: Makes sense. Will do the same for the `sdk_feature` module in the `aws-smithy-runtime` crate.

landonxjames (approved):

PR #3934, opened 2024-12-04
 Only support T as an allowed separator in RFC-3339 dates

drganjoo:
This PR modifies `aws-smithy-types::DateTime` to enforce the use of 'T' (or 't') as the date-time separator. A [recent merge](https://github.com/time-rs/time/pull/700/files#diff-78a73b9911c88961601784e674a20c89e7050de9f8763522bbaa88ed28311b51R524) in the 'time' crate introduced support for space characters as separators. While ISO 8601 does allow spaces, RFC 3339 strictly requires 'T' (or 't') as the separator. 

This change restores the behavior to match how it was before `time = 0.3.37`.

This PR also removes the temporary pinning of the time crate in protocol tests.

### DIFF
--- a/rust-runtime/aws-smithy-types/src/date_time/format.rs
+++ b/rust-runtime/aws-smithy-types/src/date_time/format.rs
@@ -407,6 +407,12 @@ pub(crate) mod rfc3339 {
             )
             .into());
         }
+        if s.len() > 10 && !matches!(s.as_bytes()[10], b'T' | b't') {
+            return Err(DateTimeParseErrorKind::Invalid(
+                "RFC-3339 only allows `T` as a separator for date-time values".into(),
+            )
+            .into());
+        }
         let date_time = OffsetDateTime::parse(s, &Rfc3339).map_err(|err| {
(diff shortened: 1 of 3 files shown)


rust-runtime/aws-smithy-types/src/date_time/format.rs
```diff
@@ -773,6 +779,43 @@ mod tests {
...
 
+    #[test]
+    fn parse_rfc3339_invalid_separator() {
+        let test_cases = [
+            ("1985-04-12 23:20:50Z", AllowOffsets::OffsetsForbidden),
+            ("1985-04-12x23:20:50Z", AllowOffsets::OffsetsForbidden),
+            ("1985-04-12 23:20:50-02:00", AllowOffsets::OffsetsAllowed),
+            ("1985-04-12a23:20:50-02:00", AllowOffsets::OffsetsAllowed),
+        ];
+        for (date, offset) in test_cases.into_iter() {
```
david-perez, line 790: nit: these `into_iter` are unnecessary

david-perez (approved):

ysaito1001 (approved):

PR #3939, opened 2024-12-05
enable hyper1 as default http client

aajtodd:
## Motivation and Context

## Description
* Add a new `default-http-connector` feature to `aws-smithy-runtime` that enables hyper 1.x with rustls+aws-lc as the default HTTP client plugin.
    * If both the legacy `connector-hyper-014-x` feature is enabled the new feature takes precedence such that the default HTTP client plugin favors hyper 1 if both features are enabled.
* Update codegen to enable the new `default-http-connector` feature instead of hyper 0.14
    * NOTE: we can't break the existing `rustls` feature flag so it has instead been updated to enable the `default-http-connector` feature.
    * Importantly this still uses rustls as the TLS provider but the crypto provider now defaults to aws-lc (which is coincidentally what rustls now defaults to as well). This is likely to break _someone_ somewhere due to build requirements changing (e.g. CMake now be required on certain platforms or other build tools needed).
* Updated `aws-config` to default to hyper1 for the default credentials chain.
    * We have two features in `aws-config`, `rustls` and `client-hyper`, that seem to be used entirely to enable the `aws-smithy-runtime` features `tls-rustls` and `connector-hyper-0-14-x` features (both are required for the default HTTP client plugin to work prior to this PR). I think the _intent_ behind these features was "the user is not providing an HTTP client explicitly so we need a default one to be available". I've added a new feature to `aws-config` for this, `default-http-connector` and updated the existing features to be synonyms. We don't use `rustls` or `hyper 0.14.x` directly in `aws-config` so I think this is a more clearly defined flag and conveys it's intent better.
* Added a new `legacy-test-util` feature flag to `aws-smithy-runtime`.  The rationale for this is that the `test-util` feature provides testing utilities for other things from `aws-smithy-runtime` but it also brings in the (now) legacy hyper 0.14 HTTP testing facilities. I've left `test-util` to mean what it does today and be backwards compatibile (for now anyway) and  in future we can ship a (breaking) change to disable the legacy test utils by default (and by extension stop compiling the legacy hyper ecosystem in all of our tests)
* Fixed an issue in `examples/pokemon-service-common` due to codegen no longer generating clients with the `aws-smithy-runtime/tls-rustls` feature enabled by default (they are using the `HyperClientBuilder::build_https()` method directly but relying on feature unification to enable the method)

## Questions

* Bikeshed any feature flag names. e.g. `aws-config/default-http-connector` could be more generic like `default-providers` or something. Today we use it to mean "we need a default HTTP client" but you can imagine a future where we need other default runtime components to exist and be configured. Perhaps that is what we want from this flag?
 

----


tools/ci-scripts/check-aws-config
```diff
@@ -42,11 +42,11 @@ cargo hack check --feature-powerset --exclude-all-features --exclude-features al
 
 echo "${C_YELLOW}## Testing each feature in isolation${C_RESET}"
 # these features are missed by the following check because they are grouped
-cargo hack test --each-feature --include-features rt-tokio,client-hyper,rustls --exclude-no-default-features
+cargo hack test --each-feature --include-features rt-tokio,client-hyper,rustls,default-http-connector --exclude-no-default-features
 
 # This check will check other individual features and no-default-features
 # grouping these features because they don't interact
-cargo hack test --feature-powerset --exclude-all-features --exclude-features allow-compilation --group-features rt-tokio,client-hyper,rustls
+cargo hack test --feature-powerset --exclude-all-features --exclude-features allow-compilation --group-features rt-tokio,client-hyper,rustls,default-http-connector
```
ysaito1001, line 49: Ah good, `default-http-connector` does not increase the number of combinations, as it can be part of the existing feature group.

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -57,6 +57,10 @@ pub fn default_http_client_plugin() -> Option<SharedRuntimePlugin> {
     #[cfg(feature = "connector-hyper-0-14-x")]
     let _default = crate::client::http::hyper_014::default_client();
 
+    // takes precedence over legacy connector if enabled
+    #[cfg(feature = "default-http-connector")]
+    let _default = aws_smithy_http_client::default_client();
```
ysaito1001, line 62: This is clever.

aws/rust-runtime/aws-config/src/imds/client.rs
```diff
@@ -984,7 +984,7 @@ pub(crate) mod test {
     #[cfg_attr(windows, ignore)]
     /// Verify that the end-to-end real client has a 1-second connect timeout
     #[tokio::test]
-    #[cfg(feature = "rustls")]
+    #[cfg(feature = "default-http-connector")]
```
ysaito1001, line 987: I believe it's ok to not include the `rustls` feature in tests unlike production code?

ysaito1001 (approved):
Great work updating various places extensively. Looks like need to get through some of CI checks but the changes look good structurally.

aws/rust-runtime/aws-config/src/imds/client.rs
aajtodd: Yeah I left it as `any(feature = "default-http-connector", feature = "rustls"))` on public code as I didn't want someone relying on `--no-default-features features=rustls` for example to break.

tools/ci-scripts/check-aws-config
aajtodd: We were in fact timing out as you suspected prior to this 😆

aajtodd: I've reduced the feature powerset for `aws-smithy-runtime` as well because it was timing out after 30 mins: https://github.com/smithy-lang/smithy-rs/pull/3939/commits/c50a3d2f06191f9430b4a37c503cd7c49ebebf11

Let me know if you see anything wrong with this but my thought process is we have several features that exist to keep things working for a legacy ecosystem we are moving from. Most of these flags are now controlling re-exports. Not testing them combinatorially is probably good/ok going forward.

Velfi (approved):

PR #3942, opened 2024-12-16
pin smithy-rs-tool-common with lockfile

aajtodd:
## Description
Seeing preview build failures across the board. Looks like `home@0.5.11` was [released a few hours ago](https://github.com/rust-lang/cargo/pull/14939) and comes with a new MSRV of 1.81. We don't have a lockfile for `smithy-rs-tool-common` for some reason. This PR adds one.

```
~/sandbox/rs/smithy-rs/tools/ci-build/smithy-rs-tool-common on main ✗                                                                            [41d648dd6]  12:14
> cargo check
error: rustc 1.78.0 is not supported by the following package:
  home@0.5.11 requires rustc 1.81
Either upgrade rustc or select compatible dependency versions with
`cargo update <name>@<current-ver> --precise <compatible-ver>`
where `<compatible-ver>` is the latest version supporting rustc 1.78.0
```

Added a lockfile and downgraded `home`:

```sh
> cargo update home@0.5.11 --precise 0.5.9
```

## Testing
Build locally + CI.

## Questions 

Lockfiles do not appear to be updated automatically for anything under `ci-build` unless I'm missing something. This is probably fine as our use of lockfiles is intended to catch issues with consumers of `smithy-rs` and the SDK not our own internal tooling. 

----


landonxjames (approved):
Locks `home` to `0.5.9`, LGTM

PR #3960, opened 2025-01-02
Update docs for non aws-config config builders

landonxjames:
## Motivation and Context

Partially addresses https://github.com/awslabs/aws-sdk-rust/issues/1226, although we might want to look into going further and actually deprecating or warning on these builder methods. 

## Testing

Doc only update

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
@@ -399,6 +399,7 @@ class ServiceConfigGenerator(
         }
 
     fun render(writer: RustWriter) {
+        val configDocs = "Constructs a config builder. Note that a config created from this builder will not have the same safe defaults as one created by the [aws-config](https://crates.io/crates/aws-config) crate."
         writer.docs("Configuration for a $moduleUseName service client.\n")
         customizations.forEach {
             it.section(ServiceConfig.ConfigStructAdditionalDocs)(writer)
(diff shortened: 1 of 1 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
```diff
@@ -399,6 +399,7 @@ class ServiceConfigGenerator(
         }
 
     fun render(writer: RustWriter) {
+        val configDocs = "Constructs a config builder. Note that a config created from this builder will not have the same safe defaults as one created by the [aws-config](https://crates.io/crates/aws-config) crate."
```
aajtodd, line 402: This is fine but wonder if it's worth trying to make this a callout. 

I can't find a doc.rs example but I recall seeing this before and this [issue](https://github.com/rust-lang/rust/issues/79710#issuecomment-1868260420) seems to indicate it's possible.


```rs
/// <div class="warning">
///
/// Warning text here
///
/// </div>
```

IDK maybe it's not worth it?

aajtodd (approved):

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/config/ServiceConfigGenerator.kt
landonxjames: Updated in https://github.com/smithy-lang/smithy-rs/pull/3960/commits/4d722e15e12348b1cea28e8b4c8b238426364db7

Screenshot 2025-01-03 at 08 35 38

PR #3964, opened 2025-01-08
fix token bucket not being set for both standard and adaptive retry modes

aajtodd:
## Motivation and Context

https://github.com/awslabs/aws-sdk-rust/issues/1234

## Description

PR adds a new interceptor registered as part of the default retry plugin components that ensures a token bucket is _always_ present and available to the retry strategy. The buckets are partitioned off the retry partition (which defaults to the service name and is already set by the default plugin). We use a `static` variable in the runtime for this which means that token buckets can and will apply to every single client that uses the same retry partition. The implementation tries to avoid contention on this new global lock by only consulting it if the retry partition is overridden after client creation. 

For AWS SDK clients I've updated the default retry partition clients are created with to include the region when set. Now the default partition for a client will be `{service}-{region}` (e.g. `sts-us-west-2`) rather than just the service name (e.g. `sts`). This partitioning is a little more granular and closer to what we want/expect as failures in one region should not cause throttling to another (and vice versa for success in one should not increase available quota in another). 

I also updated the implementation to follow the SEP a little more literally/closely as far as structure which fixes some subtle bugs. State is updated in one place and we ensure that the token bucket is always consulted (before the token bucket could be skipped in the case of adaptive retries returning a delay and the adaptive rate limit was updated in multiple branches). 

## Checklist

- [x ] For changes to the smithy-rs codegen or runtime crates, I have created a changelog entry Markdown file in the `.changelog` directory, specifying "client," "server," or both in the `applies_to` key.
- [ x] For changes to the AWS SDK, generated SDK code, or SDK runtime crates, I have created a changelog entry Markdown file in the `.changelog` directory, specifying "aws-sdk-rust" in the `applies_to` key.

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
@@ -174,8 +174,8 @@ private class AwsFluentClientDocs(private val codegenContext: ClientCodegenConte
  */
 private class AwsFluentClientRetryPartition(private val codegenContext: ClientCodegenContext) : FluentClientCustomization() {
     override fun section(section: FluentClientSection): Writable {
-        return when (section) {
-            is FluentClientSection.BeforeBaseClientPluginSetup ->
+        return when {
+            section is FluentClientSection.BeforeBaseClientPluginSetup && usesRegion(codegenContext) -> {
                 writable {
                     rustTemplate(
                         """
@@ -188,6 +188,7 @@ private class AwsFluentClientRetryPartition(private val codegenContext: ClientCo
                         "Cow" to RuntimeType.Cow,
                     )
                 }
+            }
             else -> emptySection
         }
     }
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
@@ -24,6 +24,7 @@ import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
 import software.amazon.smithy.rust.codegen.core.rustlang.writable
 import software.amazon.smithy.rust.codegen.core.smithy.RuntimeConfig
 import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType.Companion.preludeScope
 import software.amazon.smithy.rust.codegen.core.smithy.RustCrate
 import software.amazon.smithy.rust.codegen.core.smithy.generators.LibRsCustomization
 import software.amazon.smithy.rust.codegen.core.smithy.generators.LibRsSection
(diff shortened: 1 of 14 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -166,3 +168,28 @@ private class AwsFluentClientDocs(private val codegenContext: ClientCodegenConte
...
+    override fun section(section: FluentClientSection): Writable {
+        return when {
+            section is FluentClientSection.BeforeBaseClientPluginSetup && usesRegion(codegenContext) -> {
+                writable {
+                    rustTemplate(
+                        """
+                        let default_retry_partition = match config.region() {
+                            Some(region) => #{Cow}::from(format!("{default_retry_partition}-{}", region)),
+                            None => #{Cow}::from(default_retry_partition),
+                        };
```
Velfi, line 185:
```suggestion
                        let default_retry_partition = match config.region() {
                            Some(region) => #{Cow}::from(format!("{default_retry_partition}-{region}")),
                            None => #{Cow}::from(default_retry_partition),
                        };
```

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/RegionDecorator.kt
```diff
@@ -223,3 +216,14 @@ class RegionProviderConfig(codegenContext: ClientCodegenContext) : ConfigCustomi
...
+/**
+ * Test if region is used and configured for a model (and available on a service client).
+ *
+ * Services that have an endpoint ruleset that references the SDK::Region built in, or
+ * that use SigV4, both need a configurable region.
+ */
+fun usesRegion(codegenContext: ClientCodegenContext) =
+    codegenContext.getBuiltIn(AwsBuiltIns.REGION) != null ||
+        ServiceIndex.of(codegenContext.model)
+            .getEffectiveAuthSchemes(codegenContext.serviceShape).containsKey(SigV4Trait.ID)
```
Velfi, line 229: Do we ever have to worry about models supporting sigv4a but not sigv4? How do retry partitions interact with Sigv4a regions (if at all)?

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/RetryPartitionTest.kt
```diff
@@ -0,0 +1,93 @@
...
+                        let client = $moduleName::Client::from_conf(client_config);
+
+                        let _ = client
+                            .some_operation()
+                            .send()
+                            .await
+                            .expect("success");
+
+                        let log_contents = logs_rx.contents();
+                        assert!(log_contents.contains("token bucket for RetryPartition { name: \"dontcare-us-west-2\" } added to config bag"));
```
Velfi, line 54: We should consider updating the orchestrator to note the retry partition used for each retry

rust-runtime/aws-smithy-runtime/tests/retries.rs
```diff
@@ -0,0 +1,204 @@
...
+        .build();
+
+    (op, attempts)
+}
+
+/// Test we exhaust the token bucket long before we exhaust max attempts
+///
+/// see [aws-sdk-rust#1234](https://github.com/awslabs/aws-sdk-rust/issues/1234)
+#[tokio::test]
+async fn token_bucket_exhausted_before_max_attempts() {
```
Velfi, line 141: I stil wish the SEP was clearer on why 50 attempts is the right number. Why 50? Why not 42 or 20 or another number?

Velfi (approved):

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/RetryPartitionTest.kt
aajtodd: We could, is there any advantage over logging it here?

ysaito1001 (approved):
LGTM, thank for straightening the impls of `should_attempt_retry` and `calculate_backoff`!

PR #3967, opened 2025-01-13
Flexible Checksums V2

landonxjames:
## Motivation and Context

Flexible checksums V2 will be launching this week

## Checklist

----


aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpChecksumTest.kt
```diff
@@ -163,6 +180,7 @@ internal class HttpChecksumTest {
                         use #{HttpRequest};
                         use #{UaAssert};
                         use #{UaExtract};
+                        //#{PresigningMarker};
```
ysaito1001, line 183: Maybe a leftover?

ysaito1001 (approved):
Excellent work, ship it 🚀

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
```diff
@@ -183,10 +204,17 @@ where
...
+            // If this is a presigned request and the user has not set a checksum we short circuit
+            if is_presigned_req && checksum_algorithm.is_none() {
+                return Ok(());
+            }
+
             // If a checksum override is set in the ConfigBag we use that instead (currently only used by S3Express)
-            // If we have made it this far without a checksum being set we set the default as Crc32
-            let checksum_algorithm = incorporate_custom_default(state.checksum_algorithm, cfg)
-                .unwrap_or(ChecksumAlgorithm::Crc32);
+            // If we have made it this far without a checksum being set we set the default (currently Crc32)
```
aajtodd, line 215: Currently implies we could change it in the future, would that be behind a behavior version or do we think we can just change it without any concern in the future?

aajtodd (approved):

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
ysaito1001: Had the same question in my mind and my interpretation was to use a behavior version in that case. Programmatically, this PR [adds [default]](https://github.com/smithy-lang/smithy-rs/pull/3967/files#diff-e3f7eaa49dc41c50b9018415da01aa10b4ecab87e10445bc9bf1255d48dd9ce1R40) to `ChecksumAlgorithm` but we can switch the default to a different enum variant without breaking since `aws-smithy-checksums` is an unstable crate.

landonxjames: The default is not specified by the SEP, so changing it is up to us. I don't think it necessarily would need a behavior version since all of this behavior is opaque to end users, the only difference they might notice is a slight performance change with the new default algorithm. Since any default change would be preceded by performance testing to ensure that the new algorithm is equally or more performant than the old one, I don't see it being an issue.

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpChecksumTest.kt
landonxjames: Actually not a leftover, but I couldn't figure out a better way to do this. Since the `RequestChecksumInterceptor` now relies on the `PresigningMarker` type from `presigning` inlineable I needed a way to bring that inlineable into the generated test crate. I was having issues applying the presigning trait to the test model from the SEP, and since we didn't really need the presigning functionality, just the `PresigningMarker` type including it here was the easiest solution I came up with. I'll add a comment explaining that.

.changelog/flexible-checksums.md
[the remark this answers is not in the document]
landonxjames: Good idea, I'll split them up, this would impact any generated client that uses the `httpChecksum` trait (I doubt there are any, but it is possible)

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
landonxjames: Counterpoint from a HN comment: I guess sometimes the "more performant" algorithm can also be less performant for customers in more constrained environments. https://news.ycombinator.com/item?id=42691607

I think we are still ok not. putting this behind a behavior version, but worth thinking about.

.changelog/flexible-checksums.md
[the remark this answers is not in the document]
landonxjames: Split up in https://github.com/smithy-lang/smithy-rs/pull/3967/commits/332a5695cda904da56b499772e76b689107fbb43

aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/HttpChecksumTest.kt
landonxjames: Added note about this in https://github.com/smithy-lang/smithy-rs/pull/3967/commits/332a5695cda904da56b499772e76b689107fbb43

PR #3973, opened 2025-01-17
enable hyper1 behind BMV

aajtodd:
## Motivation and Context

## Description

Follow up to https://github.com/smithy-lang/smithy-rs/pull/3939 to change how we enable hyper1 as the default.

* This approach adds a new cargo feature flag on generated SDK packages (`default-https-client`) and restores the `rustls` feature to mean what it did before. Both will remain enabled as `default` features for some undetermined amount of time with the intent to remove `rustls` as a default feature as a breaking change in the future.
* Renamed feature flags to be consistent. We settled on `default-https-client`. This new flag when enabled and the appropriate min BMV is set will result in the new HTTPS stack being the default for clients. Otherwise we fallback to the old behavior (still requires `rustls` and `connector-hyper-0-14` features enabled)

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
@@ -67,7 +67,13 @@ class AwsFluentClientDecorator : ClientCodegenDecorator {
         rustCrate.withModule(ClientRustModule.client) {
             AwsFluentClientExtensions(codegenContext, types).render(this)
         }
-        rustCrate.mergeFeature(Feature("rustls", default = true, listOf("aws-smithy-runtime/default-http-connector")))
+
+        // TODO(hyper1): disable rustls as a default feature in future release
+        // NOTE: We enable both rustls and default-https-client as default features. This keeps the legacy hyper+rustls
+        // stack working as is and lets BehaviorVersion control which client you get. In a future release we will
+        // break this and disable the rustls feature by default (and break old BMV versions w.r.t http client default).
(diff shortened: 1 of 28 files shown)


landonxjames (approved):
Overall LGTM. I still think it is probably worth adding a README entry to the generated SDKs (or maybe just `aws-config`?) explaining the interplay between the feature flags and the BMV, but that isn't a blocker for merging to the feature branch.

aajtodd:
> Overall LGTM. I still think it is probably worth adding a README entry to the generated SDKs (or maybe just `aws-config`?) explaining the interplay between the feature flags and the BMV, but that isn't a blocker for merging to the feature branch.

Agreed. I'm working on customer messaging as next step and will seek feedback from team of course.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
```diff
@@ -67,7 +67,13 @@ class AwsFluentClientDecorator : ClientCodegenDecorator {
         rustCrate.withModule(ClientRustModule.client) {
             AwsFluentClientExtensions(codegenContext, types).render(this)
         }
-        rustCrate.mergeFeature(Feature("rustls", default = true, listOf("aws-smithy-runtime/default-http-connector")))
+
+        // TODO(hyper1): disable rustls as a default feature in future release
+        // NOTE: We enable both rustls and default-https-client as default features. This keeps the legacy hyper+rustls
+        // stack working as is and lets BehaviorVersion control which client you get. In a future release we will
+        // break this and disable the rustls feature by default (and break old BMV versions w.r.t http client default).
```
ysaito1001, line 74: Like the description 👍

ysaito1001 (approved):
LGTM!

PR #3975, opened 2025-01-20
IMDS retries

landonxjames:
## Motivation and Context

Addressing https://github.com/awslabs/aws-sdk-rust/issues/1233

## Description

Add ability to configure the retry classifier on the IMDS client.

## Testing

Added new unit test to ensure that a user defined retry classifer is being used. 

## Checklist

----


aajtodd (commented):
I realize this is still in draft so take feedback with grain of salt.

I'm wondering if we shouldn't expose the ability to customize the `ImdsResponseRetryClassifier` so that users don't have to start from scratch when creating a client.

e.g. (with some pseudo code)

```rs
let imds = aws_config::imds::client::builder()
      .with_retry_classifier(
           ImdsResponseRetryClassifer::default()
               .with_retry_connect_timeouts(true)   // new API(s) to customize from the default we use today
      )
      ...
      .build()
```

Or if we are super paranoid keep it hidden and just add new options to the client builder for this behavior but I think I'd be ok either way.

aajtodd (approved):

ysaito1001 (approved):

PR #3990, opened 2025-01-29
Fix flaky test in aws-sigv4

rcoh:
## Motivation and Context
The test was flaky because all invalid headers are counted as rejects which can in rare cases exceed the maximum number of failures.

The test was flaky because all invalid headers are counted as rejects which can in rare cases exceed the maximum number of failures.

## Description
Change the test to assert that the validity status matches instead of filtering those cases out

## Testing
- cargo test

----

### DIFF
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.21"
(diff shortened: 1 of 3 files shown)


ysaito1001 (commented):
Thanks for the fix! It looks like `aws/rust-runtime/Cargo.lock` being committed is a bit old with respect to the one in main, since many crates are being downgraded to older versions?

drganjoo (approved):

PR #3991, opened 2025-01-29
Exclude `transfer-encoding` header from sigv4(a) signing

landonxjames:
Add changelog

## Motivation and Context

Exclude `transfer-encoding` header from sigv4(a) signing. It is a hop by hop header and can be erased or modified by a proxy (in our particular case Cloudfront)

## Testing

Updated the existing tests for excluded headers.

## Checklist

----

### DIFF
--- a/.changelog/transfer-encoding-signing.md
+++ b/.changelog/transfer-encoding-signing.md
@@ -1,0 +1,10 @@
+---
+applies_to: ["aws-sdk-rust"]
+authors: ["landonxjames"]
+references: ["smithy-rs#3991"]
+breaking: false
+new_feature: false
+bug_fix: true
+---
+
+Exclude `transfer-encoding` header from sigv4(a) signing since it is a hop by hop header that can be modified or removed by a proxy.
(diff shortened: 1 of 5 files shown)


ysaito1001 (commented):
LGTM. Existing integration tests for S3 [capture the transfer-encoding header](https://github.com/search?q=repo%3Asmithy-lang%2Fsmithy-rs%20transfer-encoding&type=code) but I suspect they won't be broken since the way they verify expected results in tests is resilient to the changes in PR.

I am ok with leaving the `transfer-encoding` in connection recording tests. Maybe want to quickly check (e.g., by using `capture_request`) that the `transfer-encoding` header is indeed excluded from an outgoing request.

ysaito1001 (approved):

landonxjames:
>Existing integration tests for S3 [capture the transfer-encoding header](https://github.com/search?q=repo%3Asmithy-lang%2Fsmithy-rs%20transfer-encoding&type=code) but I suspect they won't be broken since the way they verify expected results in tests is resilient to the changes in PR.

Looks like all of those should be able to stay since all of them appear in a Response and this change only impacts  request signing. None of the `SignedHeaders` values reference `transfer-encoding`

PR #3996, opened 2025-02-04
Allow server SDKs to replace invalid UTF-8 character with '�'

drganjoo:
When the input contains invalid UTF-8 characters (such as surrogate pairs or invalid character ranges), these characters will be replaced with the replacement character "�" (\uFFFD). This replacement only occurs if the codegen flag 'replaceInvalidUtf8' is set to true in `smithy-build-template.json`.


landonxjames:
We were considering something like this for the client a few months ago because some customers were encountering non-utf-8 headers from S3: https://github.com/smithy-lang/smithy-rs/pull/3830. It broke some of our protocol tests and we decided not to follow it up because there wasn't consistent cross-SDK guidance about how to handle these headers.

PR #3986, opened 2025-01-24
Add `aws-smithy-observability` and `aws-smithy-observability-otel` crates

landonxjames:
## Motivation and Context

This PR is an updated version of https://github.com/smithy-lang/smithy-rs/pull/3912 featuring an Associated Type based implementation of of the SRA metrics interfaces instead of the previous `dyn Trait` based implementations. 

## Description

Add two new crates
* `aws-smithy-observability` - contains traits for our observability solution and a `global` module for managing the global telemetry provider
* `aws-smithy-observability-otel` - contains an OpenTelemetry based implementation of the traits

## Testing

Various unit tests in both new crates.

## Checklist

Note: leaving out a changelog entry for now since these crates won't offer much value until we instrument our runtime crates

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #4003, opened 2025-02-07
Add `aws-smithy-observability` and `aws-smithy-observability-otel` crates

landonxjames:
## Motivation and Context

This PR is an updated version of https://github.com/smithy-lang/smithy-rs/pull/3912 and https://github.com/smithy-lang/smithy-rs/pull/3986 moving back to the `dyn Trait` based implementation since some issues came up in the review of the Associated Types based implementation 

## Description

Add two new crates
* `aws-smithy-observability` - contains traits for our observability solution and a `global` module for managing the global telemetry provider
* `aws-smithy-observability-otel` - contains an OpenTelemetry based implementation of the traits

## Testing

Various unit tests in both new crates.

## Checklist

Note: leaving out a changelog entry for now since these crates won't offer much value until we instrument our runtime crates in some subsequent PRs

----


rust-runtime/aws-smithy-observability/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "aws-smithy-observability"
+version = "0.1.0"
```
aajtodd, line 3: What stability are we picturing here? At a minimum I would assume we want to keep as much of this unstable as possible while anything a user would interact with directly (e.g. global provider) we'd want stable.

rust-runtime/aws-smithy-observability/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+// libraries update this with detailed usage docs and examples
+
+mod attributes;
+pub use attributes::{AttributeValue, Attributes};
+mod context;
+pub use context::{Context, ContextManager, Scope};
+mod error;
+pub use error::{ErrorKind, GlobalTelemetryProviderError, ObservabilityError};
+mod global;
+pub use global::{get_telemetry_provider, set_telemetry_provider};
```
aajtodd, line 27: we may want this namespace to remain so that it's very clear just from the import but maybe it's ok as is. Just a thought.

e.g.

```rs
// without
use aws_smithy_observability::set_telemetry_provider;

// vs with
use aws_smithy_observability::global::set_telemetry_provider;
```

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::context::Context;
+use std::{fmt::Debug, sync::Arc};
+
+/// Provides named instances of [Meter].
+pub trait ProvideMeter: Send + Sync + Debug {
+    /// Get or create a named [Meter].
+    fn get_meter(&self, scope: &'static str, attributes: Option<&Attributes>) -> Arc<dyn Meter>;
+
+    /// Cast to [std::any::Any]
+    fn as_any(&self) -> &dyn std::any::Any;
```
aajtodd, line 19: Why is this needed on the trait?

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/// The entry point to creating instruments. A grouping of related metrics.
+pub trait Meter: Send + Sync + Debug {
```
aajtodd, line 23: It looks like in otel they have a [concrete meter type ](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/metrics/meter.rs#L296-L319)and make a separate trait for `InstrumentProvider`. This pattern may be useful here so that we can use generics in the factory functions

e.g.

```rs
struct Meter { ... }

impl Meter {
     fn create_gauge(&self, name: impl Into<Cow<'static, str>>, ...)

    // etc
}
```

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// Create a new Gauge.
+    #[allow(clippy::type_complexity)]
```
aajtodd, line 25: I don't care about the type complexity here but I do wonder if we shouldn't try to turn this into a builder as the return type such that we can evolve and expand the options you can set over time if needed.

This appears to be what otel is doing as well.

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -0,0 +1,112 @@
...
+/// Collects a set of events with an event count and sum for all events.
+pub trait Histogram: Send + Sync + Debug {
+    /// Record a value.
+    fn record(&self, value: f64, attributes: Option<&Attributes>, context: Option<&dyn Context>);
+}
+
+/// A counter that monotonically increases.
+pub trait MonotonicCounter: Send + Sync + Debug {
+    /// Increment a counter by a fixed amount.
+    fn add(&self, value: u64, attributes: Option<&Attributes>, context: Option<&dyn Context>);
```
aajtodd, line 88: I noticed otel uses slices of `KeyValue` for attributes, any idea why they chose that?

rust-runtime/aws-smithy-observability-otel/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+description = "Smithy OpenTelemetry observability implementation."
+edition = "2021"
+license = "Apache-2.0"
+repository = "https://github.com/awslabs/smithy-rs"
+
+[dependencies]
+aws-smithy-observability = { path = "../aws-smithy-observability" }
+opentelemetry = {version = "0.26.0", features = ["metrics"]}
+value-bag = "1.10.0"
+async-global-executor = "2.4.1"
```
aajtodd, line 16: Why do we need `async-task` and `async-global-executor`?

rust-runtime/aws-smithy-observability-otel/src/meter.rs
```diff
@@ -0,0 +1,520 @@
...
+        }
+
+        Arc::new(HistogramWrap(builder.init()))
+    }
+}
+
+/// An OpenTelemetry based implementation of the AWS SDK's [ProvideMeter] trait
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct AwsSdkOtelMeterProvider {
```
aajtodd, line 273: Drop `AwsSdk` from the name.

aajtodd (commented):
First pass, couple questions, comments, and suggestions.

rust-runtime/aws-smithy-observability/src/lib.rs
landonxjames: Yeah I've waffled on this. I believe that in original commits I had `global` in the function names to make it more explicit. I'll put them back under the `global::` namespace to it is more obvious.

rust-runtime/aws-smithy-observability/Cargo.toml
landonxjames: The `global::` APIs shouldn't need to change, even if the underlying implementation does, so I believe those will be stable. The traits spread around the package will not be stable, but I think they are unlikely to change too much since using `dyn Trait` puts a lot of [restrictions](https://doc.rust-lang.org/reference/items/traits.html#r-items.traits.dyn-compatible.associated-functions) on what the API can actually look like.

rust-runtime/aws-smithy-observability-otel/Cargo.toml
landonxjames: Those are transitive dependencies I had to pin to get things to build in CI. I can't remember quite why they were failing, but I'll figure it out and add a note

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: I assume it is so that you can have multiple `KeyValue`s with the same `Key` which our `HashMap` based implementation doesn't allow. That doesn't seem to be terribly valuable to me, but I could be mistaken and that might be a common pattern. The SRA doesn't say anything about whether we should  dedupe `Attributes` on `Key`, the `(Key, Value)` pair, or not at all.

rust-runtime/aws-smithy-observability-otel/src/meter.rs
landonxjames: Addressed in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/add98ddba49fb91a2cf0816697a184b6670768b8

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Both the concrete `Meter` type and the use of `impl Into<Cow<'static, str>>` implemented in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/f6db5456ae6eb435707fa43af157cae76f468776. This does introduce one weird kind of unergonomic thing. When assigning a `None` vaule to an `Option<impl Into<Cow<'static, str>>>` the compiler requires that you turbofish it `None::<&str>`.

rust-runtime/aws-smithy-observability/src/lib.rs
landonxjames: Reinstated the `global::` namespace in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/f6db5456ae6eb435707fa43af157cae76f468776

rust-runtime/aws-smithy-observability-otel/Cargo.toml
landonxjames: Note added in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/f6db5456ae6eb435707fa43af157cae76f468776

rust-runtime/aws-smithy-observability/src/meter.rs
aajtodd: I wondered if i it was performance related somehow.

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Just relying on `Drop` is probably ok for most users who are exporting their metrics in some way that gets flushed on `Drop`, but it would make our lives more difficult, at least for testing. If you look at the unit tests for the Otel implementation I have to extract the metrics provider and call `.shutdown()` to flush the metrics pipeline so I can assert on the collected metrics. See [this test](https://github.com/smithy-lang/smithy-rs/pull/4003/files#diff-b84906ce2d70c80752dbf5bb1dd1273c98d3fd6622aaf5e839ff748cf36b2324R63-R69), which fails if `.shutdown` is not called on the `OtelMeterProvider`. Maybe I could feature gate it as a test-util, but it seemed that it might have some use for users as well.

aajtodd: I think my issue is this (and the flush related APIs) seems solvable in other ways and conflates concerns. Why can't a caller just retain a reference to their concrete type?

landonxjames: Removed `as_any` in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/347cc74a2ac613192fab522714e3268df2f0b462

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Moved to builder pattern in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/f52f43c8d7499ed85418dd407d387b7ce6bf791d. This did introduce one rough edge in that I was unable to make the lifetimes of the `callback` function workout without making that field `pub` on the builder. Otel gets around this by having [all of the fields on the builder](https://github.com/open-telemetry/opentelemetry-rust/blob/cd9bbcfa0df4ff78c47404d192b29df69b7b7eaa/opentelemetry/src/metrics/instruments/mod.rs#L245-L261) be `pub`. If we have one field `pub` we might as well have the rest be, but that doesn't seem to be a pattern we really do in our codebase, so wanted to bring it up for discussion.

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Moved to a builder pattern so this isn't really an issue now

rust-runtime/aws-smithy-observability/src/instruments.rs
```diff
@@ -0,0 +1,246 @@
...
+            instrument_provider: meter.instrument_provider.as_ref(),
+            name,
+            description: None,
+            units: None,
+            _phantom: PhantomData::<T>,
+        }
+    }
+
+    /// Get the name.
+    pub fn get_name(&self) -> &Cow<'static, str> {
```
aajtodd, line 35: **style**:  Should we drop `get_` prefix on these?

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -6,107 +6,84 @@
...
-    fn create_async_up_down_counter(
+    pub fn create_async_up_down_counter(
         &self,
-        name: String,
-        callback: Box<dyn Fn(&dyn AsyncMeasure<Value = i64>) + Send + Sync>,
-        units: Option<String>,
-        description: Option<String>,
-    ) -> Arc<dyn AsyncMeasure<Value = i64>>;
+        name: impl Into<Cow<'static, str>>,
+        callback: Arc<dyn Fn(&dyn AsyncMeasure<Value = i64>) + Send + Sync>,
```
aajtodd, line 59: Should this be more like:

```rs

pub fn create_async_up_down_counter<F>(
    &self,
    name: ...,
     callback: F
)
    where F: Fn(&dyn AsyncMeasure<Value = i64>) + Send + Sync + 'static
{
    AsyncInsrumentBuilder::new(Self, name.into(), Arc::new(callback))
}
```

This allows callers to not have to box it up into an `Arc` themselves (you may need the `'static` bounds added though).

rust-runtime/aws-smithy-observability/src/error.rs
```diff
@@ -0,0 +1,86 @@
...
+/// The types of errors associated with [ObservabilityError]
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum ErrorKind {
+    /// An error setting the global [crate::TelemetryProvider]
+    SettingGlobalProvider,
+    /// An error getting the global [crate::TelemetryProvider]
+    GettingGlobalProvider,
+    /// Error flushing metrics pipeline
+    MetricsFlush,
```
aajtodd, line 29: If we've removed flush/shutdown as concerns from our traits and the instrumentation APIs we should drop these errors (and I think thats the right thing to do).

rust-runtime/aws-smithy-observability/src/error.rs
```diff
@@ -0,0 +1,86 @@
...
+    kind: ErrorKind,
+    source: BoxError,
+}
+
+/// The types of errors associated with [ObservabilityError]
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum ErrorKind {
+    /// An error setting the global [crate::TelemetryProvider]
+    SettingGlobalProvider,
```
aajtodd, line 25: Do we need separate errors for setting vs getting? The description of the error would probably suffice. I'm not even sure we need them at all since I expect these errors to be rare I would think. In which case they could just be `Kind::Other` probably.

landonxjames: Removed the separate errors in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/50f45cc2d5d38e0d47cc3103f89d8c27bc39d35c

rust-runtime/aws-smithy-observability/src/error.rs
landonxjames: Removed these in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/50f45cc2d5d38e0d47cc3103f89d8c27bc39d35c

rust-runtime/aws-smithy-observability/src/meter.rs
landonxjames: Made this change in https://github.com/smithy-lang/smithy-rs/pull/4003/commits/50f45cc2d5d38e0d47cc3103f89d8c27bc39d35c

aajtodd (approved):

PR #4006, opened 2025-02-10
Fix Sigv4 signing bug for endpoints with default ports

landonxjames:
## Motivation and Context

Bug from https://github.com/awslabs/aws-sdk-rust/issues/1244

## Description

Fix bug in Sigv4 signing that, when an endpoint contained a default port (80 for HTTP, 443 for HTTPS), would sign the request with that port in the `HOST` header even though Hyper excludes default ports from the `HOST` header.

## Testing

Added new unit test to confirm that default ports are appropriately stripped from HTTP and HTTPS requests.

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
+++ b/aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
@@ -389,10 +389,34 @@ impl<'a> CanonicalRequest<'a> {
         match canonical_headers.get(&HOST) {
             Some(header) => header.clone(),
             None => {
+                let port = uri.port();
+                let scheme = uri.scheme();
                 let authority = uri
                     .authority()
-                    .expect("request uri authority must be set for signing");
-                let header = HeaderValue::try_from(authority.as_str())
+                    .expect("request uri authority must be set for signing")
+                    .as_str();
+                let host = uri
+                    .host()
+                    .expect("request uri host must be set for signing");
+
+                // Check if port is default (80 for HTTP, 443 for HTTPS) and if so exclude it from the
+                // Host header when signing since RFC 2616 indicates that the default port should not be
+                // sent in the Host header (and Hyper strips default ports if they are present)
+                // https://datatracker.ietf.org/doc/html/rfc2616#section-14.23
+                // https://github.com/awslabs/aws-sdk-rust/issues/1244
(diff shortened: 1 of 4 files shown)


aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
```diff
@@ -389,10 +389,34 @@ impl<'a> CanonicalRequest<'a> {
...
-                let header = HeaderValue::try_from(authority.as_str())
+                    .expect("request uri authority must be set for signing")
+                    .as_str();
+                let host = uri
+                    .host()
+                    .expect("request uri host must be set for signing");
+
+                // Check if port is default (80 for HTTP, 443 for HTTPS) and if so exclude it from the
+                // Host header when signing since RFC 2616 indicates that the default port should not be
+                // sent in the Host header (and Hyper strips default ports if they are present)
```
aajtodd, line 404: We should probably not be leaving it to hyper or whatever the HTTP client is to set the host header

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
```diff
@@ -389,10 +389,34 @@ impl<'a> CanonicalRequest<'a> {
...
+                // https://datatracker.ietf.org/doc/html/rfc2616#section-14.23
+                // https://github.com/awslabs/aws-sdk-rust/issues/1244
+                let header_value = if let (Some(scheme), Some(port)) = (scheme, port) {
```
aajtodd, line 407: Maybe consider abstracting this to a function (would maybe make it more clear without the comment or at a glance)

```rs
let header_value = if is_port_scheme_default(scheme, port) {
    host
} else {
    authority
};

#[inline]
fn is_port_scheme_default(scheme, port) -> bool { ... }
```

aajtodd (approved):
Couple small non-blocking suggestions.

ysaito1001 (approved):
Looks good (no new feedback other than pending one)

aws/rust-runtime/aws-sigv4/src/http_request/canonical_request.rs
landonxjames: Created https://github.com/smithy-lang/smithy-rs/issues/4007 to track thishttps://github.com/smithy-lang/smithy-rs/issues/4007

PR #4010, opened 2025-02-12
Use `std::result::Result` instead of `Result`

drganjoo:
When the model contains a shape by the name of `Result`, the generated code fails to compile. For instance the following model fails:

```smithy
@restJson1
service ConstrainedService {
    operations: [SampleOperation]
}

@http(uri: "/anOperation", method: "POST")
operation SampleOperation {
    output: SampleInputOutput
    input: SampleInputOutput
    errors: [ValidationException]
}

structure SampleInputOutput {
    result: Result
}

structure Result {
    @pattern("^a-z$")
    chat: String
}
```

This PR ensures:

1. If `rustTemplate` is being used, then `#{Result}` is used
2. If `rust` is being used, then that is changed to `rustTemplate` and then `#{Result}` is used.
3. If `rustBlock` is being used, then the generated code should use `std::result::Result`


ysaito1001:
Any reason to not use [*preludeScope](https://github.com/smithy-lang/smithy-rs/blob/a51570be4ad4c5440399018f8c5e3eab25071918/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt#L196) instead of `"Result" to std.resolve("result::Result")`?

ysaito1001:
Thanks for going over many places! One question, is there any reason to not use [*preludeScope](https://github.com/smithy-lang/smithy-rs/blob/a51570be4ad4c5440399018f8c5e3eab25071918/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt#L196) instead of `"Result" to std.resolve("result::Result")`?

david-perez (approved):

drganjoo:
> Thanks for going over many places! One question, is there any reason to not use [*preludeScope](https://github.com/smithy-lang/smithy-rs/blob/a51570be4ad4c5440399018f8c5e3eab25071918/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt#L196) instead of `"Result" to std.resolve("result::Result")`?

I maintained the use of preludeScope only in locations where it was already implemented. The array contains over 20 entries, and using *preludeScope would pass all entries onto the stack. While I initially aimed to optimize compilation time, I now realize this optimization would be negligible compared to the total compilation time. I will submit a follow-up PR to consistently use preludeScope instead of `"Result" to std.resolve("result::Result")`

ysaito1001 (approved):

PR #4017, opened 2025-02-14
add secondary 'tls-rustls-webpki' and 'rustls-webpki' features that don't implicitly enable rustls-native-certs

jlizen:
## Motivation and Context
I'm trying to remove `openssl` entirely from my package's dependency closure.

Currently `aws-smithy-runtime` and `aws-config` both implicitly bring in openssl, even under their respective `rustls` features, because they have default `hyper-rustls` features enabled. That in turn enables `rustls-native-certs`, which then brings in OpenSsl.

## Description
I've handled this in a non-breaking manner by adding a new feature to both packages, `tls-rustls-webpki`, and `rustls-webki`, respectively. I made sure to explicitly  enable all default features for `hyper-rustls` besides `rustls-native-certs` (which now is moved to the `tls-rustls` feature only). 

I also added a `__rustls` feature that both `rustls/rustls-webpki` and `tls-rustls/tls-rustls-webpki` resolve to, which all the conditional compilation inside application code now checks for. This is the convention I see in the hyper ecosystem.

## Testing
I'm publishing this as a draft for initial feedback, as I need to do more testing. Still sorting out my local build. It looks like I should check in `Cargo.lock` files as well?

Related - are there any CI/CD updates needed to test the new featureset?

## Checklist

It's not clear if this merits a changelog entry... probably?

----

### DIFF
--- a/aws/rust-runtime/aws-config/Cargo.toml
+++ b/aws/rust-runtime/aws-config/Cargo.toml
@@ -17,9 +17,13 @@ client-hyper = ["aws-smithy-runtime/connector-hyper-0-14-x"]
 credentials-process = ["tokio/process"]
 default = ["client-hyper", "rustls", "rt-tokio", "credentials-process", "sso"]
 rt-tokio = ["aws-smithy-async/rt-tokio", "aws-smithy-runtime/rt-tokio", "tokio/rt"]
-rustls = ["aws-smithy-runtime/tls-rustls", "client-hyper"]
+rustls = ["aws-smithy-runtime/tls-rustls", "__rustls"]
+rustls-webpki = ["aws-smithy-runtime/tls-rustls-webpki", "__rustls"]
 sso = ["dep:aws-sdk-sso", "dep:aws-sdk-ssooidc", "dep:ring", "dep:hex", "dep:zeroize", "aws-smithy-runtime-api/http-auth"]
 test-util = ["aws-runtime/test-util"]
+
+# Enables common types used for rustls. Useless on its own.
+__rustls = ["aws-smithy-runtime/__rustls", "client-hyper"]
 
 # deprecated: this feature does nothing
 allow-compilation = []
--- a/aws/rust-runtime/aws-config/src/default_provider/credentials.rs
+++ b/aws/rust-runtime/aws-config/src/default_provider/credentials.rs
@@ -14,7 +14,7 @@ use crate::meta::credentials::CredentialsProviderChain;
 use crate::meta::region::ProvideRegion;
 use crate::provider_config::ProviderConfig;
 
-#[cfg(feature = "rustls")]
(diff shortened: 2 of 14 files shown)


rcoh:
Cc @aajtodd

jlizen:
Hmm, just now seeing https://github.com/smithy-lang/smithy-rs/pull/3985

> I had to introduce a new crate aws-smithy-default-tls to make this work. This is because we can't modify dependencies or add new features based on a feature. Doing it without a new crate seems only possible with rust-lang/cargo#1197.

Sounds like there is some context I'm missing; I'll drop you a line tomorrow, Aaron. Anyway, sounds like you have some wheels in motion that might satisfy this use case along with some grander ones. Glad to close this out if so. A feature flag just for this is a bit awkward.

aajtodd:
> Hmm, just now seeing #3985
> 
> > I had to introduce a new crate aws-smithy-default-tls to make this work. This is because we can't modify dependencies or add new features based on a feature. Doing it without a new crate seems only possible with [rust-lang/cargo#1197](https://github.com/rust-lang/cargo/issues/1197).
> 
> Sounds like there is some context I'm missing (unless maybe this was a typo, and you are referring to modifying features based on *a config flag / platform target?
> 
> I'll drop you a line tomorrow, Aaron. Sounds like you have some wheels in motion that might satisfy this use case along with some grander ones. Glad to close this out if so. A feature flag just for this is a bit awkward, though still worth it if it's the best path out of requiring rustls to use openssl for pki (in my opinion).

The context is we were/are trying to default the TLS provider to be s2n-tls for unix platforms and rustls+aws-lc for windows since s2n doesn't work on Windows. You can't do this today with feature flags alone (i.e. you can't conditionally enable different features of a dependency based on a feature flag). Instead I introduced a new crate that acts as a proxy, or you could think of it as representing the feature itself.

----

I haven't looked at the PR in detail but I would really prefer to work on this in the context of the hyper1 work. One of the biggest changes is moving all the HTTP/TLS concerns we can into a new `aws-smithy-http-client` crate with the idea being that out of the box you get a default HTTPS stack that should work for most users. If you want to do anything else you need to pull in the `aws-smithy-http-client` crate yourself and enable whatever flags you want and explicitly pass in an HTTP client to the SDK (or bring your own of course by implementing the appropriate traits). This should hopefully alleviate `aws-smithy-runtime` and by extension SDK crates from having to introduce feature flags all over the place and be load bearing for all HTTPS concerns. It almost has to be this way because features are supposed to be additive and most of these TLS concerns are not so instead we offer a default and if you want something else BYOH (Bring Your Own HTTP).

jlizen:
Thanks for the explanation! That makes a lot of sense to wait for hyper 1 since it solves this class of problem. I'm fine to go ahead and close this out in the meantime as it's just a small paper cut. 

Thanks for your eyes!

PR #4021, opened 2025-02-19
Add interceptor for collecting client metrics

landonxjames:
## Checklist

----


rust-runtime/aws-smithy-runtime/src/client.rs
```diff
@@ -53,3 +53,6 @@ pub mod sdk_feature;
 
 /// Smithy support-code for code generated waiters.
 pub mod waiters;
+
+/// Interceptor for collecting client metrics.
+pub mod metrics;
```
ysaito1001, line 58: Does the `metrics` module need to be `pub` (maybe not in this PR but in the future)?

aajtodd: What's the rationale on why it's public?

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,224 @@
...
+    interceptors::Intercept, runtime_components::RuntimeComponentsBuilder,
+    runtime_plugin::RuntimePlugin,
+};
+use aws_smithy_types::config_bag::{Layer, Storable, StoreReplace};
+use std::{borrow::Cow, sync::Arc, time::SystemTime};
+
+/// Struct to hold metric data in the ConfigBag
+#[derive(Debug, Clone)]
+struct MeasurementsContainer {
+    call_start: SystemTime,
```
aajtodd, line 21: should probably use a monotonic clock (i.e. [`Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html))

aajtodd: Ahh I see `SharedTimeSource` returns `SystemTime`...that is unfortunate. 🤔

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -275,20 +300,24 @@ pub fn default_plugins(
     let behavior_version = params
         .behavior_version
         .unwrap_or_else(BehaviorVersion::latest);
+    let service_name = params
```
aajtodd, line 303: Assuming this is the service name is probably not what we want even though it's what the default is and likely to work (for now anyway).

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -275,20 +300,24 @@ pub fn default_plugins(
...
-                .retry_partition_name
-                .expect("retry_partition_name is required"),
-        ),
+        default_retry_config_plugin(service_name.clone()),
         default_sleep_impl_plugin(),
         default_time_source_plugin(),
         default_timeout_config_plugin(),
         enforce_content_length_runtime_plugin(),
         default_stalled_stream_protection_config_plugin_v2(behavior_version),
+        metrics_runtime_plugin(
```
aajtodd, line 316: This isn't going to quite work the way we want.

The `OperationBuilder` isn't used in generated code and we apply default plugins when we [create the client](https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/sts/src/config.rs#L1296C36-L1299C74) (see also [here](https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/sts/src/client.rs#L52)).

At that point we won't have the operation name. 

We also want to be very careful to create the instruments and have them live on the client (i.e. we don't want them created per/operation, the attributes (dimensions) recorded with the metric will differentiate operations/services/etc). Default runtime plugins sort of gets us there (though you have to understand it's used differently by codegen than it is say by the runtime (e.g. IMDS cred provider in `aws-config`).

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,224 @@
...
+pub(crate) struct MetricsInterceptorInstruments {
+    pub(crate) operation_duration: Arc<dyn Histogram>,
+    pub(crate) attempt_duration: Arc<dyn Histogram>,
+}
+
+impl MetricsInterceptorInstruments {
+    pub(crate) fn new() -> Result<Self, ObservabilityError> {
+        let meter = get_telemetry_provider()?
+            .meter_provider()
+            .get_meter("MetricsInterceptor", None);
```
aajtodd, line 41: This scope is wrong, it should be something like the service name 

e.g. in Kotlin when the client is instantiated:

```kt
    private val telemetryScope = "aws.sdk.kotlin.services.s3"
    private val opMetrics = OperationMetrics(telemetryScope, config.telemetryProvider)
```

SRA doesn't talk about this too much but we should probably pick a scope that identifies it's the Rust SDK + the service so `aws_sdk_s3::Client` or something, we can discuss what makes sense though.

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,224 @@
...
+    operation_name: Cow<'static, str>,
+    // Holding a TimeSource here isn't ideal, but RuntimeComponents aren't available in
+    // the read_before_execution hook and that is when we need to start the timer for
+    // the operation.
+    time_source: SharedTimeSource,
+}
+
+impl MetricsInterceptor {
+    pub(crate) fn new(
+        service_name: impl Into<Cow<'static, str>>,
```
aajtodd, line 71: Perhaps we should just make service and operation name always available to the interceptor context so that any interceptor can pull them out.

aajtodd (commented):
Good start but I have some correctness/approach concerns. 

Also unrelated to metrics, (and this can be a follow on PR) we probably want to try to align some of our spans with the SRA as well before we ship all this.

e.g. the root operation span for all smithy services should include attributes `rpc.service` and `rpc.method` (AWS SDK services would also set `rpc.system` to `aws-api` which would need to happen in codegen)

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: Thankfully it already is, Yuki pointed me to https://github.com/smithy-lang/smithy-rs/blob/64144f4024f71b092facd40d4635d4011ea0b7e5/rust-runtime/aws-smithy-runtime-api/src/client/orchestrator.rs#L294

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: Added a similarly formatted scope in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/d7a3c3aabdd71b3c31b5ed6dd2da0a09c50b022a

Had to add it in codegen since the scope has to be an `&'static str` (required by our `ProvideMeter` trait which is inspired by the same type on the OTel `MeterProvider` trait) so couldn't construct it at runtime.

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
landonxjames: Added this bit to codegen in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/d7a3c3aabdd71b3c31b5ed6dd2da0a09c50b022a, but leaving this chunk here for the manually created credentials clients.

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
landonxjames: Changed this in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/d7a3c3aabdd71b3c31b5ed6dd2da0a09c50b022a, we only really needed the service name for building the scope and I have changed how that is constructed

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: Yep, we are stuck with wall clock time for any internal timers

rust-runtime/aws-smithy-runtime/src/client.rs
landonxjames: No reason, just an oversight, fixed

ysaito1001 (commented):
Thanks for the change!

[Semver failure](https://github.com/smithy-lang/smithy-rs/actions/runs/13684351362/job/38264170100?pr=4021#step:4:550) on `UnwindSafe` is indeed puzzling 🤔 
The check seems to have been introduced [almost 2.5 years ago](https://github.com/obi1kenobi/cargo-semver-checks/pull/31), even before the previous version of the cargo semver version we used.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/operation.rs
```diff
@@ -201,6 +203,12 @@ impl<I, O, E> OperationBuilder<I, O, E> {
         self
     }
 
+    /// Configures the metrics scope for the builder.
+    pub fn scope(mut self, scope: &'static str) -> Self {
```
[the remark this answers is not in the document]
landonxjames: I do not love the `&'static str` type here, but I don't think we have another option. The type is motivated by the type on [`ProvideMeter.get_meter()`](https://github.com/smithy-lang/smithy-rs/blob/b34250611c130cb71715ae68e8c6fae86c72677f/rust-runtime/aws-smithy-observability/src/meter.rs#L17-L20) method which takes an `&'static str`. That was in turn motivated by the OpenTelemetry [`MeterProvider.meter`](https://docs.rs/opentelemetry/0.28.0/opentelemetry/metrics/trait.MeterProvider.html#method.meter) method which also takes `&'static str`. So we are kind of stuck with `&'static str` all the way since it can't be constructed at runtime.

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -266,6 +278,18 @@ impl DefaultPluginParams {
         self.behavior_version = Some(version);
         self
     }
+
+    /// Sets the time source.
+    pub fn with_time_source(mut self, time_source: SharedTimeSource) -> Self {
```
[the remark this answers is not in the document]
landonxjames: Good suggestion, addressed in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/3f75ec208ae7add1865e96cbf5b6b312e0d583ab

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,230 @@
...
+        let interceptor = MetricsInterceptor::new(self.time_source.clone());
+        if let Ok(interceptor) = interceptor {
+            Cow::Owned(RuntimeComponentsBuilder::new("Metrics").with_interceptor(interceptor))
+        } else {
+            Cow::Owned(RuntimeComponentsBuilder::new("Metrics"))
+        }
+    }
+
+    fn config(&self) -> Option<FrozenLayer> {
+        let mut cfg = Layer::new("MetricsInstruments");
```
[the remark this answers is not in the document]
landonxjames: Moved in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/3f75ec208ae7add1865e96cbf5b6b312e0d583ab

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    fn read_before_execution(
+        &self,
+        _context: &aws_smithy_runtime_api::client::interceptors::context::BeforeSerializationInterceptorContextRef<'_>,
+        cfg: &mut aws_smithy_types::config_bag::ConfigBag,
+    ) -> Result<(), aws_smithy_runtime_api::box_error::BoxError> {
+        let mut layer = Layer::new("MetricsInterceptor");
+
+        layer.store_put(MeasurementsContainer {
```
ysaito1001, line 120: Have we considered putting `MeasurementsContainer` in the interceptor state as opposed to a new layer? Basically the interceptor state is the head of the config bag and pushed layers are part of the tail ([definition](https://github.com/smithy-lang/smithy-rs/blob/b34250611c130cb71715ae68e8c6fae86c72677f/rust-runtime/aws-smithy-types/src/config_bag.rs#L523-L526)). If it's put in the interceptor state, `get_mut` can avoid cloning it from tail and putting it in the interceptor state ([code](https://github.com/smithy-lang/smithy-rs/blob/b34250611c130cb71715ae68e8c6fae86c72677f/rust-runtime/aws-smithy-types/src/config_bag.rs#L593-L603))

ysaito1001 (approved):

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: Good suggestion, I was unaware of the `.interceptor_state()` method, updated to add the container to that layer in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/b64c154bfd69e3fcad59b694c75f26021f0ce422

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub(crate) attempt_duration: Arc<dyn Histogram>,
+}
+
+impl MetricsInterceptorInstruments {
+    pub(crate) fn new(scope: &'static str) -> Result<Self, ObservabilityError> {
+        let meter = get_telemetry_provider()?
+            .meter_provider()
+            .get_meter(scope, None);
+
+        Ok(Self{
```
aajtodd, line 43: We just need to remember that instruments should be created at the client level and re-used, not per operation. I think the way we've done it here via default plugins should result in what we want but something to keep in mind.

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    pub(crate) fn get_attrs_from_cfg(
+        &self,
+        cfg: &aws_smithy_types::config_bag::ConfigBag,
+    ) -> Option<Attributes> {
+        let operation_metadata = cfg.load::<Metadata>();
+
+        if let Some(md) = operation_metadata {
+            let mut attributes = Attributes::new();
+            attributes.set("rpc.system", AttributeValue::String("aws-api".into()));
```
aajtodd, line 83: probably don't need this on the metrics attributes, it's more tracing oriented (especially since this is AWS SDK specific anyway)

aws/rust-runtime/aws-config/src/http_credential_provider.rs
```diff
@@ -104,6 +105,7 @@ impl Builder {
             .runtime_plugin(StaticRuntimePlugin::new().with_config({
                 let mut layer = Layer::new("SensitiveOutput");
                 layer.store_put(SensitiveOutput);
+                layer.store_put(Metadata::new(path.clone(), provider_name));
```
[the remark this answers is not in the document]
landonxjames: The metrics interceptor mostly noops if the `Metadata` is not present since it uses that to attach the metrics to a specific operation. So these were included to allow the credentials clients to emit metrics separately. Will remove this if we end up only registering the metrics interceptor in codegen as mentioned in another comment.

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
landonxjames: I opted to include it for the credential clients as well just to let users get a more granular view of what is going on. Maybe it isn't particularly necessary, but it seemed like a nice to have. Especially since I believe the credentials are retrieved/cached at client creation time (and then again as they expire) so the work done might not line up with individual operations.

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: Since they are created [as part of the RuntimePlugin](https://github.com/smithy-lang/smithy-rs/pull/4021/files#diff-d6cb298e2703dbfcdd2e5fe3edf2336da1ea6afc5f1c69b17f6a9350e70c1c8eR218) it should be a per-client set of instruments

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: Since they are created [as part of the RuntimePlugin](https://github.com/smithy-lang/smithy-rs/pull/4021/files#diff-d6cb298e2703dbfcdd2e5fe3edf2336da1ea6afc5f1c69b17f6a9350e70c1c8eR218) it should be a per-client set of instruments

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: Removed in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/32ded4f470d57ba627ccd4b3895307ff5563a3e4

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -269,12 +269,16 @@ private fun baseClientRuntimePluginsFn(
                     let default_retry_partition = ${codegenContext.serviceShape.sdkId().dq()};
                     #{before_plugin_setup}
 
+                    let scope = "aws.sdk.rust.services.${codegenContext.serviceShape.sdkId()}";
```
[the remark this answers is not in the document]
landonxjames: Switched to the crate name in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/32ded4f470d57ba627ccd4b3895307ff5563a3e4

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
aajtodd: Actually I'm almost wondering if we shouldn't just make these part of the orchestrator at this point 🤔 (which I know we talked about a bit). A more unified approach to telemetry/metrics seems warranted since this interceptor approach is forcing us to expand the public API of default plugin params while only covering two possible metrics. The rest will have to be done elsewhere which will make everything feel scattered.

What if what we did by default was to create the instruments and guarantee they exist in the config bag.

e.g.

```rs
struct OperationTelemetry {
         operation_duration: Arc<dyn Histogram>,
         attempt_duration: Arc<dyn Histogram>,
         ... // other instruments
}

impl OperationTelemetry {
     fn new(scope: &'static str, provider: TelemetryProvider) -> Self { ... }
}
```

Place that by default into either runtime components or config bag at client creation in codegen. Then we can pull it out as needed in orchestrator. 

You could still do some in an interceptor if needed/desired and if we really wanted to you can still do this for credential providers since `OperationBuilder` allows setting runtime plugins (which `OperationTelemetry` or an related type could implement and add to config bag). This doesn't expand the public API of default plugin params for only two metrics and (IMO) is a more unified approach to how we'll handle the remaining metrics. 

Thoughts?

landonxjames: Ok, implemented this today in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/b6d0642dc80bcd15750857a87f664a24b0c67f8a
* Moved the `MetricsRuntimePlugin` to be registered in codegen instead of in the defaults
* Added builder for `MetricsRuntimePlugin` since it is `pub` now. 
* Registered it on the IMDS and HTTP credential providers
* Removed the additional fields in `DefaultPluginParams`

The `MetricsRuntimePlugin` is adding the `OperationTelemetry` struct to the `ConfigBag` on client creation so it will always be present and we can access the instruments in the orchestrator for collecting more granular metrics.

aws/rust-runtime/aws-config/src/http_credential_provider.rs
```diff
@@ -105,7 +107,14 @@ impl Builder {
                 let mut layer = Layer::new("SensitiveOutput");
                 layer.store_put(SensitiveOutput);
                 layer.freeze()
-            }));
+            }))
+            .runtime_plugin(
+                MetricsRuntimePlugin::builder()
+                    .with_scope("http_credential_provider")
```
aajtodd, line 113: Should this scope include crate name? `aws_config::http_credential_provider`

rust-runtime/aws-smithy-runtime/Cargo.toml
```diff
@@ -46,6 +46,8 @@ wire-mock = ["legacy-test-util", "aws-smithy-http-client/wire-mock"]
 [dependencies]
 aws-smithy-async = { path = "../aws-smithy-async" }
 aws-smithy-http = { path = "../aws-smithy-http" }
+aws-smithy-observability = { path = "../aws-smithy-observability" }
+aws-smithy-protocol-test = { path = "../aws-smithy-protocol-test", optional = true }
```
aajtodd, line 50: **correctness**: Why do we need this in our regular dependencies? I don't see any features that enable it either.

rust-runtime/aws-smithy-runtime/src/client.rs
```diff
@@ -53,3 +53,6 @@ pub mod sdk_feature;
 
 /// Smithy support-code for code generated waiters.
 pub mod waiters;
+
+/// Interceptor for collecting client metrics.
```
aajtodd, line 57: Maybe make docs more generic since this isn't just an interceptor, e..g "Runtime support for collecting client metrics"

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
```diff
@@ -285,6 +285,7 @@ fn validate_stalled_stream_protection_config(
 #[non_exhaustive]
 #[derive(Debug, Default)]
 pub struct DefaultPluginParams {
+    // The retry_partition_name is also the service_name
```
aajtodd, line 288: This is a bit misleading. Yes by default it's the service name but in codegen the retry partition name includes the region for instance.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/operation.rs
aajtodd, line 207: **correctness**: Still necessary?

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,273 @@
...
+
+        if let Ok(instruments) = instruments {
+            let mut cfg = Layer::new("Metrics");
+            cfg.store_put(instruments);
+
+            if let Some(metadata) = &self.metadata {
+                cfg.store_put(metadata.clone());
+            }
+
+            Some(cfg.freeze())
```
aajtodd, line 228: Our own [docs](https://github.com/smithy-lang/smithy-rs/blob/21c6cc7e10d1879e3d898c81a421474596fa33f3/rust-runtime/aws-smithy-runtime-api/src/client/runtime_plugin.rs#L82) on this seem to indicate we should store a frozen layer as best practice. Maybe Yuki can clarify why because I'm not seeing why it would matter but it caught my eye.

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// for operations that do not, like some of the credential providers.
+    pub fn with_metadata(mut self, metadata: Metadata) -> Self {
+        self.metadata = Some(metadata);
+        self
+    }
+
+    /// Build a [MetricsRuntimePlugin]
+    pub fn build(self) -> MetricsRuntimePlugin {
+        MetricsRuntimePlugin {
+            scope: self.scope.unwrap_or("unknown_metrics_scope"),
```
aajtodd, line 268: Is this not a required parameter? Is there a case where we wouldn't have this? Feels like if not then we should validate or otherwise make it required

aws/sdk/integration-tests/metrics/tests/metrics.rs
```diff
@@ -0,0 +1,207 @@
...
+    // Both metrics contain the method and service attributes
+    let call_duration_attributes =
+        extract_metric_attributes(&finished_metrics, "smithy.client.call.duration");
+    assert!(call_duration_attributes[0].contains(&KeyValue::new("rpc.method", "GetObject")));
+    assert!(call_duration_attributes[0].contains(&KeyValue::new("rpc.service", "s3")));
+
+    let attempt_duration_attributes =
+        extract_metric_attributes(&finished_metrics, "smithy.client.call.attempt.duration");
+    assert!(attempt_duration_attributes[0].contains(&KeyValue::new("rpc.method", "GetObject")));
+    assert!(attempt_duration_attributes[0].contains(&KeyValue::new("rpc.service", "s3")));
```
aajtodd, line 102: I don't think SRA defines this but I think you asked about capitalization, we may want to see if this should be consistent or how `rpc.service` is formed in an existing SDK or two. I don't actually know what was done in Kotlin, can't remember but feels like something we should align perhaps.

aajtodd (commented):
Overall I like this approach a bit better, hopefully everyone else feels the same. Couple small questions/suggestions to discuss or address.

aws/sdk/integration-tests/metrics/tests/metrics.rs
landonxjames: Seems like both [Go](https://github.com/aws/smithy-go/blob/b17b84403740243c69345be7cd349d057ce6cbd3/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java#L391) and [Kotlin](https://github.com/smithy-lang/smithy-kotlin/blob/0a29a00a17964b7126974744127765c52a8af2a0/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/operation/OperationTelemetry.kt#L65) just use the service name from the model. I'll ask around and see if anyone else is doing something different

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: Pretty sure freezing it just wraps it in an `Arc` to make cloning cheap https://github.com/smithy-lang/smithy-rs/blob/21c6cc7e10d1879e3d898c81a421474596fa33f3/rust-runtime/aws-smithy-types/src/config_bag.rs#L23-L25, but admittedly I just copied this behavior from existing RuntimePlugins, so good to understand it

rust-runtime/aws-smithy-runtime/src/client/metrics.rs
landonxjames: This was mostly to avoid the extra code bloat of having `build` return a `Result` which I find fairly annoying for things like these that are mostly intended for internal use and that we know will always have `scope` provided. I also considered having the `builder` method take in a `scope` to initialize with it, but a quick scan of the repo didn't show any builders using that pattern.

aws/rust-runtime/aws-config/src/http_credential_provider.rs
landonxjames: Fixed in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/7d5c349085a7b41af764db442b470fa958b3e2d4

rust-runtime/aws-smithy-runtime/Cargo.toml
landonxjames: Fixed in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/7d5c349085a7b41af764db442b470fa958b3e2d4

rust-runtime/aws-smithy-runtime/src/client.rs
landonxjames: Fixed in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/7d5c349085a7b41af764db442b470fa958b3e2d4

rust-runtime/aws-smithy-runtime/src/client/defaults.rs
landonxjames: Fixed in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/7d5c349085a7b41af764db442b470fa958b3e2d4

rust-runtime/aws-smithy-runtime/src/client/orchestrator/operation.rs
landonxjames: Removed in Fixed in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/7d5c349085a7b41af764db442b470fa958b3e2d4

aws/sdk/integration-tests/metrics/tests/metrics.rs
aajtodd: Kotlin uses the SDK ID

aajtodd (approved):
I do have a question now on how we're setting service name and why it's lower cased but no blocking comments.

aws/rust-runtime/aws-config/src/http_credential_provider.rs
```diff
@@ -110,10 +110,11 @@ impl Builder {
             }))
             .runtime_plugin(
                 MetricsRuntimePlugin::builder()
-                    .with_scope("http_credential_provider")
+                    .with_scope("aws_config::http_credential_provider")
                     .with_time_source(provider_config.time_source())
                     .with_metadata(Metadata::new(path.clone(), provider_name))
-                    .build(),
+                    .build()
+                    .expect("Scope has been set"),
```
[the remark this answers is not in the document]
landonxjames: Updated in https://github.com/smithy-lang/smithy-rs/pull/4021/commits/7786744ac2b72390f7b67fa19015e433b65761bc

PR #4023, opened 2025-02-19
Assorted code clean-up for event stream with RPC-bound protocols

ysaito1001:
## Motivation and Context
This is the final PR in the series adding support for event streaming with RPC-bound protocols. It primarily focuses on addressing code cleanup tasks that were intentionally left in the previous two PRs.

## Description
The code clean-up tasks include
- adding `DisableStalledStreamProtection` to disable stalled stream protection from at the top level of `ClientCodegenVisitor` across the models, eliminating the need for individual service-specific codegen decorators just to disable stalled stream protection.
- removing `eventStreamAllowList`.
- addressing `TODO(EventStream)` for RPC protocols.
- adding a custom doc to fluent builders whose output operations are event stream and contain more than one structure members (i.e., need to populate non-event stream members at the end of `send`).

## Testing
- Existing CI
- Added `DisableStalledStreamProtectionTest`

----


landonxjames (approved):
LGTM!

aajtodd (approved):
One thing not in this PR that we might want to cleanup is not having to maintain [this](https://github.com/smithy-lang/smithy-rs/blob/e9ed5fa6fcf47fce82dfa345395752a41f70772b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt#L159-L172). 

This most likely belongs on the [Protocol](https://github.com/smithy-lang/smithy-rs/blob/36b50b3dace7798c97f2b8e8217446f13aff80a7/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/Protocol.kt#L23) interface or possibly the `HttpBindingResolver`. 

It may be fine for now but if a customer brings a new protocol they have no way to do this without modifying `smithy-rs` (and of course if we implement a new RPC protocol we have to remember to update this).

ysaito1001:
> One thing not in this PR that we might want to cleanup is not having to maintain [this](https://github.com/smithy-lang/smithy-rs/blob/e9ed5fa6fcf47fce82dfa345395752a41f70772b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt#L159-L172).

Great suggestion. 4c78dc7 has removed the hard-coded `RPC_BOUND_PROTOCOLS`. This should allow the logic around that map to also work for a new protocol as we implement its `HttpBindingResolver`.

PR #4032, opened 2025-02-21
Custom TLS config

aajtodd:
## Motivation and Context

## Description
Introduces a new `TlsContext` that can be used to customize TLS configuration. Currently that only includes supplying custom root certificates. This PR also refactors the way caching works. Previously we cached complete connectors with default resolver when the connector settings were known to be defaulted. The reason given in code and past commits for this caching resolves around the fact that `hyper-rustls` `with_native_roots()` configuration option re-loads platform root certificates on each invocation. Rather than cache the connector, instead I've switched the caching to only cache loading the certificates. This reduces a bit of complexity and required context to understand when we can cache. 

### Questions/Discussion

1. There are a number of places we call `unwrap`/`expect` in this code. It was this way previously but I've introduced a few more. This stems from building a connector not returning a result. I don't know that I love this but I also suppose I can understand why we chose this. Now is the time to consider whether we'd approach it differently or not though or we are simply saying a valid HTTPS configuration.
2. Loading native certs is all blocking file calls though and not async friendly. I haven't changed anything here but it caught my eye. It would be nice if this were more async friendly. The mitigation we chose to this seemed to be caching the connector and paying that cost once up front at construction time if possible.
3. Both rustls and s2n take in PEM encoded certs. Rustls (via `rustls-pki-types`) provides a way to actually parse and load the certs as an independent configuration object (trust store). The issue is it converts them to DER format which makes using this for s2n not possible. It would be nice if we could load and validate certs before passing them off to rustls/s2n in a uniform way.
4. Should `TlsContext` and/or `TrustStore` provider builders? Currently they just self mutate which is a pattern we use in some places (e.g. stalled stream protection config comes to mind). 

## Testing
New unit tests + existing TLS tests

----


rust-runtime/aws-smithy-http-client/src/client/tls.rs
```diff
@@ -18,7 +19,123 @@ pub enum Provider {
...
+    }
+
+    /// Configure the trust store to use for the TLS context
+    pub fn with_trust_store(mut self, trust_store: TrustStore) -> Self {
+        self.trust_store = trust_store;
+        self
+    }
+
+    /// Build a new [TlsContext]
+    pub fn build(self) -> Result<TlsContext, BoxError> {
```
aajtodd, line 64: Do we want a custom error type here?

rust-runtime/aws-smithy-http-client/src/client/tls.rs
```diff
@@ -18,7 +19,123 @@ pub enum Provider {
...
+        self.enable_native_roots = enable_native_roots;
+        self
+    }
+
+    /// Add the PEM encoded certificate to the trust store
+    ///
+    /// This may be called more than once to add multiple certificates.
+    /// NOTE: PEM certificate contents are not validated until passed to the configured
+    /// TLS provider.
+    pub fn with_pem_certificate(mut self, pem_bytes: &[u8]) -> Self {
```
aajtodd, line 114: Maybe `impl AsRef<u8>` or `Into<Vec<u8>>`?

aajtodd: Maybe we should have this and `add_pem_certificate` return a `Result` to allow for validation errors in the future to be pulled forward?

landonxjames: >Maybe we should have this and add_pem_certificate return a Result to allow for validation errors in the future to be pulled forward?

I think that is reasonable, and likely much easier to debug for a user than succeeding here and sending a bad cert

rust-runtime/aws-smithy-http-client/src/client/tls.rs
```diff
@@ -72,56 +189,36 @@ cfg_rustls! {
...
+            use rustls_pki_types::CertificateDer;
+            use rustls_pki_types::pem::PemObject;
+            use rustls_native_certs::CertificateResult;
+
+            /// Cached native certificates
+            ///
+            /// Creating a `with_native_roots()` hyper_rustls client re-loads system certs
+            /// each invocation (which can take 300ms on OSx). Cache the loaded certs
+            /// to avoid repeatedly incurring that cost.
+            pub(crate) static NATIVE_ROOTS: once_cell::sync::Lazy<Vec<CertificateDer<'static>>> = once_cell::sync::Lazy::new(|| {
```
landonxjames, line 208: We are already using `once_cell` all over the place, so it probably doesn't matter much, but I have wondered if we should try to switch to [std::cell::LazyCell](https://doc.rust-lang.org/std/cell/struct.LazyCell.html) now that it is stable, just to cut down on dependencies.

landonxjames (approved):

rust-runtime/aws-smithy-http-client/src/client/tls.rs
ysaito1001: Yeah, we probably should.

rust-runtime/aws-smithy-http-client/src/client/tls.rs
ysaito1001: > Maybe impl AsRef<u8> or Into<Vec<u8>>?

I probably prefer the latter since that's being honest about the function needing an owned value.

ysaito1001 (approved):

rust-runtime/aws-smithy-http-client/src/client/tls.rs
aajtodd: Yeah it's only that way really because we have to store it until we convert it to either rustls or s2n configuration objects. I think if we were able to move to a common format and validate up front it maybe wouldn't be required to own it. Not 100% sure it will/would work out that way though.

aajtodd: > I think that is reasonable, and likely much easier to debug for a user than succeeding here and failing in rustls

Theres no way to move this forward (this being the possible validation errors). I'm going to leave it as is and if we ever have that capability we can at least validate in `TlsContextBuilder::build()` that the trust store configured is valid.

PR #4034, opened 2025-02-22
Add `content-length` check to request checksum interceptor

landonxjames:
Note: Note entirely sure if we want to do this since this behavior isn't really covered by the SEP. Also want to run this by Dmitriy and Sam to see what they thing since they are working on the AWS Chunked SEP at the moment. 

The check allows a user to provide an unsized body and a manually set content-length header and still get a checksum.

## Checklist

----


landonxjames:
We will not be merging this PR. After discussion with Dmitriy we agreed that using the `content-length` header for this is too hacky (and wouldn't really work for any service besides S3 since they are the only ones that provide it as a modeled value).

PR #4047, opened 2025-03-13
Fix up the order of auth scheme and endpoint resolution

ysaito1001:
## Motivation and Context
This PR updates the order of auth scheme resolution, identity resolution, endpoint resolution from
- endpoint -> auth scheme -> identity

to

- auth scheme -> identity -> endpoint

within the orchestrator. The scope of this PR only fixes up the resolution order above, and everything else will be handles by subsequent PRs. Furthermore, this PR introduces a special runtime marker but it won't be placed in the config bag by anyone yet, and that'll be done in the next PR.

## Description
This is the first in the series towards adding support for the [account-based endpoints](https://docs.aws.amazon.com/sdkref/latest/guide/feature-account-endpoints.html) in the Rust SDK. To enable the functionality, the information of account ID needs to be incorporated into endpoint resolution. However, the way the orchestrator works today is
- [endpoint -> auth scheme -> identity](https://github.com/smithy-lang/smithy-rs/blob/94dbe3b0f658fe2c69bdfbefba06cd3ecde2a668/rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs#L353-L360)

So we need to fix up this order to become

- auth scheme -> identity -> endpoint

The key aspect of the code changes is to correct the resolution order while maintaining backward compatibility. When older GA-ed SDKs (ex. `aws-sdk-s3 1.76.0`) update their dependencies to a newer `aws-smithy-runtime` crate that includes these changes, they must not be broken. This is crucial because some SDKs, such as S3, have already been relying on the incorrect resolution order: endpoint → auth scheme → identity.

We introduce a special runtime marker `AuthSchemeAndEndpointOrchestrationV2` to support both resolution orders within the [`try_attempt`](https://github.com/smithy-lang/smithy-rs/blob/94dbe3b0f658fe2c69bdfbefba06cd3ecde2a668/rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs#L345) function. This approach allows us to avoid introducing `aws-smithy-runtime` version 2.x, which would enforce only the correct resolution order and require customers to upgrade to 2.x to access the functionality of account-based endpoints.

Moving forward, all SDKs generated with the `aws-smithy-runtime` crate from this PR will, by default, place the marker in the config bag. When this marker is present, the execution flow in the `try_attempt` function will follow the correct resolution order: auth scheme → identity → endpoint. Older GA-ed SDKs, generated prior to this PR, will not have the marker in the config bag, causing the `try_attempt` function to continue using the incorrect resolution order: endpoint → auth scheme → identity.

## Testing
- Existing tests in CI (specifically, passing `Check for semver hazards` is crucial)
 
----

### DIFF
--- a/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
+++ b/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
@@ -365,11 +365,15 @@ impl fmt::Debug for ExploredList {
 #[cfg(all(test, any(feature = "test-util", feature = "legacy-test-util")))]
 mod tests {
     use super::*;
+    use crate::client::orchestrator::endpoints::{
+        StaticUriEndpointResolver, StaticUriEndpointResolverParams,
+    };
     use aws_smithy_runtime_api::client::auth::static_resolver::StaticAuthSchemeOptionResolver;
     use aws_smithy_runtime_api::client::auth::{
         AuthScheme, AuthSchemeId, AuthSchemeOptionResolverParams, SharedAuthScheme,
         SharedAuthSchemeOptionResolver, Sign,
     };
+    use aws_smithy_runtime_api::client::endpoint::SharedEndpointResolver;
     use aws_smithy_runtime_api::client::identity::{
         Identity, IdentityFuture, ResolveIdentity, SharedIdentityResolver,
     };
@@ -437,44 +441,66 @@ mod tests {
             }
         }
 
-        let mut ctx = InterceptorContext::new(Input::doesnt_matter());
-        ctx.enter_serialization_phase();
-        ctx.set_request(HttpRequest::empty());
-        let _ = ctx.take_input();
-        ctx.enter_before_transmit_phase();
+        async fn run_test(add_more_to_layer: impl Fn(Layer) -> Layer) {
+            let mut ctx = InterceptorContext::new(Input::doesnt_matter());
+            ctx.enter_serialization_phase();
+            ctx.set_request(HttpRequest::empty());
+            let _ = ctx.take_input();
+            ctx.enter_before_transmit_phase();
 
-        let runtime_components = RuntimeComponentsBuilder::for_tests()
-            .with_auth_scheme(SharedAuthScheme::new(TestAuthScheme { signer: TestSigner }))
-            .with_auth_scheme_option_resolver(Some(SharedAuthSchemeOptionResolver::new(
-                StaticAuthSchemeOptionResolver::new(vec![TEST_SCHEME_ID]),
-            )))
-            .with_identity_resolver(
-                TEST_SCHEME_ID,
-                SharedIdentityResolver::new(TestIdentityResolver),
-            )
-            .build()
-            .unwrap();
+            let runtime_components = RuntimeComponentsBuilder::for_tests()
+                .with_auth_scheme(SharedAuthScheme::new(TestAuthScheme { signer: TestSigner }))
+                .with_auth_scheme_option_resolver(Some(SharedAuthSchemeOptionResolver::new(
+                    StaticAuthSchemeOptionResolver::new(vec![TEST_SCHEME_ID]),
+                )))
+                .with_identity_resolver(
+                    TEST_SCHEME_ID,
+                    SharedIdentityResolver::new(TestIdentityResolver),
+                )
+                .with_endpoint_resolver(Some(SharedEndpointResolver::new(
+                    StaticUriEndpointResolver::http_localhost(8080),
+                )))
+                .build()
+                .unwrap();
 
-        let mut layer: Layer = Layer::new("test");
-        layer.store_put(AuthSchemeOptionResolverParams::new("doesntmatter"));
-        layer.store_put(Endpoint::builder().url("dontcare").build());
-        layer.store_put(AuthSchemeAndEndpointOrchestrationV2);
-        let cfg = ConfigBag::of_layers(vec![layer]);
+            let mut layer: Layer = Layer::new("test");
+            layer.store_put(AuthSchemeOptionResolverParams::new("doesntmatter"));
+            layer.store_put(Endpoint::builder().url("dontcare").build());
+            let layer = add_more_to_layer(layer);
+            let cfg = ConfigBag::of_layers(vec![layer]);
 
-        let (scheme_id, identity, _) = resolve_identity(&runtime_components, &cfg)
-            .await
-            .expect("success");
+            let (scheme_id, identity, _) = resolve_identity(&runtime_components, &cfg)
+                .await
+                .expect("success");
 
-        sign_request(scheme_id, &identity, &mut ctx, &runtime_components, &cfg).expect("success");
+            sign_request(scheme_id, &identity, &mut ctx, &runtime_components, &cfg)
+                .expect("success");
 
-        assert_eq!(
-            "success!",
-            ctx.request()
-                .expect("request is set")
-                .headers()
-                .get("Authorization")
-                .unwrap()
-        );
+            assert_eq!(
+                "success!",
+                ctx.request()
+                    .expect("request is set")
+                    .headers()
+                    .get("Authorization")
+                    .unwrap()
+            );
+        }
+
+        // test for correct auth and endpoint orchestration
+        run_test(|mut layer| {
+            layer.store_put(AuthSchemeAndEndpointOrchestrationV2);
+            layer
+        })
+        .await;
+
+        // test for legacy auth and endpoint orchestration (no `AuthSchemeAndEndpointOrchestrationV2` in `layer`)
+        run_test(|mut layer| {
+            layer.store_put(EndpointResolverParams::from(
+                StaticUriEndpointResolverParams::new(),
+            ));
+            layer
+        })
+        .await;
     }
 
     #[cfg(feature = "http-auth")]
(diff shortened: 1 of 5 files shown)


rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -190,6 +183,86 @@ pub(super) async fn orchestrate_auth(
...
+        .load::<EndpointResolverParams>()
+        .expect("endpoint resolver params must be set");
+
+    let endpoint = runtime_components
+        .endpoint_resolver()
+        .resolve_endpoint(params)
+        .await
+        .map_err(AuthOrchestrationError::FailedToResolveEndpoint)?;
+
+    let _ = extract_endpoint_auth_scheme_config(&endpoint, scheme_id)?;
```
aajtodd, line 261: Why do we have to call this here and discard the result?

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -190,6 +183,86 @@ pub(super) async fn orchestrate_auth(
...
+pub struct AuthSchemeOptionResolverHandlesEndpointResolution;
+
+impl Storable for AuthSchemeOptionResolverHandlesEndpointResolution {
+    type Storer = StoreReplace<Self>;
+}
+
+// Checks whether an `endpoint` resolved by `SharedEndpointResolver` in `runtime_components`,
+// using `EndpointResolverParams` from `cfg`, contains `scheme_id` in its `authSchemes` property.
+//
+// Returns `Ok` if the condition is met, with the `Ok` variant also containing the actual `endpoint` used for verification.
```
aajtodd, line 233: This returns `Ok` without the endpoint as well...description/intent here needs honed and cleaned up.

aajtodd (commented):
PR description seems to indicate from this PR forward the marker will be present in config bag but this PR does not contain that change. Maybe I misunderstood though.

Good start, mostly naming/framing questions and concerns which may actually be the hardest part but I'm going to harp on it because this code will be difficult for future us/others to follow if we don't frame it right.

ysaito1001:
> PR description seems to indicate from this PR forward the marker will be present in config bag but this PR does not contain that change. Maybe I misunderstood though.

Added a disclaimer in `Motivation and Context` for clarification. This PR only introduces the marker because `endpoint_auth_scheme_matches_configured_scheme_id` needed to refer to it (will rename the function BTW), but no one places it in the config bag yet. The next PR should take care of it.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
ysaito1001: Ah, good catch. Maybe I should either add a comment or refactor so the intent is clear. What we actually want to verify here is [this snippet](https://github.com/smithy-lang/smithy-rs/blob/fd59551fb4de790cf427cb8ad3bcd64dcd5a8f75/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs#L285-L294) within `extract_endpoint_auth_scheme_config`. At its core, we want this predicate:
```
auth_schemes.iter().any(|doc| {...})
```
but I didn't bother refactoring to extract that part out of `extract_endpoint_auth_scheme_config` since line 261 (or after 251 to be more precise) only takes care of the legacy logic.

Let me know, I can go either way (adding a comment or refactor the predicate).

ysaito1001: Ah, good catch. Maybe I should either add a comment or refactor so the intent is clear. What we actually want to verify here is [this snippet](https://github.com/smithy-lang/smithy-rs/blob/fd59551fb4de790cf427cb8ad3bcd64dcd5a8f75/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs#L285-L294) within `extract_endpoint_auth_scheme_config`. At its core, we want this predicate:
```
auth_schemes.iter().any(|doc| {...})
```
but I didn't bother refactoring to extract that part out of `extract_endpoint_auth_scheme_config` since line 261 above (or lines after 251 to be more precise) only takes care of the legacy logic.

Let me know, I can go either way (adding a comment or refactor the predicate).

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -190,6 +183,86 @@ pub(super) async fn orchestrate_auth(
...
+}
+
+// Marker indicating that endpoint resolution for determining the auth scheme option
+// has already been performed by a `AuthSchemeOptionResolver`, such as `EndpointBasedAuthSchemeOptionResolver`.
+//
+// When this marker is present in the config bag (the default for SDKs moving forward),
+// `resolve_identity` will skip endpoint resolution.
+#[doc(hidden)]
+#[derive(Clone, Debug)]
+pub struct AuthSchemeOptionResolverHandlesEndpointResolution;
```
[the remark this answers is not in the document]
ysaito1001: Thanks for brainstorming naming here! I slept on it and personally settled on something simple, something stating the fact about the updated runtime mode. How about `EndpointResolutionFollowsIdentityResolution` (or `IdentityResolutionFollowedByEndpointResolution`)?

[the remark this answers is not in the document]
ysaito1001: Thanks for brainstorming naming here! I slept on it and personally settled on something simple, something stating the fact about the updated runtime mode. How about `IdentityResolutionFollowedByEndpointResolution` (or `EndpointResolutionFollowsIdentityResolution`)?

I felt that this naming makes it easier for everywhere else to fall through, i.e. easier to comment what the marker is and what it's for, easier to comment on [the early return](https://github.com/smithy-lang/smithy-rs/pull/4047#discussion_r1993910315), and so forth.

Here is an example:
```
// Marker indicating the correct resolution order: identity resolution first,
// followed by endpoint resolution, as specified in the SRA.
//
// This marker is included in the config bag to signify the intended resolution order
// by design. When the crate was released for GA, the resolution order was reversed
// (endpoint resolution first, followed by identity resolution). However, we later
// discovered the order needed correcting in a backward compatibility manner.
// This marker signals the runtime to support both resolution orders without introducing
// `aws-smithy-runtime` version 2.x.
//
// When this marker is present in the config bag (the default for future SDKs),
// `resolve_identity` skips endpoint resolution, ensuring that `try_attempt` follows
// the correct resolution order: auth scheme → identity → endpoint.
// If the marker is absent (for backward compatibility), `try_attempt` continues using
// the incorrect resolution order: endpoint → auth scheme → identity.
//
// Even with this marker, a custom `AuthSchemeOptionResolver` may still perform endpoint
// resolution before identity resolution internally. However, that is an implementation
// detail of the custom `AuthSchemeOptionResolver`, and the rest of the runtime still
// follows the expected resolution order.
#[doc(hidden)]
#[derive(Clone, Debug)]
pub struct IdentityResolutionFollowedByEndpointResolution;
```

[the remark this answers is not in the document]
ysaito1001: Thanks for brainstorming naming here! I slept on it and personally settled on something simple, something stating the fact about the updated runtime mode. How about `IdentityResolutionFollowedByEndpointResolution` (or `EndpointResolutionFollowsIdentityResolution`)?

I felt that this naming makes it easier for everywhere else to fall through, i.e. easier to comment what the marker is and what it's for, easier to comment on [the early return](https://github.com/smithy-lang/smithy-rs/pull/4047#discussion_r1993910315), and so forth.

Here is an example:
```
// Marker indicating the correct resolution order: identity resolution first,
// followed by endpoint resolution, as specified in the SRA.
//
// This marker is included in the config bag to signify the intended resolution order
// by design. When the crate was released for GA, the resolution order was reversed
// (endpoint resolution first, followed by identity resolution). However, we later
// discovered the order needed correcting in a backward compatibility manner.
// This marker signals the runtime to support both resolution orders without introducing
// `aws-smithy-runtime` version 2.x.
//
// When this marker is present in the config bag (the default for future SDKs),
// `resolve_identity` skips endpoint resolution, ensuring that `try_attempt` follows
// the correct resolution order: auth scheme → identity → endpoint.
// If the marker is absent (for backward compatibility), `try_attempt` continues using
// the incorrect resolution order: endpoint → auth scheme → identity.
//
// Even with this marker, a custom `AuthSchemeOptionResolver` may still perform endpoint
// resolution internally to determine the most suitable auth scheme. However, that is an
// implementation detail of the custom `AuthSchemeOptionResolver`, and the runtime as a whole
// still follows the expected resolution order.
#[doc(hidden)]
#[derive(Clone, Debug)]
pub struct IdentityResolutionFollowedByEndpointResolution;
```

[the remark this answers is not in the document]
ysaito1001: Thanks for brainstorming naming here! I slept on it and personally settled on something simple, something stating the fact about the updated runtime mode. How about `IdentityResolutionThenEndpointResolution` (or `EndpointResolutionFollowsIdentityResolution`)?

I felt that this naming makes it easier for everywhere else to fall through, i.e. easier to comment what the marker is and what it's for, easier to comment on [the early return](https://github.com/smithy-lang/smithy-rs/pull/4047#discussion_r1993910315), and so forth.

Here is an example:
```
// Marker indicating the correct resolution order: identity resolution first,
// followed by endpoint resolution, as specified in the SRA.
//
// This marker is included in the config bag to signify the intended resolution order
// by design. When the crate was released for GA, the resolution order was reversed
// (endpoint resolution first, followed by identity resolution). However, we later
// discovered the order needed correcting in a backward compatibility manner.
// This marker signals the runtime to support both resolution orders without introducing
// `aws-smithy-runtime` version 2.x.
//
// When this marker is present in the config bag (the default for future SDKs),
// `resolve_identity` skips endpoint resolution, ensuring that `try_attempt` follows
// the correct resolution order: auth scheme → identity → endpoint.
// If the marker is absent (for backward compatibility), `try_attempt` continues using
// the incorrect resolution order: endpoint → auth scheme → identity.
//
// Even with this marker, a custom `AuthSchemeOptionResolver` may still perform endpoint
// resolution internally to determine the most suitable auth scheme. However, that is an
// implementation detail of the custom `AuthSchemeOptionResolver`, and the runtime as a whole
// still follows the expected resolution order.
#[doc(hidden)]
#[derive(Clone, Debug)]
pub struct IdentityResolutionThenEndpointResolution;
```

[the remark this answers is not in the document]
ysaito1001: Updated in 5ff90e2

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -190,6 +183,86 @@ pub(super) async fn orchestrate_auth(
...
+async fn endpoint_auth_scheme_matches_configured_scheme_id(
+    runtime_components: &RuntimeComponents,
+    cfg: &ConfigBag,
+    scheme_id: AuthSchemeId,
+) -> Result<Option<Endpoint>, AuthOrchestrationError> {
+    if cfg
+        .load::<AuthSchemeOptionResolverHandlesEndpointResolution>()
+        .is_some()
+    {
+        // Short-circuit if auth scheme option resolver internally handles endpoint resolution
```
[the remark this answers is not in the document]
ysaito1001: Updated comment in 40b5e26

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
ysaito1001: Added comment in 41e31d0

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
ysaito1001: Updated comment in 40b5e26

rust-runtime/aws-smithy-runtime/src/client/orchestrator/endpoints.rs
```diff
@@ -77,23 +77,32 @@ pub(super) async fn orchestrate_endpoint(
...
 }
 
-fn apply_endpoint(
+pub(super) fn apply_endpoint(
+    endpoint: &Endpoint,
+    ctx: &mut InterceptorContext,
+    cfg: &ConfigBag,
+) -> Result<(), BoxError> {
+    let endpoint_prefix = cfg.load::<EndpointPrefix>();
+    tracing::debug!(endpoint_prefix = ?endpoint_prefix, "resolving endpoint");
```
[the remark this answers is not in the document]
ysaito1001: Fixed in 12dd462

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -214,50 +208,67 @@ pub(super) fn sign_request(
...
+        // The orchestrator uses the correct auth scheme and endpoint resolution order,
+        // and no endpoint needs to be resolved within this function.
         return Ok(None);
     }
 
     let params = cfg
         .load::<EndpointResolverParams>()
         .expect("endpoint resolver params must be set");
 
+    tracing::debug!(endpoint_params = ?params, "resolving endpoint");
```
aajtodd, line 258: This should have a different message as it makes it seem like we're resolving the endpoint for normal execution flow but really its for legacy concerns. Perhaps something like `using legacy resolution order, resolving endpoint for auth` or something.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -447,7 +458,7 @@ mod tests {
         let mut layer: Layer = Layer::new("test");
         layer.store_put(AuthSchemeOptionResolverParams::new("doesntmatter"));
         layer.store_put(Endpoint::builder().url("dontcare").build());
-        layer.store_put(AuthSchemeOptionResolverHandlesEndpointResolution);
+        layer.store_put(AuthSchemeAndEndpointOrchestrationV2);
```
aajtodd, line 461: Do we have a test for legacy behavior?

aajtodd (approved):
One small suggestion but otherwise looks good. Fix and ship.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
ysaito1001: Good point, included in 5d9fbf6.

PR #4052, opened 2025-03-19
Align spans with SRA and add tests

landonxjames:
## Motivation and Context

Update our tracing spans to better align with the SRA.

## Description

* Add top level `{ServiceName}.{OperationName}` span with various SRA mandated attributes
* Make all spans debug level (was previously mix of debug and info)
* Add some useful fields to spans (attempt_number, etc.)
* Add tests for span behavior (checking nesting structure, number of times span appears, etc.)

## Testing

Added tests for span emission logic in the `integration-tests/metrics` test suite

## Checklist

----

### DIFF
--- a/.changelog/1742827344.md
+++ b/.changelog/1742827344.md
@@ -1,0 +1,11 @@
+---
+applies_to:
+- client
+authors: ["aajtodd"]
+references:
+- smithy-rs#4054
+breaking: false
+new_feature: false
+bug_fix: true
+---
+Fix traversal of operations bound to resources in several places including logic to determine if an event stream exists
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
@@ -205,7 +205,7 @@ private class AuthServiceRuntimePluginCustomization(private val codegenContext: 
                     val serviceHasEventStream =
                         codegenContext.serviceShape.hasEventStreamOperations(codegenContext.model)
                     if (serviceHasEventStream) {
-                        // enable the aws-runtime `sign-eventstream` feature
+                        // enable the aws-runtime `event-stream` feature
                         addDependency(
                             AwsCargoDependency.awsRuntime(runtimeConfig).withFeature("event-stream").toType()
                                 .toSymbol(),
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientRustModule.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientRustModule.kt
@@ -7,6 +7,7 @@ package software.amazon.smithy.rust.codegen.client.smithy
 
 import software.amazon.smithy.codegen.core.Symbol
 import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.knowledge.TopDownIndex
 import software.amazon.smithy.model.shapes.EnumShape
 import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.model.shapes.Shape
@@ -156,11 +157,12 @@ class ClientModuleDocProvider(
     private fun customizeModuleDoc(): Writable =
         writable {
             val model = codegenContext.model
+            val operations = TopDownIndex.of(model).getContainedOperations(codegenContext.serviceShape)
             docs("Operation customization and supporting types.\n")
-            if (codegenContext.serviceShape.operations.isNotEmpty()) {
+            if (operations.isNotEmpty()) {
                 val opFnName =
                     FluentClientGenerator.clientOperationFnName(
-                        codegenContext.serviceShape.operations.minOf { it }
+                        operations.minOf { it.id }
                             .let { model.expectShape(it, OperationShape::class.java) },
                         codegenContext.symbolProvider,
                     )
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientDocs.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientDocs.kt
@@ -6,7 +6,6 @@
 package software.amazon.smithy.rust.codegen.client.smithy.generators.client
 
 import software.amazon.smithy.model.knowledge.TopDownIndex
-import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.model.shapes.StringShape
 import software.amazon.smithy.rust.codegen.client.smithy.ClientCodegenContext
 import software.amazon.smithy.rust.codegen.core.rustlang.docs
@@ -59,17 +58,17 @@ object FluentClientDocs {
         writable {
             val model = codegenContext.model
             val symbolProvider = codegenContext.symbolProvider
-            if (model.operationShapes.isNotEmpty()) {
+            val operations = TopDownIndex.of(model).getContainedOperations(codegenContext.serviceShape)
+            if (operations.isNotEmpty()) {
                 // Find an operation with a simple string member shape
                 val (operation, member) =
-                    codegenContext.serviceShape.operations
-                        .map { id ->
-                            val operationShape = model.expectShape(id, OperationShape::class.java)
+                    operations
+                        .map { op ->
                             val member =
-                                operationShape.inputShape(model)
+                                op.inputShape(model)
                                     .members()
                                     .firstOrNull { model.expectShape(it.target) is StringShape }
-                            operationShape to member
+                            op to member
                         }
                         .sortedBy { it.first.id }
                         .firstOrNull { (_, member) -> member != null } ?: (null to null)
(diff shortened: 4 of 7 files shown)


aws/sdk/integration-tests/metrics/Cargo.toml
```diff
@@ -23,3 +23,6 @@ opentelemetry = { version = "0.26.0", features = ["metrics"] }
 opentelemetry_sdk = { version = "0.26.0", features = ["metrics", "testing"] }
 serial_test = "3.1.1"
 tokio = { version = "1.23.1", features = ["full", "test-util"] }
+tracing = "0.1.40"
```
aajtodd, line 26: trivial: might rename this integration test to be more general (e.g. `integration-tests/telemetry`)

aws/sdk/integration-tests/metrics/tests/metrics.rs
```diff
@@ -104,104 +96,10 @@ async fn metrics_have_expected_attributes() {
     // The attempt metric contains an attempt counter attribute that correctly increments
     assert!(attempt_duration_attributes
         .iter()
-        .find(|attrs| attrs.contains(&KeyValue::new("attempt", 1)))
+        .find(|attrs| attrs.contains(&KeyValue::new("attempt_number", 1)))
```
aajtodd, line 99: Think I would just keep `attempt`, this isn't called out in the SRA recommended metrics/attributes for this metric but seems ok to keep to me.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationGenerator.kt
```diff
@@ -130,7 +131,17 @@ open class OperationGenerator(
                             err.downcast::<#{OperationError}>().expect("correct error type")
                         })
                     };
+                    use #{Tracing}::Instrument;
                     let context = Self::orchestrate_with_stop_point(runtime_plugins, input, #{StopPoint}::None)
+                        // Create a parent span for the entire operation. Includes a random, internal-only,
+                        // seven-digit ID for the operation orchestration so that it can be correlated in the logs.
+                        .instrument(#{Tracing}::debug_span!(
+                                "$serviceName.$operationName",
+                                "rpc.system" = "aws-api",
```
aajtodd, line 140: **fix**: This (the `rpc.system` attribute) is AWS SDK specific and doesn't belong in generic smithy codegen

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -168,8 +168,11 @@ pub async fn invoke_with_stop_point(
         .maybe_timeout(operation_timeout_config)
         .await
     }
-    // Include a random, internal-only, seven-digit ID for the operation invocation so that it can be correlated in the logs.
-    .instrument(debug_span!("invoke", service = %service_name, operation = %operation_name, sdk_invocation_id = fastrand::u32(1_000_000..10_000_000)))
+    .instrument(debug_span!(
+        "invoke",
+        "rpc.service" = service_name,
```
aajtodd, line 173: Do we need these attributes now if we generate the same attributes already? I suppose for the few operations that go through `OperationBuilder` like IMDS client?

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
```diff
@@ -305,8 +308,12 @@ async fn try_op(
         trace!(attempt_timeout_config = ?attempt_timeout_config);
         let maybe_timeout = async {
             debug!("beginning attempt #{i}");
-            try_attempt(ctx, cfg, runtime_components, stop_point).await;
-            finally_attempt(ctx, cfg, runtime_components).await;
+            try_attempt(ctx, cfg, runtime_components, stop_point)
+                .instrument(debug_span!("try_attempt", "attempt_number" = i))
```
aajtodd, line 312: My vote would be to just keep `attempt` rather than `attempt_number`

aajtodd (commented):
Testing assertions are nice!

Couple small suggestions and one thing to fix w.r.t it being AWS SDK specific. Naming changes I'm on the fence on after re-reviewing the naming recommendations. I can probably go either way but wanted to bring it up for discussion.

Reviewing the set of recommended metrics it seems like we should be able to easily capture:

* `smithy.client.call.duration`
* `smithy.client.call.attempts`
* `smithy.client.call.errors`
* `smithy.client.call.resolve_endpoint_duration`
* `smithy.client.call.serialization_duration`
* `smithy.client.call.deserialization_duration`
* `smithy.client.call.auth.signing_duration`
* `smithy.client.call.auth.resolve_identity_duration`

And _maybe_:
* `smithy.client.http.bytes_sent`
* `smithy.client.http.bytes_received`

aws/rust-runtime/aws-config/src/default_provider/credentials.rs
```diff
@@ -72,7 +72,7 @@ impl DefaultCredentialsChain {
     async fn credentials(&self) -> provider::Result {
         self.provider_chain
             .provide_credentials()
-            .instrument(tracing::debug_span!("provide_credentials", provider = %"default_chain"))
+            .instrument(tracing::debug_span!("default_credentials_chain"))
```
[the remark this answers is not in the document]
landonxjames: This was mostly to prevent duplication. Since the `DefaultProviderChain` is (indirectly) calling `CredentialsProviderChain.credentials()` and that is already instrumented with:
```rs
tracing::debug_span!("credentials_provider_chain", provider = %name)
```
So now the nesting comes out with:
```
default_credentials_chain -------------------------------------------------------------
    credentials_provider_chain (provider1)--
                                              credentials_provider_chain (provider2)--
```

I am mostly ambivalent about the naming, but defaulted to either the function name or the name of the parent struct of the function if the function name seemed too generic. But I do think having the chain that actually iterates the providers be the one that adds the `provider` attribute makes the most sense.

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
landonxjames: >I suppose for the few operations that go through OperationBuilder like IMDS client?

Exactly, I initially had these removed, but then when testing was seeing `invoke` spans without any attributes to give context, so I added them back in.

aws/sdk/integration-tests/metrics/tests/utils/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+pub(crate) fn make_config(with_retry: bool) -> SdkConfig {
+    SdkConfig::builder()
+        .credentials_provider(SharedCredentialsProvider::new(Credentials::for_tests()))
+        .region(Region::new("us-east-1"))
+        .http_client(new_replay_client(2, with_retry))
+        .retry_config(RetryConfig::standard())
+        .build()
+}
+
+// Util for printing spans for debugging purposes
```
ysaito1001, line 127: Nit: maybe adding a short comment showing how one might set this up with tracing subscriber and an example output from printing layer would be nice?

landonxjames: Added explanatory comment in https://github.com/smithy-lang/smithy-rs/pull/4052/commits/bf1a0c51e0e905ae2ae68e93807ec9c8be8576a0

rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs
landonxjames: Went back to `attempt` in https://github.com/smithy-lang/smithy-rs/pull/4052/commits/bf1a0c51e0e905ae2ae68e93807ec9c8be8576a0

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationGenerator.kt
landonxjames: Removed in https://github.com/smithy-lang/smithy-rs/pull/4052/commits/bf1a0c51e0e905ae2ae68e93807ec9c8be8576a0 (but will attempt to add back to only aws clients in another commit)

aws/sdk/integration-tests/metrics/tests/metrics.rs
landonxjames: Reverted to `attempt` in https://github.com/smithy-lang/smithy-rs/pull/4052/commits/bf1a0c51e0e905ae2ae68e93807ec9c8be8576a0

aws/sdk/integration-tests/metrics/Cargo.toml
landonxjames: Renamed in https://github.com/smithy-lang/smithy-rs/pull/4052/commits/bf1a0c51e0e905ae2ae68e93807ec9c8be8576a0

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/OperationGenerator.kt
landonxjames: Moved to SDK specific decorator in https://github.com/smithy-lang/smithy-rs/pull/4052/commits/35d36580bee894b56d83ae1f176d3708ff3c4cdb

ysaito1001 (approved):

PR #4053, opened 2025-03-24
aws-smithy-runtime: Allow configuring aws-smithy-http-client manually

elichai:
Depending on latest ring + hyper.
Currently you can either depend on `aws-lc-rs` or on a legacy hyper 0.14 + legacy rustls 0.21.
This will allow the user to manually enable whatever features they need in `aws-smithy-http-client`, something like this:
```toml
aws-config = { version = "1", default-features = false }
aws-smithy-http-client = { version = "1", default-features = false, features = ["rustls-ring"] }
aws-smithy-runtime = { version = "1.8", default-features = false, features = ["aws-smithy-http-client"] }
``` 
That way `aws-smithy-runtime` will use `aws-smithy-http-client` with the `rustls-ring` feature

### DIFF
--- a/rust-runtime/aws-smithy-runtime/Cargo.toml
+++ b/rust-runtime/aws-smithy-runtime/Cargo.toml
@@ -13,10 +13,11 @@ repository = "https://github.com/smithy-lang/smithy-rs"
 [features]
 client = ["aws-smithy-runtime-api/client", "aws-smithy-types/http-body-1-x"]
 http-auth = ["aws-smithy-runtime-api/http-auth"]
+aws-smithy-http-client = ["dep:dep:aws-smithy-http-client"]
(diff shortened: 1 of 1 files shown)


aajtodd:
Thanks for the contribution. I'm not sure what problem you're trying to solve here but I don't think another feature flag is something we want to pursue right now. 

> Currently you can either depend on aws-lc-rs or on a legacy hyper 0.14 + legacy rustls 0.21.

The default HTTPS client is based on `rustls`+`aws-lc`. You can pull in the `aws-smithy-http-client` crate yourself in your application and configure it differently and pass it into the SDK configuration to override the HTTP client. This is the recommended approach as feature flags must be additive.

You can find some examples of configuring your HTTP client explicitly in the [developer guide](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/http.html) as well as this [discussion](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/http.html).

Issue #4054, opened 2025-03-24
EventStreams from resources do not enable signing


aajtodd:
Event streams that stem from an operation off a resource shape are not considered when determining whether to apply the [sigv4 customization that enables event stream signing. ](https://github.com/smithy-lang/smithy-rs/blob/3e81645f1f6b2d590105a5684eee03597fa74598/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt#L210). The utility that determines this [walks the operation shapes directly](https://github.com/smithy-lang/smithy-rs/blob/c39792b3d86958078974100a783a514ea359a65a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt#L83-L86). 

This is a common bug in Smithy codegen to not use one of the knowledge indexes (e.g. `TopDownIndex`) to walk the all operation shapes (`model.operations` returns only operations directly bound to the service).

Example:

```smithy
 $version: "2"
  namespace test
  service TestService {
      resources: [ TestResource ]
  }
  
  resource TestResource {
      operations: [ EventStreamOp ]
  }

  operation EventStreamOp {
      input := {
          events: Events
      }
  }
  
  @streaming 
  union Events {
      foo: Foo
      bar: Bar,
  }
  
  structure Foo {
      foo: String
  }
  
  structure Bar {
      bar: Long
  }

```

PR #4055, opened 2025-03-24
fix event stream signing to include resource operations

aajtodd:
## Motivation and Context
fixes https://github.com/smithy-lang/smithy-rs/issues/4054

## Description
`serviceShape.operations` only includes operations directly bound to the service and not operations bound via resources. This caused a bug with event stream signing because the operation came from a resource and didn't trigger the codegen path to enable the required feature flag in the runtime (see the ticket for linked code). This PR address the original issue and adds a test. I also searched for other places this may be happening as it's a common bug in Smithy codegen. 

## Checklist

----

### DIFF
--- a/.changelog/1742827344.md
+++ b/.changelog/1742827344.md
@@ -1,7 +1,7 @@
 ---
 applies_to:
 - client
-authors: []
+authors: ["aajtodd"]
 references:
 - smithy-rs#4054
 breaking: false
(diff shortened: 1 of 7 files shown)


ysaito1001 (approved):

landonxjames (approved):
Operation iteration strikes again

PR #4063, opened 2025-03-26
Stop `cargo check` examples in smithy-rs CI 

ysaito1001:
## Motivation and Context
Removes a CI step that runs `cargo check` against the examples from the `aws-doc-sdk-examples` repository

## Description
We've long had the said CI step due to the churn in the public API prior to GA. However, value of doing so has diminished after GA, especially when we have a separate check for semver hazards.

----


landonxjames (approved):

ysaito1001:
Created a separate PR to stop checking out the `aws-doc-sdk-examples` repo in CI: https://github.com/smithy-lang/smithy-rs/issues/4068

Attempted to do so in 6fdb143 but it affected other jobs seemingly unrelated.

PR #4064, opened 2025-03-26
remove sdkId transform by default

aajtodd:
## Motivation and Context
fixes https://github.com/awslabs/aws-sdk-rust/issues/1252

## Description
The service ID we [were setting](https://github.com/smithy-lang/smithy-rs/blob/5f7113f506f301296311ef637e40d226e0a6e548/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/ServiceEnvConfigDecorator.kt#L27) for service specific environment configuration was wrong. It was wrong in two ways:

1. The [sdkId](https://github.com/smithy-lang/smithy-rs/blob/7db3e9f08e422e386c56d5916c69ab495d739ee8/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/util/Smithy.kt#L157) extension we defined strips whitespace and lowercases the ID from the model. 
2. It also used `toSnakeCase()` which will not have the same effect as internally that uses `splitOnWordBoundaries` which uses heuristics to determine where a word boundary is and isn't restricted to spaces. 

It looks like we've always applied this transform to sdkId from what I can tell with the original use being to set the operation metadata for the service name and it was carried through when making an extension. This PR makes `sdkId()` return the ID from the model untouched. I think this is the safest default as remembering when you can use the extension vs not is error prone. Any required/desired transformations should be done localized to where it's used. In this case the `ServiceEnvConfigDecorator` doesn't need to transform sdkId as the [runtime already does it](https://github.com/smithy-lang/smithy-rs/blob/d1bbd018618da9ab8f8bdb1f27d9ec75c42d2505/aws/rust-runtime/aws-runtime/src/env_config.rs#L351)

## Questions

There were only four uses of the `sdkId()` function I could find:
1. Operation metadata
2. Span names
3. Service env config
4. Default retry partition naming

I think this is a safe change though it will result in different log output. Unclear if anyone would be predicating off service name if we expose it anywhere (e.g. in an interceptor). 

## Checklist

----

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
@@ -253,6 +253,7 @@ private fun baseClientRuntimePluginsFn(
         RuntimeType.forInlineFun("base_client_runtime_plugins", ClientRustModule.config) {
             val api = RuntimeType.smithyRuntimeApiClient(rc)
             val rt = RuntimeType.smithyRuntime(rc)
+            val serviceId = codegenContext.serviceShape.sdkId().lowercase().replace(" ", "")
             val behaviorVersionError =
                 "Invalid client configuration: A behavior major version must be set when sending a " +
                     "request or constructing a client. You must set it during client construction or by enabling the " +
@@ -266,7 +267,7 @@ private fun baseClientRuntimePluginsFn(
                     ::std::mem::swap(&mut config.runtime_plugins, &mut configured_plugins);
                     #{update_bmv}
 
-                    let default_retry_partition = ${codegenContext.serviceShape.sdkId().dq()};
+                    let default_retry_partition = ${serviceId.dq()};
                     #{before_plugin_setup}
 
                     let scope = ${codegenContext.moduleName.dq()};
--- a/.changelog/1743016749.md
+++ b/.changelog/1743016749.md
@@ -1,0 +1,12 @@
+---
+applies_to:
+- aws-sdk-rust
+authors:
+- aajtodd
+references:
+- aws-sdk-rust#1252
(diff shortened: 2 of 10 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
```diff
@@ -253,6 +253,7 @@ private fun baseClientRuntimePluginsFn(
         RuntimeType.forInlineFun("base_client_runtime_plugins", ClientRustModule.config) {
             val api = RuntimeType.smithyRuntimeApiClient(rc)
             val rt = RuntimeType.smithyRuntime(rc)
+            val serviceId = codegenContext.serviceShape.sdkId().lowercase().replace(" ", "")
```
landonxjames, line 256: Could this be useful enough elsewhere to add to `Smithy.kt` as a utility function so we don't end up duplicating this transformation?

landonxjames (approved):

ysaito1001 (approved):
> this is a safe change though it will result in different log output

I assume so, and affecting those who rely on the log output is not considered a breaking change 👍

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/client/FluentClientGenerator.kt
aajtodd: Possibly. Kotlin has [this](https://github.com/awslabs/aws-sdk-kotlin/blob/57cf6f6eb57659294569d3a360472d57b0a81177/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SdkIdTransform.kt#L19) but I think thats a bit overkill. I'm going to leave it for now and we can always revisit

PR #4079, opened 2025-04-02
Update docker-build action with use-cache input

landonxjames:
## Motivation and Context

Makes the gradle cache step in the composite `docker-build` action conditional and disables it for release builds. 

## Testing

----

### DIFF
--- a/.github/actions/docker-build/action.yml
+++ b/.github/actions/docker-build/action.yml
@@ -47,7 +47,7 @@ runs:
       # it and tag it as the base image to use for this action. This will prevent acquire-build-image
       # from attempting to download an image from ECR since it will already exist,
       # which enables testing build image modifications as part of the pull request.
-      if [[ -d smithy-rs-base-image ]] && [[ ${{ inputs.use_cache }} ]]; then
+      if [[ -d smithy-rs-base-image ]] && [[ ${{ inputs.use-cache }} ]]; then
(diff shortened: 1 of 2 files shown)


.github/actions/docker-build/action.yml
```diff
@@ -12,19 +12,26 @@ inputs:
...
+    required: false
+    default: true
+
 runs:
   using: composite
   steps:
-  - uses: actions/cache@v4
-    name: Gradle Cache
+  - name: Gradle Cache
+    if: ${{ github.event.inputs.use_cache == 'true' }}
```
landonxjames, line 25: There is some weird type coercion going on here and previous attempts to just do 
```yml
if: ${{ inputs.use_cache }}
```

did not actually stop the cache step from happening. The string comparison seems to be the only thing that actually works.

ysaito1001: Do we need the `github.event.` prefix?

ysaito1001 (approved):

.github/actions/docker-build/action.yml
landonxjames: I believe so. It is a weird quirk of action inputs that booleans are coerced to string in the `github.event` context, but not in the `inputs` context. More information available here: https://github.com/orgs/community/discussions/29796#discussioncomment-3415026

PR #4084, opened 2025-04-03
Scope `GITHUB_TOKEN` permissions

landonxjames:
## Description
This PR does a few things related to scoping our tokens:
* Add a `- uses: GitHubSecurityLab/actions-permissions/monitor@v1` to most of our actions so we can get ongoing summaries of the permissions each action is using. Some actions, like Windows tests and the TLS tests, are excluded because they are not supported or the proxy it uses breaks the test.
* Add explicit `permissions` scoping to various jobs that need it.
* Although not part of the PR I have changed our Workflow Permissions (in Settings > Actions > General > Workflow Permissions) from defaulting to Read/Write to Read Only.

## Testing

* The CI for this PR ran successfully (except the Canary, but that appears to be an issue unrelated to this PR)
* A dry-run release using the workflows from this branch succeeded https://github.com/smithy-lang/smithy-rs/actions/runs/14275005243
* Various other manually runnable actions tested against this branch:
  * Daily credentials verification: https://github.com/smithy-lang/smithy-rs/actions/runs/14288824835
  * Update lockfiles: https://github.com/smithy-lang/smithy-rs/actions/runs/14288809742
  * Invoke canary (failed but not for permissions reasons): https://github.com/smithy-lang/smithy-rs/actions/runs/14288631692

**Note:** I did not test the prod release workflow for obvious reasons. It might need permissions added next time it is invoked. I will cut a release as a follow up to this PR to see if anything needs updating

## Checklist

----

### DIFF
--- a/.github/workflows/dry-run-release-scheduled.yml
+++ b/.github/workflows/dry-run-release-scheduled.yml
@@ -11,6 +11,12 @@ on:
     # Runs 00:00 UTC every day
   - cron: 0 0 * * *
 
+permissions:
+  actions: read
+  contents: read
+  id-token: write
+  pull-requests: read
+
 jobs:
   smithy-rs-scheduled-dry-run-release:
     name: Scheduled dry-run release
--- a/.github/workflows/backport-pull-request.yml
+++ b/.github/workflows/backport-pull-request.yml
@@ -19,6 +19,7 @@ jobs:
   create-backport-pull-request:
     runs-on: ubuntu-latest
     steps:
+    - uses: GitHubSecurityLab/actions-permissions/monitor@v1
     - name: Checkout
       uses: actions/checkout@v4
       with:
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -18,6 +18,12 @@ concurrency:
 env:
   ecr_repository: public.ecr.aws/w0m4q9l7/github-awslabs-smithy-rs-ci
 
+permissions:
+  actions: read
+  contents: read
+  id-token: write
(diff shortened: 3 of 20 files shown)


.github/workflows/dry-run-release-scheduled.yml
```diff
@@ -11,6 +11,12 @@ on:
     # Runs 00:00 UTC every day
   - cron: 0 0 * * *
 
+permissions:
+  actions: read
+  contents: read
+  id-token: write
+  pull-requests: read
```
ysaito1001, line 18: Just to educate myself, what's the reason for `pull-requests` to be set to `read` when it was previously `none`? Does that mean the workflow needs to read metadata for a pull request?

landonxjames: It is because the dry-run release calls `release.yml` which in turn invokes `ci.yml` which needs the `pull-request` permission. The GH permissions of the parent job are inherited by the child job and even if that child job skips the step that actually uses that permission it still needs it from the parent. 

Interestingly I believe it was previously set to `write` rather than `none` since our previous default token permissions were set to:

>**Read and write permissions**
>Workflows have read and write permissions in the repository for all scopes.

As a side note, I think I will probably have to increase the `pull-request` permissions to `write` since the prod release will need that to cut the backport PR and prod/dry-run release both share the underlying `release.yml` action definition. Going to attempt a prod release once this is merged to finalize those permissions.

ysaito1001 (approved):
Must've been painful to go through all of these. Thanks for the update!

PR #4087, opened 2025-04-08
Source account ID for credentials providers and make it available in `Identity`

ysaito1001:
## Motivation and Context
This is part 3 in the series of supporting account ID-based routing (continuation of #4047, #4057).

## Description
This PR addresses two main changes:
1. Source account ID for credentials providers in `aws-config`
2. Pass the sourced account ID to `aws_smithy_runtime_api::client::identity::Identity` when converting AWS `Credentials` to `Identity`

Within the `provide_credentials` trait method, each credentials provider will search a specific location to obtain the account ID. The design specifies the following behavior for our supported credentials providers:
- `AssumeRoleProvider`: the account ID is included in the `AssumedRoleUser` ARN within a successful response from API calls such as `assume_role` and `assume_role_with_web_identity`.
- `SsoCredentialsProvider`: the account ID is located in the account ID field of the request when calling the `get_role_credentials` API.
- static credentials: the account ID can either correspond to the `aws_account_id` property from the config file or be found in the `account_id` field of the `Credentials` type.
- `CredentialsProcessProvider`: the account ID can be found in the `AccountId` key of the JSON process output.
- `EnvironmentVariableCredentialsProvider`: the account ID can be obtained from the environment variable `AWS_ACCOUNT_ID`
- `HttpCredentialProvider`: the account ID may be found in the `AccountId` field within a successful response.

The code changes follow the implementation guidance outlined above. Regardless of the credentials provider, once the account ID is sourced and included in `Credentials`, it will eventually be threaded through `Identity` within [`SharedCredentialsProvider::resolve_identity`](https://github.com/smithy-lang/smithy-rs/blob/a29bc6916a5fae28f8ed80bc1c1acf4ce7bd45be/aws/rust-runtime/aws-credential-types/src/provider/credentials.rs#L166-L173). This is the single-sourced location where `Credentials` is converted to `Identity`, hence it makes sense to store the account ID in `Identity` in [impl From\<Credentials\> for Identity](https://github.com/smithy-lang/smithy-rs/blob/5a1fb83bb5968c00af491493062dc08d8d5bc7e1/aws/rust-runtime/aws-credential-types/src/credentials_impl.rs#L303-L314). However, since `Identity` is a generic client concept, the account ID, an AWS SDK client-specific concept, needs to be stored in a generic manner. To support this, the PR adds a `properties` field to `Identity`, which is a hash map keyed by the type ID, similar to [`Layer`](https://github.com/smithy-lang/smithy-rs/blob/a29bc6916a5fae28f8ed80bc1c1acf4ce7bd45be/rust-runtime/aws-smithy-types/src/config_bag.rs#L335-L338).

The next PR will actually use the account ID stored in `Identity` to apply it to a service-specific endpoint parameters.

## Testing
- Updated credentials providers' tests in `aws-config` to now include the account ID in the expected outputs 
- Added new credentials providers' tests as prescribed in the design spec

----


aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -134,22 +166,27 @@ impl CredentialProcessProvider {
 /// Returns an error if the response cannot be successfully parsed or is missing keys.
 ///
 /// Keys are case insensitive.
+/// The function optionally takes `fallback_account_id` that originates from the profile section.
+/// If process execution result does not contain an account ID, the function uses it as a fallback.
 pub(crate) fn parse_credential_process_json_credentials(
     credentials_response: &str,
+    fallback_account_id: Option<&AccountId>,
```
aajtodd, line 173: Instead of `fallback` perhaps this should just be named as `profile_account_id` to be more clear.

aws/rust-runtime/aws-config/src/credential_process.rs
```diff
@@ -134,22 +166,27 @@ impl CredentialProcessProvider {
...
 ) -> Result<Credentials, InvalidJsonCredentials> {
     let mut version = None;
     let mut access_key_id = None;
     let mut secret_access_key = None;
     let mut session_token = None;
     let mut expiration = None;
+    let mut account_id = None;
```
aajtodd, line 180: why not just set this to the fallback to start with and let it get overridden if it's in the response, no need for the matching later.

aws/rust-runtime/aws-config/src/imds/credentials.rs
```diff
@@ -234,17 +235,19 @@ impl ImdsCredentialsProvider {
                 access_key_id,
                 secret_access_key,
                 session_token,
+                account_id,
```
aajtodd, line 238: This isn't expected from this IMDS endpoint though is it? Should we keep IMDS separate and do in a later PR as the SEP is still in-progress for this?

aws/rust-runtime/aws-config/src/sts/util.rs
```diff
@@ -39,3 +44,20 @@ pub(crate) fn default_session_name(base: &str, ts: SystemTime) -> String {
     let now = ts.duration_since(UNIX_EPOCH).expect("post epoch");
     format!("{}-{}", base, now.as_millis())
 }
+
+// A subset of functionality extracted from `endpoint_lib::arn::Arn::parse`.
+// `Arn` is `pub(crate)` within generated SDKs, making it inaccessible from `aws-config`.
+// As a result, a subset is inlined here, with less defensive verification
+// since it only deals with the string-form ARN returned by STS.
+fn parse_account_id(arn: &str) -> Result<AccountId, CredentialsError> {
```
aajtodd, line 52: Maybe add a TODO here if we add a public ARN parser (which we should have an issue for, if not we should create one) we might want to replace this with that one day.

aws/rust-runtime/aws-credential-types/src/attributes.rs
```diff
@@ -0,0 +1,33 @@
...
+
+//! Types representing specific pieces of data contained within credentials or within token
+
+use zeroize::Zeroizing;
+
+/// Type representing a unique identifier representing an AWS account.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct AccountId {
+    // Ensure it is zeroed in memory when dropped.
+    inner: Zeroizing<String>,
```
aajtodd, line 14: Not sure we need this zeroed out, account ID is not considered sensitive.

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
```diff
@@ -164,12 +179,102 @@ impl Credentials {
...
     pub fn session_token(&self) -> Option<&str> {
         self.0.session_token.as_deref()
     }
 }
 
+/// Builder for [`Credentials`]
+///
+/// Similar to [`Credentials::new`], the use of the builder is intended for a custom credentials provider implementation.
+/// It is __NOT__ secure to hardcode credentials into your application.
+#[derive(Default, Debug, Clone)]
```
aajtodd, line 197: Should this be deriving `Debug`? We don't for `Credentials` it seems and presumably so we don't leak credentials in logs accidentally.

rust-runtime/aws-smithy-runtime-api/src/client/identity.rs
```diff
@@ -285,14 +295,118 @@ impl Identity {
     pub fn expiration(&self) -> Option<SystemTime> {
         self.expiration
     }
+
+    /// Returns arbitrary data associated with this `Identity`.
+    pub fn get<T: Any + Debug + Send + Sync + 'static>(&self) -> Option<&T> {
```
aajtodd, line 300: `get_property` or `get_attribute`? `get` is nice and short of course but my (very mild) concern or gut feeling is this makes this API seem like a core operation on identity (to me for some reason). Also aligns with builder `property` API defined below better perhaps.

aajtodd (commented):
Overall looks great. Couple minor comments and questions.

aws/rust-runtime/aws-config/src/provider_config.rs
```diff
@@ -211,7 +217,7 @@ impl ProviderConfig {
         };
 
         let mut builder = SdkConfig::builder()
-            .retry_config(RetryConfig::standard())
+            .retry_config(self.retry_config.clone().unwrap_or(RetryConfig::standard()))
```
landonxjames, line 220: Could this be an 
```rs
if let Some(config) = self.retry_config {
  config.clone()
} else {
  RetryConfig::standard()
}
```

To avoid cloning in the case where it is `None`?

landonxjames (approved):

aws/rust-runtime/aws-config/src/credential_process.rs
ysaito1001: Renamed in 4fbd4d2.

aws/rust-runtime/aws-config/src/credential_process.rs
ysaito1001: Great suggestion. There was some mismatch involve if this were initialized to `profile_account_id` as-is, but updated in ae3f8f0.

aws/rust-runtime/aws-config/src/imds/credentials.rs
ysaito1001: Makes sense. Reverted (and left `TODO`) handling account ID for this place in a032dca.

aws/rust-runtime/aws-config/src/sts/util.rs
ysaito1001: Added `TODO` with an issue in fc1233e.

aws/rust-runtime/aws-credential-types/src/attributes.rs
ysaito1001: Wasn't sure whether we needed it, so tried being defensive initially. Thanks for the input, removed the use of `Zeroizing` in 811c6f3.

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
ysaito1001: Thanks for catching. Removed it in b93ec3a.

rust-runtime/aws-smithy-runtime-api/src/client/identity.rs
ysaito1001: Renamed to `property` in 37c6d31 (since there is already a getter `.data` so didn't prefix it with `get_` for consistency).

PR #4091, opened 2025-04-12
Incorporate account ID into service-specific endpoint params

ysaito1001:
## Motivation and Context
This is part 4 in the series of supporting account ID-based routing (continuation of https://github.com/smithy-lang/smithy-rs/pull/4047, https://github.com/smithy-lang/smithy-rs/pull/4057, and https://github.com/smithy-lang/smithy-rs/pull/4087).
 
## Description
In this PR, we update the orchestrator to pass an Identity, resolved using the resolve_identity function, to service-specific endpoint parameters in the code-generated client SDKs. To support this, we've made changes to the following APIs:
- `EndpointResolverParams::get_mut` provides a mutable reference to the underlying data
- `EndpointResolverParams` can set and get arbitrary properties keyed on a type ID
- `ConfigBag::get_mut` is refactored to extract the portion of the method that doesn't require `Clone` into a separate, reusable method
- The `ResolveEndpoint` trait introduces `finalize_params` method (defaulted to no-op) to allow the caller to incorporate a property stored in `EndpointResolverParams` into its underlying service-specific endpoint params.
 
The following describes the interaction and flow between the above APIs:
1. The orchestrator resolves an `Identity` with the `resolve_identity` function.
2. The orchestrator passes the `Identity` to `orchestrate_endpoint`.
3. The orchestrator stores the resolved Identity in the `EndpointResolverParams` by setting it as a property using `set_property`.
4. The orchestrator calls `SharedEndpointResolver::finalize_params` so that the concrete endpoint resolver struct (i.e. `DowncastParams`) incorporates the `Identity` into service-specific endpoint params.
   - If a service uses the `AccountID` built-in endpoint parameter in endpoint rules, the generated client SDK overrides the `DowncastParams::finalize_params` method to retrieve the account ID from the property of `Identity` and to set it to the service-specific endpoint parameters.
5. The orchestrator calls `SharedEndpointResolver::resolve_endpoint`, with the account ID included in endpoint parameters.
 
## Questions
I made `ResolveEndpoint::finalize_params` be async to allow for future extensibility. To be honest, I _currently_ cannot think of use cases where one wants to do something in an async manner, but I was concerned that if I made the method sync, that would prevent customers from doing so completely, unless we deprecate the method and introduce an async counterpart.
 
## Testing
If we generate the dynamodb SDK using the changes in this PR and run the following experimental test, 

<details>
<summary>Experimental test</summary>

```
#[tokio::test]
async fn experiment() {
    let (http_client, request) = aws_smithy_http_client::test_util::capture_request(None);
    let client = aws_sdk_dynamodb::Client::from_conf(
        aws_sdk_dynamodb::config::Config::builder()
            .region(aws_config::Region::new("us-east-1"))
            .credentials_provider(
                aws_credential_types::Credentials::builder()
                    .account_id("333333333333")
                    .access_key_id("ANOTREAL")
                    .secret_access_key("notrealrnrELgWzOk3IfjzDKtFBhDby")
                    .provider_name("test")
                    .build(),
            )
            .http_client(http_client)
            .build(),
    );
    let mut attr_v = std::collections::HashMap::new();
    attr_v.insert(
        ":s".to_string(),
        aws_sdk_dynamodb::types::AttributeValue::S("value".into()),
    );

    let mut kv = std::collections::HashMap::new();
    kv.insert(
        ":pk".to_string(),
        aws_sdk_dynamodb::types::AttributeValue::M(attr_v),
    );
    let _ = client
        .batch_get_item()
        .request_items(
            "arn:aws:dynamodb:us-east-1:333333333333:table/table_name",
            aws_sdk_dynamodb::types::KeysAndAttributes::builder()
                .keys(kv)
                .build()
                .unwrap(),
        )
        .send()
        .await;
    let _ = dbg!(request.expect_request());
}
``` 
</details>

`dbg!` above shows that the resulting request contains the endpoint URL with the account ID:
```
req = Request {
    ...
    uri: Uri {
        as_string: "https://333333333333.ddb.us-east-1.amazonaws.com/",
        parsed: H0(
            https://333333333333.ddb.us-east-1.amazonaws.com/,
        ),
    },
    ...
```
The DynamoDB endpoint tests include cases that validate behavior based on the account ID, like the one shown [here](https://github.com/awslabs/aws-sdk-rust/blob/d18b1585f06439909aede40fe206121379de0f20/sdk/dynamodb/tests/endpoint_tests.rs#L3027-L3058). Currently, we make these tests pass by explicitly setting the account ID on the client config builder. In the next PR, we plan to remove the `account_id` and `set_account_id` setters from the client config builder.

---

### DIFF
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
@@ -81,12 +81,32 @@ interface EndpointCustomization {
     fun customRuntimeFunctions(codegenContext: ClientCodegenContext): List<CustomRuntimeFunction> = listOf()
 
     /**
-     * Allows overriding the `finalize_params` method  in the `ResolveEndpoint` trait.
+     * Allows overriding the `finalize_params` method  in the `ResolveEndpoint` trait within
+     * service-specific EndpointResolver, which is rendered in `serviceSpecificEndpointResolver`.
      *
      * `ResolveEndpoint::finalize_params` provides a default no-op implementation,
      * and this customization enables the implementor to provide an alternative implementation.
+     *
+     * Example:
+     * ```kotlin
+     * override fun serviceSpecificEndpointParamsFinalizer(codegenContext: ClientCodegenContext, params: String): Writable? {
+     *     return writable {
+     *         rustTemplate("""
+     *             let identity = $params
+     *                 .get_property_mut::<#{Identity}>();
+     *             // do more things on `params`
+     *         """,
+     *         "Identity" to RuntimeType.smithyRuntimeApiClient(codegenContext.runtimeConfig)
+     *             .resolve("client::identity::Identity"),
+     *         )
+     *     }
+     * }
+     * ```
      */
-    fun overrideFinalizeEndpointParams(codegenContext: ClientCodegenContext): Writable? = null
+    fun serviceSpecificEndpointParamsFinalizer(
+        codegenContext: ClientCodegenContext,
+        params: String,
+    ): Writable? = null
 }
 
 /**
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
@@ -81,12 +81,12 @@ interface EndpointCustomization {
     fun customRuntimeFunctions(codegenContext: ClientCodegenContext): List<CustomRuntimeFunction> = listOf()
 
     /**
-     * Allows overriding the default implementation of a method in the `ResolveEndpoint` trait.
+     * Allows overriding the `finalize_params` method  in the `ResolveEndpoint` trait.
      *
-     * For example, `ResolveEndpoint::finalize_params` provides a default no-op implementation,
+     * `ResolveEndpoint::finalize_params` provides a default no-op implementation,
      * and this customization enables the implementor to provide an alternative implementation.
      */
-    fun overrideResolveEndpointDefaultedTraitMethods(codegenContext: ClientCodegenContext): Writable? = null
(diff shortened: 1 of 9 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/EndpointsDecorator.kt
```diff
@@ -79,6 +79,14 @@ interface EndpointCustomization {
...
      */
     fun customRuntimeFunctions(codegenContext: ClientCodegenContext): List<CustomRuntimeFunction> = listOf()
+
+    /**
+     * Allows overriding the `finalize_params` method  in the `ResolveEndpoint` trait.
+     *
+     * `ResolveEndpoint::finalize_params` provides a default no-op implementation,
+     * and this customization enables the implementor to provide an alternative implementation.
+     */
+    fun overrideFinalizeEndpointParams(codegenContext: ClientCodegenContext): Writable? = null
```
aajtodd, line 89: What happens if multiple customizations implement this?

ysaito1001: Made customizations append-able and updated the doc in 280294c.

rust-runtime/aws-smithy-runtime-api/src/client/endpoint.rs
```diff
@@ -13,32 +13,74 @@ use aws_smithy_types::endpoint::Endpoint;
...
 
 new_type_future! {
-    #[doc = "Future for [`EndpointResolver::resolve_endpoint`]."]
+    #[doc = "Future for [`ResolveEndpoint::resolve_endpoint`]."]
     pub struct EndpointFuture<'a, Endpoint, BoxError>;
 }
 
+new_type_future! {
+    #[doc = "Future for [`ResolveEndpoint::finalize_params`]."]
+    pub struct FinalizeParamsFuture<'a, (), BoxError>;
```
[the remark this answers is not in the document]
ysaito1001: Discussed offline and settled on a synchronous method. Values of other endpoint params are set in a synchronous interceptor method, so it would be consistent for this to be sync also. Updated in da38e79

aajtodd (approved):

Issue #4074, opened 2025-03-28
Promote aws-smithy-mocks-experimental


aajtodd:
Rename and re-release `aws-smithy-mocks-experimental` as `aws-smithy-mocks`

Fix existing issues:

Improvements:

Internal suggestions from Russell:
right now aws-smithy-mocks works by setting interceptors and returning modeled responses
which is nice, but can be brittle in some cases what I really want is to be able to mix and match:
* return a modeled thing
* reply with 503 3 times
* return a modeled thing

and I think you could do that by:
1. Setting a custom http client (which we need to do anyway, the current state where it leaves in a real http client causes problems in network jails). This client looks at request extensions
2. The aws-smithy-mocks interceptor sets a request extension saying what http response it would like to see next

aajtodd:
Fix https://github.com/smithy-lang/smithy-rs/issues/3926

Issue #3926, opened 2024-12-02
Stub out the HTTP client in `aws_smithy_mocks_experimental`


ysaito1001:
An SDK client created from [mock_client](https://github.com/smithy-lang/smithy-rs/blob/65035d5b25b916c5c88b0aa230c42d889e05a362/rust-runtime/aws-smithy-mocks-experimental/src/lib.rs#L98-L116) does not stub out the HTTP client if used on it's own. This will result in real requests being sent to the service.

We should probably stub out the underlying HTTP client with [infallible_client_fn](https://github.com/smithy-lang/smithy-rs/blob/65035d5b25b916c5c88b0aa230c42d889e05a362/rust-runtime/aws-smithy-runtime/src/client/http/test_util/infallible.rs#L27).

PR #4098, opened 2025-04-22
promote aws-smithy-mocks out of experimental

aajtodd:
## Motivation and Context
* https://github.com/smithy-lang/smithy-rs/issues/4074
* https://github.com/smithy-lang/smithy-rs/issues/3926

## Description
* Rename `aws-smithy-mocks-experimental` to `aws-smithy-mocks` and reset version to `0.1.0`
* Fix [#3926](https://github.com/smithy-lang/smithy-rs/issues/3926) by having `mock_client!` macro always set an HTTP client. By default it returns a dummy I'm a teapot response to aid in debugging and be easy to tell where that response is coming from. Also when a rule sets an HTTP response it is not set as an extension in the request and returned by the HTTP client. This allows for more thorough and accurate integration testing as the client/runtime will see the response coming from the HTTP client instead of being ignored and then replaced as it was before
* Added a new sequence builder API for allowing a rule to generate more than one response. This allows for testing retries for example.
* Also deleted a hyper 1.x test that is no longer needed and was ignored anyway

## Questions

* In this PR I've simply renamed/moved `aws-smithy-mocks-experimental` to `aws-smithy-mocks`. Do we want or need to publish one last version of `aws-smithy-mocks-experimental` with a deprecated API that instructs users it has moved? Alternatively we could publish an [informational advisory notice](https://github.com/RustSec/advisory-db#advisory-format) about the rename. We don't currently promote the crate anywhere and the dev guide will get updated so I'm on the fence for what we want to do.

## Testing
* Added new unit tests to exercise the API without needing an SDK client by leveraging the manual `Operation` APIs.
* Added new integration tests for S3 to cover the macros and test the library e2e.

## Checklist

----


rcoh (commented):
I am really excited about this!

rust-runtime/aws-smithy-mocks/src/interceptor.rs
```diff
@@ -0,0 +1,698 @@
...
+// Store active rule in config bag
+#[derive(Debug, Clone)]
+struct ActiveRule(Rule);
+
+impl Storable for ActiveRule {
+    type Storer = StoreReplace<ActiveRule>;
+}
+
+/// Interceptor which produces mock responses based on a list of rules
+pub struct MockResponseInterceptor {
```
landonxjames, line 32: I guess this has to be `pub` since the macro expands to add a `MockResponseInterceptor::new()`, but it feels like it should just be `pub(crate)` since we don't really intend for it to be used outside of the context of the macro. Could we update the macro to instead use a `pub` function like `add_mock_interceptor(&mut config_builder)` that takes in the config builder and adds the interceptor so we can keep it `pub(crate)`?

ysaito1001: Interesting. For an HTTP client, it does have a `pub` free function `create_mock_http_client` that returns `SharedHttpClient`, used by the macro. We could have a similar `pub` free function returning `SharedInterceptor`, used by the macro.

ysaito1001 (commented):
I love the work, especially documentation. Looks great, leaving minor nits and comments on top of Landon's.

rust-runtime/aws-smithy-mocks/src/rule.rs
```diff
@@ -0,0 +1,336 @@
...
+    where
+        F: Fn() -> HttpResponse + Send + Sync + 'static,
+    {
+        let generator = Arc::new(move || MockResponse::Http(response_fn()));
+        self.generators.push(generator);
+        self
+    }
+
+    /// Repeat the last added response multiple times (total count)
+    pub fn times(mut self, count: usize) -> Self {
```
ysaito1001, line 305: What would happen if we specified `.times(0)`? I've worked with google mocks for C++ where `TIMES(0)` means that no call is expected for the mock method.

aajtodd: Right now anything less than 1 has no additional effect.

rust-runtime/aws-smithy-mocks-experimental/src/lib.rs
```diff
@@ -3,11 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//! This crate allows mocking of smithy clients.
+//! This crate has been deprecated. Please migrate to the `aws-smithy-mocks` crate.
```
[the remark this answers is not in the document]
aajtodd: Right now the API _should_ be compatible so I would hope/think that simply changing out the crate would be enough. I didn't force users to go through the new `sequence()` builder API, they can still do:

```rust
mock!(Client::get_object)
    .then_output(|| ...);
```

If we wanted a breaking change then I'd probably consider whether we even want to keep these since a sequence of 1 is the same:

```rust
mock!(Client::get_object)
     .sequence()
     .output(|| ...)
     .build()
```

I see mild utility in keeping the existing API for convenience but mostly I kept it so that existing code should be easy to migrate and still work.

rust-runtime/aws-smithy-mocks/src/rule.rs
ysaito1001: Got it, we should probably call out in the doc stating anything less than 1 has no effect or leave `TODO`.

jlizen:
> Do we want or need to publish one last version of aws-smithy-mocks-experimental with a deprecated API that instructs users it has moved? Alternatively we could publish an [informational advisory notice](https://github.com/RustSec/advisory-db#advisory-format) about the rename. We don't currently promote the crate anywhere and the dev guide will get updated so I'm on the fence for what we want to do.

I think probably yes and yes, it's low-cost and avoids potential confusion.

rust-runtime/aws-smithy-mocks/src/interceptor.rs
rcoh: Personally, for APIs required by macros, but that I don't want to expose publicly, I put them in a doc(hidden) module named `__macro_impl` or something like that.

Historically, we've seen that you _need_ to eject out of the macros, and I think we should maintain that status.

aajtodd: The only thing is this needs the HTTP client to be the one created by `create_mock_http_client`. I made that function public to allow people to construct what the mock is doing manually but I'm not sure how we can validate the interceptor is being used in conjunction with the right HTTP client. I left this public and allowed the possibility though simply because people are already using this without the macro.

ysaito1001 (approved):

rcoh (approved):

PR #4109, opened 2025-04-30
Add support for version 2.1 of the IMDS credentials provider

ysaito1001:
**DO NOT MERGE THIS PR UNTIL READY TO RELEASE**

## Description
This PR adds support for version 2.1 of the IMDS credentials provider. With this update, if the underlying EC2 instance's IMDS supports it, the provider can now retrieve credentials that include an account ID. This account ID can then be used by [account-based endpoints](https://docs.aws.amazon.com/sdkref/latest/guide/feature-account-endpoints.html).

To support this functionality, the credentials provider first attempts to access the extended API endpoint, which ends with `-extended` (e.g., `http://169.254.169.254/latest/meta-data/iam/security-credentials-extended`). If this endpoint returns a 404, the provider falls back to the legacy API endpoint—the same one used in the IMDS credentials provider v2.0.

This "try extended API, then fall back to legacy API" pattern is applied to both retrieving the IMDS instance profile name and fetching credentials. Importantly:
- Once the provider determines that a specific API version (extended or legacy) is supported, it will stick to that version.
- Only the extended API can return credentials that include an account ID.

The PR also made the following IMDS credentials providers options configurable:
- `Disable IMDS credentials fetching`
- `IMDS instance profile name`

Both options can be configured via environment variables or a shared config file, with environment variables taking precedence over the config file.

**Note:** Feature tracking for this functionality will be handled in a separate PR.

## Testing
- Updated connection recordings in `aws-config/test-data/` to reflect changes in the execution flow within `.retrieve_credentials` (Note: This update is not intended to add new test coverage, but to align recordings with the updated flow.)
- Added unit tests to verify the behavior of the extended API, including falling back to the legacy API when necessary.
- Added unit tests to verify the configuration options for disabling IMDS credentials fetching and specifying the IMDS instance profile name.

## Checklist

----


aws/rust-runtime/aws-config/test-data/default-credential-provider-chain/imds_assume_role/http-traffic.json
```diff
@@ -89,7 +89,7 @@
             "action": {
                 "Request": {
                     "request": {
-                        "uri": "http://169.254.169.254/latest/meta-data/iam/security-credentials/",
+                        "uri": "http://169.254.169.254/latest/meta-data/iam/security-credentials-extended/",
```
aajtodd, line 92: So this test is updated to hit extended but the response isn't updated?

aws/rust-runtime/aws-config/src/imds/credentials.rs
```diff
@@ -207,28 +319,58 @@ impl ImdsCredentialsProvider {
...
-                "/latest/meta-data/iam/security-credentials/{}",
-                profile
-            ))
+            .get(format!("{uri_base}{profile}", uri_base = self.uri_base()))
             .await
-            .map_err(CredentialsError::provider_error)?;
+        {
+            Ok(credentials) => {
+                let state = &mut self.provider_state.write().unwrap();
+                state.resolved_profile = Some(profile.to_string());
```
aajtodd, line 340: isn't this set in `resolve_profile_name()` already?

aws/rust-runtime/aws-config/src/imds/credentials.rs
```diff
@@ -207,28 +319,58 @@ impl ImdsCredentialsProvider {
...
+                if state.api_version == ApiVersion::Unknown {
+                    state.api_version = ApiVersion::Extended;
+                }
+                Ok(credentials)
+            }
+            Err(ImdsError::ErrorResponse(raw)) if raw.response().status().as_u16() == 404 => {
+                // A block is required to drop the `RwLockWriteGuard`,
```
aajtodd, line 347: I don't think you need this or the other similar comments like it, this would be self evident to anyone trying to remove the block.

aws/rust-runtime/aws-config/src/imds/credentials.rs
```diff
@@ -523,10 +859,314 @@ mod test {
...
+                    provider,
+                    &[(Some("ASIARTEST"), None), (Some("ASIARTEST"), None)],
+                ))
+            },
+        )
+        .await;
+    }
+
+    #[tokio::test]
+    async fn should_return_valid_credentials_when_profile_is_unstable() {
```
aajtodd, line 1029: (comment unrelated to this particular code I chose it at random): The SEP defines a bunch of test cases are we covering all of them? Is there anyway to turn the JSON provided into test cases such that new test cases or changes to the SEP are easy? I don't care how we do it but some level of traceability to the SEP would be good here.

aws/rust-runtime/aws-config/src/imds/credentials.rs
```diff
@@ -143,29 +165,119 @@ impl ImdsCredentialsProvider {
...
+            return Ok(profile);
+        }
+
+        if let Some(profile) = &self.provider_state.read().unwrap().resolved_profile {
+            return Ok(profile.clone().into());
+        }
+
+        match self.client.get(self.uri_base()).await {
+            Ok(profile) => {
+                let state = &mut self.provider_state.write().unwrap();
```
landonxjames, line 248: `.read()` and `.write()` on a `RwLock` can panic, should we use `.try_read()` and `.try_write()` instead or do you not expect it to be an issue in practice?

aws/rust-runtime/aws-config/src/imds/credentials.rs
landonxjames: Also unrelated, but related to SEP JSON tests in general. We don't really seem to use them, when I have implemented a SEP with tests I've had translated them into Rust/Kotlin because that seemed to be the standard, but it would be nice if we could start using the JSON tests directly where possible.

landonxjames (approved):
One question about the `RwLock`, but otherwise LGTM!

aws/rust-runtime/aws-config/src/imds/credentials.rs
ysaito1001: Yes, unit tests have covered all the test cases (except for ones I deemed redundant from an equivalence testing point of view). The IDs SEP used for this feature are profile names, e.g. `my-profile-0001`.

> SEP test cases probably ought to be required to have an identifier

Agreed. If we used JSON directly, that would be the best traceability. We sort of have the test infrastructure as shown under `aws-config/test-data`. They work well with connection recording, but for the feature like this that hasn't been supported in the real world we cannot generate recording, so it's easier to write the tests in unit tests.

> Is there anyway to turn the JSON provided into test cases such that new test cases or changes to the SEP are easy? 

That's invent and simplify :) Requires some level of design for the to-be-built infrastructure to be useful not just in one SEP, but in many others in the future.

aws/rust-runtime/aws-config/test-data/default-credential-provider-chain/imds_assume_role/http-traffic.json
ysaito1001: Ah, good point, not strictly necessary since extended API can still return credentials without account ID. Admittedly, this is what happens when we manually tweak connection recording files for a feature that hasn't been supported in the real world.

aws/rust-runtime/aws-config/src/imds/credentials.rs
ysaito1001: Good question, I don't expect it to be an issue in practice. [read](https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.read) and [write](https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.write) can panic upon unwrapping a `LockResult` whenever a writer panics while holding an exclusive lock. If we look at write critical sections in our code (where we call `self.provider_state.write().`), they either update the field of `ProviderState` or return a value from the method. Therefore, the lock will not be poisoned.

In terms of `.read` vs. `.try_read` (or `.write` vs. `try_write`), the difference is whether the caller blocks or not (note that the notion of a poisoned lock equally exists for `.try_*` variants). I think we can use blocking ones because
- read critical sections and write critical sections are short (reading a field value or assigning a new value to the field), which is necessary so since we're already in an async context so we don't want to block for a long time anyway
- `try_*` variants returns `Err` we need to loop back and try acquirng the lock again, which I assume is better handled by blocking variants.

At least I switched to `.expect` with explanatory comment in bff4f26.

ysaito1001: Good question, I don't expect it to be an issue in practice. [read](https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.read) and [write](https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.write) can panic upon unwrapping a `LockResult` whenever a writer panics while holding an exclusive lock. If we look at write critical sections in our code (where we call `self.provider_state.write().`), they either update the field of `ProviderState` or return a value from the method. Therefore, the lock will not be poisoned.

In terms of `.read` vs. `.try_read` (or `.write` vs. `try_write`), the difference is whether the caller blocks or not (note that the notion of a poisoned lock equally exists for `.try_*` variants). I think we can use blocking ones because
- read critical sections and write critical sections are short (reading a field value or assigning a new value to the field), which is necessarily so since we are already in an async context and don't want to block for a long time anyway.
- `try_*` variants returns `Err` we need to loop back and try acquirng the lock again, which I assume is better handled by blocking variants.

At least I switched to `.expect` with explanatory comment in bff4f26.

ysaito1001: Good question, I don't expect it to be an issue in practice. [read](https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.read) and [write](https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.write) can panic upon unwrapping a `LockResult` whenever a writer panics while holding an exclusive lock. If we look at write critical sections in our code (where we call `self.provider_state.write().`), they either update the field of `ProviderState` or return a value from the method. Therefore, the lock will not be poisoned.

In terms of `.read` vs. `.try_read` (or `.write` vs. `try_write`), the difference is whether the caller blocks or not (note that the notion of a poisoned lock equally exists for `.try_*` variants). I think we can use blocking ones because
- read critical sections and write critical sections are short (reading a field value or assigning a new value to the field), which is necessarily so since we are already in an async context and don't want to block for a long time anyway.
- If `try_*` variants returns `Err`, we need to loop back and try acquiring the lock again, which I assume is better handled by blocking variants.

At least I switched to `.expect` with explanatory comment in bff4f26.

aws/rust-runtime/aws-config/src/imds/credentials.rs
ysaito1001: Ah, thanks for catching. Came from stale pseudo-code in the SEP (pseudo-code has already been updated correctly).

aws/rust-runtime/aws-config/src/imds/credentials.rs
ysaito1001: Updated in c510cfe

ysaito1001: Updated in fcdc1b8

jackkleeman:
Hey folks, with this change we get the following WARN on nodes that don't have IMDS 2.1:
```
WARN aws_config::imds::credentials: received 404 from IMDS when loading profile information. Hint: This instance may not have an IAM role associated.
```

Issue #4115, opened 2025-05-07
CBOR Implementation Fails to Box Recursive Union Variants


drganjoo:
The CBOR implementation is not properly handling recursive data structures by failing to apply `Box` to union variants where required. This results in uncompilable Rust code for recursive types.

## Issue Details

Using the following Smithy model with recursive type references:

```smithy
@rpcv2Cbor
service SampleService {
    operations: [
        GetRelatedProduct
    ]
}

operation GetRelatedProduct {
    input := {
        product: ProductListing
    }
    output := {
        product: ProductListing
    }
}

union ProductListing {
    Related: RelatedProduct
}

structure RelatedProduct {
    name: String,
    relatedTo: ProductListing
}
```

## Current Output

The union deserializer currently generates code without boxing the recursive type:

```rust
Ok(match decoder.str()?.as_ref() {
    "Related" => crate::types::ProductListing::Related(crate::protocol_serde::shape_related_product::de_related_product(decoder)?),
    _ => {
        decoder.skip()?;
        crate::types::ProductListing::Unknown
    }
})
```

## Expected Output

The deserializer should box recursive variants to prevent infinite size errors:

```rust
Ok(match decoder.str()?.as_ref() {
    "Related" => crate::types::ProductListing::Related(
        Box::new(crate::protocol_serde::shape_related_product::de_related_product(decoder)?)
    ),
    _ => {
        decoder.skip()?;
        crate::types::ProductListing::Unknown
    }
})
```

This implementation correctly uses `Box` to handle the recursive relationship between `ProductListing` and `RelatedProduct`.

PR #4116, opened 2025-05-08
Fix CBOR codegen for recursive types by applying Box to union variants

drganjoo:
# Fix CBOR recursive data structure handling with Box implementation

This PR addresses an issue in the CBOR codegen where recursive union data structures aren't properly boxed during deserialization, resulting in uncompilable Rust code.

## Problem

The current CBOR implementation correctly defines recursive data structures with `Box` in type definitions, but fails to apply the corresponding `Box::new` wrapper during deserialization. This causes type mismatches when deserializing recursive Smithy models.

For example, with this Smithy model:
```smithy
structure RecursiveOperationInputOutputNested1 {
    variant: FooChoice,
}
union FooChoice {
    choice1: String
    choice2: RecursiveOperationInputOutputNested1
}
```

The generated Rust type correctly uses `Box`:
```rust
pub enum FooChoice {
    Choice1(::std::string::String),
    Choice2(::std::boxed::Box<crate::model::RecursiveOperationInputOutputNested1>),
}
```

But the deserializer doesn't wrap the result with `Box::new`:
```rust
// In the deserializer function:
FooChoice::Choice2(
    crate::protocol_serde::shape_recursive_operation_input_output_nested1::de_recursive_operation_input_output_nested1(decoder)
?)
```

This leads to the compiler error:
```
error[E0308]: `?` operator has incompatible types
  --> shape_foo_choice.rs:14:9
   |
14 | /         crate::protocol_serde::shape_recursive_operation_input_output_nested1::de_recursive_operation_input_output_nested1(decoder)
15 | |     ?),
   | |_____^ expected `Box<RecursiveOperationInputOutputNested1>`, found `RecursiveOperationInputOutputNested1`
```

## Solution

This PR modifies the CBOR codegen to add `.map(Box::new)` when deserializing values for union variants that are defined with `Box`. The fixed code will generate:

```rust
FooChoice::Choice2(
    crate::protocol_serde::shape_recursive_operation_input_output_nested1::de_recursive_operation_input_output_nested1(decoder)
        .map(Box::new)?
)
```

This ensures type compatibility between the defined structure and the deserialized data.

## Testing

Added a protocol test in `rpcv2Cbor-extras.smithy` file.

Closes: #4115


landonxjames (approved):
Seems like there are no tests for the `CborParserGenerator`? Seems like we should add some, but not making my approval contingent on it since I know this is blocking a customer.

ysaito1001 (approved):
Looks good, +1 for Landon's comment on tests.

drganjoo:
> Seems like there are no tests for the `CborParserGenerator`? We should add some, but not making my approval contingent on it since I know this is blocking a customer.
>
> Edit: created #4119 to track

The [Smithy protocol test suite](https://github.com/smithy-lang/smithy/tree/main/smithy-protocol-tests/model) already contains comprehensive tests for `SmithyRpcv2Cbor`, which we use to verify our implementation.

We also maintain additional tests in [rpcv2Cbor-extras.smithy](https://github.com/smithy-lang/smithy-rs/blob/main/codegen-core/common-test-models/rpcv2Cbor-extras.smithy). These include:
1. Tests for edge cases specific to our Rust implementation
2. Tests that were identified as gaps in the upstream suite but haven't yet been contributed back

For this PR, I've added a test case to our extras file that specifically verifies recursive union handling. I initially considered contributing this test upstream to the main Smithy protocol test suite (which would benefit implementations in other languages as well), but for expediency I've included it only in our extras file for now. I plan to contribute it upstream separately, after which we can remove the duplicate test from our extras file.

david-perez (approved):

PR #4125, opened 2025-05-09
Fix redundant closure warnings in Rust 1.82

drganjoo:
## Problem
In Rust 1.82, the `#[allow(clippy::redundant_closure)]` attribute is not functioning as intended, causing warnings that are elevated to errors in our build process. Specifically, when generating code in the form:

```rust
input.map(#[allow(clippy::redundant_closure)] |v| MaybeConstrained::Constrained(v))
```

Clippy still reports a redundant closure error despite the allow attribute.

## Solution
This PR implements a temporary workaround until the Clippy issue is fixed. Instead of relying on the allow attribute, we now generate code that follows Clippy's recommendation by using function references directly:

```rust
input.map(MaybeConstrained::Constrained)
```

For collection types like `Vec` and `HashMap`, we're using `.unwrap_or_default()` instead of closure-wrapped constructors.

## Implementation Details
- Added special handling for empty collections
- Modified code generation to produce different output based on context
- Preserved existing behavior for cases where we need to use closures (e.g., with `Box::new`, `.into()`)

## Testing
- Verified code generation produces the expected output
- Confirmed build succeeds without warnings when using the new code generator

The workaround will be revisited once the Clippy issue is resolved in a future Rust version.

### DIFF
--- a/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
+++ b/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
@@ -351,9 +351,22 @@ class ServerBuilderGenerator(
                     if (!constrainedTypeHoldsFinalType(member)) varExpr = "($varExpr).into()"
 
                     if (wrapInMaybeConstrained) {
-                        conditionalBlock("input.map(##[allow(clippy::redundant_closure)] |v| ", ")", conditional = symbol.isOptional()) {
-                            conditionalBlock("Box::new(", ")", conditional = hasBox) {
-                                rust("$maybeConstrainedVariant($varExpr)")
+                        // Temporary fix for `#[allow(clippy::redundant_closure)]` not working in `rustc` version 1.82.
+                        // The issue occurs specifically when we generate code in the form:
+                        // ```rust
+                        // input.map(|v| MaybeConstrained(v))
+                        // ```
+                        // This pattern triggers a clippy warning about redundant closures, even with the allow attribute.
+                        // For this specific pattern, we directly use the constructor function reference.
+                        // Other cases like `Box::new(v)` or `v.into()` are valid closure usages and remain unchanged.
+                        if (symbol.isOptional() && varExpr == "v") {
+                            rust("input.map($maybeConstrainedVariant)")
+                        }
+                        else {
+                            conditionalBlock("input.map(##[allow(clippy::redundant_closure)] |v| ",")", conditional = symbol.isOptional()) {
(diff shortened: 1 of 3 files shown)


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGenerator.kt
```diff
@@ -351,9 +351,21 @@ class ServerBuilderGenerator(
                     if (!constrainedTypeHoldsFinalType(member)) varExpr = "($varExpr).into()"
 
                     if (wrapInMaybeConstrained) {
-                        conditionalBlock("input.map(##[allow(clippy::redundant_closure)] |v| ", ")", conditional = symbol.isOptional()) {
-                            conditionalBlock("Box::new(", ")", conditional = hasBox) {
-                                rust("$maybeConstrainedVariant($varExpr)")
+                        // Temporary fix for `#[allow(clippy::redundant_closure)]` not working in `rustc` version 1.82.
```
ysaito1001, line 354: I have created an issue for the clippy team
https://github.com/rust-lang/rust-clippy/issues/14789

Can you update the comment to include
`TODO(https://github.com/rust-lang/rust-clippy/issues/14789) ...`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerBuilderGeneratorCommon.kt
```diff
@@ -84,8 +85,18 @@ fun generateFallbackCodeToDefaultValue(
     symbolProvider: RustSymbolProvider,
     publicConstrainedTypes: Boolean,
 ) {
-    var defaultValue = defaultValue(model, runtimeConfig, symbolProvider, member)
     val targetShape = model.expectShape(member.target)
+    // Temporary fix for `#[allow(clippy::redundant_closure)]` not working in `rustc` version 1.82.
```
ysaito1001, line 89: Similarly,
```
TODO(https://github.com/rust-lang/rust-clippy/issues/14789) ...
```

ysaito1001 (approved):

landonxjames:
Note that in https://github.com/smithy-lang/smithy-rs/pull/4154 I added a [-A clippy::redundant_closure](https://github.com/smithy-lang/smithy-rs/blob/fd8b4f6ce70a4ed376d0e22da088c534cd5aff63/tools/ci-build/Dockerfile#L209) to our `RUSTFLAGS` to get around this lint. Seems like it caused a merge conflict here, but should be easy to fix and disable for your own testing.

PR #4132, opened 2025-05-14
Update UnionGenerator to handle members named Unknown

landonxjames:
## Motivation and Context

Currently when a Smithy `union` type contains a member named `unknown` it causes a codegen error because we generate two functions named `is_unknown` in the `impl` block for the enum. One to indicate if the type is the `Unknown` variant we add to all enums and the other to indicate if it is the variant for the 'unknown' member in the model (which in codegen is already renamed to `UnknownValue` in the enum). 

This change mirrors the renaming of the enum variant by changing the function for it to `is_unknown_value`.  This is achieved by first passing the member through the `SymbolProvider` to get the properly renamed symbol. 

## Testing

Added codegen test for a `union` with a member named `unknown`

## Checklist

----

### DIFF
--- a/.changelog/unknown-variants.md
+++ b/.changelog/unknown-variants.md
@@ -1,0 +1,10 @@
+---
+applies_to: ["client", "server"]
+authors: ["rcoh"]
+references: ["smithy-rs#4132"]
+breaking: false
+new_feature: false
+bug_fix: true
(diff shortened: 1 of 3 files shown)


ysaito1001 (approved):
Nice!

PR #4161, opened 2025-06-04
Code generate default auth scheme resolver and related types

ysaito1001:
## Motivation and Context
Currently, the code generator does not adhere to the auth scheme resolver implementation and configuration as outlined in our internal design. For instance:
* It is not code generated per-service, instead relies upon the singleton [StaticAuthSchemeOptionResolver](https://github.com/smithy-lang/smithy-rs/blob/554cf30e3f35bdb52dd777eeadcd44465b4f879c/rust-runtime/aws-smithy-runtime-api/src/client/auth/static_resolver.rs#L10-L14).
* The supporting trait [ResolveAuthSchemeOption](https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-runtime-api/src/client/auth.rs#L232) does not take service-specific auth scheme parameters.

Similarly, the configuration of auth scheme resolver in the Rust SDK does not allow 
* customers to override the auth scheme resolver at either the service client level or the individual operation invocation level.
* customers to register a custom auth scheme at either the service client level or the individual operation invocation level.

These misalignments make it difficult to incorporate future functional requirements.

## Description
This is the first PR in the series of auth alignment. It only adds new code to the generated SDKs, but won't be used at runtime. The new code lives in a newly created `crate::config::auth` module, containing following Rust items:
- `DefaultAuthSchemeResolver` - the default auth scheme resolver
- `ResolveAuthScheme` - trait for auth scheme resolution that is per-service-params aware
- `DowncastParams` - similar to [the endpoint resolver counterpart](https://github.com/smithy-lang/smithy-rs/blob/004a1ae7706d2a3427bfe941e95de3ebf4d09fc3/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointResolverGenerator.kt#L435), but this one brides the `aws_smithy_runtime_api::client::auth::ResolveAuthSchemeOptions` trait and the `ResolveAuthScheme` above
- `Params`, `ParamsBuilder` - types representing service-specific auth scheme parameters

Each PR, this one and subsequent ones, ensures that CI checks pass. To maintain compatibility with existing code while introducing new functionality, this change preserves current auth-related codegen constructs—for example, `software.amazon.smithy.rust.codegen.client.smithy.customize.AuthSchemeOption` and the `authOptions` method on `ClientCodegenDecorator`.
I’ve added `TODO(AuthAlignment)` comments to mark areas that will be removed or refactored before merging the `ysaito/align-auth-with-sra` feature branch into `main`.

As part of this task, we've discovered that a model transformer for S3 [adds the `@optionalAuth` trait to all of its operations](https://github.com/smithy-lang/smithy-rs/blob/004a1ae7706d2a3427bfe941e95de3ebf4d09fc3/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt#L212-L225). This should be unnecessary once the auth implementation is fully aligned with SRA, but the harm has been done. To both maintain compatibility and make `AuthIndex` work, we've temporarily added `@optionalAuth` to S3's service shape rather than individual operations. However, in a subsequent PR, we're planning to remove the `AddOptionalAuth` transformer completely (so not even annotating the service with `@optionalAuth`). To avoid breakage, we'll do necessary pluming behind the scene of the [no_credentials](https://docs.rs/aws-config/latest/aws_config/struct.ConfigLoader.html#method.no_credentials) method, that is, if credentials_provider is `None` in `SdkConfig`, then we'll inject a `NoAuthSchemeResolver` on a client config builder within `impl From<SdkConfig> for config::Builder`, assuming we'll add setters for auth scheme resolver on the client config builder in the subsequent  PR:
```
#[derive(Debug)]
struct NoAuthSchemeResolver;

impl crate::config::auth::ResolveAuthScheme for NoAuthSchemeResolver {
    fn resolve_auth_scheme<'a>(
        &'a self,
        _params: &'a crate::config::auth::Params,
        _cfg: &'a ConfigBag,
        _runtime_components: &'a RuntimeComponents,
    ) -> AuthSchemeOptionsFuture<'a> {
        AuthSchemeOptionsFuture::ready(Ok(vec![AuthSchemeOption::from(NO_AUTH_SCHEME_ID)]))
    }
}
```

## Testing
- Existing CI
- `AuthIndexTest` - a port of [Kotlin SDK counterpart](https://github.com/smithy-lang/smithy-kotlin/blob/747dafb0774ab027fb0770024756f5508d7fb86d/codegen/smithy-kotlin-codegen/src/test/kotlin/software/amazon/smithy/kotlin/codegen/model/knowledge/AuthIndexTest.kt)
- `AuthSchemeResolverGeneratorTest` - verifies basic positive cases for `DefaultAuthSchemeResolver`

----


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/HttpAuthDecorator.kt
```diff
@@ -86,6 +90,50 @@ class HttpAuthDecorator : ClientCodegenDecorator {
...
+                        override fun render(
+                            codegenContext: ClientCodegenContext,
+                            operation: OperationShape?,
+                        ) = writable {
+                            rustTemplate(
+                                """
+                                #{AuthSchemeOption}::builder()
+                                    .scheme_id($name)
+                                    .build()
+                                    .expect("required fields set")
```
ysaito1001, line 118: Considering if we should implement `From<AuthSchemeId> for AuthSchemeOption` for better ergonomics.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthIndex.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Get the set of operations that need overridden in the generated auth scheme resolver.
+     */
+    fun operationsWithOverrides(): Set<OperationShape> =
+        TopDownIndex.of(ctx.model)
+            .getContainedOperations(ctx.serviceShape)
+            .filter { op ->
+                op.hasTrait<AuthTrait>() || op.hasTrait<UnsignedPayloadTrait>() || op.hasTrait<OptionalAuthTrait>()
```
aajtodd, line 80: In Kotlin unsigned payload trait resulted in an "operation override" where we set additional signing context, this likely isn't how we do it in Rust and may not be necessary here.

aajtodd (approved):
Looks like a good start!

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthIndex.kt
ysaito1001: Ah, interesting. It's unnecessary for Sigv4, since we set [`SignableBody::UnsignedPayload` in an operation runtime plugin](https://github.com/smithy-lang/smithy-rs/blob/ede2e1f95ef2bf7a75f603c12ea1460b4b728ddd/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt#L270-L292C62) if the trait is present at an operation.

One potential advantage for doing as Kotlin did is that the unsigned payload trait is respected for other, maybe new, auth scheme IDs in the future?

Let us keep it in this PR and revisit before the `align-auth-with-sra` feature is merged to main.

PR #4166, opened 2025-06-08
Expose setter for `client_builder` in `aws_smithy_http_client::client::Builder`

kevinzwang:
## Motivation and Context

Right now [`Builder::client_builder`](https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-http-client/src/client.rs#L556) cannot actually be set, which means that users will only be able to use the default Hyper client builder.

The legacy [`HyperClientBuilder`](https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-http-client/src/hyper_legacy.rs#L553) allowed setting both the client builder and the TCP connector, so this functionality is lost when porting code over to the new, Hyper v1-based HTTP client.

## Description
Adds a method on `Builder` to set `client_builder`, similar to existing builder methods. It's implemented over a generic `Builder<Tcp>` so that it can be used when a builder has its TCP provider both set or unset.

## Testing
none

## Checklist

----

### DIFF
--- a/.changelog/1748945737.md
+++ b/.changelog/1748945737.md
@@ -1,0 +1,11 @@
+---
+applies_to:
+- client
+authors:
+- kevinzwang
+references: []
+breaking: false
+new_feature: true
+bug_fix: false
+---
+Expose setter for `client_builder` in `aws_smithy_http_client::client::Builder`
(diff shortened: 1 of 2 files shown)


kevinzwang:
Closing because configuring the `client_builder` will not actually result in the ability to plug in a custom cert verifier, which is what I am aiming to do. I think what I need is to actually build my own implementation of `HttpClient` so that I can return a custom `HttpConnector` with my own verifier. If there is a better way though do let me know.

PR #4175, opened 2025-06-12
Replace per-operation auth scheme resolver with code generated default

ysaito1001:
## Motivation and Context
The second PR in the auth alignment series (follow-up on #4161)

## Description
Changes in this PR:
- Stopped creating an auth scheme resolver per-operation and used the code-generated default auth scheme resolver:
  - Placed an `AuthSchemeOptionResolverParams` containing the operation name into the config bag in an operation runtime plugin.
  - Deleted the old `authSchemes` method on `ClientCodegenDecorator` and had the call sites fully switch to the new `authSchemeOptions` method.
- Implemented the remaining endpoint-based auth scheme resolution logic, picking up where the previous PR left off.
  - Merged modeled auth scheme option properties when constructing endpoint-resolved auth scheme options.
  - Although the functionality is minimal, started incorporating missing [SigningOptions](https://github.com/smithy-lang/smithy-rs/blob/c5013b949153efafa376f84b8d44968ce77ea61c/aws/rust-runtime/aws-runtime/src/auth.rs#L45) fields when handling `AuthSchemeEndpointConfig` in the `aws-runtime` crate.
  - Applied auth scheme option properties to the config bag in the orchestrator.

## Testing
- CI
- Disabled client protocol tests referencing SigV4 in the model
The set of codegen decorators used during client protocol tests does not include `SigV4AuthDecorator`. Since the updated implementation no longer uses ["fall-back-to-no-auth' logic](https://github.com/smithy-lang/smithy-rs/blob/c624a84d9ecc451854521e363e34cf6f1c10e009/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/AuthOptionsPluginGenerator.kt#L79-L86) per-operation, we need inject a no-auth auth scheme resolver into the client config during test setup for affected protocol tests.

- Disabled some tests due to the removal of `AddOptionalAuth` transformer from `S3Decorator`
Users who depended on the behavior of this transformer currently use the [no_credentials](https://docs.rs/aws-config/latest/aws_config/struct.ConfigLoader.html#method.no_credentials) method to bypass signing. Once we support configuring the auth scheme resolver on the client config builder, we will need to update the implementation of `no_credentials` to enable the tests in question.

----

### DIFF
--- a/aws/rust-runtime/aws-inlineable/src/endpoint_auth.rs
+++ b/aws/rust-runtime/aws-inlineable/src/endpoint_auth.rs
@@ -34,7 +34,7 @@ pub(crate) async fn resolve_endpoint_based_auth_scheme_options<'a>(
     let mut endpoint_auth_scheme_ids = Vec::new();
 
     // Note that we're not constructing the `properties` for `endpoint_auth_schemes` here—only collecting
-    // auth scheme IDs. This is because, at this stage, we're only determining which auth scheme to use.
+    // auth scheme IDs but not properties. This is because, at this stage, we're only determining which auth scheme to use.
     // Any `authSchemes` list properties that influence the signing context will be extracted later
     // in `AuthSchemeEndpointConfig`, and passed by the orchestrator to the signer's `sign_http_request` method.
     if let Some(aws_smithy_types::Document::Array(endpoint_auth_schemes)) =
@@ -60,6 +60,9 @@ pub(crate) async fn resolve_endpoint_based_auth_scheme_options<'a>(
 
 // Returns a list of merged auth scheme options from `modeled_auth_scheme_options` and `endpoint_auth_scheme_ids`,
 // copying properties from the modeled auth scheme options into the endpoint auth scheme options as they are built.
+//
+// Note: We only extract properties from the modeled auth schemes. Pulling properties from the endpoint auth schemes
+// would result in duplication; they would be added here and again in the `extract_operation_config` function during signing.
 fn merge_auth_scheme_options(
     modeled_auth_scheme_options: &[AuthSchemeOption],
     endpoint_auth_scheme_ids: Vec<AuthSchemeId>,
--- a/aws/rust-runtime/aws-runtime/src/auth.rs
+++ b/aws/rust-runtime/aws-runtime/src/auth.rs
@@ -143,6 +143,8 @@ pub struct SigV4OperationSigningConfig {
     pub signing_options: SigningOptions,
 }
 
+// TODO(AuthRefactoring): Consider implementing a dedicated struct, similar to `MergeTimeoutConfig`, that allows
+// us to implement a custom merge logic for `impl Store`, enabling fold-style merging of `SigV4OperationSigningConfig`.
 impl Storable for SigV4OperationSigningConfig {
     type Storer = StoreReplace<Self>;
 }


aws/rust-runtime/aws-inlineable/src/endpoint_auth.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    let mut endpoint_auth_scheme_options = endpoint_auth_scheme_ids
+        .into_iter()
+        .map(|id| {
+            let modelded = common_auth_scheme_options
+                .iter()
+                .find(|opt| opt.scheme_id() == &id)
+                .cloned();
+            let mut builder = AuthSchemeOption::builder().scheme_id(id);
+            builder.set_properties(modelded.and_then(|m| m.properties()));
```
aajtodd, line 85: **correctness**: Is this right? It looks like it's only taking the properties from the modeled auth schemes instead of merging the properties (with endpoint options taking precedence). i.e. only merging properties that only exist on modeled but not on endpoint.

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
```diff
@@ -3,11 +3,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-use crate::auth;
```
aajtodd, line 6: General comment, I'm a little concerned with how much sigv4/sigv4a and also in general endpoint based auth is tied to `aws/rust-runtime` instead of the generic smithy runtime. I realize some of this was already here but it's going to cause problems at some point I think.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
```diff
@@ -66,45 +64,15 @@ class SigV4AuthDecorator : ConditionalDecorator(
             override val name: String get() = "SigV4AuthDecorator"
```
aajtodd, line 62: Can't comment on the actual line but the docs for usesSigv4a are out of date, sigv4a has a trait now.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
```diff
@@ -209,21 +205,6 @@ class FilterEndpointTests(
         }
 }
 
-// TODO(P96049742): This model transform may need to change depending on if and how the S3 model is updated.
-private class AddOptionalAuth {
```
aajtodd, line 213: Why are we removing this? What was our plan again to keep s3 working without auth (since we shipped it this way)?

codegen-client-test/build.gradle.kts
```diff
@@ -61,31 +61,37 @@ data class ClientTest(
 
 val allCodegenTests = listOf(
     ClientTest("com.amazonaws.simple#SimpleService", "simple", dependsOn = listOf("simple.smithy")),
+    /*  TODO(AuthAlignment): Comment in once the no auth scheme resolver can be injected via client config
```
aajtodd, line 64: Wouldn't test credentials still work? How does this work now?

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -151,6 +151,10 @@ pub(super) async fn resolve_identity(
                         } else {
                             IdentityCache::no_cache()
                         };
+                        // Apply properties from the selected auth scheme option
+                        if let Some(properties) = options[i].properties() {
```
aajtodd: why doesn't `auth_scheme_option.properties()` work? (why is enumerate necessary I guess).

aws/rust-runtime/aws-inlineable/src/endpoint_auth.rs
ysaito1001: Yes it is (I'll update source code comments to further clarify). As the comment above says
```
Any `authSchemes` list properties that influence the signing context will be extracted later
in `AuthSchemeEndpointConfig`, and passed by the orchestrator to the signer's `sign_http_request` method.
```
endpoint auth scheme properties will be handled in the [extract_operation_config](https://github.com/smithy-lang/smithy-rs/blob/54d26cbb38364842ed6ea190bd743556a540d3ed/aws/rust-runtime/aws-runtime/src/auth/sigv4.rs#L121-L144) function (endpoint properties take precedence there). Regardless of the type of auth scheme resolver being used, `authSchemes` endpoint list property is handled in a single-source manner. For this reason, the function `resolve_endpoint_based_auth_scheme_options` simply decides which auth scheme to use (plus, preserving modeled auth scheme option properties).

Admittedly, it'd be great if we could just simply merge layers to merge fields for [SigV4OperationSigningConfig](https://github.com/smithy-lang/smithy-rs/blob/54d26cbb38364842ed6ea190bd743556a540d3ed/aws/rust-runtime/aws-runtime/src/auth.rs#L130), but that struct has a nested structure, and some fields are at the top level like `SigningName` but other fields such as `double_uri_encode` are nested in another struct `SigningOptions`. If the config bag had another concept like `StoreFold`, then we might be able to just merge `SigV4OperationSigningConfig` as we merge layers but `SigV4OperationSigningConfig` is already exposed as `StoreReplace`. Alternatively, we may be able to introduce a new struct like [MergeTimeOutConfig](https://github.com/smithy-lang/smithy-rs/blob/36a82ddd57c210f9ac64efef201b852b2c0cf7d3/rust-runtime/aws-smithy-types/src/timeout.rs#L311-L316) that has a custom merge logic for `impl Store`.

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
ysaito1001: > concerned with how much sigv4/sigv4a is tied to `aws/rust-runtime`
Right, as we discussed for decorators.

> general endpoint based auth is tied to `aws/rust-runtime`
Don't we discourage the use of endpoint based auth and only currently allowed AWS services should be able to use the endpoint based auth functionality? Will moving endpoint based auth to generic smithy runtime widen the audience, in contrast to where we want it to be?

ysaito1001: > concerned with how much sigv4/sigv4a is tied to `aws/rust-runtime`

Right, as we discussed for decorators.

> concerned with how much general endpoint based auth is tied to `aws/rust-runtime`

Don't we discourage the use of endpoint based auth and only currently allowed AWS services should be able to use the endpoint based auth functionality? From a perspective of narrowing usages, isn't endpoint based auth living in `aws/rust-runtime` a good thing?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
ysaito1001: Those who rely on this behavior currently call the [no_credentials](https://docs.rs/aws-config/latest/aws_config/struct.ConfigLoader.html#method.no_credentials) method to bypass signing. As long as we make `no_credentials` continue working, we won't break customers (I already enabled `no_auth` tests in my next feature branch).

If we can avoid a model transformer, that's preferable.

P.S.
I'm considering to restore the transformer, given that if a customer directly constructs a S3 client config without going through an `SdkConfig` (i.e. no default credentials provider chain) and expects an operation to work without the need for signing, that's a breakage if we remove the transformer.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
ysaito1001: Ah, reminiscent of experiment. It should work that way. Will switch to `auth_scheme_option.properties()`.

Addressed in c5d9bce.

codegen-client-test/build.gradle.kts
ysaito1001: I believe that test credentials pull in `SigV4AuthDecorator` so the `sigv4` is available in the runtime components, if I understand your question correctly?

aws/rust-runtime/aws-runtime/src/auth/sigv4.rs
ysaito1001: > concerned with how much sigv4/sigv4a is tied to `aws/rust-runtime`

Right, as we discussed for decorators.

> concerned with how much general endpoint based auth is tied to `aws/rust-runtime`

While I agree with your concern, don't we discourage the use of endpoint based auth and only currently allowed AWS services should be able to use the endpoint based auth functionality? From a perspective of narrowing usages, isn't endpoint based auth living in `aws/rust-runtime` a good thing?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
ysaito1001: Removed stale comment in a924d8f

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
landonxjames: >I'm considering to restore the transformer,

Could we find a list of S3 operations that actually support no-auth and scope it to that? I feel like its just `get-object`, `put-object`, `head-bucket`, and maybe a few others? 

Edit: I searched internally and externally for a simple list and could not find one, so maybe it isn't possible.

aws/rust-runtime/aws-inlineable/src/endpoint_auth.rs
landonxjames: >Alternatively, we may be able to introduce a new struct like MergeTimeOutConfig that has a custom merge logic for impl Store.

I might prefer this? Or maybe just renaming this `fn` to something like `extract_modeled_auth_schemes`, since it isn't doing the full merge. Either way it feels confusing to have that logic split up.

aws/rust-runtime/aws-inlineable/src/endpoint_auth.rs
```diff
@@ -0,0 +1,190 @@
...
+// copying properties from the modeled auth scheme options into the endpoint auth scheme options as they are built.
+fn merge_auth_scheme_options(
+    modeled_auth_scheme_options: &[AuthSchemeOption],
+    endpoint_auth_scheme_ids: Vec<AuthSchemeId>,
+) -> Vec<AuthSchemeOption> {
+    let (common_auth_scheme_options, model_only_auth_scheme_options): (Vec<_>, Vec<_>) =
+        modeled_auth_scheme_options
+            .iter()
+            .partition(|auth_scheme_option| {
+                endpoint_auth_scheme_ids.contains(auth_scheme_option.scheme_id())
```
landonxjames, line 74: The `contains` call is `O(n)` and we are calling it on each item in `modeled_auth_scheme_options` so this could end up as `O(mn)`. I imagine the list of available auth schemes are small enough that this isn't going to matter in practice?

ysaito1001: > small enough that this isn't going to matter in practice?

Exactly. The number of auth schemes present in a service is small. For a small number, a contiguous data structure like `Vec` might even outperform a node-based data structure like `HashMap` (although we don't know for sure until we measure it). But `doesn't matter in practice` is the case here.

aws/rust-runtime/aws-inlineable/src/endpoint_auth.rs
ysaito1001: I'll add `TODO` on `SigV4OperationSigningConfig` to consider using a custom merge logic for `impl Store`. BUT, it occurred to me that even with the custom merge logic for `SigV4OperationSigningConfig`, we should only take the properties from the modeled auth schemes within `resolve_endpoint_based_auth_scheme_options`, because taking the properties from the endpoint auth schemes here will end up adding them twice in the config bag; once in `resolve_endpoint_based_auth_scheme_options` and twice in the `extract_operation_config` function (the latter will always be used regardless of whether endpoint auth scheme option logic is in effect or not).

Added comments and TODO in 706f20a

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
ysaito1001: Will definitely come back to this and assess the risk of removing the transformer.

ysaito1001:
Will come to unanswered questions (like s3 transformer) in a subsequent PR.

ysaito1001:
Will come back to unanswered questions (like s3 transformer) in a subsequent PR.

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/customize/s3/S3Decorator.kt
aajtodd: We should _not_ conflate `no_credentials` with this. All that `no_credentials` does is set the credential provider to explicitly unset. S3 works with no credentials due to the previous customization, **not** because `no_credentials` was called.

codegen-client-test/build.gradle.kts
aajtodd: I was asking why we need to inject no auith scheme resolver in the first place when test credentials would work?

PR #4176, opened 2025-06-12
Add allow http and cert verifier configs to http client

kevinzwang:
## Motivation and Context
I work on [Daft](https://github.com/Eventual-Inc/Daft), which uses the AWS SDK and Smithy to read data from S3. We expose certain S3 configurations to allow them to enforce HTTPS or disable SSL verification.

Using the legacy HTTP client in `aws-smithy-http-client`, these were possible to implement because we could pass in [our own `hyper-tls` connector](https://github.com/Eventual-Inc/Daft/blob/71790a235c99c62ddbd01fa0c320d65487b02031/src/daft-io/src/s3_like.rs#L438-L439). However, the new HTTP client builder does not expose ways to do this.

## Description
These new configurations are added to the `aws-smithy-http-client` crate:
- An optional custom cert verifier to `crate::tls::rustls_provider::Provider::Rustls`
- A default-true `allow_http` parameter to `crate::tls::rustls_provider::TlsContext`

## Testing
none (let me know if there are tests that I could write, to my knowledge this is a pretty small change that should not modify existing behavior)

## Checklist

----

### DIFF
--- a/.changelog/change.md
+++ b/.changelog/change.md
@@ -1,0 +1,11 @@
+---
+applies_to:
+- client
+authors:
+- kevinzwang
+references: []
+breaking: false
+new_feature: false
+bug_fix: false
+---
+Added allow http and cert verifier configs to http client
--- a/rust-runtime/aws-smithy-http-client/examples/client-aws-lc.rs
+++ b/rust-runtime/aws-smithy-http-client/examples/client-aws-lc.rs
@@ -12,12 +12,12 @@ use aws_smithy_http_client::{
 async fn main() {
     // feature = rustls-aws-lc
     let _client = Builder::new()
-        .tls_provider(tls::Provider::Rustls(CryptoMode::AwsLc))
+        .tls_provider(tls::Provider::rustls(CryptoMode::AwsLc))
         .build_https();
 
     // feature = rustls-aws-lc-fips
     // A FIPS client can also be created. Note that this has a more complex build environment required.
     let _client = Builder::new()
-        .tls_provider(tls::Provider::Rustls(CryptoMode::AwsLcFips))
+        .tls_provider(tls::Provider::rustls(CryptoMode::AwsLcFips))
         .build_https();
 }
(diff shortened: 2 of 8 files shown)


ysaito1001:
Hi, thank you for the contribution! I have two observations/questions:

1. The public method `rustls_danger_custom_cert_verifier` exposes a type `ServerCertVerifier` from an unstable crate (rustls 0.23, which is version < 1.0). Since `aws-smithy-http-client` is a stable crate (version >= 1.0), types in method signatures cannot originate from unstable crates. Such a dependency would introduce a potential API breakage for `aws-smithy-http-client` if the unstable crate changes. Furthermore, to ensure API can evolve, the usage of `std::sync::Arc`, as  in `Arc<dyn ServerCertVerifier>`, should be abstracted behind a new type. You can see types `Shared...` (e.g. [SharedHttpClient](https://github.com/smithy-lang/smithy-rs/blob/36a82ddd57c210f9ac64efef201b852b2c0cf7d3/rust-runtime/aws-smithy-runtime-api/src/client/http.rs#L182)) are used in the codebase to hide `Arc`.

2. Looks like `rustls::client::danger::ServerCertVerifier` is not feature-gated in the `rustls` crate, but I'm wondering if we should in `aws-smithy-http-client` crate; otherwise the `rustls_danger_custom_cert_verifier` would be compiled into binary for a default cargo build 🤔

aajtodd:
Thanks for the contribution but I'm going to go ahead and close this. We can't accept this for a few reasons:
 1. We can't expose unstable 3rd party crate types in a stable crate (nor do we want to really expose 3P types at all here)
 2. We only plan to support common configuration that works across TLS providers we support consistently
 

This crate provides the default HTTPS client, it isn't meant to solve all use cases but rather to work well for a large majority. If you need very specific or lesser used HTTP/TLS configuration not provided our recommendation is to provide your own HTTP client configured as you wish using the provided HTTP traits expected on `SdkConfig`.

PR #4181, opened 2025-06-18
Http-1x update phase 1

landonxjames:
## Motivation and Context

The first steps to defaulting to `http-1x` dependencies instead of `http-02x`

## Description

For this PR I updated the dependencies on SDK crates to enable their `http-1x` features and worked backwards from there fixing the things that broke (this would be easier to see if the diff was working but it seems to be broken for feature branches after the new ECR update). This involved:
* Updating the `http_request_checksum` inlineable to use `http-1x` types (and removing the `http-1x` feature since it is no longer needed and wasn't a "real" feature to begin with
* Creating a new `pub mod content_encoding_http_1x` module in the `aws-runtime` crate. Most logic here is dedicated to implementing the `http_body_1x::Body` trait for `AwsChunkedBody`. I initially considered keeping this in the existing `content_encoding` module, but with the current feature flags and all of the changes I had to make it felt cleaner to break it out.
* Update body polling logic for `ChecksumBody` calculation and validation for `http-1x` types

## Testing

Updated existing tests and added new ones for `http-1x` types

## TODOs
Future PRs will
* Update code generation logic in codegen-core to use `http-1x`
* Remove pre-1.0 `http` dependencies (anywhere they aren't re-exported)
* Normalize dependency names to `http-1x`, `http-02x`, etc

----


aws/rust-runtime/aws-runtime/src/content_encoding_http_1x.rs
```diff
@@ -0,0 +1,648 @@
...
+                        *this.inner_body_bytes_read_so_far += data.len();
+                        Poll::Ready(Some(Ok(frame)))
+                    } else {
+                        tracing::trace!(
+                            "No more chunk data, writing CRLF + CHUNK_TERMINATOR to end the data, and the first trailer frame"
+                        );
+                        *this.state = AwsChunkedBodyState::WritingTrailers;
+                        let trailers = frame.trailers_ref();
+
+                        // NOTE: there is a subtle logic bug here (which is present in the http-02x implementation as well)
```
landonxjames: Open to fixing this in this PR (or in this feature branch at least), leaving for now mostly to motivate discussion, since this logic already exists in the current implementation.

aws/rust-runtime/aws-runtime/src/content_encoding_http_1x.rs
```diff
@@ -0,0 +1,648 @@
...
+/// Content encoding header value constants
+pub mod header_value {
+    /// Header value denoting "aws-chunked" encoding
+    pub const AWS_CHUNKED: &str = "aws-chunked";
+}
+
+/// Options used when constructing an [`AwsChunkedBody`].
+#[derive(Debug, Default)]
+#[non_exhaustive]
+pub struct AwsChunkedBodyOptions {
```
landonxjames: I couldn't decide if this deserved a full builder, but since it is `pub` it needs to be extensible to allow for some upcoming work on `AwsChunked`, one aspect of which will likely be making the chunked body size configurable. I settled for making the API builderish, but I am open to making a full builder if anyone has strong opinions.

landonxjames (commented):
Couple of pointers for discussions I would like to have.

aws/rust-runtime/aws-inlineable/src/presigning.rs
```diff
@@ -241,13 +241,11 @@ impl PresignedRequest {
             .map(|_req| body)
     }
 
-    #[cfg(feature = "http-1x")]
```
aajtodd, line 244: Do we not want this feature gated still?

aws/rust-runtime/aws-runtime/src/content_encoding_http_1x.rs
aajtodd: It seems odd to me that we have two `AwsChunkedBodyOptions` and duplicating all of this logic. Is there not a way to just separate out the http-02x and http-1x specific parts and make a more coherent/single implementation?

aws/rust-runtime/aws-runtime/src/content_encoding_http_1x.rs
aajtodd: I'd at least make it a dedicated PR for this feature branch or just open an issue and leave it for the backlog.

rust-runtime/aws-smithy-checksums/Cargo.toml
```diff
@@ -14,12 +14,14 @@ repository = "https://github.com/smithy-lang/smithy-rs"
 
 [dependencies]
 aws-smithy-http = { path = "../aws-smithy-http" }
-aws-smithy-types = { path = "../aws-smithy-types" }
+aws-smithy-types = { path = "../aws-smithy-types", features = ["http-body-1-x"]}
 bytes = "1.10.0"
 crc-fast = { version = "1.2.1" }
 hex = "0.4.3"
 http = "0.2.9"
```
aajtodd: Are we not able to remove this (and http-body:0.4.5) completely?

rust-runtime/aws-smithy-checksums/src/http.rs
```diff
@@ -47,16 +47,32 @@ pub trait HttpChecksum: Checksum + Send + Sync {
...
+        header_map.insert(self.header_name(), self.header_value_http_1x());
+
+        header_map
+    }
+
     /// Return the `HeaderName` used to represent this checksum algorithm
     fn header_name(&self) -> &'static str;
 
-    /// Return the calculated checksum as a base64-encoded `HeaderValue`
+    /// Return the calculated checksum as a base64-encoded http-02x `HeaderValue`
```
aajtodd, line 52: Given the internal facing nature of this crate (and the 0.x versioning) can we remove the legacy http all together or no?

aajtodd (approved):
Overall looks good. Couple questions though given the PR description on next steps some of them may simply be planned for future PRs already. Only thing I didn't see mentioned for future work was also changing the default SdkBody internals to http-body 1.x

aws/rust-runtime/aws-runtime/src/content_encoding_http_1x.rs
ysaito1001: wondered about that too if it can be moved to a common location if the struct does not contain http-02x/http-1x bifurcation 🤔

aws/rust-runtime/aws-inlineable/src/presigning.rs
ysaito1001: maybe the removal is tied to [this change](https://github.com/smithy-lang/smithy-rs/pull/4181/files#diff-0abc3a955199091fcd2616bc18494aebbdb7029a717c7cf9f0cdb824a9b1f989R173)?

ysaito1001 (approved):
There is probably room for refactoring, but looks great for the first stab!

aws/rust-runtime/aws-runtime/src/content_encoding_http_1x.rs
landonxjames: Tracking in https://github.com/smithy-lang/smithy-rs/issues/4188

aws/rust-runtime/aws-runtime/src/content_encoding_http_1x.rs
landonxjames: I ended up refactoring this back to a single module in https://github.com/smithy-lang/smithy-rs/pull/4181/commits/52e80bc82f3786cb647b591cc55bc87762120501 (with a smaller follow up in https://github.com/smithy-lang/smithy-rs/pull/4181/commits/b231ea4a6b77a1c20c2a622d883363b1874b06b2). As part of this I removed the `optional = true` from the `http-1x` crates and made them no longer attached to the `http-1x` feature on the `aws-runtime` crate.

aws/rust-runtime/aws-inlineable/src/presigning.rs
landonxjames: I do not think it needs to be, since as Yuki linked to we are now including `http-1x` dependency unconditionally

rust-runtime/aws-smithy-checksums/Cargo.toml
landonxjames: Removed old http crates from aws-smithy-checksums in https://github.com/smithy-lang/smithy-rs/pull/4181/commits/36d967dca47611fa5ec2637ef0c273ae85b5245b

rust-runtime/aws-smithy-checksums/src/http.rs
landonxjames: Removed in https://github.com/smithy-lang/smithy-rs/pull/4181/commits/36d967dca47611fa5ec2637ef0c273ae85b5245b

landonxjames:
The semver break here is expected since we are removing some Trait impls from `aws-smithy-checksums`. That is fine since the crate is pre-1.0

PR #4182, opened 2025-06-19
Make auth scheme and auth sheme option resolver configurable

ysaito1001:
## Testing

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialProvidersDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialProvidersDecorator.kt
@@ -39,6 +39,16 @@ class CredentialsProviderDecorator : ConditionalDecorator(
                 listOf(
                     adhocCustomization<SdkConfigSection.CopySdkConfigToClientConfig> { section ->
                         rust("${section.serviceConfigBuilder}.set_credentials_provider(${section.sdkConfig}.credentials_provider());")
+                        rustTemplate(
+                            """
+                            if ${section.sdkConfig}.credentials_provider().is_none() {
+                                ${section.serviceConfigBuilder}.set_auth_scheme_option_resolver(#{no_auth_scheme_option_resolver}());
+                            }
+                            """,
+                            "no_auth_scheme_option_resolver" to
+                                ClientRustModule.Config.auth.toType()
+                                    .resolve("no_auth_scheme_option_resolver"),
+                        )
                     },
                 )
 
--- a/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
+++ b/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
@@ -11,9 +11,13 @@ import software.amazon.smithy.rust.codegen.client.smithy.ClientCodegenContext
 import software.amazon.smithy.rust.codegen.client.smithy.customize.ClientCodegenDecorator
 import software.amazon.smithy.rust.codegen.client.smithy.generators.ServiceRuntimePluginCustomization
 import software.amazon.smithy.rust.codegen.client.smithy.generators.ServiceRuntimePluginSection
+import software.amazon.smithy.rust.codegen.client.smithy.generators.config.ConfigCustomization
+import software.amazon.smithy.rust.codegen.client.smithy.generators.config.ServiceConfig
 import software.amazon.smithy.rust.codegen.core.rustlang.Writable
 import software.amazon.smithy.rust.codegen.core.rustlang.rustTemplate
 import software.amazon.smithy.rust.codegen.core.rustlang.writable
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType
+import software.amazon.smithy.rust.codegen.core.smithy.RuntimeType.Companion.preludeScope
 import software.amazon.smithy.rust.codegen.core.smithy.customize.NamedCustomization
 import software.amazon.smithy.rust.codegen.core.smithy.customize.Section
 
@@ -48,11 +52,17 @@ typealias AuthCustomization = NamedCustomization<AuthSection>
 /**
  * Codegen decorator that:
  * * Injects service-specific default auth scheme resolver
- * * TODO(AuthAlignment): Expand this list as we add more customizations
+ * * Adds setters to the service config builder for configuring the auth scheme and auth scheme option resolver
+ * * Adds getters to the service config for retrieving the currently configured auth scheme and auth scheme option resolver
  */
 class AuthDecorator : ClientCodegenDecorator {
     override val name: String = "Auth"
     override val order: Byte = 0
+
+    override fun configCustomizations(
+        codegenContext: ClientCodegenContext,
+        baseCustomizations: List<ConfigCustomization>,
+    ): List<ConfigCustomization> = baseCustomizations + AuthDecoratorConfigCustomizations(codegenContext)
 
     override fun serviceRuntimePluginCustomizations(
         codegenContext: ClientCodegenContext,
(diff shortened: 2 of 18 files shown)


codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
```diff
@@ -88,3 +98,242 @@ private fun defaultAuthSchemeResolver(codegenContext: ClientCodegenContext): Wri
...
+                        ///
+                        /// If `auth_schemes` contains an existing [AuthSchemeId](aws_smithy_runtime_api::client::auth::AuthSchemeId) in the runtime, the current identity
+                        /// resolver and signer for that scheme will be replaced by those from `auth_schemes`.
+                        ///
+                        /// _Important:_ When introducing a custom auth scheme, ensure you override either
+                        /// [`Self::auth_scheme_option_resolver`] or [`Self::set_auth_scheme_option_resolver`]
+                        /// so that the custom auth scheme is included in the list of resolved auth scheme options.
+                        /// [The default auth scheme resolver](crate::config::auth::DefaultAuthSchemeResolver) will not recognize your custom auth scheme. Refer to an
+                        /// example for [`Self::push_auth_scheme`].
+                        pub fn extend_auth_schemes(
```
aajtodd: Do we need this? Seems like a user can do this on their own.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
```diff
@@ -88,3 +98,242 @@ private fun defaultAuthSchemeResolver(codegenContext: ClientCodegenContext): Wri
...
+                        ///
+                        /// If `auth_schemes` contains an existing [AuthSchemeId](aws_smithy_runtime_api::client::auth::AuthSchemeId) in the runtime, the current identity
+                        /// resolver and signer for that scheme will be replaced by those from `auth_schemes`.
+                        ///
+                        /// _Important:_ When introducing a custom auth scheme, ensure you override either
+                        /// [`Self::auth_scheme_option_resolver`] or [`Self::set_auth_scheme_option_resolver`]
+                        /// so that the custom auth scheme is included in the list of resolved auth scheme options.
+                        /// [The default auth scheme resolver](crate::config::auth::DefaultAuthSchemeResolver) will not recognize your custom auth scheme. Refer to an
+                        /// example for [`Self::push_auth_scheme`].
+                        pub fn set_auth_schemes(
```
aajtodd: Do we want to allow replacing all auth schemes? This seems like a very sharp tool and one that I question we should offer.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
```diff
@@ -88,3 +98,242 @@ private fun defaultAuthSchemeResolver(codegenContext: ClientCodegenContext): Wri
...
+                        /// ##      todo!()
+                        ///     }
+                        /// }
+                        ///
+                        /// let config = $moduleUseName::Config::builder()
+                        ///     .auth_scheme_option_resolver(CustomAuthSchemeResolver)
+                        ///     // other configurations
+                        ///     .build();
+                        /// ```
+                        pub fn auth_scheme_option_resolver(mut self, auth_scheme_option_resolver: impl #{ResolveAuthSchemeOptions} + 'static) -> Self {
```
aajtodd: **suggestion**: I think I'd prefer just `auth_scheme_resolver` personally.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthSchemeResolverGenerator.kt
```diff
@@ -206,4 +211,31 @@ class AuthSchemeResolverGenerator(
             )
         }
     }
+
+    private fun noAuthSchemeOptionResolver() =
+        writable {
+            rustTemplate(
+                """
+                ##[derive(Debug)]
+                struct NoAuthSchemeOptionResolver;
```
aajtodd: 1. Why not just place this in the runtime as a re-usable type
2. Why do we need a public `no_auth_scheme_option_resolver` function? Who is using it and why?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ClientProtocolTestGenerator.kt
```diff
@@ -189,12 +190,30 @@ class ClientProtocolTestGenerator(
                     }
                 }
             } ?: writable { }
+        // TODO(https://github.com/smithy-lang/smithy-rs/issues/4177):
+        //  Until the incorrect separation is addressed, we need to rely on this workaround.
+        val noAuthSchemeOptionResolver =
```
aajtodd: Allowing protocol tests to support no auth seems fine but I'm wondering why `test_credentials` wouldn't work?

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialProvidersDecorator.kt
```diff
@@ -39,6 +39,16 @@ class CredentialsProviderDecorator : ConditionalDecorator(
                 listOf(
                     adhocCustomization<SdkConfigSection.CopySdkConfigToClientConfig> { section ->
                         rust("${section.serviceConfigBuilder}.set_credentials_provider(${section.sdkConfig}.credentials_provider());")
+                        rustTemplate(
+                            """
+                            if ${section.sdkConfig}.credentials_provider().is_none() {
+                                ${section.serviceConfigBuilder}.set_auth_scheme_option_resolver(#{no_auth_scheme_option_resolver}());
```
aajtodd: **fix**: I don't think this is how we should handle no auth. If credentials aren't set it doesn't necessarily mean no auth should be the resolver, e.g. a service that supports multiple authentication schemes (sigv4 + bearer) it would be perfectly acceptable to have credentials provider be `None` as the `token_provider` may be set.

This is coupling identity resolvers with auth scheme resolution, when they need to remain independent. Users MUST be able to configure 0 or more identity resolvers (credentials provider, token provider, etc) or unset them (e.g. `no_credentials`) separate from their auth scheme resolver config. 

---

Why are we setting a `no_auth` auth scheme resolver here? What are we solving, once we answer that we can figure out a different approach.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
ysaito1001: Was originally mirroring APIs like [extend_retry_classifiers](https://docs.rs/aws-sdk-s3/1.93.0/aws_sdk_s3/config/struct.RuntimeComponentsBuilder.html#method.extend_retry_classifiers) or [extend_interceptors](https://docs.rs/aws-sdk-s3/1.93.0/aws_sdk_s3/config/struct.RuntimeComponentsBuilder.html#method.extend_interceptors), but you know what, they are unused, so let's remove this. We can always add later. Good catch.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
ysaito1001: I should probably clarify the comment. This setter replaces all schemes configured just for service config's `Layer`. Other auth schemes configured, such as those in service runtime plugins or those in operation runtime plugins, won't be replaced. Hopefully you now find it not as sharp as you might think it is.

But, as I'm writing this, it makes me wonder if we need this API like `extend_auth_schemes` discussed above.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthSchemeResolverGenerator.kt
ysaito1001: 1. Because `NoAuthSchemeResolver` has to implement code-generated trait `ResolveAuthScheme`. We could place it in the runtime without implementing the trait, but I'd find it not useful.
2. It's for encapsulation (similar to [NoCache](https://github.com/awslabs/aws-sdk-rust/blob/c689eba19d331b83d0ea19148b6a8abafbe3adbe/sdk/aws-smithy-runtime/src/client/identity/cache.rs#L80)). It's safer for a creation function to return an opaque type if possible. Whoever wants to disable signing can use
```
config_builder.auth_scheme_resolver(no_auth_scheme_resolver())
```

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
ysaito1001: Suggestion taken. SRA names the component as auth scheme resolver. While we've already exposed some types from `aws_smithy_runtime_api` such as `SharedAuthSchemeOptionResolver`, we want to give the consistent name from the design wherever possible.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/protocol/ClientProtocolTestGenerator.kt
ysaito1001: The reason `test_credentials` doesn't work is that `Sigv4Decorator` is missing while running protocol tests (from `codegen/client` as opposed to `sdk-codegen`), so even though `test_credentials` puts an identity (i.e. AWS credentials) for sigv4, the orchestrator doesn't recognize sigv4 [in this loop](https://github.com/smithy-lang/smithy-rs/blob/dfd53fb59632be8e4fd8c349f9e2d4ee59255997/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs#L141).

Things used to work before the auth alignment because of [this fallback](https://github.com/smithy-lang/smithy-rs/blob/dfd53fb59632be8e4fd8c349f9e2d4ee59255997/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/generators/AuthOptionsPluginGenerator.kt#L79-L85).

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
ysaito1001: The method removed in 4653240

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
ysaito1001: The method removed in https://github.com/smithy-lang/smithy-rs/commit/465324079a58603eedbd73480581054e0aeb9362

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
ysaito1001: Renamed in 24da2cd

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/CredentialProvidersDecorator.kt
ysaito1001: Removed this code altogether as a result of restoring that S3 transformer. Originally this was meant to compensate for the removal of that transformer, but the way it was implemented was incorrect, as you pointed out.

As discussed offline, we now end up introducing a synthetic "no auth" trait that can be applied to the service shape to overcome the limitation `optionalAuth` can only be applied to operations. Done in 5158cd1.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthSchemeResolverGenerator.kt
```diff
@@ -206,4 +211,33 @@ class AuthSchemeResolverGenerator(
...
+                        _params: &'a #{Params},
+                        _cfg: &'a #{ConfigBag},
+                        _runtime_components: &'a #{RuntimeComponents},
+                    ) -> #{AuthSchemeOptionsFuture}<'a> {
+                        #{AuthSchemeOptionsFuture}::ready(#{Ok}(vec![#{NoAuthSchemeOption:W}]))
+                    }
+                }
+
+                /// Return an auth scheme resolver that favors `noAuth`
+                pub fn no_auth_scheme_resolver() -> impl ResolveAuthScheme {
```
aajtodd: I guess it comes down to I'm not a fan of this. This is a niche use case and I'm not sure it's useful to the majority of customers that we should include it by default. If we need it for protocol tests then maybe we just place it there instead.

ysaito1001: We can. Haven't we had some number of customers requesting disabling signing, though (this convenience resolver was brought up in our team design doc, which is why I originally put it here)? Maybe do we ask them to override sigv4 auth scheme with no-signer by themselves instead of providing this resolver?

aajtodd: > Haven't we had some number of customers requesting disabling signing, though

We have though I think many times it was because they wanted to use sigv4 OR something else in which case they need to disable signing and a custom interceptor implemented their auth logic. That would no longer be necessary with the ability to register a custom auth scheme though.

ysaito1001: Gotcha, makes sense. Will place it locally in protocol tests. I prefer fewer `pub` constructs, if possible.

ysaito1001: Addressed in c94eb95

Issue #4076, opened 2025-03-31
Support the `aws.auth#sigv4a` trait


ysaito1001:
While the codegen supports sigv4a today, currently it supports through [endpoint rules](https://github.com/smithy-lang/smithy-rs/blob/00254621bb308516746e3046dc17acd153d2e253/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt#L47-L57) instead of [the trait](https://smithy.io/2.0/aws/aws-auth.html#aws-auth-sigv4a-trait).

One caveat is that our code generator does not respect the order of the auth trait entries within [the auth trait](https://smithy.io/2.0/spec/authentication-traits.html#auth-trait). This causes the incorrect priority list `[sigv4, sigv4a]` to be passed to `StaticAuthSchemeOptionResolver` even when the service is annotated with `@auth([sigv4a, sigv4])`.

Completing this task, among others, requires a fix for the code generator to respect the auth trait entries.

rcoh:
@ysaito1001 is this done?

PR #4183, opened 2025-06-20
Support SigV4A auth trait

ysaito1001:
## Motivation and Context
#4076 

## Description
Prior to this PR, the SDK determined support for SigV4A authentication based on the presence of the `authSchemes` list property in endpoint rules. With the introduction of the `aws.auth#sigv4a` trait in the model, the code generator should instead rely on the presence of this trait to determine SigV4A support.

However, some services include sigv4a in their endpoint authSchemes list without actually modeling the `aws.auth#sigv4a` trait. Simply switching to trait-based detection would break these services. To maintain backward compatibility, this PR introduces a special-case decorator, `Sigv4aAuthTraitBackfillDecorator` (ported from [the Kotlin SDK](https://github.com/awslabs/aws-sdk-kotlin/blob/a9c1922f2da871148500edb40d13859109f27606/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/SigV4AsymmetricTraitCustomization.kt)), which augments the model by injecting the sigv4a trait for affected services. 

## Testing
- CI

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
@@ -12,13 +12,10 @@ import software.amazon.smithy.model.knowledge.ServiceIndex
 import software.amazon.smithy.model.shapes.OperationShape
 import software.amazon.smithy.model.shapes.ServiceShape
 import software.amazon.smithy.model.shapes.ShapeId
-import software.amazon.smithy.rulesengine.language.EndpointRuleSet
-import software.amazon.smithy.rulesengine.traits.EndpointRuleSetTrait
 import software.amazon.smithy.rust.codegen.client.smithy.ClientCodegenContext
 import software.amazon.smithy.rust.codegen.client.smithy.auth.AuthSchemeOption
 import software.amazon.smithy.rust.codegen.client.smithy.customize.ClientCodegenDecorator
 import software.amazon.smithy.rust.codegen.client.smithy.customize.ConditionalDecorator
-import software.amazon.smithy.rust.codegen.client.smithy.endpoint.AuthSchemeLister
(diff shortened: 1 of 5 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/SigV4AuthDecorator.kt
```diff
@@ -45,15 +42,8 @@ internal fun ClientCodegenContext.usesSigAuth(): Boolean =
...
- * TODO(https://github.com/smithy-lang/smithy-rs/issues/4076): Smithy's `ServiceIndex.getEffectiveAuthSchemes` should be used instead.
- */
-internal fun ClientCodegenContext.usesSigV4a(): Boolean {
-    val endpointAuthSchemes =
-        serviceShape.getTrait<EndpointRuleSetTrait>()?.ruleSet?.let { EndpointRuleSet.fromNode(it) }
-            ?.also { it.typeCheck() }?.let { AuthSchemeLister.authSchemesForRuleset(it) } ?: setOf()
-    return endpointAuthSchemes.contains("sigv4a")
-}
+internal fun ClientCodegenContext.usesSigV4a(): Boolean =
+    ServiceIndex.of(model).getEffectiveAuthSchemes(serviceShape).containsKey(SigV4ATrait.ID)
```
aajtodd: **correctness**: `getEffectiveAuthSchemes` is likely incorrect. It returns only the auth traits in the `@auth([...])` trait if applied. We likely just want `getAuthSchemes` to get all auth traits attached to the service. Same for the `usesSigAuth` function above.

ysaito1001: Thanks for the correctness. Updated in 81e8bc2.

aajtodd (approved):

PR #4190, opened 2025-06-26
fix hyper 1.x connection refused errors not marked as retryable

aajtodd:
## Description
internal ref: `V1820425457`

Fixes a regression in the way connection issues are classified which was causing them to not be retried when using the hyper 1.x client.

The issue is the logic for [classifying the error](https://github.com/smithy-lang/smithy-rs/blob/release-2025-06-11/rust-runtime/aws-smithy-http-client/src/client.rs#L401) was never being hit because connection errors are no longer `hyper::Error` they are `hyper_util::client::legacy::Error`. 

* [hyper-util::Error](https://docs.rs/hyper-util/0.1.14/hyper_util/client/legacy/struct.Error.html)
* [hyper@1.x::Error](https://docs.rs/hyper/1.6.0/hyper/struct.Error.html)
* [hyper@0.14.x::Error](https://docs.rs/hyper/0.14.32/hyper/struct.Error.html#method.is_connect) 

## Testing
New unit test plus verification against internal package that opened the issue.

## Checklist

----

### DIFF
--- a/.changelog/1750957379.md
+++ b/.changelog/1750957379.md
@@ -1,0 +1,12 @@
+---
+applies_to:
+- client
+- aws-sdk-rust
+authors:
+- aajtodd
+references: []
+breaking: false
+new_feature: false
+bug_fix: true
+---
+Fix hyper 1.x connection refused errors not marked as retryable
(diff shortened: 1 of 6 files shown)


landonxjames (approved):

ysaito1001 (approved):
Great finding.

PR #4200, opened 2025-07-02
re-use http checksums on retry attempts

aajtodd:
## Description
A change was needed for the S3 Flexible Checksum SEP to add guidance for reusing checksums during retry attempts to prevent data corruption. When a request fails after checksum calculation is complete, SDKs must save and reuse the checksum for retry attempts rather than recalculating it. This prevents inconsistencies when payload content might change between retries, ensuring data durability in S3.

* Adds a simple cache to the checksum crate that will favor cached checksums from prior attempts if set
* Adds new integration tests to verify retry behavior and re-use of checksums

NOTE: If a user were to actually replace file contents between attempts with a different content length the Rust SDK uses the original content length set on the ByteStream. This happens early when we create the bytestream by taking the user provided content length OR calculating it from the file. We don't ever attempt to recalculate this and I see no great way of doing this. The result is a client side failure about stream length mismatch as opposed to sending the request to the server with the original checksum. 

## Checklist

----

### DIFF
--- a/aws/sdk/integration-tests/s3/tests/checksums.rs
+++ b/aws/sdk/integration-tests/s3/tests/checksums.rs
@@ -489,7 +489,6 @@ impl CapturedRequest {
 
 impl HttpConnector for CaptureConnector {
     fn call(&self, request: HttpRequest) -> HttpConnectorFuture {
-        // let handle = tokio::runtime::Handle::current();
         let captured_requests = self.captured_requests.clone();
         let inner = self.inner.clone();
         HttpConnectorFuture::new(async move {
--- a/rust-runtime/Cargo.toml
+++ b/rust-runtime/Cargo.toml
@@ -12,7 +12,7 @@ members = [
     "aws-smithy-http-auth",
     "aws-smithy-http-client",
     "aws-smithy-http-server",
-#    "aws-smithy-http-server-python",
+    "aws-smithy-http-server-python",
     "aws-smithy-http-tower",
     "aws-smithy-json",
     "aws-smithy-protocol-test",


aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
```diff
@@ -64,6 +67,8 @@ struct RequestChecksumInterceptorState {
     checksum_algorithm: Option<String>,
     /// This value is set in the model on the `httpChecksum` trait
     request_checksum_required: bool,
+    calculate_checksum: Arc<AtomicBool>,
```
landonxjames, line 70: Do we need to wrap the `AtomicBool` in an `Arc`? I thought the `AtomicBool` would be threadsafe on its own?

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
```diff
@@ -207,14 +212,17 @@ where
...
+
         if calculate_checksum {
-            // If a checksum override is set in the ConfigBag we use that instead (currently only used by S3Express)
-            // If we have made it this far without a checksum being set we set the default (currently Crc32)
-            let checksum_algorithm =
-                incorporate_custom_default(checksum_algorithm, cfg).unwrap_or_default();
+            state.calculate_checksum.store(true, Ordering::Release);
 
             // Set the user-agent metric for the selected checksum algorithm
+            // NOTE: We have to do this in modify_before_retry_loop since UA interceptor also runs
```
landonxjames, line 224: Outside the scope of this review, but...

I wish we had a better way to enforce ordering of our own interceptors. For each hook it would be nice to have something like:
```
[ordered array of pre-user interceptors]
...all user interceptors run in arbitrary order...
[ordered array of post-user interceptors]
```
Where only we could set the ordered pre/post interceptors.

aws/sdk/integration-tests/s3/tests/checksums.rs
```diff
@@ -428,3 +444,333 @@ async fn test_response_checksum_ignores_invalid_base64() {
         }
     });
 }
+
+#[derive(Debug, Clone)]
+struct CaptureHttpClient {
```
landonxjames, line 449: This and all the utils below feels like they could potentially be useful enough to live somewhere more accessible under a `test_util` flag? Not necessary now, but don't want to forget this lives here and have someone rewrite something similar down the line.

rust-runtime/Cargo.toml
```diff
@@ -12,7 +12,7 @@ members = [
     "aws-smithy-http-auth",
     "aws-smithy-http-client",
     "aws-smithy-http-server",
-    "aws-smithy-http-server-python",
+#    "aws-smithy-http-server-python",
```
landonxjames: Purposefully removed?

landonxjames (approved):
Looks good, couple of questions but no blockers

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
aajtodd: It does but if I remember right it's because it doesn't implement `Clone` (will double check though)

aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs
aajtodd: Yeah...I found this the hard way when the UA interceptor is registered at the client level but the checksum one is registered later at operation level. No way to enforce which one runs first.

rust-runtime/Cargo.toml
aajtodd: Nope. Good catch. Locally I keep running into PY03 related build issues and I got tired of it and disabled it, forgot to re-enable.

aws/sdk/integration-tests/s3/tests/checksums.rs
```diff
@@ -428,3 +444,333 @@ async fn test_response_checksum_ignores_invalid_base64() {
...
+        })
+}
+
+/// Test checksum is re-used for aws-chunked/streaming content when the content changes
+/// between retries but the content length differs
+///
+/// NOTE: Because we set the overall body size hint only once when the stream is constructed
+///       we end up throwing an error :696
+#[tokio::test]
+#[traced_test]
```
ysaito1001, line 742: I suppose the feature of #[traced_test] is used somewhere?

aws/sdk/integration-tests/s3/tests/checksums.rs
```diff
@@ -428,3 +444,333 @@ async fn test_response_checksum_ignores_invalid_base64() {
...
+        let connector = CaptureConnector {
+            inner,
+            captured_requests: self.captured_requests.clone(),
+        };
+        connector.into_shared()
+    }
+}
+
+#[tokio::test]
+#[traced_test]
```
ysaito1001, line 535: I suppose the feature of `#[traced_test]` is used somewhere?

ysaito1001 (approved):

aws/sdk/integration-tests/s3/tests/checksums.rs
aajtodd: Traced test auto dumps logs on failure IIRC which is how I was using it

aws/sdk/integration-tests/s3/tests/checksums.rs
aajtodd: It might be useful elsewhere but it is making some assumptions about how it's used so I'm not sure I want to make it more general. If it comes up again I agree we might want to consider making a more generalized utility though with whatever limitations documented.

PR #4203, opened 2025-07-10
Support selectable auth schemes

ysaito1001:
~**semver-hazards check failure will be resolved when #4211 is released from `aws-sdk-rust`**~

## Description
Add support for auth scheme preference, which allows customers to re-prioritize the order of auth schemes originally determined by the auth scheme resolver. See [the changelog](https://github.com/smithy-lang/smithy-rs/blob/2d483a0880c30683858329fc8411e9f9282de633/.changelog/1752160009.md) where the auth scheme preference can be configured.

Note: The preference list is merely a hint rather than a strict override. Any auth schemes not included in the originally resolved auth schemes will be ignored, and won't be an error.

~To support the implementation, a utility method has been added to `NowOrLater` to map an inner future.~

## Testing
Those called out in the feature spec:
- [auth_scheme_preference_should_take_the_highest_priority](https://github.com/smithy-lang/smithy-rs/pull/4203/files#diff-b45485e1d1dd937e149c9c703156aa1a7f71b2bbbd5dddb29ffba385b65c50b0R12) for `Alternative Auth Resolution`
- [test_resolve_identity](https://github.com/smithy-lang/smithy-rs/pull/4203/files#diff-db7260e3be202c63463123101814c7bf23ef47796bf23a14d67f567285039d46R891) to verify `Resolving Auth and Credentials`
- [auth scheme preference](https://github.com/smithy-lang/smithy-rs/pull/4203/files#diff-e63831c17293fd55b608a406f472155f8809239f3c3afba20bcd77bdc8da2ea3R140) to verify `Manual auth schemes configuration`
- [test for parsing auth scheme preference from environmental configuration](https://github.com/smithy-lang/smithy-rs/pull/4203/files#diff-d9b0e8d5e74b3e0542908745494b34c1784153ed4e566fe1ce581f8605e801eaR99) 

## Checklist

----


aws/rust-runtime/aws-config/src/default_provider/auth_scheme_preference.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// This function is not automatically extensible. When a new scheme is introduced,
+// it must be manually updated for the scheme to be recognized in the environment configuration.
+//
+// Furthermore, this function does not return a predefined `AuthSchemeId` like `HTTP_BASIC_AUTH_SCHEME_ID`.
+// The predefined `SchemeId`s are gated behind the `http-auth` feature, but this function returns an `AuthSchemeId`
+// that wraps the parsed string regardless of feature flags. If the feature is disabled, the returned
+// auth scheme preference is simply ignored during auth scheme resolution.
+fn runtime_api_auth_scheme_id(auth_scheme_name: &str) -> AuthSchemeId {
```
aajtodd: These scheme ID's don't look right, an `AuthSchemeId` should be the fully qualified name when constructed. It looks like even our hardcoded constants are wrong too 🤔

aajtodd (commented):
Overall looks good

aws/rust-runtime/aws-config/src/default_provider/auth_scheme_preference.rs
ysaito1001: Yeah, unfortunately hard-coded ones predates this PR, and [their hyphenated string values](https://github.com/smithy-lang/smithy-rs/blob/a0801ece565df790431f96a2f718175b8559664a/rust-runtime/aws-smithy-runtime-api/src/client/auth/http.rs#L9-L18) don't match auth scheme's shape ID... Changing those hard-coded strings is a breaking change, so it's a tough spot to be in.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthSchemeResolverGenerator.kt
```diff
@@ -112,7 +115,43 @@ class AuthSchemeResolverGenerator(
...
+                            #{None} => _fut,
+                        }
+                    }
+                }
+
+                impl DefaultAuthSchemeResolver {
+                    /// Set auth scheme preference to the default auth scheme resolver
+                    pub fn with_auth_scheme_preference(mut self, preference: impl #{Into}<#{AuthSchemePreference}>) -> Self {
+                        self.preference = #{Some}(preference.into());
+                        self
```
ysaito1001: As an alternative, we could potentially eliminate the `.with_auth_scheme_preference` public method. Instead, we could rely on the default auth scheme resolver to read the `AuthSchemePreference` from the passed-in config bag in the `resolve_auth_scheme` method.

The upshot is that it works with operation override, i.e.
```
client
    .some_operation()
    .customize()
    .config_override(<sdk_crate>::Config::builder()
        .auth_scheme_preference([...])
    )
    ...
```
The auth scheme preference set at the operation invocation level will be picked up by the default auth scheme resolver.

In contrast, the default auth scheme resolver in this PR, as of this writing, is constructed with the auth scheme preference within the service runtime plugin, which prevents it from detecting auth scheme preferences configured at the operation invocation level.

aws/rust-runtime/aws-config/src/default_provider/auth_scheme_preference.rs
aajtodd: Is it really a breaking change though? Who/how is anyone relying on the actual string values?

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthSchemeResolverGenerator.kt
aajtodd: That or do the filtering/preference application in the orchestrator after resolving the initial auth options list which would allow auth scheme preference to work with custom auth scheme resolvers.

ysaito1001: > do the filtering/preference application in the orchestrator

Aha, I like that!

> which would allow auth scheme preference to work with custom auth scheme resolvers.

Yes, and I should be able eliminate map utilities I added in 842ee41.

aws/rust-runtime/aws-config/src/default_provider/auth_scheme_preference.rs
landonxjames: This strikes me as a breaking change that is probably ok to make. It is very unlikely anyone is using it, and the most likely users would be internal. If we cant find any internal use of it then we can probably get away with it without breaking anyone.

landonxjames (approved):
Nice!

aws/rust-runtime/aws-config/src/default_provider/auth_scheme_preference.rs
ysaito1001: After thinking about this a bit more, I believe the change should be safe to make.

My initial concern was that the raw string values within `AuthSchemeId` are tightly coupled with their `Eq` and `PartialEq`. Since `#[derive(Eq, PartialEq)]` on `AuthSchemeId` relies on the inner string, changing the value from `http-api-key-auth` to `httpApiKeyAuth`, for instance, would cause two `HTTP_API_KEY_AUTH_SCHEME_ID`s coming from different versions of `aws_smithy_runtime_api` to no longer be considered equal. I was worried this mismatch could lead to a runtime error in the orchestrator, where an [auth scheme might not be found](https://github.com/smithy-lang/smithy-rs/blob/7bba9b5dacac1b138cf6b29e92cf0115e9431321/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs#L141) (or [identity resolver might not be found](https://github.com/smithy-lang/smithy-rs/blob/7bba9b5dacac1b138cf6b29e92cf0115e9431321/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs#L143)) due to a failed hash lookup caused by the differing keys.

That said, since within a given crate build tree, there's only one version of the `aws-smithy-runtime-api` crate, and the concern above shouldn't occur in practice.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthSchemeResolverGenerator.kt
ysaito1001: Updated in d07260e

aws/rust-runtime/aws-config/src/default_provider/auth_scheme_preference.rs
ysaito1001: One call site we need to update is [extract_endpoint_auth_scheme_config](https://github.com/smithy-lang/smithy-rs/blob/1a94d1d42ac7bb5e49f04e6b7c24ef4dff9ff3ac/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs#L311). An auth scheme name referenced in the endpoint `authSchemes` list property is just [sigv4](https://github.com/awslabs/aws-sdk-rust/blob/931301e8858bae5fdc8811d1728b546c57eb0923/aws-models/billing.json#L166), not `aws.auth#sigv4`.

Note we also failed to specify the fully qualified auth scheme ID for [sigv4](https://github.com/smithy-lang/smithy-rs/blob/1a94d1d42ac7bb5e49f04e6b7c24ef4dff9ff3ac/aws/rust-runtime/aws-runtime/src/auth/sigv4.rs#L33) (and sigv4a). After we've updated `AuthSchemeId::new("sigv4")` to `AuthSchemeId::new("aws.auth#sigv4")`, the line of code above in `extract_endpoint_auth_scheme_config`
```
config_scheme_id == Some(scheme_id.inner())
```
should be updated to something like
```
config_scheme_id == Some(auth_scheme_name(scheme_id))
```
where `auth_scheme_name` looks something like
```
// For instance, extracts just `sigv4` from `AuthSchemeId("aws.auth#sigv4")`.
// For `AuthSchemeId("sigv4")`, it just returns `sigv4` as-is.
fn auth_scheme_name(auth_scheme_id: &AuthSchemeId) -> &str {
    auth_scheme_id
        .inner()
        .split('#')
        .next_back()
        .unwrap_or(auth_scheme_id.inner())
}
```
This is backward compatible because the helper `auth_scheme_name` returns `sigv4` regardless of whether `AuthSchemeId::new("sigv4")` or `AuthSchemeId("aws.auth#sigv4")`.

ysaito1001: Hmm, this could be a potential breaking use case for `sigv4` and `sigv4a` (we might be able to turn the other auth schemes into fully qualified names).

Endpoint auth scheme resolution needs to be updated as part of the fully qualified name fix. There are two places: one is [extract_endpoint_auth_scheme_config](https://github.com/smithy-lang/smithy-rs/blob/1a94d1d42ac7bb5e49f04e6b7c24ef4dff9ff3ac/rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs#L311) and the other is [endpoint_auth](https://github.com/smithy-lang/smithy-rs/blob/1a94d1d42ac7bb5e49f04e6b7c24ef4dff9ff3ac/aws/rust-runtime/aws-inlineable/src/endpoint_auth.rs#L74) in `aws-inlineable`.

An auth scheme name referenced in the endpoint `authSchemes` list property is just [sigv4](https://github.com/awslabs/aws-sdk-rust/blob/931301e8858bae5fdc8811d1728b546c57eb0923/aws-models/billing.json#L166), not `aws.auth#sigv4`. In addition to HTTP-related auths, we also failed to specify the fully qualified auth scheme ID for [sigv4](https://github.com/smithy-lang/smithy-rs/blob/1a94d1d42ac7bb5e49f04e6b7c24ef4dff9ff3ac/aws/rust-runtime/aws-runtime/src/auth/sigv4.rs#L33) (and sigv4a). After we've updated `AuthSchemeId::new("sigv4")` to `AuthSchemeId::new("aws.auth#sigv4")`, the line of code above in `extract_endpoint_auth_scheme_config`
```
config_scheme_id == Some(scheme_id.inner())
```
should be updated to something like
```
config_scheme_id == Some(auth_scheme_name(scheme_id))
```
where `auth_scheme_name` looks something like
```
// For instance, extracts just `sigv4` from `AuthSchemeId("aws.auth#sigv4")`.
// For `AuthSchemeId("sigv4")`, it just returns `sigv4` as-is.
fn auth_scheme_name(auth_scheme_id: &AuthSchemeId) -> &str {
    auth_scheme_id
        .inner()
        .split('#')
        .next_back()
        .unwrap_or(auth_scheme_id.inner())
}
```
The same treatment needs to be made to `endpoint_auth` in `aws-inlineable`.

However, if an older SDK whose `aws-inlineable` does not have this update depended on a newer `aws-runtime` that includes the fully qualified name fix for `sigv4`, endpoint auth resolution would no longer work due to a name mismatch between "sigv4" and "aws.auth#sigv4". It'd be a runtime error as opposed to compile-time error, so the consequences would be pretty severe.

ysaito1001: Looks like we have three options:
#### 1. Keep the code as-is

pros:
- No breaking changes

cons:
- `AuthSchemeId` strings are not fully qualified and are incorrect
- We need to have this awkward function `runtime_api_auth_scheme_id`

#### 2. Update HTTP-related `AuthSchemeId`s to be `AuthSchemeId("httpBasicAuth")`, `AuthSchemeId("httpBearerAuth")`, and so on, but keep Sigv4 and Sigv4a as they are

pros:
- Probably no breaking changes
- We can remove `runtime_api_auth_scheme_id`

cons:
- `AuthSchemeId` strings are at least correct after the namespace part but still not fully qualified

#### 3. Update HTTP-related `AuthSchemeId`s to contain fully qualified names `AuthSchemeId("smithy.api#httpBasicAuth")`, `AuthSchemeId("smithy.api#httpBearerAuth")`, and so on, but keep Sigv4 and Sigv4a as they are

pros:
- Probably no breaking changes
- We can remove `runtime_api_auth_scheme_id`

cons:
- HTTP-related `AuthSchemeId` are now correct but Sigv4 and Sigv4a are not fully qualified, so it's inconsistent
- Auth preference re-prioritization logic needs to be updated so it only checks `AuthSchemeId`'s substring after the namespace part `#` (which is actually how it's handled in [the Kotlin SDK](https://github.com/smithy-lang/smithy-kotlin/blob/444ff8a71ad8a6658264c44796eda736aa05405c/runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/ReprioritizeAuthOptions.kt#L18-L23)).

ysaito1001: Looks like we have three options:
#### 1. Keep the code as-is

pros:
- No breaking changes

cons:
- `AuthSchemeId` strings are not fully qualified and are incorrect
- We need to have this awkward function `runtime_api_auth_scheme_id`

#### 2. Update HTTP-related and `NoAuth` `AuthSchemeId`s to be `AuthSchemeId("httpBasicAuth")`, `AuthSchemeId("httpBearerAuth")`, and so on, but keep Sigv4 and Sigv4a as they are

pros:
- Probably no breaking changes
- We can remove `runtime_api_auth_scheme_id`

cons:
- `AuthSchemeId` strings are at least correct after the namespace part but still not fully qualified

#### 3. Update HTTP-related and `NoAuth` `AuthSchemeId`s to contain fully qualified names `AuthSchemeId("smithy.api#httpBasicAuth")`, `AuthSchemeId("smithy.api#httpBearerAuth")`, and so on, but keep Sigv4 and Sigv4a as they are

pros:
- Probably no breaking changes
- We can remove `runtime_api_auth_scheme_id`

cons:
- HTTP-related `AuthSchemeId` are now correct but Sigv4 and Sigv4a are not fully qualified, so it's inconsistent
- Auth preference re-prioritization logic needs to be updated so it only checks `AuthSchemeId`'s substring after the namespace part `#` (which is actually how it's handled in [the Kotlin SDK](https://github.com/smithy-lang/smithy-kotlin/blob/444ff8a71ad8a6658264c44796eda736aa05405c/runtime/auth/http-auth/common/src/aws/smithy/kotlin/runtime/http/auth/ReprioritizeAuthOptions.kt#L18-L23)). 

---

Updating Sigv4 and Sigv4a `AuthSchemeId`s to contain fully qualified names won't be an option due to the previous comment (unless we introduce yet another "v2" marker, which I don't think it's worth for this).

aajtodd: I guess I'd lean towards 2 but open to debate.

landonxjames: I think 2 would also get my vote. Especially if we can detect users who would be broken and output a `tracing::warn` to indicate that we might break this in the future.

ysaito1001: Yes, discussed offline, and we're going for 2.

ysaito1001: Updated in fca5226. However, the `semver-hazards` CI job will fail due to this [inappropriately written test](https://github.com/smithy-lang/smithy-rs/blob/179995538bffd1f7f299e9a926ba907dae9c32a4/aws/sdk/integration-tests/s3/tests/express.rs#L478) (by me 😓).

Will create a separate PR to fix the test so the expectation does not rely on a raw string, instead should rely on `aws_smithy_runtime::client::auth::no_auth::NO_AUTH_SCHEME_ID.inner()`. Only after that PR is merged and released from the `aws-sdk-rust` repository, can this PR will pass all checks in CI.

EDIT:
[PR created](https://github.com/smithy-lang/smithy-rs/pull/4211)

aajtodd (approved):

PR #4036, opened 2025-02-24
Enable event stream operations with RPC bound protocols in client SDKs

ysaito1001:
## Motivation and Context
Addresses aws-sdk-rust#213 and aws-sdk-rust#1188.

## Description
This PR consolidates the changes from the previous sub-PRs:
- https://github.com/smithy-lang/smithy-rs/pull/4004
- https://github.com/smithy-lang/smithy-rs/pull/4015
- https://github.com/smithy-lang/smithy-rs/pull/4023

Together, this PR enables event stream operations with RPC bound protocols, including `SubscribeToShard` in Kinesis and `StartLiveTail` in CloudWatchLogs in the Rust SDK. 

## Testing
(consolidated bullets from the previous sub-PRs)
- Added a service integration test for cloudwatchlogs
- Added client SDK codegen tests in `ClientEventStreamUnmarshallerGeneratorTest` to reverify initial-response message handling
- Added client SDK codegen tests in `ClientEventStreamMarshallerGeneratorTest` to verify initial-request message handling
- Added `DisableStalledStreamProtectionTest`
- Confirmed a successful run in the release pipeline 

## Checklist

- [x] For changes to the smithy-rs codegen or runtime crates, I have created a changelog entry Markdown file in the `.changelog` directory, specifying "client," "server," or both in the `applies_to` key.
- [x] For changes to the AWS SDK, generated SDK code, or SDK runtime crates, I have created a changelog entry Markdown file in the `.changelog` directory, specifying "aws-sdk-rust" in the `applies_to` key.

----

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #4208, opened 2025-07-13
Use new `hints.mostly-unused` to speed up compilation

joshtriplett:
Most users of the AWS SDK crates will use a fraction of their API
surface area.

Nightly rustc now provides an option `-Zhint-mostly-unused` to tell it
to defer as much compilation as possible, which provides a substantial
performance improvement if most of that compilation doesn't end up
happening. Cargo plumbs this option through using the new `[hints]`
table. This will cause users of the AWS SDK crates to default to setting
`hint-mostly-unused`. (Top-level crates can override this if they wish,
using a new profile option.)

Note that setting this hint does not increase the MSRV of the AWS SDK
crates, as old versions of Cargo will ignore it. New versions of Cargo
will respect it automatically (and, until we stabilize it, Cargo will do
nothing unless you pass `-Zprofile-hint-mostly-unused` to cargo).

Some sample performance numbers: this takes `aws-sdk-ec2` compilation
time ~4m07s to ~2m04s, a ~50% speedup.

Please note that this change to the code generator is the first Kotlin code
I've ever written, and I can't seem to get gradle to run locally in order to
*test* it, so this is a draft. I've extensively tested the addition of the
`hints` table manually, but I'm currently relying on CI to verify that this
produces the expected `Cargo.toml` changes for the `aws-sdk-*` crates.

## Checklist

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
```diff
@@ -138,6 +139,7 @@ fun testRustSettings(
     moduleAuthors,
     moduleDescription,
     moduleRepository,
+    hintMostlyUnused,
```
aajtodd, line 142: order here appears off

aajtodd (commented):
This is a most welcome PR so first thank you. We've been discussing for a while now how we're going to improve compile times for the SDK crates but working on the compiler is not our area of expertise and some service models like EC2 are just large and there will only be so much we can do on our side. 

Happy to help try and work through the gradle issues, relying on CI is fine as well though.

landonxjames:
Reading through the [proposed blog post](https://github.com/rust-lang/blog.rust-lang.org/pull/1662/files) on this feature and the corresponding [Zulip thread](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Call.20for.20testing.20on.20hint-mostly-unused/with/528510286) I wonder if it would be best to scope this change to just a few of our larger more compilation heavy crates to begin? The blog says:
>Always do performance analysis when considering this hint, and only apply it if
it applies obvious and substantial wins for your users. Never apply it across
the board to all your dependencies.

Compilation improvements are a frequent ask from our users (see the most upvoted issue on the [aws-sdk-rust repo](https://github.com/awslabs/aws-sdk-rust/issues/113)), so this is definitely something we want, and I think the intent here applies to some of the more popular AWS crates, especially ones that combine control/data plane operations in a single client (EC2, S3, DDB). But I don't know that we would want this on all crates from the outset without a ton of benchmarking of customer applications.

Since this is such a high user priority if you don't feel like you have the comfort with Kotlin (or desire/free time to develop it) to scope this we can discuss taking over the work on our team. If you were interested in doing the work it would probably look something like a simple [decorator customizing the manifest](https://github.com/smithy-lang/smithy-rs/blob/1a94d1d42ac7bb5e49f04e6b7c24ef4dff9ff3ac/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/DocsRsMetadataDecorator.kt#L52) that is registered in [AwsCodegenDecorator](https://github.com/smithy-lang/smithy-rs/blob/1a94d1d42ac7bb5e49f04e6b7c24ef4dff9ff3ac/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt#L38) and only [targets a few specific services](https://github.com/smithy-lang/smithy-rs/blob/1a94d1d42ac7bb5e49f04e6b7c24ef4dff9ff3ac/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/AwsCodegenDecorator.kt#L83).

ysaito1001:
Love the overall direction! We can definitely massage this on our end, if needed.

As Landon says, having an allow list for target services and making the list configurable through codegen settings might be desirable (like the one we used to have for [event streams](https://github.com/smithy-lang/smithy-rs/pull/4036/files#diff-1add96e2a6a33ca2d558c942a3968465b35642c5fb2feabef3df30892ec67164L12)).

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/TestHelpers.kt
ysaito1001: ought be something like
```
    ...
    license,
    examplesUri,
    minimumSupportedRustVersion = null,
    hintMostlyUnused,
)
```

ysaito1001: ought to be something like
```
    ...
    license,
    examplesUri,
    minimumSupportedRustVersion = null,
    hintMostlyUnused,
)
```

joshtriplett:
I'd be happy to see this taken over by someone familiar with both Kotlin and the code generator.

And it seems fine to scope this to just the biggest crates; it wouldn't make sense, for instance, for a crate with only a couple of functions.

landonxjames:
>I'd be happy to see this taken over by someone familiar with both Kotlin and the code generator.
>
>And it seems fine to scope this to just the biggest crates; it wouldn't make sense, for instance, for a crate with only a couple of functions.

Awesome, thanks for bringing this new feature to our attention. I will take over this work in a week or two when I clear out my current tasks.

landonxjames:
Superseded by https://github.com/smithy-lang/smithy-rs/pull/4243 which should deploy later this week, thanks for bringing this to our attention (and contributing the feature to the compiler in the first place)!

PR #4210, opened 2025-07-15
Wire caching rfc

rcoh:
[Rendered](https://github.com/smithy-lang/smithy-rs/blob/wire-caching-rfc/design/src/rfcs/rfc0046_wire_cached_responses.md)

## Motivation and Context
Add an RFC to support wire-caching of response types. This will enable high performance services written in smithy-rs to more efficiently cache data.
----

### DIFF
--- a/design/src/SUMMARY.md
+++ b/design/src/SUMMARY.md
@@ -68,5 +68,6 @@
   - [RFC-0043: Identity Cache Partitions](./rfcs/rfc0043_identity_cache_partitions.md)
   - [RFC-0044: Environment-defined service configuration](./rfcs/rfc0044_env_defined_service_config.md)
   - [RFC-0045: Configurable Serde](./rfcs/rfc0045_configurable_serde.md)
+  - [RFC-0046: Wire cached responses](./rfcs/rfc0046_wire_cached_responses.md)
 - [Contributing](./contributing/overview.md)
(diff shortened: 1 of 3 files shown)


drganjoo:
Currently, we don't expose serialization methods to service teams, so implementing wire-format caching would require adding some API for services to obtain the serialized bytes.

**Potential Enhancement:**

When we do expose this capability, we could consider including validation metadata in the cache storage format:

```
{
  metadata: {
    model_version: "...",
    protocol_version: "...", 
    field_index: Map<SmithyShapeId, FieldMetadata>
  },
  wire_data: <raw_bytes>
}
```

**How this could work:**

- Code generator emits field position tracking alongside serialization
- Services store the complete format (metadata + wire bytes) in their cache
- When returning cached data, we strip the metadata and send only the raw bytes to clients
- Field index enables selective validation of cached responses without full deserialization

**Areas that need more exploration:**

- Collection handling (arrays/maps with validation requirements)
- Performance impact during serialization 
- Storage overhead scaling

This might be worth digging into as part of the serializer exposure work, but we'd want to benchmark the performance implications first.

rcoh:
including metadata is a potential avenue to explore. I know that some of the downstream customers are _extremely_ space sensitive so it would need to be opt in.

PR #4212, opened 2025-07-15
Preallocate correct-sized buffer for event stream

rcoh:
## Motivation and Context

Currently when serializing event streams, an empty buffer is used. This results in repeated reallocations as the buffer grows. However, we can actually precalculate the exact size that is required.

## Description

This PR does two things:
1. Add a MessageSizeHint that aws-smithy-http can use to determine the in-memory serialized size of the message
2. Add a benchmark of write_message. The benchmark includes an prototype optimized implementation that we can switch to at a later date.

## Testing
- Benchmark
- Test of correctness for size calculation

## Checklist

----

### DIFF
--- a/rust-runtime/Cargo.lock
+++ b/rust-runtime/Cargo.lock
@@ -228,7 +228,7 @@ checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.102",
 ]
 
 [[package]]
@@ -245,7 +245,7 @@ checksum = "e539d3fca749fcee5236ab05e93a52867dd549cc157c8cb7f99595f3cedffdb5"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.102",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


rust-runtime/Cargo.lock
```diff
@@ -2447,6 +2658,97 @@ dependencies = [
...
+
+[[package]]
+name = "pyo3-asyncio-macros"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56c467178e1da6252c95c29ecf898b133f742e9181dca5def15dc24e19d45a39"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
```
arielb1, line 2704: How did this pop up?

landonxjames: This is probably an artifact of https://github.com/smithy-lang/smithy-rs/pull/4200#discussion_r2195456323. In the PR the Python crates were accidentally excluded from the workspace. That was fixed, but I doubt that the Cargo.lock was regenerated along with that fix, so these dependencies didn't make it in.

rust-runtime/aws-smithy-eventstream/benches/write_message_performance.rs
```diff
@@ -0,0 +1,484 @@
...
+const TYPE_INT32: u8 = 4;
+const TYPE_INT64: u8 = 5;
+const TYPE_BYTE_ARRAY: u8 = 6;
+const TYPE_STRING: u8 = 7;
+const TYPE_TIMESTAMP: u8 = 8;
+const TYPE_UUID: u8 = 9;
+
+/// Optimized version of write_message to remove inline buffer. Stashing this here—this implemenetation
+/// is about 10% faster but needs more testing.
+pub fn write_message_to_optimized_v1(
```
ysaito1001, line 89: I suppose the intent here is that once the `_v1` implementation reaches sufficient performance levels, we could consider replacing the original `write_message_to` with this version?

aajtodd (approved):
Do you have numbers on improvement seen from introducing the size hint?

rust-runtime/aws-smithy-eventstream/benches/write_message_performance.rs
rcoh: its a decent bit faster, it just needs to be better tested and cleaned up

rcoh:
@aajtodd no end-to-end numbers, but for a 1KB message on the default allocator it takes it from 1us to 500ns.

PR #4217, opened 2025-07-16
Move gradle cache step out of conditional

landonxjames:
## Motivation and Context

Improve the CI build Docker image by updating the gradle binary caching logic to properly support both x86_64 and ARM64 (aarch64) architectures.

## Description

* Dynamically set JAVA_HOME based on the architecture of the build machine
* Modified the smithy-rs repository cloning process to ensure gradle binary caching works consistently

Note that this fixes a bug with our existing Dockerfile. We were manually setting `JAVA_HOME` to the `x86_64` variant of the Java binary from `yum`. This worked even on M series Macs because we never invoked Java in the image build and the image currently only runs on `x86_64` machines. But if you shelled into the container on a Mac and tried to run Gradle it would fail because yum actually downloaded the `aarch64` binary. The update to dynamically set `JAVA_HOME` accounts for this so now the container should work equally well on `x86_64` or `aarch64`.

## Testing

Was able to build on my M1 mac and shell into the container and do some basic gradle builds. It successfully builds and runs all CI on GH, so should work in any environments we care about. 

----


tools/ci-build/Dockerfile
```diff
@@ -79,15 +95,22 @@ COPY . tools/ci-build
 ARG smithy_rs_commit_hash=main
 # If the `checkout_smithy_rs_tools` arg is set to true, then the Dockerfile will acquire the tools
 # source code by checking out smithy-lang/smithy-rs/main rather than copying them from the local directory.
+# If it is false we still clone the smithy-rs repo beecause we need the gradle wrapper to install the gradle
+# binary. But in this case we delete the repo before installing the tools and use the local versions from
+# the build context.
 ARG checkout_smithy_rs_tools=false
 RUN set -eux; \
+    git clone https://github.com/smithy-lang/smithy-rs.git; \
```
aajtodd: A full clone just to get the gradle version seems...heavy. While infrequent, it also will miss changes locally or in a branch PR if we do change the gradle version.

I wonder if just setting the gradle version in the dockerfile and installing it on our own wouldn't be better 🤔 . I suppose then we have to track/check if we ever change the gradle wrapper version and detect mismatches between wrapper and docker version.

landonxjames: >I suppose then we have to track/check if we ever change the gradle wrapper version and detect mismatches between wrapper and docker version.

That was my thought, we need to be using the same gradle version from smithy-rs, so we should install from that environment. We could track the version in the Dockerfile as well, but I think they will inevitably get out of sync and cause us an issue down the line

landonxjames: FWIW `time`ing the `clone` on my dev desktop has it taking about 12 seconds, which isn't terrible in the scope of all of the `cargo install`s we do in the image
```
git clone https://github.com/smithy-lang/smithy-rs.git  11.46s user 1.55s system 233% cpu 5.570 total
```

aajtodd: Yeah the other half of the issue is if we want to make branch or local changes though to the gradle version this doesn't solve that.

landonxjames: After some offline discussion we have decided to go with this approach (because we are seeing the occasional internal build failure due to failing to download the gradle binary) and revisit this issue if it becomes a problem in the future. Gradle version bumps are sufficiently rare that we are hoping it won't be a problem in practice. 

Updated the clone to be shallow in https://github.com/smithy-lang/smithy-rs/pull/4217/commits/032876621068ecaad781a91891cf5d2bd65cd4a4

ysaito1001 (approved):

PR #4224, opened 2025-07-18
Initial setup for credential resolution feature tracking

landonxjames:
## Motivation and Context

Initial setup for tracking credential/identity resolution based features. 

## Description

The biggest change here is the addition of a new `aws-features` crate containing the `sdk_features` module that used to live in `aws-runtime`. This was done so that the `AwsSdkFeature` type could be referenced in the `aws-credential-types` crate without causing a circular dependency. 

Initially I was just going to pass along the types without downcasting, but they would have had to be downcast eventually, since in the `Credentials` typemap they are `Vec<AwsSdkFeature>` but to add them to a `Layer` we need to extract the individual `AwsSdkFeature`s from the `Vec`. This could either happen in `aws-credential-types` where `From<Credentials> for Identity` is implemented or in `aws-smithy-runtime` in the orchestrator where the `Identity` is resolved. Since these credentials types are AWS specific it seemed to make more sense to keep it in `aws/rust-runtime`.  

Other changes:
* Add a typemap to `Credentials` (and manually implement the previously derived traits for it) to carry the feature information
* Update the `From<Credentials> for Identity` implementation to extract the `AwsSdkFeatures` and pass them in a `Layer` to the `Identity`'s typemap. 
* Update `resolve_identity` in the orchestrator to extract the `FrozenLayer` from `Identity` and insert it in the `ConfigBag` so that the `UserAgentInterceptor` can extract it later.

## Testing

Added new tests around `Credentials` equality and the updated `From<Credentials> for Identity` implementation. 

**Note on failing semver test:** Failing because the `UnwindSafe` traits are no longer auto implemented for `Credentials`

```
--- failure auto_trait_impl_removed: auto trait no longer implemented ---

Description:
A public type has stopped implementing one or more auto traits. This can break downstream code that depends on the traits being implemented.
        ref: https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.41.0/src/lints/auto_trait_impl_removed.ron

Failed in:
  type Credentials is no longer UnwindSafe, in /home/build/workspace/smithy-rs/aws-sdk/sdk/aws-credential-types/src/credentials_impl.rs:29
  type Credentials is no longer RefUnwindSafe, in /home/build/workspace/smithy-rs/aws-sdk/sdk/aws-credential-types/src/credentials_impl.rs:29Cloning base
```

This is due to us adding a `HashMap<TypeId, TypeErasedBox>` to `Credentials`. We could likely wrap this in an `Arc<Mutex<>>` to keep it mutable and get back the unwind safety, but that doesn't feel like it justifies the added complexity. Open to debate on this one though. 

----

### DIFF
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -319,7 +319,7 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-runtime"
-version = "1.8.5"
+version = "1.8.6"
 dependencies = [
  "aws-smithy-async",
  "aws-smithy-http",
@@ -341,7 +341,7 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-runtime-api"
-version = "1.8.4"
+version = "1.8.5"
 dependencies = [
  "aws-smithy-async",
  "aws-smithy-types",
--- a/rust-runtime/Cargo.lock
+++ b/rust-runtime/Cargo.lock
@@ -616,7 +616,7 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-runtime"
-version = "1.8.5"
+version = "1.8.6"
 dependencies = [
  "approx",
  "aws-smithy-async",
@@ -644,7 +644,7 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-runtime-api"
-version = "1.8.4"
+version = "1.8.5"
 dependencies = [
  "aws-smithy-async",
  "aws-smithy-types",
--- a/rust-runtime/aws-smithy-runtime-api/Cargo.toml
+++ b/rust-runtime/aws-smithy-runtime-api/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-runtime-api"
-version = "1.8.4"
+version = "1.8.5"
 authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "Zelda Hessler <zhessler@amazon.com>"]
 description = "Smithy runtime types."
 edition = "2021"
--- a/rust-runtime/aws-smithy-runtime/Cargo.toml
+++ b/rust-runtime/aws-smithy-runtime/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-runtime"
-version = "1.8.5"
+version = "1.8.6"
 authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>", "Zelda Hessler <zhessler@amazon.com>"]
 description = "The new smithy runtime crate"
 edition = "2021"


aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
```diff
@@ -21,8 +26,30 @@ use crate::attributes::AccountId;
 ///
 /// When `Credentials` is dropped, its contents are zeroed in memory. Credentials uses an interior Arc to ensure
 /// that even when cloned, credentials don't exist in multiple memory locations.
-#[derive(Clone, Eq, PartialEq)]
-pub struct Credentials(Arc<Inner>);
+pub struct Credentials(Arc<Inner>, HashMap<TypeId, TypeErasedBox>);
+
+impl Clone for Credentials {
+    fn clone(&self) -> Self {
+        let mut new_map = HashMap::new();
```
ysaito1001: We should be able to use `HashMap::with_capacity` based on the size of `self.1`.

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
```diff
@@ -359,4 +441,58 @@ mod test {
...
+    fn equality_ignores_properties() {
+        #[derive(Clone, Debug)]
+        struct Foo;
+        let mut creds1 = Credentials::new(
+            "akid",
+            "secret",
+            Some("token".into()),
+            Some(UNIX_EPOCH + Duration::from_secs(1234567890)),
+            "debug tester",
+        );
```
ysaito1001: Can these be replaced with `Credentials::for_tests()` or `Credentials::for_tests_with_session_token()`?

ysaito1001 (approved):
Yeah, looks clearer to hoist AWS SDK features to a upstream crate altogether, as opposed to special casing credentials-related features with `AwsSdkCredentialsFeatures` defined in `aws-credential-types`. 

Great start!

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
landonxjames: Good catch, updated in https://github.com/smithy-lang/smithy-rs/pull/4224/commits/ddcbecafb146d017e9c32ebbe87a4c7b0145841b

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
landonxjames: Much cleaner, updated in https://github.com/smithy-lang/smithy-rs/pull/4224/commits/ddcbecafb146d017e9c32ebbe87a4c7b0145841b

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
```diff
@@ -317,12 +386,22 @@ impl From<Credentials> for Identity {
...
+            Identity::builder().property(account_id.clone())
         } else {
-            Identity::builder().data(val)
+            Identity::builder()
         };
+
         builder.set_expiration(expiry);
-        builder.build().expect("set required fields")
+
+        if let Some(features) = val.get_property::<Vec<AwsSdkFeature>>().cloned() {
```
aajtodd: Seems like it doesn't matter where this lives if `aws-credential-types` has to know about it. Alternatively could just make a new `AwsCredentialFeature` enum inside `aws-credential-types` and implement `BusinessMetric` conversions for that in `aws-runtime`. Would mean we don't need a new crate for this.

aajtodd (approved):
Looks good, brought up one possible suggestion to avoid a new crate but otherwise looks fine.

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
ysaito1001: Hmm, after thinking about the point of having a repository-like crate, I've started learning towards this idea of defining `AwsCredentialFeature` in `aws-credential-types`...

rust-runtime/aws-smithy-runtime/src/client/orchestrator/auth.rs
```diff
@@ -162,6 +162,11 @@ pub(super) async fn resolve_identity(
                             .resolve_cached_identity(identity_resolver, runtime_components, cfg)
                             .await?;
                         trace!(identity = ?identity, "resolved identity");
+                        // Extract the FrozenLayer placed in the Identity property bag by the From<Credentials> impl.
+                        // This layer contains AwsSdkFeatures for the user agent
```
ysaito1001: +1 for adding comments, but since this is in generic smithy runtime and Identity's frozen layer could be used for something else, so either I'd remove the comments or make them more abstract, so they're not tied specifically to `AwsSdkFeatures`.

landonxjames: Good call, made comment more generic in https://github.com/smithy-lang/smithy-rs/pull/4224/commits/b785a6ef3a6b9e89b8a800e379ffd9f7b730bbef

aws/rust-runtime/aws-credential-types/src/credentials_impl.rs
landonxjames: Sounds like we are leaning towards a separate enum. I still prefer only having two feature enums (AWS and Smithy) since it seems cleaner, but don't know that that preference is strong enough to justify a new crate. Will make the change today.

landonxjames: Switched to an `AwsCredentialFeature` enum and removed the `aws-features` crate in https://github.com/smithy-lang/smithy-rs/pull/4224/commits/6bcea292b6197a62fef0adf12ff1604d4bc257ce

PR #4232, opened 2025-07-24
Add fallback equality on `noAuth` scheme ID

ysaito1001:
## Motivation and Context
Implements fallback equality for no auth `AuthSchemeId` so that `AuthSchemeId::from("no_auth")` (legacy) and `AuthSchemeId::from("noAuth")` (updated) should be treated as equivalent.

## Description
In #4203, the internal raw strings of pre-defined `AuthSchemeId` were updated to better align with the Smithy spec ([discussion](https://github.com/smithy-lang/smithy-rs/pull/4203#discussion_r2198206822)). Acknowledging that this was a breaking change and that customers should not rely on these internal representations, we did receive reports of issues related to this update. After discussion, we proceeded with implementing fallback equality for no auth scheme ID to allow for a safer rollout.

## Testing
- ~Added unit tests for manually implemented traits for `AuthSchemeId`, `PartialEq`, `Hash`, and `Ord`~ (became unnecessary after 7922ef9)
- Added an auth scheme preference test to verify the legacy `no_auth` is supported

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-runtime-api/src/client/auth.rs
+++ b/rust-runtime/aws-smithy-runtime-api/src/client/auth.rs
@@ -15,7 +15,6 @@ use aws_smithy_types::config_bag::{ConfigBag, FrozenLayer, Storable, StoreReplac
 use aws_smithy_types::type_erasure::TypeErasedBox;
 use aws_smithy_types::Document;
 use std::borrow::Cow;
-use std::cmp::Ordering;
 use std::fmt;
 use std::sync::Arc;
 
@@ -140,7 +139,7 @@ impl std::error::Error for AuthSchemeOptionBuilderError {}
 /// Each auth scheme must have a unique string identifier associated with it,
 /// which is used to refer to auth schemes by the auth scheme option resolver, and
 /// also used to select an identity resolver to use.
-#[derive(Clone, Debug, Eq)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub struct AuthSchemeId {
     scheme_id: Cow<'static, str>,
 }
@@ -152,11 +151,22 @@ impl AsRef<AuthSchemeId> for AuthSchemeId {
     }
 }
 
+// Normalizes auth scheme IDs for comparison and hashing by treating "no_auth" and "noAuth" as equivalent
+// by converting "no_auth" to "noAuth".
+// This is for backward compatibility; "no_auth" was incorrectly used in earlier GA versions of the SDK and
+// could be used still in some places.
+const fn normalize_auth_scheme_id(s: &'static str) -> &'static str {
+    match s.as_bytes() {
+        b"no_auth" => "noAuth",
+        _ => s,
+    }
+}
+
 impl AuthSchemeId {
     /// Creates a new auth scheme ID.
     pub const fn new(scheme_id: &'static str) -> Self {
         Self {
-            scheme_id: Cow::Borrowed(scheme_id),
+            scheme_id: Cow::Borrowed(normalize_auth_scheme_id(scheme_id)),
         }
     }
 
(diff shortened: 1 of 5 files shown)


rust-runtime/aws-smithy-runtime-api/src/client/auth.rs
```diff
@@ -192,6 +193,47 @@ impl From<Cow<'static, str>> for AuthSchemeId {
     }
 }
 
+impl PartialEq for AuthSchemeId {
```
rcoh: can you handle this in `::new` and normalize there? Seems simpler.

ysaito1001: Interesting, we need to cover the following also
```
impl From<&'static str> for AuthSchemeId {
    fn from(scheme_id: &'static str) -> Self {
        Self::new(scheme_id)
    }
}

impl From<Cow<'static, str>> for AuthSchemeId {
    fn from(scheme_id: Cow<'static, str>) -> Self {
        Self { scheme_id }
    }
}
```
but it will help eliminate handwritten trait impls and their unit tests.

ysaito1001: Hmm, looks like we cannot unfortunately, since `new` is a const function that cannot call a non-const operator `==`.

rust-runtime/aws-smithy-runtime-api/src/client/auth.rs
```diff
@@ -192,6 +193,47 @@ impl From<Cow<'static, str>> for AuthSchemeId {
...
+        Some(self.cmp(other))
+    }
+}
+
+// Normalizes auth scheme IDs for comparison and hashing by treating "no_auth" and "noAuth" as equivalent
+// by converting "no_auth" to "noAuth".
+// This is for backward compatibility; "no_auth" was incorrectly used in pre-GA versions of the SDK and
+// could be used still in some places.
+fn normalize_auth_scheme_id(scheme_id: &str) -> &str {
+    if scheme_id == "no_auth" {
```
landonxjames: Maybe worth a `warn!` in this branch to let customers know they should use the newer scheme name?

jlizen: Just for my own clarity, is the intent here to continue supporting this old string as a fallback until the next breaking version (where presumably it would go in a migration guide)?

jlizen: (And specifically for the `warn!`, I assume we want to recommend using the `NO_AUTH_SCHEME_ID` constant rather than hardcoding a different string?)

ysaito1001: > is the intent here to continue supporting this old string as a fallback until the next breaking version

Yes, we currently don't have a plan for the 2.x version, but if we do, we should mention it in the migration guide.

> I assume we want to recommend using the NO_AUTH_SCHEME_ID constant rather than hardcoding a different string

Exactly.

I like the idea of warning somewhere, but am a bit heaitent to do it in the branch above. Since `normalize_auth_scheme_id` is called in every invocation to `PartialOrd`, `Hash`, and `PartialEq`. We don't want to overwhelm customers with the repeated warning messages. This is why I wished the Russell's suggestion worke out, because we could issue the warning only once per `AuthSchemeId` construction.

ysaito1001: I wanted to issue a warning in `new` and `impl From<Cow<'static, str>> for AuthSchemeId` but cannot do that in `new`, again because of `const`-ness.

rust-runtime/aws-smithy-runtime-api/src/client/auth.rs
ysaito1001: Let me try something like
```
const fn normalize_auth_scheme_id(s: &'static str) -> &'static str {
    match s.as_bytes() {
        b"no_auth" => "noAuth",
        _ => s,
    }
}
```

ysaito1001: Thanks for the suggestion, updated in 7922ef9. Credit goes to @landonxjames for the const function above.

rust-runtime/aws-smithy-runtime-api/src/client/auth.rs
ysaito1001: As discussed offline, issuing a warning probably matters less, since whichever no auth scheme ID string users are using, it will be normalized upon `AuthSchemeId`'s construction anyway.

rcoh (approved):

Issue #4227, opened 2025-07-22
Plus signs are not encoded correctly during aws-sigv4 signing


dnsmth:
When canonicalizing the query parameters, aws-sigv4 canonicalizes a '+' as `%20`, i.e. the escape sequence for a space, but in reality it should be canonicalized as  `%2B`. Minimal repro, adapted from the signing example:

```
        use aws_sigv4::http_request::{sign, SignableRequest, SigningParams, SigningSettings};
        use aws_sigv4::sign::v4;
        use std::time::SystemTime;

        // Set up information and settings for the signing
        // You can obtain credentials from `SdkConfig`.
        let identity = aws_credential_types::Credentials::new(
            "AKIDEXAMPLE",
            "wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY",
            None,
            None,
            "hardcoded-credentials",
        )
        .into();
        let signing_settings = SigningSettings::default();
        let signing_params = v4::SigningParams::builder()
            .identity(&identity)
            .region("us-east-1")
            .name("exampleservice")
            .time(SystemTime::now())
            .settings(signing_settings)
            .build()
            .unwrap()
            .into();
        // Convert the HTTP request into a signable request
        let signable_request = SignableRequest::new(
            "GET",
            "https://some-endpoint.some-region.amazonaws.com?c+d",
            std::iter::empty(),
            aws_sigv4::http_request::SignableBody::Bytes(&[]),
        )
        .expect("signable request");

        let mut my_req = http::Request::new("...");
        // Sign and then apply the signature to the request
        let (signing_instructions, _signature) =
            sign(signable_request, &signing_params)?.into_parts();
        signing_instructions.apply_to_request_http0x(&mut my_req);
        Ok(())
```

For this example, the request is canonicalized to 

```
/
c%20d=
host:some-endpoint.some-region.amazonaws.com
x-amz-date:20250722T155736Z

host;x-amz-date
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

But in reality, it should be 
```
/
c%2Bd=
host:some-endpoint.some-region.amazonaws.com
x-amz-date:20250722T155736Z

host;x-amz-date
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

From what I can tell, this comes from `form_urlencoded::parse` normalizing plus to space during `CanonicalRequest::Params`.

[...]

PR #4237, opened 2025-07-25
fix sigv4 canonical request order and revamp test suite

aajtodd:
## Description
Started as an investigation of https://github.com/smithy-lang/smithy-rs/issues/4227.

I found that we 1. weren't utilizing most of our test suite (we only used 4 of the test directories from the v4 suite) and 2. that we had a bug in our canonical request implementation where we sort the query keys before encoding instead of after (see https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#create-canonical-request and the `get-vanilla-query-order-encoded` test).

This PR does a few things:
1. Imported the CRT test suite and removed the old ones.
    * The previous test suite we were using is the older format that was in use and floating around before CRT updated it. The newer version of this has standardized the file format and includes both header and query versions of each test.
    * We were already using this format for the sigv4a tests so now all of our tests for v4 and v4a use the same format. It also has a few additional tests we didn't have. 
    * Using the same format as CRT (and Kotlin) will make it easier to update the test suite or add new tests
3. Updates the test utils to be common across v4 and v4a. 
4. Added new test cases we weren't running for both v4 and v4a. Also every test now generally tests both headers and query signature locations (before we were only testing for headers most of the time).
5. Fixed a bug in canonical request implementation where we sort the query keys before encoding instead of after (found by new test and pointed out by engineer on ARC team as well).
6. Migrated a few tests `double-encode-path` and `double-url-encode` that we used in several places to the newer format (only for headers not query which was missing).
7. Inlined the `iam` related test request where it is used
8. Added a new `insert_encoded` for `QueryWriter` to append an already encoded key/value pair and skip encoding.

## Testing
* During the migration I kept both the old and new test suites around. Then removed the old suite after verifying the tests that utilized the same files still passed on the new test suite.

## Checklist

----


landonxjames (approved):
Should we add more docs around [SIgnableRequest](https://docs.rs/aws-sigv4/1.3.3/aws_sigv4/http_request/struct.SignableRequest.html) to make it obvious that the uri should be encoded to address the confusion from #4227, or were there going to be more changes for that issue?

aajtodd:
> Should we add more docs around [SIgnableRequest](https://docs.rs/aws-sigv4/1.3.3/aws_sigv4/http_request/struct.SignableRequest.html) to make it obvious that the uri should be encoded to address the confusion from #4227, or were there going to be more changes for that issue?

I wasn't planning on making additional changes but yeah I can add some additional docs to signable request to indicate it should be a encoded URI.

PR #4238, opened 2025-07-28
Add `AwsCredentialFeatures` to Credential Providers 

landonxjames:
## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/user_agent/test_util.rs
+++ b/aws/rust-runtime/aws-runtime/src/user_agent/test_util.rs
@@ -5,6 +5,7 @@
 
 //! Utilities for testing the User-Agent header
 
+use aws_smithy_runtime_api::client::orchestrator::HttpRequest;
 use regex_lite::Regex;
 use std::sync::LazyLock;
 
@@ -48,6 +49,11 @@ pub fn extract_ua_values(user_agent: &str) -> Option<Vec<&str>> {
             .split(',')
             .collect()
     })
+}
+
+/// Extract the `x-amz-user-agent` header from an [HttpRequest]
+pub fn get_sdk_metric_str_from_request(req: &HttpRequest) -> &str {
+    req.headers().get("x-amz-user-agent").unwrap()
 }
 
 #[cfg(test)]
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -92,6 +92,7 @@ dependencies = [
  "aws-smithy-runtime-api",
  "aws-smithy-types",
  "tokio",
+ "tracing",
  "zeroize",
 ]
 
@@ -341,7 +342,7 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-runtime-api"
-version = "1.8.5"
+version = "1.8.6"
 dependencies = [
  "aws-smithy-async",
  "aws-smithy-types",
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -92,7 +92,6 @@ dependencies = [
  "aws-smithy-runtime-api",
  "aws-smithy-types",
  "tokio",
- "tracing",
  "zeroize",
 ]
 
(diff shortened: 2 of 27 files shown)


aws/rust-runtime/aws-runtime/src/user_agent/test_util.rs
```diff
@@ -50,6 +51,11 @@ pub fn extract_ua_values(user_agent: &str) -> Option<Vec<&str>> {
     })
 }
 
+/// Extract the `x-amz-user-agent` header from an [HttpRequest]
+pub fn get_sdk_metric_str_from_request(req: &HttpRequest) -> &str {
```
aajtodd: this seems like an unnecessary public utility function

ysaito1001:
> It seems like we made a lot of stuff public or behind a test-util feature to allow testing via sdk/integration-tests. Part of me wonders if we wouldn't be better off making these unit tests in aws-config

Good point. We should generally place tests in `sdk/integration-tests` only when tests are something specific about those target services. For instance, the test logic in `imds_ua_feature` can equally apply to any services, so probably better off having those tests in `aws-config` and making newly exposed test-utils back to `pub(crate)` 🤔

ysaito1001 (commented):
Will circle back for test parts, but looks great. Thank you for making this change 👍

landonxjames:
We discussed offline and decided that the best course of action, to avoid exposing new `pub` `test-util` functions, is to test the credentials features are added in `aws-config` and only retain some of the tests that don't need new `test-util` functionality in integration tests. This work was done in:
* https://github.com/smithy-lang/smithy-rs/pull/4238/commits/63f2134f87ebf4a649b1f6d299b80f43fcbb17c2 adding tests to `aws-config`
* https://github.com/smithy-lang/smithy-rs/pull/4238/commits/eda29124f7dcf3e0f6a6d7c00d110b13dc60128c removing the newly expose test-utils
* https://github.com/smithy-lang/smithy-rs/pull/4238/commits/37276c5e0ace204ec6c0f1a727e085b17e15d36a removing the now broken integration tests and moving the remaining ones to the S3 test crate

aws/rust-runtime/aws-runtime/src/user_agent/test_util.rs
landonxjames: Removed in https://github.com/smithy-lang/smithy-rs/pull/4238/commits/eda29124f7dcf3e0f6a6d7c00d110b13dc60128c

aws/sdk/integration-tests/s3/tests/credential_features.rs
```diff
@@ -0,0 +1,39 @@
...
+        .build();
+
+    let client = Client::from_conf(config);
+
+    let _ = client
+        .head_bucket()
+        .bucket("fake")
+        .send()
+        .await
+        .expect("XXXXXXXXXXX");
```
ysaito1001: Meant to fill in something?

aws/rust-runtime/aws-config/external-types.toml
```diff
@@ -16,9 +16,11 @@ allowed_external_types = [
    "aws_runtime::env_config::section::Profile",
    "aws_smithy_async::rt::sleep::AsyncSleep",
    "aws_smithy_async::time::TimeSource",
+   "aws_smithy_http_client::test_util::replay::StaticReplayClient",
```
aajtodd: Are these still exposed?

aajtodd (approved):

aws/rust-runtime/aws-config/external-types.toml
landonxjames: Removed in https://github.com/smithy-lang/smithy-rs/pull/4238/commits/eca63eae0d9fe548df3d899bb8cd10d5daa6f120

aws/sdk/integration-tests/s3/tests/credential_features.rs
landonxjames: Updated with text in https://github.com/smithy-lang/smithy-rs/pull/4238/commits/eca63eae0d9fe548df3d899bb8cd10d5daa6f120

PR #4253, opened 2025-08-07
add https proxy support

aajtodd:
## Motivation and Context

https://github.com/awslabs/aws-sdk-rust/issues/169

## Description
* Adds support for automatic support for common proxy env vars (`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, `NO_PROXY` and their lowercase versions) behind a new `BehaviorVersion`
* relocates the TLS providers to their own modules, these are largely the same with the biggest change being `wrap_connector` returns dedicated connector types for `rustls` and `s2n` and the addition of new `connect` modules containing the proxy tunnel logic for each TLS provider.
    * This results in a bit of repeated logic to handle proxy tunneling but I couldn't find a way to unify the implementations and abstract out just the manual TLS handshake due to type system constraints
 * updates dependency versions in `aws-smithy-http-client`
* added a new `build_with_connector_fn` that allows creating a `SharedHttpClient` by passing your own function for creating connectors. We didn't have a way to configure `aws_smithy_http_client::Connector` and turn it into a `SharedHttpClient` without going through [http_client_fn](https://github.com/smithy-lang/smithy-rs/blob/release-2025-08-04/rust-runtime/aws-smithy-runtime-api/src/client/http.rs#L102). That works but it doesn't re-use the same caching logic for connectors as we do when going through `aws_smithy_http_client::Builder`. The alternative would be to make the client builder in aws-smithy-http-client behavior version aware. Doing it the way it is in the PR allows us to externalize this though and handle it in `aws-smithy-runtime` when creating the [default https client](https://github.com/smithy-lang/smithy-rs/blob/74964aac66ce6833a422d2be446e1aa8d01ba5cc/rust-runtime/aws-smithy-runtime/src/client/http.rs#L87)

## Testing

* integration tests with mock proxy over localhost
* tested against `mitmproxy` using different combinations of env variables and behavior versions

## Checklist

----


rust-runtime/aws-smithy-http-client/Cargo.toml
```diff
@@ -65,33 +66,35 @@ legacy-test-util = [
 
 legacy-rustls-ring = ["dep:legacy-hyper-rustls", "dep:legacy-rustls", "hyper-014"]
 
-rustls-ring = ["dep:rustls", "rustls?/ring", "dep:hyper-rustls", "default-client"]
-rustls-aws-lc = ["dep:rustls", "rustls?/aws_lc_rs", "dep:hyper-rustls", "default-client"]
-rustls-aws-lc-fips = ["dep:rustls", "rustls?/fips", "dep:hyper-rustls", "default-client"]
-s2n-tls = ["dep:s2n-tls", "dep:s2n-tls-hyper", "default-client"]
+rustls-ring = ["dep:rustls", "rustls?/ring", "dep:hyper-rustls", "dep:tokio-rustls", "default-client"]
```
landonxjames, line 69: Feels kind of weird adding `tokio` dependencies behind a flag that isn't `rt-tokio`. Would it make sense to introduce a `rt-tokio` flag that adds these deps and have these features enable that feature?

landonxjames (approved):
Looks good, one question about feature management, but not a blocker. 

Feel like this should release concurrently with the change from https://github.com/smithy-lang/smithy-rs/pull/4255 because that also feels like it might be an unexpected enough change to warrant a `BehaviorVersion`

rust-runtime/aws-smithy-http-client/Cargo.toml
aajtodd: I think it depends on whether the feature is bringing in the tokio runtime as a requirement vs using tokio IO traits. (for example `hyper-rustls` also [depends](https://github.com/rustls/hyper-rustls/blob/main/Cargo.toml#L34) on tokio)

aajtodd:
> Feel like this should release concurrently with the change from #4255 because that also feels like it might be an unexpected enough change to warrant a `BehaviorVersion` (I know the behavior version can't actually get that change since it is just a compile time dependency feature, but still feels like it should be indicated somehow)

Feature flags and behavior versions are not the best match but ya we can discuss what makes sense.

rust-runtime/aws-smithy-http-client/tests/proxy_tests.rs
```diff
@@ -0,0 +1,1105 @@
...
+/// Tests all-traffic proxy configuration
+/// Verifies that both HTTP and HTTPS requests go through all-traffic proxies
+#[tokio::test]
+async fn test_all_traffic_proxy() {
+    let mock_proxy = MockProxyServer::with_response(StatusCode::OK, "all traffic proxy").await;
+
+    // Configure proxy for all traffic
+    let proxy_config = ProxyConfig::all(format!("http://{}", mock_proxy.addr())).unwrap();
+
+    // Test 1: HTTP request should go through the proxy
```
ysaito1001: Will there be more tests given this comment explicit says Test 1, like for HTTPS?

ysaito1001:
> > Feel like this should release concurrently with the change from #4255 because that also feels like it might be an unexpected enough change to warrant a `BehaviorVersion` (I know the behavior version can't actually get that change since it is just a compile time dependency feature, but still feels like it should be indicated somehow)
> 
> Feature flags and behavior versions are not the best match but ya we can discuss what makes sense.

Good point. My gut is this PR can go independently of PR4255; if we just upgraded rustls to `0.24` (say PR4255 didn't exist), we probably wouldn't add a new `BehaviorVersion` for a dependency upgrade.

rust-runtime/aws-smithy-http-client/src/client/proxy.rs
```diff
@@ -0,0 +1,811 @@
...
+struct ProxyAuth {
+    /// Username for authentication
+    username: String,
+    /// Password for authentication
+    password: String,
+}
+
+/// Errors that can occur during proxy configuration
+#[derive(Debug)]
+pub enum ProxyError {
```
ysaito1001: Does this error type need to be `#[non_exhaustive]`?

aajtodd: Yes good catch. I'm wondering if we even need this or if we have a suitable error type already, will double check.

ysaito1001 (approved):
Awesome work. While dev guide *could* be useful, the API rustdoc seems good for now.

PR #4257, opened 2025-08-09
Use `crc-fast` v1.4.0 for 2X performance boost

onethumb:
## Motivation and Context
`crc-fast` v1.4.0 doubles throughput to >100GiBs on modern x86_64 systems (Intel Ice Lake, AMD Zen4, and later) when using [Rust 1.89.0+](https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/) and its stabilized support for [AVX512](https://github.com/rust-lang/rust/pull/138940) intrinsics.

It uses conditional compilation to maintain backwards compatibility and performance for Rust versions prior to 1.89.0.

## Description
- Use latest v1.4.0 of the `crc-fast` library.
- Updated the Cargo lock files using `./gradlew aws:sdk:cargoUpdateAllLockfiles`.
- Full [release notes](https://github.com/awesomized/crc-fast-rust/releases/tag/1.4.0).

## Testing
Drop-in, non-breaking minor version update. Existing tests pass.

----

### DIFF
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -527,9 +527,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.31"
+version = "1.2.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3a42d84bb6b69d3a8b3eaacf0d88f179e1929695e1ad012b6cf64d9caaa5fd2"
+checksum = "2352e5597e9c544d5e6d9c95190d5d27738ade584fa8db0a16e130e5c2b5296e"
 dependencies = [
  "shlex",
 ]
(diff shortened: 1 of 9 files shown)


onethumb:
Pinging @landonxjames for review. Thanks! 👍

rcoh:
Consumers don't actually depend on these lockfiles—so this doesn't actually impact anything right? More of a "test that everything builds with 1.4"?

~However, this _does_ increase the MSRV beyond policy (for, I think, no downstream benefit?) since builders using newer Rustc will pick these improvements up anyway?~

edit: I see that this is happening internally via rustc_version gating.

I think the main reason to avoid merging this, at least for the next few weeks, is that it prevents a consumer from forcing the 1.3.0 dependency in the event they are broken by the 1.4.0 changes.

onethumb:
> edit: I see that this is happening internally via rustc_version gating.
> 
> I think the main reason to avoid merging this, at least for the next few weeks, is that it prevents a consumer from forcing the 1.3.0 dependency in the event they are broken by the 1.4.0 changes.

Is the "next few weeks" timeline just to give things (Rust 1.89.0, this package, etc) time to shake out? Or is there some other change happening in a few weeks that would impact this? (I'm not in any sort of rush, just trying to make sure this doesn't slip off my radar if it languishes).

FWIW, our [GH Actions workflows](https://github.com/awesomized/crc-fast-rust/actions/runs/16852627434) have good version (1.81, 1.89, stable, nightly) and architecture (x86, x86_64, aarch64, powerpc) coverage, and the internal test coverage is extensive, so I think the odds of breakage are low.

ysaito1001:
Closing in favor of https://github.com/smithy-lang/smithy-rs/pull/4297

PR #4263, opened 2025-08-13
Make `TokenBucket` and `ClientRateLimiter` configurable

ysaito1001:
## Motivation and Context
The SDK provides basic retry configuration through the [RetryConfig](https://docs.rs/aws-smithy-types/latest/aws_smithy_types/retry/struct.RetryConfig.html) struct, allowing users to set parameters like retry mode, maximum attempts, and backoff durations. However, there has been a need for more advanced configuration options, particularly for token buckets and adaptive rate limiting.

## Description
This PR makes `TokenBucket` and `ClientRateLimiter` configurable through the [RetryPartition](https://docs.rs/aws-smithy-runtime/latest/aws_smithy_runtime/client/retries/struct.RetryPartition.html) struct.

In most cases, a "default" `RetryPartition` is used where `TokenBucket` and `ClientRateLimiter` are constructed behind the scenes and registered in [static maps](https://github.com/smithy-lang/smithy-rs/blob/035fcb23a4e4f171285f9f09a3cdb2cf738d6bdc/rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs#L32-L36). If a customer wishes to configure a `TokenBucket` and/or `ClientRateLimiter`, it can be done by a custom `RetryPartition`:
```
Config::builder().retry_partition(
    RetryPartition::custom("test")
        .token_bucket(...)
        .build()
)
```

For more details, refer to the API docs on [the setter](https://github.com/smithy-lang/smithy-rs/blob/e96e2dd25f50ab5eeee1fa92a755bb7e7891bce3/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/customizations/ResiliencyConfigCustomization.kt#L271-L315) and [the RetryPartition struct itself](https://github.com/smithy-lang/smithy-rs/blob/e96e2dd25f50ab5eeee1fa92a755bb7e7891bce3/rust-runtime/aws-smithy-runtime/src/client/retries.rs#L25-L35).

### End-to-end customer experience
Customers can use a custom partition with an SDK like so
```
    let shared_config = aws_config::defaults(BehaviorVersion::latest()).load().await;
    let unlimited_token_bucket = TokenBucket::unlimited();
    let custom_retry_partition = RetryPartition::custom("my-partition")
        .token_bucket(unlimited_token_bucket)
        .build();

    let conf1 = aws_sdk_s3::config::Builder::from(&shared_config)
        .retry_partition(custom_retry_partition.clone())
        .build();
    let client1 = aws_sdk_s3::Client::from_conf(conf1);
    // call operation on `client1`, which uses `unlimited_token_bucket` in the custom retry partition


    let conf2 = aws_sdk_s3::config::Builder::from(&shared_config)
        // use the same `custom_retry_partition` as `conf1`
        .retry_partition(custom_retry_partition)
        // additional configs different from `conf1` go here...
        .build();
    let client2 = aws_sdk_s3::Client::from_conf(conf2);
    // call operation on `client2`, which uses `unlimited_token_bucket` being shared as `client1`


    let conf3 = aws_sdk_s3::config::Builder::from(&shared_config)
        // custom partition also named `my-partition` but is a different instance than `custom_retry_partition` above
        .retry_partition(
            RetryPartition::custom("my-partition")
                .token_bucket(TokenBucket::default())
                .build(),
        )
        .build();
    let client3 = aws_sdk_s3::Client::from_conf(conf3);
    // call operation on `client3`, which uses the default `TokenBucket` in the custom partition.


    let client4 = aws_sdk_s3::Client::new(&shared_config);
    // call operation on `client4`, which uses the default `TokenBucket` associated with a default retry partition named "s3-<region>"
```

### Dismissed alternatives
We considered alternative approaches for making these configurable but dismissed them for the following reasons:
#### Alt1: Make them configurable through `RetryConfig`
```
RetryConfig::standard().token_bucket(...).client_rate_limiter(...)
```
`RetryConfig` is defined in `aws-smithy-types` whereas `TokenBucket` and `ClientRateLimiter` in `aws-smithy-runtime`. Since `aws-smithy-types` is higher in the dependency hierarchy, it cannot depend on `aws-smithy-runtime` to refer to `TokenBucket` and `ClientRateLimiter`. Moving `TokenBucket` and `ClientRateLimiter` to `aws-smithy-types` was probably not ideal either, as they are more closely tied to the smithy runtime.

#### Alt2: Make them configurable through individual setters on config builder
```
Config::builder().retry_config(...).retry_partition(...).token_bucket(...).client_rate_limiter(...)
```
This significantly increases the cognitive load for customers, as they see multiple configuration options at the same level.

## Testing
- Added unit tests for `TokenBucket` in `rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs`
- Added codegen tests for configurable `TokenBucket` and `ClientRateLimiter` in `aws/sdk-codegen/src/test/kotlin/software/amazon/smithy/rustsdk/RetryPartitionTest.kt`.

## Checklist

----


rust-runtime/aws-smithy-runtime/src/client/retries.rs
```diff
@@ -15,32 +15,185 @@ mod token_bucket;
...
+/// To configure these components, use a custom retry partition created via [`RetryPartition::custom`].
+/// Custom partitions allow full control over token bucket and rate limiter.
+///
+/// Two `RetryPartition`s that are equal, as defined by the `Eq` trait, share the same token bucket and client rate limiter.
+/// This means that the token bucket and rate limiter in a custom retry partition are independent
+/// of those in a default retry partition, even if the custom partition has the same name as the default partition.
 #[non_exhaustive]
-#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+#[derive(Clone, Debug)]
 pub struct RetryPartition {
```
aajtodd, line 42: This still leaves the possibility that a user creating two different clients that wants to assign them different token buckets or rate limiters results in "first access wins" right?

e.g.

```rust
// configure client 1 with custom token bucket
let client1 = Client::builder().retry_partition(RetryPartition::custom("part1").token_bucket(...).build());

// configure second client with different token bucket but same partition name
let client2 = Client::builder().retry_partition(RetryPartition::custom("part1").token_bucket(...).build());

```

It may not be obvious that client2 will never use the given token bucket. 

----

I'm not necessarily saying we shouldn't do this I'm just trying to understand the sharp edges of this approach (all of the approaches discussed have some sort of sharp edge). Also wondering how to best mitigate this if possible.

rust-runtime/aws-smithy-runtime/src/client/retries/client_rate_limiter.rs
```diff
@@ -255,126 +263,134 @@ struct Builder {
...
-    fn maximum_bucket_capacity(mut self, maximum_bucket_capacity: f64) -> Self {
-        self.maximum_bucket_capacity = Some(maximum_bucket_capacity);
+    /// The current capacity of the token bucket. The minimum this can be is 1.0
+    pub fn current_bucket_capacity(mut self, current_bucket_capacity: f64) -> Self {
+        self.set_current_bucket_capacity(Some(current_bucket_capacity));
         self
     }
-    ///The current capacity of the token bucket. The minimum this can be is 1.0
-    fn set_current_bucket_capacity(&mut self, current_bucket_capacity: Option<f64>) -> &mut Self {
+    /// The current capacity of the token bucket. The minimum this can be is 1.0
```
aajtodd: Do we actually validate this and other constraints anywhere?

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -74,3 +94,105 @@ impl TokenBucket {
...
+    regeneration_amount: Option<usize>,
+}
+
+impl TokenBucketBuilder {
+    /// Creates a new `TokenBucketBuilder` with default values.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Sets the specified initial capacity for the builder.
```
aajtodd: initial or max?

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -64,8 +81,11 @@ impl TokenBucket {
 
     pub(crate) fn regenerate_a_token(&self) {
         if self.semaphore.available_permits() < (self.max_permits) {
-            trace!("adding {PERMIT_REGENERATION_AMOUNT} back into the bucket");
-            self.semaphore.add_permits(PERMIT_REGENERATION_AMOUNT)
+            trace!(
+                "adding {regeneration_amount} back into the bucket",
+                regeneration_amount = self.regeneration_amount
+            );
+            self.semaphore.add_permits(self.regeneration_amount)
```
aajtodd: This setting seems odd or under specified. Regeneration amount when/over what time frame? I believe this function `regenerate_a_token` primarily existed to [add back permits from nothing](https://github.com/smithy-lang/smithy-rs/blob/release-2025-08-11/rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs#L213) for specific edge case. How would you describe the behavior to someone trying to create a custom token bucket?

I'm sort of wondering if `refill_units_per_second` (possibly combined with `circuit_breaker_mode` like Kotlin has) wouldn't be more understandable for a customer in terms of behavior but I'm trying to wrap my head around this knob and if it's simpler and still does what we want/need.

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -74,3 +94,105 @@ impl TokenBucket {
...
+        assert!(bucket.acquire(&ErrorKind::TransientError).is_some());
+
+        // Should have maximum capacity
+        assert_eq!(bucket.max_permits, Semaphore::MAX_PERMITS);
+
+        // Should have zero retry costs
+        assert_eq!(bucket.retry_cost, 0);
+        assert_eq!(bucket.timeout_retry_cost, 0);
+
+        // The loop count is arbitrary; should obtain permits without limit
```
aajtodd, line 154: Could also maybe check available permits after acquiring them in this loop? They should be unchanged right?

rust-runtime/aws-smithy-runtime/src/client/retries.rs
ysaito1001: > two different clients that wants to assign them different token buckets or rate limiters results in "first access wins" right?

Yeah, this is just like any other configurations. Instead of sharing a config value, if two clients are provided a separate config value, whichever makes to the config bag wins.

ysaito1001: 8118cab clarifies the rustdoc per this discussion.

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
ysaito1001: Good catch. It's better to not expose it, at least not at this time, since the setting likely confuses customers. Updated in cf71d46.

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
ysaito1001: Yes, test added in 264bb49.

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
ysaito1001: Updated in 26f35a4

rust-runtime/aws-smithy-runtime/src/client/retries/client_rate_limiter.rs
ysaito1001: Should've inspected when opening up the builder to `pub`. The minimum value constraint is enforced in the [update_bucket_refill_rate](https://github.com/smithy-lang/smithy-rs/blob/01aece8b664968ac63354e52b88915e8f8546738/rust-runtime/aws-smithy-runtime/src/client/retries/client_rate_limiter.rs#L178-L195) function. However, `ClientRateLimiter` itself is a state machine, and I realized that we cannot expose all of the builder methods for that value constraint (and other constraints) to work properly.

b324fa6 reduces the surface area of public API for `ClientRateLimiterBuilder`. The basic idea is that customers can tweak bucket capacities and fill rates but cannot alter timestamp-related fields or whether the rate limiter is enabled (we can consider exposing them gradually, should the need arise).

rust-runtime/aws-smithy-runtime/src/client/retries.rs
ysaito1001: Alternatively, we can consider changing `RetryPartitionInner` to be
```
pub(crate) enum RetryPartitionInner {
    Named(Cow<'static, str>),
    Anonymous {
        token_bucket: Option<TokenBucket>,
        client_rate_limiter: Option<ClientRateLimiter>,
    },
}
```
Anonymous `RetryPartition`s aren't equal to each other (just like anonymous functions are all different). This way, customers cannot give a name and it may help avoid confusion that they are different even though they have the same custom partition names.

EDIT:
Since `RetryPartition` implements the `Hash` trait, the `Anonymous` variant needs some sort of unique ID, something like
```
pub(crate) enum RetryPartitionInner {
    Named(Cow<'static, str>),
    Anonymous {
        _id: Arc<()>, // cloned `RetryPartition::Anonymous`s are equal
        token_bucket: Option<TokenBucket>,
        client_rate_limiter: Option<ClientRateLimiter>,
    },
}
```

aajtodd: Hmm need to think on this, that may be worth considering. Let's discuss offline.

PR #4272, opened 2025-08-18
Custom `ValidationException` RFC

drganjoo:
[Rendered](https://github.com/smithy-lang/smithy-rs/blob/custom-validation-rfc/design/src/rfcs/rfc0047_custom_validation.md)

Motivation and Context

Add an RFC to support custom `ValidationException`. This will allow server developers to use a custom structure Smithy shape as a `ValidationException`.

----

### DIFF
--- a/design/src/SUMMARY.md
+++ b/design/src/SUMMARY.md
@@ -68,5 +68,6 @@
   - [RFC-0043: Identity Cache Partitions](./rfcs/rfc0043_identity_cache_partitions.md)
   - [RFC-0044: Environment-defined service configuration](./rfcs/rfc0044_env_defined_service_config.md)
   - [RFC-0045: Configurable Serde](./rfcs/rfc0045_configurable_serde.md)
+  - [RFC-0047: Custom ValidationException](./rfcs/rfc0047_custom_validation.md)
 - [Contributing](./contributing/overview.md)
   - [Writing and debugging a low-level feature that relies on HTTP](./contributing/writing_and_debugging_a_low-level_feature_that_relies_on_HTTP.md)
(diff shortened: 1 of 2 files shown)


design/src/rfcs/rfc0047_custom_validation.md
```diff
@@ -0,0 +1,565 @@
...
+[`intEnum` shape]: https://smithy.io/2.0/spec/simple-types.html#intenum
+[`structure` shape]: https://smithy.io/2.0/spec/aggregate-types.html#structure
+
+The user experience if this RFC is implemented
+----------------------------------------------
+
+Currently, if there is a constrained shape in the input shape closure of an operation, the operation **must** add `smithy.framework#ValidationException` to the operation's error list, otherwise a build error is raised:
+
+```
+Caused by: ValidationResult(shouldAbort=true, messages=[LogMessage(level=SEVERE, message=Operation com.aws.example#GetStorage takes in input that is constrained (https://smithy.io/2.0/spec/constraint-traits.html), and as such can fail with a validation exception. You must model this behavior in the operation shape in your model file.
```
aajtodd, line 38: if this comes from smithy how are you disabling or getting around this validation? Also this seems like something we'd want to upstream to smithy anyway right?

drganjoo: No, we do this as part of the server code here: https://github.com/smithy-lang/smithy-rs/blob/f18c70d36c40fa0f40860547394c134566704e69/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ValidateUnsupportedConstraints.kt#L224

design/src/rfcs/rfc0047_custom_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+**Breaking changes:**
+- **Adding custom validation exception traits to an existing service is a breaking change** for the API contract
+- Clients expecting the standard `ValidationException` format will receive the new custom format
+- Service teams must coordinate with client teams when migrating to custom validation exceptions
+
+How to actually implement this RFC
+----------------------------------
+
+### 1. Create validation exception traits
```
jasgin, line 232: Does it makes sense/what are the implications of getting these implemented upstream in https://github.com/smithy-lang/smithy/tree/main/smithy-model/src/main/java/software/amazon/smithy/model/traits?

PR #4274, opened 2025-08-19
Add `aws-smithy-dns` crate and `HickoryDnsResolver`

landonxjames:
## Motivation and Context

Add an optional in-memory `HickoryDnsResolver` to `aws-smithy-runtime`. This is mostly targeted at supporting customers on AL2023 since it has disabled OS level DNS caching by default. Hyper relies on OS DNS caching if no DNS resolver is provided and this has led to some DNS throttling for users. See links below for examples:
* https://www.madebymikal.com/amazon-linux-2023-dns-and-systemd-resolved-a-story-of-no-caching/
* https://repost.aws/questions/QUXJ5yuiRpQE2c2Htf4Dc9iA/dns-caching-in-amazon-linux-2023
* https://repost.aws/knowledge-center/dns-resolution-failures-ec2-linux

We leverage the [hickory-resolver](https://crates.io/crates/hickory-resolver) crate to provide this functionality. 

## Description

Add new `HickoryDnsResolver` under `aws_smithy_dns`. This wraps a `hickory_resolver::Resolver` and provides some minimal configuration options (timeouts, retries, etc.) that seem unlikely to ever go away even though the `hickory-dns` ecosystem is still pre-1x. 

## Testing

Added new integration tests at `tests/hickory.rs`. These tests spin up an in-process DNS server on a random port and query it over `localhost`. 

Had to fix `CrateSetTest.kt` since the TOML library we use in Kotlin won't parse Cargo.toml files with `cfg` statements like `[target.'cfg(not(target_family = "wasm"))'.dependencies]`. I had to add one of those statements to prevent `hickory-resolver` from being compiled for WASM targets since it was introducing a tokio feature that caused compilation to fail and that caused `aws-smithy-runtime` to fail in the stability test.

## Checklist

----


rust-runtime/aws-smithy-runtime/Cargo.toml
```diff
@@ -17,7 +17,7 @@ http-auth = ["aws-smithy-runtime-api/http-auth"]
 connector-hyper-0-14-x = ["dep:aws-smithy-http-client", "aws-smithy-http-client?/hyper-014"]
 tls-rustls = ["dep:aws-smithy-http-client", "aws-smithy-http-client?/legacy-rustls-ring", "connector-hyper-0-14-x"]
 default-https-client = ["dep:aws-smithy-http-client", "aws-smithy-http-client?/rustls-aws-lc"]
-rt-tokio = ["tokio/rt"]
+rt-tokio = ["tokio/rt", "dep:hickory-resolver"]
```
aajtodd: why would `rt-tokio` pull in `hickory-resolver`? If anything a new feature for this that enables `rt-tokio` would make more sense.

rust-runtime/aws-smithy-runtime/src/client/dns.rs
```diff
@@ -6,44 +6,13 @@
...
-            })
-        }
-    }
-}
+#[cfg(all(feature = "rt-tokio", not(target_family = "wasm")))]
+mod caching;
 
 #[cfg(all(feature = "rt-tokio", not(target_family = "wasm")))]
-pub use self::tokio::TokioDnsResolver;
+pub use self::caching::CachingDnsResolver;
```
aajtodd: We called the default/tokio resolver `TokioDnsResolver`, do we want this to be called `CachingDnsResolver` or should it be more apparent that it's tied to hickory?

Calling it `CachingDnsResolver` makes sense but it also makes me think why wouldn't someone want caching, that seems like it should be the default or even configurable on the resolver (and frankly we could make tokio DNS resolver a caching resolver by layering on top of it so the "caching" part of this isn't really the differentiating factor).

rust-runtime/aws-smithy-runtime/tests/dns_caching.rs
```diff
@@ -0,0 +1,333 @@
...
+    let second_ips = second_result.unwrap();
+
+    // Verify same IPs returned
+    assert_eq!(first_ips, second_ips);
+
+    // Verify correct IP returned
+    assert_eq!(vec![IpAddr::V4(Ipv4Addr::new(1, 2, 3, 4))], first_ips);
+
+    // Cache hit should be faster
+    assert!(second_duration < first_duration);
```
aajtodd: You could make this a stronger assertion by adding state to the test dns server on the number of queries made or responses returned, and assert on that.

rust-runtime/aws-smithy-runtime/Cargo.toml
landonxjames: I am mostly just hesitant to add Cargo features for smaller functionality like this, especially since we `cargo hack test --feature-powerset` this crate, so new features incur a pretty substantial testing cost. 

I can add a new `hickory` (or `hickory-dns`?) feature if we think it is worth it.

rust-runtime/aws-smithy-runtime/src/client/dns.rs
landonxjames: I left it generic so we could replace the underlying resolver with something that isn't hickory in the future if we desired. I feel like tokio is kind of a defacto pillar of the ecosystem, so it is fine to reference it, but hickory doesn't quite have that name recognition yet. 

No strong feelings here though and happy to change it.

rust-runtime/aws-smithy-runtime/tests/dns_caching.rs
landonxjames: Good call, I struggled with the non determinism of measuring the times, I'll add in a closure that is called on `Drop` that can assert on a record of the received queries.

aajtodd: You could also keep track of it via atomics/mutex and make it available on the handle.

rust-runtime/aws-smithy-runtime/Cargo.toml
aajtodd: We do but we also [exclude some features](https://github.com/smithy-lang/smithy-rs/blob/release-2025-08-18/rust-runtime/aws-smithy-runtime/additional-ci#L13).

IDK it just struck me as odd as that would be what we tell people to enable as a feature to get the resolver.

rust-runtime/aws-smithy-runtime/src/client/dns.rs
aajtodd: Yeah I figured and I'm conflicted on this one. Need to think about it.

rust-runtime/aws-smithy-runtime/tests/dns_caching.rs
landonxjames: Added a `query_count` method to the server handle and updated tests to assert against that in https://github.com/smithy-lang/smithy-rs/pull/4274/commits/5378147ffef1a52f8395b4645d2c85b15c201f54

rust-runtime/aws-smithy-runtime/Cargo.toml
landonxjames: Added a `hickory-dns` feature in https://github.com/smithy-lang/smithy-rs/pull/4274/commits/5378147ffef1a52f8395b4645d2c85b15c201f54

rust-runtime/aws-smithy-runtime/src/client/dns.rs
landonxjames: We discussed offline and decided to move this to an unstable `aws-smithy-dns` crate, rename it to `HickoryDnsResolver`, and keep it behind a `hickory-dns` feature. This will allow us to deprecate it if hickory development stops or moves in a direction that no longer meets our needs.

landonxjames: Relocated in https://github.com/smithy-lang/smithy-rs/pull/4274/commits/534c96deaa8011bcad345bc2963f56a9c597e772

rust-runtime/aws-smithy-dns/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "aws-smithy-dns"
+version = "0.1.0"
```
ysaito1001, line 3: Nice, can start as unstable.

.changelog/caching-dns-resolver.md
```diff
@@ -0,0 +1,12 @@
...
+  - client
+authors:
+  - landonxjames
+references:
+  - smithy-rs#4274
+breaking: false
+new_feature: true
+bug_fix: false
+---
+Add a new crate, `aws-smithy-dns` that contains a `HickoryDnsResolver`. This wraps a `hickory_resolver::Resolver` and provides some minimal configuration options (timeouts, retries, etc.) Instructions for overriding the DNS resolver on your HTTP client can be found in our documentation at https://docs.aws.amazon.com/sdk-for-rust/latest/dg/http.html#overrideDns
```
teohhanhui, line 12: If one tries to follow the example and do something like:

```rust
use aws_smithy_dns::HickoryDnsResolver;
use aws_smithy_http_client::tls::rustls_provider::CryptoMode;
use aws_smithy_http_client::{tls, Builder};

#[tokio::main]
async fn main() {
    let http_client = Builder::new()
        .tls_provider(tls::Provider::Rustls(CryptoMode::AwsLc))
        .build_with_resolver(HickoryDnsResolver::default());

    let config = aws_config::defaults(aws_config::BehaviorVersion::latest())
        .http_client(http_client)
        .load()
        .await;
}
```

You'd get this error:

```
error[E0277]: the trait bound `aws_smithy_dns::HickoryDnsResolver: std::clone::Clone` is not satisfied
```

So this is not actually usable in practice.

landonxjames: Thanks for the heads up, this was fixed in https://github.com/smithy-lang/smithy-rs/pull/4289 and should be released in the next few days

landonxjames: Until it is released you should be able to wrap it in a [SharedDnsResolver](https://docs.rs/aws-smithy-runtime-api/1.9.0/aws_smithy_runtime_api/client/dns/struct.SharedDnsResolver.html) to get around your issue.

PR #4275, opened 2025-08-19
fix sdk benchmarks and add checks to ci

aajtodd:
## Description

Yesterday's release failed internally due to sdk-perf benchmark failing to build. This PR updates the lockfiles and adds building the benchmarks to CI to close this gap between internal release pipeline and CI.

----

### DIFF
--- a/tools/ci-scripts/check-aws-sdk-benchmarks
+++ b/tools/ci-scripts/check-aws-sdk-benchmarks
@@ -1,0 +1,25 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+C_YELLOW='\033[1;33m'
+C_RESET='\033[0m'
+
+set -eux
+cd smithy-rs
+
+./gradlew :aws:sdk:assemble
+
+# Build all benchmark directories
+pushd aws/sdk/benchmarks
+for dir in */; do
+    if [ -f "$dir/Cargo.toml" ]; then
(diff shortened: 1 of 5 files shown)


tools/ci-scripts/check-aws-sdk-benchmarks
```diff
@@ -0,0 +1,25 @@
...
+C_YELLOW='\033[1;33m'
+C_RESET='\033[0m'
+
+set -eux
+cd smithy-rs
+
+./gradlew :aws:sdk:assemble
+
+# Build all benchmark directories
+pushd aws/sdk/benchmarks
```
landonxjames, line 16: In addition to this check should we add the Cargo.locks in `aws/sdk/benchmarks` to the `cargoUpdateAllLockfiles` task so it is run every week with the other updates?

landonxjames (approved):
One question, non-blocking

tools/ci-scripts/check-aws-sdk-benchmarks
ysaito1001: Good point. We can create an issue/enhancement to handle lockfile updates in the following places:
- `rust-runtime/aws-smithy-fuzz/Cargo.lock` ([relevant PR](https://github.com/smithy-lang/smithy-rs/pull/4206))
- `aws/sdk/benchmarks` (this PR)

Whether updating the above lockfiles should be done with `cargoUpdateAllLockfiles` or with separate gradle tasks probably needs discussion/thinking. But updating lockfiles in `aws-smithy-fuzz` and `aws/sdk/benchmarks` is less critical compared to the other lockfiles we usually keep track of.

PR #4279, opened 2025-08-23
Centralize the default checksum algorithm in a single place

landonxjames:
## Motivation and Context

When I initially did the checksum work I meant to have the default checksum algorithm set in a single place to make it easy to update in the future. But it seems I missed actually getting that done, so I am doing it now. 

----

### DIFF
--- a/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpRequestChecksumDecorator.kt
+++ b/aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpRequestChecksumDecorator.kt
@@ -105,7 +105,7 @@ private fun HttpChecksumTrait.requestAlgorithmMember(
 // This is the single place where the default algorithm is set. It must be all uppercased.
 // There is some special logic in HttpChecksumTest and in the http_request_checksum inlineable
 // for testing the default case, if you change this value you will have to update that logic as well.
-const val defaultAlgorithm = "CRC32"
+const val DEFAULT_ALGORITHM = "CRC32"
 
 /**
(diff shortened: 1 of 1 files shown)


aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpRequestChecksumDecorator.kt
```diff
@@ -103,6 +102,11 @@ private fun HttpChecksumTrait.requestAlgorithmMember(
     return codegenContext.symbolProvider.toMemberName(checksumAlgorithmMemberShape)
 }
 
+// This is the single place where the default algorithm is set. It must be all uppercased.
+// There is some special logic in HttpChecksumTest and in the http_request_checksum inlineable
+// for testing the default case, if you change this value you will have to update that logic as well.
+const val defaultAlgorithm = "CRC32"
```
aajtodd: **nit/style**: Kotlin coding [conventions](https://kotlinlang.org/docs/coding-conventions.html#property-names) would usually have a const like this as `DEFAULT_ALGORITHM`.

aajtodd (approved):

aws/sdk-codegen/src/main/kotlin/software/amazon/smithy/rustsdk/HttpRequestChecksumDecorator.kt
landonxjames: Updated casing in https://github.com/smithy-lang/smithy-rs/pull/4279/commits/903c42c041bfe0b75081c2a4724e3682a1a70a19

Issue #4282, opened 2025-08-26
Idle connections on Hyper 1.0 are never closed


rcoh:
In Hyper 1.0, in order to enable closing idle connections you must do two things:
1. Set [`pool_timer`](https://docs.rs/hyper-util-spider/latest/hyper_util/client/legacy/struct.Builder.html#method.pool_timer)
2. Set [`pool_idle_timeout`](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Builder.html#method.pool_idle_timeout)

Today, if you use Hyper 1.0 we [set neither of these things](https://github.com/smithy-lang/smithy-rs/blob/dd10f0602682dfabafc465dbcd9eb92d3d915c51/rust-runtime/aws-smithy-http-client/src/client.rs#L167-L170)

PR #4284, opened 2025-08-27
Set `pool_idle_timeout` at 90 seconds for default hyper client

landonxjames:
## Motivation and Context

Fixes: https://github.com/smithy-lang/smithy-rs/issues/4282

## Description

Updating our default Hyper client to be configured with a [pool_timer](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Builder.html#method.pool_timer) and a [pool_idle_timeout](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Builder.html#method.pool_idle_timeout) value of 90 seconds.

Prior to Hyper going 1.0 and the legacy client moving into its own crate a `pool_timer` was not required. Now the pool won't timeout idle connections without one. The default value for the timeout in Hyper 0.14 was [90 seconds](https://docs.rs/hyper/0.14.32/hyper/client/struct.Builder.html#method.pool_idle_timeout), so we are copying that in our current settings. 

## Testing

This is difficult to test at the moment because Hyper does not support plugging in your own pool implementation, so we cannot instrument the pool itself. There has been some movement towards allowing that, so when that happens we could add tests for this behavior. For the time being we will have to be satisfied that the Hyper client tests this itself.

## Checklist

----

### DIFF
--- a/rust-runtime/aws-smithy-http-client/src/client.rs
+++ b/rust-runtime/aws-smithy-http-client/src/client.rs
@@ -195,8 +195,8 @@ impl<Any> ConnectorBuilder<Any> {
     {
         let client_builder = self.client_builder.unwrap_or_else(|| {
             let mut builder = hyper_util::client::legacy::Builder::new(TokioExecutor::new());
+            // Explicitly setting the pool_timer is required for connection timeouts to work.
             builder.pool_timer(TokioTimer::new());
-            builder.pool_idle_timeout(Duration::from_secs(90));
 
             builder
         });
--- a/.changelog/hyper-defaults.md
+++ b/.changelog/hyper-defaults.md
@@ -4,10 +4,9 @@ applies_to:
 authors:
   - landonxjames
 references:
-  - smithy-rs#4274
+  - smithy-rs#4282
 breaking: false
 new_feature: false
 bug_fix: true
 ---
-Set the `pool_idle_timeout` for the default Hyper client to 90 seconds. This aligns with the behavior of the hyper 0.14.x client that was previously the default.
-https://github.com/smithy-lang/smithy-rs/issues/4282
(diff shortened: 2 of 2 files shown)


rust-runtime/aws-smithy-http-client/src/client.rs
```diff
@@ -193,11 +193,13 @@ impl<Any> ConnectorBuilder<Any> {
...
     {
-        let client_builder =
-            self.client_builder
-                .unwrap_or(hyper_util::client::legacy::Builder::new(
-                    TokioExecutor::new(),
-                ));
+        let client_builder = self.client_builder.unwrap_or_else(|| {
+            let mut builder = hyper_util::client::legacy::Builder::new(TokioExecutor::new());
+            builder.pool_timer(TokioTimer::new());
+            builder.pool_idle_timeout(Duration::from_secs(90));
```
ysaito1001: Might be helpful to provide a link here as well:
https://docs.rs/hyper/0.14.32/hyper/client/struct.Builder.html#method.pool_idle_timeout

to help code readers see where the value 90 comes from. It's obvious today from the PR description or from git blame, but they tend to be buried overtime.

aajtodd: Looks like it's maybe still the [default](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Builder.html#method.pool_idle_timeout)? It just doesn't take effect without the timer now.

aajtodd (approved):

rust-runtime/aws-smithy-http-client/src/client.rs
landonxjames: Good catch, I deleted manually setting the 90 seconds in favor of the default behavior

PR #4288, opened 2025-09-07
Fix H2 URI path matching in ALB health check layer

drganjoo:
## Summary
Fixed URI path matching for H2 protocol in the ALB health check layer to ensure explicit path comparison.

## Changes
- Updated `alb_health_check.rs` to explicitly match URI paths for H2 requests
- Added proper path validation logic

## Testing
- Verified H2 health check requests are properly routed
- Confirmed backward compatibility with existing functionality

### DIFF
--- a/.changelog/dns-clone.md
+++ b/.changelog/dns-clone.md
@@ -1,9 +1,0 @@
----
-applies_to: ["client"]
-authors: ["landonxjames"]
-references: ["smithy-rs#4274"]
-breaking: false
-new_feature: false
-bug_fix: true
----
-The `HickoryDnsResolver` and `TokioDnsResolver` were not `Clone` making it impossible to use them in the http_client builder's `build_with_resolver` method.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 <!-- Do not manually edit this file. Use the `changelogger` tool. -->
+September 10th, 2025
+====================
+**New this release:**
+- :bug: (client, [smithy-rs#4274](https://github.com/smithy-lang/smithy-rs/issues/4274)) The `HickoryDnsResolver` and `TokioDnsResolver` were not `Clone` making it impossible to use them in the http_client builder's `build_with_resolver` method.
+
+
 August 28th, 2025
 =================
 **New this release:**
--- a/aws/SDK_CHANGELOG.next.json
+++ b/aws/SDK_CHANGELOG.next.json
@@ -6,48 +6,6 @@
   "smithy-rs": [],
   "aws-sdk-rust": [
     {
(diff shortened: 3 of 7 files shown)


rust-runtime/aws-smithy-http-server/src/layer/alb_health_check.rs
```diff
@@ -179,3 +179,32 @@ where
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use http::Uri;
+    use tower::{service_fn, ServiceExt};
+
+    #[tokio::test]
+    async fn health_check_matches_path_with_scheme_and_authority() {
```
aajtodd: For my own edification where is the "with scheme and authority" part of this test coming from?

drganjoo: The scheme and authority are intentionally hard-coded in this test. I've renamed the test to `health_check_matches_path_when_uri_contains_scheme_authority_and_path` to better reflect its purpose. 

This test specifically verifies that the health check correctly matches requests based solely on their path component, regardless of whether the URI also includes scheme and authority components. This ensures our health check behavior remains consistent across different URI.

rust-runtime/aws-smithy-http-server/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-http-server"
-version = "0.65.5"
+version = "0.66.0"
```
rcoh: why did you make a new breaking version? isn't this going to cause problems? This change seems backwards compatible?

drganjoo: You're absolutely right - this is a backwards compatible change. I bumped to 0.66 as a precautionary measure because I wanted to avoid any potential versioning issues during integration testing. At the time, this was planned to be included in the next code generator release, so the version bump seemed appropriate. However, since this now needs to be released independently of the code generator, you're correct that 0.65.6 would be more suitable.

rcoh: well and it seems like making a new breaking change release of the libraries when we don't need it only causes lots of pain?

drganjoo: We typically never release intermediate crate versions of `aws-smithy-http-server` independently of the code generator, so bumping the major version, combined with a code-generator release, is usually the safer approach - it eliminates any risk of breaking users due to potential CI tooling issues. When releases are tied to the code generator, a major version bump has minimal downstream impact compared to accidentally introducing breaking changes. 

However, since this particular release now needs to happen independently of the code generator, I've changed it to only bump the minor version.

PR #4298, opened 2025-09-11
Track features for account based endpoints

ysaito1001:
## Motivation and Context
This PR tracks and emits metrics for the following AWS SDK features (corresponding metric value in parenthesis):
- `ACCOUNT_ID_MODE_PREFERRED` (P)
- `ACCOUNT_ID_MODE_DISABLED` (Q)
- `ACCOUNT_ID_MODE_REQUIRED` (R)
- `RESOLVED_ACCOUNT_ID` (T)

## Testing
- [x] Unit test for tracking `RESOLVED_ACCOUNT_ID`
- [x] Integration tests to be added

----


aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AccountIdEndpointParamsDecorator.kt
```diff
@@ -234,6 +238,38 @@ class AccountIdEndpointModeBuiltInParamDecorator : ConditionalDecorator(
...
                             }
                         }
+
+                        override fun trackSdkFeatures(
+                            codegenContext: ClientCodegenContext,
+                            configBag: String,
+                        ) = writable {
+                            rustTemplate(
+                                """
+                                match cfg
```
aajtodd: This could probably be turned into a `pub(crate)` function that is re-used which would reduce code size a bit since it applies to every operation.

Also for some reason this code block is [rendered](https://d2luzm2xt3nokh.cloudfront.net/codegen-diff/1fe10969f786b1b53e4e1b8a40a2994e369566a4/27102829f69b2cfe3a9ca9e69ae64d1ea40f9865/aws-sdk/index_page_15.html#file-15-4) in between endpoint params.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
```diff
@@ -108,7 +109,11 @@ class EndpointParamsInterceptorGenerator(
                         #{param_setters}
                         .build()
                         .map_err(|err| #{ContextAttachedError}::new("endpoint params could not be built", err))?;
+
+                    #{track_sdk_features:W}
```
aajtodd: This seems like an odd place to tack this on. Why do this in `EndpointParamsInterceptorGenerator`?

ysaito1001: Do you consider alternative account id customization providing its own interceptor to do the feature tracking?

Tracking it within an interceptor is a single-sourced place where config values are finalized (e.g. after operation override). It's conceivable that we'll be asked to track feature values for what's passed to endpoint params in the future; either having a centralized place to do so like here in interceptor or having each customization like account id provide its own interceptor, depending on how we slice things.

aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AccountIdEndpointParamsDecorator.kt
ysaito1001: Good point.

> rendered in between endpoint params.

Yeah, coded that way before converting a service-specific params to the generic EndpointResolverParams. We can move it if you feel strongly about.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
ysaito1001: Hmm, I wonder if we can just track it in the `set_account_id_endpoint_mode` in the service config builder although the feature won't get tracked if the method isn't called (like creating a service config without `ConfigLoader`).

ysaito1001: Thinking about this a bit more, I feel tracking features in its own interceptor achieves the optimal separation of concerns, as opposed to tacking it on in endpoint params interceptor. We don't need `pub(crate)` helper either, since the logic is rendered once in the interceptor registered in service runtime plugin.

aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AccountIdEndpointParamsDecorator.kt
ysaito1001: Fixed in 2973203 where we track features in a separate interceptor.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/endpoint/generators/EndpointParamsInterceptorGenerator.kt
ysaito1001: Fixed in 2973203 where we track features in a separate interceptor.

PR #4299, opened 2025-09-11
add `RuleBuilder::then_compute_output`

greenwoodcm:
this allows the developer to compute a mocked output using content from the input.

## Motivation and Context
Some richer use cases for stubbing/mocking require computing a stubbed output using data from the input, for instance stubbing a method that returns the item that was put.

## Description
Add a new method to `RuleBuilder` that allows for providing a function that computes the output.  That function takes in a reference to the input as an argument.  I'm not sure how this works with the mutable inputs like streamed bytes, but seems to work for the simple stuff.

## Testing
added a simple unit test

----

### DIFF
--- a/aws/sdk/integration-tests/s3/tests/mocks.rs
+++ b/aws/sdk/integration-tests/s3/tests/mocks.rs
@@ -96,12 +96,10 @@ async fn test_mock_client_compute() {
             inp.bucket() == Some("test-bucket") && inp.key() == Some("correct-key")
         })
         .then_compute_output(|input| {
+            let content =
+                format!("{}.{}", input.bucket().unwrap(), input.key().unwrap()).into_bytes();
             GetObjectOutput::builder()
-                .body(ByteStream::from(format!(
-                    "{}.{}",
-                    input.bucket().unwrap(),
-                    input.key().unwrap()
-                )))
+                .body(ByteStream::from(content))
                 .build()
         });
 
@@ -121,7 +119,7 @@ async fn test_mock_client_compute() {
         .to_vec();
 
     assert_eq!(data, b"test-bucket.correct-key");
-    assert_eq!(s3_real_object.num_calls(), 1);
+    assert_eq!(s3_computed.num_calls(), 1);
 }
 
 #[tokio::test]


aws/sdk/integration-tests/s3/tests/mocks.rs
```diff
@@ -89,6 +89,70 @@ async fn test_mock_client() {
...
+                SdkBody::from(S3_NO_SUCH_KEY),
+            )
+        });
+
+    let s3_real_object = mock!(aws_sdk_s3::Client::get_object)
+        .match_requests(|inp| {
+            inp.bucket() == Some("test-bucket") && inp.key() == Some("correct-key")
+        })
+        .then_compute_output(|input| {
+            GetObjectOutput::builder()
```
rcoh, line 101: presumably this test should actually use the input?

rust-runtime/aws-smithy-mocks/Cargo.toml
```diff
@@ -9,7 +9,7 @@ repository = "https://github.com/smithy-lang/smithy-rs"
 
 [dependencies]
 aws-smithy-types = { path = "../aws-smithy-types" }
-aws-smithy-runtime-api = { path = "../aws-smithy-runtime-api", features = ["client", "http-1x"] }
+aws-smithy-runtime-api = { path = "../aws-smithy-runtime-api", features = ["client", "http-1x", "test-util"] }
```
rcoh, line 12: what uses the `test-util` in here?

rust-runtime/aws-smithy-mocks/src/interceptor.rs
```diff
@@ -444,6 +447,45 @@ mod tests {
...
+            .rule_mode(RuleMode::Sequential)
+            .with_rule(&rule);
+
+        let operation = create_test_operation(interceptor, true);
+
+        let result = operation
+            .invoke(TestInput::new("test-bucket", "test-key"))
+            .await;
+
+        // Should succeed with the final output after retries
```
aajtodd: theres no retries in this test

rust-runtime/aws-smithy-mocks/src/interceptor.rs
```diff
@@ -444,6 +447,45 @@ mod tests {
...
+            result.err()
+        );
+        assert_eq!(
+            result.unwrap(),
+            TestOutput {
+                content: "test-bucket.test-key".to_string()
+            }
+        );
+
+        // Verify the rule was used the expected number of times (all 4 responses: 2 errors + 1 success)
```
aajtodd: copy pasta?

rust-runtime/aws-smithy-mocks/src/rule.rs
```diff
@@ -318,7 +326,38 @@ where
...
+    /// let rule = mock!(Client::get_object)
+    ///     .sequence()
+    ///     .compute_output(|req| {
+    ///         GetObjectOutput::builder()
+    ///             .body(ByteStream::from_static(format!("content for {}", req.key().unwrap_or("unknown")).as_bytes()))
+    ///             .build()
+    ///     })
+    ///     .build();
+    /// ```
+    pub fn compute_output<F>(mut self, compute_fn: F) -> Self
```
aajtodd: I don't know that we want to allow this from the sequence builder.

It allows constructing invalid rules.

e.g.

```rs
    #[tokio::test]
    async fn test_compute_output_with_retry() {
        let rule = create_rule_builder()
            .match_requests(|input| input.bucket == "test-bucket" && input.key == "test-key")
            .sequence()
            .http_status(503, None)
            .compute_output(|input| {
                TestOutput {
                    content: format!("{}.{}", input.bucket, input.key),
                }
            })
            .build();

        // Create an interceptor with the rule
        let interceptor = MockResponseInterceptor::new()
            .rule_mode(RuleMode::Sequential)
            .with_rule(&rule);

        let operation = create_test_operation(interceptor, true);

        let result = operation
            .invoke(TestInput::new("test-bucket", "test-key"))
            .await;

        assert_eq!(
            result.unwrap(),
            TestOutput {
                content: "test-bucket.test-key".to_string()
            }
        );
    }

```

Fails with 
> Input type mismatch in compute_output
thread 'interceptor::tests::test_compute_output_with_retry' panicked at aws-smithy-mocks/src/rule.rs:358:17:
Input type mismatch in compute_output


I'm wondering if we just make this `pub crate` and disallow publicly creating sequences off it (which is the only way to make a rule work with retries).

rust-runtime/aws-smithy-mocks/src/interceptor.rs
greenwoodcm: bad copy/paste thanks, will fix

aajtodd:
> lgtm. I think my only question would be whether we want this API or something like:
> 
> ```rust
> .serve(fn(&Input) -> Option<Response>)
> ```
> 
> that would combine matching and responding.

I think that'll complicate things internally and externally (e.g. which takes precedence then between `match_requests` and `serve`). Also rules have to yield a response if they match. 

I originally explored a `serve` and `serve_n` like function that allowed you to use both input and yield multiple responses:

```rs
let rule = mock!(Client::get_object)
    .serve(|input, idx| {
        let key = input.key().unwrap();
        
        if key.starts_with("important") {
            // Important keys always succeed
            Some(GetObjectOutput::builder()
                .body(ByteStream::from_static(b"important data"))
                .build())
        } else if idx < 2 {
            // First two calls for non-important keys return 503
            Some(503)
        } else {
            // Subsequent calls succeed
            Some(GetObjectOutput::builder()
                .body(ByteStream::from_static(b"regular data"))
                .build())
        }
    })
    .build();
```

However, it won't work with retries and was IMO less ergonomic/readable as the sequence API we ended up with.

greenwoodcm:
pushed a new diff that addresses the comments above

rust-runtime/aws-smithy-mocks/README.md
```diff
@@ -120,6 +129,26 @@ repeat indefinitely, making the rule never exhaust.
 The sequence builder API provides a fluent interface for defining sequences of responses.
 After providing all responses in the sequence, the rule is considered exhausted.
 
+You can also use computed outputs in sequences:
+
+```rust,ignore
+let dynamic_rule = mock!(Client::get_object)
```
aajtodd: Outdated, remove

aws/sdk/integration-tests/s3/tests/mocks.rs
greenwoodcm: yep, realized that.  and CI noticed because it threw a warning that `input` was unused :) fixed in next rev.

rust-runtime/aws-smithy-mocks/Cargo.toml
greenwoodcm: hmm, seems like nothing...will remove, thanks

aajtodd: was used for `Input::doesnt_matter`

Issue #4265, opened 2025-08-13
aws-smithy-mocks `mock_client` does not support clients that don't have a region


rcoh:
https://docs.rs/aws-smithy-mocks/latest/src/aws_smithy_mocks/lib.rs.html#146

When generating a smithy-rs client outside of the Rust SDK, it won't have a region which makes `mock_client` fail to compile.

jhchee:
Is this still being worked on?

rcoh:
@aajtodd we should add `with_test_defaults_v2` that sets the region, then aws-smithy-mocks can use it

aajtodd:
Addressed in https://github.com/smithy-lang/smithy-rs/pull/4312, will go out in next release.

Issue #4189, opened 2025-06-25
docs(aws-smithy-mocks): Document that client packages should enable `test-util` feature, provide sample `Cargo.toml`


jlizen:
It's not obvious from the `aws-smithy-docs` (module-level at least), that you need a `test-util` feature enabled on any client packages. Without it, you get something like:
```
no method named with_test_defaults found for struct aws_sdk_s3::config::Builder in the current scope
method not found in Builder
```

The correct usage would look something like: 

```
[dependencies]
aws-sdk-s3 = "1"

[test-dependencies]
aws-smithy-mocks = "0.1"
aws-sdk-s3 = { version = "1", features = ["test-util"] }
```

It would be great to document this more prominently, with a sample `Cargo.toml`. Might also be nice to include the error message as a searchable breadcrumb.

aajtodd:
Addressed in https://github.com/smithy-lang/smithy-rs/pull/4312, will go out in next release.

PR #4312, opened 2025-09-22
apply `us-east-1` as default region if not set by user in `with_test_defaults_v2()` 

aajtodd:
## Motivation and Context

* https://github.com/smithy-lang/smithy-rs/issues/4265
* https://github.com/smithy-lang/smithy-rs/issues/4189

## Description
* Adds a new `with_test_defaults_v2()` method to all clients supporting region configuration. This new method applies `us-east-1` as default region if not set by user. This allows `aws-smithy-mocks` to work for non AWS SDK generated clients. This is only set as the default if a user has not already supplied one. 
    * NOTE: We are choosing to add a new method here for backwards compatibility concerns (our own generated endpoint tests break because of precedence problems between endpoint rules and with_test_defaults).
* Clarify `test-util` feature requirement when using `aws-smithy-mocks`.

## Checklist

----

### DIFF
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -561,18 +561,18 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.48"
+version = "4.5.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e2134bb3ea021b78629caa971416385309e0131b351b25e01dc16fb54e1b5fae"
+checksum = "f4512b90fa68d3a9932cea5184017c5d200f5921df706d45e853537dea51508f"
 dependencies = [
  "clap_builder",
 ]
 
 [[package]]
 name = "clap_builder"
-version = "4.5.48"
+version = "4.5.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2ba64afa3c0a6df7fa517765e31314e983f51dda798ffba27b988194fb65dc9"
+checksum = "0025e98baa12e766c67ba13ff4695a887a1eba19569aad00a472546795bd6730"
 dependencies = [
  "anstyle",
  "clap_lex",
@@ -580,9 +580,9 @@ dependencies = [
 
 [[package]]
 name = "clap_lex"
-version = "0.7.5"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b94f61472cee1439c0b966b47e3aca9ae07e45d070759512cd390ea2bebc6675"
+checksum = "a1d728cc89cf3aee9ff92b05e62b19ee65a02b5702cff7d5a377e32c6ae29d8d"
 
 [[package]]
 name = "const-oid"
@@ -951,9 +951,9 @@ dependencies = [
 
 [[package]]
 name = "generic-array"
-version = "0.14.7"
+version = "0.14.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "85649ca51fd72272d7821adaf274ad91c288277713d9c18820d8499a7ff69e9a"
+checksum = "4bb6743198531e02858aeaea5398fcc883e71851fcbcb5a2f773e2fb6cb1edf2"
 dependencies = [
  "typenum",
  "version_check",
@@ -1039,9 +1039,9 @@ dependencies = [
 
 [[package]]
 name = "half"
-version = "2.7.0"
+version = "2.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e54c115d4f30f52c67202f079c5f9d8b49db4691f460fdb0b4c2e838261b2ba5"
+checksum = "6ea2d84b969582b4b1864a92dc5d27cd2b77b622a8d79306834f1be5ba20d84b"
 dependencies = [
  "cfg-if",
  "crunchy",
--- a/aws/rust-runtime/aws-config/Cargo.lock
+++ b/aws/rust-runtime/aws-config/Cargo.lock
@@ -772,9 +772,9 @@ dependencies = [
 
 [[package]]
 name = "generic-array"
-version = "0.14.7"
+version = "0.14.9"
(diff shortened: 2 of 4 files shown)


rust-runtime/aws-smithy-mocks/README.md
```diff
@@ -14,6 +14,30 @@ without mocking the entire client or using traits.
...
+method not found in Builder
+```
+
+Example `Cargo.toml` using the `aws-sdk-s3` crate as the service client crate under test:
+
+```toml
+[dependencies]
+aws-sdk-s3 = "1"
+
+[test-dependencies]
```
landonxjames: `[dev-dependencies]`?

rust-runtime/aws-smithy-mocks/README.md
```diff
@@ -14,6 +14,30 @@ without mocking the entire client or using traits.
 - **Response Sequencing**: Define sequences of responses for testing retry behavior
 - **Rule Modes**: Control how rules are matched and applied
 
+## Prerequisites
+
+<div class="warning">
+You must enable the `test-util` feature of the service client crate in order to use the `mock_client` macro.
```
landonxjames: I know its kind of implicit in the example below that `test-util` should only be enabled for `[dev-dependencies]`, but should we make that explicit and warn users only to use those features for testing?

landonxjames (approved):
Some questions on the README, no blockers

rust-runtime/aws-smithy-mocks/README.md
rcoh: This happens so frequently that I wonder if we should design a really nice error where the mock_client invokes a macro we define in the crates and if test-util isn't enabled it uses compile_fail! To give a precise error.

Can't do it here obviously since we'd need to add that macro first in the generated crates

aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/RegionDecorator.kt
```diff
@@ -205,6 +205,17 @@ class RegionProviderConfig(codegenContext: ClientCodegenContext) : ConfigCustomi
                         *codegenScope,
                     )
                 }
+                is ServiceConfig.DefaultForTests -> {
+                    // this was added later, for backwards compat we only set a default if a region hasn't already been set by user
+                    rustTemplate(
+                        """
+                        if ${section.configBuilderRef}.config.load::<#{Region}>().is_none() {
+                            self.set_region(#{Some}(#{Region}::new("us-east-1")));
```
rcoh, line 213: Is it worth a BMV for this or other opt-out? I feel like there is a very high probability that _someones_ unit test will be broken by this and although the fix won't be hard they might appreciate a way to hold back to keep their tests working

aajtodd: Our own unit tests are [broken](https://github.com/smithy-lang/smithy-rs/actions/runs/17924115834/job/50966322553) by this and I'm not sure whether we want to ship this right now as we have introduced a precedence problem for defaults between `with_test_defaults` and endpoint builtins.

---

Generated test code:

```
// Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
#![cfg(feature = "test-util")]
#[::tokio::test]
async fn operation_input_test_test_operation_1() {
    /* documentation: region should fallback to the default */
    /* builtIns: {} */
    /* clientParams: {} */
    let (http_client, rcvr) = ::aws_smithy_http_client::test_util::capture_request(None);
    let conf = {
        #[allow(unused_mut)]
        let mut builder = endpoint_test_service::Config::builder().with_test_defaults().http_client(http_client);
        builder.build()
    };
    let client = endpoint_test_service::Client::from_conf(conf);
    let _result = dbg!(
        client
            .test_operation()
            .set_bar(::std::option::Option::Some(
                endpoint_test_service::types::Bar::builder()
                    .set_f(::std::option::Option::Some("blah".to_owned()))
                    .build()
            ))
            .send()
            .await
    );
    let req = rcvr.expect_request();
    let uri = req.uri().to_string();
    assert!(
        uri.starts_with("https://prod.us-east-2.api.myservice.aws.dev/"),
        "expected URI to start with `https://prod.us-east-2.api.myservice.aws.dev/` but it was `{}`",
        uri
    );
}
```

* Model: https://github.com/smithy-lang/smithy-rs/blob/3c11815e5e4d0ec8c048d4bde7af94b710039672/aws/sdk-adhoc-test/models/single-static-endpoint.smithy

failure:

```
running 1 test
test operation_input_test_test_operation_1 ... FAILED

failures:

---- operation_input_test_test_operation_1 stdout ----
[endpoint-test-service/rust-client-codegen/tests/endpoint_tests.rs:15:19] client.test_operation().set_bar(::std::option::Option::Some(endpoint_test_service::types::Bar::builder().set_f(::std::option::Option::Some("blah".to_owned())).build())).send().await = Ok(
    TestOperationOutput {
        _request_id: None,
    },
)

thread 'operation_input_test_test_operation_1' panicked at endpoint-test-service/rust-client-codegen/tests/endpoint_tests.rs:28:5:
expected URI to start with `https://prod.us-east-2.api.myservice.aws.dev/` but it was `https://prod.us-east-1.api.myservice.aws.dev/foo`


failures:
    operation_input_test_test_operation_1

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```


----

this test is generated by: [https://github.com/smithy-lang/smithy-rs/blob/2016a67af3beeb4c976da9fb3d867f15822e[…]/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt](https://github.com/smithy-lang/smithy-rs/blob/2016a67af3beeb4c976da9fb3d867f15822ea038/aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/endpoints/OperationInputTestGenerator.kt)
Interestingly it shows nothing for builtins

but effectively it's relying on the default builtin value of us-east-2which it would not be now because with_test_defaults sets a different default when unset

PR #4314, opened 2025-09-23
Add codegen-serde unit tests to CI

rcoh:
This PR adds the codegen-serde module to the GitHub Actions test matrix & fixes broken tests.

Along the way, I found some broken tests which I also fixed.

## Changes
- Added `check-serde-codegen-unit-tests` script in `tools/ci-scripts/`
- Added corresponding target to `ci.mk`
- Added the test to the GitHub Actions workflow matrix in `.github/workflows/ci.yml`

This follows the same pattern as other codegen unit test modules (codegen-core, codegen-client, codegen-server) and ensures that codegen-serde tests are run as part of CI.

### DIFF
--- a/codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeProtocolTestTest.kt
+++ b/codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeProtocolTestTest.kt
@@ -5,9 +5,7 @@
 
 package software.amazon.smithy.rust.codegen.serde
 
-import org.junit.jupiter.api.Test
 import org.junit.jupiter.params.ParameterizedTest
-import org.junit.jupiter.params.provider.Arguments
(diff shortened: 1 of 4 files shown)


codegen-serde/src/test/kotlin/software/amazon/smithy/rust/codegen/serde/SerdeProtocolTestTest.kt
```diff
@@ -5,46 +5,21 @@
...
-        @JvmStatic
-        fun testedModels(): Stream<Arguments> =
-            Stream.of(
-                Arguments.of(ShapeId.from(ServiceShapeId.REST_JSON)),
-                Arguments.of(ShapeId.from("com.amazonaws.constraints#ConstraintService")),
-            )
-    }
-
-    @Test
-    fun testRestJson() {
```
rcoh, line 38: this test is duplicated with the test below

ysaito1001 (approved):

Issue #4308, opened 2025-09-17
Fix issue with smithy union serialization in multiple protocols


landonxjames:
https://github.com/smithy-lang/smithy-rs/pull/4306 fixes a bug in `XmlBindingTraitSerializerGenerator.kt` exposed by a change to the [S3-Control model ](https://github.com/awslabs/aws-sdk-rust/commit/effdaf98b3eb49e8900838176c18f72db48644ff#diff-101f9ff3d2e67a2122db6bc609ccf4baffc6aee548757139badd8d1c41ab11ab).
The model change introduced a smithy union type that contained a unit struct as a member. This generated a match arm like:
```
crate::types::ObjectEncryptionFilter::Sses3(inner) => {...}
```

But `inner` is never used since there is nothing to serialize for a unit struct, so this caused a [warning and CI failure](https://github.com/smithy-lang/smithy-rs/actions/runs/17771171675/job/50511115119#step:4:1340).

A quick review of other `${protocol}SerializerGenerator` classes shows that they likely also contain the same bug:
* CBOR: https://github.com/smithy-lang/smithy-rs/blob/b570622ee6b7071071164826f8670f59f9cf0da7/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/CborSerializerGenerator.kt#L551-L555
* JSON: https://github.com/smithy-lang/smithy-rs/blob/b570622ee6b7071071164826f8670f59f9cf0da7/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt#L559-L562
* Query: https://github.com/smithy-lang/smithy-rs/blob/b570622ee6b7071071164826f8670f59f9cf0da7/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/QuerySerializerGenerator.kt#L361-L365

We should fix this for all of our protocols and add some codegen tests for it (the updated S3-control model implicitly tests it for XML, but explicit tests would be better).

ysaito1001:
Completed by https://github.com/smithy-lang/smithy-rs/pull/4333

PR #4316, opened 2025-09-24
Add union serialization tests for JSON, Query, and CBOR protocols

vcjana:
## Add test cases to expose unused variable warnings in union serialization
- Tests create minimal Smithy models with unions containing unit structs
- Use TestWorkspace pattern with `project.compileAndTest()` to validate generated code
- Cover JSON, AWS Query, and CBOR serialization protocols
- Tests designed to help validate serialization bug fixes

## Motivation and Context
This change addresses the need for comprehensive test coverage to expose unused variable warnings in union serialization across different protocols. The tests are designed to validate the serialization bug fix and ensure that union members with
unit structs are handled correctly without generating unused variable warnings in the generated Rust code.

This work supports the ongoing effort to improve code generation quality and eliminate compiler warnings in generated SDK code.

## Description
Added three new test cases across existing serializer test files to expose potential unused variable warnings in union serialization:

1. JsonSerializerGeneratorTest.kt: Added union with unit struct demonstrates serialization bug() and union with unit struct demonstrates cbor serialization bug() tests
2. AwsQuerySerializerGeneratorTest.kt: Added union with unit struct demonstrates query serialization bug() test

Each test:
• Creates a minimal Smithy model with a union containing both unit struct and data members
• Uses the TestWorkspace pattern with `testSymbolProvider(model)` and `project.compileAndTest()`
• Renders both the Unit structure and Union using renderWithModelBuilder() and UnionGenerator
• Validates that the generated Rust code compiles successfully
• Designed to expose unused variable warnings that would occur with the serialization bug

The tests follow the established patterns in the codebase and integrate seamlessly with existing test infrastructure.

## Testing
• Local development environment with Gradle 8.14.3
• Kotlin compilation and test execution via ./gradlew :codegen-core:test 
## Checklist

----


codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/AwsQuerySerializerGeneratorTest.kt
```diff
@@ -325,4 +328,33 @@ class AwsQuerySerializerGeneratorTest {
         }
         project.compileAndTest()
     }
+
+    @Test
+    fun `union with unit struct demonstrates query serialization bug`() {
```
landonxjames: I would probably change this to something like "Union with unit struct doesn't cause unused variable warning" or something along those lines, and maybe add a comment with a link to the [issue](https://github.com/smithy-lang/smithy-rs/issues/4308) so there is some context around what bug this is covering.

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGeneratorTest.kt
```diff
@@ -342,4 +344,62 @@ class JsonSerializerGeneratorTest {
         }
         project.compileAndTest()
     }
+
+    @Test
+    fun `union with unit struct demonstrates serialization bug`() {
+        val model =
```
landonxjames: Since you are using the same model for both of these tests I would split it out to live outside of the tests so it isn't copied in multiple places. 

But maybe the more salient point is I don't know that we need both of these tests? They look exactly the same to me besides the test names?

tools/ci-build/sdk-lockfiles/Cargo.lock
```diff
@@ -933,7 +933,7 @@ dependencies = [
 
 [[package]]
 name = "sdk-lockfiles"
-version = "0.1.3"
```
landonxjames, line 936: Should probably rebase your branch from main since https://github.com/smithy-lang/smithy-rs/pull/4315 just merged so there aren't all these lockfile diffs

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGeneratorTest.kt
```diff
@@ -342,4 +344,62 @@ class JsonSerializerGeneratorTest {
...
+
+        // Render the Union
+        project.moduleFor(model.lookup<UnionShape>("test#TestUnion")) {
+            UnionGenerator(model, codegenContext.symbolProvider, this, model.lookup("test#TestUnion")).render()
+        }
+        project.compileAndTest()
+    }
+
+    @Test
+    fun `union with unit struct demonstrates cbor serialization bug`() {
```
landonxjames: Ahh I see now, the CBOR test shouldn't be in this file, but there doesn't seem to be a `CborSerializerGeneratorTest` file. @ysaito1001 might know why that is missing?

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/AwsQuerySerializerGeneratorTest.kt
ysaito1001, line 1: We want to add a test in `QuerySerializerGeneratorTest` instead of `AwsQuerySerializerGeneratorTest`

ysaito1001 (commented):
We need to make sure that the test crates generated by codegen tests issue the warning
```
unused variable: `inner`
```
without the fix (and shouldn't issue the warning WITH the fix).

When I ran one of the codegen tests, I didn't see the `protocol_serde` module rendered where the unused warning in question should be issued ([occurrence in s3control](https://github.com/smithy-lang/smithy-rs/actions/runs/17772523288/job/50513747569#step:4:1414) for reference)

Could we double-check?

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGeneratorTest.kt
ysaito1001: Didn't realize till now. `CborSerializerGenerator` was originally bootstrapped by the server team's effort. After some time, client side support was added, and it's been assumed the generator was functional. We can certainly add a new test file `CborSerializerGeneratorTest`

ysaito1001:
Good attempt at the latest commit 4cba820!  We need to address the following:
- `union with unit struct doesn't cause unused variable warning` in `JsonSerializerGeneratorTest` now renders the `protocol_serde` module, which is good, but it does not seem to issue unused variable warning (note that code fixing for `JsonSerializerGenerator` isn't commited to the PR yet, so the test should issue the unused warning if written as expected)? If we execute the test and inspect the generated test crate, the variable `inner` is used in blocks that immediately follow:
```
        crate::test_model::TestUnion::UnitMember(inner) => {
            #[allow(unused_mut)]
            let mut object_1 = object_2.key("unitMember").start_object();
            crate::protocol_serde::shape_unit::ser_unit(&mut object_1, inner)?;
            object_1.finish();
        }
        crate::test_model::TestUnion::DataMember(inner) => {
            object_2.key("dataMember").string(inner.as_str());
        }
```
So we probably need to iterate on the test model so that `inner` won't be used.
- Now that `union with unit struct doesn't cause unused variable warning` in `JsonSerializerGeneratorTest` renders the `protocol_serde` module, the test [is complained by the compiler](https://github.com/smithy-lang/smithy-rs/actions/runs/17992241871/job/51185352635?pr=4316#step:4:648) because the rendered Rust constructs are unused in the test. We can refer to another test `generates valid serializers`, and see how `project.lib {...}` produces Rust code that does use serializer's code.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGenerator.kt
```diff
@@ -555,14 +555,28 @@ class JsonSerializerGenerator(
                     }
                     rustBlock("match input") {
                         for (member in context.shape.members()) {
+                            val memberShape = model.expectShape(member.target)
                             val variantName =
                                 if (member.isTargetUnit()) {
                                     "${symbolProvider.toMemberName(member)}"
+                                } else if (memberShape.isStructureShape &&
+                                    memberShape.asStructureShape().get().allMembers.isEmpty()
```
rcoh: PR description implies this is a test-only PR. Is this intentional?

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGeneratorTest.kt
```diff
@@ -342,4 +344,57 @@ class JsonSerializerGeneratorTest {
...
+                let input = crate::test_input::TestOpInput::builder()
+                    .union(TestUnion::UnitMember(Unit::builder().build()))
+                    .build()
+                    .unwrap();
+
+                // This will generate and use the serialization code that should not have unused variable warnings
+                let _serialized = ${format(operationGenerator!!)};
+                let _result = _serialized(&input);
+
+                // Test that the code compiles and runs - this validates our fix works
```
rcoh: Delete ai comment

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/CborSerializerGenerator.kt
```diff
@@ -547,14 +547,23 @@ class CborSerializerGenerator(
...
+                                } else if (isEmptyStruct) {
+                                    // Empty structures don't use the inner variable
+                                    "$memberName(_inner)"
                                 } else {
-                                    "${symbolProvider.toMemberName(member)}(inner)"
+                                    "$memberName(inner)"
                                 }
                             rustBlock("#T::$variantName =>", unionSymbol) {
-                                serializeMember(MemberContext.unionMember("inner", member))
+                                val innerRef = if (isEmptyStruct) "_inner" else "inner"
```
landonxjames: Is this necessary? I think that for the `isEmptyStruct` case it doesn't really matter what we pass to `serializeMember` because there are no members to serialize. 

This comment applies to all of the generators since it looks like there is similar code in all of the updates.

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGeneratorTest.kt
```diff
@@ -342,4 +344,59 @@ class JsonSerializerGeneratorTest {
         }
         project.compileAndTest()
     }
+
+    @Test
+    fun `union with unit struct doesn't cause unused variable warning`() {
+        // Regression test for https://github.com/smithy-lang/smithy-rs/issues/4308
+        // This test ensures that union serialization with unit structs compiles without unused variable warnings.
+        val model = RecursiveShapeBoxer().transform(OperationNormalizer.transform(QuerySerializerGeneratorTest.unionWithUnitStructModel))
```
landonxjames: Two things here:
* I don't think some of these transforms are needed, at least the `RecursiveShapeBoxer`, since the model being tested doesn't have any recursive shapes.
* Feels kind of weird importing the model from the `QuerySerializerGeneratorTest`. I would either split that out into a shared `SerializerGeneratorTestUtils` class or just copy it into each test, either is fine.

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/JsonSerializerGeneratorTest.kt
```diff
@@ -342,4 +344,59 @@ class JsonSerializerGeneratorTest {
...
+
+                // Verify the serialization worked
+                let output = std::str::from_utf8(serialized.bytes().unwrap()).unwrap();
+                assert!(output.contains("none"));
+                """,
+            )
+        }
+
+        // The test passes if the generated code compiles without unused variable warnings
+        project.compileAndTest()
```
landonxjames, line 400: I'm not entirely sure the comment above is true. In the lines below we disable failing on warnings for `compileAndTest`:
https://github.com/smithy-lang/smithy-rs/blob/052538d6a60f933841bbf8a71b298ffcc1fec4cb/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/Rust.kt#L387-L390

If the generated code here is relatively clean we could probably add an option to `compileAndTest` that allows failing on warnings, but if there are a bunch of warnings beyond the one we are trying to fix we might need a different approach.

landonxjames (dismissed):
Couple more comments, looks like it is coming along. Lets get together early next week to iron out some things

codegen-core/src/test/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/AwsQuerySerializerGeneratorTest.kt
vcjana: Changed the function name to `union with empty struct doesn't cause unused variable warning` and added a comment to that links to the issue

vcjana:
Closing this PR based on team feedback. The fix was applied too broadly - 
JsonSerializerGenerator doesn't need changes since it always uses the inner variable. Will create a new PR that targets only QuerySerializerGenerator.

PR #4322, opened 2025-10-01
Fix bug where servers rejected event stream content type

rcoh:
## Motivation and Context
There is a bug in the generated server SDK where for operations with an output event stream, it rejects `application/vnd.amazon.eventstream`, instead expecting `application/cbor`.

This is incorrect. However, we need to maintain backwards compatibility with existing clients. This adds a new fallback option that protocols can set to handle this for RpcV2Cbor.

## Description
- Small refactoring to extract a mime-type generator
- Allow accepting multiple headers as required.

## Testing
I briefly went down the path of trying to get this to work as a protocol test but ultimately it became too much of a yak shave.

I added a unit test that should be sufficient.

## Checklist

----

### DIFF
--- a/.changelog/1759353705.md
+++ b/.changelog/1759353705.md
@@ -1,0 +1,14 @@
+---
+applies_to:
+  - server
+authors:
+  - rcoh
+references: [ ]
+breaking: false
+new_feature: false
+bug_fix: true
+---
+
+Fix bug where servers rejected `application/vnd.amazon.evenstream` ACCEPT header for RPCv2Cbor
+
+This change allows this header while also allowing `application/cbor` for backwards compatibility.


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/protocols/ServerHttpBoundProtocolGenerator.kt
```diff
@@ -228,43 +229,34 @@ class ServerHttpBoundProtocolTraitImplGenerator(
...
-                                });
-                                """
-                        }
-                    rustTemplate(init, *codegenScope)
+                    val legacyContentType = httpBindingResolver.legacyBackwardsCompatContentType(operationShape)
+                    if (legacyContentType != null) {
+                        // For operations with legacy backwards compatibility, accept both content types
+                        rustTemplate(
+                            """
+                            if !#{SmithyHttpServer}::protocol::accept_header_classifier(request.headers(), &#{ContentType}) &&
```
drganjoo, line 241: I wonder what happens if someone sends a comma separated `Accept`?

```
Accept: application/cbor, application/json
```

drganjoo (approved):

PR #4330, opened 2025-10-03
feat(rust): Add tags to AssumeRoleProviderBuilder

c-thiel:
## Motivation and Context
See https://github.com/awslabs/aws-sdk-rust/issues/1366

## Description
See https://github.com/awslabs/aws-sdk-rust/issues/1366

## Checklist

----


c-thiel:
Not really sure about the changelog. Please change it as desired.

aws/rust-runtime/aws-config/src/sts/assume_role.rs
```diff
@@ -198,6 +200,15 @@ impl AssumeRoleProviderBuilder {
         self
     }
 
+    /// Set the session tags
+    /// 
+    /// A list of session tags that you want to pass. Each session tag consists of a key name and an associated value.
+    /// For more information, see `[Tag]`.
+    pub fn tags(mut self, tags: Vec<Tag>) -> Self {
```
ysaito1001: nit: you could use `tags: impl Into<Vec<Tag>>` as a parameter to allow a bit more flexibility.

Thank you for contributing!

c-thiel: Gladly, let me fix that. I was using `Vec` only because other endpoints, such as `policy_arns` are equally unflexible.

rcoh (approved):

rcoh:
thanks!

ysaito1001:
Just realized one thing. [This CI check](https://github.com/smithy-lang/smithy-rs/actions/runs/18226703996/job/51904390504#step:4:1862) complains that the `tag` method now exposes a type that previously wasn't exposed. However, the method cannot take the `Tag` type from `aws_sdk_sts` directly either because `aws-config` should only depend on so-called runtime crates publicly (those listed in its allowed [external-types](https://github.com/c-thiel/smithy-rs/blob/f166e6cbe5bc1aad79c8a1c0953f6725b654b981/aws/rust-runtime/aws-config/external-types.toml#L6-L38)) but not on generated SDK crates like `aws-sdk-sts` (ok to depend on them privately as implementation details).

One way to work around this is for the `tag` method to take plain `String`s for key & value for tags, like
```
pub fn tags<K, V>(mut self, tags: impl IntoIterator<Item = (K, V)>) -> Self 
where
    K: Into<String>,
    V: Into<String>,
{...}
```
This is why a sibling method `policy_arns` takes a vec of plain `String`s instead of a vec of `PolicyDescriptorType`s.

Alternatively, we could define own tag type somewhere in the `aws-config` crate, and let the `tags` method take a vec of that tag type instead.

c-thiel:
> Just realized one thing. [This CI check](https://github.com/smithy-lang/smithy-rs/actions/runs/18226703996/job/51904390504#step:4:1862) complains that the `tag` method now exposes a type that previously wasn't exposed. However, the method cannot take the `Tag` type from `aws_sdk_sts` directly either because `aws-config` should only depend on so-called runtime crates publicly (those listed in its allowed [external-types](https://github.com/c-thiel/smithy-rs/blob/f166e6cbe5bc1aad79c8a1c0953f6725b654b981/aws/rust-runtime/aws-config/external-types.toml#L6-L38)) but not on generated SDK crates like `aws-sdk-sts` (ok to depend on them privately as implementation details).
> 
> One way to work around this is for the `tag` method to take plain `String`s for key & value for tags, like
> 
> ```
> pub fn tags<K, V>(mut self, tags: impl IntoIterator<Item = (K, V)>) -> Self 
> where
>     K: Into<String>,
>     V: Into<String>,
> {...}
> ```
> 
> This is why a sibling method `policy_arns` takes a vec of plain `String`s instead of a vec of `PolicyDescriptorType`s.
> 
> Alternatively, we could define own tag type somewhere in the `aws-config` crate, and let the `tags` method take a vec of that tag type instead.

@ysaito1001 

I made the change, but we have a small problem here: Due to the absolutely awful builder of `Tag`, there is not typesafe constructor for this dead-simple type.  The build method only fails if either key or value is None, which in our implementation cannot be the case. Hence I currently use unwrap, even though this is not nice.
The alternative would be to expose the Error type, this is however a type the sts crate again.

aws/rust-runtime/aws-config/src/sts/assume_role.rs
```diff
@@ -198,6 +200,24 @@ impl AssumeRoleProviderBuilder {
...
+    /// For more information, see `[Tag]`.
+    pub fn tags<K, V>(mut self, tags: impl IntoIterator<Item = (K, V)>) -> Self 
+    where
+        K: Into<String>,
+        V: Into<String>{
+        self.tags = Some(
+            tags.into_iter()
+                // Unwrap won't fail as both key and value are specified.
+                // Currently Tag does not have an infallible build method.
+                .map(|(k, v)| Tag::builder().key(k).value(v).build().unwrap())
```
rcoh:
```suggestion
                .map(|(k, v)| Tag::builder().key(k).value(v).build().expect("this is unreachable: both k and v are set"))
```

rcoh (approved):

Issue #4325, opened 2025-10-02
smithy-rs servers to not support `initial-request` event stream messages


rcoh:
When communicating with other clients (e.g. Java) an initial request is _always_ sent. This is in-spec.

When the server receives an initial request, it errors out. It needs to:

1. Determine if an initial request is actually required for this operation
2. When it is, wait for an initial request to be sent prior to invoking the handler. It must fill in relevant fields of input based in the initial message.
3. Send an initial response (potentially an empty one if no modeled fields exist)

The initial request is:

```
:content-type: application/cbor, :message-type: event, :event-type: initial-request
```

rcoh:
The minimal change to support empty initial requests is this:
```rust
Ok({  
    #[allow(unused_mut)]  
    let mut input = crate::input::streaming_operation_input::Builder::default();  
    #[allow(unused_variables)]  
    let ::aws_smithy_runtime_api::http::RequestParts {  
        uri, headers, body, ..  
    } = ::aws_smithy_runtime_api::http::Request::try_from(request)?.into_parts();  
    let mut receiver =  
        crate::protocol_serde::shape_streaming_operation_input::de_events_payload(  
            &mut body.into().into_inner(),  
        )?;  
   // ADD THIS ONE LINE HERE.
    if let Some(initial_event) = receiver.try_recv_initial(aws_smithy_http::event_stream::InitialMessageType::Request).await.map_err(|ev_error|aws_smithy_http_server::protocol::rpc_v2_cbor::rejection::RequestRejection::ConstraintViolation(format!("{ev_error}").into()))? {  
        // parse intial event as the "initial message frame"  
    }  
    input = input.set_events(receiver);  
    input.build()?  
})
```

PR #4332, opened 2025-10-04
Add ability to write `integration-tests` for servers

rcoh:
## Motivation and Context

This `integration-tests` directory makes it possible to write
Rust code that you can run directly against the generated servers
from the protocol test models. This makes it easier to test things
like event streams behavior and other specifics that are hard to
test with protocol tests alone.

## Description

I added a test for #4325 (showing the broken behavior) that we can use when we fix this issue. This will also make it easy to test a number of other open issues that have been found.

To make this easy I also added a `EventStreamSender::once` method for convenience. The `ManualEventStream` sender could be upstreamed at a later date so it is more broadly usable but I didn't want to pin down a public API for it right now.

----


codegen-server-test/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ plugins {
     java
     alias(libs.plugins.smithy.gradle.base)
     alias(libs.plugins.smithy.gradle.jar)
+    id("com.dorongold.task-tree") version "2.1.1"
```
aajtodd: should be consistent with how we define dependencies by placing them in the plugins section of libs.versions.toml

aajtodd (approved):

codegen-server-test/build.gradle.kts
rcoh: oh derp, let me back this change out actually. I didn't mean to add this in the final version (unless you want to keep it in)

Issue #1700, opened 2022-09-02
Handle Smithy 2.0 integer enum shapes


DavidOgunsAWS:
[IntEnum](https://awslabs.github.io/smithy/2.0/spec/simple-types.html#intenum) shapes have been added to Smithy 2.0. In order to support Smithy 2.0 models, changes are need in client and server side code generation.

* [EnumGenerator](https://github.com/awslabs/smithy-rs/blob/main/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/generators/EnumGenerator.kt) invoked in [CodegenVisitor](https://github.com/awslabs/smithy-rs/blob/main/codegen/src/main/kotlin/software/amazon/smithy/rust/codegen/smithy/CodegenVisitor.kt#L194) needs to be extended substantially or a new class for IntEnums.
 * Also [ServerEnumGenerator](https://github.com/awslabs/smithy-rs/blob/main/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerEnumGenerator.kt) invoked in [ServerCodegenVisitor](https://github.com/awslabs/smithy-rs/blob/main/codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/ServerCodegenVisitor.kt#L190) needs similar support for integers

Note: looks like i32 should be the representation https://awslabs.github.io/smithy/2.0/spec/simple-types.html#integer

jdisanti:
Refactoring the `EnumGenerator` to work off of enum shapes instead of traits is being tracked in https://github.com/awslabs/smithy-rs/issues/1671

PR #4337, opened 2025-10-08
Fixing bug with enums as eventHeaders

landonxjames:
## Motivation and Context

Fixing a bug that was discovered with smithy `enum` and `intEnum` shapes that are marked as `@eventHeader`. This came up in an internal investigation around a failed preview build.

## Description

Added handling of `EnumShape` in `EventStreamUnmarshallerGenerator` and `EventStreamMarshallerGenerator`.

## Testing

Added new `enum` and `intEnum` shapes to the evenstream test model. They are only included in client tests since the server throws a validation error on enums in event streams (see https://github.com/smithy-lang/smithy/issues/1388 for more info).

Also tested against the model from the failed preview build and a `cargo test --all-features` was successful. 

----


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestModels.kt
```diff
@@ -129,6 +135,26 @@ object EventStreamTestModels {
...
+                model =
+                    fillInBaseModel(
+                        this.protocolShapeId,
+                        extraEventHeaderMembers = "@eventHeader enum: TheEnum,\n@eventHeader intEnum: FaceCard,",
+                        extraShapes = """    enum TheEnum {
+                            FOO
+                            BAR
+                            }
+
+                            intEnum FaceCard {
```
aajtodd, line 151: Do we support intEnum yet? https://github.com/smithy-lang/smithy-rs/issues/1700

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt
```diff
@@ -261,6 +262,7 @@ class EventStreamUnmarshallerGenerator(
                     is IntegerShape -> rustTemplate("#{expect_fns}::expect_int32(header)?", *codegenScope)
                     is LongShape -> rustTemplate("#{expect_fns}::expect_int64(header)?", *codegenScope)
                     is BlobShape -> rustTemplate("#{expect_fns}::expect_byte_array(header)?", *codegenScope)
+                    is EnumShape -> rustTemplate("#{expect_fns}::expect_string(header)?.as_str().into()", *codegenScope)
```
rcoh, line 265: does this handle `intEnum`?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/testutil/EventStreamTestModels.kt
landonxjames: No, but I still think this should stay in the test so that when we do support them we get an obvious test failure if they don't work as `@eventHeader`s instead of a false positive.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/parse/EventStreamUnmarshallerGenerator.kt
landonxjames: I don't think any codegen actually supports `intEnum` (see https://github.com/smithy-lang/smithy-rs/issues/1700) so it isn't handled here. I did add it to the test model though so when we do support it we will get an obvious error in these tests if it doesn't work as an `@eventHeader`.

Issue #4338, opened 2025-10-08
setter method (`set_meta`) conflicts with built in `set_meta` for error metadata in generated client


nated0g:
When using `rust-client-codegen`, if you have a shape like:

```
structure Foo {
   meta: Bar
}
```

The struct member `meta: Option<Bar>` would conflict with the built in `meta: ::aws_smithy_types::error::ErrorMetadata`, and so it is correctly substituted as `meta_value: Option<Bar>`.

(reference: https://github.com/smithy-lang/smithy-rs/blob/a87ffdf9ec72e24044c9681201a803ff0dea7b3c/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/ClientReservedWords.kt#L23)

*however*, the generated setter method on the builder does **not** get renamed, and so duplicate `set_meta` methods are added to the `impl` block.

(reference: https://github.com/smithy-lang/smithy-rs/blob/a87ffdf9ec72e24044c9681201a803ff0dea7b3c/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt#L138)

This leads to an invalid crate which will fail to compile with the following error:

```
error[E0592]: duplicate definitions with name `set_meta`
```

haydenbaker:
[fix: use symbol-provider for accessors in member-builder #4341](https://github.com/smithy-lang/smithy-rs/pull/4341) is a naive fix - ideally, when non-member "fields" are added to the builder, they would perhaps follow a different pattern. Perhaps the rust team can weigh in.

PR #4341, opened 2025-10-09
fix: use symbol-provider for accessors in member-builder

haydenbaker:
## Motivation and Context
- https://github.com/smithy-lang/smithy-rs/issues/4338
- needs testing
- ideally, wouldn't need to do this, but would require changing the strategy for adding non-members to the builder (ex: meta field for error_metadata)
- naively strips a `r##` prefix from accessors, but we should probably be more clever here

## Checklist

----

### DIFF
--- a/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
+++ b/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
@@ -134,11 +134,14 @@ class OperationBuildError(private val runtimeConfig: RuntimeConfig) {
     }
 }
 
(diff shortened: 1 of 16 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/generators/BuilderGenerator.kt
```diff
@@ -134,11 +134,14 @@ class OperationBuildError(private val runtimeConfig: RuntimeConfig) {
     }
 }
 
-// Setter names will never hit a reserved word and therefore never need escaping.
```
rcoh, line 137: 😅

landonxjames:
Closing in favor of https://github.com/smithy-lang/smithy-rs/pull/4342

PR #4342, opened 2025-10-10
fix: builder accessor method naming conflicts

nated0g:
## Motivation and Context
https://github.com/smithy-lang/smithy-rs/issues/4338

## Description
fixes a couple stray compilation errors in @haydenbaker's PR (https://github.com/smithy-lang/smithy-rs/pull/4341), and adds a test, adds changelog entry

## Testing

Added a test in `FluentClientGeneratorTest.kt` to verify the bug no longer exists

## Checklist

----

### DIFF
--- a/.changelog/fix-builder-accessor-naming.md
+++ b/.changelog/fix-builder-accessor-naming.md
@@ -1,0 +1,9 @@
+---
+applies_to: ["client"]
+authors: ["haydenbaker", "nated0g"]
+references: ["smithy-rs#4338"]
+breaking: false
+new_feature: false
+bug_fix: true
+---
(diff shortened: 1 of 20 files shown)


landonxjames (approved):
Assuming that commit I did for the `ktlint` issues works this all looks good to me. Will require a review from the server team as well.

rcoh (approved):

rcoh:
I think this will probably cause breaking changes because setters that almost certainly exist in the wild for fields like `default` will end up being renamed (see the `Client Test` codegen diff above)

I think we need to make this context aware so that we only rename setters where the setter itself would actually conflict.

nated0g:
> I think this will probably cause breaking changes because setters that almost certainly exist in the wild for fields like `default` will end up being renamed (see the `Client Test` codegen diff above)
> 
> I think we need to make this context aware so that we only rename setters where the setter itself would actually conflict.

alright finally coming back to this because I'm tired of manually tweaking codegen output for a client I've had to regenerate a few times now 😅

@rcoh I think my latest should address your concern

Issue #4345, opened 2025-10-10
servers must send `initial-response` in all cases


rcoh:
To ensure compatibility with all language clients, servers need to send `initial-response` events in all cases over RPC based protocols (e.g. not restJson)

rcoh:
## Minimal Code Change Identified

Successfully identified the minimal change needed in generated code. The failing test now passes with this modification to `ser_streaming_operation_with_initial_data_http_response`:

```rust
// BEFORE: Just send event stream
let adapter = output.events.into_body_stream(marshaller, error_marshaller, signer);
adapter

// AFTER: Prepend initial-response message
use ::futures_util::StreamExt;
let initial_message = {
    let mut headers = Vec::new();
    headers.push(::aws_smithy_types::event_stream::Header::new(
        ":message-type",
        ::aws_smithy_types::event_stream::HeaderValue::String("event".into()),
    ));
    headers.push(::aws_smithy_types::event_stream::Header::new(
        ":event-type", 
        ::aws_smithy_types::event_stream::HeaderValue::String("initial-response".into()),
    ));
    headers.push(::aws_smithy_types::event_stream::Header::new(
        ":content-type",
        ::aws_smithy_types::event_stream::HeaderValue::String("application/cbor".into()),
    ));
    let payload = hyper::body::Bytes::from(vec![0xbf, 0xff]); // Empty CBOR map
    ::aws_smithy_types::event_stream::Message::new_from_parts(headers, payload)
};

let mut buffer = Vec::new();
::aws_smithy_eventstream::frame::write_message_to(&initial_message, &mut buffer).unwrap();
let initial_message_stream = futures_util::stream::iter(vec![Ok(buffer.into())]);

let adapter = output.events.into_body_stream(marshaller, error_marshaller, signer);
initial_message_stream.chain(adapter)
```

## Two Cases to Handle

1. **Empty initial response** (current test case): Send empty CBOR map `{0xbf, 0xff}`
2. **Dataful initial response**: Need to serialize non-eventstream output members using the correct protocol serializer

Next: Implement this pattern in the server codegen (`ServerHttpBoundProtocolPayloadGenerator.kt`)

rcoh:
## ✅ Protocol Test Added

Successfully added a new protocol test to validate initial-response with actual data:

### Changes Made

1. **New Smithy Operation** in `rpcv2Cbor-extras.smithy`:
   ```smithy
   operation StreamingOperationWithInitialResponse {
[...]

PR #4352, opened 2025-10-16
Add support for `initial-response` event streams on servers

rcoh:
## Motivation and Context

[Event Stream](https://smithy.io/2.0/spec/streaming.html#event-streams) operations over RPC protocols (that do not rely on HTTP headers), use `initial-request` and `initial-response` objects to send data fields that are not bound to the event stream itself.

However, there are two major bug in the current implementation:
- #4353: Clients error if an initial-response is sent when one is not expected
- #4345: Servers do not send initial responses (ever)

## Description
This adds two things:
1. Optional (and disabled by default) sending of an initial-response message in all cases (even when the resulting message would be empty). Servers that communicate with `smithy-java` clients MUST enable this setting currently. This is disabled by default because clients in the wild today do not accept an empty `initial-response`.
3. Support for sending non-empty initial response messages

## Testing
- Extensive unit testing

## Checklist

----

### DIFF
--- a/codegen-core/common-test-models/rpcv2Cbor-extras.smithy
+++ b/codegen-core/common-test-models/rpcv2Cbor-extras.smithy
@@ -19,6 +19,7 @@ service RpcV2CborService {
         StreamingOperation
         StreamingOperationWithInitialData
         StreamingOperationWithInitialResponse
+        StreamingOperationWithOptionalData
     ]
 }
 
@@ -65,6 +66,22 @@ structure StreamingOperationWithInitialResponseInput {
 structure StreamingOperationWithInitialResponseOutput {
     @required
     responseData: String
+    events: Events
+}
+
+operation StreamingOperationWithOptionalData {
+    input: StreamingOperationWithOptionalDataInput,
+    output: StreamingOperationWithOptionalDataOutput,
+    errors: [ValidationException]
+}
+
+structure StreamingOperationWithOptionalDataInput {
+    optionalData: String
+    events: Events
+}
+
+structure StreamingOperationWithOptionalDataOutput {
+    optionalResponseData: String
     events: Events
 }
 
--- a/codegen-server-test/integration-tests/eventstreams/tests/structured_eventstream_tests.rs
+++ b/codegen-server-test/integration-tests/eventstreams/tests/structured_eventstream_tests.rs
@@ -28,9 +28,18 @@ struct StreamingOperationWithInitialDataState {
 }
 
 #[derive(Debug, Default, Clone)]
+struct StreamingOperationWithOptionalDataState {
+    optional_data: Option<String>,
+    events: Vec<Events>,
+    #[allow(dead_code)]
+    num_calls: usize,
+}
+
+#[derive(Debug, Default, Clone)]
 struct ServerState {
     streaming_operation: StreamingOperationState,
     streaming_operation_with_initial_data: StreamingOperationWithInitialDataState,
+    streaming_operation_with_optional_data: StreamingOperationWithOptionalDataState,
 }
 
 struct TestServer {
@@ -44,6 +53,7 @@ impl TestServer {
         let handler_state = state.clone();
         let handler_state2 = state.clone();
         let handler_state3 = state.clone();
+        let handler_state4 = state.clone();
 
         let config = RpcV2CborServiceConfig::builder().build();
         let app = RpcV2CborService::builder::<hyper0::Body, _, _, _>(config)
@@ -58,6 +68,10 @@ impl TestServer {
             .streaming_operation_with_initial_response(move |input| {
                 let state = handler_state3.clone();
                 streaming_operation_with_initial_response_handler(input, state)
+            })
+            .streaming_operation_with_optional_data(move |input| {
+                let state = handler_state4.clone();
+                streaming_operation_with_optional_data_handler(input, state)
             })
             .build_unchecked();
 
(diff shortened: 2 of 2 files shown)


codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/EventStreamMarshallerGenerator.kt
```diff
@@ -109,6 +109,25 @@ open class EventStreamMarshallerGenerator(
...
+        return RuntimeType.forInlineFun("initial_response_from_payload", eventStreamSerdeModule) {
+            rustBlockTemplate(
+                """
+                pub(crate) fn initial_response_from_payload(
+                    payload: #{Bytes}
+                ) -> #{Message}
+                """,
+                *codegenScope,
+            ) {
+                rustTemplate("let mut headers = #{Vec}::new();", *codegenScope)
```
landonxjames: Worth a `with_capacity()`?

AGENTS.md
```diff
@@ -113,22 +117,76 @@ val mimeType = RuntimeType.forInlineFun("APPLICATION_JSON", module) {
 
 ⚠️ **Footgun**: Name collisions mean only one implementation gets generated.
 
+## GitHub CLI Integration
```
landonxjames, line 120: Feel like since this section isn't project specific it should live somewhere more generic for the agent rather than in the AGENTS file of each project?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/EventStreamMarshallerGenerator.kt
ysaito1001: If so, `renderInitialRequestGenerator` above (hidden in collapsible section in diff) may benefit, too.

codegen-core/common-test-models/rpcv2Cbor-extras.smithy
```diff
@@ -51,6 +52,22 @@ structure StreamingOperationWithInitialDataOutput {
...
+    input: StreamingOperationWithInitialResponseInput,
+    output: StreamingOperationWithInitialResponseOutput,
+    errors: [ValidationException]
+}
+
+structure StreamingOperationWithInitialResponseInput {
+    events: Events
+}
+
+structure StreamingOperationWithInitialResponseOutput {
```
drganjoo, line 66: Should we add an operation with optional fields to future-proof our tests? Smithy specs state: "Clients and servers MUST NOT fail if an initial-request or initial-response is not received for an initial message that contains only optional members." This would help test our handling of this requirement when we implement client support.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/CborSerializerGenerator.kt
```diff
@@ -267,7 +280,16 @@ class CborSerializerGenerator(
     }
 
     override fun unsetStructure(structure: StructureShape): RuntimeType =
-        UNREACHABLE("Only clients use this method when serializing an `@httpPayload`. No protocol using CBOR supports this trait, so we don't need to implement this")
+        ProtocolFunctions.crossOperationFn("cbor_json_unset_struct_payload") { fnName ->
```
drganjoo: Is this change to `unsetStructure` essential for initial response functionality, or just a nice-to-have improvement? If it's necessary for initial responses, shouldn't we similarly update the `unsetUnion` function which has the same "UNREACHABLE" implementation?

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/EventStreamMarshallerGenerator.kt
```diff
@@ -109,6 +109,25 @@ open class EventStreamMarshallerGenerator(
         }
     }
 
+    fun renderInitialResponseGenerator(contentType: String): RuntimeType {
```
drganjoo, line 112: Isn't this functionality server-specific? If so, shouldn't we move it out of the core module and keep it only in server-related modules?

codegen-server-test/build.gradle.kts
```diff
@@ -56,6 +56,14 @@ val allCodegenTests = "../codegen-core/common-test-models".let { commonModels ->
             "smithy.protocoltests.rpcv2Cbor#RpcV2CborService",
             "rpcv2Cbor_extras",
             imports = listOf("$commonModels/rpcv2Cbor-extras.smithy"),
+            extraConfig = """, "codegen": { "alwaysSendEventStreamInitialResponse": true } """,
+        ),
+        CodegenTest(
+            "smithy.protocoltests.rpcv2Cbor#RpcV2CborService",
+            "rpcv2Cbor_extras_no_initial_response",
+            imports = listOf("$commonModels/rpcv2Cbor-extras.smithy"),
```
drganjoo, line 64: Do we have tests for other protocols? or are we only implementing this for `rpcv2Cbor` for now?

rcoh: its implemented for all protocols but currently only tested with cbor.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/protocols/serialize/EventStreamMarshallerGenerator.kt
rcoh: you _could_ do that for ideological reasons, but all the library functions it needs are in here so its cleaner IMO

codegen-core/common-test-models/rpcv2Cbor-extras.smithy
rcoh: yeah I'll add that to the model. There are a couple of other small improvements to make to the tests here, I'll open a tickets:

1. Make it possible to use a generated client (not just the manually implemented one)
2. Make it possible to actually _generate_ these tests to make it easy to run them against multiple protocols.

PR #4356, opened 2025-10-20
Add support for Unframing SigV4 signed messages for Servers

rcoh:
# SigV4 Event Stream Support for Server SDK

## Problem

Clients wrap event stream messages in SigV4 envelopes with signature headers (`:chunk-signature`, `:date`), but servers couldn't parse these signed messages because they expected the raw event shape, not the envelope.

## Solution

Added server-side SigV4 event stream unsigning support that automatically extracts inner messages from signed envelopes while maintaining compatibility with unsigned messages.

### Type System Changes

Event stream types are wrapped to handle both signed and unsigned messages:

- `Receiver<Events, Error>` → `Receiver<SignedEvent<Events>, SignedEventError<Error>>`
- `SignedEvent<T>` provides access to both the inner message and signature information
- `SignedEventError<E>` wraps both extraction errors and underlying event errors

### Runtime Components

**SigV4Unmarshaller**: Wraps the base event stream unmarshaller to handle SigV4 extraction:

```rust
impl<T: UnmarshallMessage> UnmarshallMessage for SigV4Unmarshaller<T> {
    type Output = SignedEvent<T::Output>;
    type Error = SignedEventError<T::Error>;
    
    fn unmarshall(&self, message: &Message) -> Result<...> {
        match extract_signed_message(message) {
            Ok(Signed { message: inner, signature }) => {
                // Process inner message with base unmarshaller
                self.inner.unmarshall(&inner).map(|event| SignedEvent {
                    message: event,
                    signature: Some(signature),
                })
            }
            Ok(Unsigned) => {
                // Process unsigned message directly  
                self.inner.unmarshall(message).map(|event| SignedEvent {
                    message: event,
                    signature: None,
                })
            }
            Err(err) => Ok(SignedEventError::InvalidSignedEvent(err))
        }
    }
}
```

### Code Generation Integration

**SigV4EventStreamDecorator**: 
- Detects services with `@sigv4` trait and event streams
- Wraps event stream types using `SigV4EventStreamSymbolProvider`
- Generates support structures (`SignedEvent`, `SigV4Unmarshaller`, etc.)
- Injects unmarshaller wrapping via HTTP binding customizations

**HTTP Binding Customization**:
- Added `BeforeCreatingEventStreamReceiver` section to `HttpBindingGenerator`
- Allows decorators to wrap the unmarshaller before `Receiver` creation
- Generates: `let unmarshaller = SigV4Unmarshaller::new(unmarshaller);`

### Usage

Server handlers receive `SignedEvent<T>` and extract the inner message:

```rust
async fn streaming_operation_handler(input: StreamingOperationInput) -> Result<...> {
    let event = input.events.recv().await?;
    if let Some(signed_event) = event {
        let actual_event = &signed_event.message;  // Extract inner message
        let signature_info = &signed_event.signature;  // Access signature if present
        // Process actual_event...
    }
}
```

## Testing

- Added `test_sigv4_signed_event_stream` that sends SigV4-wrapped events
- Verifies both signed and unsigned messages work correctly
- All existing event stream tests continue to pass

## Architecture Benefits

- **Backward Compatible**: Unsigned messages work unchanged
- **Type Safe**: Compile-time guarantees about message structure  
- **Extensible**: Pattern can be applied to other authentication schemes
- **Minimal Impact**: Only affects services with `@sigv4` trait and event streams

## Checklist

----

### DIFF
--- a/examples/pokemon-service/tests/common/mod.rs
+++ b/examples/pokemon-service/tests/common/mod.rs
@@ -3,9 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+use assert_cmd::cargo_bin;
 use std::{process::Command, time::Duration};
-
-use assert_cmd::prelude::*;
 use tokio::time::sleep;
 
 use pokemon_service::{DEFAULT_ADDRESS, DEFAULT_PORT};
@@ -13,8 +12,7 @@ use pokemon_service_client::{Client, Config};
 use pokemon_service_common::ChildDrop;
 
 pub async fn run_server() -> ChildDrop {
-    let crate_name = std::env::var("CARGO_PKG_NAME").unwrap();
-    let child = Command::cargo_bin(crate_name).unwrap().spawn().unwrap();
+    let child = Command::new(cargo_bin!()).spawn().unwrap();
 
     sleep(Duration::from_millis(500)).await;
 


codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
```diff
@@ -0,0 +1,313 @@
...
+                pub fn extract_signed_message(message: &#{Message}) -> #{Result}<MaybeSignedMessage, #{ExtractionError}> {
+                    // Check if message has chunk signature
+                    let mut chunk_signature = None;
+                    let mut timestamp = None;
+
+                    for header in message.headers() {
+                        match header.name().as_str() {
+                            ":chunk-signature" => {
+                                if let #{HeaderValue}::ByteArray(bytes) = header.value() {
+                                    chunk_signature = Some(bytes.as_ref().to_vec());
```
jasgin, line 260: I don't have a ton of context, do we need any sort of validation as part of this?

rcoh: its up to the service author at the moment to validate the signature

codegen-core/common-test-models/rpcv2Cbor-extras.smithy
```diff
@@ -6,8 +6,10 @@ use smithy.framework#ValidationException
 use smithy.protocols#rpcv2Cbor
 use smithy.test#httpResponseTests
 use smithy.test#httpMalformedRequestTests
+use aws.auth#sigv4
 
 @rpcv2Cbor
+@sigv4(name: "rpcv2-cbor")
```
drganjoo, line 12: Is there an existing test service without @sigv4 to verify the code generator works for unsigned services? If not, perhaps instead of adding this annotation directly to the .smithy file, we could add it as part of a separate test to ensure both signed and unsigned services are covered?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamDecorator.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    override fun httpCustomizations(
+        symbolProvider: RustSymbolProvider,
+        protocol: ShapeId,
+    ): List<HttpBindingCustomization> {
+        return listOf(SigV4EventStreamCustomization(symbolProvider))
+    }
+
+    override fun symbolProvider(base: RustSymbolProvider): RustSymbolProvider {
+        // We need access to the service shape to check for SigV4 trait, but the base interface doesn't provide it.
```
drganjoo, line 47: Nit: We can use the function `transformModel` to detect whether the service uses sigV4 authentication or not.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamDecorator.kt
```diff
@@ -0,0 +1,134 @@
...
+        // For now, we'll wrap all event streams and let the runtime code handle the detection.
+        return SigV4EventStreamSymbolProvider(base)
+    }
+}
+
+internal fun RustSymbolProvider.usesSigAuth(): Boolean =
+    ServiceIndex.of(model).getAuthSchemes(moduleProviderContext.serviceShape!!).containsKey(SigV4Trait.ID)
```
drganjoo, line 54: Do we need to consider `SigV4ATrait` as well?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
```diff
@@ -0,0 +1,313 @@
...
+    private fun signatureInfo(): RuntimeType =
+        RuntimeType.forInlineFun("SignatureInfo", supportModule) {
+            rustTemplate(
+                """
+                /// Information extracted from a signed event stream message
+                ##[non_exhaustive]
+                ##[derive(Debug, Clone)]
+                pub struct SignatureInfo {
+                    /// The chunk signature bytes from the `:chunk-signature` header
+                    pub chunk_signature: Vec<u8>,
```
drganjoo, line 80: `#{Vec}` instead of `Vec`

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    private fun signatureInfo(): RuntimeType =
+        RuntimeType.forInlineFun("SignatureInfo", supportModule) {
+            rustTemplate(
+                """
+                /// Information extracted from a signed event stream message
+                ##[non_exhaustive]
+                ##[derive(Debug, Clone)]
+                pub struct SignatureInfo {
```
drganjoo, line 78: I assume the handler code is responsible for determining whether the signature is sigv4 or sigv4a based by inspecting the initial request headers?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
```diff
@@ -0,0 +1,313 @@
...
+                                match self.inner.unmarshall(&inner_message) {
+                                    Ok(unmarshalled) => match unmarshalled {
+                                        #{UnmarshalledMessage}::Event(event) => {
+                                            Ok(#{UnmarshalledMessage}::Event(#{SignedEvent} {
+                                                message: event,
+                                                signature: Some(signature),
+                                            }))
+                                        }
+                                        #{UnmarshalledMessage}::Error(err) => {
+                                            Ok(#{UnmarshalledMessage}::Error(#{SignedEventError}::Event(err)))
```
drganjoo, line 189: Should `SignedEventError::Event` include the signature when the error message itself was signed? Currently the signature is being extracted but then discarded for error messages.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
drganjoo, line 260: Are there limits on event stream message/chunk sizes to prevent memory exhaustion attacks? or is this the service's responsibility via Hyper configuration? Are individual header values size-limited? or should we enforce some check here that the value is maximum 256 bytes long before we `.to_vec()` it?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
```diff
@@ -0,0 +1,313 @@
...
+                                        }
+                                    },
+                                    Err(err) => Err(err),
+                                }
+                            }
+                            Ok(MaybeSignedMessage::Unsigned) => {
```
drganjoo, line 195: If a service has the `@sigv4` trait applied to it:

1. Can a client send unsigned messages?
2. Does the service handler have to raise an error if it doesn't want to deal with unsigned messages?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamDecorator.kt
rcoh: Considered, decided it was too complex

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamDecorator.kt
rcoh: yeah, good catch

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
rcoh: yeah (or from the model). We can add a metadata field here in the future

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
rcoh: 1. If it isn't configured w/ SigV4, yeah. This is mostly for backwards compatibility.
2. Yeah it is up to the service what it wants to do with an unsigned message

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/SigV4EventStreamSupportStructures.kt
rcoh: the message has been long-parsed before this point. `Message` isn't a streaming object.

AGENTS.md
```diff
@@ -48,6 +48,26 @@ operation MyOperation {
 - **`codegen-core/common-test-models/constraints.smithy`** - Constraint validation tests with restJson1
 - **`codegen-client-test/model/main.smithy`** - awsJson1_1 protocol tests
 
+### httpQueryParams Bug Investigation
```
landonxjames, line 51: Don't think you meant to leave info for a specific bug in the AGENTS file?

examples/pokemon-service-tls/tests/common/mod.rs
```diff
@@ -65,6 +63,8 @@ fn native_tls_connector() -> NativeTlsConnector {
     let tls_connector = hyper_tls::native_tls::TlsConnector::builder()
         .min_protocol_version(Some(hyper_tls::native_tls::Protocol::Tlsv12))
         .add_root_certificate(cert)
+        .danger_accept_invalid_certs(true)
+        .danger_accept_invalid_hostnames(true)
```
rcoh:
```suggestion
```

PR #4357, opened 2025-10-20
include the protocol name in crate metadata

arielb1:
include it as `package.metadata.smithy.protocol`

this allows to easily tell what protocol was used for a given crate

example protocol name: `aws.protocols#restJson1`

## Motivation and Context

Currently, when seeing a Smithy-generated crate for a customer, it can be annoying to find out which protocol that crate was generated for. Include the metadata in the Cargo.toml to make that easier.

cc @drganjoo you wanted this originally.

## Description

Include the protocol name in `package.metadata.smithy.protocol` in `Cargo.toml`

## Testing

Added the check to existing unit tests

Not sure whether this is the best place to put them (asking @drganjoo)

## Checklist

----

### DIFF
--- a/.changelog/1760625617.md
+++ b/.changelog/1760625617.md
@@ -1,0 +1,15 @@
+---
+applies_to:
+- client
+- aws-sdk-rust
+authors:
+- rcoh
+references: ["smithy-rs#4352", "smithy-rs#4353"]
+breaking: false
+new_feature: false
+bug_fix: true
+---
+Update clients to allow `initial-response` events to be accepted on event streams, even when know modeled initial response exists.
+
+This is required for spec compliance, backwards compatibility, and compatibility with non-smithy-rs based servers that
+MAY unconditionally send `initial-response` messages.
--- a/.changelog/1760625769.md
+++ b/.changelog/1760625769.md
@@ -1,0 +1,23 @@
+---
+applies_to:
+- server
+authors:
+- rcoh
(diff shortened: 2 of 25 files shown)


rcoh:
I think this breaks compilation for the `FuzzTargetGenerator`, fyi.

I think I would prefer making an easy structured way to emit smithy metadata via customization. It will set us up well for the future to make it easy to emit more metadata here—I'm worried if we go this route we'll just add more and more parameters to this.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/CodegenDelegator.kt
```diff
@@ -138,6 +138,7 @@ open class RustCrate(
     fun finalize(
         settings: CoreRustSettings,
         model: Model,
+        protocolId: String,
```
rcoh: use `ShapeId` here — that way it's machine readable and unambiguous

arielb1: done

arielb1:
> I'm worried if we go this route we'll just add more and more parameters to this.

Currently we do want (most) crates to have a protocol parameter.

If we end up with a bunch of mostly-required parameters, we should put them into a struct.

arielb1:
Fixed formatting

landonxjames (approved):

PR #4372, opened 2025-10-31
  [HTTP 1.x] Add legacy crates to maintain http@0.x support during migration

drganjoo:
This PR introduces two new legacy crates (`aws-smithy-legacy-http` and `aws-smithy-legacy-http-server`) that maintain support for `http@0.x`
  while the codebase transitions to `http@1.x`.

  ## Background

  As part of the HTTP 1.x migration effort (see branch `feature/http-1.x`), we're adding support for generating server SDKs that depend on `http@1.x` via a new `http-1x` codegen flag. However, we need to maintain backward compatibility with `http@0.x` for existing users.

  ## Why Legacy Crates?

  Initially, we attempted to pin dependencies to the last versions that supported `http@0.x`. However, this approach failed because:
  - The codegen tests produce SDKs with path dependencies on crates
  - These crates transitively depend on the same crates we were trying to pin
  - This created version conflicts that couldn't be resolved with simple pinning

  ## Solution

  Creating dedicated legacy crates solves this problem by:
  1. **Isolating http@0.x dependencies** - Clear separation between http@0.x and http@1.x codepaths
  2. **Easier maintenance** - We can port changes back and forth between versions without checking out old commits
  3. **Simpler development workflow** - Developers don't need to switch branches/commits to maintain http@0.x support
  4. **Clearer intent** - Makes it explicit which code supports which HTTP version

  ## Changes in this PR

  - Added `aws-smithy-legacy-http` crate (forked from `aws-smithy-http` with http@0.x support)
  - Added `aws-smithy-legacy-http-server` crate (forked from `aws-smithy-http-server` with http@0.x support)
  - Updated `rust-runtime/Cargo.toml` to include both legacy crates in the workspace

  ## Note

  This PR is split out from the main `http@1.x` work to make it easier for reviewers to evaluate just the legacy crate additions separately from the  codegen changes. By default, `http@0.x` remains the default - the `http-1x` flag must be explicitly enabled to use `http@1.x`.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,6 +4,7 @@
 /codegen-server-test/                               @smithy-lang/smithy-rs-server
 /codegen-server/                                    @smithy-lang/smithy-rs-server
 /rust-runtime/aws-smithy-http-server/               @smithy-lang/smithy-rs-server
+/rust-runtime/aws-smithy-legacy-http-server/        @smithy-lang/smithy-rs-server
 
 # Python Server
 /codegen-server-test/python/                        @smithy-lang/smithy-rs-server
--- a/tools/ci-build/publisher/src/package.rs
+++ b/tools/ci-build/publisher/src/package.rs
@@ -331,6 +331,7 @@ mod tests {
             package("aws-smithy-http-server", &[]),
             package("aws-smithy-http-server-python", &[]),
             package("aws-smithy-http-server-typescript", &[]),
+            package("aws-smithy-legacy-http-server", &[]),
         ];
         for pkg in server_packages {
             assert_eq!(&["aws-sdk-rust-ci"], pkg.expected_owners());
--- a/tools/ci-build/sdk-lints/src/lint_cargo_toml.rs
+++ b/tools/ci-build/sdk-lints/src/lint_cargo_toml.rs
@@ -60,6 +60,7 @@ const SERVER_CRATES: &[&str] = &[
     "aws-smithy-http-server",
     "aws-smithy-http-server-python",
     "aws-smithy-http-server-typescript",
+    "aws-smithy-legacy-http-server",
 ];
 
 /// Check crate licensing
--- a/tools/ci-build/sdk-lockfiles/src/audit.rs
+++ b/tools/ci-build/sdk-lockfiles/src/audit.rs
@@ -124,6 +124,7 @@ const SERVER_SPECIFIC_RUNTIMES: &[&str] = &[
     "aws-smithy-http-server",
     "aws-smithy-http-server-python",
     "aws-smithy-http-typescript",
+    "aws-smithy-legacy-http-server",
 ];
 
 fn new_dependency_for_aws_sdk(crate_name: &str) -> bool {
--- a/tools/ci-build/smithy-rs-tool-common/src/package.rs
+++ b/tools/ci-build/smithy-rs-tool-common/src/package.rs
@@ -322,6 +322,7 @@ mod tests {
             package("aws-smithy-http-server", &[]),
             package("aws-smithy-http-server-python", &[]),
             package("aws-smithy-http-server-typescript", &[]),
+            package("aws-smithy-legacy-http-server", &[]),
         ];
         for pkg in server_packages {
             assert_eq!(&["aws-sdk-rust-ci"], pkg.expected_owners());


rcoh:
can you target mainline with this instead?

rust-runtime/aws-smithy-legacy-http/README.md
```diff
@@ -0,0 +1,12 @@
+# aws-smithy-http
+
+Core HTTP primitives for service clients generated by [smithy-rs](https://github.com/smithy-lang/smithy-rs) including:
+- HTTP Body implementation
+- Endpoint support
+- HTTP header deserialization
+- Event streams
+- `ByteStream`: _(supported on crate feature `rt-tokio` only)_ a misuse-resistant abstraction for streaming binary data
+
```
rcoh: fix READMEs / crate descriptions

rust-runtime/aws-smithy-legacy-http/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+    "Russell Cohen <rcoh@amazon.com>",
+]
+description = "Smithy HTTP-0x logic for smithy-rs."
+edition = "2021"
+license = "Apache-2.0"
+repository = "https://github.com/smithy-lang/smithy-rs"
+
+[features]
+event-stream = ["aws-smithy-eventstream"]
+rt-tokio = ["aws-smithy-types/rt-tokio"]
```
rcoh, line 16: hmmm...I guess we still need these features? event streams currently use aws-smithy-http based types. I guess the client won't use the legacy crate right? We should have a test case that ensures generated clients never include clients.

rust-runtime/aws-smithy-legacy-http/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+
+[dependencies]
+aws-smithy-eventstream = { path = "../aws-smithy-eventstream", optional = true }
+aws-smithy-runtime-api = { path = "../aws-smithy-runtime-api", features = ["client", "http-02x"] }
+aws-smithy-types = { path = "../aws-smithy-types", features = ["byte-stream-poll-next", "http-body-0-4-x"] }
+bytes = "1.10.0"
+bytes-utils = "0.1"
+# TODO(hyper1) - Complete the breaking changes by updating to http 1.x ecosystem fully in this crate. Also remove hyper 0.14 from dev
+http-02x = { package = "http", version = "0.2.9" }
+http-1x = { package = "http", version = "1" }
```
rcoh, line 26: we don't need a lot of this stuff right? I guess I wonder if we should prune out the parts of aws-smithy-http-legacy we don't need. maybe its OK.

rust-runtime/aws-smithy-legacy-http/Cargo.toml
drganjoo: The generated clients would never use these aws-smithy-legacy-* crates because the Kotlin class that adds these dependencies is defined in the `HttpDependencies` data class within the server module. There are no CargoDependencies defined for the legacy crates in the core module either. The client code has no reference to them unless someone explicitly adds them at a later stage.

I will add a test in the client to ensure they are not added.

rust-runtime/aws-smithy-legacy-http/README.md
drganjoo: Ah yes! fixing these.

rust-runtime/aws-smithy-legacy-http/Cargo.toml
drganjoo: I would like to keep it as it originally existed to ensure we don't break compatibility for users who still want to generate http@0.x-based SDKs. Therefore, these two crates are an exact copy of their original state before we decided to remove http@0.x support from them.

drganjoo:
> can you target mainline with this instead?

It now targets main.

ysaito1001 (approved):
Reviewed `CrateSet.kt` and `aws-smithy-legacy-http` crate. They look good, with the latter essentially the same code with `aws-smithy-http` but updated to clarify it's for `http@0.x/hyper@0.x` support

ysaito1001:
CI has passed, but you may want to consider updating the following places as well
- [sdk-lints](https://github.com/smithy-lang/smithy-rs/blob/3b94437814e710958546961b3b09da1deae5d726/tools/ci-build/sdk-lints/src/lint_cargo_toml.rs#L60)
- [CODEOWNDERS](https://github.com/smithy-lang/smithy-rs/blob/3b94437814e710958546961b3b09da1deae5d726/CODEOWNERS)
- [sdk-lockfiles](https://github.com/smithy-lang/smithy-rs/blob/3b94437814e710958546961b3b09da1deae5d726/tools/ci-build/sdk-lockfiles/src/audit.rs#L123)

drganjoo:
> CI has passed, but you may want to consider updating the following places as well
> 
>     * [sdk-lints](https://github.com/smithy-lang/smithy-rs/blob/3b94437814e710958546961b3b09da1deae5d726/tools/ci-build/sdk-lints/src/lint_cargo_toml.rs#L60)
> 
>     * [CODEOWNDERS](https://github.com/smithy-lang/smithy-rs/blob/3b94437814e710958546961b3b09da1deae5d726/CODEOWNERS)
> 
>     * [sdk-lockfiles](https://github.com/smithy-lang/smithy-rs/blob/3b94437814e710958546961b3b09da1deae5d726/tools/ci-build/sdk-lockfiles/src/audit.rs#L123)

A good catch. Let me add that. Thanks.

PR #4375, opened 2025-11-02
Add BuildError support to aws_json ResponseRejection

drganjoo:
## Summary
- Add missing `From` implementation for `aws_smithy_types::error::operation::BuildError` to `ResponseRejection` enum in aws_json protocol
- Add test to verify AwsJson protocols compile with streaming operations
- Bump aws-smithy-http-server version from 0.65.8 to 0.65.9

## Motivation
Without this implementation, generated code that uses the `?` operator with `BuildError` (e.g., when serializing HTTP payloads for streaming operations) fails to compile with errors like:

```
the trait `From<aws_smithy_types::error::operation::BuildError>` is not 
implemented for `aws_json::rejection::ResponseRejection`
```

This occurs specifically in operations with streaming blob outputs where the payload serialization can return a `BuildError`.

## Changes
- Added `Build` variant to `ResponseRejection` enum in `rust-runtime/aws-smithy-http-server/src/protocol/aws_json/rejection.rs`
- Added `AwsJsonCompilationTest` to verify the fix works for both AwsJson1.0 and AwsJson1.1
- Verified test fails without the fix by commenting out the Build variant
- Bumped crate version to 0.65.9 in `rust-runtime/aws-smithy-http-server/Cargo.toml`
- Added changelog entry in `.changelog/1762045039.md`

This change aligns the aws_json protocol with other protocols like rest_json_1 which already have this variant.

## Test plan
- [x] Added compilation test that verifies AwsJson protocols work with streaming operations
- [x] Verified test fails without the fix
- [x] Verified test passes with the fix
- [x] Ran ktlintFormat to ensure code style compliance

### DIFF
--- a/rust-runtime/aws-smithy-http-server/src/protocol/aws_json/rejection.rs
+++ b/rust-runtime/aws-smithy-http-server/src/protocol/aws_json/rejection.rs
@@ -9,6 +9,8 @@ use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum ResponseRejection {
+    #[error("error building HTTP response: {0}")]
+    Build(#[from] aws_smithy_types::error::operation::BuildError),
     #[error("error serializing JSON-encoded body: {0}")]
     Serialization(#[from] aws_smithy_types::error::operation::SerializationError),
     #[error("error building HTTP response: {0}")]
--- a/.changelog/1762045039.md
+++ b/.changelog/1762045039.md
@@ -1,0 +1,12 @@
+---
+applies_to:
+- server
+authors:
+- drganjoo
+references:
+- smithy-rs#9999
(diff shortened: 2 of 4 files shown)


rust-runtime/aws-smithy-http-server/src/protocol/aws_json/rejection.rs
```diff
@@ -9,6 +9,8 @@ use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum ResponseRejection {
+    #[error("error building HTTP response: {0}")]
```
rcoh, line 12: this is actually a breaking change because this enum is not marked non_exhaustive. I don't know how much breakage will exist in practice.

Issue #3362, opened 2024-01-12
Update smithy-rs servers' dependencies on `http`, `http-body`, `hyper`, and `tower-http`.


david-perez:
Generated server SDKs use:

```
[dependencies.http]
version = "0.2.9"
[dev-dependencies.hyper]
version = "0.14.12"
```

And `aws-smithy-http-server` uses:

```
http = "0.2"
http-body = "0.4"
hyper = { version = "0.14.26", features = ["server", "http1", "http2", "tcp", "stream"] }
tower-http = { version = "0.3", features = ["add-extension", "map-response-body"] }
```

This is a breaking change because types from these crates are exposed in the `tower::Service` the generated SDKs export. `http`, `http-body`, `hyper` crates went 1.x.

Middleware authors using the 0.x versions of the `http` and `http-body` will be affected.

cayman-amzn:
Looks like we might have this covered as apart of https://github.com/awslabs/aws-sdk-rust/issues/977

djc:
Is this a duplicate of #1925?

david-perez:
I think the only overlap of types depending on these crates used by both client and server is that servers ([unfortunately](https://github.com/smithy-lang/smithy-rs/issues/1649)) use `SdkBody`, but `SdkBody` has a [`from_body_1_x`](https://docs.rs/aws-smithy-types/latest/aws_smithy_types/body/struct.SdkBody.html#method.from_body_1_x), so we could give this a go server-side.

It'd be better if the upgrade in both server and client went out in the same smithy-rs release though, to minimize user disruption.

nated0g:
the server sdks being on these older dependency versions is causing quite a bit of pain for us as we attempt to migrate a couple large rust services over from latest axum, which uses the newer http, hyper, etc, to the smithy-rs server sdks.

How much work would you estimate there is to do to get this closed out? You mention that it would be better for client/server to go out in the same release, but presumably the server stuff could be worked on separately and put up on a branch

Happy to submit a PR or something if you can point us in the right direction

drganjoo:
Thanks for reaching out, and apologies for the delayed response. 

We are actively working on migrating the client runtime crates to Hyper 1, which should be released in the next few weeks. Following that, we plan to update the code generator to support Hyper 1 for both client and server-side implementations in a single release.

[...]

PR #4376, opened 2025-11-02
Add http-1x codegen support for server SDK generation

drganjoo:
This commit introduces comprehensive http-1x support to the server code generator,
enabling generation of server SDKs that work with either http@0.2.x or http@1.x
based on the `http-1x` codegen flag.

### New Infrastructure
- **HttpDependencies.kt**: New data class that ensures cohesive HTTP dependency
  selection, preventing accidental mixing of HTTP 0.x and 1.x versions. Provides
  helper methods for accessing HTTP types (Request, Response, HeaderMap, etc.)

- **MultiVersionTestFailure.kt**: Test infrastructure for tracking failures by
  HTTP version, enabling tests to run against both http@0 and http@1

### ServerCodegenConfig Updates
- Add `http1x` boolean field to control HTTP version selection (config key: `http-1x`)
- Defaults to `false` (http@0.2.x)

### ServerCodegenContext Updates  
- Add `isHttp1()` method to check if http-1x is enabled
- Add `httpDependencies()` method to provide version-appropriate HTTP dependencies
  throughout code generation

### ServerRequiredCustomizations
- Update to create and manage HttpDependencies based on http-1x setting
- For http@0: use aws-smithy-legacy-http-server
- For http@1: use aws-smithy-http-server

### Protocol & Test Generator Updates
- **ServerProtocolTestGenerator**: Support running protocol tests against both
  HTTP versions, with configurable test runtime mode (AS_CONFIGURED, BOTH, etc.)
- **ServerHttpBoundProtocolGenerator**: Use HttpDependencies for all HTTP type
  references instead of hardcoded dependencies

### Generator Updates
Updated generators to use HttpDependencies for type resolution:
- ServerRootGenerator: Use httpDeps for dependency selection
- ServerServiceGenerator: Pass HttpDependencies to child generators  
- ServerHttpSensitivityGenerator: Use httpDeps for HTTP types
- ServerOperationGenerator, ServerBuilderGenerator: Use contextual HTTP deps

### Test Infrastructure
- **ServerCodegenIntegrationTest**: Add multi-version test support with
  ability to run tests against both HTTP versions
- **Http1xDependencyTest**: New test validating correct dependency selection
  based on http-1x flag
- Update existing tests to work with multi-version infrastructure

### Core Changes
- HttpBindingGenerator: Accept HttpDependencies parameter for version-appropriate
  type generation
- EventStream generators: Use contextual HTTP dependencies
- SmithyTypesPubUseExtra: Conditional imports based on HTTP version

## Testing
All protocol tests can now run against both HTTP versions to ensure compatibility.
Tests track failures by HTTP version for easier debugging.

Relates to #3362


.changelog/1762113734.md
```diff
@@ -0,0 +1,14 @@
...
+- drganjoo
+references:
+- smithy-rs#3362
+breaking: false
+new_feature: true
+bug_fix: false
+---
+Add http-1x codegen support for server SDK generation
+
+    Enables generation of server SDKs compatible with either http@0.2.x/hyper@0.x or http@1.x/hyper@1.x based on the http-1x codegen flag.
```
rcoh: need more detail in the changelog message--e.g. changelog needs to say what the actual flag is, how to enable it etc.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/RustServerCodegenPlugin.kt
```diff
@@ -78,7 +78,11 @@ class RustServerCodegenPlugin : ServerDecoratableBuildPlugin() {
                 if (includeConstrainedShapeProvider) ConstrainedShapeSymbolProvider(it, serviceShape, constrainedTypes) else it
             }
             // Generate different types for EventStream shapes (e.g. transcribe streaming)
-            .let { EventStreamSymbolProvider(rustSymbolProviderConfig.runtimeConfig, it, CodegenTarget.SERVER) }
+            .let {
+                // Get the correct smithy-http dependency based on http-1x flag
+                val httpDeps = HttpDependencies.create(settings.codegenConfig.http1x, rustSymbolProviderConfig.runtimeConfig)
+                EventStreamSymbolProvider(rustSymbolProviderConfig.runtimeConfig, it, CodegenTarget.SERVER, httpDeps.smithyHttp)
```
rcoh: I would recommend that we actually make `RuntimeConfig` "do the right thing". Or make it so that you _MUST_ pass an http version setting when getting the dependencies that depend on it.

You could incorporate the `http1x` setting into `RuntimeConfig` potentially, then make things like `smithyHttp` http version aware?

I'm worried that the way the code works right now will be somewhat confusing for folks to deal with and create these really confusing compilation errors.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/ServerRequiredCustomizations.kt
```diff
@@ -38,45 +43,118 @@ class ServerRequiredCustomizations : ServerCodegenDecorator {
...
+                    """
+                pub use #{DateTime};
+                pub use #{Format} as DateTimeFormat;
+                """,
+                    "DateTime" to httpDeps.smithyTypesModule().resolve("DateTime"),
+                    "Format" to httpDeps.smithyTypesModule().resolve("date_time::Format"),
+                )
+            }
+            if (hasStreamingOperations(model, codegenContext.serviceShape)) {
+                println("[ServerRequiredCustomizations.pubUseSmithyPrimitivesHttp0x] Adding rt-tokio with http-body-0-4-x")
```
rcoh: did you want to leave these printlns?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerHttpSensitivityGenerator.kt
```diff
@@ -17,25 +17,23 @@ import software.amazon.smithy.model.traits.HttpQueryParamsTrait
...
 import software.amazon.smithy.rust.codegen.core.util.orNull
-import software.amazon.smithy.rust.codegen.server.smithy.ServerCargoDependency
 import java.util.Optional
 
 /** Models the ways status codes can be bound and sensitive. */
-class StatusCodeSensitivity(private val sensitive: Boolean, runtimeConfig: RuntimeConfig) {
+class StatusCodeSensitivity(private val sensitive: Boolean, private val smithyHttpServer: RuntimeType) {
     private val codegenScope =
         arrayOf(
-            "SmithyHttpServer" to ServerCargoDependency.smithyHttpServer(runtimeConfig).toType(),
```
rcoh, line 38: I think we should probably have `ServerCargoDependency.smithyHttpServerAuto` that automatically gives you the right one. I _think_ the http version would actually fit pretty naturally into `RuntimeConfig` as an optional parameter.

rcoh: But if you just want to ship things like it is, I get that. I'm just kinda worried about the long term fallout of having it be non-centralized especially as we eventually try to get rid of the old codegen settings.

.changelog/1762113734.md
```diff
@@ -0,0 +1,29 @@
...
+- server
+authors:
+- drganjoo
+references:
+- smithy-rs#3362
+breaking: false
+new_feature: true
+bug_fix: false
+---
+Add support for generating Server SDKs with http@1 and hyper@1, the latest major versions of Rust's foundational HTTP crates. By default, generated SDKs target http@0 and hyper@0 based crates.
```
rcoh, line 12:
```suggestion
Add support for generating Server SDKs with http@1 and hyper@1, the latest major versions of Rust's foundational HTTP crates. By default, generated SDKs target http@0 and hyper@0 based crates.

This new feature is subject to change.
```

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
```diff
@@ -55,12 +55,30 @@ value class CrateVersionMap(
...
 /**
  * Prefix & crate location for the runtime crates.
  */
 data class RuntimeConfig(
     val cratePrefix: String = "aws",
     val runtimeCrateLocation: RuntimeCrateLocation = RuntimeCrateLocation.path("../"),
+    /**
+     * HTTP version to use for code generation. Defaults to Http1x as that is the default for Clients..
+     */
+    val httpVersion: HttpVersion = HttpVersion.Http1x,
```
rcoh, line 81: as long as there is a test somewhere this is fine, but I am a little bit worried about accidentally enabling this at some point

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
```diff
@@ -289,10 +307,51 @@ data class RuntimeType(val path: String, val dependency: RustDependency? = null)
...
-        val HttpResponse1x = Http1x.resolve("Response")
-        val HttpResponseBuilder1x = Http1x.resolve("response::Builder")
+
+        /**
+         * Returns the appropriate http crate based on HTTP version.
+         *
+         * For HTTP 1.x: returns http@1.x crate
+         * For HTTP 0.x: returns http@0.2.x crate
+         */
+        fun httpForConfig(runtimeConfig: RuntimeConfig): RuntimeType =
```
rcoh: personally, I'd consider dropping the `ForConfig` on all of these. obviously doesn't really matter either way, or call them `auto`

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/CargoDependency.kt
```diff
@@ -370,6 +371,12 @@ data class CargoDependency(
         val HttpBodyUtil01x: CargoDependency =
             CargoDependency("http-body-util", CratesIo("0.1.3"))
 
+        fun hyper(runtimeConfig: RuntimeConfig) =
```
rcoh, line 374: I'd name every conditional dependency `auto` so its easy to find and its obvious if it is giving an automatic version

codegen-server-test/build.gradle.kts
```diff
@@ -33,98 +33,155 @@ smithy {
...
+    return listOf(
+        // http@0 version (legacy) - no http-1x flag, add -http0x suffix
         CodegenTest(
-            "com.aws.example#PokemonService",
-            "pokemon-service-server-sdk",
-            imports = listOf("$commonModels/pokemon.smithy", "$commonModels/pokemon-common.smithy"),
+            service,
+            "$moduleName-http0x",
+            imports = imports,
+            extraCodegenConfig = extraCodegenConfig,
```
rcoh: consider setting `http-1x: false` explicitly so this still works when we change the default

codegen-server-test/build.gradle.kts
```diff
@@ -33,98 +33,155 @@ smithy {
...
+/**
+ * Helper function to generate both http@0 (legacy) and http@1 versions of a CodegenTest.
+ *
+ * @param service The service shape ID
+ * @param moduleName Base module name (without suffix)
+ * @param imports Optional list of model imports
+ * @param extraCodegenConfig Optional additional codegen config (goes inside "codegen": {})
+ * @return List containing both http@0 (with -http0x suffix) and http@1 (no suffix) versions
+ */
+fun generateBothHttpVersions(
```
rcoh: consider making this an extension function on `CodegenTest` so you can do `codegenTests.flatMap { it.bothHttpVersions() }`. A little bit more obvious & flexible I think

codegen-server-test/integration-tests/eventstreams/tests/structured_eventstream_tests.rs
```diff
@@ -50,41 +51,28 @@ struct TestServer {
...
-                let state = handler_state3.clone();
-                streaming_operation_with_initial_response_handler(input, state)
-            })
-            .streaming_operation_with_optional_data(move |input| {
-                let state = handler_state4.clone();
-                streaming_operation_with_optional_data_handler(input, state)
-            })
+
+        let config = RpcV2CborServiceConfig::builder()
+            .layer(AddExtensionLayer::new(state.clone()))
```
rcoh, line 56: I guess cleaner? but seems like this is actually likely to be harder to debug for people?

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerRootGenerator.kt
```diff
@@ -270,5 +412,14 @@ open class ServerRootGenerator(
...
         )
+
+        // Re-export serve function for HTTP 1.x
+        if (codegenContext.settings.runtimeConfig.httpVersion == HttpVersion.Http1x) {
+            rustWriter.rust(
+                """
+                pub use crate::server::serve::serve;
+                """,
+            )
+        }
```
rcoh, line 423: not worth it here, but you can find an example in the client of doing automatic pub-use when a type is actually used

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/testutil/ServerCodegenIntegrationTest.kt
```diff
@@ -43,7 +79,120 @@ fun serverIntegrationTest(
...
+        }
+    val shouldTestHttp1 =
+        when (testCoverage) {
+            is HttpTestType.ALL -> true
+            is HttpTestType.Only -> testCoverage.version == HttpTestVersion.HTTP_1_X
+            is HttpTestType.AsConfigured -> false // Already handled above
+        }
+
+    // Check if parallel execution is enabled via environment variable
+    val runInParallel = System.getenv("PARALLEL_HTTP_TESTS")?.toBoolean() ?: false
```
rcoh: actually used?

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/UserProvidedValidationExceptionDecoratorTest.kt
```diff
@@ -382,7 +383,7 @@ internal class UserProvidedValidationExceptionDecoratorTest {
 
     @Test
     fun `code compiles with custom validation exception using optionals`() {
-        serverIntegrationTest(completeTestModelWithOptionals)
+        serverIntegrationTest(completeTestModelWithOptionals, testCoverage = HttpTestType.AsConfigured)
```
rcoh: note in this context `AsConfigured` isn't obvious which one is going to get picked. Does this basically mean "the default?"

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
drganjoo: Yes, we have a test `Http1xDependencyTest.kt.SDK defaults to http-1x disabled, and dependencies are correct` that ensures by default `http1x == false` and the generated SDK correctly depends on `http/hyper@0`.

codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/RuntimeType.kt
drganjoo: I agree. But there are already member variables with similar names without the suffix, for example, `val Http`. They are used in the codebase to directly reference Http 0.x/1.x libraries (particularly in tests where we need to test specific library versions). I'd like to rename those, for example, from `val Http` to `val Http0x` to make it explicit that callers are choosing a specific library version. However, I prefer to handle this renaming in a separate CR; otherwise, a lot more files will be included in this CR. I'll create a follow-up CR for this refactoring.

For now, this CR has the `Auto` as the suffix instead of `ForConfig`.

codegen-server-test/build.gradle.kts
drganjoo: Good idea - thanks.

codegen-server-test/build.gradle.kts
drganjoo: I've done this deliberately as a safeguard to ensure that the default value of the flag is correct. If someone accidentally sets the http-1x flag to true, downstream tests, for example, `codegen-server-test/eventstream/legacy` would fail.

codegen-server-test/integration-tests/eventstreams/tests/structured_eventstream_tests.rs
drganjoo: >  but seems like this is actually likely to be harder to debug for people?

This approach matches our documentation. I wanted to have more sample points for AI tools to learn from.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/generators/ServerRootGenerator.kt
drganjoo: Ah nice! I'll check the client code for future reference.

codegen-server/src/main/kotlin/software/amazon/smithy/rust/codegen/server/smithy/testutil/ServerCodegenIntegrationTest.kt
drganjoo: No, not at the moment. I had added this to speed up the CI, but haven't tested this with Github. I'll remove this for now.

codegen-server/src/test/kotlin/software/amazon/smithy/rust/codegen/server/smithy/customizations/UserProvidedValidationExceptionDecoratorTest.kt
drganjoo: `AsConfigured` implies that `serverIntegrationTest` should not modify any settings and should use the `codeGen` flags as set by the caller. However, in practice, it has mostly been used for the default case. I'll rename it to `Default` to better reflect its actual usage.

buildSrc/src/main/kotlin/CodegenTestCommon.kt
```diff
@@ -23,6 +23,29 @@ data class CodegenTest(
...
+/**
+ * Extension function to generate both http@0 (legacy) and http@1 versions of a CodegenTest.
+ *
+ * @return List containing both http@0 (with -http0x suffix) and http@1 (no suffix) versions
+ */
+fun CodegenTest.bothHttpVersions(): List<CodegenTest> {
+    return listOf(
+        // http@0 version (legacy) - no http-1x flag, add -http0x suffix
+        this.copy(
+            module = "${this.module}-http0x",
```
rcoh, line 35: shouldn't this set `http-1x` to false?

buildSrc/src/main/kotlin/CodegenTestCommon.kt
```diff
@@ -23,6 +23,29 @@ data class CodegenTest(
...
+        ),
+        // http@1 version - with http-1x flag, no suffix
+        this.copy(
+            extraCodegenConfig =
+                if (this.extraCodegenConfig?.isNotEmpty() == true) {
+                    """"http-1x": true, ${this.extraCodegenConfig}"""
```
rcoh, line 41: shouldn't this also produce a suffix?

PR #4384, opened 2025-11-05
Refactor `RequestChecksumInterceptor` into distinct interceptors

ysaito1001:
## Motivation and Context
This PR breaks up `RequestChecksumInterceptor` into two interceptors: `RequestChecksumInterceptor` and `AwsChunkedContentEncodingInterceptor` with additional code restructuring to support the upcoming aws-chunked content encoding enhancement.

## Description
`RequestChecksumInterceptor` previously handled two distinct responsibilities:
1. Adding checksums (body or trailer)
2. Applying aws-chunked content encoding

This refactor extracts the second responsibility (aws-chunked logic) into `AwsChunkedContentEncodingInterceptor`, aiming to enable it via the dedicated trait when available in Smithy.

In addition, both interceptors now wrap bodies in `modify_before_transmit` instead of `modify_before_signing` per design requirements.

## Tips for merging to `feature/http-1.x`

This refactor reluctantly adds `http_0x` constructs that will conflict with `feature/http-1.x`. Update these files when merging:
- `aws-inlineable/src/http_request_checksum.rs` - Apply same 1.x updates as done [in branch](https://github.com/smithy-lang/smithy-rs/compare/feature/http-1.x?expand=1#diff-7c4006e142da69a7a442cd1d152bf5d0ae05a18c93a02c9754776ef29cef5dcd) (note: `test_checksum_body_is_retryable` was removed as it tested a utility only used in streaming cases against non-streaming bodies)
- `aws-inlineable/src/aws_chunked.rs` - Use 1.x for header names, values, and body trait methods in unit tests, as in `http_request_checksum.rs`
- `aws/rust-runtime/aws-runtime/src/content_encoding.rs` - Needs 1.x updates
```
    fn size_hint(&self) -> http_body_1x::SizeHint {
        http_body_1x::SizeHint::with_exact(self.encoded_length())
    }
```
to
```
    fn size_hint(&self) -> http_body_1x::SizeHint {
        http_body_1x::SizeHint::with_exact(self.options.encoded_length())
    }
```
because the `.encoded_length()` method has been moved from `AwsChunkedBody` to `AwsChunkedBodyOptions` (due to the `content-length` header needing encoded length during signing but body creation deferred to 
`modify_before_transmit`)

## Testing
- CI

----

### DIFF
--- a/aws/rust-runtime/aws-inlineable/src/aws_chunked.rs
+++ b/aws/rust-runtime/aws-inlineable/src/aws_chunked.rs
@@ -36,7 +36,7 @@ impl fmt::Display for Error {
         match self {
             Self::UnsizedRequestBody => write!(
                 f,
-                "Only request bodies with a known size can be aws-chunk encoded."
+                "Only request bodies with a known size can be aws-chunked encoded."
             ),
         }
     }
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -113,7 +113,7 @@ dependencies = [
 
 [[package]]
 name = "aws-runtime"
-version = "1.5.14"
+version = "1.5.15"
 dependencies = [
  "arbitrary",
  "aws-credential-types",
--- a/aws/rust-runtime/aws-runtime/Cargo.toml
+++ b/aws/rust-runtime/aws-runtime/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws-runtime"
-version = "1.5.14"
+version = "1.5.15"
 authors = ["AWS Rust SDK Team <aws-sdk-rust@amazon.com>"]
 description = "Runtime support code for the AWS SDK. This crate isn't intended to be used directly."
 edition = "2021"
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,4 +17,4 @@ allowLocalDeps=false
 # Avoid registering dependencies/plugins/tasks that are only used for testing purposes
 isTestingEnabled=true
 # codegen publication version
-codegenVersion=0.1.5
+codegenVersion=0.1.6


aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AwsChunkedContentEncodingDecorator.kt
```diff
@@ -0,0 +1,93 @@
...
+import software.amazon.smithy.rust.codegen.core.util.getTrait
+import software.amazon.smithy.rust.codegen.core.util.hasStreamingMember
+
+class AwsChunkedContentEncodingDecorator : ClientCodegenDecorator {
+    override val name: String = "AwsChunkedContentEncoding"
+
+    // This decorator must decorate after any of the following:
+    // - HttpRequestChecksumDecorator
+    // - HttpRequestCompressionDecorator
+    override val order: Byte = (minOf(HttpRequestChecksumDecorator.ORDER, HttpRequestCompressionDecorator.ORDER) - 1).toByte()
```
landonxjames, line 31: If we want it to always run after wouldn't it be `maxOf() +1`? If I remember correctly higher orders run later

ex:
https://github.com/smithy-lang/smithy-rs/blob/41ea15d217921e61ef3d8bc50b068b33efa8c78a/aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt#L47-L48

aws/rust-runtime/aws-inlineable/src/aws_chunked.rs
```diff
@@ -0,0 +1,331 @@
...
+enum Error {
+    UnsizedRequestBody,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::UnsizedRequestBody => write!(
+                f,
+                "Only request bodies with a known size can be chunk-encoded."
```
landonxjames: Maybe `... can be aws-chunked encoded` to avoid confusion with `transfer-encoding: chunked`?

aws/rust-runtime/aws-inlineable/src/aws_chunked.rs
```diff
@@ -0,0 +1,331 @@
...
+        Ok(())
+    }
+
+    fn modify_before_transmit(
+        &self,
+        ctx: &mut BeforeTransmitInterceptorContextMut<'_>,
+        _runtime_components: &RuntimeComponents,
+        cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        if must_not_use_chunked_encoding(ctx.request(), cfg) {
```
landonxjames, line 124: Since `modify_before_signing` will always run before this is there any point in checking `must_not_use_chunked_encoding` again? Is there a situation where the result there could change between signing and transmit?

aws/rust-runtime/aws-runtime/src/content_encoding.rs
```diff
@@ -32,6 +33,13 @@ pub struct AwsChunkedBodyOptions {
     /// The length of each trailer sent within an `AwsChunkedBody`. Necessary in
     /// order to correctly calculate the total size of the body accurately.
     trailer_lengths: Vec<u64>,
+    /// Whether the aws-chunked encoding is disabled. This could occur, for instance,
+    /// if a user specifies a custom checksum, rendering aws-chunked encoding unnecessary.
```
landonxjames, line 37: That might render it unnecessary, but I wonder if it would have performance implications? We should probably ask about when it should be disabled based on checksum behaviors.

aws/rust-runtime/aws-inlineable/src/aws_chunked.rs
ysaito1001: I like that clarification, will update the wording.

Updated in [7826b34](https://github.com/smithy-lang/smithy-rs/commit/7826b34b1a0fcef6eb508688e4a65833c604a200#diff-de4a9d7af586d88068314e28346249ca9b742adc4cdc8b9036af6dc62da41131)

aws/rust-runtime/aws-inlineable/src/aws_chunked.rs
ysaito1001: Yeah, I believe the early return check is still necessary here. Each interceptor method performs specific work, and removing this check could create an inconsistent scenario where relevant headers (like `content-encoding`) remain unset during signing, and yet the body gets wrapped in `AwsChunkedBody`.

This is a unfortunate byproduct from splitting the logic across multiple interceptor methods; each of these methods should only execute conditionally.

aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AwsChunkedContentEncodingDecorator.kt
ysaito1001: It's actually [the opposite](https://github.com/smithy-lang/smithy-rs/blob/main/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/smithy/customize/CoreCodegenDecorator.kt#L37-L38) of what you think it is :)

landonxjames: I find that very confusing lol. 

Anyway not related to your PR but it looks like the order for `AwsFluentClientDecorator` is probably wrong then since it is supposed to run after `AwsPresigningDecorator` but has a higher order.

aws/rust-runtime/aws-runtime/src/content_encoding.rs
ysaito1001: Discussed offline. Per the design implementation guide, when a user specifies checksum and checksum type headers, we should not set content-encoding. The current implementation is [already compliant](https://github.com/smithy-lang/smithy-rs/blob/41ea15d217921e61ef3d8bc50b068b33efa8c78a/aws/rust-runtime/aws-inlineable/src/http_request_checksum.rs#L284-L292), and this PR's refactoring preserves that existing logic.

aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AwsChunkedContentEncodingDecorator.kt
ysaito1001: Thanks for pointing it out, created [an issue](https://github.com/smithy-lang/smithy-rs/issues/4389). Should be a quick fix, if nothing breaks.

PR #4385, opened 2025-11-06
Add business metrics for observability and endpoint override

vcjana:
## Motivation and Context
This PR implements business metrics tracking for observability and endpoint override features.

## Description
This PR adds five new business metrics to track SDK feature usage:

**Observability Metrics:**
- `m/4`: Observability tracing enabled
- `m/5`: Observability metrics enabled
- `m/6`: OpenTelemetry tracing integration
- `m/7`: OpenTelemetry metrics integration

**Endpoint Configuration Metrics:**
- `m/N`: Custom endpoint URL override detection

## Testing
Integration tests added in `aws/sdk/integration-tests/s3/tests/business_metrics.rs` covering all seven metrics.

## Checklist

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/observability_detection.rs
+++ b/aws/rust-runtime/aws-runtime/src/observability_detection.rs
@@ -10,7 +10,11 @@
 //! - [`crate::observability_detection::ObservabilityDetectionInterceptor`]: Detects observability features during
 //!   request processing and tracks them for business metrics in the User-Agent header.
 
+#[cfg(all(not(target_arch = "powerpc"), not(target_family = "wasm")))]
+use crate::sdk_feature::AwsSdkFeature;
+#[cfg(all(not(target_arch = "powerpc"), not(target_family = "wasm")))]
 use aws_smithy_observability_otel::meter::OtelMeterProvider;
+#[cfg(all(not(target_arch = "powerpc"), not(target_family = "wasm")))]
 use aws_smithy_runtime::client::sdk_feature::SmithySdkFeature;
 use aws_smithy_runtime_api::box_error::BoxError;
 use aws_smithy_runtime_api::client::interceptors::context::BeforeTransmitInterceptorContextRef;
@@ -40,22 +44,37 @@ impl Intercept for ObservabilityDetectionInterceptor {
         &self,
         _context: &BeforeTransmitInterceptorContextRef<'_>,
         _runtime_components: &RuntimeComponents,
-        cfg: &mut ConfigBag,
+        _cfg: &mut ConfigBag,
     ) -> Result<(), BoxError> {
-        // Try to get the global telemetry provider
-        if let Ok(provider) = aws_smithy_observability::global::get_telemetry_provider() {
-            // Use type-safe downcasting to detect OpenTelemetry meter provider
-            // This works with any ProvideMeter implementation and doesn't require
-            // implementation-specific boolean flags
-            if provider
-                .meter_provider()
-                .as_any()
-                .downcast_ref::<OtelMeterProvider>()
-                .is_some()
-            {
-                // Track that observability metrics are enabled
-                cfg.interceptor_state()
-                    .store_append(SmithySdkFeature::ObservabilityMetrics);
+        #[cfg(all(not(target_arch = "powerpc"), not(target_family = "wasm")))]
+        {
+            // Try to get the global telemetry provider
+            if let Ok(provider) = aws_smithy_observability::global::get_telemetry_provider() {
+                let meter_provider = provider.meter_provider();
+
+                // Check if this is an OpenTelemetry meter provider
+                let is_otel = meter_provider
+                    .as_any()
+                    .downcast_ref::<OtelMeterProvider>()
+                    .is_some();
+
+                // Check if this is a noop provider (we don't want to track noop)
+                let is_noop = meter_provider
+                    .as_any()
+                    .downcast_ref::<aws_smithy_observability::noop::NoopMeterProvider>()
+                    .is_some();
+
+                if !is_noop {
+                    // Track generic observability metrics (for any non-noop provider)
+                    _cfg.interceptor_state()
+                        .store_append(SmithySdkFeature::ObservabilityMetrics);
+
+                    // If it's specifically OpenTelemetry, track that too
+                    if is_otel {
+                        _cfg.interceptor_state()
+                            .store_append(AwsSdkFeature::ObservabilityOtelMetrics);
+                    }
+                }
             }
         }
 
@@ -73,7 +92,9 @@ mod tests {
     use aws_smithy_runtime_api::client::runtime_components::RuntimeComponentsBuilder;
     use aws_smithy_types::config_bag::ConfigBag;
 
+    #[cfg(all(not(target_arch = "powerpc"), not(target_family = "wasm")))]
     #[test]
+    #[serial_test::serial]
     fn test_detects_noop_provider() {
         let mut context = InterceptorContext::new(Input::doesnt_matter());
         context.enter_serialization_phase();
@@ -93,15 +114,33 @@ mod tests {
             .read_before_signing(&ctx, &rc, &mut cfg)
             .unwrap();
 
-        // Should not track any features for noop provider since it doesn't downcast to OtelMeterProvider
-        let smithy_features: Vec<_> = cfg.load::<SmithySdkFeature>().collect();
-        assert_eq!(smithy_features.len(), 0);
+        // Should not track any features for noop provider
+        let smithy_features: Vec<_> = cfg
+            .interceptor_state()
+            .load::<SmithySdkFeature>()
+            .cloned()
+            .collect();
+        assert_eq!(
+            smithy_features.len(),
+            0,
+            "Should not track Smithy features for noop provider"
+        );
 
-        let aws_features: Vec<_> = cfg.load::<AwsSdkFeature>().collect();
-        assert_eq!(aws_features.len(), 0);
+        let aws_features: Vec<_> = cfg
+            .interceptor_state()
+            .load::<AwsSdkFeature>()
+            .cloned()
+            .collect();
+        assert_eq!(
+            aws_features.len(),
+            0,
+            "Should not track AWS features for noop provider"
+        );
     }
 
+    #[cfg(all(not(target_arch = "powerpc"), not(target_family = "wasm")))]
     #[test]
+    #[serial_test::serial]
     fn test_custom_provider_not_detected_as_otel() {
         use aws_smithy_observability::meter::{Meter, ProvideMeter};
         use aws_smithy_observability::noop::NoopMeterProvider;
@@ -150,24 +189,27 @@ mod tests {
             .read_before_signing(&ctx, &rc, &mut cfg)
             .unwrap();
 
-        // Should NOT track any features for custom provider since it doesn't downcast to OtelMeterProvider
-        // The new implementation only emits metrics when OTel is detected
+        // Should track generic observability metrics for custom provider
         let smithy_features: Vec<_> = cfg
             .interceptor_state()
             .load::<SmithySdkFeature>()
             .cloned()
             .collect();
         assert!(
-            !smithy_features.iter().any(|f| *f == SmithySdkFeature::ObservabilityMetrics),
-            "Should not detect custom provider as having observability metrics (only OTel is tracked)"
+            smithy_features.contains(&SmithySdkFeature::ObservabilityMetrics),
+            "Should detect custom provider as having observability metrics"
         );
 
-        // Verify no AWS-specific features are tracked for custom provider
+        // Should NOT track AWS-specific observability metrics for custom provider
         let aws_features: Vec<_> = cfg
             .interceptor_state()
             .load::<AwsSdkFeature>()
             .cloned()
             .collect();
+        assert!(
+            !aws_features.contains(&AwsSdkFeature::ObservabilityOtelMetrics),
+            "Should NOT track OTel-specific metrics for custom provider"
+        );
         assert_eq!(
             aws_features.len(),
             0,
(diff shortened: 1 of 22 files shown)


aws/sdk/integration-tests/s3/tests/business_metrics.rs
ysaito1001: Why has this test been removed?

aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/EndpointOverrideDecoratorTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    fun `decorator is registered in AwsCodegenDecorator list`() {
+        // Verify that EndpointOverrideDecorator is in the DECORATORS list
+        val decoratorNames = DECORATORS.map { it.name }
+        assert(decoratorNames.contains("EndpointOverride")) {
+            "EndpointOverrideDecorator should be registered in DECORATORS list"
+        }
+    }
+
+    @Test
+    fun `generated code includes endpoint override interceptor registration`() {
```
landonxjames: This test doesn't quite test what the name suggests it does. It tests that it compiles, but it would compile just find if the EndpointOverride interceptor wasn't inserted at all. Same with the test below. I would try to write a test that actually ensures the interceptor is inserted, likely by making a request with an overridden endpoint and checking the user-agent on that request.

aws/rust-runtime/aws-runtime/src/user_agent/metrics.rs
```diff
@@ -222,6 +227,10 @@ impl ProvideBusinessMetric for AwsSdkFeature {
             S3Transfer => Some(BusinessMetric::S3Transfer),
             SsoLoginDevice => Some(BusinessMetric::SsoLoginDevice),
             SsoLoginAuth => Some(BusinessMetric::SsoLoginAuth),
+            ObservabilityMetrics => Some(BusinessMetric::ObservabilityMetrics),
```
landonxjames: The Observability metrics should probably all be in `SmithySdkFeature` instead of `AwsSdkFeature`. Nothing about the observability interfaces is Aws specific.

aws/rust-runtime/aws-runtime/src/observability_detection.rs
```diff
@@ -0,0 +1,113 @@
...
+    ) -> Result<(), BoxError> {
+        // Try to get the global telemetry provider
+        if let Ok(provider) = aws_smithy_observability::global::get_telemetry_provider() {
+            // Check if it's not a noop provider
+            if !provider.is_noop() {
+                // Track that observability metrics are enabled
+                cfg.interceptor_state()
+                    .store_append(AwsSdkFeature::ObservabilityMetrics);
+
+                // PRAGMATIC APPROACH: Track tracing based on meter provider state
```
landonxjames: I would say we should probably just not emit this one since we don't actually support it, and it could be confusing.

aws/rust-runtime/aws-runtime/src/observability_detection.rs
```diff
@@ -0,0 +1,113 @@
...
+                    .store_append(SmithySdkFeature::ObservabilityTracing);
+
+                // Check if it's using OpenTelemetry
+                if provider.is_otel() {
+                    cfg.interceptor_state()
+                        .store_append(AwsSdkFeature::ObservabilityOtelMetrics);
+
+                    // If using OpenTelemetry for metrics, likely using it for tracing too
+                    cfg.interceptor_state()
+                        .store_append(AwsSdkFeature::ObservabilityOtelTracing);
```
landonxjames: Same with this one, and this could be especially misleading since users can hook their `tracing` crate traces up to OTel.

aws/rust-runtime/aws-runtime/src/observability_detection.rs
```diff
@@ -0,0 +1,113 @@
...
+            .read_before_signing(&ctx, &rc, &mut cfg)
+            .unwrap();
+
+        // Should not track any features for noop provider
+        let features: Vec<_> = cfg.load::<AwsSdkFeature>().collect();
+        assert_eq!(features.len(), 0);
+    }
+
+    // Note: We cannot easily test non-noop providers without creating a custom meter provider
+    // implementation, which would require exposing internal types. The noop test above
```
landonxjames: I don't think this would require leaking any internal types? The OTel meter provider already exists so anything require to implement another one should already be `pub`. If something is missing we probably need to make it `pub` anyway so users can implement their own providers.

aws/rust-runtime/aws-runtime/src/observability_plugin.rs
```diff
@@ -0,0 +1,60 @@
...
+
+impl ObservabilityRuntimePlugin {
+    /// Create a new ObservabilityRuntimePlugin
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+impl RuntimePlugin for ObservabilityRuntimePlugin {
+    fn config(&self) -> Option<FrozenLayer> {
```
landonxjames: None of this really seems necessary since all of these features are also being set in the `ObservabilityDetectionInterceptor` above. Is there a reason to attempt to set them in both the plugin and the interceptor?

aws/sdk/integration-tests/s3/tests/business_metrics.rs
```diff
@@ -1,63 +0,0 @@
-/*
```
landonxjames, line 1: Why delete these tests?

aws/sdk/integration-tests/s3/Cargo.toml
```diff
@@ -22,6 +22,7 @@ aws-runtime = { path = "../../build/aws-sdk/sdk/aws-runtime", features = ["test-
 aws-sdk-s3 = { path = "../../build/aws-sdk/sdk/s3", features = ["test-util", "behavior-version-latest"] }
 aws-smithy-async = { path = "../../build/aws-sdk/sdk/aws-smithy-async", features = ["test-util", "rt-tokio"] }
 aws-smithy-http = { path = "../../build/aws-sdk/sdk/aws-smithy-http" }
+aws-smithy-observability = { path = "../../build/aws-sdk/sdk/aws-smithy-observability" }
```
landonxjames: This new deps in this crate don't seem to be used? There is already a separate [crate for telemetry integration tests](https://github.com/smithy-lang/smithy-rs/tree/main/aws/sdk/integration-tests/telemetry), so if you were going to add tests for that I would put them there.

rust-runtime/aws-smithy-observability/src/lib.rs
```diff
@@ -25,7 +25,7 @@ mod error;
 pub use error::{ErrorKind, GlobalTelemetryProviderError, ObservabilityError};
 pub mod global;
 pub mod meter;
-mod noop;
+pub mod noop;
```
landonxjames, line 29: If we have to make this `pub` (presumably for some of the tests you added above?) then we should probably also make it `#[doc(hidden)]` just to avoid confusing users since this isn't really meant for external consumption.

rust-runtime/aws-smithy-observability/src/meter.rs
```diff
@@ -14,9 +14,18 @@ use crate::{attributes::Attributes, instruments::ProvideInstrument};
...
 
 /// Provides named instances of [Meter].
-pub trait ProvideMeter: Send + Sync + Debug {
+pub trait ProvideMeter: Send + Sync + Debug + 'static {
     /// Get or create a named [Meter].
     fn get_meter(&self, scope: &'static str, attributes: Option<&Attributes>) -> Meter;
+
+    /// Downcast to `Any` for type inspection.
+    ///
+    /// The default implementation returns a reference to `()`, which will fail
```
landonxjames: Couldn't the default implementation just return `&self`? Feels like that should work, but also wouldn't be surprised if there was some type wonkiness that didn't allow it.

rust-runtime/aws-smithy-observability/src/provider.rs
```diff
@@ -13,27 +13,44 @@ use crate::{meter::ProvideMeter, noop::NoopMeterProvider};
 #[non_exhaustive]
 pub struct TelemetryProvider {
     meter_provider: Arc<dyn ProvideMeter + Send + Sync>,
+    is_otel: bool,
```
landonxjames: The `TelemetryProvder` should probably not have any fields/data related to specific implementations. Imagine that in the future `FooMetrics` is the hot new metrics framework. Then we have to add an `is_foo_metrics` field, and this could keep happening.

If we really need this it should probably be a `name: String` or something that can be set.

rust-runtime/aws-smithy-observability/src/provider.rs
```diff
@@ -61,10 +80,17 @@ impl TelemetryProviderBuilder {
         self
     }
 
+    /// Mark this provider as using OpenTelemetry.
+    pub fn with_otel(mut self, is_otel: bool) -> Self {
```
landonxjames: I don't really like this, it would allow the `TelemetryProvider` to lie about what implementation it is using. There is nothing actually tying this to the use of our OTel implementation.

rust-runtime/aws-smithy-runtime/src/client/orchestrator/endpoints.rs
```diff
@@ -47,6 +47,7 @@ impl ResolveEndpoint for StaticUriEndpointResolver {
     fn resolve_endpoint<'a>(&'a self, _params: &'a EndpointResolverParams) -> EndpointFuture<'a> {
         EndpointFuture::ready(Ok(Endpoint::builder()
             .url(self.endpoint.to_string())
+            .property("is_custom_endpoint", true)
```
landonxjames: This property name doesn't seem to be referenced anywhere else in this PR?

gradle.properties
```diff
@@ -12,6 +12,8 @@ org.gradle.jvmargs=-Xmx1024M
 smithy.rs.runtime.crate.stable.version=1.1.7
 # Version number to use for the generated unstable runtime crates
 smithy.rs.runtime.crate.unstable.version=0.60.6
+# Version number for aws-smithy-observability crate
```
landonxjames: This seems unnecessary. How is it being used?

aws/rust-runtime/aws-runtime/src/observability_detection.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn read_before_signing(
+        &self,
+        _context: &BeforeTransmitInterceptorContextRef<'_>,
+        _runtime_components: &RuntimeComponents,
+        cfg: &mut ConfigBag,
+    ) -> Result<(), BoxError> {
+        // Try to get the global telemetry provider
+        if let Ok(provider) = aws_smithy_observability::global::get_telemetry_provider() {
+            // Check if it's not a noop provider
+            if !provider.is_noop() {
```
landonxjames: This isn't quite enough information to know that the provider is our OTel impl. This just says it isn't the noop impl. But a user could implement the traits their self and provide a different implementation (say for the [metrics](https://crates.io/crates/metrics) crate), or we could provide another non-OTel impl in the future that would break this assumption.

landonxjames: I think the better approach here would be to use the `as_any` method you added to the `ProvideMeter` trait. Something along the lines of `provider.as_any().downcast_ref::<OtelMeterProvider>()` which returns an `Option<&OtelMeterProvider>` and then only emit the metric if it is `Some(_)`. That would allow us to get rid of the `is_otel` and `is_noop` stuff and wouldn't break if users were using some third implementation.

aws/rust-runtime/aws-runtime/src/observability_plugin.rs
landonxjames, line 1: The runtime plugin and the interceptor could probably just go in the same file to keep things together.

aws/rust-runtime/aws-runtime/src/user_agent/interceptor.rs
```diff
@@ -147,6 +148,14 @@ impl Intercept for UserAgentInterceptor {
                 .map(|m| ua.add_business_metric(m));
         }
 
+        // Also load AWS SDK features from interceptor_state (where interceptors store them)
+        let aws_sdk_features_from_interceptor = cfg.interceptor_state().load::<AwsSdkFeature>();
```
landonxjames: I think the `cfg.load::<AwsSdkFeature>()` should capture all of the `AwsSdkFeature` in the cfg bag so it shouldn't be necessary to grab them from this single layer? I might be wrong about though, will have to check in the morning, but @ysaito1001 probably knows.

ysaito1001: Correct, it's covered by the code above from line 144 to line 149. We can remove this newly added lines of code.

vcjana: I'll remove the code. I was being overly cautious about the ConfigBag layering, but now I understand that `cfg.load()` traverses all layers including interceptor_state automatically.

aws/rust-runtime/aws-runtime/src/observability_detection.rs
vcjana: I'll remove the tracing tracking since we don't actually support it yet.

aws/rust-runtime/aws-runtime/src/observability_plugin.rs
vcjana: Moving the `ObservabilityRunTimePlugin` into the observability_detection.rs

rust-runtime/aws-smithy-observability/src/meter.rs
vcjana: The issue is that Self in a trait method doesn't have a known size at compile time. We need to add a where Self: Sized bound.

vcjana:
Closing this PR as I've created separate PRs for both observability and endpoint override.

PR #4386, opened 2025-11-06
Fix missing Kotlin stdlib versions in published POMs

drganjoo:
### Problem

Published Maven POMs were missing explicit version declarations for Kotlin stdlib dependencies (e.g., kotlin-stdlib-jdk8), causing dependency resolution issues for consumers of these artifacts.

### Solution

This PR adds explicit version management for Kotlin stdlib dependencies in two places:

1. Resolution Strategy (smithy-rs.kotlin-conventions.gradle.kts):
    - Added a resolution strategy that forces all kotlin-stdlib-* dependencies to use the explicit version from the version catalog
    - Ensures consistent Kotlin stdlib versions across all modules at build time
  
2. POM Version Mapping (smithy-rs.publishing-conventions.gradle.kts):
    - Configured versionMapping for published POMs to include resolved versions from the runtime classpath
    - Ensures that transitive Kotlin dependencies have explicit versions in generated POMs
    - Applied to both java-api and java-runtime usages

###  Testing

  - Build passes with explicit versions resolved for Kotlin stdlib dependencies
  - Published POMs should now include proper version declarations for all Kotlin stdlib artifacts

### DIFF
--- a/buildSrc/src/main/kotlin/smithy-rs.publishing-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/smithy-rs.publishing-conventions.gradle.kts
@@ -3,7 +3,6 @@ plugins {
     signing
 }
 
-// Workaround per: https://github.com/gradle/gradle/issues/15383
 val Project.libs get() = the<org.gradle.accessors.dm.LibrariesForLibs>()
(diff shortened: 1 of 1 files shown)


aajtodd (approved):
Odd that this is even necessary but seems ok.

PR #4388, opened 2025-11-07
Add then_compute_response method to aws-smithy-mocks

greenwoodcm:
Adds a new method that allows developers to stub arbitrary responses (outputs, errors, or HTTP responses) based on input request properties.

Unlike then_compute_output which only returns successful outputs, this method enables conditional response generation where the response type can vary based on the input.

Changes:
- Make MockResponse enum public to allow user construction
- Add RuleBuilder::then_compute_response() method
- Add ResponseSequenceBuilder::compute_response() internal method
- Export MockResponse from lib.rs
- Add integration test demonstrating conditional responses
- Update README with usage example

Example usage:
```
  mock!(Client::get_object) .then_compute_response(|req| { if req.key() == Some("error-key") { MockResponse::Error(...) } else { MockResponse::Output(...) } })
```

## Motivation and Context
Allow for stubbing/faking that sends failures back to the caller

## Testing
crate tests

## Checklist

----

### DIFF
--- a/aws/rust-runtime/Cargo.lock
+++ b/aws/rust-runtime/Cargo.lock
@@ -479,9 +479,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.44"
+version = "1.2.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "37521ac7aabe3d13122dc382493e20c9416f299d2ccd5b3a5340a2570cdeb0f3"
+checksum = "35900b6c8d709fb1d854671ae27aeaa9eec2f8b01b364e1619a40da3e6fe2afe"
 dependencies = [
  "find-msvc-tools",
  "shlex",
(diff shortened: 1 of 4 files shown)


.changelog/1762538321.md
```diff
@@ -0,0 +1,11 @@
+---
+applies_to: ["aws-smithy-mocks"]
```
aajtodd: should be "client", this is failing to parse with unexpected value

greenwoodcm:
thanks for the consideration @aajtodd !  i updated the changelog and bumped to `0.3.0` in the `Cargo.toml`.  i don't know if you'd prefer a patch version bump, i seem to remember we bumped the minor version on my last PR so following that protocol here.  happy to make this a patch bump instead, i'll defer to your guidance on that.

rust-runtime/aws-smithy-mocks/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-mocks"
-version = "0.2.0"
+version = "0.3.0"
```
aajtodd: Actually in your PR it was a patch bump this was bumped recently to 0.2.0 when we introduced with_test_defaults_v2. This can be a patch to 0.2.1.

greenwoodcm: ok will change

PR #4392, opened 2025-11-13
Avoid emitting repeated business metrics

ysaito1001:
## Motivation and Context
P334648004

## Description
At times, the Rust SDK emits the business metric looking as follows:
```
M/E,b,0,0,0,0,0,0
```
The issue is due to [the retry loop](https://github.com/smithy-lang/smithy-rs/blob/54f42235c9bce0be0207d5d00a6b35c2501e81d1/rust-runtime/aws-smithy-runtime/src/client/orchestrator.rs#L290) in the orchestrator, where any features added to the config bag within `try_attempt` persist throughout subsequent retry iterations (we do have the [rewind](https://github.com/smithy-lang/smithy-rs/blob/54f42235c9bce0be0207d5d00a6b35c2501e81d1/rust-runtime/aws-smithy-runtime-api/src/client/interceptors/context.rs#L367) method but it currently does not touch the config bag at all).

This PR uniquifies business metric values emitted in `UserAgentInterceptor`. It also fixes a bug related to the order for emitting `AwsCredentialFeature` where it previously emitted their metric values in a reversed order.

## Dismissed alternative
We discussed an alternative approach involving a separate interceptor that runs `modify_before_attempt_completion` that removes `AwsCredentialsFeature` from config bag so the next retry loop can start from the clean slate. This would require extra code, and the approach employed in this PR is simpler. Plus, should the need for the cleanup interceptor arise, it can co-exist with the fix of this PR.

## Testing
- CI
- Unit tests in `UserAgentInterceptor`
- Codegen test in `UserAgentDecoratorTest`

----

### DIFF
--- a/aws/rust-runtime/aws-runtime/src/user_agent/interceptor.rs
+++ b/aws/rust-runtime/aws-runtime/src/user_agent/interceptor.rs
@@ -29,7 +29,8 @@ macro_rules! add_metrics_unique {
     ($features:expr, $ua:expr, $added:expr) => {
         for feature in $features {
             if let Some(m) = feature.provide_business_metric() {
-                if $added.insert(m.clone()) {
+                if !$added.contains(&m) {
+                    $added.insert(m.clone());
                     $ua.add_business_metric(m);
                 }
             }
@@ -42,7 +43,8 @@ macro_rules! add_metrics_unique_reverse {
         let mut unique_metrics = Vec::new();
         for feature in $features {
             if let Some(m) = feature.provide_business_metric() {
-                if $added.insert(m.clone()) {
+                if !$added.contains(&m) {
+                    $added.insert(m.clone());
                     unique_metrics.push(m);
                 }
             }


aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/UserAgentDecoratorTest.kt
```diff
@@ -145,6 +145,82 @@ class UserAgentDecoratorTest {
...
+                    val rc = context.runtimeConfig
+                    val moduleName = context.moduleUseName()
+                    rustTemplate(
+                        """
+                        use $moduleName::config::{AppName, Credentials, Region, SharedCredentialsProvider};
+                        use $moduleName::{Config, Client};
+
+                        let http_client = #{StaticReplayClient}::new(vec![
+                            #{ReplayEvent}::new(
+                                #{http}::Request::builder()
```
ysaito1001, line 162: Will replace this with `http-1.x` construct and backport to the `feature/http-1.x` branch.

aws/rust-runtime/aws-runtime/src/user_agent/interceptor.rs
```diff
@@ -25,6 +25,34 @@ use crate::sdk_feature::AwsSdkFeature;
 use crate::user_agent::metrics::ProvideBusinessMetric;
 use crate::user_agent::{AdditionalMetadata, ApiMetadata, AwsUserAgent, InvalidMetadataValue};
 
+macro_rules! add_metrics_unique {
+    ($features:expr, $ua:expr, $added:expr) => {
+        for feature in $features {
+            if let Some(m) = feature.provide_business_metric() {
+                if $added.insert(m.clone()) {
```
landonxjames: Wonder if we could avoid the clone by doing a `.contains(&m)` and only cloning/inserting if it is not already in? Seems like in most cases (the many repeated 0s) the insert will fail so the clone would be immediately dropped

ysaito1001: A good suggestion. It's like `look before you leap` (LBYL), and `easier to ask forgiveness than permission` (EAFP) in Python. The current code is latter, which always involves clone, but the former allows us to use a borrow to "look before" and to clone only when necessary. ~Will update.~ Updated in 7152591

PR #4394, opened 2025-11-13
Add support for success reward and fractional token mgmt

annahay4:
## Motivation and Context
Provides the ability to grant reward tokens for all successful requests.
Adds fractional token management, which is required for cases where a customer wants to grant a fraction of a token for a success.

## Description
This PR adds configurable token bucket success rewards and fractional token management to the retry strategy. Key changes include:

- Added success_reward field to TokenBucket with a default value of 0.0
- Implemented reward_success() method that accumulates fractional tokens and converts them to full tokens when they reach 1.0 or more
- Modified the standard retry strategy to call reward_success() on successful requests
- Added success_reward() builder method to configure the reward amount
- Implemented fractional token tracking using Arc<Mutex<f64>> to safely accumulate partial tokens across requests

## Testing
Ran unit tests.
Added unit tests covering:

- Fractional token accumulation and conversion to full tokens (test with 0.4 reward requiring 3 successes to generate 1 token)
- Max capacity enforcement when adding reward tokens
- Builder configuration with custom success reward values

## Checklist

- [X] For changes to the smithy-rs codegen or runtime crates, I have created a changelog entry Markdown file in the `.changelog` directory, specifying "client," "server," or both in the `applies_to` key.

----

### DIFF
--- a/aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/RetryPartitionTest.kt
+++ b/aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/RetryPartitionTest.kt
@@ -114,7 +114,7 @@ class RetryPartitionTest {
                     "RetryPartition" to RuntimeType.smithyRuntime(ctx.runtimeConfig).resolve("client::retries::RetryPartition"),
                     "RuntimeComponents" to RuntimeType.runtimeComponents(ctx.runtimeConfig),
                     "TokenBucket" to RuntimeType.smithyRuntime(ctx.runtimeConfig).resolve("client::retries::TokenBucket"),
-                    "MAXIMUM_CAPACITY" to RuntimeType.smithyRuntime(ctx.runtimeConfig).resolve("client::retries::token_bucket::MAXIMUM_CAPACITY"),
+                    "MAXIMUM_CAPACITY" to RuntimeType.smithyRuntime(ctx.runtimeConfig).resolve("client::retries::MAXIMUM_CAPACITY"),
                 )
             crate.integrationTest("custom_retry_partition") {
                 tokioTest("test_custom_token_bucket") {
--- a/rust-runtime/aws-smithy-runtime/src/client/retries.rs
+++ b/rust-runtime/aws-smithy-runtime/src/client/retries.rs
@@ -18,7 +18,7 @@ use std::fmt;
 pub use client_rate_limiter::{
     ClientRateLimiter, ClientRateLimiterBuilder, ClientRateLimiterPartition,
 };
-pub use token_bucket::{TokenBucket, TokenBucketBuilder};
+pub use token_bucket::{TokenBucket, TokenBucketBuilder, MAXIMUM_CAPACITY};
 
 use std::borrow::Cow;
 


rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -55,6 +61,8 @@ impl TokenBucket {
             max_permits: Semaphore::MAX_PERMITS,
             timeout_retry_cost: 0,
             retry_cost: 0,
+            success_reward: 0.0,
+            fractional_tokens: Arc::new(Mutex::new(0.0)),
```
aajtodd: Can probably avoid a mutex here with something like https://github.com/rust-lang/rust/issues/72353#issuecomment-1093729062

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
```diff
@@ -210,8 +210,11 @@ impl RetryStrategy for StandardRetryStrategy {
             .unwrap_or(false);
         update_rate_limiter_if_exists(runtime_components, cfg, is_throttling_error);
 
-        // on success release any retry quota held by previous attempts
+        // on success release any retry quota held by previous attempts and award success tokens
         if !ctx.is_failed() {
+            // When a request succeeds, we grant an award, if present
+            token_bucket.reward_success();
```
aajtodd: This will reward every successful attempt even retries, is that what you were going for?

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
```diff
@@ -210,8 +210,11 @@ impl RetryStrategy for StandardRetryStrategy {
...
+
             if let NoPermitWasReleased = self.release_retry_permit() {
```
aajtodd: Do we need to worry about [`add_permits`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.add_permits) panicking if we exceed `MAX_PERMITS`? e.g. since we are now rewarding on success (or I believe eventually you want to refill on time as well) then on `Drop` a released permit puts it's tokens back.

aajtodd (commented):
Couple small comments/questions.

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
annahay4: Yes, that is the intention

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
annahay4: This looks great - I'll set this up and run some tests.

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
annahay4: Yes, the CI testing revealed that `let tokens_to_add = amount.min(self.max_permits - self.semaphore.available_permits());` was causing panics in the case where available > max. I've updated this in a new commit so that in the case where available > max, we just return without adding permits. 

In the case where we have a race condition that results in the number of available permits being greater than max after a `Drop` occurs, it will self-correct when a sufficient number of tokens are acquired. My initial assessment is that this is something that we are at risk of regardless of where the drop and add_permits are in relation to each other, due to multi-threading.

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -21,6 +25,44 @@ pub struct TokenBucket {
     max_permits: usize,
     timeout_retry_cost: u32,
     retry_cost: u32,
+    success_reward: f64,
+    fractional_tokens: AtomicF64,
+}
+
+pub struct AtomicF64 {
```
aajtodd: I'd keep this pub(crate) at most

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -77,12 +123,39 @@ impl TokenBucket {
...
-            self.semaphore.add_permits(PERMIT_REGENERATION_AMOUNT)
+        self.add_tokens(PERMIT_REGENERATION_AMOUNT);
+    }
+
+    pub(crate) fn reward_success(&self) {
+        // Verify that fractional tokens have not become corrupted
+        if !self.fractional_tokens.load().is_finite() {
+            tracing::error!("Fractional tokens corrupted to: {}", self.fractional_tokens.load());
+            // If corrupted, reset to the number of permits the bucket was created with
+            self.fractional_tokens.store(self.max_permits as f64);
```
aajtodd: Why wouldn't you reset to zero?

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -77,12 +123,39 @@ impl TokenBucket {
...
+        }
+
+        if self.success_reward > 0.0 {
+            self.fractional_tokens.store(self.fractional_tokens.load() + self.success_reward);
+        }
+
+        let full_tokens_accumulated = self.fractional_tokens.load().floor();
+        if full_tokens_accumulated >= 1.0 {
+            self.add_tokens(full_tokens_accumulated as usize);
+            self.fractional_tokens.store(self.fractional_tokens.load() - full_tokens_accumulated);
```
aajtodd: You call load three different times in this function, just call load once:

```rs
let fractional_tokens = self.fractional_tokens.load();
let full_tokens_accumulated = fractional_tokens.floor();
```

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -77,12 +123,39 @@ impl TokenBucket {
...
+        // Verify that fractional tokens have not become corrupted
+        if !self.fractional_tokens.load().is_finite() {
+            tracing::error!("Fractional tokens corrupted to: {}", self.fractional_tokens.load());
+            // If corrupted, reset to the number of permits the bucket was created with
+            self.fractional_tokens.store(self.max_permits as f64);
+            return;
+        }
+
+        if self.success_reward > 0.0 {
+            self.fractional_tokens.store(self.fractional_tokens.load() + self.success_reward);
```
aajtodd: Could make this a single store as well by doing the calculations up front and then you have a single load and single store.

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
annahay4: The initial state of the bucket is the last known healthy state. Because the corruption would've happened at no fault to the user, we don't want to punish them by not allowing retries. Therefore, returning to the last known healthy state is the safest path.

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
annahay4: ack, will update

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
annahay4: ack, will update

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
annahay4: ack, will update

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -89,19 +128,31 @@ impl TokenBucket {
...
             self.add_tokens(full_tokens_accumulated as usize);
+            self.fractional_tokens.store(self.fractional_tokens.load() - full_tokens_accumulated);
         }
-    } 
+    }
 
     fn add_tokens(&self, amount: usize) {
-        let tokens_to_add = amount.min(self.max_permits - self.semaphore.available_permits());
+        let available = self.semaphore.available_permits();
+        if available >= self.max_permits {
```
annahay4, line 259: @aajtodd this addresses the concern you brought up in your comment "I don't know that we can ship this without accounting for this somehow. You're bound to hit it at some point otherwise." - with this check, we ensure we don't do any math that will result in an overflow/underflow error.

aajtodd: `add_tokens` is only called by our internal logic though not when an `OwnedSemaphorePermit` is dropped (which [internally calls `add_permits`](https://docs.rs/tokio/latest/src/tokio/sync/semaphore.rs.html#1202-1206). 

If we accumulate some fractional tokens then at some point as permits are returned we could overflow right?

annahay4: To make sure I'm answering the right question - are you worried about the overflow on L155, or overflow as in the bucket having more permits than it should?

aajtodd: I was worried about [`add_permits` ](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.add_permits) panicking if we exceeded the maximum permits for the semaphore but I can see now I misread it (it panics if you exceed [Semaphore::MAX_PERMITS](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#associatedconstant.MAX_PERMITS) not the amount configured when the bucket was created).

annahay4: I see. In that case, I think we should make a cap on max size - as is, the max usize is greater than max_permits so we will need a cap either way (1.8x10^19(usize max) vs 2.3x10^18 (semaphore max permits)). I'd be shocked if someone needed a token bucket anywhere near that big. Do you have any opinions on what a reasonable max would be? I'd propose 1 million tokens max (10^6).

aajtodd: How would you enforce a cap? Isn't `max_permits` already our cap?

annahay4: We need to enforce a max value for max_permits

aajtodd: Right how would you enforce it given how owned permits work on `Drop`? You can enforce it here in `add_tokens` but that isn't called when dropping permits. Feel free to ping me offline if you need/want to discuss it in detail.

rust-runtime/aws-smithy-mocks/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-mocks"
-version = "0.2.1"
+version = "0.2.2"
```
aajtodd, line 3: Why did this change?

rust-runtime/aws-smithy-runtime/temp_test.rs
aajtodd, line 1: remove?

rust-runtime/aws-smithy-runtime/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aws-smithy-runtime"
-version = "1.9.4"
+version = "1.9.5"
```
aajtodd, line 3: unchanged?

aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/RetryPartitionTest.kt
```diff
@@ -139,7 +139,7 @@ class RetryPartitionTest {
                             ) -> Result<(), #{BoxError}> {
                                 self.called.fetch_add(1, Ordering::Relaxed);
                                 let token_bucket = cfg.load::<#{TokenBucket}>().unwrap();
-                                let expected = format!("permits: {}", tokio::sync::Semaphore::MAX_PERMITS);
+                                let expected = format!("permits: 500000000");
```
aajtodd: Should we expose this as a public constant `MAX_TOKENS`/`MAX_PERMITS` so we don't have to have magic numbers in tests

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -64,6 +118,16 @@ impl TokenBucket {
     }
 
     pub(crate) fn acquire(&self, err: &ErrorKind) -> Option<OwnedSemaphorePermit> {
+        // We have to handle the case where the number of permits in the semaphore exceeds the intended
+        // max. This can occur when the bucket is already at max capacity and then an OwnedSemaphorePermit
```
aajtodd: > this can occur when the bucket is already at max capacity (success reward is > 0), ...

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -77,10 +141,40 @@ impl TokenBucket {
...
+        self.add_tokens(PERMIT_REGENERATION_AMOUNT);
+    }
+
+    pub(crate) fn reward_success(&self) {
+        let mut calc_fractional_tokens = self.fractional_tokens.load();
+        // Verify that fractional tokens have not become corrupted
+        if !calc_fractional_tokens.is_finite() {
+            tracing::error!("Fractional tokens corrupted to: {}", calc_fractional_tokens);
+            // If corrupted, reset to the number of permits the bucket was created with
+            self.fractional_tokens.store(self.max_permits as f32);
```
aajtodd: I still think this is "wrong", we're rewarding success and every whole token is converted and added to our actual tokens/permits. Setting this to max permits is just going to immediately on the next go around fill up the bucket to max permits. I think this should be zero.

aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/RetryPartitionTest.kt
annahay4: Yes

rust-runtime/aws-smithy-runtime/temp_test.rs
annahay4: Yes - didn't see this snuck in :(

rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
annahay4: I see what you're saying! Thank you for explaining more, I'll change this

rust-runtime/aws-smithy-runtime/src/client/retries/strategy/standard.rs
```diff
@@ -51,14 +51,28 @@ impl StandardRetryStrategy {
         Default::default()
     }
 
-    fn release_retry_permit(&self) -> ReleaseResult {
+    fn release_retry_permit(&self, token_bucket: &TokenBucket) -> ReleaseResult {
         let mut retry_permit = self.retry_permit.lock().unwrap();
         match retry_permit.take() {
             Some(p) => {
-                drop(p);
+                // Retry succeeded: reward success and forget permit if configured, otherwise release permit back
```
aajtodd, line 58: I think this logic made more sense outside of `release_retry_permit`. Adding it here overloads the function's responsibility.

annahay4: That makes sense to me - I can update this

aajtodd:
Closing as this was merged separately and released in https://github.com/smithy-lang/smithy-rs/releases/tag/release-2025-12-16

(that release will be updated by today's release with changes from https://github.com/smithy-lang/smithy-rs/pull/4461)

PR #4403, opened 2025-11-18
Expose methods to indicate if token bucket is full or empty

annahay4:
Make public methods to indicate if token bucket is full or empty.

## Motivation and Context
Customers may want to adjust other behaviors if their token bucket is depleted or full. This allows them to get signal on this.

## Description
Made two new public methods: is_token_bucket_empty and is_token_bucket_full. Each returns a boolean.

## Testing
Ran cargo test on this package

## Checklist

----


rust-runtime/aws-smithy-runtime/src/client/retries/token_bucket.rs
```diff
@@ -87,6 +87,16 @@ impl TokenBucket {
     pub(crate) fn available_permits(&self) -> usize {
         self.semaphore.available_permits()
     }
+
+    /// Returns true if the token bucket is full, false otherwise
+    pub fn is_token_bucket_full(&self) -> bool {
```
aajtodd: Why not expose available_permits instead? I believe a previous commit did that. 

Also I'd suggest `is_full` and `is_empty`, the methods belong to `TokenBucket` so we don't need `token_bucket` in the name.

annahay4: I am planning to merge this into main, because we are working on another feature that relies on these calls being available, and will need this functionality regardless of the exact token bucket implementation.

As for the available_permits, I wanted a caller to be able to get the info on if the bucket is empty/full without needing to know what capacity the bucket was configured with. I know that a customer wouldn't specifically need is_empty to be implemented over available_permits, but it seemed cleaner to expose these as a pair.

PR #4406, opened 2025-11-19
generate and check rust-version for Smithy packages

arielb1:
## Motivation and Context

Previously to this CR, many Smithy packages, including <https://crates.io/crates/aws-config>, did not declare their MSRV. This causes problems because the Cargo MSRV-aware resolver is by design not fully backtracking.

For example, the current version of aws-config (1.8.10) depends on aws-sdk-sts 1.90.0, which has a pretty high MSRV (1.88). Because of how the MSRV-aware resolver works, as long as aws-config does not declare an MSRV, everyone using rustc < 1.88 with an `aws-config = "1"` dependency will pick aws-config 1.8.10 (because the resolver is eager), and then fail to compile because all aws-sdk-sts versions that work with 1.8.10 don't work with rustc < 1.88.

This means that everyone using Smithy and updating their Cargo.lock had to use rustc 1.88, which is quite annoying churn, especially for libraries that want to have a fixed MSRV.

If aws-config 1.8.10 had a correctly-configured MSRV, then for example cargo using rustc 1.87 would pick aws-config 1.8.9, which can use aws-sdk-sts 1.89, which would keep working.

I don't think we should yank aws-config 1.8.10 as all users have already updated their rustc, but this will prevent this problem from happening during the next MSRV bump.

## Description

Added a script that uses `cargo msrv` to verify MSRVs and changed MSRVs to obey it.

## Testing

This does not change any code. The MSRV is tested by the CI script.

## Checklist

----

### DIFF
--- a/tools/ci-scripts/check-rust-runtimes
+++ b/tools/ci-scripts/check-rust-runtimes
@@ -4,6 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+
+C_RED='\033[1;31m'
 C_YELLOW='\033[1;33m'
 C_RESET='\033[0m'
 
@@ -15,6 +17,16 @@ cd smithy-rs
 
 echo -e "# ${C_YELLOW}Auditing runtime crate version numbers...${C_RESET}"
 runtime-versioner audit --no-fetch
+
+# Compute MSRV from channel, ignoring patch version
+MAIN_MSRV=$(sed -rn 's/channel = "(1[.][0-9]+)([.][0-9]+)?"/\1/p' rust-toolchain.toml)
+
+if [ -z "$MAIN_MSRV" ]; then
+    echo -e "# ${C_RED}Unable to detect MSRV in rust-toolchain.toml${C_RESET}"
+    exit 1
+else
+    echo -e "# ${C_YELLOW}MSRV is ${MAIN_MSRV}${C_RESET}"
+fi
 
 for runtime_path in \
     "rust-runtime" \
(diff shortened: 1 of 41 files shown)


landonxjames:
I think most of this change makes sense. My one worry is the inconsistent `rust-version` tags in each crate. Not sure how you settled on the current `rust-version` in each one, but presumably it is the minimum viable version? If we are going to do this I think we should align all of those with the current workspace MSRV in `rust-toolchain.toml`. 

Would also like to see some tooling for updating this. Right now our MSRV bumps are a manual process and have to change it in ~5 places. This substantially increases that burden and makes it more likely we would miss an update somewhere. It seems like the CI update would catch that, but the feedback loop there is pretty slow.

arielb1:
> My one worry is the inconsistent rust-version tags in each crate. Not sure how you settled on the current rust-version in each one, but presumably it is the minimum viable version? If we are going to do this I think we should align all of those with the current workspace MSRV in rust-toolchain.toml.

I used `cargo msrv` to compute the MSRV.

I was just worried that setting the workspace MSRV could be too high. Would you be fine with making the workspace MSRV be the MSRV anywhere?

tools/ci-scripts/check-rust-runtimes
```diff
@@ -41,6 +53,14 @@ do
     cargo +"${RUST_NIGHTLY_VERSION}" minimal-versions check --direct --all-features
 
     for crate_path in *; do
+        if [[ -f "${crate_path}/Cargo.toml" ]]; then
+            # verify that the MSRV is the main MSRV
+            package_msrv=$(sed -rn 's/rust-version = "([^"]+)"/\1/p' "${crate_path}/Cargo.toml")
+            if [[ "${package_msrv}" != "${MAIN_MSRV}" ]]; then
+                echo -e "## ${C_RED}MSRV mismatch in ${crate_path} - toolchain MSRV is ${MAIN_MSRV}, crate MSRV is ${package_msrv}${C_RESET}"
```
landonxjames, line 60: Nit: For errors I generally think it is a better experience to aggregate all of them and then fail rather than failing when the first one is encountered. Lets the developer see all of the issues and fix them in one update rather than going back and forth with the CI. 

Maybe set a bool if an error is encountered and only `exit 1` if that bool is true after the full loop so the MSRV issue is printed for all of the crates before exiting?

landonxjames (approved):
One nit with the CI script, not blocking.

Issue #4413, opened 2025-11-20
NoAuthRuntimePlugin is extremely confusing


rcoh:
As a user, I frequently want to take a client and simply disable auth – maybe to talk to a local version of a service, maybe the service has a different auth mechanism, etc.

This is possible today, but it requires writing your own RuntimePlugin. While this on its own is fine, it is severely compounded by the fact that there is a `NoAuthRuntimePlugin` but _It Does Not Work_. (More specifically, adding it does not actually disable authentication).

Recommendations:
- I believe that plugin is probably in use. It is basically never what a user would want, consider making it doc-hidden
- Provide a new `NoAuthRuntimePlugin` that works properly:

```rust
impl<R> ResolveAuthSchemeOptions for NoAuthAuthSchemeOptionResolver<R>
where
    R: ResolveAuthSchemeOptions,
{
    fn resolve_auth_scheme_options_v2<'a>(
        &'a self,
        params: &'a AuthSchemeOptionResolverParams,
        cfg: &'a ConfigBag,
        runtime_components: &'a RuntimeComponents,
    ) -> AuthSchemeOptionsFuture<'a> {
        let inner_future =
            self.inner
                .resolve_auth_scheme_options_v2(params, cfg, runtime_components);
 
        AuthSchemeOptionsFuture::new(async move {
            let mut options = inner_future.await?;
            options.push(AuthSchemeOption::from(NO_AUTH_SCHEME_ID));
            Ok(options)
        })
    }
}
```

ysaito1001:
Released in [release-2026-01-14](https://github.com/smithy-lang/smithy-rs/releases/tag/release-2026-01-14)

PR #4415, opened 2025-11-20
Add `NoAuthRuntimePluginV2` and deprecate `NoAuthRuntimePlugin`

ysaito1001:
## Motivation and Context
https://github.com/smithy-lang/smithy-rs/issues/4413

## Description
[NoAuthRuntimePlugin](https://github.com/smithy-lang/smithy-rs/blob/646f151bb1470c2e22820cb0d1ceb8a87245af49/rust-runtime/aws-smithy-runtime/src/client/auth/no_auth.rs#L31) was broken after aligning the auth implementation with SRA. 

After the SRA alignment, the code generator follows models more faithfully; it injects `noAuth` into the default auth scheme resolver when an operation is annotated with `@optionalAuth` or `@auth([])`. However, this broke `NoAuthRuntimePlugin` since it wasn't actually configuring the auth scheme option resolver.

While fixing `noAuth` issues in the originating Smithy model is ideal (by adding [appropriate auth traits](https://smithy.io/2.0/spec/authentication-traits.html#optionalauth-trait)),  that may not always be possible, e.g. when a Smithy-convertible model doesn't support it natively. 

This PR deprecates the existing `NoAuthRuntimePlugin` and introduces `NoAuthRuntimePluginV2` that configures the auth scheme option resolver for `noAuth`.

**Note**:
`NoAuthRuntimePlugin` is registered as a default client runtime plugin in `FluentClientGenerator.kt`. This was a hack from before the auth implementation was aligned with SRA. The plugin continues to be registered by default for backward compatibility, but it's broken due to the lack of an appropriate auth scheme option resolver. `NoAuthRuntimePluginV2` is not added by default and should only be used as an escape hatch. The ideal fix is to annotate operations with
the appropriate auth trait in the Smithy model whenever possible.

While users could implement the functionality of `NoAuthRuntimePluginV2` themselves, the multiple attempts have been seen in the wild (with the confusion), and some customers cannot update their models for various reasons. It's reasonable to provide this utility to avoid repeated issues.

## Testing
Added integration test in `AuthDecoratorTest.kt`

Ignore Semver compliance check since `NoAuthRuntimePlugin` has been `#[doc(hidden)]`, and the check couldn't get the information out of it.

## Checklist

----

### DIFF
--- a/aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
+++ b/aws/codegen-aws-sdk/src/main/kotlin/software/amazon/smithy/rustsdk/AwsFluentClientDecorator.kt
@@ -60,6 +60,7 @@ class AwsFluentClientDecorator : ClientCodegenDecorator {
                     AwsPresignedFluentBuilderMethod(codegenContext),
                     AwsFluentClientDocs(codegenContext),
                     AwsFluentClientRetryPartition(codegenContext),
+                    AwsFluentClientEnableRetries(codegenContext),
                 ).letIf(codegenContext.serviceShape.id == ShapeId.from("com.amazonaws.s3#AmazonS3")) {
                     it + S3ExpressFluentClientCustomization(codegenContext)
                 },
@@ -193,6 +194,28 @@ private class AwsFluentClientRetryPartition(private val codegenContext: ClientCo
                         *preludeScope,
                         "Cow" to RuntimeType.Cow,
                     )
+                }
+            }
+            else -> emptySection
+        }
+    }
+}
+
+/**
+ * Enables retries by default for AWS SDK clients when awsSdkBuild is true.
+ *
+ * Only applies to real AWS SDK builds (controlled by awsSdkBuild in JSON settings).
+ */
+private class AwsFluentClientEnableRetries(private val codegenContext: ClientCodegenContext) : FluentClientCustomization() {
+    override fun section(section: FluentClientSection): Writable {
+        if (!codegenContext.sdkSettings().awsSdkBuild) {
+            return emptySection
+        }
+
+        return when (section) {
+            is FluentClientSection.CustomizeDefaultPluginParams -> {
+                writable {
+                    rust(".with_is_aws_sdk(true)")
                 }
             }
             else -> emptySection
--- a/aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/HttpChecksumTest.kt
+++ b/aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/HttpChecksumTest.kt
@@ -456,6 +456,7 @@ internal class HttpChecksumTest {
                         .region(Region::from_static("doesntmatter"))
                         .with_test_defaults()
                         .http_client(http_client)
+                        .retry_config(#{RetryConfig}::disabled())
                         .build();
 
                     let client = $moduleName::Client::from_conf(config);
@@ -489,6 +490,7 @@ internal class HttpChecksumTest {
                 "tokio" to CargoDependency.Tokio.toType(),
                 "capture_request" to RuntimeType.captureRequest(rc),
                 "http_1x" to CargoDependency.Http1x.toType(),
+                "RetryConfig" to RuntimeType.smithyTypes(rc).resolve("retry::RetryConfig"),
             )
         }
     }
--- a/aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/TimeoutConfigMergingTest.kt
+++ b/aws/codegen-aws-sdk/src/test/kotlin/software/amazon/smithy/rustsdk/TimeoutConfigMergingTest.kt
@@ -65,6 +65,7 @@ class TimeoutConfigMergingTest {
                         let operation = Duration::from_secs(4);
                         let http_client = infallible_client_fn(|_req| http::Response::builder().body(SdkBody::empty()).unwrap());
                         let sdk_config = SdkConfig::builder()
+                            .behavior_version(aws_smithy_runtime_api::client::behavior_version::BehaviorVersion::v2026_01_12())
                             .timeout_config(
                                 TimeoutConfig::builder()
                                     .connect_timeout(connect_timeout)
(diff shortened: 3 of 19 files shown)


ysaito1001:
> Is it because `ResolveAuthScheme` is specific to each generated crate?

Yes, that's for one; While we can code generate a custom no-auth resolver for them, it can only wrap the default auth scheme resolver, not the one dynamically configured in runtime components. For two, we can narrow the usage of `NoAuthRuntimePluginV2` only to generic clients since the `.runtime_plugin` method is only exposed to them but not to AWS SDK. A custom no auth scheme resolver would be available to both generic and AWS SDK clients.

rcoh:
Really? I thought we enabled runtime plugins for the SDK...

ysaito1001:
> Really? I thought we enabled runtime plugins for the SDK...

[case in point](https://github.com/smithy-lang/smithy-rs/blob/main/aws/sdk/build.gradle.kts#L104), though it's configurable at codegen time

ysaito1001:
@rcoh, do you still feel strongly about original recommendations or do you consider us doc-hiding `NoAuthRuntimePlugin` and only exposing no auth scheme resolver, knowing the non-functional `NoAuthRuntimePlugin` still registers no auth scheme in the orchestrator?

FYI, we kind of have a [NoAuthSchemeResolver](https://github.com/smithy-lang/smithy-rs/blob/d38a6c58c2c1746d3a22f2832f0b583b910826e8/codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthSchemeResolverGenerator.kt#L214-L237), but its usage is restricted to protocol tests.

rcoh:
I don't really care the exact mechanism. But:
1. Having a `NoAuthRuntimePlugin` that doesn't work on its own is very confusing.
2. The errors you get don't point you at all in the right direction:
```
called `Result::unwrap()` on an `Err` value: DispatchFailure(DispatchFailure { source: ConnectorError { kind: Other(None), source: NoMatchingAuthSchemeError(ExploredList { items: [ExploredAuthOption { scheme_id: AuthSchemeId { scheme_id: "sigv4" }, result: NoIdentityResolver }], truncated: false }), connection: Unknown } })
```

Having a well documented cookbook about "here's how you disable auth" would probably do the trick that is linked from that error message.

The `NoAuthSchemeResolver` doesn't do the right thing since it needs to compose with the existing resolver.

If we wanted to, you could consider adding a method to `config` like `allow_no_auth` to the config that automates this process.

ysaito1001:
Sounds good. Taking both @aajtodd and @rcoh's input into account, I'll proceed with:
- Defining `allow_no_auth` on the client config builder — I prefer this over `disable_auth` since the latter implies the no-auth scheme takes highest priority, while `allow_no_auth` better conveys that it's appended to the end of the auth scheme resolver's list.
- Adding a discussion page that describes `allow_no_auth` and linking to it from the error message.
- Doc-hiding the existing `NoAuthRuntimePlugin`.

rcoh:
Actually one more consideration—we need to make sure that a runtime plugin can enable this or there is another ay for a service to vend a "working" client to customers which often involves doing some amount of custom settings. That's one thing that's nice about the existing RuntimePlugin approach — its easy for someone to provide a wrapper around the generated client that adds the runtime plugin (because a runtime plugin is always addable).

ysaito1001:
> Actually one more consideration—we need to make sure that a runtime plugin can enable this or there is another ay for a service to vend a "working" client to customers which often involves doing some amount of custom settings. That's one thing that's nice about the existing RuntimePlugin approach — its easy for someone to provide a wrapper around the generated client that adds the runtime plugin (because a runtime plugin is always addable).

OK, I was thinking to remove `NoAuthRuntimePluginV2` and instead code generate `pub(crate)` counterpart used privately by `allow_no_auth` but sounds like we still want to keep `NoAuthRuntimePluginV2` as-is in the PR, exposed as `pub` from the aws-smithy-runtime crate.

codegen-client/src/main/kotlin/software/amazon/smithy/rust/codegen/client/smithy/auth/AuthDecorator.kt
```diff
@@ -281,6 +283,22 @@ private class AuthDecoratorConfigCustomizations(private val codegenContext: Clie
                             self.runtime_components.set_auth_scheme_option_resolver(#{Some}(auth_scheme_resolver.into_shared_resolver()));
                             self
                         }
+
+                        /// Add [NoAuthScheme](aws_smithy_runtime::client::auth::no_auth::NoAuthScheme) as a fallback for operations that don't require authentication
+                        ///
+                        /// The auth scheme resolver will use this when no other auth schemes are applicable.
```
rcoh: this comment is a little confusing. Does the operation need to be modeled with noauth for this to work?

I would probably write this in a way that is more focused on the customer impact of enabling this instead as the top line that references internal details separately

ysaito1001: 👍 

> Does the operation need to be modeled with noauth for this to work

No, it doesn't need to, which is the whole point of this API, used as an escape hatch to enable noauth even when the operation isn't modeled as such.

ysaito1001: Updated in fd7388d

rcoh (approved):

rust-runtime/aws-smithy-runtime/src/client/auth/no_auth.rs
```diff
@@ -59,6 +71,92 @@ impl RuntimePlugin for NoAuthRuntimePlugin {
...
+                    NO_AUTH_SCHEME_ID,
+                    SharedIdentityResolver::new(NoAuthIdentityResolver::new()),
+                )
+                .with_auth_scheme(SharedAuthScheme::new(NoAuthScheme::new())),
+        )
+    }
+}
+
+impl RuntimePlugin for NoAuthRuntimePluginV2 {
+    fn runtime_components(
```
rcoh, line 110: I think this needs to use order = nestedComponents

ysaito1001: Good catch. Makes sense since it wraps existing runtime components to append no auth scheme at the end. Fixed in 80fcc1e.

landonxjames (approved):

